PR #118, opened 2018-05-25
feat: add basic support for helm charts

edvald:
Implementation note: We're not actually using helm to perform the
install, since it proved unreliable and opaque. We instead use helm
to render the templates and provide an array of specs that we feed
to kubectl apply. This also avoids the need for installing tiller,
which will be removed in Helm 3 either way. We may reconsider this
approach when Helm 3 is out and stable.


edvald:
Found a bug after rebase-ing, looking into it

edvald:
This npm package-lock issue will be the end of me

PR #139, opened 2018-06-02
fix: compatibility issue with Node 10

Deepakkothandan:


edvald:
Hey @Deepakkothandan and thanks for the PR! This is the first outside PR we receive, exciting times :)

I do have a couple of comments and questions though, will submit a quick review.

package.json
```diff
@@ -107,11 +107,14 @@
...
     "tmp-promise": "^1.0.4",
-    "ts-node": "^6.0.3",
+    "ts-node": "^6.0.5",
     "tslint": "^5.10.0",
     "tslint-microsoft-contrib": "^5.0.3",
     "typescript": "^2.8.3",
     "typescript-formatter": "^7.2.0",
+    "typescript-node": "^0.1.3",
+    "typescript-register": "^1.1.0",
+    "typescript-require": "^0.2.10",
```
edvald: These have been deprecated in favor of `ts-node` and would quite possibly conflict as well. Did you run into problems with `ts-node`?

src/util/index.ts
```diff
@@ -29,7 +29,9 @@ import chalk from "chalk"
 import hasAnsi = require("has-ansi")
 
 // shim to allow async generator functions
-(<any>Symbol).asyncIterator = (<any>Symbol).asyncIterator || Symbol.for("Symbol.asyncIterator")
+if (typeof (Symbol as any).asyncIterator === "undefined") {
+  (Symbol as any).asyncIterator = Symbol.asyncIterator || Symbol("asyncIterator")
+}
```
edvald: Can you explain what this fixes?

Also, it would be good to have these separate issues in separate comments (but same PR is fine). Makes it easier to understand from the git history.

package.json
Deepakkothandan: @edvald I am using a mac with nvm. npm install of garden-cli was resulting in failure both for node version 8 and 10.

So, I cloned the repo to run it directly. When I was running `gulp tsc/watch`,  tsc tried to load these packages. Even after installing ts-node, gulp was still trying to require these packages and without installing these packages, I was not able to get the compiler working.

src/util/index.ts
Deepakkothandan: After resolving the packages, I had the following error
```
(<any>Symbol).asyncIterator = (<any>Symbol).asyncIterator || Symbol.for("Symbol.asyncIterator")
                           ^
TypeError: Cannot assign to read only property 'asyncIterator' of function 'function Symbol() { [native code] }'
    at Object.<anonymous> (/Users/deepak/devel/node/garden/src/util/index.ts:32:28)```

Since the property is readonly, I changed the code to only assign it when it doesn't exist.

edvald:
Got it! That's very helpful. Seems there are three separate issues at play here: 1) the npm package manifest is incorrect, 2) gulp isn't playing nice with Typescript on your setup, I'd like to look at that in more detail, and 3) the read-only error in `util/index.ts`. 

Your fix for the last one seems good, am just gonna test it locally quickly to validate when I'm back at my laptop.

For the gulp issue, would you mind trying to upgrade your global `gulp` package and/or installing `ts-node` globally to see if the problem persists? It appears gulp tries more than one different way to load `gulpfile.ts` if it encounters it, and I suspect the issue comes up if you're using the global `gulp` and not the local one installed as a dev dependency.

edvald:
Fixed the `garden-cli` npm package here: #140

src/util/index.ts
edvald: Ok, issue confirmed, appears this came up in Node 10 which we don't yet automatically test against (but should). Could you remove the redundant `Symbol.asyncIterator || ` from `(Symbol as any).asyncIterator = Symbol.asyncIterator || Symbol("asyncIterator")` and defer the gulp issue to another PR? Then we can get this merged and push a patch release.

Deepakkothandan:
@edvald Thank you for the fix, I was trying to use `build/src/bin/garden.js` for the bin directive, that's why I could not get it to work. New thing learnt today :)

edvald:
Tested on Node 8 and 10, seems 👍

Mind if I rename the PR to "Fix Node.js 10 compatibility issue"?

Deepakkothandan:
@edvald Feel free to rename the PR. I will check on the gulp dependency issue and update you ASAP

Deepakkothandan:
@edvald I remove the node installation, reinstalled node, and tsc compiled successfully, without requiring the deprecated dependencies. I believe it has something to do with sourcing the bash profile for nvm. So, I am not creating an issue/PR for that. Thank you

PR #143, opened 2018-06-03
chore: switch to commitlint to validate commit messages

edvald:


eysi09:
Do we want to change the commit message max length back to 100? Seems commitlint defaults to 72 characters whereas Angular had 100 IIRC.

edvald:
I actually thought it was 72 before (had Sourcetree configured that way). So I don't really mind either way, would also be fine with 100.

eysi09:
Let's just stick to the defaults then.

PR #147, opened 2018-06-05
fix(vsc): handle weird stat behavior by wrapping it

eysi09:
Looks like a bug in fs or fs-extra.

Fails silently: `await Bluebird.map(["a"], el => el).map(stat)`
Works: `await Bluebird.map(["a"], el => el).map(el => stat(el))`

Where the `stat` above is imported from `fs-extra`.


eysi09:
I've opened an issue for this in the `fs-extra` repo: https://github.com/jprichardson/node-fs-extra/issues/591

PR #152, opened 2018-06-06
fix(cli): enforce single character option aliases

### LINKED ISSUE #151: garden cli fails when using -log global option but not when using --loglevel
When running garden with the `-log` option, garden fails with the below error, but it works when using `--loglevel`

```
node ~/devel/node/garden/dist/src/bin/garden.js status -log verbose

USAGE

  garden status [options]

Global options
  -h, --help        Show help
                    [boolean]
....

  -log, --loglevel  Set logger level.
                    [enum] [default: info] [error, warn, info, verbose, debug, silly]


  -o, --output      Output command result in specified format (note: disables progress
                    logging).
                    [enum] [json, yaml]

Value "" is invalid for argument o or output. Choices are: json, yaml
```

eysi09:
Fixes #151

### DIFF
--- a/src/cli/cli.ts
+++ b/src/cli/cli.ts
@@ -76,7 +76,7 @@ export const GLOBAL_OPTIONS = {
   }),
   env: new EnvironmentOption(),
   loglevel: new ChoicesParameter({
-    alias: "log",
+    alias: "l",
     choices: enumToArray(LogLevel),
     help: "Set logger level.",
     defaultValue: LogLevel[LogLevel.info],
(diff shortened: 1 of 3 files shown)


src/cli/cli.ts
```diff
@@ -300,6 +300,11 @@ export class GardenCli {
 }
 
 export async function run(): Promise<void> {
-  const cli = new GardenCli()
-  return cli.parse().then(result => shutdown(result.code))
+  try {
+    const cli = new GardenCli()
+    return cli.parse().then(result => shutdown(result.code))
```
edvald: I think it'd make sense to change this line along the way to
```
const result = await cli.parse()
shutdown(result.code)
```

eysi09: Good catch! I threw in a finally statement there as well.

PR #157, opened 2018-06-07
feat(build): Handle config changes in auto-reload.

thsig:
Before this change, each module directory was watched separately for
changes during auto-reload, and didn't properly handle module- or
project-level configuration changes since the auto-reload watch was
started.

Here, we consolidate the module-level watches into a single
project-level watch, and call the originating command with a fresh
PluginContext when configuration changes. This includes the addition or
removal of modules during the lifetime of the watch.

### DIFF
--- a/src/commands/base.ts
+++ b/src/commands/base.ts
@@ -13,6 +13,7 @@ import {
 import { PluginContext } from "../plugin-context"
 import { TaskResults } from "../task-graph"
 import { LoggerType } from "../logger/types"
+import { ProcessResults } from "../process"
 
 export class ValidationError extends Error { }
 
@@ -128,6 +129,7 @@ export interface CommandConstructor {
 
 export interface CommandResult<T = any> {
   result?: T
+  restartRequired?: boolean
   errors?: GardenError[]
 }
 
@@ -172,19 +174,24 @@ export abstract class Command<T extends Parameters = {}, U extends Parameters = 
 }
 
 export async function handleTaskResults(
-  ctx: PluginContext, taskType: string, result: TaskResults,
+  ctx: PluginContext, taskType: string, results: ProcessResults,
 ): Promise<CommandResult<TaskResults>> {
-  const failed = Object.values(result).filter(r => !!r.error).length
+  const failed = Object.values(results).filter(r => !!r.error).length
 
   if (failed) {
     const error = new RuntimeError(`${failed} ${taskType} task(s) failed!`, {
-      result,
+      results,
     })
     return { errors: [error] }
-  } else {
-    ctx.log.info("")
+  }
+
+  ctx.log.info("")
+  if (!results.restartRequired) {
     ctx.log.header({ emoji: "heavy_check_mark", command: `Done!` })
-    return { result }
+  }
+  return {
+    result: results.taskResults,
+    restartRequired: results.restartRequired,
   }
 }
 
--- a/src/plugin-context.ts
+++ b/src/plugin-context.ts
@@ -20,7 +20,6 @@ import {
   Garden,
 } from "./garden"
 import { EntryStyle } from "./logger/types"
-import { TaskResults } from "./task-graph"
 import {
   PrimitiveMap,
   validate,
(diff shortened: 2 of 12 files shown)


src/commands/base.ts
```diff
@@ -172,7 +173,7 @@ export abstract class Command<T extends Parameters = {}, U extends Parameters =
 }
 
 export async function handleTaskResults(
-  ctx: PluginContext, taskType: string, result: TaskResults,
+  ctx: PluginContext, taskType: string, result: TaskResults, restartRequired: boolean,
```
edvald: This could also accept a `ProcessResult`, tiny bit cleaner.

src/plugin-context.ts
```diff
@@ -477,44 +491,69 @@ export function createPluginContext(garden: Garden): PluginContext {
...
+          if (restartNeeded) {
+            /*
+              In case a change with configChanged === false occurs after a change with configChanged === true,
+              but within the same 1000 msec window in the while loop below.
+            */
+            return
+          }
+
+          ctx.log.debug({ msg: `Files changed for modules ${changedModules.map(m => m.name).join(", ")}` })
+          restartNeeded = await ctx.handleChanges(autoReloadDependants, process, changedModules, configChanged)
```
edvald: Shouldn't we stop the watcher here if `restartNeeded===true`?

src/plugin-context.ts
```diff
@@ -477,44 +491,69 @@ export function createPluginContext(garden: Garden): PluginContext {
...
       }
+
+      return {
+        taskResults: {}, // TODO: Return latest results for each task baseKey processed between restarts?
+        restartRequired: true,
+      }
+    },
+
+    // Returns true if the command that requested the watch needs to be restarted.
+    handleChanges: async (
```
edvald: Does this belong on the `ctx` object? Feels rather specific to be on the internal API.

src/plugin-context.ts
```diff
@@ -477,44 +491,69 @@ export function createPluginContext(garden: Garden): PluginContext {
...
+
         })
 
       registerCleanupFunction("clearAutoReloadWatches", () => {
         watcher.end()
         ctx.log.info({ msg: "\nDone!" })
       })
 
-      while (true) {
+      while (!restartNeeded) {
```
edvald: This could be done differently, by wrapping the above code in a `new Promise((resolve, reject) => ...`, returning that and resolving when `restartNeeded===true`.

src/watch.ts
```diff
@@ -165,6 +196,29 @@ export class FSWatcher {
     })
   }
 
+  processChangedFiles(files: ChangedFile[], relModuleRoots: RelativeModuleRoots): ProcessedChanges {
```
edvald: Should this be `private`?

edvald:
A few comments, but overall looks 👌

src/plugin-context.ts
thsig: Good point, looks like I forgot to actually stop the watcher. Haha.

src/plugin-context.ts
thsig: Yeah, good point. I'll move this method and `processModules` / `processServices`) to another module.

src/watch.ts
thsig: Yes indeed.

src/plugin-context.ts
thsig: Good point, will look into that.

src/types/KeyedSet.ts
```diff
@@ -0,0 +1,51 @@
+/*
```
edvald: I'd suggest putting this under `src/util` and sticking to lowercase for filenames

src/watch.ts
```diff
@@ -102,7 +137,7 @@ export class FSWatcher {
       this.client.command(args, (error: Error, result: object) => {
         if (error) {
           // TODO: Error logging
-          console.error(error)
+          console.error("Error while executing watcher.command, args were", args, "error is:", error)
```
edvald: Any reason not to use our lovely logger?

thsig: No reason at all — will make the change now.

PR #161, opened 2018-06-08
docs: add a simple project walkthrough

eysi09:


docs/examples/simple-project.md
```diff
@@ -0,0 +1,315 @@
...
+        ├── app.js
+        ├── main.js
+        ├── package.json
+        └── test
+            └── integ.js
+
+5 directories, 7 files
+ ```
+
+As you can see the project consists of two super simple services and their accompanying Dockerfiles. One of the core tenets of multi-service backends is being able to pick the right tool for the job, and therefore we have a Node.js service and a Golang service, that we can pretend solve for us specific problems.
```
edvald: "... that we can pretend solve for us specific problems" sounds slightly odd, but I like how you're framing it. Maybe just restructure the sentence slightly?

docs/examples/simple-project.md
```diff
@@ -0,0 +1,315 @@
...
+project:
+  name: simple-project
+  global:
+    providers:
+      - name: container
+  environments:
+    - name: local
+      providers:
+        - name: local-kubernetes
+    - name: dev
```
edvald: Maybe we skip the `dev` bit for the example? It feels like an advanced use-case and we don't support it super well at the moment.

docs/examples/simple-project.md
```diff
@@ -0,0 +1,315 @@
...
+```sh
+$ touch garden.yml
+```
+
+and add the following configuration:
+
+```yaml
+project:
+  name: simple-project
+  global:
```
edvald: This key is not necessary I think. (If it is, that's a bug that we need to fix).

docs/examples/simple-project.md
```diff
@@ -0,0 +1,315 @@
...
+  environments:
+    - name: local
+      providers:
+        - name: local-kubernetes
+    - name: dev
+      providers:
+        - name: google-app-engine
+```
+
+Above, we've specified the name of our project, a global provider, and an [environment](../other/glossary.md#Environment)-specific provider override.
```
edvald: This doesn't seem super clear. Maybe by making the changes I suggested above, we can simplify this as well. Something like "... and configured the project to use the `local-kubernetes` plugin for local development."

docs/examples/simple-project.md
```diff
@@ -0,0 +1,315 @@
...
+    versionString: 2c8818986d-1528373640
+    latestCommit: 2c8818986d
+    dirtyTimestamp: 1528373640
+```
+
+Under the `module` directive we can now specify how to run our service:
+
+```yaml
+module:
+  description: Go service container
```
edvald: -> `Node service container`

docs/examples/simple-project.md
```diff
@@ -0,0 +1,315 @@
...
+Calling our `go-service` from our `node-service` is straightforward from within the application code. Crack open `services/node-service/app.js` with your favorite editor and add the following:
+
+```javascript
+const request = require('request-promise')
+
+// Host and port are exposed as environment variables
+const {
+  GO_SERVICE_SERVICE_HOST: host,
+  GO_SERVICE_SERVICE_PORT: port,
+} = process.env
```
edvald: This syntax may seem a little foreign to some users. Also, does this actually work? The auto-generated environment variables should always have a `GARDEN_` prefix.

docs/examples/simple-project.md
```diff
@@ -0,0 +1,315 @@
...
+
+```yaml
+module:
+  description: Node service container
+  ...
+  services:
+    - name: node-service 
+      command: [npm, start]
+    ...
+    dependencies:
```
edvald: Wrong indent here.

docs/getting-started/README.md
```diff
@@ -64,14 +64,14 @@ Once you have the dependencies set up, simply run:
 
 ## Using the CLI
 
-With the CLI installed, we can now try out a few commands using the [hello-world](https://github.com/garden-io/garden/tree/master/examples/hello-world) example from this repository. The example consists of a container service that runs an [Express](http://expressjs.com/) app, a serverless function, and an npm library package.
+With the CLI installed, we can now try out a few commands using the [hello-world](https://github.com/garden-io/examples/tree/master/examples/hello-world) project from our Github [examples repository](https://github.com/garden-io/examples). The example consists of a container service that runs an [Express](http://expressjs.com/) app, a serverless function, and an npm library package.
```
edvald: Would it make sense to call the repo `garden-examples` instead of just `examples`?

docs/getting-started/README.md
```diff
@@ -64,14 +64,14 @@ Once you have the dependencies set up, simply run:
...
 
 _Note: check if Kubernetes is running with `kubectl version`. You should see both a `Client Version` and a `Server Version` in the response. If not, please start it up before proceeding._
 
-Clone the repo and change into the `examples/hello-world`  directory:
+Clone the repo and change into the `hello-world`  directory:
 
-    git clone https://github.com/garden-io/garden.git &&
-    cd garden/examples/hello-world
+    git clone https://github.com/garden-io/examples.git &&
```
edvald: The `&&` isn't necessary here, I think?

docs/getting-started/README.md
```diff
@@ -85,8 +85,4 @@ And that's it! The services are now running on the Garden framework. You can see
 
     garden call hello-container/hello
 
-Check out our [Commands](../guides/commands.md) guide for other features like auto-reload, streaming service logs, running tests, and lots more. 
-
-## What's next
-
-Kick the tires of our [examples](https://github.com/garden-io/garden/tree/master/examples/hello-world) to get a feel for how projects are configured, or head to the [Guides](../guides/README.md) section for a deep dive into the Garden framework.
+Check out our [Commands](../guides/commands.md) guide for other features like auto-reload, streaming service logs, running tests and lot's more, or see how a Garden project is configured from scratch in our [Simple Project](../guides/simple-project.md) guide.
```
edvald: "lot's" -> "lots"

docs/examples/simple-project.md
eysi09: It does work. However, I wasn't 100% sure how this should be handled in general. Is this the way to go about it? Also, the `GARDEN_` prefix seems to missing, the only env var I got with that prefix was: `GARDEN_SERVICES_GO_SERVICE_HOST`. Is that a bug on our end or did I mess up the config?

RE the syntax, good point, I'll fix that.

docs/examples/simple-project.md
eysi09: Good catch!

docs/getting-started/README.md
eysi09: Yes, I think it does. We might have a bunch of things under the `garden-io` organisation in the future. Should we take it a step further and call it `garden-cli-examples` or `garden-framework-examples`?

docs/getting-started/README.md
eysi09: Whoops

docs/examples/simple-project.md
eysi09: The name key?

eysi09: Or the project key?

docs/examples/simple-project.md
eysi09: Agreed. I'll remove the `global` and `dev` directives and update the text. This example should really be the bare minimum.

docs/getting-started/README.md
eysi09: ...changed it to `garden-examples` for now.

docs/examples/simple-project.md
```diff
@@ -0,0 +1,307 @@
...
+* [Configure the project](#project-wide-configuration)
+* [Configure individual modules](#module-configuration)
+* [Deploy the project locally](#deploying)
+* [Have the services communicate with one another](#inter-service-communication)
+* [Manage service dependencies](#dependencies)
+* [Test services](#testing)
+
+The code for this tutorial can be found on Github in our [examples repo](https://github.com/garden-io/garden-examples). We'll use the `simple-project-raw` branch and work our way from there.
+
+First, let's clone the examples repo, checkout to the `simple-project-raw` branch and take a look inside:
```
edvald: It feels more correct to say "checkout a branch" or even "switch to", is that just me?

docs/examples/simple-project.md
```diff
@@ -0,0 +1,307 @@
...
+5 directories, 7 files
+ ```
+
+As you can see the project consists of two super simple services and their accompanying Dockerfiles. One of the core tenets of multi-service backends is being able to pick the right tool for the job, and therefore we have a Node.js service and a Golang service, that we can pretend have different responsibilities.
+
+The task at hand is to configure these services so that they can run on the Garden framework.
+
+## Project-wide configuration
+
+To begin with, every project needs a project-wide `garden.yml` [configuration file](../guides/configuration.md#Config) at the root level. There we define, among other things, the name of the project, and the [providers](../other/glossary.md#Provider) used for each [plugin](../other/glossar.md#Plgun) the project requires.
```
edvald: I'm guessing `Plgun` in the last link is a typo?

docs/examples/simple-project.md
```diff
@@ -0,0 +1,307 @@
...
+    - name: node-service 
+      command: [npm, start]
+      ports:
+        - name: http
+          containerPort: 8080
+      endpoints:
+        - paths: [/]
+          port: http
+```
+The [services](../guides/configuration.md#Services) directive defines the services exposed by the module, namely our containerized Node.js server. The sub-directives tell Garden how to start the service and which endpoints to expose.
```
edvald: Maybe add a small note on this being specific to container modules? Maybe not necessary here, but I somehow feel that's worth emphasizing in the docs.

docs/examples/simple-project.md
```diff
@@ -0,0 +1,307 @@
...
+
+## Deploying
+
+With this configuration we're almost ready to deploy. First, we'll need to create a user namespace for our environment with the login command:
+
+```sh
+$ garden login
+# CLI output...
```
edvald: Is this meant as a placeholder? If it is, then it would be good to show the actual output, but if not it feels unnecessary.

docs/examples/simple-project.md
```diff
@@ -0,0 +1,307 @@
...
+      endpoints:
+        - paths: [/]
+          port: http
+```
+
+Run the deploy command again, this time only for the `go-service`:
+
+```sh
+$ garden deploy go-service
+# CLI output...
```
edvald: Same note as ^

docs/examples/simple-project.md
```diff
@@ -0,0 +1,307 @@
...
+```
+
+Run the deploy command again, this time only for the `go-service`:
+
+```sh
+$ garden deploy go-service
+# CLI output... 
+```
+
+Another way to verify that our services are up and running is to have a look at the service logs. We can either get an aggregate from all our services or pass the command a list of services. This time we're only interested in our `go-service`:
```
edvald: Suggestion: "We can either get an aggregate from all our services, _by running `garden logs`,_ or we can specify a list of services."

docs/examples/simple-project.md
```diff
@@ -0,0 +1,307 @@
...
+## Inter-service communication
+
+Calling our `go-service` from our `node-service` is straightforward from within the application code. Crack open `services/node-service/app.js` with your favorite editor and add the following:
+
+```javascript
+const request = require('request-promise')
+
+// Host and port are available as environment variables
+const goServiceHost = process.env.GO_SERVICE_SERVICE_HOST;
+const goServicePort = process.env.GO_SERVICE_SERVICE_PORT;
```
edvald: I figured out the issue with this - these environment variables are specific to k8s. So this will indeed work but can be either configured more explicitly, or simplified:

```javascript
const goServiceEndpoint = `http://go-service/`;
```

Because the way we've set this up, the k8s service discovery always uses the actual service name as a hostname, and the port is the default in this case.

docs/examples/simple-project.md
eysi09: It's a new Garden primitive.

eysi09:
Updated as per @edvald's comments above.

PR #182, opened 2018-06-25
fix: module versions are now handled properly across multiple repos

edvald:
This refactor resolves issues with using plugin modules as dependencies,
and is important for multi-repo support. Basically we now hash versions
together instead of using the latest commit, which would of course only
work within a single repo.

One thing to look at here are how the versions are hashed together (see `src/vcs/base.ts`). Please let me know if you think it could/should be done differently.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,6 @@ node_modules
 
 # Runtime files
 .garden
-.garden-version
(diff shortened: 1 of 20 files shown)


eysi09:
Looks good to me.

PR #185, opened 2018-06-27
Better env var support in config

edvald:


examples/hello-world/services/hello-container/garden.yml
```diff
@@ -16,6 +16,8 @@ module:
           port: http
       dependencies:
         - hello-function
+      env:
+        FUNCTION_ENDPOINT: ${dependencies.hello-function.outputs.endpoint}
```
eysi09: Should this be documented somewhere? The `outputs.endpoint` feels kind of cryptic?

eysi09:
Should change the following in the `hello-world/hello-container/app.js` example from `const functionEndpoint = process.env.GARDEN_SERVICES_HELLO_FUNCTION_ENDPOINT` to `const functionEndpoint = process.env.FUNCTION_ENDPOINT`?

Issue #171, opened 2018-06-14
Move package root to subdirectory


edvald:
I found out the hard way that lerna doesn't really support having a package in the repo root. I figure we'll also have multiple packages in the repo at some point, so I'm thinking we should move all the stuff we end up packaging and releasing as `garden-cli` to a `garden-cli` subdirectory.

We'd keep some stuff like `bin`, `docs` and some config stuff in the root. Figure it'd also make the top-level directory a bit cleaner, it's a bit crowded at this point.

Any objections...?

thsig:
Sounds good to me.

PR #186, opened 2018-06-27
Release 0.2.0

edvald:


edvald:
Our scripts still need improvement, but I'll get back to it after we do #171

PR #187, opened 2018-06-30
fix: fixed more issues with cross-repo versioning

edvald:
Turns out I missed a beat on the last fix. Hopefully this is the rest
of it.

To explain a bit - the .garden-version files need to contain the tree version for the directory, not the module version. Otherwise the mechanism falls apart, I should have realized that in testing.


gulpfile.ts
```diff
@@ -87,37 +88,44 @@ function die() {
...
+        console.error(msg + "\n" + outputWithError)
+        return reject(msg)
+      }
+
+      Bluebird.map(<any>results.result, async (module) => {
+        const path = (<any>module).path
+        const relPath = relative(__dirname, path)
+        const versionFilePath = join(__dirname, relPath, ".garden-version")
+
+        const treeVersion = await execa.stdout("ts-node", [join("src", "vcs", "git.ts"), path])
```
eysi09: I'm a little bit confused by this. Why execute the script like this, rather than just importing some function from the module that can return the tree version?

edvald: Yeah, I've just wanted to avoid importing directly from the code in general, partly because the gulp process then keeps the same code even when you make changes to the sources. Although in this case it's maybe not necessary, since the functions in question are fairly stable/static... Yeah maybe I'll just change it.

PR #189, opened 2018-07-01
feat: add create commands. wip

eysi09:
A command for creating a new garden project with modules, or for creating individual modules within a project. A couple of things I was hoping to get feedback on:

1. In `cli/cli.ts`, the `MOCK_CONFIG` object is used to initialize `garden` since there's no `garden.yml` file in new projects. Is there a better way around this?

2. Config templates for new projects/modules are created in `src/commands/configTemplates`. Should this be something the plugins themselves expose?


ellenkorbes:
Minor nit, but can we rename `clit.ts` to something else?

edvald:
Haha, that's thankfully a typo.

eysi09:
Haha, whoops. Fixed the text.

docs/reference/commands.md
```diff
@@ -74,6 +74,68 @@ Note: Currently only supports HTTP/HTTPS endpoints.
...
+input.
+
+Examples:
+
+    garden create project # creates a new Garden project in the current directory (project name defaults to
+    directory name)
+    garden create project my-project # creates a new Garden project named my-project in the current directory
+    garden create project --module-dirs=services,more-services
+    # creates a new Garden project and looks for modules in the services and more-services directories
+    garden create project my-project --new  # creates a new Garden project in a new directory named my-project
```
edvald: How about instead having a `--path` option, that defaults to `.` and attempts to create the directory if it doesn't exist? That kinda covers both cases nicely, and has more explicit behavior.

docs/reference/commands.md
```diff
@@ -74,6 +74,68 @@ Note: Currently only supports HTTP/HTTPS endpoints.
...
+| Argument | Alias | Type | Description |
+| -------- | ----- | ---- | ----------- |
+  | `--module-dirs` |  | string | Relative path to modules directory. Use comma as a separator to specify multiple directories
+  | `--new` |  | boolean | If true, creates a new directory. Otherwise assumes current working directory is the project directory
+
+### garden create module
+
+Creates scaffolding for a new Garden project.
+
+The Create command walks the user through setting up a new Garden project and generates scaffolding based on user
```
edvald: Copy/pasted from the `garden project` docs?

docs/reference/commands.md
```diff
@@ -74,6 +74,68 @@ Note: Currently only supports HTTP/HTTPS endpoints.
...
+
+| Argument | Required | Description |
+| -------- | -------- | ----------- |
+  | `moduleName` | No | The name of the module, (defaults to current directory name)
+
+##### Options
+
+| Argument | Alias | Type | Description |
+| -------- | ----- | ---- | ----------- |
+  | `--new` |  | boolean | If true, creates a new directory. Otherwise assumes current working directory is the module directory
```
edvald: Same as above, feels to me like a `--path` option would work better

docs/reference/commands.md
```diff
@@ -74,6 +74,68 @@ Note: Currently only supports HTTP/HTTPS endpoints.
...
+  | `--type` |  | string | Type of module. Check out &#x27;https://docs.garden.io&#x27; for available types
```
edvald: Might need to change the escaping in the template to allow the backticks (in Handlebars `{{something}}` HTML-escapes, `{{{something}}}` (triple braces) doesn't escape).

src/commands/base.ts
```diff
@@ -145,6 +145,7 @@ export abstract class Command<T extends Parameters = {}, U extends Parameters =
   arguments?: T
   options?: U
 
+  runWithoutConfig: boolean = false
```
edvald: Maybe `runWithoutProject` or `noProject`?

src/commands/create/project.ts
```diff
@@ -0,0 +1,195 @@
...
+      errors.push(err)
+    }
+
+    if (errors.length === 0) {
+      task.setSuccess()
+    } else {
+      task.setWarn({ msg: "Finished with errors", append: true })
+    }
+
+    ctx.log.info("Project created! Be sure to check out our docs at `https://docs.garden.io`")
```
edvald: Do backticks have the desired effect in the console?

eysi09: I think that was just suppose to be a single quote. But better yet, we can use this: https://github.com/sindresorhus/terminal-link

docs/reference/commands.md
eysi09: Yeah I thought about that but abandoned it because it kind of conflicted with the global `--root` option. But you're right though, it is more explicit.

So if, hypothetically, someone where to pass in both `--root` and `--path` we'd just join those two together right?

Also, if there's already a directory at `--path` we just initialise the project inside that dir as opposed to throwing an error?

edvald: Yeah that's what I figure. You can resolve the path, with the root directory as the base. If the user provides an absolute path, then that also works. 

And yes - if the directory exists, use that, if it's not a directory, fail, otherwise create the directory.

eysi09: Actually, the second item is a bit fuzzy, that's also why I went for explicitly stating whether it's a new project or not.

E.g. if you try to git clone into a non-empty directory you get an error. So a user who _wants_ to create a new project and thus uses the `--path` option wouldn't probably expect it to fail if there's already something at that path.

However, the name of the option kind of makes it sound like you might just as well be initialising the project in some existing directory at the given path.

Perhaps it would then make more sense to have something like:
```sh
garden create [path/to/dir] [--modules-dirs] [--name]
```
So pretty much same is git.

edvald: Hmm, right, I don't think the same semantics as git apply because you may be creating a project where your existing sources are.

Just to be clear, would you then default to creating the project in the current directory? And the path option is just a positional arg instead of an option?

eysi09: I got some feedback from @ellenkorbes regarding this and we decided that the git-like approach above would make sense, in particular since the name is already optional and just defaults to the current directory.   

If there's already a directory at the path—which is always the case when the path isn't specified, namely the current directory)—we assume that that's intentional and add the configs anyway. There's already logic in place that ensure existent `garden.yml` files don't get overwritten so it should be harmless.

eysi09: Whoops, somehow missed your comment above. But yes, you would default to creating the project in the current directory. So you have your project, and you want to "gardenify" it. Again, kind of like how you do `git init` (or e.g. `npm init`) within an existing project. 

Since project names aren't required and default to the root dir name we figured it would make sense to have the path (in most cases probably just the name of the directory) as the argument. 

So basically:

```sh
# I want to turn my project into a garden project
garden create project

# I want to start a new garden project from scratch
garden create project my-project
```

src/commands/create/project.ts
```diff
@@ -93,19 +93,31 @@ export class CreateProjectCommand extends Command<typeof createProjectArguments,
...
+    if (args.projectDir) {
+      projectDir = validate(args.projectDir.trim(), joiIdentifier(), {context: "project directory" })
+      projectRoot = join(projectRoot, projectDir)
+      if (!(await pathExists(projectRoot))) {
+        try {
+          await mkdir(projectRoot)
+        } catch ({ message }) {
+          throw new ParameterError(`Unable to make directory at ${projectRoot}`, { message })
+        }
+      }
```
eysi09: Here we can replace the `pathExists` and `mkdir` calls with `ensureDir`. 

Also the error should bubble up and is caught in the `cli` so it boils down to just `ensureDir(projectRoot)`.

src/commands/create/project.ts
```diff
@@ -93,19 +93,31 @@ export class CreateProjectCommand extends Command<typeof createProjectArguments,
...
-      args.projectName ? args.projectName.trim() : parse(projectRoot).base,
-      joiIdentifier(),
-      { context: "project name" },
-    )
+    let projectName: string
+    if (opts.name) {
+      projectName = opts.name
+    } else {
+      projectName = projectDir
+    }
```
eysi09: This could be reduced to just: `const projectName = opts.name || projectDir`

Issue #101, opened 2018-05-22
Switch from watchman to chokidar


edvald:
Watchman is good for performance but we want the basic installation process to be more straightforward. We can use https://github.com/paulmillr/chokidar, which is performant and well maintained (it's used by VSCode among others).

thsig:
I looked into replacing Watchman with chokidar in our FS watching logic, should be relatively straightforward.

thsig:
Similarly, it shouldn't be too hard to support both of them.

edvald:
As discussed AFK, we're going to abandon watchman in favor of Chokidar, which is well supported and portable. Renaming the issue accordingly.

PR #192, opened 2018-07-03
refactor: remove dependency on watchman

thsig:
Instead of using FB's watchman library to monitor FS changes in watch
commands, we now use the pure-JS library chokidar.


thsig:
This PR resolves https://github.com/garden-io/garden/issues/101.

src/watch.ts
```diff
@@ -117,131 +92,144 @@ async function uniqueDependencyModules(module: Module): Promise<Module[]> {
...
-        if (error) {
-          this.ctx.log.error(`Error while executing watcher.command, args were ${args}, error is: ${error}`)
-          rej(error)
-        }
+    const onFileChanged = this.makeFileChangedHandler(modules, changeHandler)
 
-        res(result)
-      })
+    this.watcher = watch(projectRoot, {
+      ignored: /(^|[\/\\])\../, // just ignore dotfiles for now
```
edvald: Does this implicitly ignore files in `.gitignore`? If not we should probably add that, and could add `.gardenignore` support along the way (which would work the same).

src/watch.ts
```diff
@@ -117,131 +92,144 @@ async function uniqueDependencyModules(module: Module): Promise<Module[]> {
...
+
+  }
+
+  private invalidateCached(module: Module) {
+    // invalidate the cache for anything attached to the module path or upwards in the directory tree
+    const cacheContext = pathToCacheContext(module.path)
+    this.ctx.invalidateCacheUp(cacheContext)
+
+    this.ctx.invalidateCache(cacheContext)
+    this.ctx.invalidateCacheDown(cacheContext)
```
edvald: Hmm, I don't think you want to call all three of these?

src/watch.ts
```diff
@@ -117,131 +92,144 @@ async function uniqueDependencyModules(module: Module): Promise<Module[]> {
...
+
+      let changedModule: Module | null = null
+
+      for (const module of modules) {
 
-      const changedModuleName = findKey(relModuleRoots,
-        (relPath) => f.name.startsWith(relPath))
+        if (module.path.startsWith(dirPath)) {
+          // at least one module's root dir was removed
+          await this.invalidateCachedForAll()
```
edvald: Why invalidate for all?

src/watch.ts
```diff
@@ -117,131 +92,144 @@ async function uniqueDependencyModules(module: Module): Promise<Module[]> {
...
-    }
+      if (basename(filePath) === "garden.yml") {
+        await this.invalidateCachedForAll()
+        return changeHandler(null, true)
+      }
+
+      const changedModule = modules.find(m => filePath.startsWith(m.path)) || null
+
+      if (changedModule) {
+        await this.invalidateCachedForAll()
```
edvald: Why invalidate for all?

src/watch.ts
thsig: Ah, this is a leftover from when I was debugging that I seem to have forgotten to clean up (same for the other two below, https://github.com/garden-io/garden/pull/192#discussion_r199767084 & https://github.com/garden-io/garden/pull/192#discussion_r199767466.

I'll fix these now.

Issue #89, opened 2018-05-08
Use numeric log levels instead of (or in addition to) named levels?


edvald:
Just a thought. I for one always get confused as to which is more verbose, "verbose" or "debug". And I can imagine us wanting to set multiple levels at some point.

For example, in `kubectl` you can set `--v=8` to dump a lot of logs. We could for example alias the named levels to the numeric levels (0=error, 1=warning, 2=info etc.), and perhaps make `l` an alias for `--loglevel` to make it more concise.

eysi09:
Fixed in #194

PR #194, opened 2018-07-03
feat: allow numeric log levels (#89)

eysi09:


src/cli/cli.ts
```diff
@@ -77,8 +89,8 @@ export const GLOBAL_OPTIONS = {
   env: new EnvironmentOption(),
   loglevel: new ChoicesParameter({
     alias: "l",
-    choices: enumToArray(LogLevel),
-    help: "Set logger level.",
+    choices: logLevelChoices,
+    help: "Set logger level, values can be either string or numeric",
```
edvald: Would it make sense to give a couple of examples to see how numbers map to the string names? (e.g. `-l=debug` == `-l=?`)

eysi09: Good point. I'll add those.

PR #202, opened 2018-07-10
docs: add missing generated commands docs

eysi09:
Not sure what happened but the generated docs were out of sync.


edvald:
Hmm yea, might have happened in a rebase. We could use a slightly better process for this, maybe validating specifically in the CI.

eysi09:
Also wondering if the `generate-docs` script should be a part of the pre push step and not when `watch` is running.

PR #204, opened 2018-07-10
fix(create-module-command): type option should be an enum

eysi09:
Looks like another rebase regression. This is how it should've been in the first place.


docs/reference/commands.md
```diff
@@ -136,7 +136,7 @@ Examples:
 | Argument | Alias | Type | Description |
 | -------- | ----- | ---- | ----------- |
   | `--name` |  | boolean | Assigns a custom name to the module. (Defaults to name of the current directory.)
-  | `--type` |  | string | Type of module. Check out &#x27;https://docs.garden.io&#x27; for available types
+  | `--type` |  | `container` `function` `npm-package`  | Type of module.
```
edvald: Just noticed - `function` is not a valid module type.

PR #208, opened 2018-07-12
refactor(k8s): make deployment and status checks much more robust

edvald:
This fully removes the unused `kubernetes-specs` module type and in
turn leverages Helm properly. Additionally several improvements were
made to rollout status checks, which now support most standard k8s
objects.


garden-cli/src/plugins/kubernetes/api.ts
```diff
@@ -124,3 +136,56 @@ function proxyApi<T extends Core_v1Api | Extensions_v1beta1Api | RbacAuthorizati
     },
   })
 }
+
+export async function apiReadBySpec(namespace: string, context: string, spec: KubernetesObject) {
+  // this is just awful, sorry. any better ideas? - JE
```
thsig: My first thought was something along these lines (a lookup table instead of a `switch`) to remove some of the repetition, but I'm not sure it's any clearer:
```javascript
const readMethods = {
  "ConfigMap":             {api: "core", name: "readNamespacedConfigMap"},
  "Endpoints":             {api: "core", name: "readNamespacedEndpoints"},
  "LimitRange":            {api: "core", name: "readNamespacedLimitRange"},
  "PersistentVolumeClaim": {api: "core", name: "readNamespacedPersistentVolumeClaim"},
  "Pod":                   {api: "core", name: "readNamespacedPod"},
  "PodTemplate":           {api: "core", name: "readNamespacedPodTemplate"},
  "ReplicationController": {api: "core", name: "readNamespacedReplicationController"},
  "ResourceQuota":         {api: "core", name: "readNamespacedResourceQuota"},
  "Secret":                {api: "core", name: "readNamespacedSecret"},
  "Service":               {api: "core", name: "readNamespacedService"},
  "ServiceAccount":        {api: "core", name: "readNamespacedServiceAccount"},
  "DaemonSet":             {api: "ext",  name: "readNamespacedDaemonSet"},
  "Deployment":            {api: "ext",  name: "readNamespacedDeployment"},
  "Ingress":               {api: "ext",  name: "readNamespacedIngress"},
  "ReplicaSet":            {api: "ext",  name: "readNamespacedReplicaSet"},
  "StatefulSet":           {api: "apps", name: "readNamespacedStatefulSet"},
  "Role":                  {api: "rbac", name: "readNamespacedRole"},
  "RoleBinding":           {api: "rbac", name: "readNamespacedRoleBinding"},
}

const apiInitMethods = {
  core: coreApi,
  ext: extensionsApi,
  apps: appsApi,
  rbac: rbacApi,
}

export async function apiReadBySpec(namespace: string, context: string, spec: KubernetesObject) {
  const name = spec.metadata.name
  const readMethod = readMethods[spec.kind]

  if (readMethod) {
    return apiInitMethods[readMethod.api](context)[readMethod.name](name, namespace)
  } else {
    throw new ConfigurationError(`Unsupported Kubernetes spec kind: ${spec.kind}`, {
      spec,
    })
  }

}
```

edvald: Yeah, the problem with this is in the type-safety, which is made fairly difficult by how they lay out the API library (which is not idiomatic in terms of JS/TS). So I think we should let this slide for now.

edvald: But your idea is indeed how I'd prefer the API to be laid out.

PR #213, opened 2018-07-17
docs: main README for src folder

ellenkorbes:
This is just the first stub for our docs. Definitely not final.


edvald:
@ellenkorbes we should close this one, right?

ellenkorbes:
Yeah. Moving it into the new docs.

PR #217, opened 2018-07-24
feat: add Windows support (experimental)

edvald:
I tried testing this best I could, but it is possible there are
more issues lurking. If anyone can give this a proper spin
(on an actual Windows machine and not in a VM) that would be
most appreciated!


thsig:
I'm not sure I have access to a Windows machine, but I'll do a "read-only" pass through this for starters.

edvald:
Yeah I mean, that's kinda the best we can do at the moment. We'll call for some feedback from our users as well.

docs/introduction/getting-started.md
```diff
@@ -37,9 +37,61 @@ brew tap garden-io/garden
...
+Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Hyper-V -All
+```
+
+This will require a restart.
+
+#### Step 2: Docker for Windows
+
+Install the Edge version of [Docker for Windows](https://www.docker.com/docker-windows).
+
+Once installed, open the Docker for Mac preferences, go to the Kubernetes section,
```
thsig: Mac -> Windows

edvald: Good catch, fixed.

garden-cli/src/plugins/kubernetes/api.ts
```diff
@@ -33,8 +34,7 @@ const configs: { [context: string]: KubeConfig } = {}
 
 function getConfig(context: string): KubeConfig {
   if (!kubeConfigStr) {
-    kubeConfigStr = readFileSync(process.env.KUBECONFIG || join(process.env.HOME || "/home", ".kube", "config"))
-      .toString()
+    kubeConfigStr = readFileSync(join(homedir(), ".kube", "config")).toString()
```
drubin: We should still allow `KUBECONFIG` here as an override. Many users use this to allow them to separate contexts by file. 

Why don't we also fix the node-client to also support loading from windows? https://github.com/kubernetes-client/javascript/blob/master/node-client/src/config.ts#L299 or just fix them to support windows?

edvald: Ah crap, yeah I didn't intend to remove that. And that's a good idea, yes.

PR #219, opened 2018-07-30
Multi repos

eysi09:


garden-cli/src/commands/link.ts
```diff
@@ -0,0 +1,94 @@
...
+  help = "Link a local directory to your Garden project"
+  arguments = linkArguments
+
+  description = dedent`
+    After linking a module, Garden will read the source from the it's local directory as opposed to from the remote uri.
+    Only available for modules that haven an external source.
+
+    Examples:
+
+        garden link module-a,module-b path/to/module-a,path/to/module-b
```
edvald: Is there a good reason to allow multiple links in one command? This seems a little contrived.

garden-cli/src/commands/link.ts
```diff
@@ -0,0 +1,94 @@
...
+export type LinkArguments = ParameterValues<typeof linkArguments>
+
+export class LinkCommand extends Command<typeof linkArguments> {
+  name = "link"
+  help = "Link a local directory to your Garden project"
+  arguments = linkArguments
+
+  description = dedent`
+    After linking a module, Garden will read the source from the it's local directory as opposed to from the remote uri.
+    Only available for modules that haven an external source.
```
edvald: I'm thinking we may want to call this command `link module` because it specifically applies to that use case, whereas we'll likely want a separate command to link project level sources.

garden-cli/src/commands/link.ts
eysi09: Yeah but I figured it was consistent with the other CLI commands. Usually when the command accepts module or service names as arguments the user can also supply a list of names.

garden-cli/src/commands/link.ts
eysi09: Good point.

eysi09: Given that, should the key in the garden config be `moduleSources` (or `externalModules`) instead of `sources`?

Alternatively we could stick to a single key and a single command and deduce from the imported config whether it's a project or a module. From a user perspective I think having a single command would be nice. That being said, it might make sense to have different config keys to keep things more explicit.

examples/hello-world-ext-sources/garden.yml
```diff
@@ -0,0 +1,20 @@
...
+  environments:
+    - name: local
+      providers:
+        - name: local-kubernetes
+        - name: local-google-cloud-functions
+    - name: dev
+      providers:
+        - name: google-app-engine
+        - name: google-cloud-functions
+          default-project: garden-hello-world
```
edvald: I'd suggest calling this example just `ext-sources` and removing the `dev` environment and the `environmentDefaults` to keep it simple

examples/hello-world-ext-sources/libraries/hello-npm-package/garden.yml
```diff
@@ -0,0 +1,4 @@
+module:
+  description: Hello world npm package
+  type: npm-package
+  name: hello-npm-package
```
edvald: This is probably also unnecessary for the purposes of this example

examples/hello-world-ext-sources/garden.yml
eysi09: Ah yes, I forgot that example in there. Was just using it on my end to test things out. But I'll add a super simple ext-sources example for integration testing.

garden-cli/src/garden.ts
```diff
@@ -592,27 +600,39 @@ export class Garden {
...
+      const path = await this.loadExtSourcePath({ name, repositoryUrl, sourceType: "project" })
+      extSourcePaths.push(path)
+    }
 
-      const parsedPath = parse(item.path)
+    const dirsToScan = [this.projectRoot, ...extSourcePaths]
 
-      if (parsedPath.base !== MODULE_CONFIG_FILENAME) {
-        continue
+    for (const dir of dirsToScan) {
```
edvald: Could be better to use `Bluebird.map` here for concurrency

garden-cli/src/garden.ts
```diff
@@ -592,27 +600,39 @@ export class Garden {
...
+        if (parsedPath.base !== MODULE_CONFIG_FILENAME) {
+          continue
+        }
 
-      modulePaths.push(parsedPath.dir)
+        modulePaths.push(parsedPath.dir)
+      }
     }
 
     for (const path of modulePaths) {
```
edvald: Same applies to my old code here actually (to use `Bluebird.map`)

garden-cli/src/garden.ts
```diff
@@ -749,6 +767,28 @@ export class Garden {
...
+
+    const linkedSources = await getLinkedSources(this.localConfigStore, sourceType)
+
+    const linked = findByName<LinkedSource>(linkedSources, name)
+
+    if (linked) {
+      return linked.path
+    }
+
+    const path = await this.vcs.ensureRemoteSource({ name, sourceType, url: repositoryUrl, logger: this.log })
```
edvald: Maybe better to use the same naming as for the plugin handlers for `logEntry`?

garden-cli/src/types/config.ts
```diff
@@ -6,22 +6,25 @@
...
 import * as yaml from "js-yaml"
 import { readFile } from "fs-extra"
-import { defaultEnvironments, ProjectConfig, projectSchema } from "./project"
+import { defaultEnvironments, ProjectConfig, projectSchema, SourceConfig } from "./project"
 import { omit } from "lodash"
+import { Garden } from "../garden"
 
 const CONFIG_FILENAME = "garden.yml"
 
+export const getRelPath = (projectRoot: string, path: string) => relative(projectRoot, join(path, CONFIG_FILENAME))
```
edvald: This name is a bit broad for what it actually does (wouldn't matter for a local function, but worth mentioning because it's exported)

garden-cli/src/types/config.ts
```diff
@@ -143,3 +141,31 @@ export async function loadConfig(projectRoot: string, path: string): Promise<Gar
...
 }
+
+export async function loadProjectConfig(projectRoot: string): Promise<GardenConfig> {
+  const spec = await readConfig(projectRoot)
+  return resolveConfig(spec, projectRoot, projectRoot)
+}
+
+export async function loadModuleConfig(
+  path: string,
+  gardenCtx: Garden,
```
edvald: `gardenCtx` seems like the wrong name here?

garden-cli/src/util/util.ts
```diff
@@ -394,6 +400,10 @@ export function findByName<T extends ObjectWithName>(array: T[], name: string):
   return find(array, ["name", name])
 }
 
+export function onlyRemote(module: Module): boolean {
```
edvald: Seems like a rather specific function to have here?

garden-cli/src/vcs/git.ts
```diff
@@ -71,63 +71,72 @@ export class GitHandler extends VcsHandler {
...
-      } else if (latestCommit === NEW_MODULE_VERSION) {
-        return 0
-      } else {
-        // clean versions are sorted by their commit distance from HEAD
-        return await this.getOffsetFromHead(latestCommit)
-      }
-    }
-    let positions = {}
+  // TODO Wrap the cli around this process so that we can add spinners while cloning and other delights
+  async ensureRemoteSource({ url, name, logger, sourceType }: {
```
edvald: This should be added as an abstract method on the base class as well

package.json
```diff
@@ -56,5 +56,9 @@
       "pre-push": "npm run check-package-lock && npm test && npm run integ"
     }
   },
-  "snyk": true
+  "snyk": true,
+  "dependencies": {
+    "deepmerge": "^2.1.1",
+    "parse-github-url": "^1.0.2"
```
edvald: These should be in `garden-cli/package.json` I think

edvald:
A few comments but overall looks very good. I'll take some time to properly try this out tomorrow/Monday and see if more feedback comes up.

One thing that we definitely need though, is more documentation. I'd suggest writing a guide specifically for this, and mentioning that guide in some of the introductory docs as well.

Great work, this'll be a really good addition to the framework.

eysi09:
Yes, agreed! Just wanted to get this out before moving on to the documentation. This functionality could also make for a great example project. I was thinking about combining this with the `create` command and writing a guide for something like the [Sock shop microservice demo](https://github.com/microservices-demo/microservices-demo).

I'll also follow up with a PR for integration tests for the git functionality.

garden-cli/src/types/config.ts
eysi09: Yes, the `export` was a mistake.

eysi09:
@edvald All comments have been resolved.

garden-cli/src/config-store.ts
```diff
@@ -6,29 +6,32 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import { resolve } from "path"
-import { ensureFile, readFile } from "fs-extra"
 import * as Joi from "joi"
 import * as yaml from "js-yaml"
+import keymirror = require("keymirror")
```
edvald: Is this a necessary library to add?

garden-cli/src/garden.ts
```diff
@@ -592,33 +602,49 @@ export class Garden {
...
+          continue
+        }
+
+        paths.push(parsedPath.dir)
       }
 
-      modulePaths.push(parsedPath.dir)
-    }
+      return paths
+    })).reduce((acc, val) => acc.concat(val), []).filter(Boolean) // Flatten and remove empty
```
edvald: There's also a `flatten` helper in lodash that I often use

garden-cli/src/vcs/git.ts
```diff
@@ -71,63 +69,55 @@ export class GitHandler extends VcsHandler {
...
 
-    return this.repoRoot
+    return fullPath
   }
 
-  private async git(args, cwd?: string) {
-    if (!cwd) {
-      cwd = this.projectRoot
+  async updateRemoteSource({ url, name, sourceType, logEntry }: RemoteSourceParams) {
+    const { branch } = parseGithubUrl(url)
```
edvald: Is this actually Github specific, or is the name of the library confusing?

garden-cli/test/src/commands/unlink.ts
```diff
@@ -0,0 +1,111 @@
...
+  removeLocalConfig,
+} from "../../helpers"
+import { PluginContext } from "../../../src/plugin-context"
+import { LinkSourceCommand } from "../../../src/commands/link/source"
+import { UnlinkSourceCommand } from "../../../src/commands/unlink/source"
+
+describe("UnlinkCommand", () => {
+
+  describe("UnlinkModuleCommand", () => {
+
```
edvald: Just a very small nit, but I notice you sometimes put a linebreak after the start of a function - is there some pattern in it? Just tickles my OCD a bit.

eysi09: Just random line breaks I'm afraid. Will fix.

eysi09:
Added fixes for the comments above in a separate commit to make it easier to review the changes. I'll squash the whole thing once it's ready.

PR #238, opened 2018-08-18
chore: update to TypeScript 3.0

edvald:


edvald:
Interesting. After this change the first test takes a long time to run. Same happens locally, except it barely goes under the timeout. I'll investigate.

edvald:
Ok, worked it out. @thsig can you review?

PR #239, opened 2018-08-18
[WIP] Support TLS and custom domains/subdomains (take 2)

edvald:
Posting this for early review on the semantics (again).

Some notes on the implementation:
* The notion of domains/subdomains is gone. The biggest impact is that we no longer (and actually _can't_ anymore) automatically assign subdomains to services, instead defaulting to always use the same default hostname.
* In turn, hostnames can now be set arbitrarily by the user on individual services.
* TLS certificates are still configured at the plugin level. We may want to have some official framework-level notion of certs later, but I didn't find an immediate reason or an obvious design for that just yet.

Still TO-DO (pending design approval):
* Testing
* Update examples
* Configurable TLS for OpenFaaS plugin
* Extended validation on routing rules, e.g. warn when rules conflict.
* We may want to support path rewriting, now that we have the same hostname by default on all services.

cc @drubin


edvald:
In the absence of design comments, I think I'll press on. Let me know if you find issues with the spec as it stands.

drubin:
@edvald I think the high-level design makes sense.

I still feel using the domain's could lead to a leaky abstraction :/ but I think its the best choice given, where we are and being able to prove this is useful and valuable.

edvald:
Hmm how do you mean, using the domains? We just configure hostnames and certs now.

Note that with this approach we could still add a feature to explicitly pick a certificate for an endpoint, but I decided against it as part of this PR because we'd need to consider certs as a primitive in the framework right away.

drubin:
@edvald Exactly! 

I am saying that until we do eventually make certificates a primitive it might feel a little missing, but something we can improve on later.

edvald:
Ok cool, then we're on the same page :)

edvald:
I'm thinking I'll pull out the incomplete tasks (provider docs, OpenFaaS TLS etc.) into a separate PR, so I'm calling this one ready for review.

examples/local-tls/README.md
```diff
@@ -0,0 +1,57 @@
+# Local TLS example project
```
drubin: I Love this. I think every example project should have this.

edvald: Agreed! I plan on doing exactly that

examples/local-tls/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+```sh
+kubectl create secret tls tls-garden-dev --key garden.dev+1-key.pem --cert garden.dev+1.pem
+```
+
+_The filenames above will be different if you used a different hostname._
+
+### Step 4 - Configure the hostname in your hosts file
+
+Add the `garden.dev` hostname to the hosts file on your machine, and have it point to the IP of your local cluster.
```
eysi09: Maybe add that _having a hostname point to a IP_ translates to this in the user's hostfile (assuming *nix):
```127.0.0.1 garden.dev```

Comment (edvald) [reply examples/local-tls/README.md]: Actually the IP will vary when using minikube, which is partly why I opted to not get into that, and suggest using a tool.

File: garden-cli/src/plugins/kubernetes/api.ts [hunk 3]
```diff
@@ -95,30 +69,121 @@ export class KubernetesError extends GardenBaseError {
...
+  /**
+   * Wrapping the API objects to deal with bugs.
+   */
+  private proxyApi<T extends K8sApi>(api: T, config): T {
+    api.setDefaultAuthentication(config)
+
+    return new Proxy(api, {
+      get: (target: T, name: string, receiver) => {
+        if (!(name in Object.getPrototypeOf(target))) { // assume methods live on the prototype
+          return Reflect.get(target, name, receiver)
```
eysi09: Not an issue per se but I'm just a bit confused by all this.

eysi09:
Just read this over and the code looks good to me. However, I did notice that if I set the hostname at the service level I get the following error:

```
✖ Sending HTTP GET request to http://hostname-test/hello

getaddrinfo ENOTFOUND hostname-test hostname-test:80
```
Am I doing something wrong here? If that's the case I guess the error could be more helpful.

garden-cli/src/plugins/kubernetes/api.ts
edvald: Yeah, this is some magicky ES6 stuff, which I actually first added many PRs ago. Essentially this wraps an object, keeping its original shape. Those two lines at the top of the get handler make it  a passthrough for anything that isn't a method (i.e. doesn't do anything when accessing properties). It's something I got off a couple of examples I found.

edvald:
Is `hostname-test` in your hosts file?

edvald:
We're just showing the standard error message, which simply indicates that hostname couldn't be found. I guess we could wrap that particular message to make it clearer and provide suggestions, but it'll vary case-by-case whether the correct suggestion is to add something to the hosts file, fix the configured hostname, fix some plugin configuration etc.

eysi09:
It was not. Works after I add it 👍 

Is this an error we should try to catch though and add something like "Are you missing a host file entry"?

edvald:
See my comment above - we couldn't really just suggest that, because that's very specific to the `local-kubernetes` setup (and I don't see us delegating the call or the error handling to plugins for now). So it'd have to be a somewhat elaborate message, which I'm not entirely sure is a good idea, plus I don't feel we should put too much into the `garden call` helper since it's usefulness is fairly limited by its nature.

eysi09:
Missed your comment above. But yeah, perhaps wrapping the error is tricky. Maybe add something regarding this to the description with the hostname key?

edvald:
That's a good idea, will do!

PR #240, opened 2018-08-21
Delete service command

eysi09:


eysi09:
PR updated with respect to the comments above.

PR #242, opened 2018-08-24
feat: allow commands from subdirectories (WIP)

### LINKED ISSUE #136: Enable running garden commands from subdirectories in a project directory
Currently you can only run from the root directory of a project. We should add a step when starting up to walk up the directory tree from the current working directory and find the closest `garden.yml` that has a `project` key.

The logic for this should probably be in the `Garden.factory()` method, since it will involve parsing `garden.yml` files.

ellenkorbes:
Closes #136.


garden-cli/src/garden.ts
```diff
@@ -185,37 +186,46 @@ export class Garden {
...
         throw new ConfigurationError(`Supplied config does not contain a project configuration`, {
-          projectRoot,
+          currentDirectory,
           config,
         })
       }
     } else {
-      config = await loadConfig(projectRoot, projectRoot)
-      parsedConfig = await resolveTemplateStrings(config, new ProjectConfigContext())
+      let pathArray = currentDirectory.split(require("path").sep)
```
eysi09: Rather than `require("path").sep`, you should import it at the top of the file like so: `import { sep } from "path"`. 

In this case `sep` has already been imported so this could just be `split(sep)`

garden-cli/src/garden.ts
```diff
@@ -185,37 +186,46 @@ export class Garden {
...
+      for (let i = 0; i < pathArray.length; i++) {
+        config = await loadConfig(currentDirectory, currentDirectory)
+        if (!config) {
+          pathArray.pop()
+          pathArray.unshift("/")
+          currentDirectory = join.apply(null, pathArray)
```
eysi09: Not a big deal but in Imo it's cleaner to do `join(...pathArray)`.

garden-cli/src/garden.ts
```diff
@@ -185,37 +186,46 @@ export class Garden {
...
+        }
+      }
 
-      if (!parsedConfig.project) {
-        throw new ConfigurationError(`Path ${projectRoot} does not contain a project configuration`, {
-          projectRoot,
-          config,
-        })
+      if (!config || !config.project) {
+        throw new ConfigurationError(`Could not find config file in directory ${currentDirectory} or below`,
```
eysi09: Shouldn't this be `...or above`. Or use something similar to the git error message when you run a git command outside a git repo: `Not a git repository (or any of the parent directories)`

garden-cli/src/config/base.ts
```diff
@@ -48,18 +48,21 @@ export const configSchema = Joi.object()
 
 const baseModuleSchemaKeys = Object.keys(baseModuleSpecSchema.describe().children)
 
-export async function loadConfig(projectRoot: string, path: string): Promise<GardenConfig> {
+export async function loadConfig(projectRoot: string, path: string): Promise<GardenConfig | undefined> {
```
eysi09: This function should probably stay the same, see my comment below.

garden-cli/src/garden.ts
```diff
@@ -185,37 +186,46 @@ export class Garden {
     this.taskGraph = new TaskGraph(this.pluginContext)
   }
 
-  static async factory(projectRoot: string, { env, config, logger, plugins = [] }: ContextOpts = {}) {
+  static async factory(currentDirectory: string, { env, config, logger, plugins = [] }: ContextOpts = {}) {
```
eysi09: Again, this function should probably still receive the `projectRoot`. I'll explain below.

garden-cli/src/garden.ts
```diff
@@ -185,37 +186,46 @@ export class Garden {
...
     let parsedConfig: GardenConfig
 
     if (config) {
       parsedConfig = <GardenConfig>validate(config, configSchema, { context: "root configuration" })
 
       if (!parsedConfig.project) {
         throw new ConfigurationError(`Supplied config does not contain a project configuration`, {
-          projectRoot,
+          currentDirectory,
```
eysi09: Again, we should probably stick to `currentDirectory` instead of `projectRoot` here.

garden-cli/src/garden.ts
```diff
@@ -263,10 +273,10 @@ export class Garden {
       variables: merge({}, environmentDefaults.variables, envConfig.variables),
     }
 
-    const buildDir = await BuildDir.factory(projectRoot)
+    const buildDir = await BuildDir.factory(currentDirectory)
 
     const garden = new Garden(
-      projectRoot,
+      currentDirectory,
```
eysi09: It would probably be clearer to still use the `projectRoot` variable name since that value isn't necessarily the current directory. So something like:

```
let projecRoot: string
// look for the project root and when you've found it:
projectRoot = currentDirectory
```

garden-cli/src/garden.ts
```diff
@@ -185,37 +186,46 @@ export class Garden {
...
       }
     } else {
-      config = await loadConfig(projectRoot, projectRoot)
-      parsedConfig = await resolveTemplateStrings(config, new ProjectConfigContext())
+      let pathArray = currentDirectory.split(require("path").sep)
+      for (let i = 0; i < pathArray.length; i++) {
+        config = await loadConfig(currentDirectory, currentDirectory)
+        if (!config) {
+          pathArray.pop()
+          pathArray.unshift("/")
```
eysi09: This should be `sep` instead of `"/"` right?

garden-cli/src/garden.ts
```diff
@@ -185,37 +186,46 @@ export class Garden {
...
-          projectRoot,
+          currentDirectory,
           config,
         })
       }
     } else {
-      config = await loadConfig(projectRoot, projectRoot)
-      parsedConfig = await resolveTemplateStrings(config, new ProjectConfigContext())
+      let pathArray = currentDirectory.split(require("path").sep)
+      for (let i = 0; i < pathArray.length; i++) {
```
eysi09: Just a thought, but instead of splitting the path into an array like this you could use the `join` method to go up one dir like so: `join(path, "..")`. 

So this could possibly be something like:

```javascript
let path = currentDirectory
while (path !== sep) {
  // load config and then:
  if (!config) {
    path = path.join("..")
  } else {
    break
  }
}

File: garden-cli/src/config/base.ts [hunk 10]
```diff
@@ -48,18 +48,21 @@ export const configSchema = Joi.object()
...
   const absPath = join(path, CONFIG_FILENAME)
   let fileData
   let spec: any
 
+  // loadConfig returns null if config file is not found in the given directory
   try {
     fileData = await readFile(absPath)
   } catch (err) {
-    throw new ConfigurationError(`Could not find ${CONFIG_FILENAME} in directory ${path}`, err)
+    // throw new ConfigurationError(`Could not find ${CONFIG_FILENAME} in directory ${path}`, err)
```
eysi09: We should remove this right?

eysi09:
Had a few comments, most of them nit picks. Otherwise looks good.

eysi09:
One thing about this approach is that Garden could possibly be reading and parsing several module configs before finding the project config, only to go back to reading and parsing the same module configs again. I'm assuming that the performance hit is pretty trivial but I've often wondered if we should have different names for module configs and the project config. E.g. `garden-project.yml` and `garden-module.yml`. 

I've also noticed it gets a little tricky when writing docs, always having to specify whether or not it's the top-level / project-level config or not. @edvald, any thoughts on this?

Alternatively, we might consider splitting the `loadConfig` function into `readConfig(dir)` and `resolveConfig(jsonSchema)`. That way, `loadConfig` won't have to both read _and_ parse module configs while looking for the project config. (And we can do away with returning `undefined` which feels a bit hacky in the context of `loadConfig` (but makes sense for `readConfig` imo))

edvald:
I don't really think parsing all the found YAMLs is that much of an issue. Parsing the configs is quick, and in most cases we might hit at most one or two non-project configs. So I think this approach is fine and there's no meaningful performance hit.

Separating the configs is another discussion perhaps, and you may recall that's how it was first set up, but we arrived at the current setup for a few reasons. One is the possibility that we'll have single-module projects, esp. if we introduce the notion of nested projects. Another was that we imagined the two config types might have top-level keys in common, although the schema hasn't really evolved that way. And lastly it's a subjective aesthetical thing - `garden-module.yml` feels rather verbose imo.

But I could be swayed on this. In some ways it'd be clearer when looking at a directory to see from the filename if it's a module or a project (or both, if there are two files). On the other hand the current setup hasn't bothered me much - but that's just me.

There is the option of allowing both - `garden.yml` files as they are, as well `garden-project.yml` or `garden-module.yml` files whose schemas are narrowed to projects and modules, respectively. But I'm always hesitant to add multiple ways to configure the same thing, because that's another source of confusion.

Anyone else feel like the current setup (using `garden.yml` files to configure both projects and modules) is confusing?

eysi09:
A priori I thought it might make things a bit clearer to have separate filenames but I didn't consider the single module project scenario. That's a really good point.

In any case I don't think we should allow both. In general, I'm not a fan of having multiple ways to do the same thing, that just adds to the decision fatigue.

garden-cli/src/config/base.ts
```diff
@@ -48,18 +48,20 @@ export const configSchema = Joi.object()
...
 
+  // loadConfig returns null if config file is not found in the given directory
   try {
     fileData = await readFile(absPath)
   } catch (err) {
-    throw new ConfigurationError(`Could not find ${CONFIG_FILENAME} in directory ${path}`, err)
+    return undefined
   }
 
+  // This is duplicated because I was getting compile errors without it.
```
eysi09: Should this comment be here?

eysi09:
@ellenkorbes This change will actually causes some issues with the logger since it depends on the project root and is initialised before the `Garden` class. I'll throw in a quick workaround before merging - I say workaround because changes we have in the pipelines will make this a non-issue.

garden-cli/src/garden.ts
```diff
@@ -185,37 +186,46 @@ export class Garden {
...
     } else {
-      config = await loadConfig(projectRoot, projectRoot)
-      parsedConfig = await resolveTemplateStrings(config, new ProjectConfigContext())
-
-      if (!parsedConfig.project) {
-        throw new ConfigurationError(`Path ${projectRoot} does not contain a project configuration`, {
-          projectRoot,
-          config,
-        })
+      let sepCount = (currentDirectory.match(new RegExp(sep, "g")) || []).length
```
edvald: No big deal, but this is slightly more complex than necessary. `currentDirectory.split(sep).length` would do the same.

garden-cli/src/garden.ts
```diff
@@ -185,37 +186,46 @@ export class Garden {
...
+      for (let i = 0; i < sepCount; i++) {
+        config = await loadConfig(currentDirectory, currentDirectory)
+        if (!config || !config.project) {
+          currentDirectory = join(currentDirectory, "..")
```
edvald: Better to use `resolve(currentDirectory, "..")` here, for readability in messages etc.

garden-cli/src/garden.ts
eysi09: In that case it would have to be `currentDirectory.split(sep).length - 1`

eysi09:
Just realised that the project root might be wrong when the config is passed to the garden factory function. Would also have happened before these changes. @edvald is that going to cause problems?

edvald:
I don't think that's a concern, but we could add a quick check to make sure `config.path === currentDirectory`

eysi09:
Yeah but technically the only thing that matters is that supplied config path is the root path right? Regardless of the `currentDirectory`.

PR #243, opened 2018-09-03
fix(k8s): don't deploy system services when running minikube

edvald:
cc @ellenkorbes - can you give this a spin?


ellenkorbes:
It works!

PR #254, opened 2018-09-07
Docs simple project

eysi09:


eysi09:
This needs to be updated pending some of the PRs we have open.

examples/simple-project-start/services/go-service/webserver/main.go
```diff
@@ -10,7 +10,7 @@ func handler(w http.ResponseWriter, r *http.Request) {
 }
```
ellenkorbes: Here shouldn't we just stick to simply `fmt.Fprintf(w, "Hello from Go!")`? Also there's an extra blank space on line 226 on `simple-project.md`.

eysi09: Good catch. I've updated accordingly.

examples/local-tls/services/go-service/webserver/main.go
```diff
@@ -6,7 +6,7 @@ import (
 )
 
 func handler(w http.ResponseWriter, r *http.Request) {
-	fmt.Fprintf(w, "Hello %s from Go!", r.URL.Path[1:])
+	fmt.Print("Hello from Go!")
 }
```
ellenkorbes: `fmt.Print` will print to the screen, I suggest `fmt.Fprint` into `w` so it prints into the http response.

eysi09: Ah yes. Fixed.

PR #264, opened 2018-09-07
fix(openfaas): better cross-platform support

edvald:
This _should_ fix the issues on Windows, but it needs validation (which I can't currently do myself).


ellenkorbes:
I tried testing this but I'm having all sorts of trouble getting things to run on Windows. Let's take a look at this on Monday.

edvald:
That's not necessarily a Windows-related bug. I'll take a look. I'm planning to have something ready for Klang on Monday btw.

edvald:
I think this will do the trick - note that there is a breaking change in there, the K8s env needs to be reset (that fix solves other edge-case problems as well, so it needs to happen).

PR #265, opened 2018-09-08
refactor: get rid of native OpenSSL dependency

### LINKED ISSUE #261: Replace x509 library with PKI.js
Mainly because https://github.com/PeculiarVentures/PKI.js has no native dependencies, but it also seems more complete.

Part of #256

edvald:
By removing the x509 library, in favor of the (admittedly odd) certpem/pkijs libs

Closes #261

### DIFF
--- a/garden-cli/package.json
+++ b/garden-cli/package.json
@@ -28,6 +28,7 @@
     "async-lock": "^1.1.3",
     "axios": "^0.18.0",
     "bluebird": "^3.5.1",
+    "certpem": "^1.1.2",
     "chalk": "^2.4.1",
     "child-process-promise": "^2.2.1",
     "chokidar": "^2.0.4",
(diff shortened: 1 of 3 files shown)


garden-cli/src/plugins/kubernetes/ingress.ts
```diff
@@ -181,14 +181,28 @@ async function getCertificateHostnames(api: KubeApi, cert: IngressTlsCertificate
...
 
     try {
-      const subject = x509.getSubject(crt)
-      const hostnames = uniq([
-        ...(subject.commonName ? [subject.commonName] : []),
-        ...x509.getAltNames(crt),
-      ])
+      // Note: Can't use the certpem.info() method here because of multiple bugs.
+      // And yes, this API is insane. Crypto people are bonkers. Seriously. - JE
+      const certInfo = certpem.debug(crtData)
```
drubin: My js isn't that good but I think debug is an alias for `getCertInfo` maybe we can use that so it makes more logical sense? 

https://git.coolaj86.com/coolaj86/cert-info.js/src/branch/master/index.js#L68

garden-cli/package.json
```diff
@@ -28,6 +28,7 @@
     "async-lock": "^1.1.3",
     "axios": "^0.18.0",
     "bluebird": "^3.5.1",
+    "certpem": "^1.1.2",
```
drubin: I think considering how little we use the native implementation this is a good way for now, also native openssl is *always* a screw up with miss matched versioning so anything that avoids this feels good.

PR #269, opened 2018-09-10
refactor(logger): clean up code + enable empty log entries

eysi09:
I did a little refactor on the Logger in preparation for some other changes that I haven't gotten around to implementing. So I'm just pushing this as is to keep the code from going stale. Changes include: 

* Split `LogNode` and `LogEntry` into separate files
* Remove `logger/types.ts` file
* Use union types instead of enums
* Allow empty log entries

Most notably we can now pass around empty entries to update them later. 

Example:

```javascript
ctx.log.info("Before")
const emptyForNow = ctx.log.info()
ctx.log.info("After")
```
prints:
```sh
Before
After
```
and then
```javascript
emptyForNow.setState("In between")
```
prints:
```sh
Before
In between
After
```


drubin:
How does this magic work with file-based logging?  Does it simply add a new log entry after (I assume we had this problem before and figured out a way of handling it?)

edvald:
This is great, we should use this (in a separate PR) to do a bit of an overhaul of how we pass loggers around, and improve the nesting.

Re: The file writer, I actually don't think we should do the "magic" at all there. There's no real point to it imo, and I can't think of a single file logger that buffers and overwrites parts of the log.

eysi09:
Yeah, agreed. My point was not to overwrite anything but rather that we technically could use the complete graph to write a log _after_ a garden "cycle" (assuming Garden is in watch mode).

PR #270, opened 2018-09-10
Release 0.6.0

eysi09:


CHANGELOG.md
```diff
@@ -1,3 +1,60 @@
+<a name=""></a>
+# [](https://github.com/garden-io/garden/compare/v0.3.0...v) (2018-09-10)
```
edvald: Hmm, this doesn't look right. Something's up with the changelog generation.

CHANGELOG.md
```diff
@@ -1,3 +1,60 @@
...
+* **config:** Endpoint names will have to be removed from module configs.
+* **config:** Module name no longer defaults to directory name but must be explicitly
+set in the module's garden.yml file. Any existing garden.yml module files
+without a name key must therefore be updated to expclitily provide the
+module name.
+
+
+
 <a name="0.5.0"></a>
 # [0.5.0](https://github.com/garden-io/garden/compare/v0.3.0...v0.5.0) (2018-09-05)
```
edvald: Seems to have happened last time as well, didn't notice it then. We should fix the bug and manually fix this in the changelog.

eysi09: Indeed, didn't notice. I'll fix the changelog and look into the underlying problem.

eysi09:
I figured out was wrong with the changelog, it was a bit of a doozy.

So, `conventional-changelog` uses `git-semver-tags` to get an ordered list of previous versions which it does by simply running `git log --decorate --no-color` and look for strings that start with `tag: ` and look like semver.

However, we were previously amending the release commits when adding the changelog to them which caused any `tag: `-like strings to disappear from the `git log --decorate` result.

In general we should be aware that `conventional-changelog` reads the current version from `package.json` and previous versions from `git log --decorate`.

The workflow should therefore be to bump the version, then generate the changelog and finally commit and tag. Since lerna bumps the version, commits it, and tags the commit in one go, we need to reset the commit before generating the changelog and make sure we re-create the tag after amending the commit.

PR #280, opened 2018-09-13
feat: windows installer

ellenkorbes:
This is a PowerShell script for the installation on Windows.

It does:

```
# 1. Checks whether Hyper-V is enabled.
# 2. Checks whether Docker is installed.
# 3. Checks whether Kubernetes is the default orchestrator for Docker.
# 4. Installs Stern.
# 5. Installs the Chocolatey package manager.
# 6. Installs the listed dependencies using Chocolatey if they're not already 
#    present (currently git, nodejs, rsync, and kubernetes-helm).
# 7. Installs the windows-build-tools NPM package.
# 8. Install the garden-cli NPM package.
```

Ideally we should be able to run it the same way as the Chocolatey installer, just by copying the command from our website and pasting it on the shell.

The Chocolatey package is in progress, should be done tomorrow. It's basically the same thing wrapped in a Chocolatey package, except without the bit that installs Chocolatey itself.

There are some comments and a couple unused functions that I'll leave there until we get to a finished version, as they may still be useful.

One thing we need to decide on about the Chocolatey package is the uninstall script. When installing, we're adding *a bunch* of stuff. When uninstalling, what should we remove? I don't think it makes sense to remove say node and git from the user's machine, but then again we did install them. I'm partial to just doing a npm uninstall on the garden-cli and that's that.

Okay, now for the bad news:

- It Works On My Machine®. We need clean Windows install to test it at least, be it VM or physical.
- While it does install correctly, the thing it is installing doesn't quite work just yet. There are still bugs in our NPM release that need to be ironed out for this to work right on Windows.
- If I am to get all the documentation ready before the Secret Meeting® on the 25th, I'll need to stand down from this and someone else will have to take point. I can finish this work with the installers but I don't think I can do all the bug hunting *and* the docs in a timely fashion.

### DIFF
--- a/docs/introduction/getting-started.md
+++ b/docs/introduction/getting-started.md
@@ -2,6 +2,12 @@
 
 This guide will walk you through setting up the Garden framework.
 
+Please follow the guide for your operating system:
+
+* [macOS](#macos)
+* [Windows](#windows)
+* [Linux (or manual installation on other platforms)](#linux-manual-installation)
+
 ## Installation
 
 ### macOS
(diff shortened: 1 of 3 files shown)


edvald:
A couple of notes to start:
1) I don't actually think we should install Chocolatey itself, but maybe check if it is installed.
2) I was going to get rid of the stern installation dependency, by running it as a docker container. I'll try and get that in a PR shortly.
3) I think we can safely just install the choco deps, Chocolatey will surely just do nothing if they're already up to date?

And a question: Which NPM bugs are you referring to, specifically?

ellenkorbes:
Thing is, we need Chocolatey to install the dependencies. If it's not there then we need to install it first. 

And the reason I'm checking if the CLI commands are available is because you might have them already from say a manual install without Chocolatey, in which case I'm not sure whether Chocolatey would detect them or overwrite your previous installation.

The one NPM bug I can think of off the top of my mind is that when running the hello-world example the plug-in tries to run a bootstrap script, and it's called with "./bootstrap" and Windows throws an error there. If we make it simply "bootstrap" it works. I recall there being a second error after that one's fixed but I'll have to check.

edvald:
My point is, we can generally assume that our users already have Chocolatey, so the install step may be superfluous. But I suppose it doesn't hurt.

That said, while this script is a good place to start, we're not going to ask our users to clone the repository. We'll just publish a Chocolatey package, so there's no reasonable way for them to run this script.

I'm also wondering if it's sensible to block the package installation if Hyper-V, docker, K8s etc. is not installed, but we should rather check it at runtime and log warnings at installation time.

The bootstrap script thing has been fixed in #264, just need to test that and merge if it works. Could you quickly try that one? Are there other errors that remain? We already removed that annoying OpenSSL dependency.

ellenkorbes:
> we can generally assume that our users already have Chocolatey

Can we?

> we're not going to ask our users to clone the repository.

Indeed we're not. We're gonna ask them to copy a line of text from our website and paste it into their terminal, [the same way the Chocolatey installer works](https://chocolatey.org/install).

> I'm also wondering if it's sensible to block the package installation

We should at least block if they're using Windows versions where virtualization doesn't work then, e.g. Windows 10 Home.

> The bootstrap script thing

Will test.

edvald:
Well, the same way we expect mac users to have homebrew. And I think it's enough to point people to the chocolatey installation docs.

> We should at least block if they're using Windows versions where virtualization doesn't work then, e.g. Windows 10 Home.

I would agree, but there will be Garden providers that don't required all that Kubernetes / Docker for Desktop jazz. So I still think it should be a warning.

PR #284, opened 2018-09-18
docs: link specifically to edge version

ellenkorbes:
I always find it a bit of a hassle to find the specific link to the Edge version of Docker, so I thought I'd use it in the documentation instead of the generic Docker For Windows one.


drubin:
@ellenkorbes Minor issue, your commit should be `docs:`  since the changelog is auto-generated from this. 

ref https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#type

ellenkorbes:
Fixed. Thanks @drubin.

drubin:
@ellenkorbes Maybe we need to make it more clear you need the edge version only the URL contains edge and I not sure people will notice their versions are different from edge?

ellenkorbes:
The text immediately preceding this link is 

> Docker - We strongly recommend using the _Edge version_

so IMO we're fine in that regard.

PR #286, opened 2018-09-19
fix: incl. ingresses & services in delete command.

thsig:
When deleting a container service via the delete command, also delete
its k8s service and ingress.

This fixes problems previously caused by dangling k8s
ingresses/services.

### DIFF
--- a/garden-cli/src/plugins/kubernetes/deployment.ts
+++ b/garden-cli/src/plugins/kubernetes/deployment.ts
@@ -323,18 +323,31 @@ export async function createDeployment(
 }
 
 export async function deleteContainerService(
-  { namespace, provider, serviceName, deploymentOnly, logEntry },
+  { namespace, provider, serviceName, logEntry },
+) {
+
+  const context = provider.config.context
+  await deleteContainerDeployment({ namespace, provider, serviceName, logEntry })
+  await deleteObjectsByLabel({
+    context,
+    namespace,
+    labelKey: "service",
+    labelValue: serviceName,
+    objectTypes: ["deployment", "service", "ingress"],
+    includeUninitialized: false,
+  })
+
+}
+
+export async function deleteContainerDeployment(
+  { namespace, provider, serviceName, logEntry },
 ) {
 
   let found = true
-  const context = provider.config.context
   const api = new KubeApi(provider)
 
   try {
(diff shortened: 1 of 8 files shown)


garden-cli/src/plugins/kubernetes/kubectl.ts
```diff
@@ -193,3 +198,28 @@ export async function applyMany(
     return result.output
   }
 }
+
+const defaultObjectTypesForDelete = ["deployment", "service", "ingress"]
```
edvald: I think it makes sense to simply make this a required parameter and not have defaults.

garden-cli/src/plugins/kubernetes/deployment.ts
```diff
@@ -324,17 +322,19 @@ export async function createDeployment(
...
 
-export async function deleteContainerService({ namespace, provider, serviceName, logEntry }: {
-  namespace: string,
-  provider: KubernetesProvider,
-  serviceName: string,
-  logEntry?: LogEntry,
-}) {
-  const api = new KubeApi(provider)
+export async function deleteContainerService(
+  { namespace, provider, serviceName, deploymentOnly, logEntry },
```
edvald: Instead of this flag, I'd just implement this logic specifically in the OpenFaaS handler. And actually I'd suggest using the faas-cli command there instead (and perhaps make a quick helper for it, because of all that `docker run` stuff we do there now).

garden-cli/src/plugins/kubernetes/kubectl.ts
```diff
@@ -193,3 +198,28 @@ export async function applyMany(
...
+
+export async function deleteObjectsByLabel(
+  context: string, namespace: string, labelKey: string, labelValue: string,
+  { includeUninitialized = false, objectTypes = defaultObjectTypesForDelete }: DeleteOptions = {},
```
edvald: Because we have so many parameters here now, I'd suggest putting all of them in an object. Too many positionals are just confusing to write.

garden-cli/src/plugins/kubernetes/deployment.ts
thsig: Yeah, makes sense.

PR #287, opened 2018-09-20
feat(k8s): support remote container registries and remote clusters

edvald:
This changes the deployment mechanics to include a push step before
deploying, which is performed if the provider has a `pushModule`
handler. In turn the existing "push" mechanics have been renamed to
"publish", which more accurately describes the previous implementation.


edvald:
@ellenkorbes some documentation updates here

PR #291, opened 2018-09-21
Hot-reload functionality for local k8s

thsig:
### Summary

When the `local-kubernetes` provider is used, `container` modules can now be configured to hot-reload their running services when the module's sources change (i.e. without redeploying).

For example, services that can be run with an FS watcher that automatically applies changed sources to the running application (e.g. nodemon, Django, Ruby on Rails, and many other web app frameworks) are a natural fit for this feature.

For those who like to read along, much of the action takes place in:

 https://github.com/garden-io/garden/blob/hot-reload/garden-cli/src/plugins/kubernetes/deployment.ts#L392-L460

and

https://github.com/garden-io/garden/blob/hot-reload/garden-cli/src/plugins/kubernetes/actions.ts#L230-L267

To see this in action, the `hot-reload` example project is a good place to start.

### Implementation

The current hot reload implementation (only for `container` modules in projects using the `local-kubernetes` provider) consists of the following:

* A set of k8s `emptyDir` volumes, representing the subdirs of the module to be synced. These are mounted at the `target` paths specified in the module's `garden.yml` (under `hotReload.sync`). The most common case will probably be a single `source` / `target` pair representing the module's top-level directory (resulting in one volume).
* A k8s init container using the service's image, which populates the above volumes according to the specifications in the module's `garden.yml`.
* An rsync sidecar container, deployed in the same pod as the service's own container, which runs an rsync server.
* Each of the volumes is mounted in the service's container (at the paths specified in the module's `garden.yml`), and in the rsync sidecar container.
* A `hotReload` module action, called via Garden's FS watcher when sources change for a module that is configured for hot reloading. `rsync` is used to sync the module's updated source files to the appropriate target paths inside each of its services' rsync sidecar containers.
* Each target path inside the rsync sidecar containers is contained in one of the above volumes. This therefore results in the updated source files being copied into the services' running containers.

### Semantics / usage

Currently, modules configured for hot reloading are only deployed in hot reload configuration (i.e. with the rsync sidecar, volumes and init container described above) when deployed via `garden deploy -w` or `garden dev` (and not, for example, when deployed via `garden deploy` without the `-w` flag).

Subsequently deploying a service (belonging to a module configured for hot reloading) via `garden deploy` (without the watch flag) results in the service being redeployed in "standard configuration", i.e. without the rsync sidecar, init container and sync volumes.

Since hot reloading currently happens via Garden's FS watcher, hot reloading only occurs while a `garden deploy -w` / `garden build -w` / `garden dev` command is running.

When a module is hot-reloaded, a build task for the module is simultaneously added, but unlike "cold" reloading (the default), no deploy task is added. Tasks for the hot reloaded module's build/service dependencies/dependants are added in the same way as during a cold reload.

#### Changes to command options (to be completed before merging)

Currently, all services belonging to modules configured for hot reloading are deployed in hot reload configuration. This doesn't scale well for systems with many hot-reload-capable services, since the FS watching overhead adds up quite quickly.

Therefore, the user should be able to specify which modules to deploy with hot reloading enabled when running `garden dev|deploy|build`.

One approach here is a comma-separated list of module names, something like: `garden dev --hot-reload="this-module,that-module"`. Any other ideas/opinions on this?

If the user doesn't specify a list of modules to hot reload, should the default be to deploy all hot-reload-capable modules' services with hot reloading turned on? Or the opposite (i.e. all services deployed without hot reloading)?

#### Remote Kubernetes (for a separate PR)

Architecturally, the current implementation can be naturally extended to support remote k8s, the only difference being that the `rsync` calls triggered by Garden's FS watcher would be pointed at the remote k8s cluster. However, when using the `rsync://`/client-server protocol (which we're using), there's no built-in encryption, so we'd have to set up some sort of secure tunneling. `rsync`'s default, `ssh`-based strategy is automatically encrypted, but probably slower.


garden-cli/src/plugins/kubernetes/actions.ts
```diff
@@ -217,6 +227,45 @@ export async function execInService(params: ExecInServiceParams<ContainerModule>
...
 }
 
+export async function hotReload(
+  { ctx, runtimeContext, module }: HotReloadParams<ContainerModule>,
+): Promise<HotReloadResult> {
+  const hotReloadConfig = module.spec.hotReload
+
+  if (!hotReloadConfig) {
+    return {}
+  }
```
edvald: Should the plugin action be called at all if it is not configured?

I feel like the framework should know ahead of time if a module can be hot-reloaded. I'd suggest doing it the same way as `testConfigs` and `serviceConfigs`, which are populated by the `validate` module action handlers.

garden-cli/src/plugins/kubernetes/deployment.ts
```diff
@@ -321,9 +241,237 @@ export async function createDeployment(
...
+    }
+  })
+
+  container.volumeMounts = (container.volumeMounts || []).concat(syncMounts)
+
+  if (container.ports.find(p => p.containerPort === RSYNC_PORT)) {
+    throw new Error(deline`
+      Service ${serviceSpec.name} is configured for hot reload, but uses port ${RSYNC_PORT},
+      which is reserved for internal use while hot reload is active. Please remove
+      ${RSYNC_PORT} from your services' port config.`)
```
edvald: Would be nicer and more robust to automatically allocate a free port

garden-cli/src/types/plugin/params.ts
```diff
@@ -205,6 +211,7 @@ export const getTestResultParamsSchema = moduleActionParamsSchema
  * Service actions
  */
 export interface GetServiceStatusParams<T extends Module = Module> extends PluginServiceActionParamsBase<T> {
+  watch?: boolean
```
edvald: This name feels ambiguous - are we indicating that the service is being watched, or that we should keep monitoring the status? Maybe `watching` would be better?

garden-cli/src/commands/deploy.ts
```diff
@@ -82,18 +88,57 @@ export class DeployCommand extends Command<Args, Opts> {
...
-        garden,
-        module,
-        serviceNames,
-        force: true,
-        forceBuild: true,
-        includeDependants: true,
-      }),
+      changeHandler: async (module) => {
+
+        const hotReload = !!module.spec.hotReload
```
edvald: This doesn't really work - `spec` is not typed here. We should instead add an optional `hotReloadConfig` (which itself can have a `spec` field, which can include module-type-specific stuff) to `ModuleConfig` and check for its presence.

garden-cli/src/commands/deploy.ts
```diff
@@ -82,18 +88,57 @@ export class DeployCommand extends Command<Args, Opts> {
...
+
+        if (hotReload) {
+          hotReloadAndLog(module, garden)
```
edvald: Missing `await`

garden-cli/src/commands/dev.ts
```diff
@@ -67,6 +68,12 @@ export class DevCommand extends Command {
     const tasksForModule = (watch: boolean) => {
       return async (module: Module) => {
 
+        const hotReload = !!module.spec.hotReload
+
+        if (watch && hotReload) {
+          hotReloadAndLog(module, garden)
```
edvald: Missing `await`

garden-cli/src/commands/build.ts
```diff
@@ -72,6 +73,10 @@ export class BuildCommand extends Command<BuildArguments, BuildOptions> {
       watch: opts.watch,
       handler: async (module) => [new BuildTask({ garden, module, force: opts.force })],
       changeHandler: async (module: Module) => {
+        if (module.spec.hotReload) {
+          hotReloadAndLog(module, garden)
```
edvald: Missing `await`

garden-cli/src/plugins/container.ts
```diff
@@ -409,6 +482,11 @@ export const gardenPlugin = (): GardenPlugin => ({
 
         return { pushed: true }
       },
+
+      async hotReload(_: HotReloadParams) {
+        return {}
+      },
```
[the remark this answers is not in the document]
edvald: We should instead explicitly check for the presence of a hotReload handler, when a module is configured to be hot-reloaded.

garden-cli/src/hotReloadScheduler.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    const prom = new Promise((resolver) => {
+      if (pendingRequest) {
+        pendingRequest.addResolver(resolver)
+      } else {
+        this.pending[moduleName] = new HotReloadRequest(hotReloadHandler)
+      }
+    })
+
+    this.process(moduleName)
```
edvald: Missing `await`

garden-cli/src/hotReloadScheduler.ts
```diff
@@ -0,0 +1,82 @@
...
+ * by this implementation.
+ */
+export class HotReloadScheduler {
+  private pending: { [moduleName: string]: HotReloadRequest }
+
+  constructor() {
+    this.pending = {}
+  }
+
+  requestHotReload(moduleName: string, hotReloadHandler: HotReloadHandler): Promise<any> {
```
edvald: Is there a good reason to use `any` here?

garden-cli/src/commands/deploy.ts
thsig: I omitted the `await` here (and in the other commands mentioned below) on purpose, to perform the hot reloading in parallel with the other tasks (since the two are independent of each other).

That said, since hot reloading should generally be quite fast, this may not make much of a difference. What do you think?

edvald: That's going to cause uncaught promise rejections if something goes wrong, so that doesn't really work... So you either need to wait for the promise or come up with another way to handle exceptions.

thsig: Ah, good point. I'll make the appropriate changes here.

garden-cli/src/config/common.ts
```diff
@@ -28,20 +28,32 @@ export const enumToArray = Enum => (
...
+export const userIdentifierRegex = /^(?!garden)[a-z][a-z0-9]*(-[a-z0-9]+)*$/
 export const envVarRegex = /^(?!GARDEN)[A-Z_][A-Z0-9_]*$/
 
 export const joiIdentifier = () => Joi.string()
   .regex(identifierRegex)
   .max(63)
   .description(
-    "Valid RFC1035/RFC1123 (DNS) label (may contain lowercase letters, numbers and dashes, must start with a letter, " +
+    "Valid RFC1035/RFC1123 (DNS) label (may contain lowercase letters, numbers and dashes, must start with a letter," +
+    " " +
```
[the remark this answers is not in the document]
thsig: Yes indeed.

garden-cli/src/plugins/container.ts
```diff
@@ -8,10 +8,14 @@
 
 import * as Joi from "joi"
 import * as childProcess from "child-process-promise"
+import dedent = require("dedent")
+import deline = require("deline")
```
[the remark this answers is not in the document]
thsig: For the `description` strings in this file, we can.

In general, though, `deline` removes leading whitespace, which isn't always desirable, since sometimes one wants indentation / leading whitespace in part of the string (which works with `dedent`).

garden-cli/src/plugins/container.ts
```diff
@@ -333,6 +382,30 @@ export async function validateContainerModule({ moduleConfig }: ValidateModulePa
...
+      for (const t2 of targets) {
+        if (t2.startsWith(t) && t !== t2) {
+          invalidTargetPairs.push(`${t} is a subdirectory of ${t2}.`)
+        }
+      }
+    }
+
+    if (invalidTargetPairs.length > 0) {
+      throw new ConfigurationError(
+        dedent`Hot reload configuration invalid - a target may not be a subdirectory of another target.\n
```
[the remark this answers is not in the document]
thsig: > "Hot reload configuration invalid" -> "Invalid hot reload configuration"?

Yeah, that's better.

> Also, maybe it would be nice to explain why this is invalid, and to mention this in the docs/reference.

I.e. explain why `targets` may not be subdirectories of other `targets`?

garden-cli/src/plugins/container.ts
```diff
@@ -333,6 +382,30 @@ export async function validateContainerModule({ moduleConfig }: ValidateModulePa
...
+        ${invalidTargetPairs.join("\n")}`,
+        { invalidTargetPairs },
```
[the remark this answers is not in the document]
thsig: Yeah, good idea.

garden-cli/src/plugins/generic.ts
```diff
@@ -126,6 +128,10 @@ export async function buildGenericModule({ module }: BuildModuleParams<GenericMo
   return output
 }
 
+export async function reloadGenericModule(_: HotReloadParams<GenericModule>): Promise<HotReloadResult> {
+  return {}
+}
```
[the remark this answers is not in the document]
thsig: No, good point — removed.

garden-cli/src/types/plugin/params.ts
thsig: Turns out that this field became unnecessary in this interface later on, I've removed it now.

garden-service/src/commands/deploy.ts
```diff
@@ -69,10 +72,23 @@ export class DeployCommand extends Command<Args, Opts> {
...
+      garden.log.error({ msg: "No services found. Aborting." })
       return { result: {} }
     }
 
+    let watch
+    const hotReloadServiceNames = opts["hot-reload"] || []
+    const hotReloadModuleNames = await getHotReloadModuleNames(garden, hotReloadServiceNames)
+
+    if (opts["hot-reload"]) {
+      if (!validateHotReloadOpt(garden, hotReloadServiceNames)) {
```
eysi09: Should this throw an error instead of just returning?

garden-service/src/commands/dev.ts
```diff
@@ -58,20 +76,28 @@ export class DevCommand extends Command {
...
-      }
+      garden.log.info({ msg: "No modules found in project." })
       garden.log.info({ msg: "Aborting..." })
       return {}
     }
 
+    const hotReloadServiceNames = opts["hot-reload"] || []
+    const hotReloadModuleNames = await getHotReloadModuleNames(garden, hotReloadServiceNames)
+
+    if (opts["hot-reload"] && !validateHotReloadOpt(garden, hotReloadServiceNames)) {
```
eysi09: Same as for the deploy command, shouldn't this throw an error?

garden-service/src/actions.ts
```diff
@@ -229,7 +229,10 @@ export class ActionHelper implements TypeGuard {
 
   async hotReload<T extends Module>(params: ModuleActionHelperParams<HotReloadParams<T>>)
     : Promise<HotReloadResult> {
-    return this.callModuleHandler(({ params, actionType: "hotReload" }))
+    // return this.callModuleHandler(({ params, actionType: "hotReload" }))
```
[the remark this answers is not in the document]
thsig: Yessir.

garden-service/src/commands/deploy.ts
thsig: Yeah, good point. Will make that change.

garden-service/src/commands/dev.ts
thsig: Yes indeed.

garden-service/src/types/plugin/params.ts
```diff
@@ -203,18 +203,27 @@ export const getTestResultParamsSchema = moduleActionParamsSchema
...
-  watch?: boolean
   runtimeContext: RuntimeContext
+  verifyHotReloadStatus?: hotReloadStatus
 }
+// export interface GetServiceStatusParams<T extends Module = Module> extends
+//   Omit<Temp, "buildDependencies">
+//   {
+//     watch?: boolean
+//     runtimeContext: RuntimeContext
+//   }
```
eysi09: Remove this?

PR #292, opened 2018-09-21
refactor: rename logEntry to log and require for tasks, cmds and actions

edvald:
This leverages the recent changes to the logger that allow for empty
log entries.

_Note: This currently has an issue where the section part of log entries doesn't appear. Need input from @eysi09 to complete this._

### DIFF
--- a/garden-service/src/cli/cli.ts
+++ b/garden-service/src/cli/cli.ts
@@ -32,6 +32,7 @@ import {
   GardenError,
   PluginError,
   toGardenError,
+  InternalError,
 } from "../exceptions"
 import { Garden, ContextOpts } from "../garden"
 
@@ -67,17 +68,14 @@ const OUTPUT_RENDERERS = {
   },
 }
 
-const logLevelKeys = getEnumKeys(LogLevel)
-// Allow string or numeric log levels
-const logLevelChoices = [...logLevelKeys, ...range(logLevelKeys.length).map(String)]
+const FILE_WRITER_CONFIGS = [
+  { filename: "development.log" },
+  { filename: ERROR_LOG_FILENAME, level: LogLevel.error },
+  { filename: ERROR_LOG_FILENAME, level: LogLevel.error, path: ".", truncatePrevious: true },
+]
 
-const getLogLevelFromArg = (level: string) => {
-  const lvl = parseInt(level, 10)
-  if (lvl) {
-    return lvl
-  }
-  return LogLevel[level]
-}
+const GLOBAL_OPTIONS_GROUP_NAME = "Global options"
+const DEFAULT_CLI_LOGGER_TYPE = LoggerType.fancy
 
 // For initializing garden without a project config
 export const MOCK_CONFIG: GardenConfig = {
@@ -99,6 +97,11 @@ export const MOCK_CONFIG: GardenConfig = {
   },
 }
 
+const getLogLevelNames = () => getEnumKeys(LogLevel)
+const getNumericLogLevels = () => range(getLogLevelNames().length)
+// Allow string or numeric log levels as CLI choices
+const getLogLevelChoices = () => [...getLogLevelNames(), ...getNumericLogLevels().map(String)]
+
 export const GLOBAL_OPTIONS = {
   root: new StringParameter({
     alias: "r",
@@ -113,7 +116,7 @@ export const GLOBAL_OPTIONS = {
   env: new EnvironmentOption(),
   loglevel: new ChoicesParameter({
     alias: "l",
-    choices: logLevelChoices,
+    choices: getLogLevelChoices(),
     help:
       "Set logger level. Values can be either string or numeric and are prioritized from 0 to 5 " +
       "(highest to lowest) as follows: error: 0, warn: 1, info: 2, verbose: 3, debug: 4, silly: 5.",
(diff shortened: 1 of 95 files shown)


eysi09:
I haven't been able to reproduce the problem with the sections. Something like `emptyEntry.setState({ section: mySection, msg: "hello" })` works on my end. Were there any particular steps to reproduce?

The issue with indentation of empty entries has been fixed in #302.

edvald:
The issue isn't with `setState`, it's when I create a new line:
```
const log = garden.log.info()
log.info({ section: "foo" })
```

eysi09:
You're right. Sorry. Fixed in: #316

garden-service/src/cli/cli.ts
```diff
@@ -238,9 +238,11 @@ export class GardenCli {
           contextOpts.config = MOCK_CONFIG
         }
         garden = await Garden.factory(root, contextOpts)
+        const log = garden.log.info()
```
eysi09: Creating this entry here causes some log output to be printed before the command header. Like here:

```
g build
    ✖ openfaas--builder         → Building
    ✔ hello-npm-package         → Building → Done (took 0.1 sec)
    ✔ hello-container           → Building hello-container:vd8c433df4d... → Done (took 1.2 sec)
Build 🔨```

Comment (edvald) [reply garden-service/src/cli/cli.ts]: Good catch. I'll clean that up, and look for similar issues.

**Comment by eysi09** [comment]
There are currently two issues:

**Issue 1**

The commands use a placeholder log entry (and pass them onwards) whereas the `Garden` class uses the root logger (and passes it onwards). This means that everything logged by the  `Garden` class, and all the methods that receive the root logger from it, will always be either above or below everything that descends from the command placeholder entry.

This results in something like this:
```
Logging stuff from the Garden class
    Build 🔨

    ✔ go-service                → Building go-service:611c61cddc... → Done (took 0.5 sec)

    Done! ✔️
```
or something like this:
```
    Build 🔨

    ✔ node-service              → Building node-service:611c61cddc... → Done (took 0.7 sec)
    ✔ go-service                → Building go-service:611c61cddc... → Done (took 0.8 sec)

    Done! ✔️

Logging stuff from the Garden class
```
depending on whether or not the placeholder log entry is created before or after the Garden class is initialised. In the second example the "Logging stuff..." text gets repeatedly pushed down as more content is printed above which looks a little bit awkward. 

**Solution**

We can of course move the `Build` and `Done` messages around to ensure they always get printed first and last. By I just wanted to double check that this is indeed the intended behaviour.

**Issue 2**

The logs printed by the placeholder log entry all get indented since they're the children of the empty entry. This is because the original motivation behind allowing entries to create child entries was to create subtasks.

**Solutions**

* Never indent subtasks unless the option is explicitly set (we'd need to update all current subtasks)
* Check if parent is empty and if so skip indentation at the render level. This feels  a bit hacky since `someEntry.info("am I indented?")` might either be indented or not, depending on how `someEntry` was created. 
* Continue to distinguish between the logger and individual log entries. This would be the most idiomatic way imo.
* Set indentation to `-1` on the empty entry that gets passed to the commands. This is probably the most simple solution.

Note that `setState` prevents indenting since that just updates the current (possibly empty) entry. However, that won't work if we need to create child entries from that entry.

**Comment by eysi09** [comment]
I just update the branch with a fix for issue 2, opting for solution d).

So now the main concern is how we handle issue 1.

**Comment by edvald** [comment]
Re Issue 1, I think we should have Garden accept a LogEntry instead of a Logger. I think that would sort the issue out more neatly. Then `garden.log` does the expected thing.

**Comment by eysi09** [comment]
Updated the PR with a commit that I think should solve most of our issues for now.

I actually went ahead and remove the ability to create an empty log entry like so: `logger.info()` or `logEntry.info()` and instead added a `placeholder` method that does the same with the following signature: `placeholder(level: LogLevel = LogLevel.info): LogEntry`. 

I think that makes things a little bit clearer. My main concern was that a function that receives a log entry doesn't really know what to do with it, since it wasn't clear from the context whether the entry was simply an empty placeholder that should be populated or a task with a spinner that should be updated. Type safety-wise, this is still an issue but at least the intent is a little bit clearer.

I think it would make sense to continue to use the `logger` as before within the commands and have them set up the entries that get passed to the tasks and actions. That could be an intermediary step before we implement something event based, but I skipped that for now so that we can finally get this merged.

Regarding the log entry control flow in general, this is kind of how I envisioned it (and why I've been on the fence about making everything a log entry):
```javascript
const entry = logger.info({msg: "running task..." status: "active"})
try {
  // pass the entry along so that it can be updated while task is running
  runTask(entry)
} catch (err) {
  entry.setState(err.message)
}
entry.setSuccess("whoop")
```
However, the entries are kind of all over the place right now. Also, when everything is an entry, you might end up calling `setState` on empty placeholders which doesn't really make sense.

But as I mentioned above, I think some of this can be reigned in by initialising the placeholder entries at the command level and structure the whole thing a bit more.

eysi09:
@edvald, if the last commit looks okay to you I can go ahead and squash and merge.

eysi09:
...or perhaps I'll leave is as two commits since there's a lot of conversation around this.

edvald:
👍 approved

PR #300, opened 2018-09-24
fix: handle all promises and add no-floating-promises linting rule

edvald:
This addresses a few bugs, including error handling issues in the
TaskGraph.


eysi09:
Looks good. @thsig do you have any comments on replacing the `OperationQueue` with the `p-queue` library?

PR #304, opened 2018-09-25
refactor: remove explicit helm and stern installation dependencies

edvald:
We now automatically fetch helm when needed via the new tool helper.

Note: We still install Helm via package dependencies, because it's
really no bother for the user, but we don't need to explicitly list
it in docs.

Before completing this and using for more tools, I wanted to float the idea and see if it makes sense to everyone...

### DIFF
--- a/docs/introduction/getting-started.md
+++ b/docs/introduction/getting-started.md
@@ -65,7 +65,7 @@ The things the script will check for are the following:
   installed, and whether Kubernetes has been enabled as the default orchestrator.
 * Node.js - The script will install it via Chocolatey if it is missing, but note that _if you already have Node.js
   installed, please make sure it is version 8.x or newer._
-* Git, rsync, Helm and stern. The script will install those if they are missing.
+* Git and rsync. The script will install those if they are missing.
 
 To run the script, open PowerShell as an Administrator and run:
 
@@ -83,8 +83,6 @@ You need the following dependencies on your local machine to use Garden:
 * [Docker](https://docs.docker.com/)
 * Git
 * rsync
-* stern
-* [Helm](https://github.com/kubernetes/helm)
 * Local installation of Kubernetes and kubectl
 
 #### Step 1: Docker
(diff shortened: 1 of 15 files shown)


garden-cli/src/plugins/kubernetes/helm.ts
```diff
@@ -221,11 +222,44 @@ async function build({ ctx, module, logEntry }: BuildModuleParams<HelmModule>):
...
+    },
+    linux: {
+      url: "https://storage.googleapis.com/kubernetes-helm/helm-v2.10.0-linux-amd64.tar.gz",
+      sha256: "0fa2ed4983b1e4a3f90f776d08b88b0c73fd83f305b5b634175cb15e61342ffe",
+      extract: {
+        format: "tar",
+        executablePath: ["linux-amd64", "helm"],
+      },
+    },
+    win32: {
```
ellenkorbes: Isn't it a bit confusing to call it `win32` when we specifically only allow things to run on x86?

edvald: This is standard Node.js terminology (as output by `os.platform()`). And for whatever reason, `win32` is just what it's commonly called, even if it's actually running `x64`

ellenkorbes:
FYI this seems to work fine on Linux.

garden-cli/support/install.ps1
```diff
@@ -5,12 +5,11 @@
 # 1. Checks whether Hyper-V is enabled.
 # 2. Checks whether Docker is installed.
 # 3. Checks whether Kubernetes is the default orchestrator for Docker.
-# 4. Installs Stern.
-# 5. Check if Chocolatey is installed, installs it if missing.
-# 6. Installs the listed dependencies using Chocolatey if they're not already
+# 4. Check if Chocolatey is installed, installs it if missing.
+# 5. Installs the listed dependencies using Chocolatey if they're not already
 #    present (currently git, nodejs, rsync, and kubernetes-helm).
```
eysi09: Shouldn't `kubernetes-helm` be removed from this list and from `chocDeps` on [here](https://github.com/garden-io/garden/blob/7bcb7c074f4b273e9191b230fe51981429c36696/garden-cli/support/install.ps1#L141)

edvald: I debated that, since users are likely to want it installed. But we could remove it, sure.

PR #306, opened 2018-09-25
chore(examples): make node example services build faster

edvald:


edvald:
Yep. There's just one python service, and it was already done right there.

PR #310, opened 2018-09-27
Experimental container build

edvald:
This is just a step in the direction, and the container is only run if you run `garden-service/bin/garden`, but I wanted to get this step merged before moving on with the go CLI.

Note that I've renamed `garden-cli` to `garden-service` in preparation for next steps (and to avoid annoying git conflicts when that time comes). After merging this you will need to change any relevant aliases (if any) and re-run `npm link` in the `garden-service` directory.

I suggest looking at individual commits when reviewing.


garden-service/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# system dependencies
+RUN apk add --no-cache \
+  bash \
+  curl \
+  docker \
+  git \
+  openssl \
+  rsync \
+  && curl https://raw.githubusercontent.com/helm/helm/master/scripts/get | bash \
```
drubin: Can we not change this to a simple wget `https://storage.googleapis.com/kubernetes-helm/helm-v2.11.0-linux-amd64.tar.gz`

This script is none deterministic as to what versions it installs.

garden-service/Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM node:10.11.0-alpine
+
+WORKDIR /project
+ENTRYPOINT [ "garden" ]
```
drubin: Slightly meta question, (and maybe this comes later)?

Do we want to have the entry point something like `tini` and make the CMD the full `garden dev` as the command since that actually handles signals and reaping up processes?

Maybe I have been bitten a bit in terms of trying to run docker images with entry points set to their commands like `terraform` does(this doesn't work on ci requires overriding it).

garden-service/Dockerfile
edvald: Yea good point

garden-service/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+  && chmod +x /usr/local/bin/kubectl
+
+# npm dependencies
+ADD package.json package-lock.json /garden/
+
+RUN apk add --no-cache --virtual .deps \
+  python2 \
+  make \
+  g++ \
+  rsync \
```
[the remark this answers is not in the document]
edvald: Good catch, will fix

garden-service/Dockerfile
edvald: I was just saving it for later but yes, definitely. I can do it now actually, might as well.

edvald: (Commenting here for posterity) Turns out tini is not necessary anymore. Also looked into running as root, and that won't work until we're syncing code instead of mounting host directories.

PR #314, opened 2018-10-01
Multi repo example project

eysi09:


docs/reference/config.md
```diff
@@ -35,11 +35,10 @@ module:
 
   description:
 
-  # A remote repository URL to fetch the module from. Garden will read the garden.yml config from
-  # the local module. Currently only supports git servers.
+  # A remote respository URL. Currently only supports git servers. Must contain a hash part
+  # pointing to a specific branch or tag with the format: <git remote url>#<branch|tag>
```
edvald: Would a commit hash also work?

garden-service/src/config/module.ts
```diff
@@ -80,12 +80,7 @@ export const baseModuleSpecSchema = Joi.object()
       .description("The name of this module.")
       .example("my-sweet-module"),
     description: Joi.string(),
-    repositoryUrl: joiRepositoryUrl()
-      .description(
-        "A remote repository URL to fetch the module from. Garden will read the garden.yml config" +
-        " from the local module." +
-        " Currently only supports git servers.",
-      ),
+    repositoryUrl: joiRepositoryUrl(),
```
edvald: Might it be worth explaining here in a description how this differs from declaring a source at the project level?

examples/remote-sources/README.md
```diff
@@ -0,0 +1,100 @@
...
+```sh
+garden deploy
+```
+Garden will continue to use the version originally downloaded. Use the `update-remote sources|modules|all` command to fetch the latest version of your remote sources and modules:
+```sh
+garden update-remote modules jworker
+```
+If you however change the repository URL of your remote source or module (e.g. switch to a different tag or branch), Garden will automatically fetch the correct version.
+
+If you want to make changes to the remote source you can link it to a local directory with the `link source|module` command:
```
edvald: It feels like this could more clearly state that you would clone the repo to a local directory and point to your local clone.

Side note (for future work) - would it be nice to have a `garden clone` helper?

docs/reference/config.md
eysi09: Not at the moment. We'd need to treat commit hashes as a special case when cloning and fetching and therefore figure out whether the string after the hash symbol is a commit hash or just a tag/branch name.

eysi09: Not at the moment.

At first I thought that that would mean we'd have to download the entire git history (which is not the case for branches/tags) but according to this [SO post](https://stackoverflow.com/questions/14872486/retrieve-specific-commit-from-a-remote-git-repository/30701724#30701724) it's actually possible to fetch a single commit. At least on Github, it depends on the Git server configuration.

The syntax also seem to be different from fetching branches/tags so we'd need to be able to a commit id apart from a branch/tag name.

ellenkorbes:
Hi, I've added a couple *very tiny* changes here.

PR #319, opened 2018-10-05
fix: add missing parenthesis to windows install script

eysi09:


garden-service/support/install.ps1
```diff
@@ -137,7 +137,7 @@ if ((CheckChocolatey) -eq $false) { return }
 # name.
 $chocDeps = (("git","git"),
              ("rsync","rsync"),
-             ("node","nodejs"),
+             ("node","nodejs")),
```
ellenkorbes: Hi, we need to also remove the trailing comma on the last line.

eysi09: I had one job :(

PR #329, opened 2018-10-12
[WIP] feat(cli): experimental go frontend CLI

edvald:
This is just a step towards the Go CLI. We _should_ be able to merge this without any impact on the current way of using garden, but since this introduces some changes to the build flows we may want to review and merge this.

To get this working, you need to do the following:
1. Move your garden repo directory to `$GOPATH/src/github.com/garden-io/garden`
2. Install `dep` (`brew install dep`)
3. Run `dep ensure` in your repo directory
4. Run `gulp build` to build the CLI

You may then want to alias `garden` to the built binary (`garden-cli/build/garden`) when developing.

This is all highly experimental, there are several steps remaining until we get to our final destination of having a single-binary installation.


garden-cli/docker/sync/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM alpine:3.8
+
+# system dependencies
+RUN apk add --no-cache unison
+
+RUN mkdir host-mount
+
+CMD cp -r host-mount/. project/ \
```
edvald: If the volume already exists, this may fail. We could also do an initial rsync between the directories here (or if unison has a "one-way" sync command, that would also work).

garden-cli/docker/sync/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM alpine:3.8
```
edvald: I feel like this could live under `garden-sync` instead of `garden-cli/docker/sync`

garden-cli/dockerutil/docker.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	return cmd.Run()
+}
+
+func HasVolume(volumeName string) bool {
+	args := []string{"volume", "inspect", volumeName}
+	err := Exec(args, true)
+	if err != nil {
+		return false
```
edvald: Probably want a TODO here to distinguish between different potential errors.

garden-cli/main.go
```diff
@@ -43,8 +43,31 @@ func main() {
...
 
-	err = runGardenService(projectName, gitRoot, relPath, os.Args[1:])
+	args := os.Args[1:]
+	watch := fsWatchEnabled(args)
+
+	if watch {
+		err = runSyncContainer(projectName, gitRoot, relPath)
+		if err != nil {
+			os.Exit(1)
+		}
```
edvald: Probably better to just do `check(err)` here, or use `log.Fatal(<some clearer message>)`. Same below.

edvald: Also, we'd need to make sure the initial copy is complete before running the service below. We may need to run the sync container once with the copy/rsync command and wait for that to complete, before starting the unison process.

garden-cli/run.go
```diff
@@ -25,29 +38,16 @@ func runGardenService(projectName string, gitRoot string, relPath string, args [
 			"--volume", fmt.Sprintf("%s/.docker:/root/.docker", homeDir),
 			"--volume", fmt.Sprintf("%s:/root/.garden", gardenHomeDir),
 			"--volume", fmt.Sprintf("%s/.kube:/root/.kube", homeDir),
-			// Mount the project directory.
-			// TODO: sync into a garden-sync container instead
-			"--volume", fmt.Sprintf("%s:/project:delegated", gitRoot),
+			"--volume", fmt.Sprintf("%s/.git:/root/garden/.git", gardenHomeDir),
```
edvald: Hmm, this doesn't seem right. `~/.garden` is not supposed to contain a git root.

garden-cli/main.go
```diff
@@ -43,8 +43,31 @@ func main() {
...
+	}
+
+	err = runGardenService(projectName, gitRoot, relPath, watch, args)
 	if err != nil {
 		os.Exit(1)
 	}
 }
+
+// FIXME: We need a proper way to check if the command requires file system watch. This is not it.
+func fsWatchEnabled(args []string) bool {
```
edvald: I'm a little skeptical about having this check. If we have a long-running sync container, it's also probably an unnecessary optimization. If we always start by doing an initial copy when creating the volume, then we should be able to assume that _if the sync container is already running_, that the synced volume is up to date. So we then don't have to have separate flows.

eysi09: How would we like to manage the sync containers in general? Should they just run along indefinitely, even though a user might have several projects and thus several unison processes running? Or should there be some clean up mechanism?

edvald: We might want to implement some sort of timeout mechanism, so that they shut down after a certain time of inactivity. A little tricky to implement though, so I kinda filed that as an optimization we could make later. That said, the unison process shouldn't consume a lot of resources if the user isn't working on the project.

eysi09: With the approach above, I figured we'd stop the sync container after each run (and perhaps clean up any dangling sync containers) and then run it again next time a command requires file system watch. That would increase the start time of long running commands but not the other ones and there would always only be one sync process running.

garden-cli/sync.go
```diff
@@ -0,0 +1,38 @@
...
+		return nil
+	}
+
+	dockerArgs := []string{
+		"run", "-d", "--rm",
+		// Mount docker socket and configuration directories.
+		"--volume", "/var/run/docker.sock:/var/run/docker.sock",
+		"--volume", fmt.Sprintf("%s/.docker:/root/.docker", homeDir),
+		// Mount the project directory.
+		"--volume", fmt.Sprintf("%s:/host-mount:delegated", gitRoot),
```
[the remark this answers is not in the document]
edvald: Looking at the docs, I actually think `:delegated` is more appropriate. `:cached` has slightly stronger guarantees in terms of making writes from the container visible on the host, but I don't think that's very important for us.

And yes, this is a mac-only optimization, but the flag is ignored elsewhere.

garden-cli/run.go
```diff
@@ -1,19 +1,32 @@
...
+
+	"github.com/garden-io/garden/garden-cli/dockerutil"
 )
 
-// TODO: we may want to use one of the docker go client libs instead of shelling out here, but this works for now so...
-func runGardenService(projectName string, gitRoot string, relPath string, args []string) error {
+func runGardenService(projectName string, gitRoot string, relPath string, watch bool, args []string) error {
 	homeDir := getHomeDir()
 	gardenHomeDir := getGardenHomeDir()
 	containerName := "garden-run-" + randSeq(6)
```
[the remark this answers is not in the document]
edvald: This container isn't persistent for now. We create a new one for each command. When we have an actual running service we will change this.

garden-cli/run.go
```diff
@@ -0,0 +1,53 @@
...
+	dockerArgs := append(
+		[]string{
+			"run", "-i", "--tty", "--rm",
+			// Give the container direct access to the host network.
+			"--net", "host",
+			// Mount docker socket and configuration directories.
+			"--volume", "/var/run/docker.sock:/var/run/docker.sock",
+			"--volume", fmt.Sprintf("%s/.docker:/root/.docker", homeDir),
+			"--volume", fmt.Sprintf("%s:/root/.garden", gardenHomeDir),
+			"--volume", fmt.Sprintf("%s/.kube:/root/.kube", homeDir),
```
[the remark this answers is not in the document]
edvald: We're running as root. We had to do that when deploying as a K8s pod (trying to recall the exact reason...), but we can maybe see if it's still necessary.

`/garden` is just a folder we create, not a user directory (which would then be `/home/garden`).

garden-cli/sync.go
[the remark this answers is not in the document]
drubin: You are right. I miss read the docs 
I only read  `:cached` 

> As such, cached typically improves the performance of read-heavy workloads, at the cost of some temporary inconsistency between the host and the container.

With out seeing 
>For directories mounted with cached, the host’s view of the file system is authoritative; writes performed by containers are immediately visible to the host, but there may be a delay before writes performed on the host are visible within containers.

garden-cli/run.go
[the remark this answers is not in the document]
drubin: yes I am suggesting we switch this out sooner rather than later. I rand into issues where things broke and I had dangling containers.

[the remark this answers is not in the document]
edvald: But I guess the concern does stand for watching commands... One thing we could do would be using the parent process PID instead of the random letters, that would make it easy to see which containers have an attached CLI process.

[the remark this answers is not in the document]
drubin: mmm we should just use a predictable name based on project (how ever we define that, name, path/*) 

PID is pretty much random too. I am suggesting we need idempotency across runs (i.e command stopping and restarting)

garden-cli/main.go
eysi09: ...on second thought, that's probably going to be too slow.

garden-cli/run.go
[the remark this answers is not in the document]
drubin: we might need to run as root to build docker images. Guess its fine for now (unless we start writing files from the container back to the host os)

garden-cli/run.go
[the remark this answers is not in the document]
edvald: The main point is that you can currently run multiple containers in parallel, so we can't have a single fixed name. When we take the next step and the CLI is sending commands to a long-running service, instead of spinning up a container per command, then a fixed name will be how to do it.

garden-cli/run.go
drubin: Should this be `/root/.git` it's the users git config (but why do we need this?)

garden-cli/main.go
edvald: Different tradeoffs I guess, but I think the differing flows depending on the command, and the potentially long startup times are the worse choice. Especially when we'll be going over the network (although the implementation will be somewhat different then anyway).

garden-cli/run.go
edvald: I think it's just a mistake, we shouldn't need this at all.

garden-cli/run.go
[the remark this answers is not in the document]
drubin: Ah sorry I totally miss understood this. I assumed when we ran all the commands they check for an existing container and exec into it. 

Why do we do this instead of just starting up a container(if not already running) and exec'ing into it ? It seems like a simpler considering we are trying to watch events and state?

This would mean you would be able to behave more like client-> server model (albeit still with the exec instead of api)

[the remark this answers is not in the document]
edvald: We might do that as an optimization actually - should for sure be faster to start. Just wanted to get this working and then make that one of the next steps.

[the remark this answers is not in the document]
drubin: I believe it will result in simpler easier to maintain code that is easier to develop on. It's also how I believed we did it originally.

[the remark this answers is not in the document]
drubin: It also means we can test things with simple `docker exec -it` as a rapid verification method.

garden-cli/sync.go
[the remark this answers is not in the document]
drubin: I think we should keep an eye on this. `:cached` has huge performance wins and regardless of the consistency a few ms latency for improved cpu might make more sense.

garden-cli/run.go
eysi09: Yeah, that one just slipped in accidentally. My bad.

garden-cli/run.go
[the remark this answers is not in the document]
edvald: We did plan on doing it that way, just haven't gotten there yet. @eysi09 up to you if we change it in this PR or later.

.circleci/config.yml
```diff
@@ -23,9 +23,11 @@ jobs:
       - run:
           name: Install apt dependencies
           command: |
-            sudo add-apt-repository ppa:gophers/archive
             sudo apt-get update
-            sudo apt install rsync cmake libicu-dev pkg-config golang-1.11-go
+            sudo apt -y install software-properties-common
```
edvald: This installs _a lot_ of stuff, maybe we can go a cheaper route with this. @drubin can you help us get the CI passing here? 👼

drubin: I am busy fixing this actually on my Fork https://github.com/drubin/garden/pull/1.

I am splitting every thing out so we can also leverage caching and multi jobs running at the same time, it will simplify all this if we can actually leverage the pre-built images for what they were meant for.

drubin: If we *need* the CLI to pass I can always backport the building configs into this branch? but we still actively trying to fix a bunch of things so thought it could wait a few hours?

eysi09: Ok great. Just wanted to see if the flag above would get it running.

edvald: Oh no stress, let's just wait then until you're done with your thing. Btw, _very_ happy about everything you just said .)

.circleci/config.yml
```diff
@@ -6,45 +6,65 @@ version: 2
...
+      - run: export
+
       - run:
           name: Install apt dependencies
-          command: sudo apt install rsync cmake libicu-dev pkg-config
+          command: |
+            sudo apt-get update
+            sudo apt -y install software-properties-common
+            # Need software-properties-common for the add-apt-repository command
+            sudo add-apt-repository -y ppa:gophers/archive
```
drubin: You need another `sudo apt-get update`

.circleci/config.yml
drubin: I think we should leave this for now. No point trying to optimise it further when we are going to replace it all any way. if you add another `sudo apt-get update` after `add-apt-repository` it should pass.

garden-cli/docker/sync/Dockerfile
drubin: Is it possible  for the mount to not exist before the container starts up? Never seen that.

drubin: Also can we use absolute paths `/host-mount` it makes it more explicit. (although paths are relative to the working directly which would be `/` by default.

garden-cli/docker/sync/Dockerfile
drubin: I was coming here to suggest the same. It will make publishing easier too.

Out of interest why we we prefix folders with `garden-` it's in the garden repo after all?

edvald: Dunno, just feels clearer to me. No real reason beyond that.

garden-cli/kubernetes/bootstrap.go
```diff
@@ -0,0 +1,255 @@
+package kubernetes
```
drubin: Can we delete this file? There is no need to merge a commented out file?

garden-cli/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM alpine:3.8
```
drubin: Do you think we can move this out to `garden-sync` before merging? If its too much effort we can do it after.

I think it would clean things up technically we wont publish this as garden-cli its just a dependancy that the go cli will use just like garden-service?

edvald: +1 on that. These are basically two separate modules with different code bases.

PR #337, opened 2018-10-23
feat: experimental single-binary build via zeit/pkg

edvald:
_We may or may not want to merge this. I just had it half-baked and happened upon a solution to a problem I was struggling with, so I figured I'd get it done. We can have this as a fallback if our timeline with the Go CLI starts to stretch out._

To try this, install pkg (`npm install -g pkg`), run `npm run build`,
then `./garden-service/scripts/build-pkg.sh` and run garden via the
appropriate binary in `garden-service/dist`.

I needed to make a fork of the K8s JS client to get rid of the
remaining native dependency. There's an open PR to get that change into their next release.

Caveat: The macos build doesn't include the native fsevents binding,
so the npm package (or homebrew installation) would still be preferred
on mac.


garden-service/src/constants.ts
```diff
@@ -8,9 +8,11 @@
 
 import { resolve, join } from "path"
 
+export const isPkg = !!(<any>process).pkg
```
eysi09: Does this need to be exported?

PR #338, opened 2018-10-23
feat(k8s): print error logs when container fails to start

### LINKED ISSUE #137: kubernetes: Output logs from container when it crashes on startup
Currently we get a less-than-helpful `BackOff - back-off restarting failed container` when a container doesn't successfully start when deploying. 

A reasonable workaround is to run the troubled service with `garden run service <service-name>` to see what it spits out, but it would be better if we'd get the service log output when this particular error happens.

edvald:
Closes #137

### DIFF
--- a/garden-service/src/plugins/kubernetes/status.ts
+++ b/garden-service/src/plugins/kubernetes/status.ts
@@ -30,6 +30,8 @@ import { KubernetesProvider } from "./kubernetes"
 import { isSubset } from "../../util/is-subset"
 import { LogEntry } from "../../logger/log-entry"
 import { getContainerServiceStatus } from "./deployment"
+import { V1ReplicationController, V1ReplicaSet } from "@kubernetes/client-node"
+import dedent = require("dedent")
 
 export interface RolloutStatus {
   state: ServiceState
@@ -37,11 +39,14 @@ export interface RolloutStatus {
   lastMessage?: string
   lastError?: string
   resourceVersion?: number
+  logs?: string
 }
 
 interface ObjHandler {
   (api: KubeApi, namespace: string, obj: KubernetesObject, resourceVersion?: number): Promise<RolloutStatus>
 }
+
+const podLogLines = 20
 
 // Handlers to check the rollout status for K8s objects where that applies.
 // Using https://github.com/kubernetes/helm/blob/master/pkg/kube/wait.go as a reference here.
@@ -62,16 +67,11 @@ const objHandlers: { [kind: string]: ObjHandler } = {
   },
 
   ReplicaSet: async (api, namespace, obj) => {
-    const res = await api.core.listNamespacedPod(
-      namespace, undefined, undefined, undefined, true, obj.spec.selector.matchLabels,
-    )
-    return checkPodStatus(obj, res.body.items)
+    return checkPodStatus(obj, await getPods(api, namespace, (<V1ReplicaSet>obj).spec.selector.matchLabels))
   },
+
   ReplicationController: async (api, namespace, obj) => {
-    const res = await api.core.listNamespacedPod(
-      namespace, undefined, undefined, undefined, true, obj.spec.selector,
-    )
-    return checkPodStatus(obj, res.body.items)
+    return checkPodStatus(obj, await getPods(api, namespace, (<V1ReplicationController>obj).spec.selector))
   },
 
   Service: async (api, namespace, obj) => {
@@ -95,6 +95,7 @@ const objHandlers: { [kind: string]: ObjHandler } = {
 
 async function checkPodStatus(obj: KubernetesObject, pods: V1Pod[]): Promise<RolloutStatus> {
   for (const pod of pods) {
+    // TODO: detect unhealthy state (currently we just time out)
     const ready = some(pod.status.conditions.map(c => c.type === "ready"))
     if (!ready) {
       return { state: "deploying", obj }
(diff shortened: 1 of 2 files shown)


garden-service/src/plugins/kubernetes/status.ts
```diff
@@ -476,3 +485,27 @@ export function removeNull<T>(value: T | Iterable<T>): T | Iterable<T> | { [K in
...
+  return res.body.items
+}
+
+/**
+ * Get a formatted list of log tails for each of the specified pods. Used for debugging and error logs.
+ */
+async function getPodLogs(api: KubeApi, namespace: string, pods: V1Pod[]): Promise<string> {
+  const allLogs = await Bluebird.map(pods, async (pod) => {
+    const res = await api.core.readNamespacedPodLog(
+      pod.metadata.name, namespace, undefined, false, 1000000, undefined, false, undefined, 20,
```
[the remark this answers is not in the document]
edvald: I just put the 1MB limit for weirdo edge cases, there's also a parameter there that limits it to 20 lines.

garden-service/src/plugins/kubernetes/status.ts
```diff
@@ -476,3 +485,27 @@ export function removeNull<T>(value: T | Iterable<T>): T | Iterable<T> | { [K in
...
+    )
+    return `****** ${pod.metadata.name} ******\n${res.body}`
```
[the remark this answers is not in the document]
edvald: Yep, good idea, will add that.

garden-service/src/plugins/kubernetes/status.ts
```diff
@@ -353,7 +362,7 @@ export async function waitForObjects({ ctx, provider, service, objects, logEntry
     const now = new Date().getTime()
 
     if (now - startTime > KUBECTL_DEFAULT_TIMEOUT * 1000) {
-      throw new Error(`Timed out waiting for ${service.name} to deploy`)
+      throw new DeploymentError(`Timed out waiting for ${service.name} to deploy`, { statuses })
```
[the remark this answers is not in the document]
edvald: We don't handle that specifically. I can see how that comes out with this flow - I'm not exactly sure if/where K8s gives useful information to debug that type of error.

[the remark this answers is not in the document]
edvald: PR updated. I checked this particular case and K8s now returns a clearer error message when commands are invalid or not found. If we later come across other cases where the logs are not helpful, we can add those cases.

eysi09:
Not sure why that remote sources test is failing but I can take a look at that.

garden-service/src/plugins/kubernetes/status.ts
```diff
@@ -476,3 +506,27 @@ export function removeNull<T>(value: T | Iterable<T>): T | Iterable<T> | { [K in
...
+  return res.body.items
+}
+
+/**
+ * Get a formatted list of log tails for each of the specified pods. Used for debugging and error logs.
+ */
+async function getPodLogs(api: KubeApi, namespace: string, podNames: string[]): Promise<string> {
+  const allLogs = await Bluebird.map(podNames, async (name) => {
+    const res = await api.core.readNamespacedPodLog(
+      name, namespace, undefined, false, 1000000, undefined, false, undefined, podLogLines,
```
drubin: If there was an issue where some one writes a binary file to a log or something we are going to flood their terminal and make it unusable which would be worse than cutting off some erroneous log lines.

We have a limit on lines so 99% of the time this will be under 50kb? (thumb suck) making this like orders of magnitude bigger.

garden-service/src/plugins/kubernetes/status.ts
```diff
@@ -30,19 +30,24 @@ import { KubernetesProvider } from "./kubernetes"
...
   lastError?: string
   resourceVersion?: number
+  logs?: string
 }
 
 interface ObjHandler {
   (api: KubeApi, namespace: string, obj: KubernetesObject, resourceVersion?: number): Promise<RolloutStatus>
 }
 
+const podLogLines = 20
```
drubin: Question about stylistic type script?. Is making constant in uppercase a thing?

edvald: The uppercase style is not idiomatic in TS, but we still have some of those around. Might wanna make it consistent at some point.

garden-service/src/plugins/kubernetes/status.ts
edvald: It was more a safety measure in my mind, but I can lower the limit, sure.

garden-service/src/plugins/kubernetes/status.ts
```diff
@@ -180,6 +181,29 @@ export async function checkDeploymentStatus(
       }
       out.state = "unhealthy"
       out.lastError = `${event.reason} - ${event.message}`
+
+      // TODO: fetch logs for the pods in the deployment
+      if (event.involvedObject.kind === "Pod") {
+        const logs = await getPodLogs(api, namespace, [event.involvedObject.name])
+
+        out.logs = dedent`
+          <Showing last ${podLogLines} for the pod. Run the following command for complete logs:>
```
eysi09: Should the text be: `<Showing last ${podLogLines} log lines for the pod. Run...`?

garden-service/src/plugins/kubernetes/status.ts
```diff
@@ -180,6 +181,29 @@ export async function checkDeploymentStatus(
...
+          kubectl -n ${namespace} --context=${api.context} logs ${event.involvedObject.name}
+
+          ${logs}
+        `
+      } else {
+        const pods = await getPods(api, namespace, statusRes.spec.selector.matchLabels)
+        const logs = await getPodLogs(api, namespace, pods.map(pod => pod.metadata.name))
+
+        out.logs = dedent`
+          <Showing last ${podLogLines} per pod in this ${obj.kind}. Run the following command for complete logs:>
```
eysi09: Same as above, `<Showing last ${podLogLines} log lines per pod in this...`.

PR #340, opened 2018-10-25
refactor(go-cli): use docker sdk and keep containers running

eysi09:
An improved version of the Go CLI.

* Use Docker SDK where possible instead of shelling out
* Ensure sync is complete before running the command
* Keep the garden-service container running

### DIFF
--- a/garden-cli/constants.go
+++ b/garden-cli/constants.go
@@ -1,0 +1,10 @@
+package main
+
+// SyncImage is which docker image to use for syncing
+const SyncImage = "gardenengine/garden-sync:latest"
+
+// ServiceImage is which docker image to use for garden service
+const ServiceImage = "gardenengine/garden-service:latest"
+
+// ProjectPath is where to find the code inside ServiceImage
+const ProjectPath = "/project"
--- a/garden-cli/dockerutil/docker.go
+++ b/garden-cli/dockerutil/docker.go
@@ -1,12 +1,40 @@
 package dockerutil
 
 import (
+	"context"
 	"log"
 	"os"
 	"os/exec"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/api/types/volume"
+	"github.com/docker/docker/client"
+	"github.com/garden-io/garden/garden-cli/util"
 )
 
-// TODO: Use the Docker Golang SDK instead of shelling out here
+func RunContainer(
+	containerConfig container.Config, hostConfig container.HostConfig, containerName string,
+) (container.ContainerCreateCreatedBody, error) {
+	ctx := context.Background()
+	cli, err := client.NewEnvClient()
+	util.Check(err)
+
+	resp, err := cli.ContainerCreate(ctx, &containerConfig, &hostConfig, nil, containerName)
+	util.Check(err)
+
+	err = StartContainer(resp.ID)
+
+	return resp, err
+}
+
+func StartContainer(containerID string) error {
+	cli, err := client.NewEnvClient()
+	util.Check(err)
+
+	return cli.ContainerStart(context.Background(), containerID, types.ContainerStartOptions{})
+}
 
 func Exec(args []string, silent bool) error {
 	binary, err := exec.LookPath("docker")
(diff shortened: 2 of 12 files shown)


garden-cli/dockerutil/docker.go
```diff
@@ -26,31 +54,69 @@ func Exec(args []string, silent bool) error {
...
+
+	err = cli.ContainerStop(context.Background(), id, nil)
+	return err
+}
+
+func CreateVolume(volumeName string) error {
+	cli, err := client.NewEnvClient()
+	util.Check(err)
+
+	_, err = cli.VolumeCreate(context.Background(), volume.VolumeCreateBody{
```
drubin: Currently this doesn't seem to compile? 

```# github.com/garden-io/garden/garden-cli/dockerutil
dockerutil/docker.go:89:50: undefined: volume.VolumeCreateBody
Exited with code 2
```

garden-cli/run.go
```diff
@@ -2,51 +2,64 @@ package main
...
-		projectVolume = gitRoot
+		bindMounts := []string{
+			"/var/run/docker.sock:/var/run/docker.sock",
+			fmt.Sprintf("%s/.docker:/root/.docker", homeDir),
+			fmt.Sprintf("%s:/root/.garden", gardenHomeDir),
+			fmt.Sprintf("%s/.kube:/root/.kube", homeDir),
+		}
+
+		containerConfig := container.Config{
+			Image:      fmt.Sprintf("%s/%s:latest", DOCKER_REPO, SERVICE_IMAGE),
```
drubin: Can we please move (the tag) to the constants? I was thinking it might make more sense to just make this image configurable?

```
SERVICE_IMAGE = "gardenengine/garden-service:latest"
SYNC_IMAGE =  "gardenengine/garden-sync:latest"
```

I can't see why we would ever need to parse/split these? Also an image is a full entity changing the `DOCKER_REPO` (also I think docker calls it an org) would involve switching both images at the same time. 

This will allow us to later move these to "privately/internally hosted" images in future once we create a config system to move these out of the compiled version.

garden-service/Dockerfile
```diff
@@ -1,7 +1,13 @@
...
 
 WORKDIR /project
-ENTRYPOINT [ "garden" ]
+
+# The current implementation of Garden expects the Helm and OpenFaas binaries at these specific paths.
+ARG HELM_PATH=/.garden/tools/helm/778f4f359fd60ecb/
+ARG FAASCLI_PATH=/.garden/tools/faas-cli/826489a1731741fd/
+
+RUN mkdir -p ${HELM_PATH}
+RUN mkdir -p ${FAASCLI_PATH}
```
eysi09: Move the two `mkdir` steps to the `RUN apk add` step

garden-service/Dockerfile
```diff
@@ -1,7 +1,13 @@
 FROM node:10.11.0-alpine
 
 WORKDIR /project
-ENTRYPOINT [ "garden" ]
+
+# The current implementation of Garden expects the Helm and OpenFaas binaries at these specific paths.
+ARG HELM_PATH=/.garden/tools/helm/778f4f359fd60ecb/
```
eysi09: Check if Helm path hash needs updating.

garden-service/Dockerfile
```diff
@@ -11,13 +14,15 @@ RUN apk add --no-cache \
   git \
   openssl \
   rsync \
-  && curl -L https://storage.googleapis.com/kubernetes-helm/helm-v2.11.0-linux-amd64.tar.gz | tar xvz --strip-components=1 -C /usr/local/bin linux-amd64/helm \
-  && chmod +x /usr/local/bin/helm \
-  && helm init --client-only \
+  && mkdir -p ${HELM_PATH} \
+  && curl -L https://storage.googleapis.com/kubernetes-helm/helm-v2.10.0-linux-amd64.tar.gz | tar xvz -C ${HELM_PATH}  \
```
edvald: This should be v2.11.0

garden-cli/run.go
```diff
@@ -2,51 +2,64 @@ package main
...
+				Type:   mount.TypeVolume,
+				Source: volumeName,
+				Target: PROJECT_PATH,
+			},
 		}
-		projectVolume = volumeName
-	} else {
-		projectVolume = gitRoot
+		bindMounts := []string{
+			"/var/run/docker.sock:/var/run/docker.sock",
```
drubin: Should we also mount `~/.git` and `~/.ssh` all our examples assume public git projects I doubt this will be the case going forward.

drubin: Just for clarification If we mount `~/.git` and `~/.ssh` it will allow the container to pull down private git repo's (which I assume would be the biggest use case)

eysi09: Yeah I added a code comment to for clarification

garden-cli/constants.go
```diff
@@ -0,0 +1,5 @@
+package main
+
+const SYNC_IMAGE = "gardenengine/garden-sync:latest"
```
drubin: This isn't very `go` like. I think their convention is `const SyncContainer` I think we should try to stick to the go conventions.

PR #342, opened 2018-10-26
fix(git): error when running before first commit is made in repo (#324)

### LINKED ISSUE #324: Garden doesn't work until the first commit has been made
Fails with:

```
Command failed: git diff-index --name-only HEAD /Users/eysi/code/garden-io/sock-shop/services/front-end
fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

edvald:
Closes #324

### DIFF
--- a/garden-service/src/vcs/git.ts
+++ b/garden-service/src/vcs/git.ts
@@ -60,17 +60,31 @@ export class GitHandler extends VcsHandler {
       ]) || NEW_MODULE_VERSION
     } catch (err) {
       if (err.code === 128) {
-        // not in a repo root, return default version
+        // not in a repo root, use default version
         commitHash = NEW_MODULE_VERSION
+      } else {
+        throw err
       }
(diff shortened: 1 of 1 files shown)


garden-service/src/vcs/git.ts
```diff
@@ -60,17 +60,31 @@ export class GitHandler extends VcsHandler {
...
+    } catch (err) {
+      if (err.code === 128) {
+        // no commit in repo, use default version
+        commitHash = NEW_MODULE_VERSION
+      } else {
+        throw err
+      }
+    }
+
+    const newFiles = (await git("ls-files", ["--other", "--exclude-standard", path])).split("\n")
```
drubin: Will this not need to be put in a try catch too?

PR #351, opened 2018-10-31
Docs references

eysi09:


garden-service/src/commands/update-remote/modules.ts
```diff
@@ -35,8 +35,8 @@ export class UpdateRemoteModulesCommand extends Command<Args> {
   arguments = updateRemoteModulesArguments
 
   description = dedent`
-    Remote modules are modules that have a repositoryUrl field
-    in their garden.yml config that points to a remote repository.
+    This command updates remote modules, i.e. modules that have a \`repositoryUrl\` field
```
drubin: `This command` feels a little awkward when displaying help text I think the general context provides which command it's about?

Maybe? 
```
Updates remote modules, i.e. modules that have a \`repositoryUrl\` field...
```

thsig: Yeah, agree with that. E.g. the description shown in `git help status` (on my machine) starts with

> Displays paths that have differences between...

and not

> This command displays paths that have differences between...

garden-service/src/docs/commands.ts
```diff
@@ -19,7 +19,7 @@ import { describeParameters } from "../commands/base"
 
 export function generateCommandReferenceDocs(docsRoot: string) {
   const referenceDir = resolve(docsRoot, "reference")
-  const outputPath = resolve(referenceDir, "commands.md")
+  const outputPath = resolve(referenceDir, "command-reference.md")
```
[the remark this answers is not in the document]
eysi09: cc @ellenkorbes 

I changed the code to match the current file names. Alternatively we can revert the filename changes and leave the code as is.

garden-service/src/commands/update-remote/modules.ts
eysi09: Yes, agreed. All instances of `This command...` have been removed.

garden-service/src/docs/commands.ts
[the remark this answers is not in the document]
drubin: I think it makes sense to revert the file name changes it makes ugly urls. But let's do that in another PR?

drubin:
@eysi09 I think the changes are an improvement, but we need to fix the CI linter before merging.

Issue #353, opened 2018-10-31
Garden attempts to get status of K8s service before environment is prepared 


eysi09:
Running some Garden commands (e.g. `deploy` and `init`) when the `garden-system` hasn't been initialised throws `Got error from Kubernetes API - services "garden-system--ingress-controller-nginx-ingress-controller" not found`. 

This is because:

1. The commands above trigger a call to `prepareEnvironment` in (`src/actions.ts`).
2. The `prepareEnvironment` function calls `getEnvironmentStatus` before invoking the `prepareEnvironment` handler for the relevant plugins.
3. That eventually results in the `Helm` plugin calling `checkObjectStatus` (`kuberntetes/status.ts`) for `garden-system` K8s objects.
4. Calling `checkObjectStatus` throws a 404 error for missing service objects (they haven't been prepared cf. 1. above). 

At first glance, `getEnvironmentStatus` should be called after the `prepareEnvironment` handlers are invoked. However, the handlers depend on the return value from `getEnvironmentStatus`. So this one requires a bit more thought.

eysi09:
This issue can be worked around for now with #354.

edvald:
I actually think #354 addresses this correctly. Closing.

PR #354, opened 2018-10-31
fix(k8s): don't throw if api returns 404 when checking object status

eysi09:
Circumvents #353


garden-service/src/plugins/kubernetes/status.ts
```diff
@@ -301,10 +300,24 @@ export async function checkObjectStatus(
...
-      ? await handler(api, namespace, obj, prevStatus && prevStatus.resourceVersion)
+    let status: RolloutStatus
+    if (handler) {
+      try {
+        status = await handler(api, namespace, obj, prevStatus && prevStatus.resourceVersion)
+      } catch (err) {
+        // We handle 404s specifically since this might be invoked before some objects are deployed
+        // See: https://github.com/garden-io/garden/issues/353
+        // TODO: Figure out whether we'll need this check after the issue #353 above has been resolved
+        if (err.code && err.code === 404) {
```
thsig: Simplify to `if (err.code === 404) {` ?

eysi09: Oc, 'twas late.

PR #357, opened 2018-11-01
Cleaner service builds

drubin:
Removes dependency on git-submodules and simplifies garden-service dependancies.

We can either merge this now using our fork or wait for upstream to merge my PR there? Should still be this week, if not we can go forward with our fork. 

We just need to change 

`"@kubernetes/client-node": "kubernetes-cliet/javascript#hash",`

### DIFF
--- a/garden-service/package.json
+++ b/garden-service/package.json
@@ -22,7 +22,7 @@
     "static"
   ],
   "dependencies": {
-    "@kubernetes/client-node": "file:vendor/github.com/garden-io/javascript/node-client",
+    "@drubin/client-node": "0.7.1-rc1",
     "ansi-escapes": "^3.1.0",
     "async-exit-hook": "^2.0.1",
     "async-lock": "^1.1.3",
@@ -111,6 +111,7 @@
     "@types/unzip": "^0.1.1",
     "@types/unzipper": "^0.9.1",
     "@types/wrap-ansi": "^3.0.0",
+    "@types/ws": "^6.0.1",
     "chai": "^4.2.0",
     "gulp": "^4.0.0",
     "gulp-cached": "^1.1.1",
@@ -145,7 +146,6 @@
     "dist": "npm run build",
     "generate-docs": "gulp generate-docs",
     "integ": "./test/integ/run",
-    "preinstall": "./scripts/installVendorDeps.sh",
     "prepare": "npm run snyk-protect",
     "snyk-protect": "snyk protect",
     "test": "npm run clean && gulp test",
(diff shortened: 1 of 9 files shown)


garden-service/package.json
```diff
@@ -111,6 +111,7 @@
     "@types/unzip": "^0.1.1",
     "@types/unzipper": "^0.9.1",
     "@types/wrap-ansi": "^3.0.0",
+    "@types/ws": "^6.0.1",
```
eysi09: Why do we need this?

edvald: That may be in there by accident

drubin: It's required or should we push this to the k8s repo?

drubin: It fails to compile with out it. I am not really sure the best place to put these types. Do they go in each npm package or do we add then when we use them?

drubin: @eysi09 is this still an issue? I would really like to fix it properly.

drubin: This seems to be required to reference the project via github because it compiles the project while its pulling the source. in `prepare`

We are going to leave this here and once they release an official package (should be some time next week) I will remove it.

drubin:
The docker image (`garden-service`) doesn't contain the bundled `dist` files for `kubernetes/node-client`. :/ so going to hold off merging this till next week and they make a release or I will publish it to `@drubin/node-client`

eysi09:
@drubin Out of curiosity, do you know what the issue is? If I run `npm i` locally with `"@kubernetes/client-node": "garden-io/javascript#c65b666"` it installs the package along with the `dist` directory.

If I run the same command in a Docker container it leaves out the `dist` directory which is what I would expect to begin with since it's not a part of the git repo.

drubin:
@eysi09 I honestly have no idea. Thats why I said we should wait till next week. 

My assumption was we don't have the right build tools inside the container to run npm install even, I did try adding typescript and ts-node but it didn't seem to help.

drubin:
> If I run the same command in a Docker container it leaves out the dist directory which is what I would expect to begin with since it's not a part of the git repo. 

This gets compiled during install time. using the `prepare` script https://github.com/kubernetes-client/javascript/pull/143/files 

But then I read the docs on Friday https://docs.npmjs.com/misc/scripts 
>prepack: run BEFORE a tarball is packed (on npm pack, npm publish, and when installing git dependencies) 

> prepare: Run both BEFORE the package is packed and published, and on local npm install without any arguments (See below). This is run AFTER prepublish, but BEFORE prepublishOnly.

So still not sure if that should have  not be `prepack` but I didn't have time to test it.

drubin:
@eysi09 I have  published our package under `@drubin/0.7.1-rc1` which is current master at `45e0987`  as we decided once they publish their 0.7.1 finally we can easily switch out.

The tests pass and the docker image contains a built binary. 

Once we merge this we can work on all the other stuff around release automation.

PR #359, opened 2018-11-01
fix: fix broken doc links

thsig:
The URLs that were updated here were pointing to non-existent paths at docs.garden.io.

PR #360, opened 2018-11-02
Add automatic markdown link validator 

### LINKED ISSUE #330: Validate links in docs
Our docs contain a lot of links and these tend to break. Perhaps someone re-names an example project and forgets to change the corresponding links to it. 

We should have some way of automatically testing all the links in docs to ensure they stay valid.

drubin:
Adds automatic markdown linter for links. 

Can be run locally 

`npm run check-docs` 

Thinking we should split `check-docs` and `lint-docs` but kinda on the fence here

This also happens during CI now as well.

Closes #330 

Will rebase once #359  is merged.

### DIFF
--- a/bin/check-docs
+++ b/bin/check-docs
@@ -1,0 +1,6 @@
+#!/bin/bash -e
+
+# linter before git check so we can lint docs even if they are 'dirty'
+find docs -name '*.md' -type f -exec markdown-link-check  {} \;
+
+git diff --quiet HEAD -- docs/ || (echo 'generated docs are not up-to-date!' && exit 1)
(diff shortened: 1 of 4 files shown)


bin/check-docs
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -e
+
+# linter before git check so we can lint docs even if they are 'dirty'
+find docs -name '*.md' -not -path 'docs/reference/glossary.md' -type f -exec markdown-link-check  {} \;
```
eysi09: Why does it skip the Glossary?

drubin: This fails and I don't know where to link it to now with the updated scheme. 

`[✖] ../guides/configuration.md#project-configuration`

I meant to add a comment but forgot, if we fix this I will remove the glossary. It references a comment from doc-overhaul that we would deal with glossary later.

drubin: I found it it's now under `../using-garden/configuration-files#project-configuration` fixed and remove the exception.

PR #364, opened 2018-11-08
refactor(go-cli): replace unison with mutagen (wip)

eysi09:
First take on the [Mutagen](https://github.com/havoc-io/mutagen/) implementation. Tests are still missing so this should be considered work in progress. 

The previous Go CLI implementation used [Unison](https://www.cis.upenn.edu/~bcpierce/unison/) and did the syncing inside the sync container, similar to [this strategy](https://github.com/EugenMayer/docker-sync/wiki/8.-Strategies#native_osx-osx) that docker-sync uses. That however will not work on Windows. 

Therefore we opted for doing the syncing from the host and into the `garden-sync` container, using Mutagen as our synchronisation tool. The flow looks like this:

1.  Create a named Docker volume unique to the project
2. Create a `garden-sync` container unique to the project and mount it with the Docker volume
3. Use Mutagen to sync from the host into the Docker volume
4. Create a `garden-service` container unique to the project
5. Run the command inside the `garden-service` container

### DIFF
--- a/garden-cli/main.go
+++ b/garden-cli/main.go
@@ -9,77 +9,122 @@ import (
 
 	"github.com/garden-io/garden/garden-cli/dockerutil"
 	"github.com/garden-io/garden/garden-cli/util"
+	"github.com/pkg/errors"
 )
+
+type Dependency struct {
+	bin          string
+	errorMessage string
+}
 
 // A CLI for running Garden commands in a garden-service container.
 //
 // For a new project the flow is as follows:
 //
-// 	1. Create a named garden-sync volume for the project
-// 	2.1 Create and start a garden-sync container, mount the named volume and bind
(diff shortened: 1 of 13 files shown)


garden-cli/main.go
```diff
@@ -9,77 +9,127 @@ import (
...
+		},
+		{
+			bin: "kubectl",
+			errorMessage: "Could not find kubectl - " +
+				"Garden requires a configured local Kubernetes cluster and for kubectl to be configured to access it",
+		},
+		{
+			bin:          "mutagen",
+			errorMessage: "Could not find mutagen - Garden requires mutagen to be installed in order to run.",
+		},
```
edvald: I think it would be nice to bundle this. Also, do we still need kubectl here?

PR #365, opened 2018-11-09
WIP: chore(release): create releases from ci

drubin:
Use circleci to create releases for both master and tags.

I will fully update the release docs but I wanted to get this out there for feedback so long. 

**To b updated just before merging**

- [ ] Change backage name back to `garden-cli`
- [ ] Re enable commit workflow 
- [ ] Downgrade version to match current master

We now have 3 flows

## Every Commit
Mostly the same as before except we also build docker images (verifies they still build).

1. We build and test the `garden-service` 
1. Build docker image `garden-sync`
1. Build docker `garden-service`
1. `garden-cli` unit tests

screenshot 2018-11-08 13 40 57

## Master 
1. We build/compile  the `garden-service`
1. Build docker image `garden-sync` and release it as `:master`
1. Build docker `garden-service` and release it as `:master`
1. Build `garden-cli` with goreleaser and save Artifacts to circleci (all platforms)
1. Build our zeit/pkg package of `garden-service` save Artifacts  to circleci (linux and windows)

screenshot 2018-11-08 15 54 50

## Tag's 

1. We build/compile  the `garden-service`
1. Build docker image `garden-sync` and release it as `:$TAG`
1. Build docker `garden-service` and release it as `:TAG`
1. Publish `garden-service` as npm package with `$TAG`
1. Build `garden-cli` with goreleaser and save Artifacts to circleci (all platforms)
1. Build our zeit/pkg package of `garden-service` save Artifacts  to circleci (linux and windows)

screenshot 2018-11-09 11 31 01

## Outstanging things 

1. Change log auto generation (I feel like this should happen before we push the tag)
1. Updating brew formulars (Would require circle to be able to write to another git repo..)
1. Automatically updated these to github releases (I think for now we can download them and upload them, since they were built from ci, and checksums created there)


.circleci/config.yml
```diff
@@ -38,6 +85,65 @@ commands:
...
+        description: Directory to build in
+        type: string
+        default: .
+    steps:
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run:
+          name: Build Docker Image
+          command: |
+            # Just build it with the approprate tag
```
eysi09: appropriate

.circleci/config.yml
```diff
@@ -38,6 +85,65 @@ commands:
...
+    steps:
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - deploy:
+          name: Release docker image
+          command: |
+            # These values are inject from context docker
+            echo $DOCKER_PASS | docker login -u $DOCKER_USER --password-stdin
+            # Switches between git tag and master for releases
+            # To make a bit smarter later
```
eysi09: _TODO_, right? Sorry about the nit pick but I guess that's the only non-trivial comment typo.

eysi09:
Looks good! Regarding the changelog, we're currently using [git-chglog](https://github.com/git-chglog/git-chglog). It's way better than what we had previously ([conventional-changelog](https://github.com/conventional-changelog/conventional-changelog/)). However, they both depend on the latest tag being available when generating the changelog.

For instance, they both generate a link to compare the latest release with the one before it on Github. If the changelog were generated before tagging that link would be broken until the tag is created. Obviously not a big deal but I guess it highlights the idea that the changelog should come after the tag (at least according to the library authors).

I didn't see anything better out there when I made the change to git-chglog but maybe I missed something.

.circleci/config.yml
drubin: It wasn't supposed to be a todo actually but I will just remove it. the "smarter" part was switching between git tag and master.

garden-service/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "garden-cli",
-  "version": "0.8.0-rc2",
+  "name": "@drubin/garden-cli",
+  "version": "0.8.0-rc10",
```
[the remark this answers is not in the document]
drubin: See the 1st bullet point in the PR description I was just leaving this here incase there was feedback I needed to retest

package.json
```diff
@@ -55,9 +55,9 @@
     "preversion": "npm test",
     "release": "./bin/publish",
     "release-canary": "./bin/publish-canary",
-    "test": "npm run lint && npm run clean && gulp test",
-    "ci-build": "gulp garden-service:build",
-    "ci-test": "gulp test"
+    "test": "gulp test",
```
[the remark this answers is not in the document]
drubin: Oops I didn't notice notice this change would affect that.

I have mixed feelings on this.. It feels odd to have this lint/test on every push considering many of these commits don't actually make code changes. Surely we trust our devs to run tests were appropriate (and for every thing else we have CI?) I mostly run push with `--no-verify` since almost all my changes have not touched any thing that would affect tests.  

But I will add it back for now since the point of this wasn't to change the entire workflow.

garden-service/bin/build-pkg.sh
```diff
@@ -29,13 +29,16 @@ rm -rf linux-amd64
 mkdir linux-amd64
 mv garden-cli-linux linux-amd64/garden
 cp -r ../tmp/dist/static linux-amd64
-tar -czf garden-${version}-${commit_hash}-linux-amd64.tar.gz linux-amd64
-
-echo "  -> win-amd64"
```
drubin: Incase this wasn't clear this was to match the go binary and the general standard of using `windows-amd64` for release binaries (I guess this come's from Go builds)

Also Removed the hash's because our public releases should be easily scriptable and predictable names using a hash doesn't give us this and git tag should be enough.

.circleci/config.yml
```diff
@@ -59,60 +165,178 @@ jobs:
...
+    steps:
+      - checkout
+      - npm_install
+      - *attach-node-workspace
+      - run:
+          name: Authenticate with registry
+          command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > ~/.npmrc
+      - run:
+          name: Publish package
+          command: cd garden-service && npm publish
```
drubin: @eysi09 Should we maybe add a `--tag=beta` if the version contains RC? 

Then we can easily just git tag things with RC and it wont affect latest?

Issue #276, opened 2018-09-11
Workflows / Tasks 📋


edvald:
## Objective and motivation

Most backend systems need some degree of imperative bootstrapping before they're considered functional.
In the abstract this can be many types of things, perhaps the most common being database migrations and data
fixtures, but one can imagine other tasks like generating keys, pulling information from external systems etc.

## Design considerations

Rather than attempt to implement fine-grained support for those specific types of tasks, we should focus on
workflows in the abstract, as part of the development workflow and runtime dependency graph, and also consider
how the more fine-grained tasks could be further facilitated by dedicated plugins.

In terms of flows, users need to be able to define tasks to be executed in certain sequence, and often in relation
to service deployment.

In many cases the user only cares about the execution order, but in some cases they may also need outputs from one
task to be passed on to other tasks, and even services (as files, environment variables etc.). For this we can look
to how some CI systems are configured (e.g. CircleCI), and workflow engines like [Argo](https://argoproj.io).

## Design proposal

I propose the addition of _tasks_ as a primitive in our plugin framework and API. _(This notably differs from the
notion of task in the terms of our TaskGraph, but that's internal implementation detail and we may simply rename that
component.)_

A task is configured similarly to services, and plugins are able to define and run tasks if

1) The given module type's schema allows for configuration of tasks.
2) The _parseModule()_ handler for the module type correctly populates the output `ModuleConfig` via a new
   `taskConfigs` key (analogous to `testConfigs` and `serviceConfigs`).
3) They provide a `runTask()` action handler for the module type.

Note that 3) will often be decoupled from 1) and 2), as in the case of the `container` plugin, and the `kubernetes`
plugin.

The `runTask()` handler will be very similar to the `runModule()` handler, but I feel it's wise to separate them
in case the plugin needs to behave differently between them. Plugins can of course re-use most of the code internally,
as appropriate.

I imagine configuring `container` tasks would look something like this:

```yaml
module:
  name: db
  type: container
  services:
    - name: db
      dependencies:
        - db-bootstrap
  ...
  tasks:
    - name: db-prepare
      command: [npm, run, prepare]
    - name: db-migrations
      command: [npm, run, notify]
      dependencies:
        - db
  ...
```

Here the execution flow would be
1) Run `db-prepare` and wait for result.
2) Deploy `db` and wait until healthy.
3) Run `db-migrations` and wait for result.

First thing to note here is that *tasks share a namespace with services*. This is perhaps debatable, but otherwise
[...]

PR #367, opened 2018-11-11
feat: tasks/workflows implemented

thsig:
# Summary

Added tasks to the framework. In a future PR (hopefully soon), the task classes (`BuildTask` etc.) will be renamed, to disambiguate from the task concept in the internal code base.

Tasks are defined under the new `tasks` field in a module's `garden.yml`.

For background and motivation, check out the initial design proposal: https://github.com/garden-io/garden/issues/276

This PR also introduces the `DependencyGraph` class.

# User-level semantics

In short, a task is a **user-defined command** — or, more generally, a **one-off computation** (the specifics depend on the provider) — which belongs to a module, and is executed in the context of its module. Tasks are integrated into the framework's dependency system, and may depend on other tasks or services.

Like services, tasks have an implicit build dependency on their parent module.

Task names and service names inhabit the same namespace, so they must be mutually unique (this is checked during config validation).

A name listed under the `dependencies` field of a service, task or test spec in a module's `garden.yml` can be a service name or a task name.

When a service _S_ with task dependencies is deployed with hot reloading enabled, those tasks will be run when it is initially deployed, but will not be re-run when that _S_'s module is hot-reloaded.

Furthermore, if the project has services that depend on _S_, which in turn were deployed with hot reloading enabled, and have their own task dependencies, those task dependencies will be ignored and thus not re-run when _S_'s module is hot-reloaded (unless included by one or more service dependants of _S_ that were not deployed with hot reloading enabled, and which also depend on those tasks).

In simpler terms, hot reload-enabled services' task dependencies are ignored, except when those services are initially deployed.

These calculations of task dependencies of a hot-reloaded service's service dependants (which, in turn, may be deployed with hot reloading enabled) when its module's sources change weren't compatible with the simplifications/assumptions underlying the old `computeAutoReloadDependants` approach. The `DependencyGraph` class addresses this.

# Implementation

A new task class, `TaskTask` (cringe, I know — this is to be renamed as soon as possible, once certain pending changes on other branches have been merged), is introduced, along with a new `run` command type (`RunTaskCommand`), and the appropriate framework-level method for running tasks (`runTask`).

The actual execution/running of a task doesn't involve much new code at all. For k8s, `runTask` just calls `runModule` with the appropriate arguments. For the generic plugin type, `runTask` runs the task's command from the module directory on the host machine.

Most of the changes involve framework-level types, specs, methods and config handling — and the aforementioned changes to the framework's dependency/dependant calculation logic.

## `DependencyGraph`

Properly integrating tasks into the framework's dependency/dependant logic required a more fine-grained method for resolving dependants when a module's sources change, particularly when one or more services are deployed with hot reloading enabled.

The `DependencyGraph` class was introduced primarily to address these new requirements, but also to centralize the recursive "graph-traversal" logic for computing dependencies/dependants that was previously implemented ad-hoc in various parts of the codebase. These changes should also make it easier to change the framework's dependency/dependant semantics in the future.

`DependencyGraph` is fundamentally a graph data structure that represents all the **static dependency/dependant relationships** described in a project's `garden.yml` files. It's not a replacement for the runtime dependency/dependant calculations performed by the framework (which need to be aware e.g. of command options). Rather, it is a more sophisticated, centralized way for commands, tasks (as in `BuildTask` etc.) and other parts of the framework to query the dependency/dependant relationships described in the project's module-level `garden.yml` files.

# Review notes

The following should definitely be reviewed closely — that said, I'm sure there are other aspects of these changes that merit close examination as well.

- `DependencyGraph`: https://github.com/garden-io/garden/blob/workflows/garden-service/src/dependency-graph.ts
- `TaskTask`: https://github.com/garden-io/garden/blob/workflows/garden-service/src/tasks/task.ts
- `RunTaskCommand`: https://github.com/garden-io/garden/blob/workflows/garden-service/src/commands/run/task.ts
- `DeployTask`: https://github.com/garden-io/garden/blob/workflows/garden-service/src/tasks/deploy.ts
- `tasks/helpers.ts`: https://github.com/garden-io/garden/blob/workflows/garden-service/src/tasks/helpers.ts

# Open questions
- How verbose should task execution be? Should there be a CLI option for the relevant commands to log stdout from task execution to the terminal?
- Currently, the `RunTaskCommand` logs task execution output to the terminal, but `TaskTask` does not (since they're generally run as part of the build/deploy/test cycle, and too much verbosity in that context might be distracting and repetitive). Do those defaults make sense?
- Also, should we log stdout from task execution to a `garden-tasks.log` file in project root?

# Example

To illustrate (taken from the `tasks` example project https://github.com/garden-io/garden/tree/workflows/examples/tasks):

```yaml
module:
  name: hello
  description: Greeting service
  type: container
  services:
    - name: hello
      command: [npm, start]
      # ...
      dependencies:
        - node-migration
  tasks:
    - name: node-migration
      # This migration creates the users table.
      command: [knex, migrate:latest]
      dependencies:
        - postgres
```

```yaml
module:
  description: postgres container
  type: container
  name: postgres
  image: postgres:9.4
  services:
    - name: postgres
    # ...
```

```yaml
module:
  name: user
  description: User-listing service written in Ruby
  type: container
  services:
    - name: user
      command: [ruby, app.rb]
      # ...
      dependencies:
        - ruby-migration
  tasks:
    - name: ruby-migration 
      command: [rake, db:migrate]
      dependencies:
        # The users table has to exist before we use ruby-migration
        # to insert records into it.
        - node-migration
```
When this project is `garden deploy`-ed, `node-migration` is run once `postgres` is up. Once `node-migration` finishes, `hello` is deployed and `ruby-migration` is run. When `ruby-migration` finishes, `user` is deployed.


drubin:
Really nice detailed overview thank you! 

Minor questions about high level design before digging into details

```
  tasks:
    - name: node-migration
      # This migration creates the users table.
      # Sleep to give postgres time to get ready.
      command: ["sleep 5 && knex migrate:latest"]
      dependencies:
        - postgres-service
```

Why do we need to manually wait for postgres to get ready? Surely this `dependencies` takes into account if the service is healthy and passes the health check? 

If we need a sleep here because we aren't using the built in ready checks we already have we are going to have race conditions where things fail some times and some times just works. 

Our naming convention means that tasks/services/modules all have the same scope and they now need to be prefixed/suffixed  did we think about automatically doing this for users? `service/postgres`  these service names should not have to include a `-service` at the end since we k8s uses this for DNS / service discovery or at least for tasks not being in the same scope.

Most applications should work out of the box when npm connects to `postgres` dns name with out need to manually configure ENV configurations.  (This is the standard workflow for docker-compose / k8s/containers)

docs/reference/command-reference.md
```diff
@@ -446,9 +446,9 @@ This is useful for debugging or ad-hoc experimentation with modules.
 
 Examples:
 
-    garden run module my-container           # run an ad-hoc instance of a my-container container and attach to it
-    garden run module my-container /bin/sh   # run an interactive shell in a new my-container container
-    garden run module my-container --i=false /some/script  # execute a script in my-container and return the output
+    garden run module my-container                                   # run an ad-hoc instance of a my-containercontainer and attach to it
+    garden run module my-container /bin/sh                           # run an interactive shell in a newmy-container container
```
edvald: Missing space/breaks here

docs/reference/commands.md
```diff
@@ -494,6 +494,32 @@ Examples:
 | -------- | ----- | ---- | ----------- |
   | `--force-build` |  | boolean | Force rebuild of module.
 
+### garden run task
+
+Run a task (in the context of its parent module).
+
+This is useful for re-running tasks on the go, for example after writing/modifying database migrations.
```
edvald: "on the go" seems a little odd. "ad-hoc" maybe?

docs/reference/config.md
```diff
@@ -646,34 +679,35 @@ module:
...
-  # are passed to the `--hot-reload` option in a call to the `deploy` or `dev` command.
+  # A list of tasks that can be run from this container module. These can be used as dependencies
+  # for services (executed before the service is deployed) or for other tasks.
   #
   # Optional.
-  hotReload: 
-    # Specify one or more source files or directories to automatically sync into the running
-    # container.
+  tasks: 
+    # The task specification for a generic module.
```
edvald: Need to change the phrasing here to not say "generic module".

examples/tasks/hello-service/garden.yml
```diff
@@ -0,0 +1,24 @@
...
+          port: http
+      dependencies:
+        - node-migration
+  tests:
+    - name: unit
+      command: [npm, test]
+  tasks:
+    - name: node-migration
+      # sleep to give postgres time to get ready
+      command: ["sleep 5 && knex migrate:latest"]
```
edvald: Would be better to use a proper health check on the postgres container (see https://medium.com/@nieldw/kubernetes-probes-for-postgresql-pods-a66d707df6b4 for example)

garden-service/src/actions.ts
```diff
@@ -300,12 +311,22 @@ export class ActionHelper implements TypeGuard {
 
   //endregion
 
+  //===========================================================================
+  //region Workflow Methods
+  //===========================================================================
+
+  async runWorkflow(params: WorkflowActionHelperParams<RunWorkflowParams>): Promise<RunWorkflowResult> {
```
edvald: It would feel clearer to me to call this `runTask` in the plugin interface, since a Workflow would indicate a sequence of actions, and we might at some point expose something like that at the engine level.

garden-service/src/config/workflow.ts
```diff
@@ -0,0 +1,65 @@
...
+  joiIdentifier,
+} from "./common"
+
+export interface WorkflowSpec { }
+
+export interface BaseWorkflowSpec extends WorkflowSpec {
+  name: string
+  dependencies: string[]
+  timeout: number | null
+}
```
edvald: I'm pondering the naming a bit... Going a bit back and forth in my head but I feel like we should prioritize clarity in the plugin and user interface, and go for "task" rather than "workflow" there. Internally we can refer to a task as a`WorkflowTask` to disambiguate. What do you think?

garden-service/src/dependency-graph.ts
```diff
@@ -0,0 +1,364 @@
...
+  }
+
+  async getDependenciesForMany(
+    nodeType: DependencyGraphNodeType, names: string[], recursive: boolean, filterFn?: DependencyRelationFilterFn,
+  ): Promise<DependencyRelations> {
+    return this.toRelations(flatten(
+      names.map(name => this.getDependencyNodes(nodeType, name, recursive, filterFn))))
+  }
+
+  async getDependantsForMany(
```
edvald: All of these methods should have comments to explain the distiction - this feels a bit confusing.

garden-service/src/plugins/generic.ts
```diff
@@ -155,12 +167,51 @@ export async function testGenericModule({ module, testConfig }: TestModuleParams
...
+    const commandResult = await execa.shell(
+      command.join(" "),
+      {
+        cwd: module.buildPath,
+        env: { ...process.env, ...mapValues(module.spec.env, v => v.toString()) },
+      },
+    )
+
+    result["completedAt"] = new Date()
+    result["output"] = commandResult.stdout + commandResult.stderr
```
edvald: We shouldn't use bracket notation on typed objects.

garden-service/src/types/plugin/plugin.ts
```diff
@@ -41,7 +46,7 @@ import {
   runModuleParamsSchema,
   testModuleParamsSchema,
   getTestResultParamsSchema,
-  publishModuleParamsSchema,
+  publishModuleParamsSchema, getWorkflowStatusParamsSchema, runWorkflowParamsSchema,
```
edvald: Formatting

garden-service/src/types/plugin/plugin.ts
```diff
@@ -61,17 +66,12 @@ import {
   hotReloadResultSchema,
   runResultSchema,
   ServiceActionOutputs,
+  WorkflowActionOutputs,
   setSecretResultSchema,
   testResultSchema,
   validateModuleResultSchema,
-  publishModuleResultSchema,
+  publishModuleResultSchema, workflowStatusSchema, runWorkflowResultSchema,
```
edvald: Formatting

garden-service/src/util/util.ts
```diff
@@ -378,6 +382,15 @@ export function pickKeys<T extends object, U extends keyof T>(obj: T, keys: U[],
   return picked
 }
 
+export function throwOnMissingNames<T extends ObjectWithName>(names: string[], entries: T[], description: string) {
```
edvald: Doesn't appear to be used?

garden-service/src/actions.ts
edvald: Or perhaps `runWorkflowTask` if we want to be very clear.

edvald:
Mostly looks great! Just some naming concerns, and a few smaller comments.

docs/reference/commands.md
thsig: Yeah, that's better.

examples/tasks/hello-service/garden.yml
thsig: Yeah, true that. Will make that change.

docs/reference/config.md
thsig: Should we then make the same change to `genericTestSchema` and `genericModuleSchema`?

Maybe we should just replace these with blank strings, since _"A task specification"_ etc. is a bit redundant.

garden-service/src/util/util.ts
thsig: It's used in a couple of places in the `Garden` class:

https://github.com/garden-io/garden/blob/e95695895a07dd475d34c2826a9637236874eb01/garden-service/src/garden.ts#L695

https://github.com/garden-io/garden/blob/e95695895a07dd475d34c2826a9637236874eb01/garden-service/src/garden.ts#L710

garden-service/src/dependency-graph.ts
thsig: Added some comments here:

https://github.com/garden-io/garden/pull/367/commits/46d4d9ab7269449fc2876b0a020b58e7c57ac5e5#diff-c4922521099309baabe316ccf4f1c28d

Are these satisfactory now, or do you see something that needs clarification?

thsig:
> Really nice detailed overview thank you!
> 
> Minor questions about high level design before digging into details
> 
> ```
>   tasks:
>     - name: node-migration
>       # This migration creates the users table.
>       # Sleep to give postgres time to get ready.
>       command: ["sleep 5 && knex migrate:latest"]
>       dependencies:
>         - postgres-service
> ```
> Why do we need to manually wait for postgres to get ready? Surely this `dependencies` takes into account if the service is healthy and passes the health check?
> 
> If we need a sleep here because we aren't using the built in ready checks we already have we are going to have race conditions where things fail some times and some times just works.
> 
> Our naming convention means that tasks/services/modules all have the same scope and they now need to be prefixed/suffixed did we think about automatically doing this for users? `service/postgres` these service names should not have to include a `-service` at the end since we k8s uses this for DNS / service discovery or at least for tasks not being in the same scope.
> 
> Most applications should work out of the box when npm connects to `postgres` dns name with out need to manually configure ENV configurations. (This is the standard workflow for docker-compose / k8s/containers)

Thanks!

Just noting, since you mentioned this, that the `tasks` example project has now been modified to use a proper health check on the `postgres` service. @drubin

garden-service/test/data/test-project-b/module-b/garden.yml
```diff
@@ -14,4 +14,4 @@ module:
   build:
     command: [echo, B]
     dependencies:
-      - module-a
+      - module-a
```
drubin: Missing new line.

thsig: Good catch. In general, our `fix-format` npm script doesn't touch on `.yml` files. We should find a good library for that and add it to `fix-format`.

drubin: https://editorconfig.org/ :) we should add it. I will add it tomorrow.

garden-service/src/tasks/task.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    let result: RunTaskResult
+    try {
+      result = await this.garden.actions.runTask({
+        task,
+        logEntry,
+        runtimeContext,
+        interactive: false,
+      })
+    } catch (err) {
```
eysi09: Should we print the error message to the console if the task fails (instead of just writing it to the `error.log`)?

For example, this:
```yaml
  tasks:
    - name: test
      command: [foo]
```
prints:
```sh
Failed running task test in module node-service. Here is the output:
————————————————————————————————————————————————————————————————————————————————
kubectl exited with code 127
————————————————————————————————————————————————————————————————————————————————

1 deploy task(s) failed!
```

However, in `error.log` we can see the following:
```
 29 code: 127
 30 output: |
 31   --namespace=garden--default--simple-project: line 1: foo: not found
 32   pod default/run-node-service-1542720590485 terminated (Error)
```

eysi09:
FYI: Did some manual "regression" testing like we spoke about:

`gatsby-hot-reload`: hot reload functionality
`sock-shop-mono`: tested changing the code while `garden dev` was running
`local-tls`: tested calling the endpoint via https
`remote-sources`: tested deploying and linking/unlinking modules
`hello-world`: tested breaking the integration test while `garden dev` was running

Didn't notice anything unusual.

cc @drubin

garden-service/src/tasks/task.ts
eysi09: ...so presumably, we could print some thing like this:

```
Failed running task test in module node-service. Here is the output:
———————————————————————————————————————————
kubectl exited with code 127  

foo: not found
————————————————————————————————————————————

1 deploy task(s) failed!
```

docs/reference/config.md
edvald: Or we can just phrase it more concretely than "specification". Like, "a list of tasks that can be run in this module", or something to that effect.

garden-service/src/dependency-graph.ts
edvald: This seems good to me

garden-service/src/tasks/task.ts
edvald: 👍

garden-service/test/src/tasks/helpers.ts
```diff
@@ -0,0 +1,230 @@
...
+      expect(await sortedBaseKeysWithDependencies(tasks)).to.eql([
+        "build.build-dependency",
+        "build.good-morning",
+        "deploy.good-morning",
+        "push.good-morning",
+        "task.good-morning-task",
+      ].sort())
+    })
+
+    describe("returns the correct set of tasks for the changed module and its dependants", () => {
```
edvald: This should be `it(...)`, right?

edvald: Ah hmm, I think I get it now, but it is a little confusing. Maybe this should be `context(...)` instead with a bit of a clearer description?

thsig: Good point. Just pushed a change along those lines — is this fine now?

garden-service/src/plugins/kubernetes/kubectl.ts
```diff
@@ -52,25 +52,25 @@ export class Kubectl {
...
+      opts.push(`--kubeconfig=${this.configPath}`)
     }
 
     if (tty) {
-      ops.push("--tty")
+      opts.push("--tty")
     }
 
-    return ops.concat(args)
+    return args.concat(opts)
```
drubin: This doesn't seem to work if your default kubeconfig namespace is not the same as the project namespace. This is expected behaviour if you are running `garden dev` for the first time because garden creates that namespace. 

The issue here is that `--namespace=garden--default--tasks --context=docker-for-desktop` are being appended AFTER the command. while knex ignores these other commands may not be so linent 

The logic inside the kubectl.ts file should split generic args and command args or be smarter about the way it parses these things. 

Side note (for future) I think it would be better to avoid running tasks via kubectl and rather as once-off jobs this way we can use the same library functions instead of string/command manipulation. 

The Jobs api also has retries and ability to debug them after they have completed we can also then use this same logic to define repeating jobs that happen every x minutes/seconds.

```Error Details:
cmd: kubectl
args:
  - run
  - run-hello-service-1542800925274
  - '--image=hello-service:b2ffe410d1'
  - '--restart=Never'
  - '--command'
  - '--rm'
  - '-i'
  - '--quiet'
  - '--env=GARDEN_VERSION=b2ffe410d1'
  - '--env=GARDEN_SERVICES_POSTGRES_SERVICE_HOST=postgres-service'
  - '--'
  - /bin/sh
  - '-c'
  - 'knex migrate:latest'
  - '--namespace=garden--default--tasks'
  - '--context=docker-for-desktop'```

```
Using environment: development
  Error: getaddrinfo ENOTFOUND postgres-service postgres-service:5432
      at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:67:26)
  pod default/run-hello-service-1542800925274 terminated (Error)
```

File: examples/tasks/README.md [hunk 16]
```diff
@@ -0,0 +1,34 @@
...
+
+Garden takes care of deploying the project's services and running the project's tasks in the specified dependency order:
+
+When this project is `garden deploy`-ed, `node-migration` is run once `postgres-service` ([defined here](https://github.com/garden-io/garden/tree/workflows/examples/tasks/postgres/garden.yml)) is up.
+
+Once `node-migration` finishes, `hello-service` is deployed and `ruby-migration` is run. When ruby-migration finishes, `user-service` is deployed.
+
+## Usage
+
+The simplest way to see this in action is to run `garden deploy` or `garden dev` in the project's top-level directory.
```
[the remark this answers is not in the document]
drubin: Something like `garden call hello-service/hello`

PR #370, opened 2018-11-14
fix: skip build task during hot-reloading

thsig:
Previously, the module being hot-reloaded was rebuilt following the hot
reload operation.

This is an unnecessary step, and is removed in this commit to reduce CPU
load.


drubin:
I assume we never actually took into account if the path being changed was in the hot reload paths? 

So this would always rebuild the containers but not actually update the deployment ever?

thsig:
Yeah, it wasn't supposed to update the deployment. The rationale for rebuilding on hot reload was to have the container ready for the next time it was deployed without hot reloading enabled. There's really no good reason for that, in hindsight, and it also ends up creating too many images.

PR #371, opened 2018-11-14
fix: fixes to hot reload source/target handling

drubin:
Fixes issues with handling subpaths as hotreload sync targets with differences between the way `cp` and `rysync` and `k8s` handle the paths. 

Solves issues with trying to sync `src` into `/app/src` and other edge cases where hotreloading would only work after a first change was made to the source. 

Also originally both the source and target needed to be in the exact correct format for things to work regarding trailing slashes. 

@thsig I also made 2 small changes that should have zero impact on code but really bugged me.

1. Changed the sync container name to not include `-container` it's annoying to suffix things by what they are.
2. Changed the volume to be called `garden-sync` so removed `-volume` and the `-${service-name}` this volume isn't a global variable its only shared with in this deployment so is there a reason to add the service name to it?  

As far as I know with garden services result in multiple deployments + services?

### DIFF
--- a/docs/reference/config-files-reference.md
+++ b/docs/reference/config-files-reference.md
@@ -663,11 +663,15 @@ module:
         # top-level directory. Must be a relative path if provided. Defaults to the module's
         # top-level directory if no value is provided.
         #
+        # Example: "src"
+        #
         # Optional.
         source: .
 
         # POSIX-style absolute path to sync the directory to inside the container. The root path
         # (i.e. "/") is not allowed.
+        #
+        # Example: "/app/src"
         #
         # Required.
         target:
--- a/garden-service/src/plugins/kubernetes/deployment.ts
+++ b/garden-service/src/plugins/kubernetes/deployment.ts
@@ -7,7 +7,7 @@
  */
 
 import deline = require("deline")
-import { resolve } from "path"
+import { resolve, normalize } from "path"
 import {
   extend,
   get,
@@ -16,9 +16,8 @@ import {
   toPairs,
 } from "lodash"
 import { DeployServiceParams, GetServiceStatusParams, PushModuleParams } from "../../types/plugin/params"
-import { Module } from "../../types/module"
 import { RuntimeContext, Service, ServiceStatus } from "../../types/service"
-import { helpers, ContainerModule, ContainerService } from "../container"
+import { helpers, ContainerModule, ContainerService, SyncSpec } from "../container"
 import { createIngresses, getIngresses } from "./ingress"
 import { createServices, RSYNC_PORT, RSYNC_PORT_NAME } from "./service"
 import { waitForObjects, compareDeployedObjects } from "./status"
@@ -399,15 +398,18 @@ function configureVolumes(deployment, container, spec): void {
 */
 function configureHotReload(deployment, container, serviceSpec, moduleSpec, env, imageId) {
 
-  const syncVolumeName = `garden-sync-volume-${serviceSpec.name}`
-  const targets = moduleSpec.hotReload.sync.map(pair => pair.target)
+  const syncVolumeName = `garden-sync`
 
-  const copyCommand = `cp -r ${targets.join(" ")} /.garden/hot_reload`
+  // We're copying the target folder, not just its contents
+  const syncConfig = moduleSpec.hotReload.sync
+  const targets = syncConfig.map(pair => removeTrailingSlashes(pair.target))
+
+  const copyCommands = makeCopyCommands(syncConfig)
 
   const initContainer = {
     name: "garden-sync-init",
     image: imageId,
-    command: ["sh", "-c", copyCommand],
+    command: ["sh", "-c", copyCommands],
     env,
     imagePullPolicy: "IfNotPresent",
     volumeMounts: [{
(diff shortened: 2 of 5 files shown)


garden-service/src/plugins/kubernetes/deployment.ts
```diff
@@ -467,16 +469,51 @@ function configureHotReload(deployment, container, serviceSpec, moduleSpec, env,
...
+ * the rsync container to update.
+ *
+ * This needs to deal with nested pathing as well as root.
+ * this will first create the path and then copy the contets from the
+ * docker image into the shared volume as a base for the rsync command
+ * to update.
+ *
+ * @param syncConfig
+ */
+export function makeCopyCommands(syncConfig: Array<any>) {
```
drubin: I am not really sure what types to pass in here?

garden-service/src/plugins/kubernetes/deployment.ts
```diff
@@ -467,16 +469,51 @@ function configureHotReload(deployment, container, serviceSpec, moduleSpec, env,
...
+ * Ensure that there's no trailing slash
+ *
+ * converts /src/foo/ into /src/foo
+ * @param path
+ */
+export function removeTrailingSlashes(path: string) {
+  return path.replace(/\/*$/, "")
+}
+
+export function rsyncSourcePath(modulePath: string, sourcePath: string) {
```
drubin: I also replaced this module with more generic path variable so we can test it in isolation. This function doesn't need access to any thing but a path.

docs/reference/config-files-reference.md
```diff
@@ -663,12 +663,16 @@ module:
...
+        # Example: "src"
+        #
         # Optional.
         source: .
 
         # POSIX-style absolute path to sync the directory to inside the container. The root path
         # (i.e. "/") is not allowed.
         #
+        # Example: "/app/src"
+        #
```
drubin: I think in future we need to ensure there are working examples like this it would solve many issues regarding to how to format paths/strings.

thsig:
One general thing: Maybe we should include more of the PR description in the commit message (here, and in general)? Just to make the commit history more descriptive without having to refer back to finished PRs to refer to something in the description (this applies to a lot of the team's older PRs).

E.g. add this after the "title line"?
> Fixes issues with handling subpaths as hot-reload sync targets with differences between the way cp and rsync and k8s handle the paths.
>
> Solves issues with trying to sync src into /app/src and other edge cases where hot-reloading would only work after a first change was made to the source.
>
> Also, originally both the source and target needed to be in the exact correct format for things to work regarding trailing slashes.

docs/reference/config-files-reference.md
thsig: Yeah, good point.

garden-service/src/plugins/kubernetes/deployment.ts
thsig: `syncConfig: SyncSpec[]` makes most sense here.

garden-service/src/plugins/kubernetes/deployment.ts
```diff
@@ -467,16 +469,51 @@ function configureHotReload(deployment, container, serviceSpec, moduleSpec, env,
...
-export function rsyncSourcePath(module: Module, sourcePath: string) {
-  return resolve(module.path, sourcePath)
+/**
+ * Creates the initial copy command for the sync init container.
+ *
+ * This handles copying the contents from source into a volume for
+ * the rsync container to update.
+ *
+ * This needs to deal with nested pathing as well as root.
+ * this will first create the path and then copy the contets from the
```
thsig: • "this will" -> "This will"
• "... path and then..." -> "... path, and then ..."
• "contets" -> "contents"

garden-service/src/plugins/kubernetes/deployment.ts
thsig: `SyncSpec[]` is the way to go here.

garden-service/src/plugins/kubernetes/deployment.ts
```diff
@@ -399,15 +398,18 @@ function configureVolumes(deployment, container, spec): void {
 */
 function configureHotReload(deployment, container, serviceSpec, moduleSpec, env, imageId) {
 
-  const syncVolumeName = `garden-sync-volume-${serviceSpec.name}`
-  const targets = moduleSpec.hotReload.sync.map(pair => pair.target)
+  const syncVolumeName = `garden-sync`
```
thsig: Is it OK that `syncVolumeName` is no longer unique on the level of service names? E.g. when two services in two different modules are deployed with hot reloading enabled?

garden-service/test/src/plugins/kubernetes/deployment.ts
```diff
@@ -0,0 +1,82 @@
...
+      it(`handles paths correctly for ${path[0]}`, () => {
+        expect(rsyncTargetPath(path[0])).to.eql(path[1])
+      })
+    }
+  })
+
+  describe("makeCopyCommands", () => {
+    // Test cases for each type
+    const configs: any[] = [
+      // Source is Missing slash
```
thsig: "Missing" -> "missing"

garden-service/src/plugins/kubernetes/deployment.ts
drubin: `syncVolumeName` name is unique per deployment, and we create a seperate deployment for every service/module right? 

If this assumption holds true then its fine. 

You might be confusing volumes or PV's which are unique per namespace and different deployments can reverence this.

But in this case` emptydir` its unique per deployment. similar to how container name is unique per pod.

PR #373, opened 2018-11-15
fix: allow env variables to be lower case

### LINKED ISSUE #369: Lower case env variables are not supported
`Error validating module website: key "a_a_a" is not allowed at path .services[0][env][a_a_a]`

```
module:
  description: Minimal Gatsby example
  name: website
  type: container
  hotReload:
    sync:
      - target: /app/src/
        source: src
  services:
    - name: website
      command: [ls, /app/src]
      hotReloadCommand: [npm, run, develop]
      env:
        GATSBY_WEBPACK_PUBLICPATH: /
        a_a_a: bbb
        #GATSBY_WEBPACK_PUBLICPATH: http://gats.local.app.garden/
      ports:
        - name: http
          containerPort: 8000
      ingresses:
        - path: /
          port: http
```

Changing the blow key to `A_A_A: bbb` works

drubin:
We should allow the same set of variables that bash/docker/k8s would allow. 

Fixes #369 

Also https://www.ietf.org/rfc/rfc2119.txt regarding usage of may/must :)

### DIFF
--- a/garden-service/src/config/common.ts
+++ b/garden-service/src/config/common.ts
@@ -31,7 +31,7 @@ export const joiPrimitive = () => Joi.alternatives().try(Joi.number(), Joi.strin
 export const absolutePathRegex = /^\/.*/ // Note: Only checks for the leading slash
 export const identifierRegex = /^[a-z][a-z0-9]*(-[a-z0-9]+)*$/
 export const userIdentifierRegex = /^(?!garden)[a-z][a-z0-9]*(-[a-z0-9]+)*$/
-export const envVarRegex = /^(?!GARDEN)[A-Z_][A-Z0-9_]*$/
+export const envVarRegex = /^(?!garden)[a-z_][a-z0-9_]*$/i
 
 export const joiIdentifier = () => Joi.string()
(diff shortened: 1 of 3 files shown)


garden-service/src/config/common.ts
```diff
@@ -31,7 +31,7 @@ export const joiPrimitive = () => Joi.alternatives().try(Joi.number(), Joi.strin
 export const absolutePathRegex = /^\/.*/ // Note: Only checks for the leading slash
 export const identifierRegex = /^[a-z][a-z0-9]*(-[a-z0-9]+)*$/
 export const userIdentifierRegex = /^(?!garden)[a-z][a-z0-9]*(-[a-z0-9]+)*$/
-export const envVarRegex = /^(?!GARDEN)[A-Z_][A-Z0-9_]*$/
+export const envVarRegex = /^(?!GARDEN)[A-Za-z_][A-Za-z0-9_]*$/
```
edvald: I think maybe we should disallow `garden` in lowercase as well. Something like `/^(?!garden)[a-z_][a-z0-9_]*$/i` (note the `/i`).

PR #378, opened 2018-11-19
fix(create-command): add project key to generated config and fix tests

eysi09:
The `project` key was missing at the top of the project level `garden.yml` file generated by the create command. Furthermore, the tests were validating the generated config against the projectSchema instead of the entire configSchema.

### DIFF
--- a/garden-service/src/commands/create/config-templates.ts
+++ b/garden-service/src/commands/create/config-templates.ts
@@ -7,13 +7,12 @@
  */
 
 import { capitalize, camelCase, uniq } from "lodash"
-import * as Joi from "joi"
 
 import { DeepPartial } from "../../util/util"
 import { ContainerModuleSpec } from "../../plugins/container"
 import { GcfModuleSpec } from "../../plugins/google/google-cloud-functions"
 import { ProjectConfig } from "../../config/project"
-import { BaseModuleSpec, ModuleConfig, baseModuleSpecSchema } from "../../config/module"
+import { ModuleConfig } from "../../config/module"
 
 /**
  * Ideally there would be some mechanism to discover available module types,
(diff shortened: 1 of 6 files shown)


garden-service/src/commands/create/config-templates.ts
```diff
@@ -85,22 +73,34 @@ export function npmPackageTemplate(_moduleName: string): any {
...
 }
 
-export const moduleTemplate = (name: string, type: ModuleType): Partial<BaseModuleSpec> => ({
-  name,
-  type,
-  description: `${titleize(name)} ${noCase(type)}`,
-})
+export const moduleTemplate = (name: string, type: ModuleType): ModuleTemplate => {
+  const moduleTypeTemplate = {
+    container: containerTemplate,
```
drubin: @eysi09  minor nitpick but this hash looks really broken mixing `container` and `"npm-package"` maybe we can stick to using a single type of object notion?  either quoting all objects or not quoting them.

edvald: You have to quote keys with a dash in them, and our linter complains if we unnecessarily quote strings. I personally prefer this over having a mix of quoted or non-quoted strings, or quoting all keys.

drubin: I am aware that you have to quote keys with a dash. 

But if you have have lists with differing structures they require mental overhead to process the difference. This is the exact point of standards/linting. 

This although valid JS and passes a linter is harder to read.

edvald: Unless the linter can conditionally require quoted strings depending on if any of the keys on the object requires it (I'm not aware of that option), the only real alternative would be to always quote strings, and I'd rather not. This, as it is in this commit, is the only style the linter allows in the current configuration.

drubin: https://palantir.github.io/tslint/rules/object-literal-key-quotes/  `consistent-as-needed`

Is there a good reason for going against the defaults?

edvald: Ah I wasn't aware of that one. Wouldn't mind that. The default is "always" which I don't like, personally, but "consistent-as-needed" seems fine to me if you'd prefer to switch.

Issue #328, opened 2018-10-12
Support .gardenignore in subdirectories


edvald:
`.gardenignore` files should have broadly the same semantics as `.gitignore` files. Currently we only support one file in the project root directory. We should scan for `.gardenignore` files in subdirectories and include those when filtering files to include in builds etc.

edvald:
Actually, we may also need to do this for our internal handling of `.gitignore` files, might wanna look into that along the way.

drubin:
Does Garden already honour .dockerignore files?

edvald:
We use docker for building containers, so yes, by extension.

drubin:
@edvald I meant in deciding when to trigger a rebuild.

edvald:
I see - no we don't refer to the .dockerignore file for that. I'm not sure how we'd implement that at the framework level, since the framework doesn't know what a `.dockerignore` file is.

PR #385, opened 2018-11-22
(WIP) feat: support ignorefiles in subdirectories

thsig:
Fixes https://github.com/garden-io/garden/issues/328.

Previously, getIgnorer only considered .gitignore and .gardenignore
files located in a project's root directory.

We now scan for .gitignore and .gardenignore files in all subdirectories
of the project's root directory, and prefix rules with the relative path
(from project root) to the directory containing them.


garden-service/test/src/util/ignorer.ts
```diff
@@ -0,0 +1,61 @@
...
+    it("includes rules from ignorefiles in a given directory", () => {
+      expect(ignorer.ignores("file-from-root-gitignore")).to.eql(true)
+      expect(ignorer.ignores("file-from-root-gardenignore")).to.eql(true)
+    })
+
+    it("only applies rules from a subdir's .gitignore to files/folders within that subdir", () => {
+      expect(ignorer.ignores("file-from-b-subdir-gitignore")).to.eql(false)
+      expect(ignorer.ignores("b/file-from-b-subdir-gitignore")).to.eql(false)
+      expect(ignorer.ignores("b/b-subdir/file-from-b-subdir-gitignore")).to.eql(true)
+    })
```
edvald: I think I'd like to see additional tests that test for ignoring files in subdirectories (e.g. a line like `somedir/my-ignored-file.txt`) and maybe a couple of variants on wildcards, to be sure that behavior works as expected as well.

thsig: Yeah, good idea. Any other test cases you'd like to suggest?

CC @drubin @eysi09

thsig: Added a few more tests here: https://github.com/garden-io/garden/pull/385/commits/a7ac21af8d14b9ec14d8cfb2ada9abac266e8569

Do these cover what you had in mind? @edvald

garden-service/src/garden.ts
```diff
@@ -828,6 +832,30 @@ export class Garden {
...
+  private async scanIgnorefiles() {
+    const gitignorePaths: string[] = []
+    const gardenignorePaths: string[] = []
+
+    for await (const item of scanDirectory(this.projectRoot)) {
+      if (!item) {
+        continue
+      }
+
+      const basename = parse(item.path).base
```
drubin: `basename` would be cleaner here I I think

garden-service/test/src/util/ignorer.ts
drubin: Negation so `!` 
Verify it handles the difference between things starting with `/foo` and `foo` and `/foo/` and `foo/`
Not sure how the underlaying library does this but we are doing some serious path munging with it

garden-service/src/util/ignorer.ts
```diff
@@ -0,0 +1,41 @@
...
+  ig.add(rules)
+  ig.add([
+    "node_modules",
+    ".git",
+    GARDEN_DIR_NAME,
+  ])
+  return ig
+}
+
+export async function rulesForIgnorer(projectRoot: string, ignorefilePaths: string[]) {
```
drubin: This needs a comment why we are doing this instead of just feeding the file directly. 

The whole logic should have a description

I think we do something like this?

1. Parse files for `.gardenignore` `.gitignore` (this feels a little odd, docker/* doesn't assume things based on gitignore so not sure how expected this behaviour actually is)
1. DO magic to merge them
1. Use this to compare if files should be watched?

garden-service/src/util/ignorer.ts
```diff
@@ -0,0 +1,41 @@
...
+  const commentRegex = /^#/
+  return flatten(await Bluebird.map(ignorefilePaths, async (ignorefilePath) => {
+    const relativePath = relative(projectRoot, dirname(ignorefilePath))
+    const rulePrefix = relativePath ? `${relativePath}/` : ""
+    return (await readFile(ignorefilePath)).toString()
+      .split(endOfLineRegex)
```
drubin: Why are we not doing this inversely `ig.add(eadFile(ignorefilePath))` and loop over and modify them? This way we delegate parsing the file to the actual parser.

See https://www.npmjs.com/package/ignore#addignorefilepath

garden-service/src/util/ignorer.ts
thsig: Yeah, good point. I'll add a comment explaining the approach here.

What we're doing here is prefixing rules from ignorefiles in subdirectories with the relative path to the subdir containing the ignorefile.

So the rule `*.js` in `foo/.gitignore` would be changed to `foo/*.js` before being added to the ignorer.

If we simply read the files in as strings, the ignorer wouldn't be aware of this, and would presumably register all the rules at the "top level", which we don't want.

garden-service/src/util/ignorer.ts
thsig: See https://github.com/garden-io/garden/pull/385#discussion_r235885137.

thsig:
Keeping this PR as is for now.

This implementation might ultimately be more confusing than the current semantics (i.e. only considering ignorefiles in project root).

When the time is right, we should revisit this.

edvald:
Came across this, which might be helpful in some way (whenever we get back to this, no big rush imo): https://stackoverflow.com/questions/466764/git-command-to-show-which-specific-files-are-ignored-by-gitignore

edvald:
Superseded by #698

PR #391, opened 2018-11-28
feat: experimental HTTP API

edvald:
This is a very minimal first implementation, just to get the ball
rolling. This exposes most commands (excluding any interactive or
perpetually running commands) via an HTTP API, served by the
`garden serve` command.

BREAKING CHANGE:

The `exec`, `run module` and `run test` commands no longer default to
`--interactive=true`. In order to run those interactively, e.g. attach
to shells, you must now explicitly specify the `-i` flag.


drubin:
@edvald regarding the breaking change. 

I think test / run are fine to default to interactive false but I think the exec's use-cased feels more interactive/debugging.

I love that docker-compose's exec command defaults to interactive=true it was one of my favourite things about using it vs docker's exec or kubectl's exec not having to type `-it`

Is there a way to keep this behaviour on the CLI but make it not for the http api, I get this might actually be confusing fro developers on garden but I think it makes the interface much nicer for users using the CLI? Or does it simply introduce too much complexity?

garden-service/src/server.ts
```diff
@@ -0,0 +1,196 @@
...
+import dedent = require("dedent")
+import Joi = require("joi")
+import getPort = require("get-port")
+import { Command, Parameters } from "./commands/base"
+import { validate } from "./config/common"
+import { coreCommands } from "./commands/commands"
+import { mapValues, pick } from "lodash"
+import { Garden } from "./garden"
+
+export const DEFAULT_SERVER_PORT = 8000
```
drubin: We should pick a more unique port this one will almost always be taken I know we use `get-port` but this random factor makes it a little harder to make it predictable to build api's for it.

Maybe we shouldn't even try to magically pick a port number and if its in use require they explicitly pick another we can suggest one for the user though.

Also this default value should be in the docs.

edvald: Sure, fine to change the default, but the "magical" thing is kinda necessary when starting the server implicitly, e.g. in `garden dev`. We could make it fully explicit with the `garden serve` command though, that would make sense.

Suggestion on a good default port?

garden-service/src/commands/serve.ts
```diff
@@ -0,0 +1,51 @@
...
+    Starts an HTTP server that exposes Garden commands and events.
+  `
+
+  arguments = serveArgs
+  options = serveOpts
+
+  async action({ garden, opts }: CommandParams<Args, Opts>): Promise<CommandResult<{}>> {
+    await startServer(garden, opts.port)
+
+    while (true) {
```
drubin: Can we please add a comment here. I assume its because our actions can't "end" but this feels hacky and deserves a comment.

garden-service/src/server.ts
drubin: I think this magical port thing will make it less useful to users who want to bookmark or other wise have predictable urls. 

Why don't we just allow `--port` to `dev` I am a strong believer in the 2 commands behaving the same predictable way.

This random part goes against predictable codified predictable workflows, maybe we should rather read it from the `garden.yaml` file? (then its codified in a project) and we don't need to have params passed in. (obviously with sane defaults)


Thumb suck

garden-service/src/commands/serve.ts
edvald: Yep, done.

garden-service/src/server.ts
edvald: I've modified the default, and made it so that `garden serve` always uses a fixed port, and the "implicit" servers created in watch commands always have random ports. 

I think that's probably clearer, and imo good enough for the context of this PR, which is more geared towards experimentation. 

I didn't want to add explicit parameters to the `garden dev` command or to the `garden.yml` config just yet because we need to work out more of the design before deciding what the top-level interface looks like.

edvald:
Re: the `exec` command being interactive by default or not... I kinda debated that and wound up deciding that it's best to go for consistency, not least because people already expect it in common tools like kubectl. Having separate parameter defaults for the CLI and API felt rather messy as well... I don't feel super strongly about it, but this made sense to me. A 3rd option would be to simply remove `exec` from the API, but that felt odd to me as well. 

@thsig @eysi09 any thoughts on this?

eysi09:
I'm tempted to say that we should default to `--interactive=true` even though it's not consistent with docker and kubectl. Feels like that's the intention when you run `garden exec` and thus fits into the dev-friendly paradigm. On the other hand we just had a lengthy discussion about not introducing code complexity. But if it's not too much hassle, I vote for `-i=true` along with a comment about why the http api behaves differently.

edvald:
Yep, I concede that. I've updated the PR accordingly, and without the breaking change.

garden-service/src/server.ts
```diff
@@ -0,0 +1,234 @@
...
+  if (!port) {
+    port = await getPort()
+  }
+
+  // TODO: secure the server
+  app.listen(port)
+
+  const url = `http://localhost:${port}`
+
+  garden.log.info({
```
drubin: The order of loading this means the dashboard link is only visible after every thing has been deployed. 

I assume we we later move things around we can ensure this happens either in parallel or before?

garden-service/src/server.ts
```diff
@@ -0,0 +1,234 @@
...
+    emoji: "sunflower",
+    msg: chalk.cyan("Garden dashboard and API server running on ") + url,
```
drubin: Stupid thing but can we please change this to 

`${chalk.cyan("Garden dashboard and API server running on")} ${ur}l`

Notice the space is outside of the colour function. this allows the link to not look broken. 

screenshot 2018-11-29 11 46 17

edvald: 👍 will fix

garden-service/src/server.ts
edvald: Yea we could move it now actually

edvald:
@drubin updated to start server earlier (to print the URL) and fixed the link formatting issue.

drubin:
@edvald I guess I more meant for it to appear right at the top in a stable predictable place kinda like the directly after `Good afternoon! Let's get your environment wired up...` 

but currently it appears directly below `local-kubernetes          → Configured` but because that has sub logs it kinda bounces around which actually feels worse...

I think though unless it's an easy fix we should leave it till later.

garden-service/src/server.ts
drubin: This still seems to cause an issue it might be the extra space before the end of the `cyan`

drubin: Lets chalk it up to an iterm2 bug... I tried moving the spacing around and it didn't help.

Issue #389, opened 2018-11-26
Editing `garden.yml` file causes `garden dev` to stop


drubin:
Editing any `garden.yml` file causes the process to stop working either crashing or simply exiting.

If we can't live reload/compute dependencies I think its fine to exit but we should at least print a friendly error message and tell them what command to re-run (i.e what params it was called with)

## Steps to reproduce 

1. Start any project with `garden dev`
1. Make a change to any `garden.yml` file
1. `garden dev` stops

PR #392, opened 2018-11-28
fix: fix dev command terminating on config change

thsig:
Resolves https://github.com/garden-io/garden/issues/389.

Also added a `headerRendered` instance var to commands accepting the watch flag, to prevent commands re-rendering their header on restart.

I decided not to add `headerRendered` to the base `Command` class for starters. Thoughts?

### DIFF
--- a/garden-service/src/cli/cli.ts
+++ b/garden-service/src/cli/cli.ts
@@ -7,36 +7,26 @@
  */
 
 import * as sywac from "sywac"
-import { merge, intersection, range } from "lodash"
+import { intersection, merge, range } from "lodash"
 import { resolve } from "path"
 import { safeDump } from "js-yaml"
 import { coreCommands } from "../commands/commands"
-import stringify = require("json-stringify-safe")
 
 import { DeepPrimitiveMap } from "../config/common"
-import {
-  getEnumKeys,
-  shutdown,
-  sleep,
-} from "../util/util"
+import { getEnumKeys, shutdown, sleep } from "../util/util"
 import {
   BooleanParameter,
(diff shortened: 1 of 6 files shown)


edvald:
How about adding a `printHeader()` method to the Command interface, and have the CLI call that?

eysi09:
What was the underlying issue?

thsig:
> How about adding a `printHeader()` method to the Command interface, and have the CLI call that?

Yeah, that would be better. This was my initial thought, but I thought I'd start simple and ask for feedback in case this had any bearing on the upcoming changes to our logging system.

thsig:
> What was the underlying issue?

The `dev` command was returning `{}` (from its `action` instance method) instead of calling `handleTaskResults` like e.g. the `deploy` command.

To decide whether to restart the command or just complete it, the CLI checks whether the return value of the command's action has `restartRequired = true`. Since the `dev` command's return value never had `restartRequired = true`, the CLI assumed the command was done.

https://github.com/garden-io/garden/pull/392/files#diff-e117e7480ef1501a63555e1ffc7e2ae1L134

thsig:
> How about adding a `printHeader()` method to the Command interface, and have the CLI call that?

Just updated the PR with changes along these lines. Is this satisfactory? @edvald

https://github.com/garden-io/garden/compare/18bbec321922a7fea147edd9c566f4a393e14b3a..0274c095adff0ab487ae6487b55be45ea2122ccb

thsig:
@edvald: I've moved the `printHeader` call to the CLI as we discussed.

Issue #386, opened 2018-11-23
Dependency relation not transitive?


eysi09:
If `service-a` depends on `service-b`, and `service-b` depends on` service-c`, `service-a` does not appear to depend on `service-c`.

I noticed this when trying out the _[sock-shop-mono](https://github.com/garden-io/sock-shop-mono/commits/master)_ example where `front-end` depends on `orders` and `orders` depends on `shipping`, at the service level.

Now, when I run `garden dev` and then change the `shipping` service, `front-end` will deploy before `shipping`. Whereas I'd expect `shipping` to deploy, then `orders`, then `front-end`. (Note that on the current _sock-shop-mono_ `master` branch, `shipping` is a dependency of `front-end` so that should be removed to see issue.)

If the dependency relation is not supposed to be transitive we should probably document that because I think that's what people would assume.

[Here's a digram](https://github.com/microservices-demo/microservices-demo.github.io/blob/HEAD/assets/Architecture.png) of the sock-shop service dependencies.

edvald:
This is simply a bug, likely introduced recently. @thsig can you investigate?

edvald:
Or well at least it sounds like a bug from the description, but looking at it again there are some possible explanations, depending on the exact configuration and state (not 100% clear to me reading the description). In any case this needs to be inspected - and I'd suggest tests to be written to validate.

thsig:
I'll look into this (and write tests as appropriate).

PR #397, opened 2018-11-30
fix: fixed dependency bug & simplified TaskGraph

thsig:
Fixes https://github.com/garden-io/garden/issues/386.

This bug was fixed here as a side-effect of simplifying the implementation of `TaskGraph`'s dependency logic.

Previously, dependency relationships of nodes in `TaskGraph` were updated by adding/removing from the dependencies/dependants arrays of the appropriate `TaskNode`s when nodes were added or removed.

This logic was somewhat opaque and difficult to maintain/debug. Here, we change the implementation to simply rebuild the dependency relationships of all nodes in `TaskGraph`'s index whenever a node is added or removed. We make use of a mutex/lock to prevent race conditions, and cache config-level dependency relationships for performance.

With this implementation, we no longer need to track dependants or handle predecessors (since `addTaskInternal` and `processTasksInternal` are never mutually concurrent), and the logic is hopefully simpler and easier to maintain.

### DIFF
--- a/garden-service/src/task-graph.ts
+++ b/garden-service/src/task-graph.ts
@@ -79,6 +79,10 @@ export class TaskGraph {
 
     // this.taskDependencyCache will already have been populated at this point (happens in addTaskInternal).
     for (const node of taskNodes) {
+      /**
+       * We set the list of dependency nodes to the intersection of the set of nodes in this.index with
+       * the node's task's dependencies (from configuration).
+       */
       node.clear()
       const taskDeps = this.taskDependencyCache[node.getKey()] || new Set()
       node.setDependencies(taskNodes.filter(n => taskDeps.has(n.getBaseKey())))
--- a/garden-service/src/task-graph.ts
+++ b/garden-service/src/task-graph.ts
@@ -9,7 +9,8 @@
 import * as Bluebird from "bluebird"
 import * as PQueue from "p-queue"
 import chalk from "chalk"
-import { merge, padEnd, pick } from "lodash"
+import * as yaml from "js-yaml"
+import { merge, padEnd, pick, flatten } from "lodash"
 import { BaseTask, TaskDefinitionError } from "./tasks/base"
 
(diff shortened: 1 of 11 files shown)


garden-service/src/task-graph.ts
```diff
@@ -62,63 +73,83 @@ export class TaskGraph {
...
+      node.clear()
+      const taskDeps = this.taskDependencyCache[node.getKey()] || new Set()
+      node.setDependencies(taskNodes.filter(n => taskDeps.has(n.getBaseKey())))
     }
 
-    this.index.addNode(node)
-    await this.addDependencies(node)
+    const newRootNodes = taskNodes.filter(n => n.getDependencies().length === 0)
+    this.roots.clear()
+    this.roots.setNodes(newRootNodes)
```
edvald: How about inlining this logic in the process method, instead of applying it when adding tasks? That, combined with making `addNodeWithDependencies()` synchronous (by moving the async call to getDependencies outside of the method), should get rid of the need for the lock. Also, we could simply remove the separate `roots` index, and just use the computed `newRootNodes` directly in the process method.

garden-service/src/task-graph.ts
```diff
@@ -62,63 +73,83 @@ export class TaskGraph {
...
+   * Should only be called inside a this.lock block.
+   */
+  private async rebuild() {
+    const taskNodes = this.index.getNodes()
+
+    // this.taskDependencyCache will already have been populated at this point (happens in addTaskInternal).
+    for (const node of taskNodes) {
+      node.clear()
+      const taskDeps = this.taskDependencyCache[node.getKey()] || new Set()
+      node.setDependencies(taskNodes.filter(n => taskDeps.has(n.getBaseKey())))
```
edvald: This needs some explanation.

thsig: Added an explanatory comment here just now.

Issue #388, opened 2018-11-26
Garden dev behaviour differs from deploy


drubin:
The calculation of tasks differs between `dev` and `deploy` in unexpected ways. This was partially fixed here https://github.com/garden-io/garden/pull/370 

My understanding is that `garden deploy -w` behaves the same as `garden dev` but with out tests? 

---

This workflow is using the `gatsby-hot-reload` example project

Each one of these hot reloading + rebuilding were done because I ran 

`garden get status` (See #387)

```sh
garden dev --hot-reload=website

Good morning! Let's get your environment wired up...

✔ website                   → Building website:2c12317af0... → Done (took 36.2 sec)
✔ website                   → Deploying → Ready
✔ website                   → Hot reloading → Done (took 316 ms)
✔ website                   → Building website:2c12317af0... → Done (took 0.7 sec)
✔ website                   → Hot reloading → Done (took 144 ms)
✔ website                   → Building website:2c12317af0... → Done (took 0.4 sec)
✔ website                   → Hot reloading → Done (took 132 ms)
✔ website                   → Building website:2c12317af0... → Done (took 0.4 sec)
```

Notice the difference when running 

`garden get status` with `deploy` instead of `dev`

```sh
garden deploy --hot-reload=website
Deploy 🚀

✔ website                   → Building website:2c12317af0... → Done (took 0.5 sec)
✔ website                   → Checking status → Version 2c12317af0 already deployed
✔ website                   → Hot reloading → Done (took 103 ms)
✔ website                   → Hot reloading → Done (took 118 ms)
✔ website                   → Hot reloading → Done (took 103 ms)
```

---

But I would expect the differences between dev and deploy + hot-reload/watch to be almost the same.

Why do we have totally different workflows/code instead of calling the same calculation logic with different params?

We should 

1. Fix the issue where their task list is calculated differently
1. Fix the documentation to highlight the differences between the `dev` and `deploy`
1. Fix issue #387

thsig:
> 2. Fix the documentation to highlight the differences between the dev and deploy

Regarding 2. do you have any suggestions for how to change the description strings for the `dev` and `deploy` commands? At first glance, they seem pretty clear to me, but I'm perhaps a bit too familiar with how all this works. @drubin

edvald:
@thsig Is this still an issue?

drubin:
[...]

PR #398, opened 2018-11-30
fix: make dev cmd consistent with deploy cmd

thsig:
Fixes https://github.com/garden-io/garden/issues/388.

Don't return a singleton build task by default in the dev command. This makes the dev command's behavior more consistent with the deploy command in watch mode.


drubin:
Thanks for this I think this might solve some of the issues. 

But it appears the structure/code is still slightly different between dev/deploy. My understanding is that the code should be fairly similar if not exactly the same and shareable but it's not. 

I don't fully understand why? Why can't abstract it away so that when we fix/change something we change it in one place?

We have had a few bugs recently about the differences between these 2 commands behaving differently and every time we patch the logic.

Is there a solution of wrapping this login in a shared function that could potentially solve this once and for all?

thsig:
@drubin: True, the structure of `dev` vs `deploy` is still different in a couple of ways, mostly in that `dev` uses both the `getTestTasks` and `getTasksForModule` helpers to generate the task list in its `handler` and `changeHandler`, where `deploy` uses only `getTasksForModule`.

On the other hand, the `test` command only uses the `getTestTasks` handler, and the `build` command simply returns build tasks and doesn't use either of these helpers.

We could consolidate `getTestTasks` into `getTasksForModule` and add an `includeTests` param, but I thought it was clearer to have the commands mix and match them explicitly.

That said, `getTasksForModule` is a bad name, just couldn't think of anything better to describe its return value.

PR #403, opened 2018-11-30
docs: add new questions to FAQ

ellenkorbes:


docs/faqs.md
```diff
@@ -1,6 +1,128 @@
...
+
+If I am an architect trying to keep up with development, can I then use garden to get an overview of what is going on? 
+
+### Which languages does Garden work with? C#, Python, etc? 
+
+**Garden works with any language you can run in a Docker container. That includes C# and Python.**
+
+However, some *stacks* are better suited than others for Garden. Interpreted languages like Python and NodeJS tend to work very well.
+
+Some languages, on the other hand, might be less suited for Garden. This could be because of long compile times, or because they're resource heavy and running many of them on a local machine will slow things down (e.g. Java).
```
edvald: "...might _currently_ be less suited for Garden."

docs/faqs.md
```diff
@@ -1,6 +1,128 @@
...
+
+This will help users more easily migrate across platforms when their needs (almost inevitably) change or new technology emerges.
+
+If I am an architect trying to keep up with development, can I then use garden to get an overview of what is going on? 
+
+### Which languages does Garden work with? C#, Python, etc? 
+
+**Garden works with any language you can run in a Docker container. That includes C# and Python.**
+
+However, some *stacks* are better suited than others for Garden. Interpreted languages like Python and NodeJS tend to work very well.
```
edvald: I'd also say Go works pretty well, because it's not so resource intensive (unlike Java) and because it compiles quickly.

docs/faqs.md
```diff
@@ -1,6 +1,128 @@
...
+
+Some languages, on the other hand, might be less suited for Garden. This could be because of long compile times, or because they're resource heavy and running many of them on a local machine will slow things down (e.g. Java). 
+
+That's why we are working on supporting remote development, in which case the services run in a remote cluster and not on the user's machine.
+
+Furthermore, for some stacks it might take a long time to re-build and re-deploy services. If those services are frequently changed during development, running them with Garden might be undesirable.  Garden can however work well if the services being worked on can re-build and re-deploy in a reasonable time and/or support hot reloading, and they merely depend on the slower services.
```
edvald: I think there's more nuance to this. When working on heavy/slow modules like that, you may not want to use `garden dev` but rather limit your scope to the things you're working on directly, or manually run deploys/builds. It should still be useful if you need to track dependencies across things.

docs/faqs.md
```diff
@@ -1,6 +1,128 @@
...
+
+**Garden operates in the pre-commit stage of the development workflow, and focuses on the as-you-code experience** e.g. by providing short feedback loops and running tests *as you develop*.
+
+It currently only runs locally—although we're working on remote capabilities—and when it comes to tests, it's better suited for things like smoke testing.
+
+CI, on the other hand, runs remotely, after you commit your code, and it's better suited for running unit and integration tests.
+
+### Is it possible to tag end-to-end tests so we don't have to re-run all tests every time?
+
+You can split them up and run tests by their names with the garden test command.The YAML could look like this:
```
edvald: Missing space

docs/faqs.md
```diff
@@ -1,6 +1,128 @@
...
+tests:
+  - name: light
+    command: [npm, run, e2e-light]
+  - name: full
+    command: [npm, run, e2e-full]
+```
+
+And you'd run them with `garden test front-end --name=light`, assuming the service is called front-end.
+
+In `garden dev` mode, all tests for a given service are run when a file watch event is triggered. To disable these you'd have to comment them out from the `garden.yml` file.
```
edvald: We should make an issue for that, and link it here.

docs/faqs.md
```diff
@@ -1,6 +1,128 @@
...
+
+More fine grained control over when and how tests run could be a feature for the future.
+
+### How about Docker Swarm?
+
+**Garden doesn't have a Docker Swarm plugin** and can't deploy to a Swarm cluster. Rather, at the moment, it deploys services to a Kubernetes cluster.
```
edvald: We do actually have limited support for it... Maybe we could say that we have a rough version of it, and that if people are interested we could consider stabilizing it?

docs/faqs.md
drubin: We had this debate before the compilation time inside docker slows things down because out side of docker go can leverage `incremental builds.` but inside every thing starts with a clean slate. 

Its not to say that it doesn't work its just noticeably slower than building these things outside of container. 

We can and should solve this with nice and easy to follow recipes but the standard flows as they are right now are slower. (this has nothing to do with Garden and more to do with Docker/layering and multi-stage builds or building outside on the host machine)

I think we should split the answer though.

1. All languages work do work.
1. A+ Grade
1. B Grade 

Or something similar to show the experience is less amazing at the moment.

docs/faqs.md
ellenkorbes: What does
> limit your scope to the things you're working on directly
mean?

docs/faqs.md
edvald: I really don't think we want to suggest commenting stuff out of configs. We should just be clear about the different commands and when you'd want to use each.

docs/faqs.md
```diff
@@ -1,6 +1,130 @@
...
+tests:
+  - name: light
+    command: [npm, run, e2e-light]
+  - name: full
+    command: [npm, run, e2e-full]
+```
+
+And you'd run them with `garden test front-end --name=light`, assuming the service is called front-end.
+
+In `garden dev` mode, all tests for a given service are run when a file watch event is triggered. If the system you're working on is very large, you might want to use more specific commands.
```
edvald: Not sure "`garden dev` mode" is a good way to phrase it?

docs/faqs.md
```diff
@@ -1,6 +1,130 @@
...
+### Is it possible to tag end-to-end tests so we don't have to re-run all tests every time?
+
+You can split them up and run tests by their names with the `garden test` command. The YAML could look like this:
+
+```yaml
+tests:
+  - name: light
+    command: [npm, run, e2e-light]
+  - name: full
+    command: [npm, run, e2e-full]
```
edvald: Perhaps a better example would be splitting up tests by _what_ they're testing, as opposed to how "thoroughly". You could have separate test-suites with different dependencies, to avoid testing everything when one service changes. That's the benefit of the dependency graph - as your stack gets more complex, you can make your graph more granular. Similar to how you'd split a library out of a large codebase to reduce its weight.

This would probably warrant a blog post on its own, once we have some good real-world examples to reference.

edvald:
Ah, oops, thought the last two comments were also outdated. They're pretty minor though.

docs/faqs.md
ellenkorbes: I honestly don't know if I should change the wording here.

PR #404, opened 2018-11-30
Event bus and websocket endpoint

edvald:
Added an event bus on `garden.events`, and just to kick things off I'm emitting some events from the TaskGraph.

The WebSocket endpoint forwards any event emitted to `garden.events` to open
connections, and allows commands to be executed via the socket as well.


eysi09:
Could you add idiot proof instructions for how to interact with the API endpoints? E.g. the shape of the JSON payload and the `command.subCommand` syntax.

garden-service/src/util/util.ts
```diff
@@ -249,6 +250,38 @@ export function splitFirst(s: string, delimiter: string) {
   return [parts[0], parts.slice(1).join(delimiter)]
 }
 
+/**
+ * Recursively process all values in the given input,
+ * walking through all object keys _and array items_.
+ */
+export function deepMap<T extends object, U extends object = T>(
```
eysi09: This doesn't appear to be used anywhere.

garden-service/src/commands/build.ts
```diff
@@ -22,7 +22,7 @@ import { Module } from "../types/module"
 import { logHeader } from "../logger/util"
 
 const buildArguments = {
-  module: new StringsParameter({
+  modules: new StringsParameter({
```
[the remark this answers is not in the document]
edvald: Added a commit for that.

garden-service/src/commands/call.ts
```diff
@@ -24,7 +24,7 @@ import dedent = require("dedent")
 
 const callArgs = {
   serviceAndPath: new StringParameter({
-    help: "The name(s) of the service(s) to call followed by the ingress path (e.g. my-container/somepath).",
+    help: "The name(s) of the service to call followed by the ingress path (e.g. my-container/somepath).",
```
eysi09: Why remove the plural `(s)` from `service(s)`?

edvald: We can't call more than one service here.

eysi09: Ah yes. So `name` should also be singular than.

eysi09: ...I'll go ahead and fix it and merge.

PR #405, opened 2018-11-30
refactor: rename `generic` plugin to `exec`

edvald:
This naming is much more clear, since basically the plugin executes
commands in a shell, and the name `generic` doesn't give any indication
as to how it works.

BREAKING CHANGE:
Projects using the `generic` module type need to update the relevant
`garden.yml` files, to reference the `exec` module type instead.


edvald:
One thing worth mentioning, is that we might later have a separate `local` plugin (name TBD), because we may want to have a clear indication of different execution environments when using a remote cluster... Not sure how it'll look but something to consider along the way.

drubin:
@edvald I was thinking about what you said, I don't think `shell` shows what we actually want this to do really shells feel like they are interactive.

Regarding the remote/local. I think since this command currently runs locally on the users machine I think we should explicitly call it `LocalExec` and then when we do remote containers/services `RemoteExec`? 

My worry is that if we start calling this something generic `shell` it will be have a different behaviour when when split things.

We can also do `Exec` with a param for remote/local where we default to local (for now). I think this maps better to the fact that the `Exec` module has a `command` in it. 

`shell` just doesn't really sit well as a name choice but not sure `LocalExec` is right either.

edvald:
Exec makes sense to me, let's go for that!

edvald:
Let's hold off on merging this until we're working on 0.9, or merge this to a new 0.9 branch.

PR #408, opened 2018-12-03
feat: added get graph command

thsig:
This command returns a serialized representation of the project's
DependencyGraph.

### DIFF
--- a/garden-service/test/src/commands/get/get-graph.ts
+++ b/garden-service/test/src/commands/get/get-graph.ts
@@ -1,0 +1,25 @@
+import { expect } from "chai"
+import { dataDir, makeTestGarden } from "../../../helpers"
+import { GetGraphCommand } from "../../../../src/commands/get/get-graph"
+import { resolve } from "path"
+
+describe("GetGraphCommand", () => {
+  const pluginName = "test-plugin"
+  const provider = pluginName
+
+  // TODO: Switch to a stable topological sorting algorithm that's more amenable to testing.
+  it("should get the project's serialized dependency graph", async () => {
+    const garden = await makeTestGarden(resolve(dataDir, "test-project-dependants"))
+    const log = garden.log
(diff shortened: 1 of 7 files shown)


garden-service/test/src/commands/get/get-graph.ts
```diff
@@ -0,0 +1,23 @@
...
+    const command = new GetGraphCommand()
+
+    const res = await command.action({
+      garden,
+      log,
+      args: { provider },
+      opts: {},
+    })
+
+    expect(Object.keys(res.result!).sort()).to.eql(["nodes", "relationships"])
```
drubin: Why is there a sort here? the order should be predictable right?

thsig: Maybe the sort is unnecessary, but I find the key ordering guarantee a bit vague:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys#Examples

So I've always assumed I had to ensure the key order myself when using `Object.keys`.

drubin: My point is if the function should return a predictable list, then the function should do the sorting not the tests... it hides hidden logic around ordering. (in this test it doesn't matter)

I can see that you are actually trying to get the keys out of the "object"  but I guess we should be explicitly checking that the 2 objects are equal?

This test doesn't feel like it tests a valid expected result? 

There is a bunch of logic in the `dependency-graph.ts` that does filtering / sorting but never actually tests tested?

thsig: Without a stable topological sorting algorithm (didn't find a JS lib that implemented one of those), testing the ordering means essentially re-implementing the logic in the test, so I'm skipping that for now (added a TODO comment about this).

I think we should just merge this one for now.

PR #412, opened 2018-12-03
docs: new readme

ellenkorbes:


README.md
```diff
@@ -1,63 +1,73 @@
...
-
 ![](docs/garden-banner-logotype-left-2.png)
 
-*Welcome! Garden is a full-featured development framework for containers and serverless backends, designed to make
-it easy to develop and test distributed systems.*
-<br><br>
+Garden is an open-source development engine for Kubernetes, containers and serverless backends, designed to make it easy to test and develop multi-service systems.
 
-### Status
+- **Focus on what matters:** Garden continuously builds, tests and deploys your changes into your own persistent development environment as you code. Easily define and continuously run tests in the background and focus on writing code. Develop with confidence, no matter how complex your stack gets.
```
edvald: I feel like the last sentence in this paragraph should support the heading. Maybe just omit it?

README.md
```diff
@@ -1,63 +1,73 @@
...
+- [Reference](https://docs.garden.io/reference), for the glossary, commands reference, configuration files reference, and template strings reference.
+- [FAQs](https://docs.garden.io/faqs).
 
-Please read the [Motivation](https://docs.garden.io/home#motivation) section in our documentation
-for a brief discussion on why we're building Garden.
+## Features
 
+- Garden keeps track of all interdependencies between your services.
+- It automatically re-builds, re-deploys, and re-tests as you code.
+- A dependency graph makes sure only what's needed gets worked on, resulting in really fast feedback loops.
```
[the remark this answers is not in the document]
ellenkorbes: My point here is that feedback loops are quick *because* of the dependency graph. Not trying to re-make that same point about the dependencies, but emphasize a consequence of it. I'll rewrite it.

As for "get worked on," I'm trying to think of a term I can use to replace "re-build, re-deploy, and re-test" because repeating those all the time gets old real quick. I'll come up with something else.

README.md
```diff
@@ -1,63 +1,73 @@
...
+- Hot reload makes feedback nearly instantaneous in compatible services.
```
edvald: This feels like it needs more clarity. Maybe something to the tune of "Use _hot reloading_ to synchronise code into deployed containers while you work, to rapidly update static files and services that support in-place reloading."

README.md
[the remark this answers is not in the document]
edvald: Ok got it. That does make sense as a separate point, just lacks a bit more clarity then.

As a general point, I think it's fine for these points to be slightly more verbose when needed. I'd rather prioritise clarity over brevity in this context.

README.md
```diff
@@ -1,63 +1,75 @@
...
 
-![](docs/garden-banner-logotype-left-2.png)
 
-*Welcome! Garden is a full-featured development framework for containers and serverless backends, designed to make
-it easy to develop and test distributed systems.*
-<br><br>
+Garden is an open-source development engine for Kubernetes, containers and serverless backends, designed to make it easy to test and develop multi-service systems.
 
-### Status
+- **Focus on what matters:** Garden continuously builds, tests and deploys your changes into your own persistent development environment as you code. Easily define and continuously run tests in the background and focus on writing code.
```
edvald: Maybe add an Oxford comma before that last part? Makes it easier to parse.

README.md
```diff
@@ -1,63 +1,75 @@
...
+- [FAQs](https://docs.garden.io/faqs).
 
-Please read the [Motivation](https://docs.garden.io/home#motivation) section in our documentation
-for a brief discussion on why we're building Garden.
+## Features
 
+- Garden keeps track of all interdependencies between your services.
+- It can automatically re-build, re-deploy, and re-test your services as you code.
+- Due to Garden's dependency graph, you get really fast feedback loops: It makes sure only what's needed gets re-built, re-deployed, and/or re-tested.
+- Hot reload lets you near-instantaneously update code and static files on containers as they run, on services that support in-place reloading.
```
edvald: _in_ containers/services.

PR #413, opened 2018-12-03
Fix log tailing issues

### LINKED ISSUE #411: Garden logs -t doesn't handle service restarts
### Current Behavior

When running `garden logs -t` and `garden dev` and you make a change to a service garden doesn't detect this and stops being able to follow the logs

There is also a Deprecation warning from moment?

```
Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
Arguments:
[0] _isAMomentObject: true, _isUTC: false, _useUTC: false, _l: undefined, _i: rpc, _f: undefined, _strict: undefined, _locale: [object Object]
Error
    at Function.createFromInputFallback (/app/garden-service/node_modules/moment/moment.js:320:98)
    at configFromString (/app/garden-service/node_modules/moment/moment.js:2368:15)
    at configFromInput (/app/garden-service/node_modules/moment/moment.js:2594:13)
    at prepareConfig (/app/garden-service/node_modules/moment/moment.js:2577:13)
    at createFromConfig (/app/garden-service/node_modules/moment/moment.js:2544:44)
    at createLocalOrUTC (/app/garden-service/node_modules/moment/moment.js:2631:16)
    at createLocal (/app/garden-service/node_modules/moment/moment.js:2635:16)
    at hooks (/app/garden-service/node_modules/moment/moment.js:12:29)
    at Stream.proc.stdout.pipe.on (/app/garden-service/build/plugins/kubernetes/actions.js:295:29)
    at Stream.emit (events.js:182:13)
    at drain (/app/garden-service/node_modules/through/index.js:36:16)
    at Stream.stream.queue.stream.push (/app/garden-service/node_modules/through/index.js:45:5)
    at emit (/app/garden-service/node_modules/split/index.js:37:14)
    at Stream.<anonymous> (/app/garden-service/node_modules/split/index.js:60:7)
    at _end (/app/garden-service/node_modules/through/index.js:65:9)
    at Stream.stream.end (/app/garden-service/node_modules/through/index.js:74:5)
    at Socket.onend (_stream_readable.js:628:10)
    at Object.onceWrapper (events.js:273:13)
    at Socket.emit (events.js:187:15)
    at endReadableNT (_stream_readable.js:1094:12)
    at process._tickCallback (internal/process/next_tick.js:63:19)
result                    →                          → error: code = Unknown desc = Error: No such container: fb49980718a85976f30d95d21cc38b66efe344825a8cef48708a311a468967bc
vote                      →                          → error: code = Unknown desc = Error: No such container: c35073d998796767d1b351a1a0e965491e2a1e80239b3834ae0ae00319d8b6ea
db                        →                          → error: code = Unknown desc = Error: No such container: f561dc0e83bb93b8d281727e4ca40909f2c0295982b0095e63788098c769fb91
```

 ### Expected behavior

To be able to get the logs as things change (thats the point fo being able to view the logs with -t)

 ### Reproducible example

1. `garden deploy`
1. `garden logs -t` (keep runinng)
1. Edit a file to trigger a rebuild `echo " " > services/vote/templates/index.html`
1. `garden deploy`

 ### Suggested solution(s)

na.

 ### Additional context

Na 
 ### Your environment

Garden version: 0.8.0

edvald:
Closes #411 

Also adds support for log tailing for OpenFaaS functions.

### DIFF
--- a/garden-service/src/plugins/kubernetes/logs.ts
+++ b/garden-service/src/plugins/kubernetes/logs.ts
@@ -8,27 +8,35 @@
 
 import * as split from "split"
 import moment = require("moment")
-import Stream from "ts-stream"
 
-import { GetServiceLogsResult, ServiceLogEntry } from "../../types/plugin/outputs"
+import { GetServiceLogsResult } from "../../types/plugin/outputs"
 import { GetServiceLogsParams } from "../../types/plugin/params"
 import { ContainerModule } from "../container"
 import { getAppNamespace } from "./namespace"
 import { splitFirst } from "../../util/util"
 import { BinaryCmd } from "../../util/ext-tools"
 import { kubectl } from "./kubectl"
-import { ContainerService } from "../../../tmp/dist/build/plugins/container"
-import { LogEntry } from "../../logger/log-entry"
 
-export async function getServiceLogs(
-  { ctx, log, service, stream, tail }: GetServiceLogsParams<ContainerModule>,
-) {
+interface GetKubernetesLogsParams extends GetServiceLogsParams {
+  context: string
+  namespace: string
+  selector: string
+}
+
+export async function getServiceLogs(params: GetServiceLogsParams<ContainerModule>) {
+  const { ctx, service } = params
   const context = ctx.provider.config.context
   const namespace = await getAppNamespace(ctx, ctx.provider)
+  const selector = `service=${service.name}`
 
-  const proc = tail
-    ? await tailLogs(context, namespace, service, stream, log)
-    : await getLogs(context, namespace, service, stream)
+  return getKubernetesLogs({ ...params, context, namespace, selector })
+}
+
+export async function getKubernetesLogs(params: GetKubernetesLogsParams) {
+  // Currently Stern doesn't support just returning the logs and exiting, it can only follow
+  const proc = params.tail
+    ? await tailLogs(params)
+    : await getLogs(params)
 
   return new Promise<GetServiceLogsResult>((resolve, reject) => {
     proc.on("error", reject)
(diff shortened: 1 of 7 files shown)


garden-service/src/plugins/kubernetes/logs.ts
```diff
@@ -0,0 +1,123 @@
...
+  context: string
+  namespace: string
+  selector: string
+}
+
+export async function getServiceLogs(params: GetServiceLogsParams<ContainerModule>) {
+  const { ctx, service } = params
+  const context = ctx.provider.config.context
+  const namespace = await getAppNamespace(ctx, ctx.provider)
+  const selector = `service=${service.name}`
```
drubin: Side note we should really come up with a better name here or annotate it with `io.garden.service` service is way too generic.  (obviously something we shouldn't do now)

edvald: Yes I agree, noticed the same thing, just didn't want to lump that in here.

garden-service/src/plugins/kubernetes/logs.ts
```diff
@@ -0,0 +1,123 @@
...
+    .on("data", (line) => {
+      if (!line || line[0] !== "{") {
+        return
+      }
+      const obj = JSON.parse(line)
+      const [timestampStr, msg] = splitFirst(obj.message, " ")
+      try {
+        timestamp = moment(timestampStr).toDate()
+      } catch { }
+      void stream.write({ serviceName: service.name, timestamp, msg: msg.trimRight() })
```
drubin: I understand why this  (`serviceName: service.name` instead of pod name) is here but its really a nice feature being able to easily see when the deployment happens is insanely useful.

Is it not possible to default to the pod name with coloured output?

drubin: This also hides which container the logs come from with mutli pod containers (i.e sidecar).

Do we run a separate stern process for each service?

edvald: The main point was to have a consistent structure across all providers. We could (and should) add an optional metadata field that the provider can populate, and possibly just prefix the log message with the coloured pod name in the K8s provider. But yea let's do it separately.

edvald: Oh and yes, we do run a separate stern process per service, for the moment.

Please file an issue with feedback and suggestions, happy to improve this.

PR #415, opened 2018-12-03
fix(k8s): remove replicasets and daemonsets when deleting services

### LINKED ISSUE #379: Deleting a container service doesn't remove the replica sets
This seems to be a limitation of the Kubernetes API, see [this issue](https://github.com/kubernetes/kubernetes/issues/34052).

edvald:
Closes #379

### DIFF
--- a/garden-service/src/plugins/kubernetes/deployment.ts
+++ b/garden-service/src/plugins/kubernetes/deployment.ts
@@ -528,7 +528,7 @@ export async function deleteContainerService(
     namespace,
     labelKey: "service",
     labelValue: serviceName,
-    objectTypes: ["deployment", "service", "ingress"],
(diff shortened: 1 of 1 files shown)


garden-service/src/plugins/kubernetes/deployment.ts
```diff
@@ -528,7 +528,7 @@ export async function deleteContainerService(
     namespace,
     labelKey: "service",
     labelValue: serviceName,
-    objectTypes: ["deployment", "service", "ingress"],
+    objectTypes: ["deployment", "replicaset", "pod", "service", "ingress", "daemonset"],
```
drubin: Should we clean up the configmaps in the --meta namespace too?

edvald: Hmm, I think they should be disconnected usually. But we can revisit that later maybe.

PR #416, opened 2018-12-03
Dashboard front-end pt1

eysi09:
Added basic scaffolding and a _Overview_ and _Logs_ section. See readme for details on how to run, build and develop.


garden-dashboard/gulpfile.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * copyright (c) 2018 garden technologies, inc. <info@garden.io>
```
edvald: Probably doesn't matter, but I wonder why this header is lowercased?

garden-dashboard/src/api/index.ts
```diff
@@ -0,0 +1,130 @@
...
+  version?: string
+  state?: string
+  runningReplicas?: number
+  ingresses?: ServiceIngress[],
+  lastMessage?: string
+  lastError?: string
+  createdAt?: string
+  updatedAt?: string
+  detail?: any
+}
```
edvald: I'm looking at all these copied types, and also thinking that it might make sense to include dashboard under the `garden-service` project. Going forward I'd love to have this separation, but preferably only after we A) run `garden-service` as an actual singleton service, and B) we've exported the shared types in a library that we can import across the services. We'd also have issues including the dashboard in a container build like this.

garden-dashboard/src/styles/variables.ts
```diff
@@ -0,0 +1,72 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+export const tablet = 768
+export const desktop = 992
+export const big = 1680
+
+export const MOBILE_SECTIONS_MARGIN = "48px"
+// TODO Find proper fonts and set weights
```
edvald: I guess we may not be allowed to distribute JT Marnie? Or is that solvable?

gulpfile.ts
```diff
@@ -16,10 +16,10 @@ import { isString, clone, extend } from "lodash"
 const gulp = require("gulp")
 const checkLicense = require("gulp-license-check")
 
-const tsSources = ["garden-service/src/**/*.ts"]
-const pegjsSources = "src/*.pegjs"
+const tsSources = ["garden-service/src/**/*.ts", "dashboard/src/**/*.ts*"]
```
edvald: This should be `garden-dashboard`?

garden-dashboard/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "scripts": {
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test",
+    "eject": "react-scripts eject"
+  },
+  "eslintConfig": {
+    "extends": "react-app"
+  },
+  "proxy": "http://localhost:9777",
```
edvald: How does this property work?

garden-dashboard/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "axios": "^0.18.0",
+    "classnames": "^2.2.6",
+    "emotion": "^10.0.0",
+    "flexboxgrid": "^6.3.1",
+    "lodash": "^4.17.11",
+    "node-sass": "^4.10.0",
+    "react": "^16.6.3",
+    "react-dom": "^16.6.3",
+    "react-router-dom": "^4.3.1",
+    "typescript": "^3.1.6"
```
edvald: This is a dev dependency, no?

garden-dashboard/gulpfile.ts
eysi09: Ah, I've probably pressed `u` instead of `y`. That is, lowercased the text instead of yanking it.

garden-dashboard/src/api/index.ts
eysi09: Unfortunately we can't import content from outside the `garden-dashboard/src` directory. This is a limitation of `create-react-app`. We might be able to workaround it by "ejecting" the app. I say we revisit later.

garden-dashboard/package.json
eysi09: Ah yes, this was because of CORS error. I've removed this and instead added a proxy handler to `setupProxy.js`. See also readme.

Issue #237, opened 2018-08-17
Add status line at bottom in `garden dev`


edvald:
Currently it doesn't quite feel clear that the command is waiting for code changes. 

We should at least add a fixed line that says something like "Waiting for changes to code" (like we had in our prototype), but it would be even better if the line changes when processing is happening. For that we'd need to rig some basic event emitter, but we'll want to do that anyway and we can keep it simple to start.

@eysi09 this one might be for you :)

PR #417, opened 2018-12-04
refactor: nicer logging when watching for changes / hot reloading

### LINKED ISSUE #402: Log entry for hot reloading is misplaced
This seems to be a regression introduced in https://github.com/garden-io/garden/pull/292.

screenshot 2018-11-30 at 08 32 14

The log entry for the hot reload operation at the bottom of the screenshot isn't updated when hot reloading finishes, but instead the updated log entry is rendered above the command header (and the in-progress log entry is still shown with a spinner).

edvald:
Closes #402 and #237

### DIFF
--- a/garden-service/src/cli/cli.ts
+++ b/garden-service/src/cli/cli.ts
@@ -266,6 +266,10 @@ export class GardenCli {
       // entries (i.e. print new lines).
       const log = logger.placeholder()
 
+      // We pass a separate placeholder to the action method, so that commands can easily have a footer
+      // section in their log output.
+      const logFooter = logger.placeholder()
+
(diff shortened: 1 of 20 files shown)


drubin:
@edvald The formatting looks a little broken if we make a change to a `garden.yml` file the gear icon feels like its in the middle of nowhere and the -> don't line up.

Before with no output (it just carries on reloading)
screenshot 2018-12-04 14 55 00

This branch after making a change to `garden.yml` 
screenshot 2018-12-04 14 53 43


This is reproducible in the `hot-reload` example. 

Just run `garden-local dev` and make a change to any `garden.yml` file.

edvald:
Good catch! I'll fix that

PR #422, opened 2018-12-07
feat(k8s): make hot reloading work for remote clusters

edvald:
This amends the hot reloading flow for K8s containers to use port-forwarding instead of finding the hostname via ingress (which wasn't going to work in all cases anyway), which makes it work securely across the wire.

I also happened upon issues with the "hot reload scheduler" which wasn't handling promise errors correctly. I made a `HotReloadTask` instead, which serves the same purpose - which was to queue up and deduplicate rapid successions of sync attempts - but captures and handles errors properly (and does away with superfluous code along the way).


garden-service/src/plugins/kubernetes/hot-reload.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export async function hotReload(
+  { ctx, log, runtimeContext, module, buildDependencies }: HotReloadParams<ContainerModule>,
+): Promise<HotReloadResult> {
+  const hotReloadConfig = module.spec.hotReload!
+  const services = module.services
+
+  if (!await waitForServices(ctx, log, runtimeContext, services, buildDependencies)) {
+    // Service deployment timed out, skip hot reload
+    return {}
```
drubin: This feels bad to just ignore errors? can't we throw an exception or make it more visible it didn't work?

garden-service/src/plugins/kubernetes/deployment.ts
```diff
@@ -443,6 +443,9 @@ function configureHotReload(deployment, container, serviceSpec, moduleSpec, env,
     name: "garden-rsync",
     image: "eugenmayer/rsync",
     imagePullPolicy: "IfNotPresent",
+    env: [
+      { name: "ALLOW", value: "127.0.0.1 192.168.0.0/16 172.16.0.0/12" },
```
drubin: What are these values for?

Why do you have both local loopback and private ip ranges but are ignoring 10.0.0/8.

edvald: I just took the defaults from the container and added 127.0.0.1, not being entirely sure if different situations call for different values.

garden-service/src/plugins/kubernetes/hot-reload.ts
edvald: This is not new in the PR (see the old code), maybe @thsig can shed light on this.

garden-service/src/plugins/kubernetes/deployment.ts
edvald: Added a comment to explain.

edvald:
Oh and I should add - I started this because I was working on providing hot-reloading to user-defined Helm charts. The changes turned out to be necessary for that, but I split this part out of the PR because it's useful by itself.

garden-service/src/plugins/kubernetes/hot-reload.ts
edvald: But if I understand correctly, the user will see an error bubble up elsewhere, and an error thrown here would be superfluous.

drubin: We should at the least log it...

This behaviour is the exact same as the success result of returning an empty hash.

garden-service/src/plugins/kubernetes/deployment.ts
drubin: again we should add 10.0.0.0/8 here if we are assuming that we don't connect to the local IP else this wont work on any deployment that uses 10.0.0.0 as a POD ip range. 

Also, I think its safer to just do `ALLOW: value "0.0.0.0/0"` and have k8s deal with the security of the pod it logically makes more sense and maps to the bind ip of all web servers already running.

garden-service/src/plugins/kubernetes/hot-reload.ts
thsig: Looking back at this, I think we should throw an exception here (or in `waitForServices`) if a timeout occurs.

I think I wrote something of the sort at the time, but it seems to have ended up being omitted when I was refactoring/pruning the original hot reload branch at the time.

thsig:
This moving of the hot reload operation to `HotReloadTask` is a good idea.

The original thought was: Since a hot reload operation doesn't depend on any other tasks, and since it should be executed as soon as possible (regardless of whether capacity was available in `TaskGraph`), that it made sense to execute it outside the task system.

However, in hindsight, this approach (i.e. wrapping hot-reloading in a task) is simpler / "less special", and fits better into the general infrastructure.

edvald:
@drubin Updated with changes related to your comments. Please take a look.

garden-service/src/plugins/kubernetes/hot-reload.ts
```diff
@@ -0,0 +1,79 @@
...
+  const namespace = await getAppNamespace(ctx, ctx.provider)
+
+  const procs = await Bluebird.map(services, async (service) => {
+    // Forward random free local port to the remote rsync container.
+    const rsyncLocalPort = await getPort()
+
+    const targetDeployment = `deployment/${service.name}`
+    const portMapping = `${rsyncLocalPort}:${RSYNC_PORT}`
+
+    log.debug(
```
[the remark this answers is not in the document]
edvald: Hmm. This entry shows up for me, can't think of a way debug it on my end. Does this happen in other places?

edvald:
I made a delightfully dumb mistake there 🤦‍♂️ Updated commit, and tested again, works for me. Might be worth a double-check when you have a moment.

drubin:
The performance is about 3x worse than the nodeport version currently on master which is already noticeable.

120ms is on the border of it being too slow 400ms is simply not ok, and this will be much worse for remote clusters.

Current master (I tested it once but its always been roughly 100s - 150ms, and I have tested this a lot)

`✔ node-service              → Hot reloading... → Done (took 145 ms)`

This branch

```
✔ node-service              → Hot reloading... → Done (took 351 ms)
✔ node-service              → Hot reloading... → Done (took 420 ms)
```

I think the finding an open port and port-forwarding can't happen on every change we need to do them when we deploy the service. 

@edvald I think if we move the port-forwarding to a single tunnel and stream the changes down that we could get this down to acceptable limits? What do you think?

edvald:
Managing a single tunnel might be tricky, but worth exploring. I can take a look at that, but would prefer to defer it to another PR. This PR also addresses potential bugs when working locally, and I would be okay with an increased response time while we dig into it.

edvald:
I just took a stab at persisting tunnels, and noticed it'll be much cleaner to implement once I'm done with my Helm branch, because it changes the `hotReload` plugin action to work at a service level, as opposed to per-module. We can either hold off on this, or make a separate PR if/when we find something that improves the performance measurably.

edvald:
Re-visiting this. @drubin would you be okay with me splitting the re-used tunnels into a separate issue? Otherwise this needs to wait for another week or so, until my WIP Helm changes are merged.

eysi09:
@drubin, regarding your requested changes, it appears @edvald already [addressed them](https://github.com/garden-io/garden/pull/422#issuecomment-448306167). I've also tested this on my end and everything seems to work fine, apart from the perf regression. 

The perf regressions themselves are addressed in #458 which is also a part of the 0.9 milestone. With that in mind, should we merge this one?

garden-service/src/plugins/kubernetes/hot-reload.ts
```diff
@@ -0,0 +1,79 @@
...
+  const hotReloadConfig = module.spec.hotReload!
+  const services = module.services
+
+  await waitForServices(ctx, log, runtimeContext, services, buildDependencies)
+
+  const namespace = await getAppNamespace(ctx, ctx.provider)
+
+  const procs = await Bluebird.map(services, async (service) => {
+    // Forward random free local port to the remote rsync container.
+    const rsyncLocalPort = await getPort()
```
thsig: If the `getPort()` call were moved outside this `Bluebird.map` , could the `concurrency: 1` limit be removed for the rest of this function body (since the ports would already have been generated sequentially)?

Or do the port-forward `kubectl` calls themselves also need to be run in sequence?

Regardless, I'm sure this is fine for now, given that we're planning to re-use connections (https://github.com/garden-io/garden/issues/458).

PR #429, opened 2018-12-10
fix: do not run dependant tasks unless updated services depend on them

edvald:
Tasks should not run if they are leaf nodes after watch triggers.
Otherwise tasks that have no dependants get run automatically, and we
want users to be able to define tasks that are only run manually.


edvald:
cc @ellenkorbes

PR #430, opened 2018-12-10
refactor(k8s): minor change to speed up container deploys

edvald:


garden-service/src/plugins/kubernetes/deployment.ts
```diff
@@ -232,8 +232,9 @@ export async function createDeployment(
     deployment.spec.strategy = {
       type: "RollingUpdate",
       rollingUpdate: {
-        maxUnavailable: "34%",
-        maxSurge: "34%",
+        // This is optimized for fast re-deployment.
+        maxUnavailable: "100%",
```
drubin: I think we should explicitly set it to `1` instead of 100% since we are already only handling single pods for now, if some one does scale this this has unexpected behaviour like doubling the required resources. 

We should also maybe set `terminationGracePeriodSeconds` to 5 seconds. Any development pod that handles this should be enough and any that doesn't wont terminate gracefully with in the default 30 seconds any way. (this should improve things even more).

What is curious for me is that during deployments do we want to show the failure during deploy, or do we want to keep the service "up".

I have found it pretty annoying that during deploy's there is such a long time for the pods to be added back to the ingress controller but I do prefer that to showing "stale" outdated api's

edvald: Good idea. And yea, I think the main thing is to surface problems quickly. Any more ideas on how to do that would be great, for now I'll update with the suggestions above.

drubin: Change the health checks too they don't make super much sense. Takes a long time for things to become healthy?

edvald:
I've updated this one @drubin

drubin:
@edvald Cool lets do the others in another PR when we have more. 

Just rebase and merge when you are ready.

PR #431, opened 2018-12-10
feat(versioncmd): adds version command to return current version

drubin:
Adds a version command so it matches all a bunch of other tools.

`garden version` and `garden -v` return the same result.

### DIFF
--- a/docs/reference/commands.md
+++ b/docs/reference/commands.md
@@ -801,3 +801,12 @@ Throws an error and exits with code 1 if something's not right in your garden.ym
 
     garden validate 
 
+### garden version
+
+Show&#x27;s the current cli version.
+
+
+##### Usage
+
+    garden version 
+
--- a/garden-service/src/cli/helpers.ts
+++ b/garden-service/src/cli/helpers.ts
@@ -173,3 +173,8 @@ export function failOnInvalidOptions(argv, ctx) {
     ctx.cliMessage(`Received invalid flag(s): ${invalid.join(", ")}`)
   }
 }
+
+export function getPackageVersion(): String {
+  const version = require("../../package.json").version
+  return version
+}
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.md
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.md
@@ -31,7 +31,7 @@ about: Create a report to help make garden better
 <!-- PLEASE FILL THIS OUT -->
 
 <!-- Please run and copy and paste the results  -->
-`garden -v`
+`garden version`
 
 `kubectl version`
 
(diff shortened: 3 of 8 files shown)


garden-service/src/cli/helpers.ts
```diff
@@ -173,3 +173,8 @@ export function failOnInvalidOptions(argv, ctx) {
     ctx.cliMessage(`Received invalid flag(s): ${invalid.join(", ")}`)
   }
 }
+
+export function getVersion(): String {
+  const version = require("../../package.json").version
```
edvald: Does this work in a packaged build?

garden-service/src/cli/helpers.ts
```diff
@@ -173,3 +173,8 @@ export function failOnInvalidOptions(argv, ctx) {
     ctx.cliMessage(`Received invalid flag(s): ${invalid.join(", ")}`)
   }
 }
+
+export function getVersion(): String {
```
edvald: I think this should have a more verbose name, to distinguish from other types of versions we have all over the place.

docs/reference/commands.md
```diff
@@ -786,3 +786,12 @@ Throws an error and exits with code 1 if something's not right in your garden.ym
 
     garden validate 
 
+### garden version
+
+Show version number
```
edvald: I think this needs further explanation, to state that it's the version of the Garden CLI, as opposed to project, module etc.

garden-service/src/cli/helpers.ts
drubin: This is exactly how we calculate it from garden -v

docs/reference/commands.md
drubin: Exactly the same help text as - v didn't want to make them differ although I am happy to change this.

garden-service/src/cli/helpers.ts
edvald: Ok cool, just wasn't sure.

docs/reference/commands.md
edvald: Also, we should use proper punctuation in help text (I forget that all the time, but it looks much nicer in the docs).

drubin: Updated both.

eysi09:
Actually, setting the `implicitCommand` flag to `true` in this [line](https://github.com/garden-io/garden/blob/master/garden-service/src/cli/cli.ts#L186) allows you to call `garden version` with the desired outcome. However, that won't print `version` in the command list when running `garden --help`. Same applies to `help`. My apologies for not having noticed this sooner.

drubin:
I don't have a preference, although right now I would err on leaving it there. Would want to make it as easy as possible to get the version out also feels a bit premature to just remove it.

PR #433, opened 2018-12-10
Allow explicitly picking port for api in dev cmd

drubin:
via GARDEN_SERVER_PORT env variable

`GARDEN_SERVER_PORT=9778 GARDEN_ENABLE_SERVER=1 garden dev`

This should solve a bunch of hacks getting the dashboard port to update every time you run the a command.

### DIFF
--- a/garden-service/src/process.ts
+++ b/garden-service/src/process.ts
@@ -144,7 +144,9 @@ export async function processModules(
 
   // Experimental HTTP API and dashboard server.
   if (process.env.GARDEN_ENABLE_SERVER === "1") {
-    await startServer(garden, log)
+    // allow overriding automatic port picking
+    const port = Number(process.env.GARDEN_SERVER_PORT) || undefined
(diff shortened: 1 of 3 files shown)


garden-service/src/process.ts
```diff
@@ -144,7 +144,9 @@ export async function processModules(
 
   // Experimental HTTP API and dashboard server.
   if (process.env.GARDEN_ENABLE_SERVER === "1") {
-    await startServer(garden, log)
+    // allow overriding automatic port picking
+    const port = Number(process.env.GARDEN_SERVER_PORT) || undefined
```
eysi09: Should we validate that `process.env.GARDEN_SERVER_PORT` is a valid (port) number? `Number` just returns `NaN` if the parameter is not a number.

drubin: It will default to the random port if it isn't valid and it will throw and error if it can't bind to the port so I don't think it's nessary.

PR #437, opened 2018-12-10
Fix test outputs

edvald:


edvald:
FYI @eysi09 and @drubin - That `-i` flag was important. Without it we don't get log outputs or exit codes. I added a comment to explain.

drubin:
@edvald shouldn't we then add this to the exec thing as well? I can imagine people doing `garden exec xxx` and wanting to know if it was successful

PR #440, opened 2018-12-11
Release 0.8.1

edvald:


CHANGELOG.md
```diff
@@ -1,70 +1,162 @@
 
+<a name="v0.8.1"></a>
```
drubin: Why is the changelog so different from the previous versions?

thsig: Maybe @eysi09 knows more? IIRC, he's the changelog czar.

eysi09: Are you referring to the `-rc` tags? It's because the changelog generator looks at the local tags to and generates it based on them.

So to get rid of those `-rc` logs, you need to first remove them from your local git tag list, and then generate the changelog. They get re-added on `git fetch`. 

My apologies for not having documented this better. Also, wasn't sure if we actually wanted to keep the tag logs or not. 

At a quick glance I can't see if it's possible to specify the tag format.

eysi09: I also just noticed that it's possible to generate the changelog before creating the tag by using the `--next-tag` flag. I'll update the docs as soon as I have the time.

drubin: I mean it modifying much older release notes.

eysi09: Yeah it's the same issue, it's picking up the `-rc` tags. I think I'll just submit a PR for to the repo with something like a `--tag-regex` flag. That should fix all our issues and reduce the changelog generation to just a single step. No tag mangling.

eysi09: ...but for now I'll just fix the changelog manually so that we can get this merged.

drubin: https://stackoverflow.com/questions/3867619/how-to-get-last-git-tag-matching-regex-criteria

I think their approach would work for us too `git-chglog $(git describe --tags $(git rev-list --tags --max-count=1))`

we just need to replace so it includes the 2 latest stable releases.

drubin: Also I don't think we should manually mangle this just to get it merged we should just fix it properly instead of duplicating the work. We have already released `0.8.1`.

Its going to bite us the next time we do a release and we haven't fixed this.

eysi09: Yeah, I'm working on it. Another thing I noticed, we filter on the following commit types:  **feat**, **fix**, **perf**, **refactor**, whereas the angular commit convention has the following types: **feat**, **fix**, **perf**, **refactor**, **docs**, **style**, **test** and **chore*.

So basically we're filtering out **docs**, **style**, **test**, and **chore**.  Any thoughts on this? Shouldn't we just include all of them?

edvald: I quite like this setup, personally, since users will be less likely to care about the filtered types. But I don't feel strongly about it, so feel free to make a call.

drubin: @eysi09 I also think we should copy the relevant parts of the angular docs and include them in our `contributing.md`. Linking to external docs that could change feels wrong it also makes it another click to see what we support. 

I think docs is good to filter from change log I am not sure it makes sense to filter any thing that actually touches code/config. Adding additional tests/examples doesn't actually affect the binary.

The point of a change log is to get an overview of any changes that might impact your usage so any thing that fits that should be included. 

Same goes for library upgrades? I think I have been putting them under `chore`, I am not sure they should be (unless they are dev dependancies)

eysi09: The changelog now only prepends the entries for `v0.8.1-rc1` and `v0.8.1` and leaves the older entries as they were.

CONTRIBUTING.md
```diff
@@ -74,22 +87,24 @@ Currently we maintain the following:
...
-   Browse to the job marked `release-service-pkg`, open the _Artifacts_ tab and download the listed artifacts.
-9. Go to our Github [Releases tab](https://github.com/garden-io/garden/releases) and click the **Draft a new release** button.
-10. Fill in the **Tag version** and **Release title** fields with the new release version (same as you used for the tag).
-11. Upload the downloaded artifacts.
-12. Write release notes (not necessary for RCs). The notes should _at least_ contain the changelog.
-13. Click the **Publish release** button.
-14. Push the branch and make a pull request.
-15. If you're making an RC, you're done! Otherwise, you need to update Homebrew package: `gulp update-brew`
+4. Create a new tag for the version: `git tag -a v0.8.0-rc2 -m 'chore(release): release v0.8.0-rc2'`
+5. Get the changelog for the tag with: `git-chglog v0.8.0-rc2` and prepend the output to `CHANGELOG.md`. Alternatively you can generate the entire `CHANGELOG.md` file with `git-chglog -o=CHANGELOG.md v0.1.0..v0.8.0-rc2`, but make sure that older entries do not change (this can happen if you have local tags that do not match the remote release tags).
```
drubin: Are these steps still the same?

eysi09: I've updated them compared to the original PR. Does this not make sense?

Sidenote: I figured we'd just keep the `-rc` entries in the changelog, otherwise we'd have to manually remove them (given our current tooling).

Issue #420, opened 2018-12-06
Add a get tasks command


thsig:
We should add a `description` field to task definitions, and a `get tasks` command that outputs a project-wide list of all tasks (and their descriptions).

The output could look something like this:
```yaml
some-module:
  migrate-db:
    description: Migrate the DB to the latest version. Idempotent.
    command: ...
  download-dev-db:
    description: Download the latest internal dataset to populate your local dev database.
    command: ...
some-other-module:
  ...
```

PR #444, opened 2018-12-17
feat: added get tasks command

thsig:
Added a get tasks command, which lists all the project's tasks by module. Also added a `description` field to task configs.

Implements https://github.com/garden-io/garden/issues/420.

Example output (when run in the `tasks` example project):
```yaml
$ garden get tasks
- tasks for module hello:
    - name: node-migration
      command:
         - knex
         - 'migrate:latest'
      description: Creates the users table.
      dependencies:
         - postgres
- tasks for module postgres: []
- tasks for module user:
    - name: ruby-migration
      command:
         - rake
         - 'db:migrate'
      description: Populates the users table with a few records.
      dependencies:
         - node-migration
```

### DIFF
--- a/garden-service/src/commands/get/get-tasks.ts
+++ b/garden-service/src/commands/get/get-tasks.ts
@@ -1,0 +1,98 @@
+/*
+ * Copyright (C) 2018 Garden Technologies, Inc. <info@garden.io>
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import chalk from "chalk"
+import * as dedent from "dedent"
+import * as indentString from "indent-string"
+import { sortBy, omit, uniq } from "lodash"
+import {
+  Command,
+  CommandResult,
+  CommandParams,
+  StringsParameter,
+} from "../base"
+import { logHeader } from "../../logger/util"
+import { Task } from "../../types/task"
+
+const getTasksArgs = {
+  tasks: new StringsParameter({
+    help: "Specify task(s) to list. Use comma as a separator to specify multiple tasks.",
+  }),
+}
+
+type GetTasksArguments = typeof getTasksArgs
+
+export function prettyPrintTask(task: Task): string {
+  let out = dedent`
+  ${chalk.cyan.bold(task.name)}
+    ${printField("command", (task.spec.command || []).join(" "))}
+  `
+
+  if (task.spec.description) {
+    out += "\n" + indentString(printField("description", task.spec.description), 2)
+  }
+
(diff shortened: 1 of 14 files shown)


garden-service/src/config/task.ts
```diff
@@ -26,6 +27,7 @@ export const baseTaskSpecSchema = Joi.object()
     name: joiIdentifier()
       .required()
       .description("The name of the task."),
+    description: Joi.string().optional(),
```
thsig: Just noticed that the description string for this field is missing. Adding that now.

edvald:
First thing I notice is that the YAML output looks highly unusual. The machine-readable (`--output=yaml`) will be practically unusable like this. I'd suggest rather making a sanely formatted object, then doing something other than the highlighted YAML to print out to the console.

thsig:
@edvald Yeah, good point. How about:

- Replacing the `tasks for module [module name]` key with just the module name in the command result; and
- Logging the tasks for each module as YAML, and using a "vanilla"/ non-YAML heading to indicate which module the tasks come from?

Or did you have some other format in mind?

garden-service/src/commands/get/get-tasks.ts
```diff
@@ -0,0 +1,49 @@
...
+        tasks.filter(t => t.module.name === m.name),
+        t => t.name)
+        .map(t => ({
+          ...omit(t.config.spec, ["timeout"]),
+          name: t.name,
+          description: t.config.spec.description,
+          dependencies: t.config.spec.dependencies,
+        }))
+
+      taskListing.push({ [`tasks for module ${m.name}`]: tasksForModule })
```
edvald: This could easily be a `map` instead of a for-loop. Also gets rid of the `any[]` type above.

garden-service/src/commands/get/get-tasks.ts
```diff
@@ -0,0 +1,49 @@
...
+  Command,
+  CommandResult,
+  CommandParams,
+} from "../base"
+
+export class GetTasksCommand extends Command {
+  name = "tasks"
+  help = "Lists the tasks defined in your project's modules."
+
+  async action({ garden, log }: CommandParams): Promise<CommandResult> {
```
edvald: A nice-to-have would be an argument or two to filter the tasks, not a must for the PR though.

thsig: Yeah, true that. E.g. a list of modules. Any other filtering options come to mind?

edvald: Maybe an optional argument for task name(s)? Similar to other commands

edvald:
The human-readable output can be done in a variety of ways. Don't really have a strong opinion on it, just that it shouldn't really be YAML (we've been wanting to change other command outputs like that too).

But the formal output should be something that's easy to digest. I'd suggest a flat list, with a module name on each item, rather than nested lists.

PR #446, opened 2018-12-19
Adds Service port config option to container module

### LINKED ISSUE #400: Allow service port definition to differ from container port
## Problem 

A very common usecase is the container to listen on a none privliged port i.e 8080 but have the service connect to `http://user/` this is achived by port mapping in both kubernetes and docker-compose. 

Our default option of defaulting to the container port is a great default but pretty limiting when you are working with pre-built containers/modules and pre-built code that makes assumptions.

### Introduce `servicePort`
Possible solution

```
ports: 
        #
        # Required.
        - # The name of the port (used when referencing the port elsewhere in the service
          # configuration).
          #
          # Required.
          name:
​
          # The port number on the container defaults to the service endpoint's port
          #
          # Required.
          containerPort:

          # The port number on the service defaults to `containerPort`
          #
          # optional 
          servicePort:
​
```

### Mapping values 

Use a similar syntax to docker and docker-compose where you map a port. 

ie `docker run -p 8080:9090` 

```
ports: 
        #
        # Required.
        - # The name of the port (used when referencing the port elsewhere in the service
          # configuration).
          #
          # Required.
          name:
​
          # The port number on the container defaults to the service endpoint's port
          #
          # Required.
          containerPort:  "80:9090"

          # The port number on the service defaults to `containerPort`
          #
          # optional 
          servicePort:
​
```

My personal preference is for option of introducing a `servicePort`

drubin:
Adds a `servicePort` configuration that if not there will default to `containerPort` this is the value that the service will be exposed on. 

I will fix  and improve all the docs / help text.

I would also like to add a doc that explains the differences between our service ports and how they work and what they map to in their native terms. 

fixes #400

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -643,15 +643,27 @@ module:
           # Required.
           name:
 
-          # The protocol of the service container port.
+          # The protocol of the port.
           #
           # Optional.
           protocol: TCP
 
-          # The port number on the service container.
+          # The port exposed on the container by the running procces. This will also be the
+          # default value for `servicePort`.
+          # `servicePort:80 -> containerPort:8080 -> process:8080`
+          #
+          # Example: "8080"
           #
           # Required.
           containerPort:
+
+          # The port exposed on the service. Defaults to `containerPort` if not specified.
+          # `servicePort:80 -> containerPort:8080 -> process:8080`
+          #
+          # Example: "80"
+          #
+          # Optional.
+          servicePort: <containerPort>
 
           hostPort:
 
--- a/garden-service/src/plugins/container.ts
+++ b/garden-service/src/plugins/container.ts
@@ -60,6 +60,8 @@ export interface ServicePortSpec {
   name: string
   protocol: ServicePortProtocol
   containerPort: number
+  // Defaults to containerPort
+  servicePort: number
   hostPort?: number
   nodePort?: number | null
 }
@@ -164,7 +166,7 @@ const healthCheckSchema = Joi.object()
       .description("Set this to check the service's health by checking if this TCP port is accepting connections."),
   }).xor("httpGet", "command", "tcpPort")
 
-const portSchema = Joi.object()
+export const portSchema = Joi.object()
   .keys({
     name: joiIdentifier()
       .required()
@@ -172,10 +174,21 @@ const portSchema = Joi.object()
     protocol: Joi.string()
       .allow("TCP", "UDP")
       .default(DEFAULT_PORT_PROTOCOL)
-      .description("The protocol of the service container port."),
+      .description("The protocol of the port."),
     containerPort: Joi.number()
       .required()
-      .description("The port number on the service container."),
+      .example("8080")
+      .description(deline`
+        The port exposed on the container by the running procces. This will also be the default value
+        for \`servicePort\`.
+
+        \`servicePort:80 -> containerPort:8080 -> process:8080\``),
+    servicePort: Joi.number().default((context) => context.containerPort, "<containerPort>")
+      .example("80")
+      .description(deline`The port exposed on the service.
+        Defaults to \`containerPort\` if not specified.
+
+        \`servicePort:80 -> containerPort:8080 -> process:8080\``),
     hostPort: Joi.number()
       .meta({ deprecated: true }),
     nodePort: Joi.number()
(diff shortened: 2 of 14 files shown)


garden-service/src/plugins/container.ts
```diff
@@ -60,6 +60,10 @@ export interface ServicePortSpec {
   name: string
   protocol: ServicePortProtocol
   containerPort: number
+  // Defaults to containerPort
+  // TODO: this should not be optional
+  // `src/plugins/local/local-google-cloud-functions.ts:40:15` breaks
+  servicePort?: number
```
drubin: @edvald if I make this required it starts throwing warnings in `src/plugins/local/local-google-cloud-functions.ts:40:15` and I can't really figure it out. 

I think since we have a default value making it not optional makes sense.

garden-service/src/plugins/container.ts
```diff
@@ -164,6 +168,10 @@ const healthCheckSchema = Joi.object()
       .description("Set this to check the service's health by checking if this TCP port is accepting connections."),
   }).xor("httpGet", "command", "tcpPort")
 
+// Generates a servicePort configuration which defaults to the containerPort
+const generateServicePort = (context) => context.containerPort
+generateServicePort.description = "The port number service will be exposed on. Defaults to `containerPort` if specified"
```
edvald: Hmm, I don't quite understand this. Why "generate" it?

garden-service/src/plugins/container.ts
```diff
@@ -175,7 +177,12 @@ const portSchema = Joi.object()
       .description("The protocol of the service container port."),
     containerPort: Joi.number()
       .required()
-      .description("The port number on the service container."),
+      .description(deline`
+        The port exposed on the container.
+        Will also be default port used for \`servicePort\` unless specified`),
+    servicePort: Joi.number().default((context) => context.containerPort, "<first specified environment>")
```
drubin: @edvald I can't get the doc generator to pull the description out if its a function.

I tried a bunch of options 

1. Second param to the default function
1. As the docs say adding a description field to the function

This seems to be the way we do it in other places, but that isn't working any tips?

```js
  dependencies: joiArray(buildDependencySchema)
        .description("A list of modules that must be built before this module is built.")
        .example([
          [{ name: "some-other-module-name" }],
          {},
        ]),
    })
      .default(() => ({ dependencies: [] }), "{}")
      .description("Specify how to build the module. Note that plugins may define additional keys on this object."),
```

We should add tests for the `docs` generator too :)

edvald: The docs generator currently doesn't try to pull out the second parameter to the `.default()` function. It instead calls the default value, if it is a function. The reason it's failing for you in this case is that you're using the `context` input variable to the default callable, which we've never done before and the docs generator doesn't support it.

The line to look at is `garden-service/src/docs/config.ts:89`. This needs to be augmented to handle this case somehow, for example by supplying an empty object to the callable, and falling back to the default parameter description (e.g. `"<first specified environment>"`) when the callable returns `undefined`.

eysi09:
> Regarding the docs I have updated them but I think we need a separate issue to properly document exposing a service because its a complex topic and I don't think we can do it justice in the pure reference guide.

I second that. Perhaps a comprehensive guide covering networking in general? I.e., how to configure different port types, how services communicate inside the cluster, exposing services, hostnames, etc. Feels like we (read: I) repeatedly struggle with this when setting up new projects. I'll create an issue for this where we can iterate on the structure and content.

drubin:
> Perhaps a comprehensive guide covering networking in general?  

For me it's also about knowing what the garden terminology maps to. For example what we call a service kubernetes calls a "deployment" and what we call a "port" is actually a service (kinda). 

So maybe also a few guides aimed at people coming from different backgrounds.

docs/reference/config.md
```diff
@@ -643,16 +643,23 @@ module:
...
           name:
 
-          # The protocol of the service container port.
+          # The protocol of the port.
           #
           # Optional.
           protocol: TCP
 
-          # The port number on the service container.
+          # The exposed exposed port on the container. Will also be the default `servicePort`.
```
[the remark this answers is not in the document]
drubin: https://github.com/garden-io/garden/pull/446/files#diff-8244d579ea2c1aaeafa671f998bfdd6dR10 

I added this in the sample project. `Maps service:80 -> container:8080`

I don't think what ever we do will do it justice other than a long winded doc explaining this. 

Will fix the typo and improve the text a little but its always going to "feel" under documented until we add a exposing services doc.

But maybe I should just do that as part of this PR I think it will maybe remove a bunch of these questions.

PR #447, opened 2018-12-19
fix(exec): missing tasks key in module schema

edvald:
A simple omission, seems @thsig had implemented the actual task action for the plugin.

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -301,6 +301,41 @@ module:
   env: 
     {}
 
+  # A list of tasks that can be run in this module.
+  #
+  # Optional.
+  tasks: 
+    # A task that can be run in this module.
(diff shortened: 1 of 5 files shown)


drubin:
@edvald This looks good, but I think we should start adding tests for high level features modules support. 

Can we maybe start with this one? i.e add a test where an exec module has a task and its a valid config, as we grow and things change I suspect it's going to get confusing which modules support which features.

Makes sense?

edvald:
Yea makes sense. I'm doing something like that for the Helm module type in another branch, just testing local parts and unit testing, but the `exec` module is much easier to work with. I'll add those.

edvald:
@drubin Added explicit tests to validate the module parsing.

PR #448, opened 2019-01-08
Fix doc links

eysi09:


docs/home.md
```diff
@@ -1,87 +0,0 @@
-[![CircleCI](https://circleci.com/gh/garden-io/garden/tree/master.svg?style=svg&circle-token=ac1ec9984d093f91e594e5a0a03b34cec2c2a093)](https://circleci.com/gh/garden-io/garden/tree/master)
```
drubin: Why did we remove this?

README.md
```diff
@@ -1,9 +1,8 @@
 [![CircleCI](https://circleci.com/gh/garden-io/garden/tree/master.svg?style=svg&circle-token=ac1ec9984d093f91e594e5a0a03b34cec2c2a093)](https://circleci.com/gh/garden-io/garden/tree/master)
 
 <p align="center">
-<img src="docs/logo.png" width="66%">
-</center>
-
+  <img src="docs/logo.png" width="66%">
+</p>
```
drubin: Thanks!

drubin:
```
FILE: examples/vote/README.md
[✖] http://vote.local.app.garden/
[✖] http://result.local.app.garden
```

I think we might need to add an ignore pattern for `*.local.app.garden` 

https://www.npmjs.com/package/markdown-link-check#config-file-format 
```
ignorePatterns: An array of objects holding regular expressions which a link is checked against and skipped for checking in case of a match.
```

docs/home.md
eysi09: Because the new main README.md replaces it (cc @ellenkorbes).

eysi09:
Yup. Missed those because curl-ing for them locally actually returns a valid response since I have the example running.

docs/home.md
eysi09: The readme is included, under _Introduction_. And what links does it break? There was one link to it in the table of contents that I removed.

eysi09: Also, cc @ellenkorbes who suggested we'd remove the _Home_ section.

Issue #410, opened 2018-12-03
Include filename when we have invalid scheme in garden.yml


drubin:
### Current Behavior

When a user makes a mistake in a `garden.yml` file with invalid syntax we don't show which file contains the error.

```
link module 🔗

Error validating module user: key "servicePort" is not allowed at path .services[0][ports][0][servicePort], key .services[0][ports] does not contain 1 required value(s)
module:

See error.log for detailed error message
```

### Expected behavior

We should include which file could not be correctly parsed. 

### Your environment

```
garden -v
0.8.0
```

drubin:
@thsig  I am not sure how easy it is to include the filename since I think that type of parsing happens higher up the chain, but even just the module/project would be an amazing improvement to knowing which file is invalid.

thsig:
Hadn't started on this, but config-related error messages do seem to include the path to the failing file on `master`. Looks like this was fixed since 0.8.

drubin:
@thsig The issue includes which version number this happens in. So it wasn't fixed in `0.8.0` 

Adding an invalid key ie `foo: bar` to a config in a module results in the following output on latest code

```
garden dev

Good morning! Let's get your environment wired up...

✔ local-kubernetes          → Configured
    ✔ ingress-controller        → Preparing chart... → Done (took 0.9 sec)
    ✔ kubernetes-dashboard      → Preparing chart... → Done (took 0.8 sec)
    ✔ default-backend           → Deploying version f4da12648f... → Done (took 4.5 sec)
    ✔ kubernetes-dashboard      → Deploying version f4da12648f... → Done (took 5.8 sec)
    ✔ ingress-controller        → Deploying version f4da12648f... → Done (took 21.7 sec)
✔ openfaas                  → Configured
    ✔ system                    → Preparing chart... → Done (took 1.2 sec)
    ✔ system                    → Deploying version f4da12648f... → Done (took 22.3 sec)

Error validating module hello-container: key "foo" is not allowed at path [foo]

See error.log for detailed error message
```

Same thing with `garden validate`

We don't link to the file. I am going to reopen this.

PR #450, opened 2019-01-09
fix: add path to module validation error messages

thsig:
When config validation fails for a module, the relative path (from project root) to the failing config file is included in the error message.

Resolves https://github.com/garden-io/garden/issues/410.

For example:
```
Error validating module user at user/garden.yml: key "foo" is not allowed at path [foo]
```
instead of
```
Error validating module user: key "foo" is not allowed at path [foo]
```
which is the current behaviour on `master`.

### DIFF
--- a/garden-service/src/config/common.ts
+++ b/garden-service/src/config/common.ts
@@ -11,6 +11,7 @@ import * as Joi from "joi"
 import * as uuid from "uuid"
 import { ConfigurationError, LocalConfigError } from "../exceptions"
 import chalk from "chalk"
+import { relative } from "path"
 
 export type Primitive = string | number | boolean
 
@@ -113,8 +114,39 @@ const joiOptions = {
 }
 
 export interface ValidateOptions {
-  context?: string
+  context?: string // Descriptive text to include in validation error messages, e.g. "module at some/local/path"
   ErrorClass?: typeof ConfigurationError | typeof LocalConfigError
+}
+
+export interface ValidateWithPathParams<T> {
+  config: T,
+  schema: Joi.Schema,
+  path: string, // Absolute path to the config file, including filename
+  projectRoot: string,
+  name?: string, // Name of the top-level entity that the config belongs to, e.g. "some-module" or "some-project"
+  configType?: string // The type of top-level entity that the config belongs to, e.g. "module" or "project"
+  ErrorClass?: typeof ConfigurationError | typeof LocalConfigError
+}
+
+/**
+ * Should be used whenever a path to the corresponding config file is available while validating config
+ * files.
+ *
+ * This is to ensure consistent error messages that include the relative path to the failing file.
+ */
+export function validateWithPath<T>(
+  { config, schema, path, projectRoot, name, configType = "module", ErrorClass }: ValidateWithPathParams<T>,
+) {
+
+  const validateOpts = {
+    context: `${configType} ${name ? name + " " : ""}in ${relative(projectRoot, path)}`,
+  }
+
+  if (ErrorClass) {
+    validateOpts["ErrorClass"] = ErrorClass
+  }
+
+  return validate(config, schema, validateOpts)
 }
 
 export function validate<T>(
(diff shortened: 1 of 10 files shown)


garden-service/src/config/common.ts
```diff
@@ -113,10 +116,29 @@ const joiOptions = {
...
+export function validateModuleConfig<P extends ValidateModuleParams, S extends Joi.Schema>(
+  { ctx, moduleConfig }: P,
+  moduleConfigSchema: S,
+  prefix = "module",
+) {
+  const configRelPath = relative(ctx.projectRoot, join(moduleConfig.path, MODULE_CONFIG_FILENAME))
+  return validate(
+    moduleConfig.spec,
+    moduleConfigSchema,
+    { context: `${prefix} ${moduleConfig.name} at ${configRelPath}` })
```
drubin: should this not be "in" at a filename feels a bit odd?

thsig: Yeah, good point. Will make that change.

thsig:
Any further comments on this one, @drubin?

drubin:
For me the only missing thing would be a test. I know its just a "display" value, but its something that I would like to never be removed by mistake. The error should always include the correct file name. 

Not sure how easily it would be to add a test that proves this works, but from just reading the code its hard to get that.

@thsig would it be possible to test this? have an actual case being parsed with an error?

garden-service/src/config/common.ts
```diff
@@ -113,10 +116,29 @@ const joiOptions = {
...
+  moduleConfigSchema: S,
+  prefix = "module",
+) {
+  const configRelPath = relative(ctx.projectRoot, join(moduleConfig.path, MODULE_CONFIG_FILENAME))
+  return validate(
+    moduleConfig.spec,
+    moduleConfigSchema,
+    { context: `${prefix} ${moduleConfig.name} in ${configRelPath}` })
+}
+
```
drubin: Thanks for providing docs for `validateModuleConfig` <3

Maybe also add a comment here when we should use just `validate` of if its just for internal / tests?

thsig:
Yeah. I thought about adding tests for this, but module config validation is sometimes done in contexts where a path is not available (e.g. when generating a module/project with the `new` command), so I thought it would be cleaner to recommend `validateModuleConfig` for those cases where it's available (mostly the plugin-specific module validation methods).

With this contextuality in mind, I'm not quite sure how to test "when the path is available during module config validation, it should be included in the error message".

That said, I'll make that doc comment a bit more instructive as you suggest.

drubin:
>Yeah. I thought about adding tests for this, but module config validation is sometimes done in contexts where a path is not available (e.g. when generating a module/project with the new command), so I thought it would be cleaner to recommend validateModuleConfig for those cases where it's available (mostly the plugin-specific module validation methods). 

But can we add a test for `validateModuleConfig` is what I originally meant.

Adding a test for functions not only ensures the functionality works but it also serves as an example for expected usage.

garden-service/test/src/config/base.ts
```diff
@@ -1,14 +1,30 @@
...
 
-  // TODO: test more cases + error cases
+  it("should not throw an error if no file was found", async () => {
+    const parsed = await loadConfig(projectPathA, resolve(projectPathA, "non-existent-module"))
+
+    expect(parsed).to.eql(undefined)
+  })
+
+  it("should throw a config error if the file couldn't be parsed°", async () => {
+    // throw new ConfigurationError(`Could not parse ${CONFIG_FILENAME} in directory ${path} as valid YAML`, err)
```
drubin: we should remove this line.

garden-service/test/src/config/common.ts
```diff
@@ -0,0 +1,43 @@
...
+      config: validModuleConfig,
+      schema: baseModuleSpecSchema,
+      name: validModuleConfig.name,
+      path: validModuleConfig.path,
+      projectRoot: projectPathA,
+    })
+  })
+
+  it("should include the module's relative path in the error message for invalid config", async () => {
+    validateWithPath({
```
drubin: Seems like there is a missing actual test there? Where do we  validate the return value/exception?

thsig:
I think this should be good now — if so, can we get it merged?

thsig:
Aha, good point. Let me know if there's anything else here that needs fixing/improving!

PR #451, opened 2019-01-09
refactor: use events for file watching instead of callbacks

edvald:
I started this as part of another branch, but figured I'd split this
into a new PR. This should have no functional differences to the prior
implementation, apart from exposing events that can be used in more
places. Added tests for the Watcher class along the way.


edvald:
It looks like the watcher just isn't working in CI, or at least partially failing. I'll need to investigate, no idea why this happens.

garden-service/src/events.ts
```diff
@@ -48,7 +48,27 @@ export class EventBus extends EventEmitter2 {
...
+  // Internal test/control events
+  _restart: string,
   _test: string,
+
+  // Watcher events
+  configAdded: {
+    path: string,
+  },
+  projectConfigChanged: {},
+  moduleConfigChanged: {
```
drubin: What are the difference between `moduleConfigChanged` and  `moduleChanged`

I think these events could definitely use a bit more of inline comments explaining how each one is used?

drubin: I can't really see how we go from having this type here to having a 

`garden.events.on("projectConfigChanged"....`

garden-service/src/process.ts
```diff
@@ -110,32 +108,40 @@ export async function processModules(
...
+      resolve()
+    })
+
+    garden.events.on("projectConfigChanged", () => {
+      log.info({ symbol: "info", msg: `Project configuration changed, reloading...` })
+      resolve()
+    })
+
+    garden.events.on("configAdded", (event) => {
+      log.info({ msg: `Garden config added at ${event.path}, reloading...`, symbol: "info" })
```
drubin: Can we stick to consistent order of params?  previous 2 examples are `symbol:, msg:` this is `msg:, symbol` makes it easier for humans to read it.

garden-service/src/events.ts
edvald: I've clarified the name a bit, to make the distinction clear. `moduleChanged` -> `moduleSourcesChanged`. Not sure I understand the second question?

garden-service/src/process.ts
edvald: Yea, fixed.

garden-service/src/events.ts
drubin: where does the event `projectConfigChanged` come from? I can't really see how these types map to where they used or what fires them.

I guess I just don't understand how it worked with out this and what adding this changed... 

I guess I would have expected these to be constants instead of referencing the strings "moduleSourcesChanged" in places.

edvald: The string keys are what make these strongly typed in the above `EventBus` class. We use mapped types to enforce the shape of the event payloads, and we wouldn't be able to do that with constants.

But, you're exactly right, we're missing the `projectConfigChanged` event from the Watcher. I missed it because I was translating from the prior implementation, which didn't have a 1:1 mapping to how I set this up. I'll fix that..

garden-service/test/helpers.ts
```diff
@@ -36,13 +36,15 @@ import { helpers } from "../src/vcs/git"
...
 import { EventBus, Events } from "../src/events"
-import { ValueOf } from "../src/util/util"
+import { ValueOf, Ignorer } from "../src/util/util"
 import { SourceConfig } from "../src/config/project"
 import { BuildDir } from "../src/build-dir"
 import timekeeper = require("timekeeper")
 
 export const dataDir = resolve(__dirname, "data")
 export const examplesDir = resolve(__dirname, "..", "..", "examples")
+export const tmpDir = resolve(__dirname, "..", "..", "tmp")
```
drubin: Where is this used?

edvald: It actually isn't anymore, since I changed the tests to not actually write to the file system.

garden-service/src/events.ts
edvald: Should be all set now.

PR #457, opened 2019-01-11
refactor(k8s): shorten default namespace names

### LINKED ISSUE #334: Namespace and helm release names can become too long
We need to rethink how we generate namespace names and release names. The current approach is simple but can fall apart when deploying to nested namespaces (e.g. the openfaas plugin dependencies).

It should remain deterministic, and ideally we should avoid having lookup tables somewhere. So we should consider how we might have a length-limited prefix that's user readable, and a hash suffix that makes it unique across users, projects etc.

We should limit the overall length based on Kubernetes identifier limitations, and the max length of Helm release names (which may be the same).

edvald:
BREAKING CHANGE:

After updating, any project that doesn't have an explicitly
configured namespace will be installed into a new namespace,
and the old namespace needs to be manually removed.

closes #334 
closes #50 

cc @drubin - consider it a birthday present :)


docs/reference/config.md
```diff
@@ -176,8 +176,8 @@ project:
         {}
 
       # Valid RFC1035/RFC1123 (DNS) label (may contain lowercase letters, numbers and dashes, must
-      # start with a letter, and cannot end with a dash) and additionally cannot contain
-      # consecutive dashes, or be longer than 63 characters.
+      # start with a letter, and cannot end with a dash), cannot contain consecutive dashes or
+      # start with `garden`, or be longer than 63 characters.
```
drubin: Maybe we should fix the docs for `joiIdentifier` so they match `joiUserIdentifier`. they should be the same.

garden-service/src/plugins/kubernetes/kubernetes.ts
```diff
@@ -153,6 +148,12 @@ const configSchema = kubernetesConfigBase
     ingressHttpsPort: Joi.number()
       .default(443)
       .description("The external HTTPS port of the cluster's ingress controller."),
+    namespace: Joi.string()
+      .default(undefined, "<username>--<project name>")
+      .description(
+        "Specify which namespace to deploy services to (defaults to <username>--<project name>). " +
+        "Note that the framework generates other namespaces as well with this name as a prefix.",
```
drubin: This is no longer valid? the other generated namespaces start with `garden-` as well? `garden-system`

garden-service/src/plugins/kubernetes/local.ts
```diff
@@ -58,6 +58,12 @@ export interface LocalKubernetesConfig extends KubernetesBaseConfig {
 
 const configSchema = kubernetesConfigBase
   .keys({
+    namespace: Joi.string()
+      .default(undefined, "<project name>")
+      .description(
+        "Specify which namespace to deploy services to (defaults to the project name). " +
```
drubin: Does it not default to `env-project-name`?

garden-service/src/plugins/kubernetes/namespace.ts
```diff
@@ -66,7 +66,7 @@ export async function getNamespace(
       )
     }
 
-    namespace = `garden--${username}--${ctx.projectName}`
+    namespace = `${username}--${ctx.projectName}`
```
drubin: I am really confused how we go from this to `garden--default--simple-project`

Does `username` map to "env" or does `${ctx.projectName}` have implicit `default-` but if that is the case how does `username` get removed?

docs/reference/config.md
edvald: They are explicitly different. The latter enforces the constraint that the identifier should not start with `garden`, the former does not.

garden-service/src/plugins/kubernetes/kubernetes.ts
edvald: `garden-system` is specific to the `local-kubernetes` plugin and wouldn't be documented here. What this refers to are the `--metadata` and `--openfaas` branches, for example, that are created as needed.

garden-service/src/plugins/kubernetes/local.ts
edvald: Hmm, not sure what you're referring to.

garden-service/src/plugins/kubernetes/namespace.ts
edvald: > I am really confused how we go from this to `garden--default--simple-project`

This change removes the `garden--` prefix.

> Does username map to "env"

Not quite sure what you mean by "env" here?

I think your confusion may be coming from the difference between the configuration for `local-kubernetes` and `kubernetes`. The `local-kubernetes` sets a default namespace, so that the line you highlighted doesn't come into play (the first branch of the if-else-statement is used instead). For remote `kubernetes`, I decided not to remove the existing default behaviour of using a username, because the expectation is that the cluster might be shared by many users.

garden-service/src/plugins/kubernetes/namespace.ts
```diff
@@ -89,9 +89,8 @@ export function getMetadataNamespace(ctx: PluginContext, provider: KubernetesPro
   return getNamespace({ ctx, provider, suffix: "metadata" })
```
drubin: Why don't we use the `suffix` param that is passed into the function?

edvald: We do. Line 72. Unless I'm misunderstanding the question?

edvald: Oh sorry, line number might be off (was looking at the file in a different branch)

garden-service/src/plugins/kubernetes/namespace.ts
drubin: I was under the impression `default` was always an environment not a "user" I don't think this is clearly enough explained. 

Also the above code is pretty confusing and needs a commenting highliting where these variables come from, (they get created differently in different providers). 

This abstraction/defaults aren't obvious

garden-service/src/plugins/kubernetes/namespace.ts
drubin: We check if the value exists but the param is a "string" I would assume if the value is a string it would use it as a string (ie `return getNamespace({ ctx, provider, suffix })`

garden-service/src/plugins/kubernetes/kubernetes.ts
drubin: @edvald  When I deploy this locally I get a namespace of `simple-project` I thought the intention was to still keep the `default-simple-project` there?

edvald: This is the desired outcome. The `local-kubernetes` provider sets the default namespace name to the project name. The prefix is used when there is no namespace name set in the provider config, which applies if you're using the `kubernetes` provider and don't specify a namespace name. I considered the prefix to be superfluous for a local, single-user cluster.

PR #465, opened 2019-01-16
WIP: fix(k8s): split command and args

drubin:
fix(k8s): split command and args so they match the semantics of kubernetes

By making this split we actuallly need to make very little code changes but it changes have knock on effect which I wanted to get feedback on. 

Currently this isn't the only place we use commands/args. We do it with `hotreloadCommand` and all task comamnds.

Currently although broken they behave in the same way. When doing this split we will need to provide a way for users to use custom args/commands in those as well. 

Because `args` actually maps more closely to docker's CMD most of our tasks could potentially look like this

```yaml
tasks:
    - name: ruby-migration
      args: [rake, db:migrate]
```

Since we assume the entrypoint would be `bundle exec` or some form of init handler.

@thsig @edvald Thoughts? I haven't actually gone and fully updated tasks and hotreloading wanted to get some throughts before going further.

BREAKING CHANGE:
After upgrading, any project which uses a custom `command`
will need to change their garden.yaml to use the `args` instead.

Example with a Dockerfile

```docker
ENTRYPOINT["/bin/echo ]
```

and garden.yml of

    command: ["hello", "worlld"]

would become

    args: ["hello", "world]
    command: ["/bin/echo] # this is line is optional

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -547,10 +547,29 @@ module:
       outputs: 
         {}
 
-      # The arguments to run the container with when starting the service.
+      # The command to run the container with when starting the service.
+      # Equivalent to
+      # Docker's entrypoint
+      # Kubernetes's command
+      #
+      # Example:
+      #   - /bin/echo
       #
       # Optional.
       command: 
+        -
+
+      # The arguments to pass to the command run when starting the service.
+      # Equivalent to
+      # Docker's cmd
+      # Kubernetes's args
+      #
+      # Example:
+      #   - hello
+      #   - world
+      #
+      # Optional.
+      args: 
         -
 
       # Whether to run the service as a daemon (to ensure only one runs per node).
--- a/garden-service/src/plugins/container.ts
+++ b/garden-service/src/plugins/container.ts
@@ -84,6 +84,7 @@ export interface ServiceHealthCheckSpec {
 
 export interface ContainerServiceSpec extends BaseServiceSpec {
   command: string[],
+  args: string[],
   daemon: boolean
   ingresses: ContainerIngressSpec[],
   env: PrimitiveMap,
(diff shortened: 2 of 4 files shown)


edvald:
Re: `hotReloadCommand`, I'd simply suggest renaming it to `hotReloadArgs`. We can later add the option of overriding the command, if we think it's necessary.

Did you consider ways to alert/warn the user that they might need to update their config?

drubin:
Other than versioning the configs and forcing users to manually opt in to v2 I don't think there is any way for us to figure out which version they wrote the file with.

Also I looked through(and ran) all our examples and unless the user is currently using a custom endpoint that is not a `init` command making use of our command as a pure args it will just continue to "work"..  (Which is arguably worse than breaking)

We could add a log entry to any config that contains `command` but how long do we leave it there and after when do we remove it? (that seems like just requiring a min version of the config files)

My issue with calling `hotreloadArgs` is if you look at all our examples they actually behave more like `cmd` 

The other option is we go back to docker style of `entrypoint` and then we can either just add custom entrypoints to tasks/hotreload, but we went against that because with k8s it seemed to make more sense.

> Re: hotReloadCommand, I'd simply suggest renaming it to hotReloadArgs. We can later add the option of overriding the command, if we think it's necessary.

I am not a huge fan of not doing this now at the same time, because its a little unclear which command the tasks would use, Would they use the default docker's entrypoint (would currently with our code) or would they make use of the overridden command configured above. (which is what I would expect)

edvald:
Hmm yes. I did suggest a while back to only do the rename from `command` to `args` in 0.9, and adding the new `command` field in 0.10. This would basically break any existing config referencing the field, forcing the user to address it as long as they don't happen to skip 0.9.x entirely. This would also allow us to rename `hotReloadCommand` -> `hotReloadArgs` in this PR without confusion.

The downside is of course that we delay the new feature of overriding entrypoints until 0.10. But I think I'd prefer that over the potential confusion and bug hunting resulting in changing behaviour on existing fields without visible warning.

edvald:
Closing in favour of #475

PR #467, opened 2019-01-17
feat(k8s): proper support for Helm charts

edvald:
This brings Helm support close to parity with the container module
type. The vote-helm example project demonstrates its use.

This required some framework changes and additions, in order to
allow modules that contain manifests to reference sources from other
modules.

BREAKING CHANGES:

We now install Helm's Tiller service in each app namespace, instead
of cluster-wide, and we no longer prefix Helm release names.

TODO:
- [x] Update existing tests
- [x] Add tests
- [x] Add guide for using Helm modules
- [x] Add testModule handler for helm module type
- [x] Add getServiceLogs handler for helm module type
- [x] Fix namespace issue with OpenFaaS plugin bootstrapping


edvald:
Beyond the two open TODO items above, this is ready for review. Happy weekend!

edvald:
The guide should be a helpful place to start the review: https://github.com/garden-io/garden/blob/47c5d671cc7320fdf4ca5c31487be9c099ec6ac7/docs/using-garden/using-helm-charts.md

garden-service/src/plugins/kubernetes/hot-reload.ts
```diff
@@ -8,72 +8,268 @@
...
+ * Adds an rsync sidecar container, an emptyDir volume to mount over module dir in app container,
+ * and an initContainer to perform the initial population of the emptyDir volume.
+ */
+export function configureHotReload({
+  target, hotReloadSpec, hotReloadArgs, containerName,
+}: ConfigureHotReloadParams) {
+  const kind = <HotReloadableKind>target.kind
+
+  // We use this to check whether the target is configured for hot-reloading.
+  set(target, ["metadata", "annotations", "garden.io/hotReload"], "true")
```
thsig: We can leave this here for now, but on the other hand, I see no reason not to just remove it.

garden-service/src/plugins/kubernetes/hot-reload.ts
```diff
@@ -8,72 +8,268 @@
...
+  // We're copying the target folder, not just its contents
+  const syncConfig = hotReloadSpec.sync
+  const syncTarget = syncConfig.map(pair => removeTrailingSlashes(pair.target))
+
+  const copyCommands = makeCopyCommands(syncConfig)
+
+  const initContainer = {
+    name: "garden-sync-init",
+    image: mainContainer.image,
+    command: ["/bin/sh", "-c", copyCommands],
```
thsig: Speaking of getting rid of the assumption of `/bin/sh` and `cp` being available in the container: Should we mention in the reference docs that `/bin/sh` and the `cp` command need to be available in the init container for hot reloading to work? Or should we have the init container use the app container's image as a base image, adding a layer to ensure that `/bin/sh` and `cp` are available?

garden-service/src/config/service.ts
```diff
@@ -34,13 +38,21 @@ export const baseServiceSchema = Joi.object()
...
   // Plugins can add custom fields that are kept here
   spec: T
 }
 
 export const serviceConfigSchema = baseServiceSchema
   .keys({
+    sourceModuleName: joiIdentifier()
+      .optional()
+      .description(deline`
+        The \`validate\` module action should specify this, if the service's code sources are contained in a
```
thsig: Just a thought: "specify" -> "populate"?

garden-service/src/plugins/kubernetes/helm/common.ts
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  return container
+}
+
+/**
+ * This is a dirty hack that allows us to resolve Helm template strings ad-hoc.
+ * Basically this writes a dummy file to the chart, has Helm resolve it, and then deletes the file.
+ */
+async function renderHelmTemplateString(
```
thsig: `renderHelmTemplateString` may be called (through `findServiceResource`) when hot-reloading a Helm service (i.e. if `resourceSpec.targetName` is a template string).

Could the `resourceSpec`'s `targetName` reasonably be a template in these cases? And if so, is it worth thinking about cacheing the rendered `targetName` after the first call to `renderHelmTemplateString`, since the file writing and Helm CLI call might add some time to the hot reload task?

If that's the case, and it's tricky to implement, then maybe we should log this as a perf improvement issue for looking into later. And/or we might mention in the docs, that we recommend using literal strings for hot-reloadable resource names when possible, for performance reasons.

garden-service/src/plugins/kubernetes/helm/run.ts
```diff
@@ -0,0 +1,45 @@
...
+  // find the relevant resource, and from that the container image to run
+  const chartResources = await getChartResources(ctx, module, log)
+  const resource = await findServiceResource({ ctx, log, module, chartResources, resourceSpec: task.spec.resource })
+  const container = getResourceContainer(resource)
+
+  const res = await runPod({
+    context,
+    namespace,
+    module,
+    envVars: { ...runtimeContext.envVars },
```
thsig: Replace with `runtimeContext.envVars`?

garden-service/src/tasks/deploy.ts
```diff
@@ -112,14 +112,25 @@ export class DeployTask extends BaseTask {
     })
 
     // TODO: get version from build task results
-    const { versionString } = this.version
-    const hotReloadEnabled = includes(this.hotReloadServiceNames, this.service.name)
+    let version = this.version
+    const hotReload = includes(this.hotReloadServiceNames, this.service.name)
+
+    // add a suffix to the deployed version when hot-reloading
```
thsig: This should probably be uncommented (and the tests re-run, just in case).

garden-service/test/src/actions.ts
```diff
@@ -413,6 +413,7 @@ const testPlugin: PluginFactory = async () => ({
       },
 
       getBuildStatus: async (params) => {
+        console.log(params.module.services)
```
thsig: Remove `console.log`.

garden-service/src/plugins/kubernetes/hot-reload.ts
edvald: Hmm yes, I think we should document it. Any thoughts on where?

garden-service/src/config/service.ts
edvald: 👍

garden-service/src/plugins/kubernetes/helm/common.ts
edvald: I've put in a TODO, but I don't think this is going to be a common issue. Also, the rendering is surprisingly snappy (tens of msec) so it's not a major concern.

garden-service/src/tasks/deploy.ts
edvald: Actually, I just needed to remove the commented code. Because we now always pass a `hotReload` parameter to the `getServiceStatus` handlers, we do the appropriate comparison there, and the returned `state` should be `"outdated"` if the toggle has changed between deployments.

edvald:
@eysi09 can you pull and check again?

thsig:
@eysi09: This is working for me now.

PR #470, opened 2019-01-22
chore(release): package using zeit/pkg for macOS

### LINKED ISSUE #453: Package Garden into an executable for OSX with zeit/pkg
Currently we package Garden into an executable binary for Windows and Linux. For OSX we still use the NPM package (by way of Homebrew). We originally left OSX out because of issues with native modules. This should be fixable by compiling on OSX in CI (perhaps @edvald can fill in the details).

As a benefit to this we can ditch the NPM package and simplify our release process.

edvald:
We were already doing this for Linux and Windows. This simply extends
that to macOS, and updates the homebrew package accordingly.

Closes #453


.circleci/config.yml
```diff
@@ -288,35 +288,6 @@ jobs:
           paths:
             # Save the built output for future
             - dashboard/build/
-  release-service-npm:
```
eysi09: ❤️

edvald:
Yes. Binary works fine, and I tested the `update-brew` best I could (short of pushing to the homebrew-garden repo).

PR #473, opened 2019-01-22
Provider docs

eysi09:
Generate reference docs for providers and module types automatically. We probably need to fix the text that's printed before the auto-generated content but I've been staring at this for too long so I decided to make a PR to get some feedback. We might also want to re-consider some of the content and it's structure.

I'll post a link to our docs site with this version once I've set it up.


eysi09:
The Gitbooks version is [here](https://docs.garden.io/v/provider-docs/reference/config). However, the Module Type and Provider docs only point to the Github "blob". I haven't figured out why, all the links are relevant and should work as usual.

edvald:
A few thoughts (not all rooted in this PR btw, but kind of exposed with the new view):
1) A lot of the keys could use examples. We should add those.
2) Would it make sense to split the `garden.yml` reference between module and project? At least it would be good to have a quick menu at the top, also to hop to the full YAML reference.
3) "Allowed values" is a little unclear. Maybe omit the column when it doesn't apply?
4) It would be good to add a type for arrays, e.g. `array[string]`

garden-service/src/docs/templates/base-config.hbs
```diff
@@ -0,0 +1,10 @@
+## garden.yml reference
+
+Below is the schema refernce for the `garden.yml` configuration files. For an introduction to configuring a Garden project,
```
edvald: Typo: "refernce"

eysi09:
All right, this should be ready for a proper review. I've added tests and resolved point 2) - 4) above.

docs/README.md
```diff
@@ -16,6 +16,8 @@
   * [TLS project](./examples/tls-project.md)
   * [Remote sources project](./examples/remote-sources.md)
 * [Reference](./reference/README.md)
+  * [Module Types](./reference/module-types/README.md)
```
edvald: Is it possible to nest the sidebar further?

docs/reference/module-types/README.md
```diff
@@ -0,0 +1,6 @@
+# Module Types
+
+* [Container](./container.md)
```
edvald: It would be helpful to have a paragraph describing each here. Fine to omit just now, but would be good to make an issue for it.

docs/reference/module-types/container.md
```diff
@@ -0,0 +1,518 @@
...
+
+POSIX-style path of the directory to sync to the target, relative to the module's top-level directory. Must be a relative path if provided. Defaults to the module's top-level directory if no value is provided.
+
+| Type | Required |
+| ---- | -------- |
+| `string` | No
+
+Example:
+```yaml
+hotReload:
```
edvald: I'd suggest leaving the proper indent here, so that this can be copy-pasted.

docs/reference/module-types/container.md
```diff
@@ -0,0 +1,518 @@
...
+```yaml
+tasks:
+  - args:
+    - rake
+    - 'db:migrate'
+```
+
+## Complete schema
+```yaml
+buildArgs:
```
edvald: Definitely good to add `module:` at the top here. +1 for omitting the comments here btw, looks nice and clear.

docs/reference/module-types/container.md
```diff
@@ -0,0 +1,518 @@
...
+    ingresses:
+      - hostname:
+        path: /
+        port:
+    env:
+      {}
+    healthCheck:
+      httpGet:
+        path:
+
```
edvald: Extra space here.

docs/reference/module-types/container.md
```diff
@@ -0,0 +1,518 @@
...
+    outputs:
+      {}
+    args:
+    daemon: false
+    ingresses:
+      - hostname:
+        path: /
+        port:
+    env:
+      {}
```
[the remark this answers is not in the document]
eysi09: Ah yes, it was like this previously and I figured that was by design.

docs/reference/module-types/README.md
eysi09: Yes, that would also match our other "Table of Contents" sections.

docs/reference/module-types/container.md
eysi09: Won't that look funny for really nested keys?

docs/reference/module-types/container.md
eysi09: Should we also add `provider[]` to the provider references? So the schema looks like:

```
{ name: "openfaas", schema: Joi.object().keys({ provider: joiArray(openfaasConfigSchema) }) }
```

docs/reference/module-types/container.md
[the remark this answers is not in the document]
eysi09: It's actually pretty hard to catch since this is just the default value for that key and the rule is to print non-primitive values in the line below. So we'd need to make an exception for non-primitive default values that just so happen to be `{}` which feels kind of messy. (Of course this kind of string manipulation is always going to be messy.)

docs/README.md
edvald: I think it would be nice to have the individual module types and providers in here (again, fine to leave for another PR). What do you think?

docs/reference/module-types/container.md
edvald: We don't have that many of those (thankfully). Plus you could use ellipses to make it a little more palatable, e.g.
```yaml
module:
  ...
    ...
      something: nested
```

docs/reference/module-types/container.md
edvald: That made more sense with the comments in there. I'd rather get rid of all the blank lines. (Could even do that as a post-process step for this particular case).

docs/reference/module-types/container.md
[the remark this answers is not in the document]
edvald: It's messy anyway, as you said. I'd actually say it's fine to special-case `[]` and `{}`.

docs/reference/module-types/container.md
edvald: Hmm, not sure what you mean.

docs/reference/module-types/container.md
eysi09: Actually this is not an issue for module types once we add the module type to the top:
```
module:
  hotReload:
    sync:
      - source: "src"
```

eysi09: Now the examples all parent keys with ellipsis between parent and child, like so:

```yaml
project:
  ...
  sources:
    - repositoryUrl: "git+https://github.com/org/repo.git#v2.0"
```

docs/reference/module-types/container.md
eysi09: I've removed the blank lines for the case when there are no comments.

docs/reference/module-types/README.md
eysi09: I'll add a separate issue for this.

docs/reference/module-types/container.md
eysi09: > Would it make sense to put in example values instead of defaults here, and defaults in inline comments?

Sometimes the default value and the example are the same value. Looks something like:

```yaml
project:
  name: "my-sweet-project"
  # Default value: ''
  defaultEnvironment: ''
  # Default value:
  #   providers: []
  #   variables: {}
  environmentDefaults:
    providers: []
    variables: {}
    # Default value:
    #   []
    providers:
      - name: "local-kubernetes"
    # Default value:
    #   {}
    variables: {}
  # Default value:
  #     name: local
  #     providers:
  #       - name: local-kubernetes
  #     variables: {}
  environments:
    - name: local
      providers:
        - name: local-kubernetes
      variables: {}
    # Default value: []
    - providers:
        - name: "local-kubernetes"
      # Default value: {}
      variables: {}
      name:
  # Default value: []
  sources:
    - name:
      repositoryUrl: "git+https://github.com/org/repo.git#v2.0"
```

My suggestion would be to leave this as is and address any other improvements in a separate issue.

edvald: Sounds good

docs/reference/providers/kubernetes.md
```diff
@@ -0,0 +1,346 @@
...
+          imagePullSecrets:
+            - name:
+              namespace: default
+          tlsCertificates:
+            - name:
+              hostnames:
+              secretRef:
+                name:
+                namespace: default
+          name:
```
edvald: Just noticed this, should have a fixed name in this case.

Issue #401, opened 2018-11-30
Don't use `/bin/sh -c` when executing commands


drubin:
### Problem 

Currently when you configure a custom command  garden does some magic and appends that to the entry point but also runs it in a shell. 

garden.yaml
```yaml
command: ["user", "--port=80"]
```

Dockerfile
```docker 
...
ENTRYPOINT user
...
```

Result

```sh
PID   USER     TIME   COMMAND
    1 root       0:00 /bin/sh -c user user --port=80
```

Proxying commands in a sh shell has big implications to how the commands behave. 

1. containers that run `FROM scratch` i.e most go containers simply wont work. 
2. We loose all ability to propagate the signals to the process as the user intended them.
3. We aren't PID1, PID1 is special and commands/entry points  that were chosen by the user should be honoured.

### Places we do this 

1. When passing custom `command: []`
2. Hot-reloading when we initially run the copy 
3. Test execution 
4. Possibly others

## Solution 
We need to differentiate entrypoint(docker) + command (or args as k8s calls it)

1. Support both entrypoint and commands / args
1. We should delegate this as much as possible to upstream providers and not do variable munging on them we should be passing both of these command + args directly upstream and be able to leverage their ability to parse the defaults in docker and do the merging.
1. Stop making assumptions. about shell's users should explicitly opt in to calling things via a shell.

edvald:
This is actually entirely handled by plugins, and it's only in the `kubernetes` plugin's `runModule` action that this is in place. So that in and of itself is easy to address.

The other issue is how we should introduce the separation between entrypoint and args/command. I do agree, that's something we'd want to address. My suggestion:

1. Rename `command` to either `args` or `arguments` (I can go either way personally) at the framework level, as well as in the `container` module spec. That feels like a more general term, that can apply across many providers.
2. Add an `entrypoint` or `command` (again, could go for either... might make most sense to stick to K8s semantics?) to the `container` module type and support in relevant providers.

Does that make sense? These are easy changes to make in code, but we should make sure we feel good about the layout.

We could also avoid a deprecation or breaking change by not doing 1) and just adding `entrypoint` to the container module type, but I think we're still okay to make breaking changes, just need to be careful and be sure we feel like the result will hold up longer-term.

edvald:
@drubin can I delegate the `/bin/sh -c` thing to you? We should do that separately from the entrypoint feature/refactor.

drubin:
[...]

PR #475, opened 2019-01-23
feat: rename command to args for container type

thsig:
BREAKING CHANGE:

After updating, the following configuration fields for container modules must be renamed as indicated:

  * `service.command` -> `service.args`
  * `service.hotReloadCommand` -> `service.hotReloadArgs`
  * `test.command` -> `test.args`
  * `task.command` -> `task.args`

This is done in preparation for a new configuration option (using the key `command`, hence the rename) that's planned for release soon, whereby users can override the current default of running
tests / tasks / ad-hoc commands inside containers via `/bin/sh -c [args]`.

See also: https://github.com/garden-io/garden/issues/401


docs/reference/config.md
```diff
@@ -482,9 +482,9 @@ module:
 
   # When this field is used, the files or directories specified within are automatically synced
   # into the running container when they're modified. Additionally, any of this module's services
-  # that define a `hotReloadCommand` will be run with that command instead of the one specified in
-  # their `command` field. Services are only deployed with hot reloading enabled when their names
-  # are passed to the `--hot-reload` option in a call to the `deploy` or `dev` command.
+  # that define `hotReloadArgs` will be run with args instead of those specified in their `args`
```
edvald: This sentence doesn't parse right.

docs/reference/config.md
```diff
@@ -693,7 +693,7 @@ module:
   #
   # Optional.
   tests: 
-    # The test specification of an exec module.
+    # Required configuration for module tests.
```
edvald: This feels less clear in terms of user-facing docs.

docs/reference/config.md
```diff
@@ -756,10 +760,14 @@ module:
       # Optional.
       timeout: null
 
-      # The command that the task should run inside the container.
+      # The shell arguments used to run the task inside the container.
```
edvald: Is "shell arguments" the right term here? These are generally arguments passed to the container's entrypoint, which may or may not be the shell.

docs/using-garden/hot-reload.md
```diff
@@ -46,4 +46,4 @@ You can configure several such `source`/`target` pairs, but note that the `sourc
         target: /app/bar
 ```
 
-Lastly, `hotReloadCommand` determines which command should be run inside the container (when deployed with hot reloading enabled). If no `hotReloadCommand` is specified, `command` is also used in hot reload mode.
+Lastly, `hotReloadArgs` specifies the arguments to use to run the container (when deployed with hot reloading enabled). If no `hotReloadArgs` is specified, `args` is also used in hot reload mode.
```
edvald: "... _are_ specified, ..."

examples/tasks/README.md
```diff
@@ -2,7 +2,7 @@
 
 This example uses dependency-aware database migrations to demonstrate Garden's _tasks_ functionality.
 
-Tasks are defined under `tasks` in a module's `garden.yml`. They're currently only supported for `container` modules, and consist of a `name`, a `command` and `dependencies`.
+Tasks are defined under `tasks` in a module's `garden.yml`. They're currently only supported for `container` modules, and consist of a `name`, `args` (to run the container) and `dependencies`.
```
edvald: This needs to be updates. Tasks are supported in a bunch of modules now.

examples/tasks/hello/garden.yml
```diff
@@ -15,12 +15,12 @@ module:
...
   tasks:
     - name: node-migration
-      command: [knex, migrate:latest]
+      args: [knex, migrate:latest]
       description: Creates the users table.
       dependencies:
         - postgres
     - name: foo-task
-      command: [echo hello]
+      args: [echo hello]
```
edvald: Missing comma - should be `[echo, hello]`

garden-service/src/plugins/google/google-app-engine.ts
```diff
@@ -37,6 +37,10 @@ export interface GoogleAppEngineServiceSpec extends ContainerServiceSpec {
 
 export interface GoogleAppEngineModule extends ContainerModule<ContainerModuleSpec, GoogleAppEngineServiceSpec> { }
 
+function getAppEngineProject<T extends GoogleAppEngineModule>(service: Service<T>, provider: Provider) {
```
edvald: What prompted this change? (Just curious, seems fine)

garden-service/src/plugins/kubernetes/container/run.ts
```diff
@@ -101,7 +101,7 @@ export async function runContainerService(
   return runContainerModule({
     ctx,
     module: service.module,
-    command: service.spec.command || [],
+    command: service.spec.args || [],
```
edvald: Might want to change the parameter name here as well.

docs/reference/config.md
edvald: Ah, now I see it was incorrect before (thought this was actually on the exec module). But still, phrasing seems a little odd, if only because the very next line reads "Optional". Maybe we should just omit this particular description field?

thsig: Yeah, let's just omit this. It's a bit too tautological and doesn't really need explaining.

docs/reference/config.md
thsig: How about _The arguments used to run the task inside the container_?

examples/tasks/README.md
thsig: How about just omitting _(to run the container)_ here?

docs/reference/config.md
```diff
@@ -481,10 +481,12 @@ module:
   image:
 
   # When this field is used, the files or directories specified within are automatically synced
-  # into the running container when they're modified. Additionally, any of this module's services
-  # that define a `hotReloadCommand` will be run with that command instead of the one specified in
-  # their `command` field. Services are only deployed with hot reloading enabled when their names
-  # are passed to the `--hot-reload` option in a call to the `deploy` or `dev` command.
+  # into the running containers of any of this module's services (and services with this module as
+  # their `sourceModule`) deployed with hot reloading enabled  when they're modified.
```
edvald: This is a very long sentence 😬

docs/reference/config.md
```diff
@@ -481,10 +481,12 @@ module:
...
+  # Additionally, any of this module's services that define `hotReloadArgs` will be run with those
```
edvald: Is this necessary/relevant to mention here?

Issue #458, opened 2019-01-11
Re-use connections to improve hot-reload performance


edvald:
Discussed in #422. Basically the issue is that after that change we create a port forward for every sync call when hot-reloading. We can reduce the latency considerably by managing and re-using the tunnels.

PR #476, opened 2019-01-24
fix(perf): reuse port-forwards when hot-reloading

thsig:
Added a simple in-process cache of port-forwards allocated for syncing
during hot-reloading.

This results in dramatic (often 50%) speedups for hot reloads subsequent
to the initial one, during the execution of the enclosing Garden command.

Addresses https://github.com/garden-io/garden/issues/458.

### DIFF
--- a/garden-service/src/plugins/kubernetes/hot-reload.ts
+++ b/garden-service/src/plugins/kubernetes/hot-reload.ts
@@ -7,6 +7,7 @@
  */
 
 import * as Bluebird from "bluebird"
+import { ChildProcess } from "child_process"
 import * as execa from "execa"
 import { V1Deployment, V1DaemonSet, V1StatefulSet, V1ObjectMeta } from "@kubernetes/client-node"
 import { HotReloadServiceParams } from "../../types/plugin/params"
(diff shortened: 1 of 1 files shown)


eysi09:
What does the speedup translate to in terms of milliseconds, for, say, the vote  example?

thsig:
I did a couple of runs in the **vote** project, with the `vote` service configured for hot reloading (stopping and then re-running `garden dev --hot-reload=vote` after the first run). Here are the numbers:

**First run**
* Initial hot reload: 335ms.
* Next five hot reloads: 201ms, 206ms, 145ms, 216ms, 228ms (average: 119ms).
* Average speedup compared to initial: 64%.

**Second run**
* Initial hot reload: 327ms.
* Next five hot reloads: 154ms, 142ms, 158ms, 229ms, 215ms (average: 179ms).
* Average speedup compared to initial: 45%.

I got similar distributions and speedups in the **gatsby-hot-reload** example project, but a slightly larger speedup on average.

garden-service/src/plugins/kubernetes/hot-reload.ts
```diff
@@ -272,3 +222,99 @@ export function rsyncTargetPath(path: string) {
...
+
+}
+
+async function getLocalRsyncPort(ctx: PluginContext, log: LogEntry, targetDeployment: string): Promise<number> {
+
+  let rsyncLocalPort
+
+  if (registeredPortForwards[targetDeployment]) {
+    rsyncLocalPort = registeredPortForwards[targetDeployment].rsyncLocalPort
+    log.debug(`Reusing local port ${rsyncLocalPort} for ${targetDeployment} sync container`)
```
edvald: I think it would be good to add a check here, to see if the process is still up.

thsig: Yeah, good point. Just added a check for that — was this what you had in mind?

edvald: Yep, I think that's good for now.

PR #484, opened 2019-01-25
fix: improved error messages when deps are missing

thsig:
Output a more informative error message when invalid/unknown
dependencies are present in a module's configuration.


garden-service/src/garden.ts
```diff
@@ -837,14 +837,18 @@ export class Garden {
       const moduleConfigContext = new ModuleConfigContext(
         this, this.log, this.environment, Object.values(this.moduleConfigs),
       )
-      this.moduleConfigs = await resolveTemplateStrings(this.moduleConfigs, moduleConfigContext)
 
-      await this.detectCircularDependencies()
+      this.validateDependencies()
```
thsig: I moved this above the `resolveTemplateStrings` call, since `resolveTemplateStrings` was throwing an (understandably uninformative) exception when dependencies were missing.

Is this OK?

garden-service/src/util/validate-dependencies.ts
```diff
@@ -7,23 +7,104 @@
...
-/*
-  Implements a variation on the Floyd-Warshall algorithm to compute minimal cycles.
+  const missingDepsError = detectMissingDependencies(moduleConfigs, serviceNames, taskNames)
+  const circularDepsError = detectCircularDependencies(moduleConfigs)
+
+  let errMsg = ""
+  let detail = {}
+
+  if (missingDepsError) {
+    errMsg = errMsg + missingDepsError.message
```
eysi09: `errMsg += missingDepsError.message` would be slightly terser.

garden-service/src/util/validate-dependencies.ts
```diff
@@ -7,23 +7,104 @@
...
+    detail = merge(detail, missingDepsError.detail)
+  }
+
+  if (circularDepsError) {
+    errMsg = errMsg + "\n" + circularDepsError.message
```
eysi09: Same as above with `+=`.

thsig:
In https://github.com/garden-io/garden/pull/484/commits/bb0fcf92ad4c7c951f111b9552ddbb79e4631af6, I changed the tests to more granularly test `detectMissingDependencies` and  `detectCircularDependencies` (instead of just expecting an exception during `scanModules`). Also addressed @eysi09's comments.

I didn't squash these changes right away, since it's probably easier to read through the changes addressing the PR comments this way.

If there are no other comments, I'd like to squash this into one commit before merging.

CC @edvald

garden-service/src/util/validate-dependencies.ts
```diff
@@ -111,10 +111,10 @@ export function detectCircularDependencies(moduleConfigs: ModuleConfig[]): Confi
...
   const tasks: TaskConfig[] = []
 
-  /*
-    There's no need to account for test dependencies here, since any circularities there
-    are accounted for via service dependencies.
-    */
+  /**
+   * There's no need to account for test dependencies here, since any circularities there
+   * are accounted for via service dependencies.
+   */
```
edvald: Hmm, can you explain that?

garden-service/test/src/validate-dependencies.ts
```diff
@@ -0,0 +1,96 @@
+import { expect } from "chai"
```
edvald: This file should be under `test/src/util`, right?

thsig: Ah yes, true that. Will move it.

garden-service/src/util/validate-dependencies.ts
thsig: In short, because builds/deployments/tasks/tests can't currently depend on tests, so dependencies listed in test configs can't — by themselves — introduce circularities.

I'll add this explanation to the comment.

PR #488, opened 2019-01-28
fix(k8s): configure RBAC properly for Tiller

edvald:


edvald:
@ellenkorbes I think this might fix your issue. Can you check when you have a moment?

PR #490, opened 2019-01-29
fix(core): certain template strings could not be resolved in configs

edvald:
This addresses a design issue, where runtime values were being computed
to interpolate module configurations. Because service names and outputs
cannot really be resolved until after all module configs are resolved,
we needed to remove the `services` template string context, as well
as fix the module resolution flow.

BREAKING CHANGE:

Module configurations using the `services` template key need to be
updated to use `modules` instead.

The (admittedly poorly supported) google-cloud-function module type has
been changed to include only one function per module. This is more
consistent with other module types, and avoids complex refactoring
to fit with the changes in the templating context.


garden-service/src/plugins/google/google-app-engine.ts
```diff
@@ -20,34 +17,42 @@ import {
...
     prepareEnvironment,
   },
   moduleActions: {
     container: {
+      async configure(params: ConfigureModuleParams<ContainerModule>) {
+        const config = await configureContainerModule(params)
+
+        // TODO: we may want to pull this from the service status instead, along with other outputs
+        const project = params.ctx.provider.config.project
+        const endpoint = `https://${GOOGLE_CLOUD_DEFAULT_REGION}-${project}.cloudfunctions.net/${config.name}`
```
thsig: Should the region here be override-able via config?

edvald:
Added another commit, which I think takes this thing home:

**fix(core): using module version in templates didn't work with watch**

This is a pretty major refactor, where resolving raw configs is
deferred to usage time, and the handling of resolved configuration is
moved to the ConfigGraph class (née DependencyGraph). Basically the
Garden class now holds the raw configuration, and it is fully resolved
when building a new instance of ConfigGraph, which should then be done
whenever module sources are modified (otherwise resolved template
strings, e.g. version tags, may be stale).

Also moved some handling of actions to the ActionHandler to further
reduce the size of the Garden class.

garden-service/src/garden.ts
```diff
@@ -146,16 +108,13 @@ const scanLock = new AsyncLock()
...
 export class Garden {
   public readonly log: LogEntry
-  public readonly actionHandlers: PluginActionMap
-  public readonly moduleActionHandlers: ModuleActionMap
-  public dependencyGraph: DependencyGraph
+  public dependencyGraph: ConfigGraph
 
   private readonly loadedPlugins: { [key: string]: GardenPlugin }
   private moduleConfigs: ModuleConfigMap
+  private pluginModules: ModuleConfig[]
```
thsig: Suggest renaming this to `pluginModuleConfigs`.

garden-service/src/garden.ts
```diff
@@ -516,123 +472,55 @@ export class Garden {
...
-    const serviceDeps = await dg.getDependenciesForMany("service", serviceNames, true)
-    const taskDeps = await dg.getDependenciesForMany("task", taskNames, true)
-
-    const modules = [
-      ...(await this.getModules(moduleNames)),
-      ...(await dg.modulesForRelations(await dg.mergeRelations(buildDeps, serviceDeps, taskDeps))),
-    ]
-
-    return sortBy(uniqByName(modules), "name")
+  async getConfigGraph() {
```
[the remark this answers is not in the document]
edvald: I added a note in the function doc to explain, but maybe `resolveConfigGraph()` would make sense?

Issue #377, opened 2018-11-16
getContainerServiceStatus returns state === ready even though pod is unhealthy


eysi09:
To reproduce:

1. cd into `simple-project`
2. open `services/node-service/package.json`
3. change line 7 to `"start": "node pain.js", (this breaks the start command)`
4. run `garden deploy node-service`
5. notice that the deployment is successful, also that `garden get status` returns `state: ready` for the `node-service`, even though `kubectl get pods --namespace=garden--default--simple-project` shows that the status for the pod is `CrashLoopBackOff`

PR #491, opened 2019-01-29
fix(status): return more correct/granular statuses

thsig:
Fixes https://github.com/garden-io/garden/issues/377.

Previously, deployed but unhealthy services would be reported as "ready" by the `get status` command. This is fixed here.


eysi09:
Looks good:

```
  hello-container:
    ingresses:
      - path: /hello
        port: 80
        hostname: hello-world.local.app.garden
        protocol: http
    state: unhealthy
```

PR #487, opened 2019-01-28
chore(release): bump version to 0.9.0-rc1

eysi09:

PR #495, opened 2019-01-31
chore(release): add a release script

eysi09:


eysi09:
We should wait for #487 before merging.

Issue #312, opened 2018-10-01
Garden create command


ellenkorbes:
Currently the types available for modules via `garden create` are hard-coded. Ideally they should be dynamically generated so that we always have all possible values present there without having to add them one by one.

cc @eysi09

eysi09:
Thanks for catching this one.

For this to be truly dynamic, we need to implement the following:

1. The command imports and parses all the builtin plugins to get a map between all available module types and all providers for that particular type.
2. The command prompts the user for a module type based on the map from  **step 1**.
3. The command prompts the user for a provider based on the map from **step 1** and the selected module type.
4. Plugins implement something like `getSampleConfig` per module type so that we can populate the module config.
5. As an added bonus, plugins have help text to go with the selection in **step 3**. 

The above wasn't possible when we first added the feature due to the architecture of the plugin system but I don't think there's anything in the way now. As long as it aligns with our future plans for the plugin system. 

cc @edvald

eysi09:
Actually, the `create module` command should just call a `createModule` (or similar) handler on the relevant plugins. The plugins themselves are responsible for creating the module and the scope could expand beyond just generating the `garden.yml` file.

In the meantime we should remove the command since it's not very useful at the moment, see #492.

eysi09:
This was closed with #1582

PR #497, opened 2019-01-31
refactor(commands): remove create commands

### LINKED ISSUE #492: Remove Garden create commands (for now)
The `create project|module` commands are more confusing than they are useful at the moment. We should remove them for now. There's a separate issue for re-implementing them, #312.

eysi09:
BREAKING CHANGE:

After this, the `create project` and `create module` commands will no
longer be available. We're removing them for now because currently
they're more confusing than they are useful. There's an open Github
issue (#312) for properly implementing the "create" functionality.

Closes #492.


eysi09:
@edvald I dismissed your review since I noticed that the docs mention the `create` commands. I've updated them accordingly.

PR #498, opened 2019-01-31
Fix serve command logs

eysi09:
Also renames `preserveLevel` to `childEntriesInheritLevel`. Although it really should just be `bequeathLevel`.


eysi09:
`bestowLevelUponChildren`?

Issue #468, opened 2019-01-21
Remove garden-cli and garden-sync from build process


eysi09:
Currently we have several unnecessary build steps, both local and in CI:

* build garden-cli go binary
* build and release garden-sync docker image
* build and release garden-service docker image

These are remnants of our experiment with running the Garden service in a container with a Go CLI in front of it. 

We should remove these steps as they cause unnecessary complexity. Furthermore we should add a readme to the `garden-cli` and `garden-sync` repos explaining their purpose.

eysi09:
We should also use to chance to either close or merge #364.

eysi09:
Closed in #515.

PR #515, opened 2019-02-05
refactor: remove experimental Go CLI

eysi09:
The experimental Go CLI, previously in the `garden-cli` and `garden-sync`
directories, has been moved to a separate repo: https://github.com/garden-io/experimental-go-cli.

Moving the Go CLI simplifies the build process. It also simplifies
development since users no longer need to have Go installed and the
source code in their `$GOPATH`.

Furthermore, the Go CLI works just fine as a standalone module since the docker
image it uses is available on DockerHub (https://hub.docker.com/r/gardenengine/garden-service)
and is updated on every merge to master. Therefore there's no real reason to
co-locate it with the rest of the Garden source code.


eysi09:
The [experimental-go-cli repo](https://github.com/garden-io/experimental-go-cli) is being set up at time of writing.

eysi09:
Closes #468

PR #516, opened 2019-02-05
fix(core): error in actions.getStatus helper

edvald:
Occurred only when services depended on tasks, so it was missed
in testing.


eysi09:
Looks good, but I'm not quite sure what it fixes.

edvald:
Yeah sorry, I did this kind of on the fly so I skimped on the description. The issue was using `getServices` to retrieve dependencies when some of the dependencies were tasks and not services.

Issue #512, opened 2019-02-05
garden dev process crashes when reloading invalid config


edvald:
### Current Behavior

When I added a module or project config while running `garden dev` or other commands with `--watch`, and the changes I make don't result in a valid configuration, the process crashes:

https://www.dropbox.com/s/ck5x8kefgtgfshv/Screenshot%202019-02-04%2017.15.49.png?dl=0

### Expected behavior

I'd expect to be shown the validation error, but for the process to keep running.

### Reproducible example

1. In the `simple-project` example project, run `garden dev`
2. Add some invalid key, e.g. `foo: bar`, to the project `garden.yml`.

### Suggested solution(s)

This needs a bit of thought in terms of how/where to catch these validation errors, and how to resume execution when the errors are fixed. Not 100% sure what the most elegant approach is, but I'd say this is important to fix quickly.

### Your environment

Running 0.9.0-rc2

PR #530, opened 2019-02-11
feat: don't restart command when config is invalid

thsig:
Fixes https://github.com/garden-io/garden/issues/512.

Before restarting a watch-mode command when a config file is changed/added/removed, attempt creating a new Garden instance using the project's (and its modules') updated config before performing the restart.

If a configuration error was introduced by the change, log an error message and keep the existing Garden instance instead of restarting.

### DIFF
--- a/garden-service/src/process.ts
+++ b/garden-service/src/process.ts
@@ -19,6 +19,8 @@ import { Garden } from "./garden"
 import { LogEntry } from "./logger/log-entry"
 import { startServer } from "./server/server"
 import { ConfigGraph } from "./config-graph"
+import { dedent } from "./util/string"
+import { ConfigurationError } from "./exceptions"
 
 export type ProcessHandler = (graph: ConfigGraph, module: Module) => Promise<BaseTask[]>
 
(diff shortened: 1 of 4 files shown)


garden-service/src/process.ts
```diff
@@ -164,3 +179,32 @@ export async function processModules(
...
+  garden: Garden, log: LogEntry, changedPath: string, operationType: "added" | "changed" | "removed",
+): Promise<boolean> {
+
+  try {
+    const nextGarden = await Garden.factory(garden.projectRoot, garden.opts)
+    await nextGarden.getConfigGraph()
+  } catch (error) {
+    if (error instanceof ConfigurationError) {
+      const msg = deline`
+        Encountered configuration error after ${changedPath} was ${operationType}. Keeping existing configuration
```
eysi09: I'd suggest we log the ConfigurationError to the terminal as well. In particular since this is in watch mode so the user would have to kill the process and open `error.log` to see what they did wrong.

eysi09: ...just that message that is, not the details.

PR #533, opened 2019-02-11
Fix server routes

eysi09:


eysi09:
We need this for the `0.9.1` release.

PR #535, opened 2019-02-13
feat: allow multiple modules in a single file

thsig:
More than one module can now be defined in the same garden.yml file.

This is useful e.g. where more than one Dockerfile is used to build the same container (e.g. for development and production).

Within a garden.yml file, each module definition lives within a YAML document (separated by `---`).


garden-service/src/cli/cli.ts
```diff
@@ -70,10 +70,10 @@ const DEFAULT_CLI_LOGGER_TYPE = LoggerType.fancy
 
 // For initializing garden without a project config
 export const MOCK_CONFIG: GardenConfig = {
-  version: "0",
   dirname: "/",
   path: process.cwd(),
   project: {
+    apiVersion: "0",
```
edvald: Maybe we should make the default value, even though it's not used, more like K8s versions. Something like `garden.io/v0`.

docs/using-garden/configuration-files.md
```diff
@@ -192,6 +192,48 @@ in your tests.
...
+  ...
+  tests:
+    - name: unit
+      args: [npm, test]
+    - name: integ
+      args: [npm, run, integ-prod]
+      dependencies:
+        - hello-function
+        - prod-integration-testing-backend
+```
```
edvald: Looking at this example, it does feel to me that it would feel more natural using K8s-style definitions, like this:
```yaml
kind: Module
description: Hello world container - dev configuration
type: container
dockerfile: Dockerfile-dev
...
tests:
- name: unit
  args: [npm, test]
- name: integ
  args: [npm, run, integ-dev]
  dependencies:
    - hello-function
    - dev-integration-testing-backend
---
kind: Module
description: Hello world container - production configuration
type: container
dockerfile: Dockerfile-prod
...
tests:
- name: unit
  args: [npm, test]
- name: integ
  args: [npm, run, integ-prod]
  dependencies:
    - hello-function
    - prod-integration-testing-backend
```

We'd keep compatibility by having the top-level `module` and `project` keys as alternate forms, but move to a more standard style of representation (which notably wasn't really so commonplace when we first started on Garden, before CRDs took off), and as a bonus we drop an indent level. What do you think?

thsig: Yeah, good idea. Will make the corresponding changes.

edvald: Following on from that, we might actually consider changing the documented/recommended configuration style. We'd make a separate issue for it though, since we'd want to merge those changes after 0.10 is out.

eysi09: Would this also apply to the project config?

edvald: I'd say so, yes.

thsig:
I've added support for the flat config style in the second commit: https://github.com/garden-io/garden/commit/9151167591c9b1c20fb25020a12378377b8d5636

PR #541, opened 2019-02-15
fix: clean up field usage in example projects

thsig:
Cleaned up the usage of command / hotReloadCommand / hotReloadArgs in
example projects.


eysi09:
Looks good, although I'm not sure how to best handle the `module.services[].args` key. Should we always leave it out if it's not needed? I think we should figure out what the best practice is and document this specifically. 

@edvald, any thoughts?

edvald:
I'd generally leave it out if it's not needed, personally. Keep the config concise where we can.

PR #545, opened 2019-02-18
fix(templates): add prefix to versionString

thsig:
Added the "version-" prefix to the versionString field on ModuleVersion.

This should resolve an issue where module versions would be read as numbers when used as template variables when the first several characters of the commit hash were numbers.

@eysi09 and I have come across this error a couple of times during CI. Here's an example:

screenshot 2019-02-18 at 14 06 22

### DIFF
--- a/garden-service/src/config/config-context.ts
+++ b/garden-service/src/config/config-context.ts
@@ -188,7 +188,7 @@ class EnvironmentContext extends ConfigContext {
 }
 
 const exampleOutputs = { endpoint: "http://my-service/path/to/endpoint" }
-const exampleVersion = "v17ad4cb3fd"
+const exampleVersion = "v-v17ad4cb3fd"
 
 class ModuleContext extends ConfigContext {
   @schema(
--- a/garden-service/src/vcs/base.ts
+++ b/garden-service/src/vcs/base.ts
@@ -43,6 +43,7 @@ interface NamedTreeVersion extends TreeVersion {
 }
 
 const versionStringSchema = Joi.string()
+  .regex(/^v/)
   .required()
   .description("String representation of the module version.")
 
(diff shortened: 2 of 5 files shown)


garden-service/src/vcs/base.ts
```diff
@@ -221,8 +222,18 @@ export async function writeModuleVersionFile(path: string, version: ModuleVersio
   await writeFile(path, JSON.stringify(version))
 }
 
+/**
```
eysi09: Shouldn't the comment apply to the `addVersionPrefix` function? We sometimes create the `versionString` with the `version-` prefix without calling the `getVersionString` function.

thsig:
@edvald: Before I merge — do you have any comments/thoughts on this?

edvald:
My main issue is that "version-" feels needlessly verbose. Why not just "v"?

Also, we should log a separate issue for that issue described in the screenshot, which could come up when rendering other template strings.

thsig:
> My main issue is that "version-" feels needlessly verbose. Why not just "v"?

Yeah, good point. Wanted it to be self-evident why it was being added, but on second thought "v" would be better.

> Also, we should log a separate issue for that issue described in the screenshot, which could come up when rendering other template strings.

True, I'll write that up. This issue could come up where-ever the interpolated value is (or could sometimes be) a number.

garden-service/src/config/config-context.ts
```diff
@@ -188,7 +188,7 @@ class EnvironmentContext extends ConfigContext {
 }
 
 const exampleOutputs = { endpoint: "http://my-service/path/to/endpoint" }
-const exampleVersion = "v17ad4cb3fd"
+const exampleVersion = "v-v17ad4cb3fd"
```
edvald: We should remove the "v" prefixing in the `hashVersions` function, which is now superfluous. Otherwise, looks good!

PR #553, opened 2019-02-19
Fix minikube issues

edvald:
This addresses a few issues that have affected minikube users, but (for some reason, still investigating) not Docker for Desktop users.


garden-service/src/garden.ts
```diff
@@ -441,15 +441,16 @@ export class Garden {
...
-    if (configureHandler) {
-      const configureOutput = await configureHandler({ config: providerConfig })
-      providerConfig = configureOutput.config
-    }
 
     if (plugin.configSchema) {
       providerConfig = validate(providerConfig, plugin.configSchema, { context: `${pluginName} configuration` })
     }
 
+    if (configureHandler) {
```
eysi09: Not sure I understand the context here, why did you change the order?

eysi09:
Should we also address the hardcoded `ingressClass` value:
https://github.com/garden-io/garden/blob/be5448381ed1a9c9e1b8808a3279204c8e54ecb2/garden-service/src/plugins/kubernetes/local.ts#L143
in this commit?

garden-service/src/garden.ts
edvald: Oh right, I should have explained. Thing is, the configSchema may specify default values, which are lost if the clauses are swapped back.

edvald:
Yes, good catch. I'll add that.

edvald:
@eysi09 added a commit to fix the hardcoded ingress class. Could you test this on your setup?

eysi09:
Will do.

eysi09:
Yes, toggling the `setupIngressController` now works.

PR #554, opened 2019-02-20
fix: fixed bug in configuring flat-config modules

thsig:
Also simplified vote-helm project to put image and chart module configs in the same file, for those Helm modules that don't have a `Chart.yml` or `values.yml`.

This also serves as a simple demonstration of multiple modules being configured in one file.


eysi09:
I'm probably missing something but why does it matter if the modules have a `Chart.yaml` and/or `values.yaml` files?

edvald:
It actually doesn't matter technically. Any particular reason for not doing this with the vote service as well @thsig? Also, what exactly was the bug? Seems unclear from the changes.

thsig:
> I'm probably missing something but why does it matter if the modules have a `Chart.yaml` and/or `values.yaml` files?

No fundamental reason, my knee-jerk reaction was that the user might want to keep the Helm config separate from the underlying container. Should I merge the rest of these module directories too?

> Any particular reason for not doing this with the vote service as well @thsig?

I.e. for merging the `redis` and `postgres` services?

> Also, what exactly was the bug? Seems unclear from the changes.

In certain circumstances (IIRC, for flat-format Helm module configs), the module's path wasn't being set during configuration. Not sure how that slipped past the unit tests, but it was a simple fix.

edvald:
> No fundamental reason, my knee-jerk reaction was that the user might want to keep the Helm config separate from the underlying container. Should I merge the rest of these module directories too?

Got it. Yeah I guess it makes sense to separate the modules when you have the chart sources in there too.

> In certain circumstances (IIRC, for flat-format Helm module configs), the module's path wasn't being set during configuration. Not sure how that slipped past the unit tests, but it was a simple fix.

👍

edvald:
Should we hold off on merging until after release? The examples won't work pre 0.9.2.

thsig:
Replaced by https://github.com/garden-io/garden/pull/556.

Issue #540, opened 2019-02-15
taskPending event emitted when it shouldn't


eysi09:
### Current Behavior

When changing the `services/api/app.py` file in the `vote` example, a `taskPending` event for the `vote` unit test is fired. 

### Expected behavior

Event shouldn't fire.

### Reproducible example

1. Run `garden dev` in the `vote` example.
2. Open the dashboard and browse to the graph page.
3. Change `examples/vote/services/app.py` and save.
4. Notice that the `test.vote.unit` task starts processing. It's also possible to inspect the console logs in the browser or add a log statement here: https://github.com/garden-io/garden/blob/master/garden-service/src/task-graph.ts#L99. 

Furthermore, the `taskComplete` event never fires.

### Additional context

Here's the relevant config for the `vote` service:

```
  tests:
    - name: unit
      args: [npm, run, test:unit]
    - name: integ
      args: [npm, run, test:integ]
      dependencies:
        - api
```

Notice that the `integ` test should run but the `unit` test should not. 

### Your environment

`garden version`

Latest `master`. 

`kubectl version`

N/A

`docker version`

N/A

PR #555, opened 2019-02-21
fix: don't emit taskPending if task is skipped

thsig:
Fixes https://github.com/garden-io/garden/issues/540.

This fixes an issue in `TaskGraph` where `taskPending` events were emitted for tasks with up-to-date cached results. This led to the task graph UI on the dashboard displaying the nodes corresponding to skipped tasks in a permanent pending state.

This was addressed by only emitting the event if the task was, in fact, added to `TaskGraph`'s index.


thsig:
Added a test case for this just now.

PR #569, opened 2019-02-26
docs: add a "custom ingress controller" guide and example

eysi09:


eysi09:
@thsig @edvald Could you double check that the example project works on your end?

eysi09:
As of #571, Garden supports service annotations those guide doesn't really make sense.

We should however write a guide on how to use the Helm base charts for arbitrary K8s objects and CRDs.

PR #571, opened 2019-02-27
feat(k8s): support service and ingress annotations on container module

edvald:


garden-service/src/plugins/container/config.ts
```diff
@@ -59,7 +61,12 @@ export interface ServiceHealthCheckSpec {
   tcpPort?: string,
 }
 
+interface Annotations {
+  [name: string]: string
```
eysi09: How do we deal with non-string values? Something like:

```yaml
kind: Service
metadata:
  name: httpbin
  annotations:
    getambassador.io/config: |
      ---
      apiVersion: ambassador/v1
      kind:  Mapping
      name:  httpbin_mapping
      prefix: /httpbin/
      service: httpbin.org:80
      host_rewrite: httpbin.org
spec:
...
```

edvald: Kubernetes doesn't allow non-string values on annotations. Your example is actually a string, so that would work fine.

PR #573, opened 2019-02-27
fix(k8s): don't abort deployment on FailedMount warning

edvald:


ellenkorbes:
I can't reproduce the errors I got the other day anymore, I think this is good to merge.

PR #576, opened 2019-02-27
fix(vcs): untracked files didn't update version timestamp correctly

### LINKED ISSUE #565: Dirty timestamp not always correct
## Bug

The `getTreeVersion` function in `vcs/git.ts` doesn't properly join file path and file name when looking at timestamps on dirty files.

The `join` call below:

https://github.com/garden-io/garden/blob/9628da7ff4f22bfe6ee15a7cf42dfc6c5dd2bc84/garden-service/src/vcs/git.ts#L91

only joins the repo root with the file name which can be a non-existing path if the file itself is not at the repo root.

As a result, the `stats` array either includes the wrong file stats or none at all if the path doesn't exist because of the following `filter` function:

https://github.com/garden-io/garden/blob/9628da7ff4f22bfe6ee15a7cf42dfc6c5dd2bc84/garden-service/src/vcs/git.ts#L92  

### Current Behavior

See above.

### Expected behavior

Timestamp should update.

### Reproducible example

1. `cd examples/simple-project/`
2. `touch services/node-service/foo.txt`
3. Add a console log below this line

https://github.com/garden-io/garden/blob/9628da7ff4f22bfe6ee15a7cf42dfc6c5dd2bc84/garden-service/src/vcs/git.ts#L91

like so: `console.log(await Bluebird.map(dirtyFiles, file => join(repoRoot, file)))`

4. Notice that the path to `foo.txt` is wrong, namely it is: `'/Users/eysi/code/garden-io/garden/foo.txt'`
5. (Notice also that the `repoRoot` is the Garden root, not the example root.)

### Suggested solution(s)

Make sure we do the join correctly. Also, should that filter be there? Shouldn't the path always exist.

We should also test this properly, there's a lot of codepaths to the version string and this logic has been buggy before.

### Additional context

N/A

### Your environment

Latest `master`.

edvald:
Closes #565

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -185,6 +185,11 @@ jobs:
       - npm_install
       - *attach-workspace
       - run:
+          name: Configure git (needed for some tests)
+          command: |
+            git config --global user.name "Garden CI"
+            git config --global user.email "admin@garden.io"
+      - run:
           name: Make sure generated docs are up-to-date
           command: npm run check-docs
       - run:
--- a/garden-service/src/vcs/base.ts
+++ b/garden-service/src/vcs/base.ts
@@ -13,7 +13,7 @@ import * as Joi from "joi"
 import { validate } from "../config/common"
 import { join } from "path"
 import { GARDEN_VERSIONFILE_NAME } from "../constants"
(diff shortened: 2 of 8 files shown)


eysi09:
Code LGTM.

PR #579, opened 2019-02-28
Dev workflow pt2

eysi09:


bin/check-docs
```diff
@@ -1,8 +1,18 @@
...
-find docs examples -name '*.md' -type f | xargs -n 1 ./node_modules/.bin/markdown-link-check --config markdown-link-check-config.json
+
+# Use "|| true" so we don't exit on empty
+modified_docs=$(git diff --name-status master docs README.md) || true
+modified_examples=$(git diff --name-status master examples | grep "examples.*\.md$") || true
+
+# Only check links if docs or examples were modified
+if !([ -z "$modified_docs" ] && [ -z "$modified_examples" ]); then
+  # Using xargs to exit 1 if link is broken, see: https://github.com/tcort/markdown-link-check/issues/57
+  find docs examples -name '*.md' -type f -not -path "*/.garden/*" | xargs -n 1 ./node_modules/.bin/markdown-link-check --config markdown-link-check-config.json
```
edvald: I think we should rather just limit to the README.md files in the examples directories. This doesn't account for cases where `node_modules` are installed outside of `.garden` directories etc.

eysi09: Good point.

eysi09: I've fixed this.

PR #581, opened 2019-02-28
docs(examples): add Ambassador example

eysi09:


examples/ambassador/README.md
```diff
@@ -0,0 +1,65 @@
+# Ambassador example project
+
+This example project demonstrates how to use the [Ambassador API Gateway](https://www.getambassador.io/) instead of the default Nginx ingress controller. Ambassador is an open source, Kubernetes-Native, API Gateway built on the [Envoy Proxy](https://www.envoyproxy.io/). Services are configured via [annotations](https://docs.garden.io/reference/module-types/container#module-services-annotations) which Ambassador reads to configure its Envoy Proxy.
+
+Even though we chose Ambassador for this project, the same principles apply to e.g. [Traefik](https://traefik.io/), which also supports configuring route mappings via service annotations.
+
+The project is based on our [simple-project example](https://github.com/garden-io/garden/tree/v0.9.0-docfix.2/examples/simple-project) and installs Ambassador via the [Helm module type](ihttps://docs.garden.io/reference/module-types/helm). To learn more about using Helm charts with Garden, take a look at our [Helm user guide](https://docs.garden.io/using-garden/using-helm-charts).
```
edvald: Broken link to the Helm module type here

examples/ambassador/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Usage
+
+This project doesn't require any specific set up and can be deployed in a single step with  the `deploy` command:
```
edvald: Extra space ("with  the")

examples/ambassador/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+```sh
+garden deploy
+```
+
+Since we're letting Ambassador handle our ingresses, we don't define any in our `garden.yml` config files. Therefore, the `call` command won't work with this set up. Instead, we can use `curl` to check on our services.
```
edvald: Would be great to add an issue for this. It would be good to be able to extract ingress info from annotations, CRDs etc. for a handful of popular ingress controllers.

edvald: Side note: I believe this issue would not come up for Traefik.

examples/ambassador/garden.yml
```diff
@@ -0,0 +1,6 @@
+project:
+  name: ambassador
+  environments:
+    - name: local
+      providers:
+        - name: local-kubernetes
```
edvald: I'd like to see the `setupIngressController: false` line here, even though they don't conflict due to the different port.

examples/ambassador/README.md
```diff
@@ -0,0 +1,65 @@
...
+curl localhost:8080/node-service/hello-node
+```
+
+which should return
+
+```sh
+Hello from Node service!
+```
+
+Note that we've configured the Ambassador service to listen on port `8080` since the default Nginx ingress controller already occupies the `80`, the default port. Here's the relevant configuration from `ambassador/garden.yml`:
```
edvald: Typo: "the 80".

Also, we might add a note that we're doing this to avoid conflicts with _other examples_, since some of them install another ingress controller by default. Just to make it clear that it's not a technical issue outside of that context.

examples/ambassador/README.md
eysi09: We could off course add the ingresses to the `garden.yml` file. In that case the call command would work as usual. Should I rather do that?

eysi09:
I added another section about configuration. Not sure if we think that's to verbose for and example readme. If that's the case I can just revert back to what we had before plus the changes you requested.

examples/ambassador/README.md
edvald: I suppose... But that's a little dirty, duplicating the definitions. I'd say we rather just log an issue, since we want that feature anyway, and it's not very hard to do.

PR #589, opened 2019-03-04
fix(examples): use different ingress hostnames for vote-helm example

edvald:
This avoids confusion/conflict when trying out both the vote and
vote-helm examples.


eysi09:
LGTM. Did you check that the example works with the changes?

edvald:
Yep, tested and works fine on my side.

PR #592, opened 2019-03-05
chore: added integ tests & tools

thsig:
Wrote helper logic for integration testing Garden. Also wrote up a simple pre-release test suite using those tools.

The integration test suite can be run with `npm run integ-full`. This script supports the `--binPath` option for overriding the Garden binary used (which defaults to `garden-service/static/bin/garden`), and the `--showLog` option to show full log output for watch-mode commands (useful for debugging integ tests).

The old `integ` test script is still there, since the full suite is a bit too heavy to run on every push.

I'd like to merge this into `master` soon, since moving the unit tests into `test/unit` makes this branch a bit of a hassle to rebase when new unit tests are added, so maybe we could address any non-critical comments in a follow-up PR (unless they're quick fixes).


garden-service/src/cli/cli.ts
```diff
@@ -103,13 +111,19 @@ export const GLOBAL_OPTIONS = {
...
     hints:
       "[enum] [default: info] [error || 0, warn || 1, info || 2, verbose || 3, debug || 4, silly || 5]",
     defaultValue: LogLevel[LogLevel.info],
   }),
+  jsonLog: new BooleanParameter({
+    help: deline`
+      Format log entries as JSON. Unlike the default format, log entries are not updated in-place when this
+      format is used.`,
+    defaultValue: false,
+  }),
```
edvald: This means we have separate ways to enable JSON logging, and to set the logger type (via the `GARDEN_LOGGER_TYPE` env var. Doesn't have to be in this PR, but maybe @eysi09 we could make an issue and settle on a canonical way to do this?

garden-service/src/cli/cli.ts
```diff
@@ -127,11 +141,12 @@ function initLogger({ level, logEnabled, loggerType, emoji }: {
...
-      writers.push(new FancyTerminalWriter())
-    } else if (loggerType === LoggerType.basic) {
-      writers.push(new BasicTerminalWriter())
-    }
+    writers.push(new WRITER_CLASSES[loggerType]())
+    // if (loggerType === LoggerType.fancy) {
+    //   writers.push(new FancyTerminalWriter())
+    // } else if (loggerType === LoggerType.basic) {
+    //   writers.push(new BasicTerminalWriter())
+    // }
```
edvald: This should not be commented out, I'm assuming.

garden-service/src/cli/cli.ts
```diff
@@ -233,7 +247,14 @@ export class GardenCli {
...
       const root = resolve(process.cwd(), parsedOpts.root)
-      const { emoji, env, loglevel, silent, output } = parsedOpts
+      const { emoji, env, loglevel, jsonLog, silent, output } = parsedOpts
+
+      let loggerType
+      if (jsonLog) {
+        loggerType = LoggerType.json
+      } else {
+        loggerType = command.loggerType || DEFAULT_CLI_LOGGER_TYPE
+      }
```
edvald: @thsig Was there a reason to add a CLI arg and not to just use the env var?

garden-service/src/commands/build.ts
```diff
@@ -86,6 +86,8 @@ export class BuildCommand extends Command<BuildArguments, BuildOptions> {
       },
     })
 
+    // log.serialize(2)
+
```
edvald: Remove commented line.

garden-service/src/commands/validate.ts
```diff
@@ -28,6 +28,10 @@ export class ValidateCommand extends Command {
     const graph = await garden.getConfigGraph()
     await graph.getModules()
 
+    // const rawLog = await runGarden(garden.projectRoot, ["build", "-l", "4"])
+    // console.log(rawLog.join("\n"))
+    // console.log(findTasks({ entries: rawLog, taskType: "build",  taskName: "node-module" }))
```
edvald: Remove commented lines, or replace with `log.silly/verbose`

garden-service/src/garden.ts
```diff
@@ -164,7 +165,7 @@ export class Garden {
     this.pluginModuleConfigs = []
     this.registeredPlugins = {}
 
-    this.taskGraph = new TaskGraph(this, this.log)
+    this.taskGraph = new TaskGraph(this, this.log) // TODO: Add integ logging option
```
edvald: What would that look like? Would we not just use the log level?

garden-service/src/cli/cli.ts
thsig: No real reason — I've just tended to use env vars less often myself, so I didn't think of that option. Would that be a cleaner approach here, in your opinion?

garden-service/src/logger/log-entry.ts
```diff
@@ -189,6 +200,22 @@ export class LogEntry extends LogNode {
...
+    if (msg) {
+      const out = section ? `${section}: ${msg}` : msg
+      console.log(out)
+    }
+
+    for (const childEntry of this.children) {
+      childEntry.serialize(logLevel)
+    }
+  }
+
```
edvald: Can you explain the addition of this method? A docstring would be helpful.

garden-service/src/garden.ts
thsig: This is a leftover from another approach to adding the JSON logging option, which I abandoned. Will remove this comment.

garden-service/src/plugins/kubernetes/api.ts
```diff
@@ -208,6 +205,20 @@ export class KubeApi {
...
+ * outside of a particular Garden project (and thus use BaseKubeApi instead).
+ */
+export class KubeApi extends BaseKubeApi {
+
+  constructor(public provider: KubernetesProvider) {
+    super(provider.config.context)
+  }
+
+}
+
```
edvald: I'd actually suggest just changing the `KubeApi` class to accept the `context` directly, and modifying the calling code accordingly, instead of the separation.

garden-service/src/plugins/kubernetes/init.ts
```diff
@@ -261,27 +261,31 @@ export async function cleanupEnvironment({ ctx, log }: CleanupEnvironmentParams)
   return {}
 }
 
-export async function deleteNamespaces(namespaces: string[], api: KubeApi, log: LogEntry) {
+export async function deleteNamespaces<A extends BaseKubeApi>(namespaces: string[], api: A, log?: LogEntry) {
```
edvald: See comment above, but just as an FYI the generic here is not necessary, since it has no impact on the interface beyond the constructor. You'd get the same result by just changing `KubeApi` to `BaseKubeApi`.

garden-service/src/logger/writers/json-terminal-writer.ts
```diff
@@ -0,0 +1,43 @@
...
+import { LogEntry, TaskMetadata } from "../log-entry"
+import { Logger } from "../logger"
+import { Writer } from "./base"
+import { formatForJSON } from "../renderers"
+
+export interface JSONLogEntry {
+  msg: string,
+  section?: string,
+  durationMs?: number,
+  taskMetadata?: TaskMetadata,
```
edvald: This feels questionable, in terms of coupling. Perhaps a more generic `metadata` on the `UpdateOpts` interface would be more appropriate. @eysi09, thoughts?

garden-service/src/tasks/base.ts
```diff
@@ -23,7 +23,7 @@ export interface TaskParams {
 }
 
 export abstract class BaseTask {
-  abstract type: string
+  abstract type
```
edvald: Why remove the `string` type?

garden-service/test/integ-helpers.ts
```diff
@@ -0,0 +1,120 @@
...
+    "local-tls",
+    "multiple-modules",
+    "remote-k8s",
+    "remote-sources",
+    "simple-project",
+    "simple-project-start",
+    "tasks",
+    "vote",
+    "vote-helm",
+  ]
```
edvald: Is there a reason to hardcode, over just listing the directory?

garden-service/test/integ-helpers.ts
```diff
@@ -0,0 +1,120 @@
...
+}
+
+export type TaskLogEntryResult = {
+  startedIndex: number | null,
+  completedIndex: number | null,
+  errorIndex: number | null,
+  executionTimeMs?: number,
+}
+
+export function findTasks(entries: JSONLogEntry[], baseKey: string, status?: TaskLogStatus): TaskLogEntryResult[] {
```
edvald: A docstring here would be helpful, as well as for the other exported functions. The semantics of this are a little opaque from just looking at the code.

garden-service/src/cli/cli.ts
thsig: Yeah, that's a good point. I don't have a strong opinion on env vars VS CLI options, but I've leaned slightly toward using options so far (at least for commonly-used settings).

garden-service/src/cli/cli.ts
thsig: No, I'll delete these.

garden-service/test/integ/src/pre-release.ts
```diff
@@ -0,0 +1,198 @@
...
+export function changeFileStep(path: string, description: string): WatchTestStep {
+  return {
+    description, // Mandatory, because we don't want to print the absolute path
+    action: async () => {
+      await execa.shell(`echo "" >> ${path}`)
+    },
+  }
+}
+
+export const commandReloadedStep: WatchTestStep = {
```
edvald: Super minor thing, but would suggest making this a callable like the above steps, just for consistency (even if no params are needed atm).

garden-service/test/integ/src/pre-release.ts
```diff
@@ -0,0 +1,198 @@
...
+
+      await gardenWatch.run({ testSteps })
+
+    })
+
+  })
+
+  describe("vote-helm: helm & dependency calculations", () => {
+
+    it.only("runs the dev command", async () => {
```
edvald: Left `it.only` here. We gotta be careful with those, our tests were only running one test for a few days last week... I use the `-g` parameter on mocha instead.

garden-service/test/integ/src/pre-release.ts
```diff
@@ -0,0 +1,198 @@
...
+      const testSteps = [
+        dashboardUpCheck,
+        changeFileStep(resolve(voteHelmProjectPath, "api-image/app.py"), "change api-image/app.py"),
+        taskCompletedStep("build.api-image", 2),
+        taskCompletedStep("build.api", 2),
+        taskCompletedStep("deploy.api", 2),
+        taskCompletedStep("deploy.vote", 2),
+      ]
+
+      await gardenWatch.run({ testSteps })
```
edvald: This is nice and elegant 👍 ✨

garden-service/test/run-garden.ts
```diff
@@ -0,0 +1,195 @@
...
+ *   true: condition has passed (proceed to next step)
+ *   false: condition has failed (terminates the watch command)
+ */
+export type WatchTestCondition = (logEntries: JSONLogEntry[]) => Promise<boolean | null>
+
+export type WatchTestAction = (logEntries: JSONLogEntry[]) => Promise<void>
+
+export const DEFAULT_CHECK_INTERVAL_MS = 500
+
+export class GardenWatch {
```
edvald: This could use a docstring as well, describing the idea and flow.

garden-service/src/cli/cli.ts
edvald: I tend to agree, but having two separate mechanisms depending on which type you want feels off. We should go with consistency, and aim to unify the two later.

garden-service/src/cli/cli.ts
```diff
@@ -59,6 +61,12 @@ const OUTPUT_RENDERERS = {
   },
 }
 
+const WRITER_CLASSES = {
+  [LoggerType.basic]: BasicTerminalWriter,
+  [LoggerType.fancy]: FancyTerminalWriter,
+  [LoggerType.json]: JSONTerminalWriter,
```
eysi09: We usually don't uppercase abbreviations. E.g. `KubeApi` as opposed to `KubeAPI`. So this case would be `JsonTerminalWriter`.

garden-service/src/cli/cli.ts
eysi09: I agree. We should stick to the env var for now and open a separate issue for setting the logger type via CLI options.

garden-service/src/cli/cli.ts
eysi09: Same here. We should skip this extra `if` statement and just set the type via an env var.

garden-service/src/logger/writers/json-terminal-writer.ts
eysi09: Yes, I agree.

garden-service/src/tasks/base.ts
thsig: No reason, just a leftover from a spike I did at some point. Will revert this.

garden-service/src/cli/cli.ts
thsig: I replaced the `--jsonLog` option with a `--loggerType` option (`basic | fancy | json`).

PR #594, opened 2019-03-05
fix(openfaas): update faas-cli to 0.8.3

edvald:


ellenkorbes:
LGTM. Fixes the workaround of having to put the handler in a subdirectory. FYI since recompiling to try this new version I'm not running into the it-deploys-before-building issue anymore.

Issue #399, opened 2018-11-30
Building a container with a huge amount of standard output crashes the build


drubin:
Because we proxy the output of `docker build` and display it on failure if a container produces a huge amount of logs we can't handle it. 

```
Failed building shipping. Here is the output:
————————————————————————————————————————————————————————————————————————————————
stdout maxBuffer exceeded
————————————————————————————————————————————————————————————————————————————————
```

We need to figure out how we want to handle logs from commands. (kubectl and others also have the same problems.

We should figure out an abstract solution to handling this type of sinario when the stream output is unpredictable but we should handle it in the same way regardless of the setup.

Possible options are

1) store temp buffer of recent logs
2) write raw logs to file
3) Do some combo of showing the buffer and the link to the raw logs

edvald:
I suggest we do the following, either together or separately. The first fixes the bug, the others are more an enhancement:
1. Wrap `execa` to only keep the last X lines and last Y bytes of logs in memory, instead of just buffering the whole thing. This also serves to reduce spam in the console.
2. In the same wrapper, emit each written line as a log event that can be routed to UIs or elsewhere.
3. Pipe the logs to some form of storage. Could be just files to begin with, but no matter what we need to have automated cleanup in place.

edvald:
Fixed in #602

PR #602, opened 2019-03-07
fix: stream container build output and cap max buffer size

eysi09:
This is a quick fix for #399 (addresses point one in this [comment](https://github.com/garden-io/garden/issues/399#issuecomment-443343277)).

A few things:

* I opted not to append the `stderr` to the `output` string in the `spawn` function since the error is already printed at the top of the error message. This also conforms to how the container build output was displayed before this change.
* We now only print the last hundred lines of the output in case of an error. Should this be more?
* The `MAX_BUFFER_SIZE` is pretty arbitrary, any thoughts?


edvald:
Nope, looks good to me.

PR #607, opened 2019-03-08
Spring boot hot reload

thsig:
Added an example project demonstrating hot reloading for a Spring Boot Java service.


examples/spring-boot-hot-reload/devtools/src/main/java/sample/devtools/MyController.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.servlet.ModelAndView;
+
+@Controller
+public class MyController {
+
+	@PostConstruct
+	public void slowRestart() throws InterruptedException {
+		Thread.sleep(5000);
+	}
```
edvald: What does this do?

garden-service/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -270,15 +270,15 @@ function deploymentConfig(service: Service, configuredReplicas: number, namespac
...
+    periodSeconds: 2,
     timeoutSeconds: 3,
     successThreshold: 2,
-    failureThreshold: 5,
+    failureThreshold: 10,
   }
 
   container.livenessProbe = {
-    initialDelaySeconds: 15,
+    initialDelaySeconds: 90,
```
edvald: This should likely not be here?

thsig: No, true that. It's a leftover from when the probe settings PR was still under review.

PR #609, opened 2019-03-08
feat: add maven-container plugin type

edvald:
Pushing this a little early, to support a customer. We'll want to add an example for this, and some further documentation.

This plugin is a specialised version of the container plugin, that adds a couple of parameters, and a pre-build step ahead of building the actual container. The motivation being that it's very difficult and/or prohibitively slow to build large Java projects inside containers, especially if they're organised in a way that doesn't jive with typical docker build context layouts.

### DIFF
--- a/garden-service/src/plugins/maven-container/maven-container.ts
+++ b/garden-service/src/plugins/maven-container/maven-container.ts
@@ -1,0 +1,202 @@
+/*
+ * Copyright (C) 2018 Garden Technologies, Inc. <info@garden.io>
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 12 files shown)


garden-service/src/plugins/maven-container/maven-container.ts
```diff
@@ -0,0 +1,208 @@
...
+    "package",
+    "--batch-mode",
+    "-DskipTests",
+    "--projects", ":" + artifactId,
+    "--also-make",
+  ]
+  const mvnCmdStr = "mvn " + mvnArgs.join(" ")
+
+  log.setState(`Creating jar artifact...`)
+  await mvn(ctx.projectRoot, [
```
thsig: Use `mvnArgs` here instead.

PR #616, opened 2019-03-11
docs: add a simple istio example

eysi09:


examples/istio/README.md
```diff
@@ -0,0 +1,89 @@
+# Istio example project
+
+This example project demonstrates how to set up [Istio](https://istio.io/), the open-source service mesh, with a Garden project. Below you'll find brief instructions on how to install Istio via [Helm](https://helm.sh). It is recommended that you also look at the [official guide](https://istio.io/docs/setup/kubernetes/quick-start/) for different install options and in-depth explanations.
+
+Note that if you install Istio by different means, we still recommend that you [change the default port number](#change-http2-port-number) that the Istio listens on to prevent conflict with the default Garden ingress controller.
```
edvald: I'd suggest framing this differently. The default ingress controller can be disabled, and people won't usually be running/deploying multiple examples. Maybe you could rather mention that the example uses a different port to avoid conflicting with other examples, but also set the `setupIngressController` flag to `false`.

eysi09: Yeah good point.

eysi09: I've updated the text.

edvald: Maybe update this paragraph as well, for clarity, or reference Step 2 below.

examples/istio/README.md
```diff
@@ -0,0 +1,99 @@
...
+      name: http2
+      nodePort: 31380
+```
+
+### Step 3 - Install Istio components via Helm
+
+Still in the Istio directory, we can now render the template and install the Istio components into the `istio-system` namespace:
+
+```sh
+helm template install/kubernetes/helm/istio --name istio --namespace istio-system > $HOME/istio.yaml
```
edvald: Hmm, just noticed this—why `$HOME/istio.yaml` and not just `istio.yaml`

eysi09: It's just from the Istio docs. Didn't give it any thought.

Issue #586, opened 2019-03-01
For module configs defining multiple modules, only the first module is rebuilt/redeployed on watch changes


thsig:
### Current Behavior

When running `garden build -w` in the `multiple-modules` example project, changing `node-service/app.js` results in a rebuild of the `dev` module, but not the `prod` module.

Moving the `prod` module's configuration above the `dev`  module's configuration in `node-service/garden.yml` reverses this, i.e. the `prod` module is rebuilt, but not the `dev` module.

### Expected behavior

In the above scenario, both the `prod` and `dev` modules should be rebuilt.

### Your environment

Latest `master`.

PR #617, opened 2019-03-11
fix: add tasks for all affected modules on watch

thsig:
Fixes https://github.com/garden-io/garden/issues/586.

Before this commit, changing a file belonging to more than one module (i.e. where both/all modules are defined in the same `garden.yml`) would only add tasks for the first module defined in that `garden.yml`.

After this change, tasks are added for all modules in the file when their shared sources change.


garden-service/test/src/watch.ts
```diff
@@ -119,15 +133,15 @@ describe("Watcher", () => {
   it("should emit a moduleRemoved event if a directory containing a module is removed", async () => {
     emitEvent("unlinkDir", modulePath)
     expect(garden.events.log).to.eql([
-      { name: "moduleRemoved", payload: { name: "module-a" } },
+      { name: "moduleRemoved", payload: {} },
```
edvald: Should there not be a name when a module is removed?

PR #621, opened 2019-03-12
fix: use abs target paths in HR copy commands

thsig:
This fixes various issues with hot reload sync configuration and simplifies the internal logic.

One issue resolved by this change is that `Dockerfile`s with `WORKDIR`s nested under one of the sync targets no longer cause problems with hot reloading.

### DIFF
--- a/garden-service/src/plugins/kubernetes/hot-reload.ts
+++ b/garden-service/src/plugins/kubernetes/hot-reload.ts
@@ -17,7 +17,7 @@ import { getAppNamespace } from "./namespace"
 import { kubectl } from "./kubectl"
 import getPort = require("get-port")
 import { RuntimeError, ConfigurationError } from "../../exceptions"
-import { resolve as resolvePath, normalize } from "path"
+import { resolve as resolvePath, normalize, dirname } from "path"
 import { Omit, registerCleanupFunction } from "../../util/util"
 import { deline } from "../../util/string"
 import { set } from "lodash"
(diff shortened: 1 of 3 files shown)


garden-service/test/src/plugins/kubernetes/hot-reload.ts
```diff
@@ -143,34 +143,22 @@ describe("removeTrailingSlashes", () => {
...
-    [
-      [
-        { source: "src1", target: "/app/src1" },
-        { source: "src2", target: "/app/src2" },
-      ],
-      "mkdir -p /.garden/hot_reload/app/src1/ && cp -r src1/. /.garden/hot_reload/app/src1/ && " +
-      "mkdir -p /.garden/hot_reload/app/src2/ && cp -r src2/. /.garden/hot_reload/app/src2/",
+      ["/app/src1", "/app/src2"],
+      "mkdir -p /.garden/hot_reload/app && cp -r /app/src1/ /.garden/hot_reload/app/src1/ && " +
+      "mkdir -p /.garden/hot_reload/app && cp -r /app/src2/ /.garden/hot_reload/app/src2/",
```
edvald: This is quite difficult to read. Could we use more verbose keyed objects instead, for clarity?

edvald: Or better yet, split into multiple tests.

garden-service/src/plugins/kubernetes/hot-reload.ts
```diff
@@ -178,31 +176,24 @@ export async function hotReloadContainer(
...
  *
- * @param syncConfig
+ * @param syncTargets
  */
-export function makeCopyCommands(syncConfig: FileCopySpec[]) {
-  const commands = syncConfig.map(({ source, target }) => {
-    const adjustedSource = `${removeTrailingSlashes(source)}/.`
-    const absTarget = normalize(`/.garden/hot_reload/${target}/`)
-    return `mkdir -p ${absTarget} && cp -r ${adjustedSource} ${absTarget}`
+export function makeCopyCommands(syncTargets: string[]) {
```
edvald: Should be singular, right? Since it ends up as a single concatenated command, I mean.

PR #641, opened 2019-03-18
fix: use unique names for a service's ingresses

thsig:
Before this fix, services defining multiple ingresses would only have the last configured ingress deployed to the cluster (since each ingress for a given service had `metadata.name = service.name`).


garden-service/test/unit/src/plugins/kubernetes/container/ingress.ts
```diff
@@ -788,7 +788,7 @@ describe("createIngressResources", () => {
       apiVersion: "extensions/v1beta1",
       kind: "Ingress",
       metadata: {
-        name: service.name,
+        name: ingressName(service.name, 0),
```
edvald: I'd prefer having the actual string here, as the tested output. Otherwise 👍

PR #643, opened 2019-03-18
improvement(k8s): better deployment status checking

edvald:


thsig:
If I run `garden deploy` and then `garden get status` in the `hello-world` project, I get the following error:
```
events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: write EPIPE
    at _errnoException (util.js:1022:11)
    at WriteWrap.afterWrite [as oncomplete] (net.js:880:14)```

edvald:
I've fixed the above issue. Was an elusive one—apparently Node dislikes it when a process exits while you're still writing to stdin.

PR #646, opened 2019-03-18
feat(k8s): cache task results

thsig:
Added caching for task results. Similarly to the caching for test runs that was already in place, these are stored by version as `ConfigMap`s.

The version used to index these results includes the task's module's current version, along with the module versions of the task's runtime dependencies (if any). This is the same formula that is used for
computing test task versions.

A versioning bug was also fixed for the test result caching functionality. Here, the version used was just the test config's module's version, whereas now the corresponding test task's version
is used to index it instead (the same as is now also used to cache task results).

### DIFF
--- a/garden-service/src/plugins/kubernetes/task-results.ts
+++ b/garden-service/src/plugins/kubernetes/task-results.ts
@@ -1,0 +1,81 @@
+/*
+ * Copyright (C) 2018 Garden Technologies, Inc. <info@garden.io>
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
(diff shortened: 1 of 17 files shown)


garden-service/src/plugins/kubernetes/task-results.ts
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Store a task run result as a ConfigMap in the cluster.
+ *
+ * TODO: Implement a CRD for this.
+ */
+export async function storeTaskResult(
+  { ctx, taskName, taskVersion, result }:
+    { ctx: PluginContext, taskName: string, taskVersion: ModuleVersion, result: RunTaskResult },
+): Promise<RunTaskResult> {
+  const k8sCtx = <KubernetesPluginContext>ctx
```
edvald: You could also do `provider: KubernetesProvider = ctx.provider` here, since that's the only thing the type cast applies to.

PR #660, opened 2019-03-22
Fix init issue

edvald:


garden-service/src/plugins/maven-container/maven-container.ts
```diff
@@ -111,6 +111,13 @@ export async function configureMavenContainerModule(params: ConfigureModuleParam
 
   const configured = await configureContainerModule({ ...params, moduleConfig: containerConfig })
 
+  const hasOwnDockerfile = await pathExists(containerHelpers.getDockerfileSourcePath(moduleConfig))
+
+  if (!configured.spec.dockerfile && !hasOwnDockerfile) {
```
edvald: I just noticed this can be simplified to just use the modified `hasDockerfile` helper. Will update.

PR #667, opened 2019-03-26
fix: include resolved config in module version

thsig:
Module versions now also include a serialized version of the resolved module config (hashed together with versions / dirty timestamps).

This fixes an issue where changes to project variables used in a module's config would not result in the module's version changing when recomputed.

This was the case because the version was based only on the latest commit and dirty timestamp of the module in question, and those of its dependencies, which doesn't factor in changes to project-level variables.

### DIFF
--- a/garden-service/src/config/module.ts
+++ b/garden-service/src/config/module.ts
@@ -7,6 +7,7 @@
  */
 
 import dedent = require("dedent")
+import stableStringify = require("json-stable-stringify")
 import * as Joi from "joi"
 import { ServiceConfig, ServiceSpec } from "./service"
 import {
@@ -146,3 +147,7 @@ export const moduleConfigSchema = baseModuleSpecSchema
       .description("The module spec, as defined by the provider plugin."),
   })
   .description("The configuration for a module.")
+
+export function serializeConfig(moduleConfig: ModuleConfig) {
+  return stableStringify(moduleConfig)
+}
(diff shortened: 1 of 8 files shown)


garden-service/src/config/module.ts
```diff
@@ -146,3 +146,7 @@ export const moduleConfigSchema = baseModuleSpecSchema
       .description("The module spec, as defined by the provider plugin."),
   })
   .description("The configuration for a module.")
+
+export function serializeConfig(moduleConfig: ModuleConfig) {
+  return JSON.stringify(moduleConfig)
```
edvald: I think we may want something like this, to get a stable hash from equivalent configs: https://www.npmjs.com/package/json-stable-stringify

garden-service/src/vcs/base.ts
```diff
@@ -247,18 +208,37 @@ export async function writeModuleVersionFile(path: string, version: ModuleVersio
...
+  const latest = last(sortBy(
+    versions.filter(v => !!v.dirtyTimestamp), v => v.dirtyTimestamp)
+    .map(v => v.dirtyTimestamp))
+  return latest || null
+}
+
+/**
+ * The versions argument should consist of moduleConfig's tree version, and the tree versions of its dependencies.
+ */
+export function hashVersions(moduleConfig: ModuleConfig, versions: NamedTreeVersion[]) {
```
edvald: Would be good to add tests for this and `getLatestDirty`, even though they are quite straightforward.

Issue #626, opened 2019-03-13
CLI returns exit code 1 when help or --version args passed


eysi09:
The following commands return exit code `1` instead of `0`:

```sh
garden --version
garden help
```

Note that `garden --help` returns `0` as expected and `garden version` returns `0` if in a project directory but fails with `Not a project directory (or any of the parent directories): ...` otherwise.

eysi09:
We need to fix this for Garden to work properly in CI.

edvald:
Hmm, why is that?

eysi09:
Well, this was kind of poorly phrased, Garden works fine in CI regardless.

The above was more of a note to self since I'm trying Garden on different CI systems and like to verify that it's properly installed on the agent by running `garden --version`. However, that causes the CI job to fail because it returns exit code `1`.

eysi09:
Closed in #669

PR #669, opened 2019-03-26
fix: ensure CLI returns correct exit code (#626)

eysi09:
...also changed LoggerType from enum to union type


garden-service/src/cli/cli.ts
```diff
@@ -97,22 +97,23 @@ export const MOCK_CONFIG: GardenConfig = {
...
   }),
-  env: new EnvironmentOption(),
-  loggerType: new ChoicesParameter({
-    choices: Object.keys(WRITER_CLASSES),
-    help: `TODO`,
+  "env": new EnvironmentOption(),
+  "logger-type": new ChoicesParameter({
+    choices: [...LOGGER_TYPES],
+    defaultValue: DEFAULT_CLI_LOGGER_TYPE,
+    help: "Set logger type.",
```
thsig: I think we should include descriptions of the available logger types. Maybe something like this? Not too happy about the formatting, but we haven't been using line breaks in these descriptions so far.

> Set logger type: fancy: updates log lines in-place when their status changes (e.g. when tasks complete), basic: appends a new log line when a log line's status changes, quiet: suppresses all log output.

thsig:
Looks good! Had one comment/thought about the docstring for the `logger-type` option.

eysi09:
Since we're changing the flag, it's technically a breaking change. However, in the previous version the flag wasn't documented at all, the help text was simply: `TODO`. So I guess we can let it slide.

garden-service/src/cli/cli.ts
eysi09: There's also the `json` type, any thoughts on that one?

PR #674, opened 2019-03-28
docs: update graphics in README

edvald:


README.md
```diff
@@ -8,6 +8,10 @@ Garden is a _development orchestrator_ for Kubernetes, containers and functions,
 
 It is centered around the [Stack Graph](https://docs.garden.io/basics/stack-graph), which allows you to fully codify how each part of your stack is built, deployed and tested—making your workflow reproducible and portable.
```
eysi09: Not a part of this PR, but I was wondering if we could rephrase this to also explain what the stack graph is. Something like:

"It is centered around the [Stack Graph](https://docs.garden.io/basics/stack-graph), a graph (or data?) structure that fully codifies how each part of your stack is built, deployed and tested—making your workflow reproducible and portable."

PR #684, opened 2019-04-01
fix(dashboard): add toggle control for sidebar 

benstov:


eysi09:
It would be good if you could rename the pull request to something more descriptive and also add some description. (When you make a PR with just one commit, Github uses the git message for the PR title and description. When the PR contains several commits, Github just uses the branch name. So in this case we get "Dashboard uistate hacks" which doesn't really describe what the PR does.)

We should also squash the last three commits.

dashboard/src/app.tsx
```diff
@@ -27,58 +26,103 @@ import { DataProvider } from "./context/data"
 import { NavLink } from "./components/links"
 
 import logo from "./assets/logo.png"
+import { ReactComponent as Hamburger } from "./assets/hamburger.svg"
```
eysi09: How does this work? Where does ReactComponent come from?

PR #690, opened 2019-04-03
feat(config): add var alias for variables template key

edvald:
Because why not?


eysi09:
Should we remove `variables` altogether for the next minor release?

edvald:
I'm not sure we want that, but we might remove it from documentation at least or flag it as deprecated.

PR #698, opened 2019-04-05
feat(core): add module include field and use content hash for versions

### LINKED ISSUE #328: Support .gardenignore in subdirectories

### LINKED ISSUE #603: Support specifying source paths in module configuration
We should add an optional `sourcePaths` field to module configurations, to enable specifying the path/paths where the module's sources reside.

These paths could e.g. be relative paths from project root, and should be independent of the location of the `garden.yml` file where the module configuration in question resides.

edvald:
This changes the version handling to no longer rely on git commit hashes
but instead use content hashes on all included files.

Also added is the `include` field on modules, which allows for control
over which files are considered part of the module sources.

Further, this improves handling of .gardenignore files, so they now
work in subdirectories as well as the project root.

Note: .gitignore files no longer factor in when evaluating module versions.
Use .gardenignore files instead (same format) to exclude files from
file watching and versioning in Garden.

Closes #328
Closes #603


garden-service/src/watch.ts
```diff
@@ -74,41 +76,70 @@ export class Watcher {
     }
   }
 
+  private makeFileAddedHandler(modules: Module[]) {
+    return this.wrapAsync(async (path: string) => {
+      this.log.debug(`Watcher: File ${path} added`)
+
+      const changedModules = await Bluebird.filter(modules, async (m) => {
```
thsig: The number of string comparisons here is linear in the number of tracked files in the project.

Maybe we should load the file paths into a `Set` instead? E.g.
```javascript
const changedModules = await Bluebird.filter(modules, async (m) => {
  const paths = Set.new(files.map(f => f.path))
  return paths.has(f.path)
})
```
Just mentioning it since the watch delay is performance-sensitive for the user.

PR #699, opened 2019-04-05
fix(run): correctly ignore task dependencies

thsig:
Before this fix, the run module and run service commands would fail if any of the corresponding module's services had a task dependency.

For example, `garden run module hello "node"` and `garden run service hello` in the `tasks` example project both failed before this change (because the `hello` service has a task dependency).

### DIFF
--- a/garden-service/src/commands/run/service.ts
+++ b/garden-service/src/commands/run/service.ts
@@ -15,9 +15,8 @@ import {
   CommandResult,
   StringParameter,
 } from "../base"
-import { printRuntimeContext } from "./run"
+import { printRuntimeContext, runtimeContextForServiceDeps } from "./run"
 import dedent = require("dedent")
-import { prepareRuntimeContext } from "../../types/service"
 import { logHeader } from "../../logger/util"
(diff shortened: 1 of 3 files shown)


garden-service/src/commands/run/service.ts
```diff
@@ -71,8 +71,11 @@ export class RunServiceCommand extends Command<Args, Opts> {
     await garden.addTask(pushTask)
     await garden.processTasks()
 
-    const dependencies = await graph.getServices(module.serviceDependencyNames)
-    const runtimeContext = await prepareRuntimeContext(garden, graph, module, dependencies)
+    // combine all dependencies for all services in the module, to be sure we have all the context we need
+    const depNames = module.serviceDependencyNames
+    const allServices = await graph.getServices()
+    const deps = allServices.filter(s => depNames.includes(s.name))
+    const runtimeContext = await prepareRuntimeContext(garden, graph, module, deps)
```
edvald: Could we deduplicate the code across the files? Looks to be exactly the same.

PR #710, opened 2019-04-10
improvement(config-graph): add more data to rendered graph nodes

eysi09:


garden-service/src/config-graph.ts
```diff
@@ -524,9 +552,47 @@ export class DependencyGraphNode {
...
+          ...renderedNode,
+        }
+      case "run":
+        return {
+          type,
+          taskName: this.name,
+          ...renderedNode,
+        }
+      case "test":
+        const { testName } = parseTestKey(this.name)
```
thsig: Maybe I'm misunderstanding something here, but couldn't this be read directly from `this.testName`?

garden-service/src/config-graph.ts
```diff
@@ -576,6 +642,15 @@ function nodeKey(type: DependencyGraphNodeType, name: string) {
   return `${type}.${name}`
 }
 
+function testKey(moduleName: string, testName: string) {
+  return `${moduleName}.${testName}`
+}
+
+function parseTestKey(key: string) {
```
thsig: If the above comment is valid, then this function can be removed.

garden-service/src/config-graph.ts
eysi09: There is no `this.testName`.

eysi09: ...but perhaps the function is a little superfluous.

PR #712, opened 2019-04-10
fix(vcs): fixed path handling for modified files

thsig:
This fixes an issue with the new version hash calculation logic where tests would not be re-run even if their underlying module sources changed.

Because the relative paths being used were relativized from different starting directories (git root VS module root), the modified file paths weren't matching with the module file paths.

### DIFF
--- a/garden-service/test/unit/src/vcs/git.ts
+++ b/garden-service/test/unit/src/vcs/git.ts
@@ -59,14 +59,42 @@ describe("GitHandler", () => {
       ])
     })
 
-    it("should return untracked files as absolute paths with hash", async () => {
-      await createFile(join(tmpPath, "foo.txt"))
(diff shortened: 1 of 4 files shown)


garden-service/test/unit/src/vcs/git.ts
```diff
@@ -59,6 +59,24 @@ describe("GitHandler", () => {
       ])
     })
 
+    it("should return different hashes before and after a file is modified", async () => {
```
eysi09: Do we also check this for untracked files?

edvald: The test below should catch those, but we should probably add another one (actually for this new test as well) to make sure we cover both path cases (calling from module root and git root).

thsig: Added some more here—was this what you had in mind?

PR #713, opened 2019-04-11
fix(config-graph): remove superfluous data from rendered graph nodes

eysi09:


eysi09:
@thsig Sorry for the back and forth on this one :)

PR #715, opened 2019-04-11
feat(dashboard): add 'more info' pane to stack graph

benstov:
- added icons to the graph
- added right pane that shows up upon one of the graph nodes of type "test"/"run"
- added command ```get task-result <task_name>```
- added command ```get test-result <test_name> <module_name>```
Screenshot 2019-04-25 at 16 03 06


dashboard/src/util/helpers.ts
```diff
@@ -12,3 +12,23 @@ import { ModuleConfig } from "../api/types"
 export function getServiceNames(moduleConfigs: ModuleConfig[]) {
   return flatten(moduleConfigs.map(m => m.serviceConfigs.map(s => s.name)))
 }
+
+export function timeConversion(millisec) {
```
edvald: Isn't there a library function we can use for this? Something in moment.js for example.

dashboard/src/api/types.ts
```diff
@@ -84,11 +86,20 @@ export interface FetchConfigResponse {
   moduleConfigs: ModuleConfig[]
 }
 
-export type RenderedNode = { type: RenderedNodeType, name: string }
+export type RenderedNode = { key: string, type: RenderedNodeType, name: string, moduleName: string }
 
-export type RenderedNodeType = "build" | "deploy" | "run" | "test" | "push" | "publish"
+export type RenderedNodeType =
```
eysi09: Any particular reason you changed the format here?

dashboard/src/app.tsx
```diff
@@ -32,8 +35,8 @@ import { UiStateProvider, UiStateContext } from "./context/ui"
 
 // Style and align properly
 const Logo = styled.img`
-  height: auto;
-  width: 100%;
+  height: 67px;
```
eysi09: Shouldn't the other value be `auto` so that the dimensions stay correct? Also we should use `rem`s here, right?

dashboard/src/components/graph/index.tsx
```diff
@@ -48,11 +49,18 @@ export interface Graph {
 const MIN_CHART_WIDTH = 200
 const MIN_CHART_HEIGHT = 200
 
-function drawChart(graph: Graph, width: number, height: number) {
+function drawChart(
```
eysi09: Same as above, why change the format?

dashboard/src/components/graph/index.tsx
```diff
@@ -48,11 +49,18 @@ export interface Graph {
...
+  graph: Graph,
+  width: number,
+  height: number,
+  onNodeSelected: (string) => void,
+) {
   // Create the input graph
   const g = new dagreD3.graphlib.Graph()
     .setGraph({})
-    .setDefaultEdgeLabel(function() { return {} })
+    .setDefaultEdgeLabel(function() {
```
eysi09: Format again. We should actually just remove this line since it doesn't do anything.

dashboard/src/components/graph/index.tsx
```diff
@@ -115,16 +125,25 @@ function drawChart(graph: Graph, width: number, height: number) {
   // svg.attr("height", Math.max(graphHeight, MIN_CHART_HEIGHT))
 
   // Center the graph
-  const xCenterOffset = (parseInt(svg.attr("width"), 10) - g.graph().width * initialScale) / 2
-  const yCenterOffset = (parseInt(svg.attr("height"), 10) - g.graph().height * initialScale) / 2
-  const zoomTranslate = d3.zoomIdentity.translate(xCenterOffset, yCenterOffset).scale(initialScale)
+  const xCenterOffset =
```
eysi09: Ok, I'm assuming all these format changes happened automatically. Perhaps it's line length setting thing. In any case we should make sure we both have the same settings so that the format doesn't keep jumping round.

dashboard/src/components/graph/index.tsx
```diff
@@ -134,28 +153,18 @@ interface State {
...
-  if (type === "test") {
-    type += ` (${nameParts[1]})`
-  }
-  return "<div class='label-wrap'><span class='name'>" +
-    nameParts[0] + "</span><br /><span class='type'>" + type + "</span></div>"
+const makeLabel = (name: string, type: string, moduleName: string) => {
+  return `
+    <div class='label-wrap'>
+      <span class='name'>${moduleName === name ? `${moduleName}` : `${moduleName} - ${name}`}</span>
+
```
eysi09: Remove empty line.

dashboard/src/components/overview.tsx
```diff
@@ -75,17 +75,32 @@ interface IngressesProp {
   ingresses: ServiceIngress[]
 }
 
+const startAndEnd = (str) => {
```
eysi09: Nice. But could you rename it to `truncateMiddle` or similar.

eysi09:
@benstov and @edvald. I was wondering if the left pane should slide over the graph itself and cover it, instead of pushing it to the side and making it shrink. Making it shrink can look weird when there's a lot of nodes. 

The downside is that you won't reach graph nodes that are under the pane unless you close the pane first.

garden-service/src/commands/get/get-task-result.ts
```diff
@@ -0,0 +1,111 @@
...
+  async action({
+    garden,
+    log,
+    args,
+  }: CommandParams<Args>): Promise<CommandResult<TaskResultOutput>> {
+    const taskName = args.name
+
+    if (!taskName) {
+      throw new ParameterError(
+        `Failed to find task, provided task name is cannot be empty.`,
```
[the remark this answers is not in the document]
benstov: i guess you're right, I forgot about this flag

dashboard/src/components/graph/index.tsx
benstov: yes, lets sync on that one.

eysi09:
Some notes on the UI from my end:

- Task graph nodes that aren't clickable shouldn't have a cursor
- Info pane needs a "close" button
- I still think the dashboard info pane should cover the graph, not push it to the side. We could even use React modal (like we used on the website) for this but just position it differently. See screenshot below. If we stick to the current layout, we should manually trigger a resize event which will re-draw the graph so that it gets centred again after the pane opens. 
- The tab/label looks a little out of place imo.
- Use different styling between *key* and *value* in the info list table and skip the colon. E.g. make "Module Name" bold and "node-service" normal.
- The terminal output font size could be smaller in my opinion.

Screenshot 2019-04-17 at 09 39 34

garden-service/src/commands/get/get-task-result.ts
```diff
@@ -0,0 +1,101 @@
+/*
```
eysi09: This command needs tests.

garden-service/src/commands/get/get-test-result.ts
```diff
@@ -0,0 +1,124 @@
+/*
```
eysi09: This command also needs tests.

dashboard/src/components/graph/index.tsx
```diff
@@ -70,10 +78,10 @@ function drawChart(graph: Graph, width: number, height: number) {
     // Round the corners of the nodes
     node.rx = node.ry = 5
     // Add more padding
-    node.paddingBottom = 20
-    node.paddingTop = 20
-    node.paddingLeft = 20
-    node.paddingRight = 20
+    node.paddingBottom = 0
```
eysi09: Can't we just skip this altogether? Or is there some default padding? If there's default padding we should change the comment above to "Add less padding".

dashboard/src/components/graph/index.tsx
```diff
@@ -115,16 +125,26 @@ function drawChart(graph: Graph, width: number, height: number) {
   // svg.attr("height", Math.max(graphHeight, MIN_CHART_HEIGHT))
 
   // Center the graph
-  const xCenterOffset = (parseInt(svg.attr("width"), 10) - g.graph().width * initialScale) / 2
-  const yCenterOffset = (parseInt(svg.attr("height"), 10) - g.graph().height * initialScale) / 2
-  const zoomTranslate = d3.zoomIdentity.translate(xCenterOffset, yCenterOffset).scale(initialScale)
+  const xCenterOffset =
```
eysi09: Can you revert the unnecessary format changes?

dashboard/src/components/graph/index.tsx
```diff
@@ -251,6 +275,10 @@ class Chart extends Component<Props, State> {
     if (prevState.filters !== this.state.filters) {
       this.drawChart()
     }
+
+    if (_prevProps.selectedGraphNode !== this.props.selectedGraphNode) {
```
eysi09: Remove underscore and add type annotation to the parameter where the function is defined. Like `  componentDidUpdate(prevProps: Props, prevState: State) {`.

I just put the underscore because originally `prevProps` wasn't used.

eysi09:
I'd suggest having a standalone `InfoPane` component that handles the switch case logic that's currently in `containers/graph.tsx`. 

That component could also be responsible for preparing the data for the child components, e.g. calculating the `startedAt` values and such. That would prevent some of the logic that's duplicated between `TaskResultNodeInfo` and `TestResultNodeInfo`. 

That being said, let's just merge this and revisit when we add more stuff to the Info pane.

dashboard/src/containers/test-result-node-info.tsx
```diff
@@ -0,0 +1,205 @@
...
+}) => {
+  const {
+    actions: { loadTestResult },
+    store: { testResult },
+  } = useContext(DataContext)
+  const {
+    actions: { clearGraphNodeSelection },
+  } = useContext(UiStateContext)
+
+  useEffect(() => loadTestResult({ name, module }, true), [])
```
eysi09: You need to add `name` and `module` to the array. Like so:

```
  useEffect(() => loadTestResult({ name, module }, true), [name, module])
```

eysi09:
Once we've resolved this last batch of comments, I think we're good to merge!

eysi09:
Oh and one more: We should label the commit type as a `feat`, not `improvement`.

PR #719, opened 2019-04-16
docs: fix a few issues and improve the generated reference docs

edvald:
1. Some docs had not been updated recently due to a bug
2. The Maven Container module type was not documented
3. We were still using the old config style
4. Added specific docs for each module type
5. Moved schema handling and description to a plugin handler

### DIFF
--- a/garden-service/src/docs/config.ts
+++ b/garden-service/src/docs/config.ts
@@ -18,54 +18,44 @@ import {
   get,
   flatten,
   padEnd,
+  uniq,
+  startCase,
 } from "lodash"
-import { containerModuleSpecSchema } from "../plugins/container/config"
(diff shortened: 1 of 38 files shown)


garden-service/src/docs/config.ts
```diff
@@ -407,31 +397,65 @@ function renderModuleTypeReference(schema: Joi.ObjectSchema, name: string) {
...
+        ],
+      },
+    },
+  })
+
   // Render provider docs
   const providerDir = resolve(referenceDir, "providers")
   for (const { name, schema } of providers) {
     const path = resolve(providerDir, `${name}.md`)
+    console.log("->", path)
```
eysi09: This is intentional, right?

edvald: Yes, it is.

PR #720, opened 2019-04-17
Support Microk8s

### LINKED ISSUE #406: Support MicroK8s
We should add explicit support for https://microk8s.io. Should be an easy addition to the `local-kubernetes` plugin, similar to the `minikube` support.

### LINKED ISSUE #711: Error with multiple entries in KUBECONFIG env var
### Current Behavior

If I set a `KUBECONFIG` environment variable with multiple paths (as is supported by `kubectl`), e.g. `KUBECONFIG=/Users/me/.kube/config:/Users/me/.kube/k3s.yaml`, Garden crashes with `can't read property server of null`.

### Expected behavior

We should read the configuration from all config files listed in `KUBECONFIG`, and we should throw a clearer error when loading the configs fails.

### Reproducible example

```
# from the garden repo
cd examples/simple-project
export KUBECONFIG=$HOME/.kube/config:$HOME/.kube/k3s.yaml # (or any other kubeconfig path)
garden deploy
```

### Workaround

Set the `KUBECONFIG` variable to the appropriate path when running `garden`. For example:
```
KUBECONFIG=$HOME/.kube/config garden
``` 
You can also alias `garden` to include the modified variable:
```
alias garden='KUBECONFIG=$HOME/.kube/config garden'
``` 

### Suggested solution(s)

We either need to implement our own handling of the variable, because the library we use doesn't support this, or split and use only the first path in the env variable.

We also need to improve error handling when initializing the Kubernetes API wrapper.

### Your environment

`garden version`
v0.9.7

edvald:
Closes #406 
Closes #711

### DIFF
--- a/docs/basics/installation.md
+++ b/docs/basics/installation.md
@@ -93,6 +93,33 @@ Use your preferred method or package manager to install `git` and `rsync`. On Ub
 
 Once you have the dependencies set up, download the Garden CLI for your platform from our [latest release](https://github.com/garden-io/garden/releases/latest) page, extract and make sure it is on your PATH. E.g. by extracting to `~/.garden/bin` and adding `export PATH=$PATH:~/.garden/bin` to your `.bashrc` or `.zshrc` file.
 
+## MicroK8s
+
+Garden can be used with [MicroK8s](https://microk8s.io) on supported Linux platforms.
+
+To install it, please follow [their instructions](https://microk8s.io/docs/).
+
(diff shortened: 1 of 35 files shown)


edvald:
@ellenkorbes can you give this a spin?

edvald:
#723 should fix the CI issue

edvald:
I've tested this myself with microk8s in an Ubuntu 18.04 VM. Seems to work as expected. I've done my own tests for both, but we should make sure nothing breaks with Docker for Desktop and Minikube along the way.

Probably best to review individual commits here btw.

PR #725, opened 2019-04-18
feat(cli): allow --hot=* in dev/deploy commands

thsig:
Passing an asterisk (`*`) to the `--hot`/`--hot-reload` option deploys all compatible services with hot reloading enabled (i.e. all services belonging to a module specifying the `hotReload` field).

Also throw an error before initializing the environment or proceeding with deployment if one or more service names passed to the `--hot`/`--hotReload` option aren't configured for hot reloading.

### DIFF
--- a/docs/reference/commands.md
+++ b/docs/reference/commands.md
@@ -146,13 +146,14 @@ Optionally stays running and automatically re-builds and re-deploys services if 
 
 Examples:
 
-    garden deploy                         # deploy all modules in the project
-    garden deploy my-service              # only deploy my-service
-    garden deploy service-a,service-b     # only deploy service-a and service-b
-    garden deploy --force                 # force re-deploy of modules, even if they're already deployed
-    garden deploy --watch                 # watch for changes to code
-    garden deploy --watch --hot-reload=my-service # deploys all services, with hot reloading enabled for my-service
-    garden deploy --env stage             # deploy your services to an environment called stage
+    garden deploy                      # deploy all modules in the project
+    garden deploy my-service           # only deploy my-service
+    garden deploy service-a,service-b  # only deploy service-a and service-b
+    garden deploy --force              # force re-deploy of modules, even if they're already deployed
+    garden deploy --watch              # watch for changes to code
+    garden deploy --hot=my-service     # deploys all services, with hot reloading enabled for my-service
+    garden deploy --hot=*              # deploys all compatible services with hot reloading enabled
+    garden deploy --env stage          # deploy your services to an environment called stage
 
 ##### Usage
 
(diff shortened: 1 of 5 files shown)


docs/reference/commands.md
```diff
@@ -171,7 +172,7 @@ Examples:
   | `--force` |  | boolean | Force redeploy of service(s).
   | `--force-build` |  | boolean | Force rebuild of module(s).
   | `--watch` | `-w` | boolean | Watch for changes in module(s) and auto-deploy.
-  | `--hot-reload` | `-hot` | array:string | The name(s) of the service(s) to deploy with hot reloading enabled. Use comma as a separator to specify multiple services. When this option is used, the command is run in watch mode (i.e. implicitly assumes the --watch/-w flag).
+  | `--hot-reload` | `-hot` | array:string | The name(s) of the service(s) to deploy with hot reloading enabled. Use comma as a separator to specify multiple services. Use * to deploy all services with hot reloading enabled (ignores services belonging to modules that don&#x27;t specify the hotReload field). When this option is used, the command is run in watch mode (i.e. implicitly assumes the --watch/-w flag).
```
edvald: I'd suggest "that don&#x27;t specify the hotReload field" -> "that don't support or haven't configured hot reloading" (since the field is specific to the container module type)

docs/using-garden/hot-reload.md
```diff
@@ -8,6 +8,8 @@ For example, services that can be run with a file system watcher that automatica
 
 Currently, services are only deployed with hot reloading enabled when their names are passed to the `--hot` option via `garden deploy` or `garden dev` commands (e.g. `garden dev --hot=foo-service,bar-service`). If these services don't belong to a module defining a `hotReload` configuration (see below for an example), an error will be thrown if their names are passed to the `--hot` option.
 
+You can also pass `*` (e.g. `--hot=*`/`--hot-reload=*`) to deploy all compatible services with hot reloading enabled (i.e. all services belonging to a module that defines a `hotReload` configuration).
```
edvald: Same as my other comment.

garden-service/src/commands/helpers.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { ConfigGraph } from "../config-graph"
+
+export async function getHotReloadServiceNames(
+  namesFromOpt: string[] | undefined, configGraph: ConfigGraph,
+) {
+  const names = namesFromOpt || []
+  if (names[0] === "*") {
+    return (await configGraph.getServices())
+      .filter(s => !!s.module.spec.hotReload)
```
edvald: This check doesn't really work except for specific module types. Might be out of scope for this PR to fix it, but we generally should never "peek" into the type-specific spec at the framework level.

garden-service/src/commands/helpers.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Returns an error message string if one or more serviceNames refers to a service that's not configured for
+ * hot reloading, or if one or more of serviceNames referes to a non-existent service. Returns null otherwise.
+ */
+export async function validateHotReloadServiceNames(
+  serviceNames: string[], configGraph: ConfigGraph,
+): Promise<string | null> {
+  const services = await configGraph.getServices(serviceNames)
+  const invalidNames = services.filter(s => !s.module.spec.hotReload).map(s => s.name)
```
edvald: Same note as above, but we should at least deduplicate the code here.

garden-service/src/commands/helpers.ts
thsig: Yeah, good point. I was thinking of this check as a proxy for "the module uses the `hotReload` field, and its module type defines the `hotReload` field", but I guess it's a bit too specific.

How would you prefer to go about this here?

edvald: I think it could be as simple as making `hotReload` a native `ModuleConfig` field, in the same way `build` is.

edvald: But we can leave that for another PR, maybe just make a note or issue for it.

PR #730, opened 2019-04-22
fix(task-graph): ensure graph node keys and event payload keys match

eysi09:
In the dashboard, we match the keys emitted with the event payload
against the node keys returned by the get graph command.

This commit adds a generic makeBaseKey function that both the Task
classes and the config graph render function use.

Previously we emitted the task key, as opposed to just the base key.

### DIFF
--- a/dashboard/package-lock.json
+++ b/dashboard/package-lock.json
@@ -6707,7 +6707,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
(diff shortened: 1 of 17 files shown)


eysi09:
I put the `makeBaseKey` function in the `tasks/base.ts` file instead of the `tasks/helpers.ts` because the latter caused some circular import issues (see e.g. [here](https://stackoverflow.com/questions/43176006/typeerror-class-extends-value-undefined-is-not-a-function-or-null)).

PR #734, opened 2019-04-24
test(garden-service): add GetTaskResultCommand and GetTestResultCommand tests

benstov:


eysi09:
We need to merge the `add-info-pane` branch first or change the base branch for this one. Currently it's based on master and also contains the add-info-pane commit (and all its file changes).

eysi09:
Also, we should change the commit scope from `garden-service` to `commands`. So the commit message would start with `test(commands): ...`.

garden-service/test/unit/src/commands/get/get-task-result.ts
```diff
@@ -0,0 +1,44 @@
+import { dataDir, makeTestGarden, expectError } from "../../../../helpers"
+import { resolve } from "path"
+import { GetTaskResultCommand } from "../../../../../src/commands/get/get-task-result"
+import { expect } from "chai"
+
+describe("GetTaskResultCommand", () => {
+  it("should not find module by module name", async () => {
```
eysi09: The message doesn't really make sense. Shouldn't it be something like: "should throw error if task not found"

garden-service/test/unit/src/commands/get/get-task-result.ts
```diff
@@ -0,0 +1,44 @@
...
+    const garden = await makeTestGarden(
+      resolve(dataDir, "test-project-dependants"),
+    )
+    const log = garden.log
+    const command = new GetTaskResultCommand()
+
+    await expectError(
+      async () =>
+        await command.action({ garden, log, args: { name }, opts: {} }),
+      "parameter",
```
eysi09: Not relevant to this PR per se but the command should throw a `NotFound` error if the task name doesn't exist.

garden-service/test/unit/src/commands/get/get-task-result.ts
eysi09: The message doesn't really make sense. Shouldn't it be something like: "should throw error if task not found"

garden-service/test/unit/src/commands/get/get-task-result.ts
```diff
@@ -0,0 +1,44 @@
...
+    )
+  })
+
+  it("should find module", async () => {
```
eysi09: Again, this need a better description. Something like: "it should return the task result".

garden-service/test/unit/src/commands/get/get-test-result.ts
```diff
@@ -0,0 +1,47 @@
+import { dataDir, makeTestGarden, expectError } from "../../../../helpers"
+import { resolve } from "path"
+import { GetTestResultCommand } from "../../../../../src/commands/get/get-test-result"
+import { expect } from "chai"
+
+describe("GetTestResultCommand", () => {
+  it("should not find module by module name", async () => {
```
eysi09: Same as above for the description.

garden-service/test/unit/src/commands/get/get-test-result.ts
```diff
@@ -0,0 +1,47 @@
...
+    const garden = await makeTestGarden(
+      resolve(dataDir, "test-project-dependants"),
+    )
+    const log = garden.log
+    const command = new GetTestResultCommand()
+
+    await expectError(
+      async () =>
+        await command.action({ garden, log, args: { name, module }, opts: {} }),
+      "parameter",
```
eysi09: Same as above for the error type.

garden-service/test/unit/src/commands/get/get-test-result.ts
```diff
@@ -0,0 +1,47 @@
...
+    )
+  })
+
+  it("should find module", async () => {
```
eysi09: Same as above for the description.

garden-service/test/unit/src/commands/get/get-task-result.ts
```diff
@@ -0,0 +1,44 @@
...
+    const command = new GetTaskResultCommand()
+
+    const res = await command.action({
+      garden,
+      log,
+      args: { name },
+      opts: {},
+    })
+
+    expect(res.result).to.exist
```
eysi09: In general we shouldn't use multiple `expect` statements in the same `it` function. If there are multiple cases they should be handled my multiple `it`s.

garden-service/test/unit/src/commands/get/get-task-result.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    if (res.result) {
```
eysi09: Again with the multiple `expect`s. Here you could just do something like:

```javascript
expect(res.result).to.eql({ output: null, name })
```

Notice also the difference between `equal` and `eql`. The former is for strictly equals (`===`) and the latter for deep objects equals. (IMO it's a terrible convention. There's also `deepEqual` which is the same as `eql` but way clearer)`

garden-service/test/unit/src/commands/get/get-test-result.ts
```diff
@@ -0,0 +1,47 @@
...
+    const command = new GetTestResultCommand()
+
+    const res = await command.action({
+      garden,
+      log,
+      args: { name, module },
+      opts: {},
+    })
+
+    expect(res.result).to.exist
```
eysi09: Same as above.

garden-service/test/unit/src/commands/get/get-test-result.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    if (res.result) {
```
eysi09: Same as above. 

The reason that it's better to have a single `expect` per `it` is that it's easier to debug. If one of the fails you don't know which one.

eysi09:
@benstov Would be good to resolve the comments and get this one approved.

eysi09:
@benstov bump. Please resolve the comments so that we can close this one.

garden-service/test/unit/src/commands/get/get-task-result.ts
benstov: thanks, this is really helpful!

Issue #735, opened 2019-04-25
Istio example fails to deploy


jpreese:
### Current Behavior

Istio example fails on the `garden deploy` step, specifically for trying to deploy the bookinfo-gateway service. Sometimes it can't find the gateway, sometimes it can't find the VirtualService. The status code is HTTP 404.

### Expected behavior

Istio deploys successfully.

### Reproducible example

Attempt to deploy the Istio example

### Your environment

`kubectl version`
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.0", GitCommit:"ddf47ac13c1a9483ea035a79cd7c10005ff21a6d", GitTreeState:"clean", BuildDate:"2018-12-03T21:04:45Z", GoVersion:"go1.11.2", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.0", GitCommit:"ddf47ac13c1a9483ea035a79cd7c10005ff21a6d", GitTreeState:"clean", BuildDate:"2018-12-03T20:56:12Z", GoVersion:"go1.11.2", Compiler:"gc", Platform:"linux/amd64"}

`docker version`
Client: Docker Engine - Community
 Version:           18.09.3
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        774a1f4
 Built:             Thu Feb 28 06:32:50 2019
 OS/Arch:           windows/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.3
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.8
  Git commit:       774a1f4
  Built:            Thu Feb 28 06:40:58 2019
  OS/Arch:          linux/amd64
  Experimental:     false

eysi09:
Hi @jpreese! Did #736 resolve this?

jpreese:
Yep!

PR #736, opened 2019-04-26
fix(k8s): use statuscodeerror type when using request-promise

jpreese:
For #735 

When calling a CRD and a non-200 status code is returned, the message is set but the status code is not. This causes the deployment status checks to fail. They are looking for 404s, which is what we want, but in the current implementation the status code is not being wrapped properly.

To resolve this, the request-promise library has a StatusCodeError for non-200 responses, so we should be setting the code by looking at that property.


edvald:
Thanks @jpreese! Looking into this now.

garden-service/src/plugins/kubernetes/api.ts
```diff
@@ -331,3 +334,16 @@ function wrapError(err) {
     throw err
   }
 }
+
+function handleRequestPromiseError(err) {
+  if (err instanceof requestErrors.StatusCodeError) {
```
edvald: I wonder if this could be handled directly in the `wrapError()` function instead? If I understand correctly the issue could be fixed by replacing `if (!err.message) {` in `wrapError()` with this line that you've added (`if (err instanceof requestErrors.StatusCodeError) {`). Same fix, just a bit less code.

jpreese: From my investigations, the problem seemed to be that we're using two different approaches to make HTTP requests. The request-promise library for CRDs, and then the Kubernetes javascript API which just uses the built in Promise.

I tried your suggested approach, but requestErrors.StatusCodeError will only be set to true for requests made through request-promise.

Maybe it would be possible to stick to one approach to reach out to the Kubernetes API so we're not handling multiple types of error responses. But thats why I split it up as I did, it's two different ways to parse the error response

edvald: Ahaa, I see! Ok cool. The idea was to use the same approach as the API library for CRDs, but I didn't realize `request-promise` threw a different type of error. Then this LGTM 👍 Thanks so much for the fix :)

We might end up getting rid of the kubernetes API library entirely, since it's kind of a mess tbh, but in any case we can refactor/clean up when we circle back to it.

eysi09:
@jpreese Thanks for the fix!

Could you please rebase master so that we can merge it?

PR #745, opened 2019-04-29
improvement(dashboard): start dashboard before init and keep same port

edvald:
We no longer restart the dashboard instance after config changes, and
instead just swap out the Garden instance from the running server.


eysi09:
LGTM.

There's some similar/duplicate logic across the commands though. Maybe we could have something like a command generator that creates similar commands based on how they describe themselves. CC @thsig .

eysi09:
Btw, I resolved some rebase conflicts related to changes I had made and force pushed.

Issue #717, opened 2019-04-15
Update contributing docs


eysi09:
Our contributing docs have become a little stale. The following needs updating:

- [x] Update commit types section in CONTRIBUTING.md
- [x] Add section about pre-commit checks to CONTRIBUTING.md
- [x] Add a step to release process for acknowledging external contributors.
- [x] Add "improvement" commit type to changelog config
- [x] Fix `garden serve` path, should be inside an example project
- [x] Fix debugging setup and update relevant docs. (Our current method of debugging doesn't work on the latest versions of Chrome. Browser extensions don't work either.)

thsig:
Here's a suggested addition to the contributing docs:
> If you’ve created a prerelease, e.g. `v1.2.3-0`, and want to create a new prerelease `v1.2.3-1` which includes changes made on master since `v1.2.3-0` was created, do the following:
> 1. `git checkout v1.2.3-0`
> 2. `git rebase master`
> 3. `./bin/release.ts prerelease`
> 
> When you’re ready to release, check out `master` again and run `./bin/release.ts` with `minor` or `patch`. This way, the version bump commits created by the prereleases are omitted from the final history.

eysi09:
Debug setup is still broken but at least it's documented.

PR #747, opened 2019-05-02
docs: update contributing docs

eysi09:
See #717


CONTRIBUTING.md
```diff
@@ -126,6 +109,26 @@ Integration tests are run with:
...
+* **fix**: A bug fix.
+* **improvement**: Changes that improve a current implementation without adding a new feature or fixing a bug.
+* **perf**: A code change that improves performance.
+* **refactor**: A code change that neither fixes a bug nor adds a feature.
+* **revert**: A commit that reverts a previous commit. It should begin with `revert: `, followed by the header of the reverted commit. In the body it should say: `This reverts commit <hash>.`, where the hash is the SHA of the commit being reverted.
+* **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing
+  semi-colons, etc).
+* **test**: Adding missing or correcting existing tests.
+
+When generating the changelog, we only include the following types: **feat**, **fix**, **refactor**, **improvement**, and **perf**. This means that any changes that the user should be aware of, should have one of these types.
```
thsig: Should **refactor** be removed from the list of types shown in the changelog? Such changes are generally invisible to the end-user.

eysi09: Yes, perhaps it's time to remove it.

We decided to include it when Garden was still flagged as alpha and more unstable. Refactors can introduce bugs so if `some-feature` doesn't work after an upgrade and the changelog contains a `refactor(some-feature)` entry, it's easy to find the offending commit. It also feels more relevant for an open source product somehow, to err on the side of transparency. 

Thoughts?

PR #748, opened 2019-05-02
improvement(dashboard): fix ui issues

benstov:
- Line breaks don't come out right in the task output section
- Change sidebar open/close icon
- Build/Run/etc checkboxes are much too large, and tend to overlap graph
- Add hover styling on nodes (to make them more obviously clickable)
- Graph jumps around when clicking a node to open the sidebar


edvald:
Sweet, thanks @benstov! I'll leave the code review for @eysi09 since I'm on the road, but looks good at first glance.

dashboard/src/components/checkbox.tsx
```diff
@@ -67,12 +69,22 @@ interface Props {
   onChange: (event: ChangeEvent<HTMLInputElement>) => void
 }
 
-const CheckBox: React.FC<Props> = ({ name, onChange, checked = false, children }) => {
+const CheckBox: React.FC<Props> = ({
```
eysi09: Looks like you line length is still not correct. It should be set to 120 characters. The code shouldn't format like this.

dashboard/src/components/graph/index.tsx
```diff
@@ -138,28 +154,41 @@ function drawChart(
...
+  const zoomTranslate = d3.zoomIdentity
+    .translate(xCenterOffset, yCenterOffset)
+    .scale(initialScale)
   svg.call(zoom.transform, zoomTranslate)
 
   const selections = svg.select("g").selectAll("g.node")
-  selections.on("click", evt => {
-    onNodeSelected(evt)
+  selections.on("click", function(evt) {
+    // tslint:disable-next-line: no-invalid-this
```
eysi09: I think you can get the element from `evt.target` or similar. So you don't have to use `this`.

garden-service/package-lock.json
```diff
@@ -412,7 +412,7 @@
     },
     "@types/accepts": {
       "version": "1.3.5",
-      "resolved": "https://registry.npmjs.org/@types/accepts/-/accepts-1.3.5.tgz",
+      "resolved": "http://registry.npmjs.org/@types/accepts/-/accepts-1.3.5.tgz",
```
eysi09: The package-lock shouldn't change. Could you please remove it from the commit.

I think this happens because you're on an older version of `npm`. Run `npm i -g npm` to update.

dashboard/src/components/checkbox.tsx
```diff
@@ -14,11 +14,13 @@ import { colors } from "../styles/variables"
 const Label = styled.label`
   display: block;
```
eysi09: Didn't we say that we wanted align the checkboxes horizontally?

dashboard/src/components/graph/index.tsx
benstov: actually   
```const selections = svg.select("g").selectAll("g.node")```
 returns an array of strings, i dont use document selector here

PR #751, opened 2019-05-03
(dashboard) Info pane refactor

eysi09:
In this PR:

* All task nodes are clickable and have an info pane.
* Nodes with test or task results also get a refresh button. We probably want to style it a little better (cc @benstov).
* `LoadWrapper` component is removed. Rendering logic is much clearer without it.
* Previous info pane gets refactored into generic containers and components.
* Use `useReducer` instead of `useState` in the `useApi` hook. It's a more idiomatic pattern for nested data structures like the Store blob. 
* Some style changes to the info pane. Keys are now bold and the `Output:` text has been removed. See screenshot below. This is of course debatable and easy to revert.

Screenshot 2019-05-03 at 20 03 34


eysi09:
Going forward, it might be nice to pull the load functions out of the `useApi` hook. If only to prevent it from growing linearly with the number of load calls. That would also make the loaders more flexible. On the flip side, we'd need to pass the `store` and `dispatch` function to all of them. So every container would need to get the `store` and `dispatch` function from the data context and then do something like `loadLogs(store, dispatch, ...otherArgs)`.  

Alternatively we could do away with the load functions and just pass the dispatch to the container components. They could then do something like: `dispatch({ type: loadLogs })`. That would reduce some code (pun kind of intended) but it feels a little to loose though.

PR #752, opened 2019-05-04
Fix hot reload param issues

edvald:


garden-service/src/config/module.ts
```diff
@@ -155,11 +154,27 @@ export interface ModuleConfig
 
 export const moduleConfigSchema = baseModuleSpecSchema
   .keys({
+    outputs: joiIdentifierMap(joiPrimitive())
+      .description("The outputs defined by the module (referenceable in other module configs)þ"),
```
eysi09: Typo, remove "þ".

PR #753, opened 2019-05-04
improvement(task-graph): raise concurrency limit

edvald:


garden-service/src/task-graph.ts
```diff
@@ -37,7 +37,7 @@ export interface TaskResults {
   [key: string]: TaskResult
 }
 
-export const DEFAULT_CONCURRENCY = 4
+export const DEFAULT_CONCURRENCY = 6
```
eysi09: Could we add a comment here about the choice of concurrency? Why are we bumping it from `4` in the first place? And why `6`?

thsig: Maybe because those are common CPU core counts?

PR #755, opened 2019-05-06
fix(dashboard): don't use PageError twice on Overview page

eysi09:


eysi09:
This is no longer an issue.

PR #764, opened 2019-05-09
document defaultHostname in using remote clusters

w9n:
Otherwise this will fail on e.g. the simple project with:

```
No hostname configured for one of the ingresses on service go-service. Please configure a default hostname or specify a hostname for the ingress.
```


eysi09:
Thank you @w9n!

PR #774, opened 2019-05-14
fix(k8s): ensure env is prepared

### LINKED ISSUE #770: (Remote) Kubernetes plugin doesn't install Tiller
### Current Behavior

First of all `getEnvironmentStatus` always return `{ ready: true }`, even if Tiller is missing. Second, even with the `--force` flag, the code path to install Tiller is never taken.

This is basically because the code paths to get Tiller status and to install Tiller in the `getEnvironmentStatus` and `prepareEnvironmentStatus` handlers respectively is never taken. 

This mean that projects that use Helm charts will fail with `Failed deploying service [...] Error: could not find Tiller`. 

### Expected behavior

Should report the correct status and install Tiller if needed. Should always install Tiller if the `--force` flag is provided.

### Reproducible example

Run `garden init` with the Kubernetes provider. Delete the project namespace first if necessary and notice that Tiller never gets deployed.

### Workaround

Install Tiller manually. 

### Your environment

On latest `master`.

### LINKED ISSUE #773: (Remote) Kubernetes plugin doesn't create system metadata namespace
### Current Behavior

Kubernetes plugin doesn't create the `garden-system--metadata` namespace which causes errors when Garden attempts to store test results there.

### Expected behavior

Should create the namespace.

### Reproducible example

Delete the `garden-system--metadata` namespace if already there and then run `garden init` for an environment that uses the `kubernetes` plugin.

### Workaround

Create the namespace manually with `kubectl create ns garden-system--metadata`. 

### Your environment

Latest master.

eysi09:
Before, Garden would not deploy Tiller to the project namespace nor create the system namespaces for the Kubernetes provider. This was because the Kubernetes provider doesn't have any system services.

I also used the chance to refactor `getEnvironmentStatus` and `prepareEnvironment` handlers for the K8s provider since the control flow was a bit hard to follow. In particular because `getEnvironmentStatus` triggers a call to `actions.getStatus` which again calls `getEnvironmentStatus` for the system context. This lead to some code paths never being executed which caused #773  and #770.

After the refactor, most of the logic is hoisted to the top level handlers and `getEnvironmentStatus` is no longer called recursively. 

Fixes #773.
Fixes #770.

### DIFF
--- a/garden-service/src/actions.ts
+++ b/garden-service/src/actions.ts
@@ -70,7 +70,7 @@ import {
   TaskActionParams,
   GetTaskResultParams,
 } from "./types/plugin/params"
-import { Service, ServiceStatus, getServiceRuntimeContext } from "./types/service"
+import { Service, ServiceStatus, getServiceRuntimeContext, ServiceStatusMap } from "./types/service"
 import { mapValues, values, keyBy, omit, pickBy, fromPairs } from "lodash"
 import { Omit } from "./util/util"
 import { processServices, ProcessResults } from "./process"
@@ -205,7 +205,12 @@ export class ActionHelper implements TypeGuard {
         msg: "Configuring...",
       })
 
-      await handler({ ...this.commonParams(handler, log), force: forcePrep, status, log: envLogEntry })
+      await handler({
+        ...this.commonParams(handler, log),
+        force: forcePrep,
+        status,
+        log: envLogEntry,
+      })
 
       envLogEntry.setSuccess({ msg: chalk.green("Ready"), append: true })
 
@@ -369,20 +374,25 @@ export class ActionHelper implements TypeGuard {
     log.verbose(`Getting environment status (${this.garden.projectName})`)
 
     const envStatus: EnvironmentStatusMap = await this.getEnvironmentStatus({ log })
+    const serviceStatuses = await this.getServiceStatuses({ log, serviceNames })
+    return {
+      providers: envStatus,
+      services: serviceStatuses,
+    }
+  }
+
+  async getServiceStatuses(
+    { log, serviceNames }: { log: LogEntry, serviceNames?: string[] },
+  ): Promise<ServiceStatusMap> {
+
     const graph = await this.garden.getConfigGraph()
     const services = keyBy(await graph.getServices(serviceNames), "name")
-
-    const serviceStatus = await Bluebird.props(mapValues(services, async (service: Service) => {
+    return Bluebird.props(mapValues(services, async (service: Service) => {
       const runtimeContext = await getServiceRuntimeContext(this.garden, graph, service)
       // TODO: The status will be reported as "outdated" if the service was deployed with hot-reloading enabled.
       //       Once hot-reloading is a toggle, as opposed to an API/CLI flag, we can resolve that issue.
       return this.getServiceStatus({ log, service, runtimeContext, hotReload: false })
     }))
-
-    return {
-      providers: envStatus,
-      services: serviceStatus,
-    }
   }
 
   async deployServices(
--- a/garden-service/src/plugins/kubernetes/init.ts
+++ b/garden-service/src/plugins/kubernetes/init.ts
@@ -13,39 +13,84 @@ import {
 } from "../../types/plugin/params"
 import { KubeApi } from "./api"
 import { getAppNamespace, prepareNamespaces, deleteNamespaces } from "./namespace"
-import { KubernetesProvider, KubernetesPluginContext } from "./kubernetes"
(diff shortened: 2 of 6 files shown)


garden-service/src/actions.ts
```diff
@@ -373,20 +373,25 @@ export class ActionHelper implements TypeGuard {
...
+
+  async getServiceStatuses(
+    { log, serviceNames }: { log: LogEntry, serviceNames?: string[] },
+  ): Promise<ServiceStatusMap> {
+
     const graph = await this.garden.getConfigGraph()
     const services = keyBy(await graph.getServices(serviceNames), "name")
-
-    const serviceStatus = await Bluebird.props(mapValues(services, async (service: Service) => {
+    return await Bluebird.props(mapValues(services, async (service: Service) => {
```
edvald: `return await` is redundant here (it does occasionally apply, where you're specifically catching errors for example, but unnecessary in this case).

garden-service/src/plugins/kubernetes/init.ts
```diff
@@ -13,61 +13,115 @@ import {
...
+  const systemServiceNames = k8sCtx.provider.config._systemServices
+
+  if (systemServiceNames.length > 0) {
+    // Check Tiller status in system namespace
+    let systemTillerReady = true
+    if (await checkTillerStatus(sysCtx, sysCtx.provider, log) !== "ready") {
+      systemTillerReady = false
+    }
 
-  if (!isSystemGarden(provider)) {
```
edvald: I think it might be necessary to set `_systemServices` explicitly to `[]` at https://github.com/garden-io/garden/blob/e0543ad65bc20abc6b638cd6f9d7dcfd0d857283/garden-service/src/plugins/kubernetes/system.ts#L66

garden-service/src/plugins/kubernetes/system.ts
```diff
@@ -227,30 +185,16 @@ export async function prepareSystemServices(
...
-
-  await installTiller(sysCtx, sysCtx.provider, log)
 
   // Deploy enabled system services
   if (serviceNames.length > 0) {
     const results = await sysGarden.actions.deployServices({
       log,
       serviceNames,
-      force,
+      force: true,
```
edvald: Why is this needed?

garden-service/src/plugins/kubernetes/init.ts
```diff
@@ -13,61 +13,115 @@ import {
...
-      ready,
-      dashboardPages: [],
-      detail: {},
-    }
+    systemReady = systemTillerReady && systemServiceStatuses.ready && sysNamespaceUpToDate
+    dashboardPages = systemServiceStatuses.dashboardPages
+  }
+
+  return {
+    ready: projectReady && systemReady,
```
edvald: Not _directly_ related to the issue, but you could throw in a cheap optimization here, by putting the separate statuses in the `details` field and checking that in the prepare handler. I was thinking of that recently, because otherwise init is much slower when the system services are ready, but you still need to install Tiller in the project namespace.

edvald:
Btw, small thing I noticed the other day: "Fixes #773 and #770" doesn't work as expected in the PR body, you need to put in separate lines for each resolved issue for the automatic closing to work.

garden-service/src/actions.ts
eysi09: Good catch.

garden-service/src/plugins/kubernetes/system.ts
eysi09: Good question. It was like this before and I considered removing it but vaguely remembered it had something to do with the issues we had around re-installing system Hem charts. Should I remove it?

edvald: That does sound vaguely familiar, but feels like the kind of thing where it should at least have a comment explaining why. I'd say we leave it as it was if we can't identify a clear reason for the change.

eysi09: Yeah I removed the default force. Now it's only applied with `garden init --force`.

garden-service/src/plugins/kubernetes/init.ts
eysi09: Done.

We're still checking for Tiller twice though for both namespaces since the `installTiller` function checks for Tiller as well. Alternatively we could include something like `projectTillerReady` and `systemTillerReady` and hoist the check to the `prepareEnvironment` function, so `installTiller` always installs Tiller. Decided to leave it as is for now though. Same applies to the `OpenFaas` provider.

eysi09:
I've resolved the comments.

PR #775, opened 2019-05-15
docs: use flat config style in example projects

thsig:
Also renamed a few example projects (and adjusted links to them in the docs accordingly):

`hello-world` -> `openfaas`
`simple-project` -> `demo-project`
`simple-project-start` -> `demo-project-start`

The hello-world.md guide was removed, since it's mostly a duplicate of
the material in demo-project.md (previously simple-project.md).

In a later PR, we'll add some how-to sections that will cover what is
omitted here from the old hello-world guide, as well as adding new
material in that format (instead of wrapping it in a guide as was done
in hello-world.md).

@eysi09: I think I'll need a bit of guidance from you around the link validation etc., like we talked about the other day. Let's look at this together when we next have a chance.


docs/reference/template-strings.md
```diff
@@ -13,31 +13,31 @@ For example, for one service you might want to reference something from another
...
 ```
 
 You can also inject a template variable into a string. For instance, you might need to include a module's
 version as part of a URI:
 
 ```yaml
-      # ...
-      env:
-        OTHER_MODULE_ENDPOINT: http://other-module/api/${modules.other-module.version}
+# ...
```
edvald: I think we match the indentation here to the above block. Same with the next block.

docs/using-garden/configuration-files.md
```diff
@@ -191,7 +191,76 @@ in your tests.
...
+To spin up this project, three services have to be deployed and two database migrations have to be run, and all this
+has to happen in the right dependency order. 
+
+To deploy the `user` service, first the `postgres` service has to be deployed, then `node-migration` has to be run (to
+create the `users` table), and finally, `ruby-migration` has to be run (to populate the `users` table).
+
+Garden takes care of this automatically when you run e.g. `garden deploy` or `garden dev`.
+
+You can also run tasks in an ad-hoc manner using `garden run task [task name]`. This is useful e.g. when migrating the
+DB schema for one of your services after you've already deployed your stack.
```
edvald: It might be worth noting here that tasks that are not depended upon by other services are not automatically run. For example, the `db-clear` task in our vote example(s) is only run when triggered manually.

eysi09:
Regarding link validation. I think we should just disable the check for this particular PR, and then re-enable it in a separate commit once we merge. It's either that or force merge a failed commit.

docs/reference/template-strings.md
thsig: Yeah, good point. I'd decreased the indent after pasting this, but on second thought, leaving the original indentation makes it clearer where it comes from.

docs/using-garden/configuration-files.md
thsig: Added a `db-clear` task to the `user` module, and added some added text along the lines you suggested.

eysi09:
This is ready for merge right? (Expect cheating the link checker).

examples/openfaas/garden.yml
```diff
@@ -0,0 +1,13 @@
...
+environmentDefaults:
+  providers:
+    - name: npm-package
+  variables:
+    my-variable: hello-variable
+environments:
+  - name: local
+    providers:
+      - name: local-kubernetes
+      - name: openfaas
```
edvald: After some recent changes, this needs to be changed to `local-openfaas`.

Issue #547, opened 2019-02-19
Helper command to collect debugging info


edvald:
It would be very useful to have a command that collects and archives information that would generally be useful for us when debugging user problems.

I suggest we make a `garden get debug-info` command (other suggestions for names welcome), that does the following:

1. Collect every garden.yml in the project, and store in a corresponding directory structure in a temporary folder. This should notably work even if the project doesn’t successfully validate.
2. For every found garden.yml, see if there is an error.log file next to it, if so copy it next to the garden.yml in the temporary directory.
3. For each provider that specifies it, call the `getDebugInfo` handler (which we need to add) The handler output schema should have an info key, which allows any object, which is then stored in the temp dir under `<provider-name>/info.json`. Later we may wish to add more specific keys, or a way to output individual files, but that can wait.
4. Collect some basic system info, as well as the Garden version into a JSON or YAML file and store at the top of the temp directory.
5. Zip the temp directory into a `garden-debug-<timestamp>.zip` file and delete the temp directory.

For step 3, we should at a minimum make a handler for the Kubernetes plugin, that does the following:
1. Calls `kubectl get all —all-namespaces —output yaml` and stores it under `clusterResources`.
2. Calls `kubectl version --output yaml` and stores it under `version`
3. Possibly other useful info to add?

We should then make sure to update our bug report issue template, once the command is released.

edvald:
Done in #778

PR #778, opened 2019-05-16
Feature: implement get debug-info

10ko:
This PR introduces the command:
```bash
$ garden get debug-info
```
As spec'd out in #547, it's a command to collect all sort of system, project and providers information in a zip file. This will be useful data users can attach to github issues.

Thanks,
E.


docs/reference/commands.md
```diff
@@ -324,6 +324,25 @@ Outputs the latest execution result of a provided test.
...
   | `name` | Yes | Test name.
 
+### garden get debug-info
+
+Outputs the status of your environment for debug purposes.
+
+Examples:
+
+garden get debug-info                # create a zip file on the root of the project with debug information
+garden get debug-info --format yaml  # outputs the provider info as yaml files (default as json)
```
edvald: Curious about the motivation for this. Do you have a use case for YAML here?

garden-service/src/actions.ts
```diff
@@ -409,6 +410,13 @@ export class ActionHelper implements TypeGuard {
     })
   }
 
+  async getDebugInfo({ log }: { log: LogEntry }): Promise<DebugInfoMap> {
+    const handlers = this.getActionHandlers("getDebugInfo")
+    const res = await Bluebird.props(mapValues(handlers, h => h({ ...this.commonParams(h, log) })))
+    log.setSuccess("Ready")
```
edvald: This line might belong in the surrounding context. Could then `return Bluebird.props...` above for brevity.

garden-service/src/commands/get/get-debug-info.ts
```diff
@@ -0,0 +1,209 @@
...
+  // Find all services paths
+  const ignorer = await getIgnorer(root)
+  const scanOpts = {
+    filter: (path) => {
+      const relPath = relative(root, path)
+      return !ignorer.ignores(relPath)
+    },
+  }
+  const paths: string[] = []
+  for await (const item of scanDirectory(root, scanOpts)) {
```
edvald: It looks like this routine duplicates code from the Garden class. Might be worth factoring out, unless there's a reason it can't be?

garden-service/src/commands/get/get-debug-info.ts
```diff
@@ -0,0 +1,209 @@
...
+  CONFIG_FILENAME,
+  ERROR_LOG_FILENAME,
+  GARDEN_DIR_NAME,
+} from "../../constants"
+import dedent = require("dedent")
+import { Garden } from "../../garden"
+import { zipFolder } from "../../util/archive"
+
+export const TEMP_DEBUG_ROOT = "tmp"
+export const SYSTEM_INFO_FILENAME = "systemInfo.json"
```
edvald: Maybe kebab-case this for consistency?

garden-service/src/commands/get/get-debug-info.ts
```diff
@@ -0,0 +1,209 @@
...
+export async function generateBasicDebugInfoReport(root: string, log: LogEntry) {
+  const tempPath = join(root, GARDEN_DIR_NAME, TEMP_DEBUG_ROOT)
+  await collectBasicDebugInfo(root, log)
+
+  log.info("Preparing archive.")
+  const outputFilename = DEBUG_ZIP_FILENAME.replace("TIMESTAMP", new Date().toISOString())
+  await zipFolder(tempPath, join(root, outputFilename), log)
+
+  await remove(tempPath)
+  log.info(`Done! Please find your report under ${root}.`)
```
edvald: Super minor, but `... at <full path to output file>` would allow me to quickly open or copy the path.

garden-service/src/commands/get/get-debug-info.ts
```diff
@@ -0,0 +1,209 @@
...
+}
+
+export async function collectBasicDebugInfo(root: string, log: LogEntry) {
```
edvald: Would be great to add docstrings explaining the functions, e.g. how this one should always work even if the debug logs command fails etc.

garden-service/src/commands/get/get-debug-info.ts
edvald: Side note: This async generator thingy isn't actually very nice (that's on me, I made it :P), I or someone willing should refactor it at some point.

garden-service/src/plugins/kubernetes/kubernetes.ts
```diff
@@ -155,6 +160,28 @@ export async function configureProvider({ projectName, config }: ConfigureProvid
...
 
+export async function debugInfo({ ctx, log }: GetDebugInfoParams): Promise<DebugInfo> {
+  const k8sContext = <KubernetesPluginContext>ctx
+  const { context } = k8sContext.provider.config
+  const appNamespace = await getAppNamespace(k8sContext, log, k8sContext.provider)
+  const appMetadataNamespace = await getMetadataNamespace(k8sContext, log, k8sContext.provider)
+
+  const namespacesList = [appNamespace, appMetadataNamespace, systemNamespace, systemMetadataNamespace]
+  const namespaces = await Bluebird.map(namespacesList, async (ns) => {
+    const out = await kubectl.stdout({ log, context, args: ["get", "all", "--namespace", ns, "--output", "json"] })
```
edvald: Might need to catch the case where the namespace doesn't exist.

garden-service/src/commands/get/get-debug-info.ts
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  // Run system diagnostic
+  await collectSystemDiagnostic(root, log)
+
+}
+
+export async function collectSystemDiagnostic(root: string, log: LogEntry) {
+
+  log.info("Collecting OS basic information.")
```
eysi09: Minor but shouldn't this be, `"Collecting basic OS information"`, or alternatively, just `"Collecting basic system information"`.

garden-service/src/commands/get/get-debug-info.ts
```diff
@@ -0,0 +1,209 @@
...
+type Opts = typeof debugInfoOptions
+
+export class GetDebugInfoCommand extends Command<Args, Opts> {
+  name = "debug-info"
+  help = "Outputs the status of your environment for debug purposes."
+
+  description = dedent`
+    Examples:
+
+    garden get debug-info                # create a zip file on the root of the project with debug information
```
eysi09: It should `at` instead of `on` in doc string: `"# create  a zip file at the root of the project...`. Right?

garden-service/src/commands/get/get-debug-info.ts
```diff
@@ -0,0 +1,209 @@
...
+    garden get debug-info --format yaml  # outputs the provider info as yaml files (default as json)
```
eysi09: For consistency, use `# output...`  instead of `# outputs...`.

garden-service/test/unit/src/commands/get/get-debug-info.ts
```diff
@@ -0,0 +1,146 @@
...
+const debugZipFileRegex = new RegExp(/debug-info-.*?.zip/)
+
+async function cleanupTmpDebugFiles(root: string) {
+  const allFiles = readdirSync(root)
+  await remove(join(root, GARDEN_DIR_NAME, TEMP_DEBUG_ROOT))
+  const deleteFilenames = allFiles.filter((fileName) => {
+    return fileName.match(debugZipFileRegex)
+  })
+  for await (const name of deleteFilenames) {
+    console.log(join(root, name))
```
eysi09: Remove `console.log`?

garden-service/test/unit/src/commands/get/get-debug-info.ts
```diff
@@ -0,0 +1,146 @@
...
+        const gardenProjectRootFiles = readdirSync(garden.projectRoot)
+        const zipFiles = gardenProjectRootFiles.filter((fileName) => {
+          return fileName.match(debugZipFileRegex)
+        })
+        expect(zipFiles.length).to.equal(1)
+      },
+    )
+  })
+
+  describe("generateBasicDebugInfo", () => {
```
eysi09: Feels like this could be split up into several `it` statements since it has multiple `expect`s. Can be hard to debug if one of them fails.

garden-service/src/commands/get/get-debug-info.ts
```diff
@@ -0,0 +1,209 @@
...
+  `
+
+  arguments = debugInfoArguments
+  options = debugInfoOptions
+
+  async action({ garden, log, opts }: CommandParams<Args, Opts>) {
```
eysi09: The command appears to run twice. See e.g. the logs:

```
✔ Ready
Collecting project configuration files.
Collecting OS basic information.
Preparing archive.
Done! Please find your report under /Users/eysi/code/garden-io/garden/examples/simple-project.
Collecting project configuration files.
Collecting OS basic information.
Preparing archive.
Done! Please find your report under /Users/eysi/code/garden-io/garden/examples/simple-project.
```

garden-service/src/commands/get/get-debug-info.ts
```diff
@@ -0,0 +1,209 @@
...
+import dedent = require("dedent")
+import { Garden } from "../../garden"
+import { zipFolder } from "../../util/archive"
+
+export const TEMP_DEBUG_ROOT = "tmp"
+export const SYSTEM_INFO_FILENAME = "systemInfo.json"
+export const DEBUG_ZIP_FILENAME = "debug-info-TIMESTAMP.zip"
+export const PROVIDER_INFO_FILENAME_NO_EXT = "info"
+
+export async function generateBasicDebugInfoReport(root: string, log: LogEntry) {
```
eysi09: Is this function used anywhere?

eysi09:
Had a few nitpicks on my own but otherwise looks great!

One thing we could improve though is the logging. E.g.: the body of the command action could look like this:

```
    const tempPath = join(garden.projectRoot, GARDEN_DIR_NAME, TEMP_DEBUG_ROOT)
    const entry = log.info({ msg: "Collecting debug info", status: "active" })
    const systemEntry = entry.info({ section: "system", msg: "collecting info", status: "active" })
    await collectBasicDebugInfo(garden.projectRoot, systemEntry)
    systemEntry.setSuccess({ msg: chalk.green(`Done (took ${systemEntry.getDuration(1)} sec)`), append: true })

    const providerEntry = entry.info({ section: "providers", msg: "collecting info", status: "active" })
    await collectProviderDebugInfo(garden, providerEntry, tempPath, opts.format)
    providerEntry.setSuccess({ msg: chalk.green(`Done (took ${providerEntry.getDuration(1)} sec)`), append: true })

    entry.setState("Preparing archive")
    const outputFilename = DEBUG_ZIP_FILENAME.replace("TIMESTAMP", new Date().toISOString())
    await zipFolder(tempPath, join(garden.projectRoot, outputFilename), log)

    await remove(tempPath)

    entry.setSuccess({ msg: "Done", append: true })
    log.info(`\nDone! Please find your report under ${garden.projectRoot}.`)

    return { result: 0 }
```

With this, we can simply remove the log statements from `actions.getDebugInfo` (and return the Bluebird map like @edvald suggested) and from `collectSystemDiagnostic`. The info statements that is, no the errors. 

Passing the system and provider entries like this also ensures that the errors get set on the appropriate entry. 

The end result looks something like this:

```
✔ Preparing archive → Done
   ✔ system                    → collecting project configuration files → Done (took 0.3 sec)
   ✔ providers                 → collecting info → Done (took 0.4 sec)

Done! Please find your report under /Users/eysi/code/garden-io/garden/examples/simple-project.
```

garden-service/src/commands/get/get-debug-info.ts
eysi09: ...in my last comment I actually suggest skipping this all together and moving most of the logging the `action` body.

garden-service/src/commands/get/get-debug-info.ts
10ko: Yes, this is the one that should have been called in the `cli.ts` catch block.
I swapped names during my refactoring and that slipped through. Thanks for pointing out.

docs/reference/commands.md
10ko: We couldn't really agree between us on what's the preferred format (yaml vs json) for the report. Some find yaml more readable others prefer json.
Unless you have some strong opinions regarding this, it seemed a simple feature to add and we added it.

garden-service/src/plugins/kubernetes/kubernetes.ts
10ko: I didn't think about that but it looks like if the namespace doesn't exist it won't throw any exception.
Instead the output will look like the following:

```json
{
    "namespace": "trash",
    "output": {
        "apiVersion": "v1",
        "items": [],
        "kind": "List",
        "metadata": {
            "resourceVersion": "",
            "selfLink": ""
        }
    }
},
```

10ko: Is that something we are happy with?

docs/reference/commands.md
edvald: Just seemed like extra overhead somehow, but all good, no strong opinions.

garden-service/src/plugins/kubernetes/kubernetes.ts
edvald: Ah ok, wasn't sure if kubectl would balk at this. In that case it's all good.

garden-service/src/commands/get/get-debug-info.ts
10ko: You are right, it was duplicated logic and I extracted it in its own function.
Regarding the generator thingy, I thought of giving it a go but I decided against since this PR is already huge.

10ko:
Dear all, 
thanks a lot for the review, I agree on everything and wrote fixes and add-ons as you requested.

I have few notes and consideration to add:

- @eysi09 The reason the command was running twice was because the garden project you tried the command against was:
  * a valid garden project, so
  * the normal command action would get called
  * at which point the command tried to collect the providers info but, more than likely, you had docker or k8s down.
  * the function `collectProviderDebugInfo(...)` would throw an exception. 
  * The exception would then get caught in the `cli.ts` where `generateBasicDebugInfoReport(...)` would get called again: I added another try/catch in the function collecting the provider info.

- The try/catch mentioned above will catch any exception thrown by any provider. Since the actions are executed through `Bluebird.props(...)`: even if one of may providers fail, all (promises) will. It would be nice to catch only the failing providers. This would require a lot of internals work and was kinda out-of-scope for this PR. If any of you think it's something we want in the future I can open an issue.
- @edvald I found a strange 404 error regarding a dependency called `tmp-promise`: the 2.0.0 version we used just got deleted from npm and github. I updated the dependecy to 2.0.1 but maybe we should investigate a replacement or if we need it at all since it doens't seem very reliable?

Thanks again and please let me know if you have any other comment.
E.

eysi09:
I believe the `tmp-promise` issue was resolved in another PR so rebasing should remove the need for that fix.

eysi09:
One thing I noticed after approving is that the debug info is not git ignored.

PR #785, opened 2019-05-19
fix(k8s): fix type error when Kubernetes is not running

### LINKED ISSUE #784: Cannot read property 'output' of undefined when Kubernetes not running
### Current Behavior

Garden fails ungracefully with `Cannot read property 'output' of undefined` if Kubernetes cluster not running.

### Expected behavior

Should give a better error example.

### Reproducible example

Run `garden deploy` for a local env with Minikube / Docker for Desktop turned off.

### Workaround

N/A

### Suggested solution(s)

Here's the offending code: https://github.com/garden-io/garden/blob/master/garden-service/src/plugins/kubernetes/namespace.ts#L129

The `detail` key is not defined on the `err` object.

### Your environment

Latest `master`

eysi09:
Fixes #784

### DIFF
--- a/garden-service/src/plugins/kubernetes/namespace.ts
+++ b/garden-service/src/plugins/kubernetes/namespace.ts
@@ -125,18 +125,21 @@ export async function prepareNamespaces({ ctx, log }: GetEnvironmentStatusParams
     // TODO: use API instead of kubectl (I just couldn't find which API call to make)
     await kubectl.exec({ log, context: kubeContext, args: ["version"] })
   } catch (err) {
-    // TODO: catch error properly
-    if (err.detail.output) {
-      throw new DeploymentError(
-        `Unable to connect to Kubernetes cluster. ` +
-        `Please make sure it is running, reachable and that you have the right context configured.`,
-        {
-          kubeContext,
-          kubectlOutput: err.detail.output,
-        },
-      )
+    let message = err.message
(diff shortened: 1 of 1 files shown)


garden-service/src/plugins/kubernetes/namespace.ts
```diff
@@ -126,13 +126,13 @@ export async function prepareNamespaces({ ctx, log }: GetEnvironmentStatusParams
     await kubectl.exec({ log, context: kubeContext, args: ["version"] })
   } catch (err) {
     // TODO: catch error properly
-    if (err.detail.output) {
+    if (err) {
```
edvald: This will always be true...

garden-service/src/plugins/kubernetes/namespace.ts
```diff
@@ -126,13 +126,13 @@ export async function prepareNamespaces({ ctx, log }: GetEnvironmentStatusParams
...
       throw new DeploymentError(
         `Unable to connect to Kubernetes cluster. ` +
         `Please make sure it is running, reachable and that you have the right context configured.`,
         {
           kubeContext,
-          kubectlOutput: err.detail.output,
+          kubectlOutput: err.stdout,
```
edvald: Probably should include stderr here too (fine to just concatenate).

PR #789, opened 2019-05-20
fix: make global CLI opts available to commands

thsig:
Make global CLI options (such as `output`) available in the `opts` command parameter.

### DIFF
--- a/garden-service/src/cli/cli.ts
+++ b/garden-service/src/cli/cli.ts
@@ -36,7 +36,7 @@ import { Writer } from "../logger/writers/base"
 import {
   envSupportsEmoji,
   failOnInvalidOptions,
-  falsifyConflictingParams,
+  negateConflictingParams,
   filterByKeys,
   getArgSynopsis,
   getKeys,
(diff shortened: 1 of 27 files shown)


garden-service/src/cli/cli.ts
```diff
@@ -36,7 +36,7 @@ import { Writer } from "../logger/writers/base"
 import {
   envSupportsEmoji,
   failOnInvalidOptions,
-  falsifyConflictingParams,
+  negateConflictingParams,
```
eysi09: ❤️

garden-service/src/server/commands.ts
```diff
@@ -73,7 +74,8 @@ export async function resolveRequest(
   const cmdLog = log.placeholder(LogLevel.silly, { childEntriesInheritLevel: true })
 
   const cmdArgs = mapParams(ctx, request.parameters, command.arguments)
-  const cmdOpts = mapParams(ctx, request.parameters, command.options)
+  const optParams = extend(clone(GLOBAL_OPTIONS), command.options)
```
eysi09: Can't you skip the clone and do `extend(...GLOBAL_OPTIONS, command.options)`.

PR #791, opened 2019-05-22
fix(dashboard): stackgraph loses ws status when drawn

benstov:
Symptom: The dashboard stack graph resets when I click a node. For example, if a test is broken, so a node has a red border, and I click it, the color goes back to green. Same applies if I click something while some nodes are pending/processing.

Fix: merged the event status into the graph object (in graph container) being passed to the graph component(graph/index)

### DIFF
--- a/dashboard/src/components/graph/index.tsx
+++ b/dashboard/src/components/graph/index.tsx
@@ -21,17 +21,17 @@ import { colors, fontMedium } from "../../styles/variables"
 import Spinner, { SpinnerProps } from "../spinner"
 import { SelectGraphNode, StackGraphSupportedFilterKeys } from "../../context/ui"
 import { WsEventMessage, SupportedEventName } from "../../context/events"
-import { Events } from "garden-cli/src/events"
 import { Extends } from "garden-cli/src/util/util"
 import { ConfigDump } from "garden-cli/src/garden"
-import { GraphOutput } from "garden-cli/src/commands/get/get-graph"
 import { FiltersButton, Filters } from "../group-filter"
 import { RenderedNodeType } from "garden-cli/src/config-graph"
+import { GraphOutputWithNodeStatus } from "../../context/data"
 
 interface Node {
   name: string
   label: string
   id: string
+  status?: string
 }
 
 interface Edge {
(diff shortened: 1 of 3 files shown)


dashboard/src/components/graph/index.tsx
```diff
@@ -93,23 +89,28 @@ function drawChart(
...
+    .setDefaultEdgeLabel(function () {
       return {}
     })
 
-  // Here we"re setting nodeclass, which is used by our custom drawNodes function
-  // below.
-
+  // // Here we"re setting nodeclass, which is used by our custom drawNodes function
+  // // below.
+  // if (nodeEl) {
```
eysi09: Remove this.

dashboard/src/containers/graph.tsx
```diff
@@ -42,6 +42,13 @@ export default () => {
   if (!config.data || !graph.data || config.loading || graph.loading) {
     return <Spinner />
   }
+  if (message && message.type === "event") {
```
eysi09: This is a good solution! The container is exactly the one that should be responsible for preparing the data for the component in this case.

eysi09:
Also, `npm run fix-format` to fix that CI issue.

dashboard/src/components/graph/index.tsx
```diff
@@ -366,24 +364,6 @@ class Chart extends Component<Props, State> {
     }
   }
 
-  // Update the node class instead of re-rendering the graph for perf reasons
```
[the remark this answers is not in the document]
benstov: We use the function getNodeClass inside the drawChart and we only rendering it if only the graph has changed (by changed i mean changed its ref - look at graph.tsx there is this code: 
`  if (message && message.type === "event") {
    const nodeToUpdate = graph.data.nodes.find(node => node.key === (message.payload && message.payload["key"]))
    if (nodeToUpdate) {
      nodeToUpdate.status = message.name
      graph.data = {...graph.data}
    }
  }`
changing the graph.data ref will trigger the drawChart.

im not so sure what is the secret, i just think that its more logical this way and maybe the god of javascript like this more so he decided to pump up the performance xD

dashboard/src/containers/graph.tsx
benstov: Thank you very much! :)
That's exactly the thought that guided me to this solution

PR #794, opened 2019-05-24
improvement(dashboard): add graph filters to global context

benstov:
Useful for many cases such as making the filters still valid if the user leave the page and comes back


dashboard/src/context/ui.tsx
```diff
@@ -84,14 +85,27 @@ const useUiState = () => {
     })
   }
 
-  const overviewToggleItemsView = (itemToToggle: OverviewSupportedFilterKeys) => {
+  const overviewToggleItemsView = (filterKey: OverviewSupportedFilterKeys) => {
```
eysi09: Not related to this PR per se but this is getting a little messy. We should use `useReducer` instead of `useState`  in the `useUiState` hook.

PR #795, opened 2019-05-28
style(dashboard): align name and type

benstov:
overview page - 
fix: the name and type is not aligned in each card.


10ko:
Hi Ben,
Thanks for the PR. Unfortunately I am not really sure how to test this? :|

One thing I notice thou: is there a specific reason why we use `px`, `rem` and `em` in the styling?

Thanks,
Emanuele.

eysi09:
Didn't we agree on `rem` and `em` at some point? In the general case we use `rem` and we only use `em` for margins and paddings that should be relative to the font size of the element in question. Does that make sense? We should also document this somewhere.

This is also the kind of PR where it would make sense to add some before and after pictures.

PR #800, opened 2019-05-29
perf: improve performance of logs command

10ko:
The default value for the logger was specified to `fancy` which has big impact on performace.

### DIFF
--- a/garden-service/src/cli/cli.ts
+++ b/garden-service/src/cli/cli.ts
@@ -111,7 +111,6 @@ export const GLOBAL_OPTIONS = {
   "env": new EnvironmentOption(),
   "logger-type": new ChoicesParameter({
     choices: [...LOGGER_TYPES],
-    defaultValue: DEFAULT_CLI_LOGGER_TYPE,
     help: deline`
       Set logger type:
       fancy: updates log lines in-place when their status changes (e.g. when tasks complete),


eysi09:
Did this make a difference? It's possible to set the logger type at the command level and the type for the logs command is `basic`. So removing the default shouldn't make a difference unless the type set at the command level isn't propagating (which we should of course fix then).

eysi09:
It's actually an open question whether the command should overwrite the user input or vice versa. E.g.,  if the log commands sets the type to `basic`, should the user be able to demand `fancy`, even if the framework knows that that's not going to work well. IMO it's a better solution to have the framework overwrite the user input, in which case logs is always `basic`. 

If we go for that one, we should maintain the `defaultValue: DEFAULT_CLI_LOGGER_TYPE` in `GLOBAL_OPTIONS`. The logic would instead go something like: _if command has a logger type, use that one_, where the logger type is set.

thsig:
> Did this make a difference? It's possible to set the logger type at the command level and the type for the logs command is `basic`. So removing the default shouldn't make a difference unless the type set at the command level isn't propagating (which we should of course fix then).

The decision point regarding the logger type used is here:

https://github.com/garden-io/garden/blob/662ea428ab5c274dbc11be0e11d662a67d45964/garden-service/src/cli/cli.ts#L253

Since the value of the logger type option has the highest precedence, a non-empty default value there (as the logic currently stands) results in that default overriding the command's logger type. This meant that the fancy logger was being used by default in the `get logs` command.

> It's actually an open question whether the command should overwrite the user input or vice versa.

The option needs to have priority over the logger type the command specifies, e.g. so we're able to use the `json` logger type for any command during integ tests.

PR #808, opened 2019-05-31
feat(k8s): in-cluster building

edvald:
Moving this from our closed-source version. This adds an option to build
container images inside Kubernetes clusters, and by extension removes the
need for Docker and Kubernetes running on developer machines.

Use it by setting `buildMode: "cluster-docker"` in the `kubernetes` provider
configuration.


edvald:
Note: This needs a little more testing before merging. I think I see some merge-related issues that the compiler isn't catching.

edvald:
Ok, resolved the issues I could find. But it'd be great to get more hands on this, to test and make sure I didn't mess anything else up in the merge.

eysi09:
Commands always fail with:

```
Plugin local-kubernetes has been updated or hasn't been configured, and requires user input.. Please r
un `garden init` and then re-run this command.
```

Even though I've already run `garden init` for the respective env.

eysi09:
Also, `getSystemServicesStatuses` always returns:

```javascript
{
  'docker-daemon':
  {
    state: 'outdated',
      version: undefined,
        detail: { remoteObjects: [Array] }
  },
  'docker-registry':
  {
    state: 'outdated',
      version: undefined,
        detail: { remoteObjects: [Array] }
  },
  'registry-proxy':
  {
    state: 'ready',
      version: 'v-cc8c98d873',
        detail: { remoteObjects: [Array] }
  }
}
```
which results in Garden initialising the env on every command.

garden-service/src/actions.ts
```diff
@@ -158,8 +158,8 @@ export class ActionHelper implements TypeGuard {
     if (!allowUserInput && needManualInit.length > 0) {
       const names = needManualInit.map(s => s.name).join(", ")
       const msgPrefix = needManualInit.length === 1
-        ? `Plugin ${names} has been updated or hasn't been configured, and requires user input.`
-        : `Plugins ${names} have been updated or haven't been configured, and require user input.`
+        ? `Provider ${names} has been updated or hasn't been configured, and requires manual initialization.`
```
eysi09: There's a dot at the end of the prefix and where the message is printed, resulting in two dots.

garden-service/src/actions.ts
```diff
@@ -158,8 +158,8 @@ export class ActionHelper implements TypeGuard {
...
+        : `Providers ${names} have been updated or haven't been configured, and require manual initialization.`
```
eysi09: Same.

eysi09:
There's still the issue with `needsManualInit` being always set to `true` for the remote provider. As a result I never get passed: 

```
Provider kubernetes has been updated or hasn't been configured, and requires manual initialization.. Please run `garden init` and then re-run this command.
```

edvald:
Ah, I misunderstood the flow, I thought that would only kick in if the environment wasn't reported as ready. I'll fix that.

garden-service/src/plugins/kubernetes/init.ts
```diff
@@ -76,20 +77,26 @@ export async function getEnvironmentStatus({ ctx, log }: GetEnvironmentStatusPar
...
+      // same cluster, that most likely shouldn't affect usage.
+      log.warn({
+        symbol: "warning",
+        msg: chalk.yellow(
+          "One or more cluster-wide services are outdated or their configuration does not match your current " +
+          "configuration. You may want to run \`garden init\` to update them, or contact your cluster admin.",
+        ),
+      })
+    } else if (systemServiceStatuses.state !== "ready") {
+      systemReady = false
```
eysi09: Could this not be accounted for where `systemReady` is set above? Would be more readable imo.

edvald: The problem would then look like this:
```
systemReady = systemTillerReady && sysNamespaceUpToDate && (
      systemServiceStatuses.state === "ready"
      ||
      (needManualInit && systemServiceStatuses.state === "outdated")
    )
``` 
Guess that's in the eye of the beholder, I'm happy to go with either.

eysi09: It's not the prettiest of assignments but IMO the control flow is slightly more readable this way. But I can swing either way of course.

eysi09:
LGTM. Did some very basic tests but still working on running the whole thing through our integ suite, got stuck on a tangent earlier today.

PR #813, opened 2019-06-03
improvement(dashboard): add dependencies to each entity card

benstov:
Screenshot 2019-06-03 at 18 39 25


eysi09:
Thanks for including the image @benstov, it's very useful. How does it look if there are multiple dependencies?

eysi09:
My bad, didn't notice the one with multiple deps. LGTM but need to fix that format to get passed CI.

PR #817, opened 2019-06-05
improvement(dashboard): add path field to each module in overview page

benstov:
While adding the path field i noticed the structure wasn't fully clear, so i aligned the Fields > Field > Key/Value throughout all of the Module component

See updated screenshots below in the comments

### DIFF
--- a/dashboard/src/components/entity-card.tsx
+++ b/dashboard/src/components/entity-card.tsx
@@ -32,15 +32,10 @@ const EntityCard = styled.div<EntityCardProps>`
     margin-right: 0;
   }
 `
-const Header = styled.div`
-  width: 100%;
-  padding: .6rem .75rem;
-  height: 3rem;
-`
 
 const Content = styled.div`
   width: 100%;
-  padding: 0rem .75rem .75rem .75rem;
+  padding: .6rem .75rem;
   position: relative;
   max-height: 10rem;
   &:empty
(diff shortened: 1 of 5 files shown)


edvald:
This feels a little visually noisy to me. Could you maybe try reducing the font size for the path and making it a lighter grey? Also, did you check how this comes out with modules contained in remote sources?

benstov:
> This feels a little visually noisy to me. Could you maybe try reducing the font size for the path and making it a lighter grey? Also, did you check how this comes out with modules contained in remote sources?

i

benstov:
> This feels a little visually noisy to me. Could you maybe try reducing the font size for the path and making it a lighter grey?

Screenshot 2019-06-06 at 14 12 50

10ko:
Hi Ben,
Thanks for the PR.
I have few comments:
 - The path has `cursor: pointer` and it's a bit misleading, since it's not a link. It's also difficult to select for some reason.
 - There seems to be a regression on the description field: we cut the word between lines when it doesn't fit (see jworker below) 
Screenshot 2019-06-06 at 15 07 20
 - Not sure if that's necessary but [this few lines](https://github.com/garden-io/garden/blob/785f97fd04f26ed47763893aab9bdb0a870e26eb/dashboard/src/components/entity-card.tsx#L112-L116) have some confusing naming? 
  ```typescript
        {state && (
              {state}
          <State state={state}>
            </State>
            {state}
          )}
          </State>
```

PR #821, opened 2019-06-06
Cli updates

### LINKED ISSUE #668: Clean up CLI flags
Currently we a `silent` flag and a `--logger-type=quiet` flag which both do the same thing. We should probably just skip the latter. We should also consider renaming `--loglevel` to `--log-level`.

We should also use the chance to clean up some of the CLI code, make it type safe and improve the way we resolve config (those cases where we have env vars, cli flags and default values for the same thing).

eysi09:
This PR contains a handful of small CLI improvements and refactors:

- Rename the `--loglevel` flag to `--log-level` for consistency.
- Remove Global Options from CLI help output and use a dedicated command instead.
- Add log header to commands, along with necessary refactors. 

Closes #668.

### DIFF
--- a/garden-service/src/commands/exec.ts
+++ b/garden-service/src/commands/exec.ts
@@ -9,7 +9,7 @@
 import chalk from "chalk"
 import { LoggerType } from "../logger/logger"
 import { ExecInServiceResult } from "../types/plugin/service/execInService"
-import { logHeader } from "../logger/util"
+import { printHeader } from "../logger/util"
 import {
   Command,
   CommandResult,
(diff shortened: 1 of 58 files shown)


garden-service/src/commands/exec.ts
```diff
@@ -63,15 +63,17 @@ export class ExecCommand extends Command<Args> {
...
+  ): Promise<CommandResult<ExecInServiceResult>> {
     const serviceName = args.service
     const command = args.command || []
 
-    logHeader({
-      log,
-      emoji: "runner",
-      command: `Running command ${chalk.cyan(command.join(" "))} in service ${chalk.cyan(serviceName)}`,
-    })
+    printHeader(
```
edvald: This feels slightly confusing, with the ambiguous names and all. Would it make sense to make this a class method instead?

If nothing else, it feels like `headerLog` and `footerLog` would make it more clear that one is a verb and the other a noun.

PR #825, opened 2019-06-07
Init flow improvement

edvald:


garden-service/src/actions.ts
```diff
@@ -161,22 +161,17 @@ export class ActionHelper implements TypeGuard {
...
-
     const entry = log.info({ section: "providers", msg: "Getting status...", status: "active" })
     const statuses = await this.getEnvironmentStatus({ pluginName, log: entry })
 
     const needManualInit = Object.entries(statuses)
       .map(([name, status]) => ({ ...status, name }))
       .filter(status => status.ready === false && status.needManualInit === true)
 
-    if (!allowUserInput && needManualInit.length > 0) {
+    if (!init && needManualInit.length > 0) {
```
thsig: Just a thought—would `manualInit` be a clearer name than `init` here? Or alternatively, using `needInit` instead of `needManualInit`? If I'm, not mistaken, they're both referring to the same flag.

garden-service/src/types/plugin/provider/prepareEnvironment.ts
```diff
@@ -34,9 +35,11 @@ export const prepareEnvironment = {
   `,
   paramsSchema: actionParamsSchema
     .keys({
-      status: environmentStatusSchema,
       force: Joi.boolean()
         .description("Force re-configuration of the environment."),
+      init: Joi.boolean()
+        .description("Set to true if the environment is being explicitly initialized via `garden init`"),
```
thsig: Missing `.` at the end.

PR #829, opened 2019-06-10
Kaniko support

edvald:
Please review individual commits here. There are some bugfixes and refactors that were necessary to get this all working.

To test, please take any project that has a remote Kubernetes environment configured and `buildMode=cluster-docker` (such as `simple-project`), and change `buildMode` to `kaniko`.

Some implementation notes:
* Moved the build-sync-data volume to NFS, to make it accessible concurrently
  by multiple readers. This is necessary because...
* Kaniko needs to be executed individually for each build, as opposed to
  running multiple processes in the same container (as with the Docker
  Daemon).
* The rest of the build flow is the same (syncing to remote, mounting the
  sync data volume) except Kaniko itself takes care of pushing to the
  registry after building.
* Because we re-use the same flows for syncing data, I moved the sync service
  and volume to a separate service.


garden-service/test/helpers.ts
```diff
@@ -301,10 +301,20 @@ export class TestGarden extends Garden {
     public readonly opts: GardenOpts,
     plugins: Plugins,
     providerConfigs: ProviderConfig[],
+    public readonly workingCopyId: string,
   ) {
     super(
-      projectRoot, projectName, environmentName, variables, projectSources,
-      buildDir, ignorer, opts, plugins, providerConfigs,
+      projectRoot,
```
eysi09: Should we refactor this into an object?

garden-service/test/unit/src/util/util.ts
```diff
@@ -118,4 +120,24 @@ describe("util", () => {
...
     })
   })
+
+  describe("splitFirst", () => {
+    it("should split string on first occurrence of given delimiter", () => {
+      expect(splitFirst("foo:bar:boo", ":")).to.eql(["foo", "bar:boo"])
+    })
+
+    it("should return the whole string as first element when no delimiter is found in string", () => {
+      expect(splitFirst("foo", ":")).to.eql(["foo", ""])
```
eysi09: Would it make more sense to return `["foo", undefined]` when `splitFirst` can't find a delimiter?

garden-service/test/unit/src/util/util.ts
```diff
@@ -118,4 +120,24 @@ describe("util", () => {
...
+    })
+  })
+
+  describe("splitLast", () => {
+    it("should split string on last occurrence of given delimiter", () => {
+      expect(splitLast("foo:bar:boo", ":")).to.eql(["foo:bar", "boo"])
+    })
+
+    it("should return the whole string as last element when no delimiter is found in string", () => {
+      expect(splitLast("foo", ":")).to.eql(["", "foo"])
```
eysi09: Same as above, `undefined` as opposed to an empty string.

garden-service/test/unit/src/util/util.ts
eysi09: So that `splitFirst("foo", ":") !== splitFirst("foo:", ":")`

garden-service/test/helpers.ts
edvald: Yes, definitely. I considered doing it here, but knew it would conflict with other PRs. Maybe the last ready PR can go ahead and do that, or we do it separately.

garden-service/test/unit/src/util/util.ts
edvald: I considered that, but it would complicate the signature so much. You'd always need to if-statement the result somehow. Plus I didn't want to change the existing behavior (I just added tests to make sure the function was doing what I expected).

garden-service/test/unit/src/util/util.ts
edvald: Same as above ^

examples/simple-project/garden.yml
```diff
@@ -11,4 +11,4 @@ project:
           context: gke_garden-dev-200012_europe-west1-b_garden-dev-1
           namespace: ${local.env.USER || "default"}-simple-project
           defaultHostname: ${local.env.USER || "default"}-simple-project.dev-1.sys.garden
-          buildMode: cluster-docker
+          buildMode: kaniko
```
eysi09: Shouldn't we default to `cluster-docker`?

edvald: Ah yes, we should, forgot that in there.

garden-service/src/plugins/kubernetes/config.ts
```diff
@@ -202,15 +220,20 @@ const tlsCertificateSchema = Joi.object()
...
+      .allow("local", "cluster-docker", "kaniko")
       .default("local")
       .description(deline`
         Choose the mechanism used to build containers before deploying. By default it uses the local docker, but you
-        can set it to 'cluster-docker' to sync files to a remote docker daemon, installed in the cluster, and build
-        container images there.
+        can set it to 'cluster-docker' or 'kaniko' to sync files to a remote docker daemon, installed in the cluster,
+        and build container images there.
 
         This is currently experimental and sometimes not needed (e.g. with Docker for Desktop), so it's not enabled
```
eysi09: Not a part of this PR but the "sometimes not needed" is a little hazy. Can we be a clearer on the when and why?

garden-service/src/plugins/kubernetes/config.ts
```diff
@@ -228,21 +251,23 @@ export const kubernetesConfigBase = providerConfigBaseSchema
       .keys({
         builder: storageSchema(defaultStorage.builder),
         registry: storageSchema(defaultStorage.registry),
+        sync: storageSchema(defaultStorage.registry),
```
eysi09: Shouldn't this be `defaultStorage.sync`?

garden-service/src/plugins/kubernetes/config.ts
edvald: Sure, I could add some color. In particular, it's not needed at all when the docker daemon is already part of the K8s cluster, as with Docker for Desktop and Minikube. It may also not be preferred if you're just deploying to a cluster, and not using it for development. I'll update the doc.

garden-service/src/plugins/kubernetes/config.ts
edvald: Good catch, yes!

eysi09: ...although it produces the same values.

garden-service/static/kubernetes/system/build-sync/garden.yml
```diff
@@ -0,0 +1,18 @@
+kind: Module
+type: helm
+name: build-sync
+description: Docker daemon used for in-cluster building
```
eysi09: Can we add a better description here?

edvald: We sure can.

eysi09:
Looks good from my end. The `jworker` module of the `vote` example won't build though, but this is likely an issue with Kaniko. We should create an issue for this though.

@thsig, could you do a quick sanity check as well once you've completed your review.

eysi09:
Integration tests passed on second try. First try failed with `Provider kubernetes has been updated or hasn't been configured, and requires manual initialization. Please run `garden init` and then re-run this command.`. See here: https://circleci.com/gh/garden-io/garden/12207

Note that the `init` command is run as a part of the integration tests. It passed the second time without any changes to the code or tests.

PR #835, opened 2019-06-13
feat(core): allow .yaml endings for Garden config files

### LINKED ISSUE #562: Allow .yaml ending for Garden config files
Currently, the project and module file names _must_ be `garden.yml`. We should allow `garden.yaml` as well.

eysi09:
With this PR, users can now name their Garden config files either `garden.yml` or `garden.yaml`.

Closes #562

### DIFF
--- a/garden-service/src/garden.ts
+++ b/garden-service/src/garden.ts
@@ -7,7 +7,7 @@
  */
 
 import Bluebird = require("bluebird")
-import { parse, relative, resolve, sep, join } from "path"
+import { parse, relative, resolve, sep } from "path"
 import { flatten, isString, cloneDeep, sortBy, set, zip } from "lodash"
 const AsyncLock = require("async-lock")
 
@@ -35,12 +35,12 @@ import { BuildDependencyConfig, ModuleConfig, baseModuleSpecSchema, ModuleResour
 import { ModuleConfigContext, ContextResolveOpts } from "./config/config-context"
 import { createPluginContext } from "./plugin-context"
 import { ModuleAndRuntimeActions, Plugins, RegisterPluginParam } from "./types/plugin/plugin"
-import { SUPPORTED_PLATFORMS, SupportedPlatform, CONFIG_FILENAME, DEFAULT_GARDEN_DIR_NAME } from "./constants"
+import { SUPPORTED_PLATFORMS, SupportedPlatform, DEFAULT_GARDEN_DIR_NAME } from "./constants"
 import { platform, arch } from "os"
 import { LogEntry } from "./logger/log-entry"
 import { EventBus } from "./events"
 import { Watcher } from "./watch"
-import { getIgnorer, Ignorer, getModulesPathsFromPath } from "./util/fs"
+import { getIgnorer, Ignorer, getModulesPathsFromPath, getConfigFilePath } from "./util/fs"
 import { Provider, ProviderConfig, getProviderDependencies } from "./config/provider"
 import { ResolveProviderTask } from "./tasks/resolve-provider"
 import { ActionHelper } from "./actions"
@@ -407,13 +407,13 @@ export class Garden {
 
       this.resolvedProviders = Object.values(taskResults).map(result => result.output)
 
-      for (const provider of this.resolvedProviders) {
-        for (const moduleConfig of provider.moduleConfigs) {
+      await Bluebird.map(this.resolvedProviders, async (provider) =>
+        Bluebird.map(provider.moduleConfigs, async (moduleConfig) => {
           // Make sure module and all nested entities are scoped to the plugin
           moduleConfig.plugin = provider.name
-          this.addModule(moduleConfig)
-        }
-      }
+          return this.addModule(moduleConfig)
+        }),
+      )
     })
 
     return this.resolvedProviders
@@ -622,9 +622,7 @@ export class Garden {
         }
       })
 
-      for (const config of rawConfigs) {
-        this.addModule(config)
-      }
+      await Bluebird.map(rawConfigs, async (config) => this.addModule(config))
 
       this.modulesScanned = true
     })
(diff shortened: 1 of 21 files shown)


garden-service/src/util/fs.ts
```diff
@@ -50,6 +53,40 @@ export async function* scanDirectory(path: string, opts?: klaw.Options): AsyncIt
...
+/**
+ * Returns the expected full path to the Garden config filename.
+ *
+ * If a valid config filename is found at the given path, it returns the full path to it.
+ * If no config file is found, it returns the path joined with the first value from the VALID_CONFIG_FILENAMES list.
+ * (The check for whether or not a project or a module has a valid config file at all is handled elsewehere.)
+ *
+ * Throws an error if there are more than one valid config filenames at the given path.
+ */
+export async function getConfigFilePath(path: string) {
```
eysi09: I kind of debated whether we should have a `getConfigFilename` function instead since in some of the uses cases it's not certain that there is actually a config file at the given path.

So having a `getConfigFilePath()` function that returns a default path that actually doesn't exist feels kind of weird. At the same time, in almost all use cases we end up creating that path any way. So if we go for the `getConfigFilename` version we usually just end up with something like:

```javascript
const configPath = join(path, await getConfigFilename(path)) // may or may not exists
``` 
instead of just:
```javascript
const configPath = await getConfigFilePath(path) // may or may not exist
```
as in the current implementation.

garden-service/src/garden.ts
```diff
@@ -623,7 +623,7 @@ export class Garden {
       })
 
       for (const config of rawConfigs) {
-        this.addModule(config)
+        await this.addModule(config)
```
edvald: This should be a Bluebird.map, otherwise we await each promise sequentially.

garden-service/src/garden.ts
```diff
@@ -411,7 +411,7 @@ export class Garden {
         for (const moduleConfig of provider.moduleConfigs) {
           // Make sure module and all nested entities are scoped to the plugin
           moduleConfig.plugin = provider.name
-          this.addModule(moduleConfig)
+          await this.addModule(moduleConfig)
```
edvald: This should be a Bluebird.map as well.

garden-service/src/util/fs.ts
```diff
@@ -50,6 +53,40 @@ export async function* scanDirectory(path: string, opts?: klaw.Options): AsyncIt
...
+  const configFilePaths = await Bluebird
+    .map(VALID_CONFIG_FILENAMES, async filename => {
```
edvald: I think we almost always use parentheses around parameters when we use the async keyword. I think that's preferable at least.

garden-service/src/util/fs.ts
```diff
@@ -50,6 +53,40 @@ export async function* scanDirectory(path: string, opts?: klaw.Options): AsyncIt
...
+      const configFilePath = join(path, filename)
+      return await pathExists(configFilePath) ? configFilePath : undefined
```
edvald: For clarity, parentheses around the pathExists call would be good

garden-service/src/util/fs.ts
eysi09: Ah ok. I always skip it when I can. I'll see if we can add lint rule for it.

garden-service/src/garden.ts
eysi09: Whoops, didn't notice. Good catch.

garden-service/src/util/fs.ts
eysi09: I'll fix this here and add a separate PR with lint rule + appropriate fixes.

garden-service/src/garden.ts
eysi09: Can I do a `Bluebird.map` for the outer loop as well?

garden-service/src/util/fs.ts
edvald: This seems fine to me.

garden-service/src/garden.ts
edvald: That should be fine, yeah.

garden-service/src/garden.ts
```diff
@@ -641,13 +641,13 @@ export class Garden {
...
-  private addModule(config: ModuleConfig) {
+  private async addModule(config: ModuleConfig) {
     const key = getModuleKey(config.name, config.plugin)
 
     if (this.moduleConfigs[key]) {
       const [pathA, pathB] = [
-        relative(this.projectRoot, join(this.moduleConfigs[key].path, CONFIG_FILENAME)),
-        relative(this.projectRoot, join(config.path, CONFIG_FILENAME)),
+        relative(this.projectRoot, await getConfigFilePath(this.moduleConfigs[key].path)),
+        relative(this.projectRoot, await getConfigFilePath(config.path)),
```
edvald: This could also be done in parallel, although it probably doesn't matter a lot.

eysi09: Added a new commit for this. Will squash on approval.

garden-service/src/garden.ts
```diff
@@ -407,13 +407,13 @@ export class Garden {
 
       this.resolvedProviders = Object.values(taskResults).map(result => result.output)
 
-      for (const provider of this.resolvedProviders) {
-        for (const moduleConfig of provider.moduleConfigs) {
+      await Bluebird.map(this.resolvedProviders, async (provider) => {
+        await Bluebird.map(provider.moduleConfigs, async (moduleConfig) => {
```
edvald: You can shorten/simplify this to
```
      await Bluebird.map(this.resolvedProviders, (provider) => 
        Bluebird.map(provider.moduleConfigs, async (moduleConfig) => {
```
Plus, and I always forget this, I suppose `Bluebird.all` is the appropriate function here since we don't care about the output.

garden-service/src/garden.ts
```diff
@@ -622,9 +622,9 @@ export class Garden {
         }
       })
 
-      for (const config of rawConfigs) {
+      await Bluebird.map(rawConfigs, async (config) => {
         await this.addModule(config)
```
edvald: Can simplify this one to `await Bluebird.all(config => this.addModule(config))`

garden-service/src/garden.ts
eysi09: Ah yes. I vaguely remembered us talking about there not being a Bluebird method for that some reason. But yes, we shouldn't map if there's no return value.

PR #836, opened 2019-06-13
feat(container): add command option

thsig:
Added `command` and `hotReloadCommand` fields to container module configs to go with `args` and `hotReloadArgs`  (on services, test configs and tasks).

For the k8s plugins, `command` corresponds to the `command` field on k8s podspecs, i.e. it's the "entrypoint" which `args` are passed to.

This is useful e.g. when some other entrypoint than `/bin/sh -c` is desired.

Closes https://github.com/garden-io/garden/issues/401.

### DIFF
--- a/garden-service/src/commands/run/module.ts
+++ b/garden-service/src/commands/run/module.ts
@@ -85,8 +85,6 @@ export class RunModuleCommand extends Command<Args, Opts> {
     const buildTask = new BuildTask({ garden, log, module, force: opts["force-build"] })
     await garden.processTasks([buildTask])
 
-    const command = args.command || []
-
     const runtimeContext = await runtimeContextForServiceDeps(garden, graph, module)
 
     printRuntimeContext(log, runtimeContext)
@@ -96,7 +94,7 @@ export class RunModuleCommand extends Command<Args, Opts> {
     const result = await actions.runModule({
       log,
       module,
-      command,
+      args: args.command || [],
       runtimeContext,
       interactive: opts.interactive,
     })
--- a/garden-service/src/plugins/kubernetes/run.ts
+++ b/garden-service/src/plugins/kubernetes/run.ts
@@ -11,6 +11,7 @@ import { kubectl } from "./kubectl"
 import { PrimitiveMap } from "../../config/common"
 import { Module } from "../../types/module"
 import { LogEntry } from "../../logger/log-entry"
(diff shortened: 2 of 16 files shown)


garden-service/test/helpers.ts
```diff
@@ -190,14 +191,15 @@ export const testPlugin: PluginFactory = (): GardenPlugin => {
...
+        ): Promise<RunTaskResult> {
           const result = await runModule({
             ctx,
             interactive,
             log,
             runtimeContext,
             module: task.module,
-            command: task.spec.command || [],
+            args: task.spec.command,
+            // command: task.spec.command,
```
eysi09: Remove comment.

eysi09:
LGTM except that one comment I'm assuming was left by accident.

garden-service/src/commands/run/module.ts
```diff
@@ -96,7 +94,7 @@ export class RunModuleCommand extends Command<Args, Opts> {
     const result = await actions.runModule({
       log,
       module,
-      command,
+      args: args.command || [],
```
edvald: This strikes me as odd. The behavior was, if anything, more natural before. I'd suggest adding a `--command` flag to this command, and keeping the positionals as `args`.

garden-service/src/plugins/kubernetes/run.ts
```diff
@@ -56,12 +62,11 @@ export async function runPod(
     ...opts,
     ...envArgs,
     "--",
-    "/bin/sh",
-    "-c",
-    commandStr,
+    ...cmd,
+    args.join(" "),
```
edvald: This needs to look different. 

First off, we should only add `--command` to the `opts` array above if a command is specified when calling the function. That flag indicates that the supplied arguments (the ones after `--`) should be regarded as a command/entrypoint as opposed to args to the default command.

We should also not supply a default for command (`defaultContainerCommand`)  and rather take care of that on the calling side, otherwise we're unable to use the container's default entrypoint.

Lastly, we should not join `args` into a single string. You should really only do that if the args should be passed to `/bin/sh -c`, which is something the caller should decide.

edvald:
This was merged prematurely, and needs to be reverted or at least fixed before we cut the next release. See comments above. There are btw many more points where the "container" term is used at the framework level, which needs to be fixed.

PR #837, opened 2019-06-13
fix(cli): ensure cli exits with code 0 when help/version called

eysi09:
This was causing our pre-commit integ tests to fail.

### DIFF
--- a/garden-service/src/cli/cli.ts
+++ b/garden-service/src/cli/cli.ts
@@ -84,15 +84,21 @@ export const MOCK_CONFIG: ProjectConfig = {
   variables: {},
 }
 
-export const GLOBAL_OPTIONS = {
-  "version": new StringParameter({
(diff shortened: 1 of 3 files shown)


garden-service/src/cli/cli.ts
```diff
@@ -84,15 +84,21 @@ export const MOCK_CONFIG: ProjectConfig = {
   variables: {},
 }
 
-export const GLOBAL_OPTIONS = {
-  "version": new StringParameter({
+// The help text for these commands is only displayed when calling `garden options`.
+// However, we can't include them with the global options since that causes the CLI
+// to exit with code 1 when they're called.
+export const HIDDEN_COMMANDS = {
```
edvald: Should this be `HIDDEN_OPTIONS`?

eysi09: Yeah good point.

PR #838, opened 2019-06-13
chore: updated bug report template

10ko:
Updated the Bug Report template to include the `garden get debug-info` command.

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.md
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.md
@@ -37,6 +37,5 @@ about: Create a report to help make garden better
 <!-- Please run and copy and paste the results  -->
 `garden version`
 
-`kubectl version`
-
-`docker version`
+<!-- Please run the "get debug-info" command and attach the output zip file -->
+`garden get debug-info`


.github/ISSUE_TEMPLATE/BUG_REPORT.md
```diff
@@ -40,3 +40,6 @@ about: Create a report to help make garden better
 `kubectl version`
 
 `docker version`
+
+<!-- Please run the get debug-info command and attach the output zip file -->
```
eysi09: Could you add quotes or back ticks around the "get debug-info" part?

.github/ISSUE_TEMPLATE/BUG_REPORT.md
```diff
@@ -40,3 +40,6 @@ about: Create a report to help make garden better
 `kubectl version`
 
 `docker version`
```
edvald: I think we should also remove both the docker and kubectl version checks. They're no longer needed.

eysi09:
I went ahead and removed the `kubectl` and `docker` version requests from the template.

thsig:
Merged.

PR #839, opened 2019-06-13
fix: revisions to `command` option

thsig:


garden-service/src/commands/run/module.ts
```diff
@@ -17,18 +17,18 @@ import {
...
 const runArgs = {
   module: new StringParameter({
     help: "The name of the module to run.",
     required: true,
   }),
   // TODO: make this a variadic arg
-  command: new StringsParameter({
-    help: "The command to run in the module.",
+  arguments: new StringsParameter({
+    help: deline`The command to run in the module. Example: 'npm run my-script'.`,
```
edvald: No need for `deline` here. Also, I'd suggest "The arguments to run the module with. ..."

garden-service/src/commands/run/module.ts
```diff
@@ -42,6 +42,12 @@ const runOpts = {
     cliOnly: true,
   }),
   "force-build": new BooleanParameter({ help: "Force rebuild of module before running." }),
+  "command": new StringsParameter({
+    help: deline`The entrypoint of the command to run in the module. For container modules, this overrides the
```
edvald: How about "The base command (aka entrypoint) to run in the module. For container modules, for example, ..."?

garden-service/src/commands/run/module.ts
```diff
@@ -73,8 +79,8 @@ export class RunModuleCommand extends Command<Args, Opts> {
     const graph = await garden.getConfigGraph()
     const module = await graph.getModule(moduleName)
 
-    const msg = args.command
-      ? `Running command ${chalk.white(args.command.join(" "))} in module ${chalk.white(moduleName)}`
+    const msg = args.arguments
+      ? `Running command ${chalk.white(args.arguments.join(" "))} in module ${chalk.white(moduleName)}`
```
edvald: I'd suggest rephrasing: "Running module <name> with arguments <args>"

garden-service/test/unit/src/commands/run/module.ts
```diff
@@ -52,7 +52,7 @@ describe("RunModuleCommand", () => {
       log,
       headerLog: log,
       footerLog: log,
-      args: { module: "module-a", command: ["my", "command"] },
+      args: { module: "module-a", arguments: ["my", "command"] },
```
edvald: Would be great to add a quick test for the command flag.

garden-service/src/plugins/kubernetes/run.ts
```diff
@@ -57,13 +55,17 @@ export async function runPod(
...
+  }
+
   const kubecmd = [
     "run", `run-${module.name}-${Math.round(new Date().getTime())}`,
     ...opts,
     ...envArgs,
     "--",
     ...cmd,
-    args.join(" "),
+    ...args,
```
edvald: This all looks good _but_ we almost definitely need to update various call sites and have them explicitly set `["/bin/sh", "-c"]` as the `command`.

PR #850, opened 2019-06-15
docs: add missing links to summary

eysi09:
If a link is not included in the "SUMMARY" file, it won't work correctly
and will point to Github instead of docs.garden.io. Our "SUMMARY" file
is the docs/README.md file. This can be configured in the
`.gitbook.yaml` config file.


edvald:
Ah, that explains that. We should maybe update it automatically, I can look into that.

eysi09:
Yes, we should definitely automate this. It should be straightforward, just didn't have the time.

PR #854, opened 2019-06-17
Fix K8s status issues

edvald:


garden-service/src/plugins/kubernetes/status.ts
```diff
@@ -471,14 +478,11 @@ export async function compareDeployedObjects(
       // Exited with non-zero code. Check for error messages on stderr. If one is there, the command was unable to
       // complete the check, so we fall back to our own mechanism. Otherwise the command worked, but one or more
       // resources are missing or outdated.
-      if (
-        !err.detail || !err.detail.result
-        || (!!err.detail.result.stderr && err.detail.result.stderr.trim() !== "exit status 1")
```
thsig: Off-topic: Lodash's `get` is a nice convenience method for this sort of thing (https://lodash.com/docs/4.17.11#get).

This conditional would have been
```typescript
if (get(err, ["detail", "result", "stderr"], "").trim() !=== "exit status 1") {
```

PR #866, opened 2019-06-21
fix(core): chokidar watcher on mac could segfault after reloading config

edvald:
Turns out chokidar and the native fsevents module don't like it when multiple
instances are created in a single process, even if one is closed before the
other starts. Fixed it by making a global instance and reconfiguring on
reload instead. See https://github.com/fsevents/fsevents/issues/273 for
a discussion on this issue.


edvald:
Tests are hanging now after they run, because the watcher keeps running. Need to work out how to fix that.

PR #867, opened 2019-06-22
Template string improvements

### LINKED ISSUE #551: Template strings: Fixnums or floats with many decimals are rendered in exponential notation.
This PR https://github.com/garden-io/garden/pull/545 fixed an issue where module versions used in template strings were rendered as numbers in exponential format (instead of a standard string of decimals) if the first several characters in the commit hash were numbers.

To illustrate, here's an example from a failed CI run (from before https://github.com/garden-io/garden/pull/545 was merged):

screenshot 2019-02-18 at 14 06 22

Practically, this may only be a problem for versions / commit hashes, but we should keep an eye open for other bugs of this kind, in case there are other template-string related edge cases we haven't tackled yet.

edvald:
Look at individual commits for details. This addresses some issues with expected behavior, improves documentation a bit and paves the way for some features that would rely on templating e.g. booleans and numbers into configs.

Closes #551

### DIFF
--- a/docs/reference/template-strings.md
+++ b/docs/reference/template-strings.md
@@ -3,12 +3,13 @@
 Below you'll find the schema for the keys available when interpolating template strings (see our
 [Configuration Files](../using-garden/configuration-files.md) guide for information and usage examples).
 
-Note that there are two sections below, because Project configs and Module configs have different keys available to
-them. Please make sure to refer to the correct section.
+Note that there are three sections below, because Project configs and Module configs have different keys available to
+them, and additional keys are available under `providers` in Project configs.
+Please make sure to refer to the correct section.
 
 ## Project configuration context
 
-The following keys are available in template strings under the `project` key in `garden.yml` files:
+The following keys are available in template strings anywhere in Project `garden.yml` config files:
 
 ```yaml
 # Type: object
@@ -29,11 +30,20 @@ local:
   # Example: "posix"
   #
   platform:
+
+  # The current username (as resolved by https://github.com/sindresorhus/username)
+  #
+  # Type: string
+  #
+  # Example: "tenzing_norgay"
+  #
+  username:
 ```
 
-## Module configuration context
+## Provider configuration context
 
-The following keys are available in template strings under the `module` key in `garden.yml` files:
+The following keys are available in template strings under the `providers` key (or `environments[].providers)
+in Project `garden.yml` config files:
 
 ```yaml
 # Type: object
@@ -54,6 +64,14 @@ local:
   # Example: "posix"
   #
   platform:
+
+  # The current username (as resolved by https://github.com/sindresorhus/username)
+  #
+  # Type: string
+  #
+  # Example: "tenzing_norgay"
+  #
+  username:
 
 # Information about the environment that Garden is running against.
 #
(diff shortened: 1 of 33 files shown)


garden-service/test/unit/src/template-string.ts
```diff
@@ -209,12 +263,20 @@ describe("resolveTemplateString", async () => {
     expect(res).to.equal("foo")
   })
 
+  it("should handle chained conditional with string fallback", async () => {
```
thsig: "handle chained" -> "handle a chained"?

garden-service/src/template-string.ts
```diff
@@ -48,11 +52,21 @@ export async function resolveTemplateString(string: string, context: ConfigConte
...
   })
 
-  const resolved = await Bluebird.all(parsed)
-  return resolved.join("")
+  const resolved: (Primitive | undefined)[] = await Bluebird.all(parsed)
+
+  const result = resolved.length === 1
+    // Return value directly if there is only one value in the output
+    ? resolved[0]
+    // Else join together all the parts as a string. Output null as a literal string and not an empty string.
```
thsig: "join together all the parts" -> "join all the parts together"?

thsig: "and not an empty string" -> "and not as an empty string"?

garden-service/src/config/config-context.ts
```diff
@@ -173,10 +174,26 @@ class LocalContext extends ConfigContext {
...
+  )
+  public username: () => Promise<string>
+
   constructor(root: ConfigContext) {
     super(root)
     this.env = process.env
     this.platform = process.platform
+    this.username = async () => {
+      const name = await username()
+      if (name === undefined) {
```
thsig: Should this be `if (!name) {` to catch `null` as well? Or is that unnecessary here?

edvald: That's unnecessary here. The function will either return a string or undefined.

garden-service/src/template-string.ts
edvald: This is a code comment. No need to poke at the grammar .)

docs/using-garden/configuration-files.md
```diff
@@ -313,10 +316,124 @@ tests:
...
+    namespace: ${local.env.CI_BRANCH || local.username || "default"}
+```
+
+This allows you to easily set default values when certain template keys are not available, and to configure your
+project based on varying context.
+
+### Numbers, booleans and null values
+
+When a template string key resolves to a number, boolean or null, the output is handled in two different ways,
+depending on whether the template string is part of surrounding string or not.
```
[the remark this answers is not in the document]
edvald: I don't think that's any clearer. But there is a missing article there, I'll fix that.

docs/using-garden/configuration-files.md
```diff
@@ -313,10 +316,124 @@ tests:
...
+    log-level: ${local.env.LOG_LEVEL || "info"}
+    namespace: ${local.env.CI_BRANCH || local.username || "default"}
+```
+
+This allows you to easily set default values when certain template keys are not available, and to configure your
+project based on varying context.
+
+### Numbers, booleans and null values
+
+When a template string key resolves to a number, boolean or null, the output is handled in two different ways,
```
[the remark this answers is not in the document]
edvald: Why?

garden-service/src/template-string.ts
thsig: Haha, no indeed—got a bit carried away there.

docs/reference/template-strings.md
```diff
@@ -1,53 +1,53 @@
...
-services:
-  - name: some-service
-    # ...
-    env:
-      OTHER_MODULE_VERSION: ${modules.other-module.version}
-```
-
-You can also inject a template variable into a string. For instance, you might need to include a module's
-version as part of a URI:
+The following keys are available in template strings anywhere in Project `garden.yml` config files:
```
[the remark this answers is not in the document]
edvald: This is meant to distinguish this section from the next section below, I'll clarify the language a bit.

PR #879, opened 2019-06-25
improvement: refactor dashboard data store

benstov:


edvald:
This needs to be rebased for it to be easily reviewable, lots of unrelated diffs to sift through :P

PR #880, opened 2019-06-25
fix(k8s): error when test+task result log exceeded 1MB

edvald:
This comes down to a limitation of K8s ConfigMap resources. I've
refactored that code a bit and ensured we limit the log size we store.

I also made sure the ConfigMap keys didn't grow too long by hashing
the key parameters.

Lastly I made the results more easily searchable by adding labels.


garden-service/src/util/string.ts
```diff
@@ -16,8 +16,34 @@ export const deline = _deline
...
+ *
+ * Optionally scan towards the next line break after trimming the bytes, and trim to there.
+ *
+ * Note that a UTF-8 character can be 1-4 bytes so this is a naive but inexpensive approach.
+ */
+export function tailString(str: string, maxLength: number, nextLine = false) {
+  const overflow = Buffer.byteLength(str, "utf8") - maxLength
+  if (overflow > 0) {
+    if (nextLine) {
+      const lineBreakIdx = str.indexOf("\n", overflow)
```
thsig: We don't have to consider any other line break sequences than `\n` here, right?

edvald: Nope, this implicitly handles CRLF as well.

PR #881, opened 2019-06-26
feat(k8s): add mechanism for cleaning up unused images in clusters

edvald:
This required some mechanism for plugins to expose commands. We should
later iterate on that, but it's a start.

The cleanup flow and command itself is described in the
In-Cluster Building guide.


docs/using-garden/in-cluster-building.md
```diff
@@ -0,0 +1,143 @@
...
+```yaml
+kind: Module
+name: my-module
+image: my-repo/my-image:v1.2.3   # <- omit the tag here if you'd like to use the Garden-generated version tag
+...
+```
+
+## Cleaning up cached images
+
+In order to avoid disk-space issues in the cluster, the `kubernetes` providers exposes a utility command:
```
thsig: "providers exposes" -> "providers expose"?

edvald: _ps. You don't need to put a question mark after every language suggestion .)_

thsig: Haha. Noted : )

PR #884, opened 2019-06-26
fix(k8s): don't store full version object with test+task results

edvald:
This addresses issues where the list of files, contained in the module
version object, would push the size of the result ConfigMap over 1MB.


garden-service/src/plugins/kubernetes/task-results.ts
```diff
@@ -83,6 +90,8 @@ export async function storeTaskResult(
       [gardenAnnotationKey("moduleVersion")]: module.version.versionString,
       [gardenAnnotationKey("version")]: taskVersion.versionString,
     },
-    data: result,
+    data: {
```
10ko: Why `data: {}`?

10ko:
Tested and fixes my issues, even when running tests *with* `node_modules` outside the container and *without* a `.gardenignore`. 
Aside the question above, LGTM. Thanks!

garden-service/src/plugins/kubernetes/task-results.ts
edvald: That's a mistake, I'll fix it.

edvald:
@10ko fixed that line. Can you give a stamp of approval? :)

PR #889, opened 2019-06-27
improvement(k8s): add explicit cluster-init command for remote clusters

edvald:
This avoids complications relating to differing provider configurations
and Garden versions, by removing cluster-wide init from the standard
`garden init` flow and making it an explicit plugin command.

`local-kubernetes` environments are not affected by this, since they are
not shared between users/CI/etc.

### DIFF
--- a/garden-service/src/plugins/kubernetes/init.ts
+++ b/garden-service/src/plugins/kubernetes/init.ts
@@ -6,9 +6,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import { KubeApi } from "./api"
+import { KubeApi, KubernetesError } from "./api"
 import { getAppNamespace, prepareNamespaces, deleteNamespaces } from "./namespace"
-import { KubernetesPluginContext, KubernetesConfig, KubernetesProvider } from "./config"
+import { KubernetesPluginContext, KubernetesConfig } from "./config"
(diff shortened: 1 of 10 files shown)


garden-service/src/plugins/kubernetes/init.ts
```diff
@@ -119,48 +113,81 @@ function checkManualInit(provider: KubernetesProvider) {
...
-
-    // Install system services
-    await prepareSystemServices({
-      log,
-      sysGarden,
-      namespace,
-      force,
-      ctx: k8sCtx,
-      serviceNames: systemServiceNames,
+  const serviceStatuses: ServiceStatus[] = status.detail && status.detail.serviceStatuses
```
thsig: Is this expression missing a default value for when `status.detail` is blank? Since there's an `&&` here.

garden-service/src/plugins/kubernetes/init.ts
```diff
@@ -119,48 +113,81 @@ function checkManualInit(provider: KubernetesProvider) {
...
+  const serviceStates = Object.values(serviceStatuses).map(s => s.state!)
```
thsig: Calling `Object.values` on an array?

edvald: Actually the type annotation is wrong, I'll fix it.

Issue #907, opened 2019-06-28
sudo: garden: command not found


marcus-sa:
### Current Behavior
Unable to run `garden get status` as instructed in your Quick Start guide due to you not supporting sudo:

```sh
$ sudo garden get status
sudo: garden: command not found
```

sudo is required when using minikube with no drivers https://github.com/kubernetes/minikube/issues/3760 

### Expected behavior
Being able to run `garden` as sudo user.

### Environment

`garden version`: 0.10.0

edvald:
Hey @marcus-sa. I believe the issue is not that Garden can't be run as sudo, but rather that `sudo garden` can't find the `garden` binary. If you run `sudo <full-path-to-garden>` or symlink `/usr/local/bin/garden` to where you have Garden installed, it should work fine.

I'm curious though, why do you need to run Garden with sudo?

marcus-sa:
@edvald hey you're probably right.
As I wrote, starting minikube with no driver's lovally requires sudo explicitly for accessing its config file in ones home directory.
Even though I ran `chmod -R 777 .` on it, unfortunately still failed.

This happens because
https://github.com/garden-io/garden/blob/ed0233b7b59d25686f9e1ca8216a71543974004b/garden-service/src/plugins/kubernetes/local/minikube.ts#L16
Isn't supposed in a local environment with no driver's specified, even though it uses docker as the container.

See these issues:
https://github.com/kubernetes/minikube/issues/2443
https://github.com/GoogleContainerTools/skaffold/issues/112

edvald:
Ah, interesting. I guess we'd need to explicitly catch that error message or otherwise check which VM driver is in use. Should be easy to do, I'll flag it for our next update.

eysi09:
Fixed with #910 (assuming the original problem with sudo was a none-issue).

ghostnet2000:
cd to where you have Garden installed, and run chmod +x garden

PR #910, opened 2019-06-29
fix(k8s): skip setMinikubeDockerEnv when vm-driver=None

dev-kpyc:
https://github.com/garden-io/garden/issues/907


edvald:
Thanks @dev-kpyc! Looks good. integ tests are failing for unrelated reasons btw, seems to be for all external contributors.

edvald:
Oops, spoke too soon. This will actually fail if you start minikube without a `--vm-driver` flag: `minikube config get vm-driver` exits with code 1 and a `Error: specified key could not be found in config` message. We'd need to explicitly catch that error, then it should be all good.

edvald:
Looks good @dev-kpyc. The integ test failures are unrelated to this. I'll do a quick manual check and then merge.

Issue #851, opened 2019-06-17
Allow fine-grained test management with 'test sets'


ellenkorbes:
As originally mentioned in #438, users might not want to run _all_ tests _all_ the time.

I suggest we allow the user to define named sets of tests to run on different scenarios. Some ideas:

- A `default` test set defines which tests run on every file change, when `garden dev` is running.
- If `default` is undefined, it contains all tests.
- An `all` test set is hard-coded to include all tests.
- A `--testset` flag allows the user to manually specify which test set to use.

If we're feeling fancy, we could allow test sets to be test sets, which should give more granular control—but could also be completely useless. 

For example, let's say I have three tests, `t1`, `t2`, and `t3`. I define a test set called `mytestset` containing those three tests. I can then specify `default: mytestset`.

OR I have three tests, `t1`, `t2`, and `t3`. I define a test set called `mytestset` containing tests `t1` and `t2`. I can then specify `default: mytestset, t3`.

Regardless of these suggestions, _any_ way to differentiate between tests to run on `garden dev` and tests to run in CI would solve the problem.

edvald:
So, we already have names on tests. And `garden test` allows you to filter by those, but not currently `garden dev`. Adding a `--test-name` flag to the dev command would be pretty easy.

Then there's the notion of test groups/sets, which could be quite nice. Basically we could have a `group` field on test specs, which defaults to "default", and then `--group/--test-group` field on the test+dev commands respectively.

How does that sound?

ellenkorbes:
Sounds good.

Do we prefer to list the groups a test belongs to when defining the tests, or would we rather have a groups field and list the tests there per group?

```
tests:
  - groups:
    - myGroup: test1, test2, test3
  - name: test1
    ...
  - name: test2
    ...
  - name: test3
    ...
```

or

```
tests:
  - name: test1
    groups: myGroup
    ...
  - name: test2
    groups: myGroup
    ...
  - name: test3
    groups: myGroup
    ...
```

?

10ko:
So as far as I am concerned the list of things we should implement/update is the following:

 - A `--skip-tests` flag to skip all the tests in `garden dev`.
 - Introduction of concept of test-groups. 
   - Group definition. I very much prefer the former approach of "*creating a groups field and list the tests there per group*" @ellenkorbes proposed. Group names are not unique.
   - `--test-groups` flag to select the groups.
[...]

PR #917, opened 2019-07-02
feat: Implement --test-names/--name for dev/test commands

10ko:


eysi09:
@10ko could you outline the design you had in mind? Since this is core framework stuff we should make sure we're all on the same page (cc @edvald).

10ko:
Hi @eysi09, I still don’t have something 100% clear in mind, this is a very early stage WIP PR. I will post some more comments and questions in #851 and then outline the design here. ✌️

docs/reference/commands.md
```diff
@@ -189,9 +189,11 @@ as you modify the code.
 Examples:
 
     garden dev
-    garden dev --hot-reload=foo-service       # enable hot reloading for foo-service
     garden dev --hot=foo-service,bar-service  # enable hot reloading for foo-service and bar-service
     garden dev --hot=*                        # enable hot reloading for all compatible services
+    garden dev --skip-tests=foo-service       # skip running any tests
```
edvald: The parameter shouldn't have a value

garden-service/src/commands/dev.ts
```diff
@@ -51,6 +51,11 @@ const devOpts = {
   "skip-tests": new BooleanParameter({
     help: "Disable running the tests",
   }),
+  "test-names": new StringsParameter({
+    help: "The name(s) of the module(s) to test (skip to test all modules). " +
```
edvald: The help string is misleading. Shouldn't it be more like "Filter the tests to run by test name (leave unset to run all tests). "? Phrasing suggestion @thsig?

garden-service/test/unit/src/commands/test.ts
```diff
@@ -68,4 +68,45 @@ describe("commands.test", () => {
       },
     })).to.be.true
   })
+
+  it.only("should only run integration tests if the option 'name' is specified with a glob", async () => {
```
edvald: Oops, left an `.only` here!

garden-service/src/commands/dev.ts
```diff
@@ -51,6 +51,11 @@ const devOpts = {
...
+      "Accepts glob patterns (e.g. integ* would run both 'integ' and 'integration')",
```
edvald: Also, missing period at end. Maybe add one for the skip-tests parameter above as well.

garden-service/test/unit/src/commands/test.ts
10ko: 🤦‍♂

docs/reference/module-types/exec.md
```diff
@@ -353,58 +351,3 @@ tests:
     command:
     env: {}
 ```
-
-## Outputs
```
edvald: Hmm, maybe you need to rebase or re-run docs generation. This shouldn't disappear.

garden-service/src/commands/dev.ts
thsig: How about this?

"Filter the tests to run by test name across all modules (leave unset to run all tests)."

I.e. to make explicit that this filters across modules (which implicitly recommends using common names like `unit` or `integ`).

PR #924, opened 2019-07-05
feat(core): add SOCKS proxy support WIP

### LINKED ISSUE #920: Allow configuring SOCKS proxy for external requests
Garden sometimes needs to fetch external dependencies at runtime. Users that need to connect through SOCKS proxies currently get stuck on this. We should add support for SOCKS proxies, configured through environment variables and/or CLI flags.

edvald:
Closes #920


eysi09:
On that topic, we should document the available environmental variables and perhaps centralise them at the code level as well (beyond just `process.env`).

Issue #894, opened 2019-06-27
Only rebuild on last save when running "garden dev"


10ko:
When running `garden dev` the filesystem is watched and any change will trigger a new build/deploy/tests run.
This also mean that if a user presses 5 times save in 5 seconds Garden will schedule 5 build/deploy/tests runs.

Feature: make Garden interrupt the current run if a user changes a file and restart it.
It might need investigation on how to do it gracefully (I can think of before/after or db to dump for tests) and other stuff.

edvald:
@thsig I've been noticing this as well. Kinda feels like it may be a regression actually?

thsig:
Yeah, sounds like a regression. I'll look into this.

thsig:
> Feature: make Garden interrupt the current run if a user changes a file and restart it.
It might need investigation on how to do it gracefully (I can think of before/after or db to dump for tests) and other stuff.

I like this idea, but we should move this into a separate Github issue if we want to pursue it.

This might involve adding plugin handlers for `cancelBuild`, `cancelDeploy` etc., which the task classes in question would wrap in a `cancel` method. The watch logic could then ask the `TaskGraph` to attempt cancelling a given task (if it supports cancelling for the plugin/handler it represents).

PR #926, opened 2019-07-05
fix(task-graph): fix task deduplication

thsig:
Resolves https://github.com/garden-io/garden/issues/894.

Fixed a regression that allowed more than one pending (i.e. not
in-progress) task to be added to `TaskGraph` for a given `key`.

This happened when the add & process queues were merged into a single
op queue (for valid reasons), and could cause repeated file changes
to result in unnecessarily repeated builds, deploys etc.

After this fix, at most one task can be pending for a given `key`.


garden-service/test/helpers.ts
```diff
@@ -325,6 +325,17 @@ export function stubModuleAction<T extends keyof ModuleActions<any>>(
   return td.replace(garden["moduleActionHandlers"][actionType][moduleType], pluginName, handler)
 }
 
+export function defer() {
```
edvald: A comment here would be nice. I had to read the code that uses this to figure out the point. That aside, I like the solution!

thsig: Good point, added a comment just now. Thanks!

PR #934, opened 2019-07-08
fix(k8s): always flatten resources of kind List

### LINKED ISSUE #905: v1/List not supported
## Bug

This issue comes out of a discussion in #814 
The resource type `v1/List` does not seem to be supported by garden currently.

```
Failed deploying service seldon-core-operator (from module seldon-core-operator). Here is the output:
————————————————————————————————————————————————————————————————————————————————
Unrecognized resource type v1/List
————————————————————————————————————————————————————————————————————————————————


1 deploy task(s) failed!
```

### Your environment

garden version: 0.10.0-1

eysi09:
Fixes #905

### DIFF
--- a/garden-service/src/plugins/kubernetes/helm/common.ts
+++ b/garden-service/src/plugins/kubernetes/helm/common.ts
@@ -24,7 +24,7 @@ import { ConfigurationError, PluginError } from "../../../exceptions"
 import { Module } from "../../../types/module"
 import { findByName } from "../../../util/util"
 import { deline } from "../../../util/string"
-import { getAnnotation } from "../util"
+import { getAnnotation, flattenResources } from "../util"
 import { KubernetesPluginContext } from "../config"
 
 /**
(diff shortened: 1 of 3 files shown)


garden-service/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -75,13 +75,15 @@ export async function createContainerObjects(
...
-  return objects.map(obj => {
+  const resources = objects.map(obj => {
     set(obj, ["metadata", "annotations", gardenAnnotationKey("generated")], "true")
     set(obj, ["metadata", "annotations", gardenAnnotationKey("version")], version.versionString)
     set(obj, ["metadata", "labels", "module"], service.module.name)
     set(obj, ["metadata", "labels", "service"], service.name)
     return obj
   })
+
+  return flattenResources(resources)
```
edvald: This is unnecessary here because we create these manifests ourselves (and we wouldn't use the List type).

Issue #936, opened 2019-07-08
Garden doesn't work in project sub directories


eysi09:
### Current Behavior

Garden commands fails with:

```
Not a project directory (or any of the parent directories): /path/to/project/some-dir
```

if run in a sub directory of a project.

### Expected behavior

Garden should work in project sub directories.

### Reproducible example

Go to `demo-project`, cd into `frontend` and run `garden validate`. 

### Workaround

Run Garden commands from root.

edvald:
Hmh. I wonder when this broke. Seems we're missing an appropriate test.

PR #940, opened 2019-07-09
fix: garden can again be run from project subdirs

thsig:
Fixes a regression that resulted in Garden commands failing to find the project-level config if run from a subdirectory of project root.

Closes https://github.com/garden-io/garden/issues/936.

Also added a unit test to prevent this regression from being repeated.


garden-service/test/unit/src/config/base.ts
```diff
@@ -313,3 +313,16 @@ describe("loadConfig", () => {
   })
 
 })
+
+describe("findProjectConfig", async () => {
+  it("should find the project config when path is projectRoot", async () => {
+    const project = await findProjectConfig(projectPathA)
+    expect(!!project).to.eq(true)
```
edvald: Small thing, but could we additionally make sure that the project path is the expected one? Just to prevent possible later regressions.

thsig: Will do.

PR #942, opened 2019-07-09
docs: clarify dependency workflow for integ in demo-project

theRealWardo:
after going through the example, I went one small step further and started changing things while using `garden dev`. a bit more specifically, when playing with the `/call-backend` response message and the assertion on it, I bumped into a scenario where the system did not behave as I expected.

I believe it is appropriate to have the integration test depend on a deployment of the `frontend` as well as the `backend`. I've updated the example and the documentation in this PR.


edvald:
Thanks @theRealWardo! Sorry about the slow response, been busy busting out the v0.10.1 release.

We had a chat about this, and you're right, the example is poorly formulated. As I see it (now that you mention it), the current integ test should really belong to the `backend` module (with a dependency on the `backend` service), and we should make another one that tests the `frontend` service, which in turn has a transitive dependency on the `backend` so we'd test the relationship between the two. Does that make sense?

It's maybe a little more work to fix, and we're happy to jump in, but if you're interested we'd also love it if you could help us make our examples clearer! We sometimes get blinded because we're too close to the project, so your perspective helps a lot :)

theRealWardo:
I am not sure I agree with your assessment. here is how I see it:

the `/frontend/test/integ.js` file includes the following:
```
      .get("/call-backend")
      .expect(200, { message: "Backend says: 'Hello from Go!'" })
```
this is a call to `GET /call-backend` which hits the **frontend** and returns a response that _wraps_ a response from the **backend**. the **frontend** includes this logic which shows how the wrapping is achieved:
```
  request.get(backendServiceEndpoint)
    .then(message => {
      message = `Backend says: '${message}'`
```

1) if the backend changes from
```
        fmt.Fprint(w, "Hello from Go!")
```
to
```
        fmt.Fprint(w, "Hello from Golang!")
```
the test should fail.

2) if the frontend changes from
```
      message = `Backend says: '${message}'`
```
to
```
      message = `The backend says: '${message}'`
```
the test should fail.

try making both of these changes *independently* and seeing what happens when you are running `garden dev`. I believe that 1 does result in a failure (as it should) but 2 does not result in a failure (and it should).

therefore, I would propose that the test is asserting on both an updated `backend` and `frontend` which should be expressed as I propose in this change. this is indeed a common pattern in my opinion, we have a [cucumber](https://cucumber.io/]) integration test suite that runs in a deployed container in our cluster and we do want **all** services updated before those integration tests run. I am of course a n00b to Garden (I cracked it open for the first time in the last couple days), but I think the change I've proposed does make sense as it is what you call the "one that tests the frontend service, which in turn has a transitive dependency on the backend so we'd test the relationship between the two". sure the backend is lacking tests, but that is a different problem I am happy to tackle in another PR if you'd like that to be part of the example, however I would add that it is valuable that this example includes a module that does not have unit tests associated with it as I hear some developers don't write tests (shocking I know).

edvald:
Oops 🤦‍♂️ sorry, I was debating you looking at different code! I was mixing up examples, which is kind of a separate issue (various same-ish examples copied all over the place instead of completely consistent), which we have on a TODO to fix.

You're right, the integ test is fine. The only comment that remains is that you only need to state a dependency on `frontend` because the `backend` dependency is transitive. Sorry about that!

theRealWardo:
no worries! that was super helpful as I didn't realize the backend could be expressed as a transitive dependency like that. thanks! I pushed an updated commit that explains that.

PR #945, opened 2019-07-09
feat: add env var for setting max task concurrency

thsig:
The `GARDEN_TASK_CONCURRENCY_LIMIT` env variable can now be used to override the task graph's default concurrency limit. Values should be integers.

Also set this env variable to 10 in our CircleCI config.

### DIFF
--- a/garden-service/src/task-graph.ts
+++ b/garden-service/src/task-graph.ts
@@ -41,6 +41,10 @@ export interface TaskResults {
 
 export const DEFAULT_CONCURRENCY = 6
 
+const concurrencyFromEnv = process.env.GARDEN_TASK_CONCURRENCY_LIMIT
+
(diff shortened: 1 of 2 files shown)


garden-service/src/task-graph.ts
```diff
@@ -41,6 +41,10 @@ export interface TaskResults {
 
 export const DEFAULT_CONCURRENCY = 6
 
+const concurrencyFromEnv = process.env.GARDEN_TASK_CONCURRENCY_LIMIT
+
+export const TASK_CONCURRENCY = (concurrencyFromEnv && parseInt(concurrencyFromEnv, 10)) || DEFAULT_CONCURRENCY
```
edvald: Small thing: Using the uppercase variable name indicates an actual constant, which isn't the intent. No worries though, fine to merge.

thsig: Yeah, good point.

PR #946, opened 2019-07-09
chore(examples): use local.username as default suffix instead of "default"

edvald:


garden-service/test/integ/src/pre-release.ts
```diff
@@ -31,7 +32,7 @@ describe("PreReleaseTests", () => {
   }
 
   function getProjectNamespaces() {
-    const ns = `${project}-testing-${process.env.CIRCLE_BUILD_NUM || "default"}`
+    const ns = `${project}-testing-${process.env.CIRCLE_BUILD_NUM || username() || "default"}`
```
eysi09: Doesn't `username()` return a promise? (https://github.com/sindresorhus/username)

eysi09: ...unless you use `username.sync()`?

edvald: ach, yes, I'll fix that

PR #961, opened 2019-07-12
fix(k8s): ensure images built remotely are tagged before publishing

### LINKED ISSUE #960: Publish command fails for images built in-cluster
### Current Behavior

Running `garden publish --env remote` fails if the image hasn't been built locally. The reason is that the image name that is pulled doesn't match the image name that's published.

### Expected behavior

Garden should publish the image.

### Workaround

Publish the image from a local environment, e.g. `garden publish my-module --env local` for local refers to an environment that uses the `local-kubernetes` provider. 

### Suggested solution(s)

We need to re-tag the pulled imaged. Something like: `docker tag pulledImage localId`.

eysi09:
**What type of PR is this?**
 
**fix**: A bug fix.

Publish command would fail for images built remotely since they will have different names than what's needed for the registry that's being published to. We might e.g. pull something like: `local.app.garden:53113/eysi-demo-project/backend:v-d8c8e24c77` but need to push something like: 
`eu.gcr.io/garden-dev-200012/backend:v-d8c8e24c77`. This PR resolves that.

Fixes #960

### DIFF
--- a/garden-service/src/plugins/kubernetes/container/publish.ts
+++ b/garden-service/src/plugins/kubernetes/container/publish.ts
@@ -12,7 +12,6 @@ import { containerHelpers } from "../../container/helpers"
 import { KubernetesPluginContext } from "../config"
 import { publishContainerModule } from "../../container/publish"
 import { getRegistryPortForward } from "./util"
-import execa = require("execa")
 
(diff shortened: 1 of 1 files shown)


eysi09:
Comment test.

PR #966, opened 2019-07-15
chore: automate release process

eysi09:
**What type of PR is this?**
* **chore**: Changes to the build process or auxiliary tools and libraries such as documentation generation.

With this PR we add more automation to our release process. In particular, we now publish releases _from CI_ as oppose to downloading the build artifacts from CI and uploading to Github manually. If the release is a pre-release it get's published automatically, otherwise a draft is created that needs to be published manually.

Furthermore, we now publish releases with the version tag `next` on every merge to `master`. 

See `CONTRIBUTING.md` for more details.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,12 @@ version: 2.1
           GARDEN_TASK_CONCURRENCY_LIMIT: "10"
     resource_class: large
 
+  # Configuration for release jobs
+  release-config: &release-config
+    docker:
+      # Image that contains ghr for publishing releases to Github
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -278,15 +307,53 @@ jobs:
   build-service-dist:
     <<: *node-config
     steps:
+      - build_service_dist:
+          version: $(./garden-service/bin/garden --version)
```
edvald: Since we're using garden itself to indicate the version, do we need to make `build-service` a dependency?

PR #970, opened 2019-07-15
fix(core): respect includes/excludes when syncing to build directory

### LINKED ISSUE #968: Garden syncs ignored files/dirs to .garden/build dir
### Current Behavior

Garden syncs everything to the `./garden/build` dir. Even if some of the content is ignored via the `include` directive or in a `.gardenignore` file.
 
### Expected behavior

It should exclude content that's ignored via `.gardenignore` or missing from the `include` list if provided.

### Reproducible example

1. Pick and example project
2. Run `rm -rf .garden/build` from the root.
3. Exclude files/dirs by adding a `.gardenignore` to a module or by using the `include` directive.
4. Run a command that stages a build.
5. Notice that nothing got excluded.

### Workaround

N/A.

edvald:
This notably cascades neatly to anything that syncs _from_ the build
directory, so there should be no further action needed in e.g.
hot-reloading or remote-building code.

Fixes #968

### DIFF
--- a/garden-service/src/build-dir.ts
+++ b/garden-service/src/build-dir.ts
@@ -21,7 +21,6 @@ import { FileCopySpec, Module, getModuleKey } from "./types/module"
 import { zip } from "lodash"
 import * as execa from "execa"
 import { normalizeLocalRsyncPath } from "./util/fs"
-import { ModuleConfig } from "./config/module"
 import { LogEntry } from "./logger/log-entry"
 
 // Lazily construct a directory of modules inside which all build steps are performed.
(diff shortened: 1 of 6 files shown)


edvald:
Tests caught issue with remote sources. I'll take a look.

garden-service/src/build-dir.ts
```diff
@@ -37,13 +36,18 @@ export class BuildDir {
     return new BuildDir(projectRoot, buildDirPath, buildMetadataDirPath)
   }
 
-  async syncFromSrc(module: ModuleConfig, log: LogEntry) {
-    await this.sync(
-      resolve(this.projectRoot, module.path) + sep,
-      await this.buildPath(module.name),
-      true,
+  async syncFromSrc(module: Module, log: LogEntry) {
+    const files = module.version.files
```
eysi09: This could be a single line.

edvald: I had previously had another step in the chain so it landed like this. Don't feel like it matters tbh.

garden-service/src/build-dir.ts
```diff
@@ -118,20 +132,28 @@ export class BuildDir {
...
     if (withDelete) {
       logMsg += " (with delete)"
     }
 
     log.debug(logMsg)
 
-    await execa("rsync", [...syncOpts, sourcePath, destinationPath])
+    let input: string | undefined
+
+    if (files !== undefined) {
```
eysi09: We could use `if (files)` here.

edvald: Technically, yes, but I tend to like this for clarity. Other languages sometimes evaluate empty lists as false, so my preference is verbosity.

PR #982, opened 2019-07-17
fix: various issues with path handling on Windows

edvald:


garden-service/src/plugins/kubernetes/container/build.ts
```diff
@@ -138,7 +138,7 @@ const remoteBuild: BuildHandler = async (params) => {
   const buildRoot = resolve(module.buildPath, "..")
   // The '/./' trick is used to automatically create the correct target directory with rsync:
   // https://stackoverflow.com/questions/1636889/rsync-how-can-i-configure-it-to-create-target-directory-on-server
-  let src = normalizeLocalRsyncPath(`${buildRoot}/./${module.name}/`)
+  let src = normalizeLocalRsyncPath(`${buildRoot}`) + `/./${module.name}/`
```
eysi09: Don't we need `sep` instead of forward slash here?

edvald: No, this is a normalized POSIX-style path

PR #986, opened 2019-07-17
fix(k8s): remote building broken with certain cluster network configs

edvald:
We now use Envoy to proxy requests from the kubelet to the in-cluster
registry when pulling images to deploy.


edvald:
Here is a workaround in the meantime for those users affected:

`kubectl -n garden-system edit daemonset/garden-registry-proxy`
and replace
`export CIDR_RANGE=$(echo $KUBERNETES_SERVICE_HOST | cut -d '.' -f 1).0.0.0/8`
with
`export CIDR_RANGE=10.0.0.0/8` (or the appropriate CIDR range, it's usually this one but you can tell by looking at the IP for one of Pods in your cluster)

If that doesn't work you could try removing `range=$CIDR_RANGE,` from the line below.

PR #994, opened 2019-07-18
fix(core): properly handle joining K8s api server url and path

### LINKED ISSUE #983: Cannot install garden services on remote cluster
## Bug
Cannot install on remote cluster.

### Current Behavior
Running `garden --env=dev plugins kubernetes cluster-init` results in the following:

`ℹ tiller                    → Waiting for resources to be ready...`
`Timed out waiting for tiller to deploy`

### Expected behavior

Garden services to be properly installed on the cluster.

### Reproducible example

1. Spin up cluster using Containership.io on DigitalOcean
2. Setup kubectl context using the `kubectl connection info` in Containership
3. Run `garden --env=dev plugins kubernetes cluster-init`

### Suggested solution(s)

I'm not yet well versed enough with Kubernetes to identify where the issue lies. I have tried installing helm/tiller manually though, and had no problems.

### Additional context
I am using a cluster created through https://containership.io/ running on DigitalOcean. I followed the instructions [here](https://docs.garden.io/using-garden/remote-kubernetes) but could not install Garden. I could not get past the *Initializing the cluster* step. 

### Your environment

`garden version`
*0.10.0*

`garden get debug-info`
[debug-info-2019-07-17T13:31:24.641Z.zip](https://github.com/garden-io/garden/files/3402081/debug-info-2019-07-17T13.31.24.641Z.zip)

eysi09:

Previously we merged the server url from the kube config with the path with the `url.resolve` function. Turns out `resolve` strips parts of the base server url away and ends up returning a bad url for some endpoints. Switched to using `proper-url-join` npm package.

Fixes #983 

Using `proper-url-join`, because, not surprisingly, `url-join` did not work. However, the former is not typed.


garden-service/src/plugins/kubernetes/api.ts
```diff
@@ -261,7 +261,7 @@ export class KubeApi {
 
   async request(log: LogEntry, path: string, opts: Omit<request.OptionsWithUrl, "url"> = {}): Promise<any> {
     const baseUrl = this.config.getCurrentCluster()!.server
-    const url = resolve(baseUrl, path)
+    const url = <string>urlJoin(baseUrl, path)
```
edvald: Quick suggestion: Maybe we should export this from `util/string.ts` and set the type there. Just cleaner and easier to import when it's needed.

PR #1004, opened 2019-07-22
fix(core): task dependencies were not automatically run ahead of tests

edvald:

Prior to this fix, tasks were incorrectly left out when computing direct dependencies of tests.

We still need more work on this subject, since you may want more fine-grained control over which tasks are run at what times, but this PR at least fixes a clear error.


garden-service/src/tasks/test.ts
```diff
@@ -69,27 +72,38 @@ export class TestTask extends BaseTask {
...
         service,
         force: false,
         forceBuild: this.forceBuild,
-      }))
-    }
+      }),
+    )
 
-    return Bluebird.all(deps)
+    return [buildTask, ...serviceTasks, ...taskTasks]
```
eysi09: Feels like we should formalise and document the relationship between different tasks.

edvald: This is essentially how we codify it. Do you mean formalize as in document?

garden-service/src/tasks/test.ts
```diff
@@ -205,6 +219,9 @@ async function getTestDependencies(graph: ConfigGraph, testConfig: TestConfig) {
 export async function getTestVersion(
   garden: Garden, graph: ConfigGraph, module: Module, testConfig: TestConfig,
 ): Promise<ModuleVersion> {
-  const moduleDeps = await graph.resolveDependencyModules(module.build.dependencies, testConfig.dependencies)
+  const moduleDeps = (await graph.resolveDependencyModules(module.build.dependencies, testConfig.dependencies))
```
eysi09: Should `resolveDependencyModules` filter on the module itself or is this a special case?

edvald: I considered that, but it would involve more refactoring and a change to the method signature, and I couldn't see any issues in other places where it's used.

PR #1009, opened 2019-07-23
fix(vcs): garden.yml changes now only affect relevant module version

edvald:
Previously we were explicitly including the module config file in the
module version file list. Now we ensure that it is _not_ included,
which is the intended behavior after we started hashing the actual
module config as part of the module version.


garden-service/src/config/module.ts
```diff
@@ -148,6 +148,7 @@ export interface ModuleConfig
 
   outputs: PrimitiveMap
   path: string
+  configPath?: string
```
eysi09: Why is this optional?

garden-service/src/config/project.ts
```diff
@@ -96,6 +96,7 @@ export interface ProjectConfig {
   kind: "Project",
   name: string
   path: string
+  configPath?: string
```
eysi09: Same as above, shouldn't the config path be required?

garden-service/src/watch.ts
```diff
@@ -106,23 +106,25 @@ export class Watcher {
     return (path: string) => {
       this.log.debug(`Watcher: File ${path} ${type}`)
 
-      const changedModules = modules.filter(m => m.version.files.includes(path))
+      const changedModules = modules
+        .filter(m => m.version.files.includes(path) || m.configPath === path)
```
eysi09: We're filtering on the `configPath` in at least three places. Should there be a generic function for this? Also feels like the type of thing one might forget.

garden-service/src/config/module.ts
edvald: Because we don't always have a config file, e.g. when plugins specify modules.

garden-service/src/watch.ts
edvald: Each filter clause is different, so I can't see how we'd make it more generic. And re: forgetting—that's what the tests are for .)

edvald:
Looking into it

PR #1010, opened 2019-07-23
fix(container): respect include field when checking for Dockerfile

edvald:
Previously, when using multiple container modules in one garden.yml,
it would be impossible to have one of them use a remote image if any
of the others had a Dockerfile in the same directory.


edvald:
Ugh, the diff seems much larger because I moved a file and split it in two. I thought git would at least detect the move. The only tests really affected are the ones I added for `hasDockerfile`.

PR #1019, opened 2019-07-24
feat(core): make module scans more configurable and ignores more robust

edvald:
This adds `dotIgnoreFiles`, `modules.include` and `modules.exclude` 
fields to the project config.

The first allows you to specify which "dot ignore" files to use across
you project. To match current behavior, `dotIgnoreFiles` defaults to
`.gitignore` and `.gardenignore`. The feature is also more robust
now because we properly use git to evaluate the files in all cases now,
including handling nested ignore files.
This meant getting rid of the `Ignorer` package, which was always quite
flimsy anyway, and instead using the VcsHandler when scanning the
project, much like module sources.

Secondly, we allow explicit filtering of which directories to scan for
modules, using the `modules.include/exclude` keys. This allows us to
control project scanning, without blanket ignoring files that we may
actually want to be part of module builds.


garden-service/src/util/fs.ts
```diff
@@ -13,14 +13,15 @@ import { pathExists, readFile, writeFile } from "fs-extra"
...
 import { ValidationError } from "../exceptions"
 import { platform } from "os"
-// NOTE: Importing from ignore/ignore doesn't work on Windows
-const ignore = require("ignore")
+import { VcsHandler } from "../vcs/vcs"
 
 const VALID_CONFIG_FILENAMES = ["garden.yml", "garden.yaml"]
 const metadataFilename = "metadata.json"
+export const defaultDotIgnoreFiles = [".gitignore", ".gardenignore"]
+export const fixedExcludes = [".git", ".garden", "debug-log-*"]
```
10ko: The original ignore was `debug-info-*` not `debug-log-*` or you perhaps meant `*.log`?

garden-service/src/util/fs.ts
```diff
@@ -104,75 +105,20 @@ export async function getChildDirNames(parentDir: string): Promise<string[]> {
...
-
-    if (!isConfigFilename(parsedPath.base)) {
-      continue
-    }
-
-    paths.push(parsedPath.dir)
-  }
-
-  return paths
+export async function findConfigPathsInPath(
```
10ko: May I suggest a more explicit name like `findModulePaths` or `findGardenModulePaths`?

10ko: Also there are missing `@param`s in the docs.

garden-service/src/garden.ts
```diff
@@ -622,6 +629,12 @@ export class Garden {
     return version
   }
 
+  async scanForConfigs(path: string) {
```
10ko: Maybe add a comment to explain reasons and context and to not mistake this function with the one down?

garden-service/src/util/fs.ts
edvald: Oops, yeah I'll fix that.

garden-service/src/util/fs.ts
edvald: These find all config paths, not just modules, so I went for the more generic term.

Also, we generally don't use `@param` annotations, feels unnecessary in most cases with TypeScript imo.

edvald:
I didn't consider the case where `include` and `exclude` are the same, but it feels like that should be fairly obvious to the user?

10ko:
Thanks Jón.  
👍LGTM!

PR #1026, opened 2019-07-24
fix: handle blank build args appropriately

### LINKED ISSUE #1023: Blank build args are handled incorrectly
### Current Behavior

Setting a non-valued build arg like this:

```yaml
buildArgs:
  CONTAINER: ${modules.base.outputs.local-image-name}
  TAG: ${modules.base.version}
  # this gets pulled from an env var
  CHECKOUT_KEY:
```

Or explicitly, like this:

```yaml
buildArgs:
  CONTAINER: ${modules.base.outputs.local-image-name}
  TAG: ${modules.base.version}
  # this gets pulled from an env var
  CHECKOUT_KEY: !!null
```

results in a Docker command incorrectly emitting `"null"` as a string like this:

```
docker build -t snapshot:v-69e99e5246 --build-arg CONTAINER=base --build-arg TAG=v-dac401d9dd --build-arg CHECKOUT_KEY=null /Users/eronarn/Code/investment_automation/.garden/build/snapshot
```

### Expected behavior

The expected Docker command is instead:

```
docker build -t snapshot:v-69e99e5246 --build-arg CONTAINER=base --build-arg TAG=v-dac401d9dd --build-arg CHECKOUT_KEY /Users/eronarn/Code/investment_automation/.garden/build/snapshot
```

This is special syntax in Docker builds: https://docs.docker.com/engine/reference/commandline/build/

> You may also use the --build-arg flag without a value, in which case the value from the local environment will be propagated into the Docker container being built:

### Your environment

```
[eronarn@Jamess-MacBook-Pro investment_automation]$ garden version
0.10.2
```

baspetersdotjpeg:
Docker has special syntax where blank build args make it pull that key from the environment. We currently pass build args explicitly as null when they are.

Fixes #1023

### DIFF
--- a/garden-service/src/plugins/container/build.ts
+++ b/garden-service/src/plugins/container/build.ts
@@ -70,7 +70,15 @@ export function getDockerBuildFlags(module: ContainerModule) {
   const args: string[] = []
 
   for (const [key, value] of Object.entries(module.spec.buildArgs)) {
-    args.push("--build-arg", `${key}=${value}`)
+
+    // 0 is falsy
+    if (value || value === 0) {
+      args.push("--build-arg", `${key}=${value}`)
+    } else {
+      // If the value of a build-arg is null, Docker pulls it from
+      // the environment: https://docs.docker.com/engine/reference/commandline/build/
+      args.push("--build-arg", `${key}`)
+    }
   }
 
   if (module.spec.build.targetImage) {


edvald:
Just a formatting error. You can verify locally with `npm run lint`.

edvald:
Ah, it's the `==` instead of `===`. JS fun.

baspetersdotjpeg:
@edvald that should do the trick 👍

PR #1030, opened 2019-07-25
chore: also remove 'edge' tag when stripping tags

eysi09:


edvald:
Can you give some context for this? Just curious

eysi09:
Yup, it's for generating a clean changelog. The changelog is basically based on the tags so having the `edge` tag in there will add a entry for that in the changelog.

The flow looks like this:

1. Pull all remote tags
2. Delete all pre-release tags locally so that we only have tags like: v0.1.1, v0.2.0, .., v0.10.2. 
3. Also delete the `edge` tag
4. Generate the changelog
5.  Pull all remote tags to reset the original state

eysi09:
Actually, this might not be needed if I tweak the command that generates the changelog. I'll close this for now.

eysi09:
Scratch that, we do need this.

PR #1038, opened 2019-07-27
feat(openfaas): enable remote building for openfaas modules

edvald:
Also updated the OpenFaaS components. I'll follow up with another
commit to make the configuration of the OpenFaaS system components
more flexible, and allow for a more secure setup. 

Decided to push this right away because OpenFaaS support is essentially
busted for remote deployments right now and causes test failure.


garden-service/src/util/fs.ts
```diff
@@ -18,7 +18,7 @@ import { ValidationError } from "../exceptions"
 import { platform } from "os"
 import { VcsHandler } from "../vcs/vcs"
 
-const VALID_CONFIG_FILENAMES = ["garden.yml", "garden.yaml"]
+const VALID_CONFIG_FILENAMES = ["garden.yml", "garden.yaml", "garden.json", "garden.json5"]
```
edvald: Oops, left that in there from an experiment, will remove

PR #1044, opened 2019-07-29
fix: exclude symlinks to directories from hashing

### LINKED ISSUE #1029: garden commands error if there's a broken symlink
### Current Behavior

Commands like `garden scan` fail with errors like this if there's a broken symlink:

`ENOENT: no such file or directory, stat '/Users/eronarn/Code/qf/app/assets/javascripts/docs'`

In my case:

```
[eronarn@ip-192-168-10-243 qf]$ ls -lath '/Users/eronarn/Code/qf/app/assets/javascripts/docs'
lrwxr-xr-x 1 eronarn staff 31 Nov 27  2018 /Users/eronarn/Code/qf/app/assets/javascripts/docs -> ../../../qdocs_build/_static/js
```

The intervening folders exist, but `_static` doesn't, so the symlink is (currently) broken.

### Expected behavior

Local symlink issues don't prevent non-build commands, like scanning the project.

### Suggested solution(s)

There should be a warning about the symlink, or potentially the module may fails building if a critical symlink can't be resolved; but project-level commands shouldn't fail on this condition.

### Additional context

Usually one would expect symlinks to resolve, but not always; like in this case where I'm working in a repo where I haven't been building the docs. The application would work without the symlink and it's there for developer convenience.

### Your environment

```
[eronarn@ip-192-168-10-243 qf]$ garden version
0.10.2
```

10ko:
This PR fixes #1029, which was caused by bad handling of symlink when initializing garden.

Fixes #1029

We use the git client to retrieve all the files from a project, from where to extract then the module config paths. The current implementation then doesn't allow the use of symlinks: the git client will never resolve a symlink if it's a directory. It will always treat it as file and (this is now fixed/prevented) would try to hash it using the node streams, failing.
From a design point of view, I am not 100% aware of all the reasons behinf choosing `git ls-files` instead of something else. What I find strange is the jump from "using a library" to retrieve files and "using bare node streams" to hash it. It would be nice to understand if it makes sense to convert the search function to the same node standard or viceversa if we can use the git client to do the hashing. 

No

### DIFF
--- a/garden-service/src/vcs/git.ts
+++ b/garden-service/src/vcs/git.ts
@@ -126,7 +126,12 @@ export class GitHandler extends VcsHandler {
         // If we can't compute the hash, i.e. the file is gone, we filter it out below
         let hash = ""
         try {
-          hash = await this.hashObject(resolvedPath) || ""
+          // "git ls-files" returns a symlink even if it points to a directory.
+          // We filter symlinked directories out, since hashObject() will fail to
+          // process them.
+          if (!(await stat(resolvedPath)).isDirectory()) {
+            hash = await this.hashObject(resolvedPath) || ""
+          }
         } catch (err) {
           // 128 = File no longer exists
           if (err.code !== 128 && err.code !== "ENOENT") {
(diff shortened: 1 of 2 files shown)


10ko:
Make sense, thanks!

PR #1046, opened 2019-07-30
fix(k8s): avoid "no deployed releases" errors after Helm install failure

### LINKED ISSUE #1045: OpenFaaS projects need two attempts to deploy
### Current Behavior

When deploying a project that uses an `openfaas` module for the first time, deploying the `openfaas-system` service fails with `CrashLoopBackOff` error. This is because some of the Pods in the chart restart rapidly until they're able to connect to the `nats` service, and we've opted to consider BackOff warnings as failures (in order to fail fast while configuring/coding).

### Expected behavior

The project deploys as normal.

### Reproducible example

Run `garden deploy` in the `openfaas` example project. Doesn't matter which environment.

### Workaround

Run `garden deploy` once, expect the error, and then everything works normal from the next deploy onwards.

### Suggested solution(s)

I've reached out on the OpenFaaS Slack to see if we could/should fix this on their side, but it seems likely _something_ like this will come up in other contexts. Here are some ideas I have that could help:

1. Some way to specify dependencies within Helm charts on our side.
2. Some option to not regard BackOff warnings (or warnings in general) as errors.

### Your environment

Latest master.

edvald:
Fixes #1045

### DIFF
--- a/garden-service/src/plugins/kubernetes/helm/deployment.ts
+++ b/garden-service/src/plugins/kubernetes/helm/deployment.ts
@@ -59,6 +59,9 @@ export async function deployService(
       "--name", releaseName,
       "--namespace", namespace,
       "--values", valuesPath,
+      // Make sure chart gets purged if it fails to install
+      "--atomic",
+      "--timeout", "600",
     ]
     if (force) {
       installArgs.push("--replace")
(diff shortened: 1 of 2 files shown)


edvald:
This may actually also fix #1045, testing that now.

edvald:
It does indeed!

PR #1051, opened 2019-07-30
feat(core): add docker command flags

wjacquette:
This adds in the ability for the user to pass in more commands into the `docker build` more generically. For example, it allows the user to use docker's `--ssh default` build flag which is a part of the most recent docker unstable version.

Expressing in `garden.yml` as the following:
```
...
commandArgs:
    ssh: default
...
```

It is possible that the created function `getDockerCommandFlags` could replace the already existing function `getDockerBuildFlags `, however, it would break already existing uses of the function as the `garden.yml` file would need to be made differently.
Currently:
```
buildArgs:
    key: value
```
With `getDockerCommandFlags` instead:
```
commandArgs:
    build-arg: key=value 
```

This would allow for one more generic function to handle docker build flags, as opposed to have it be restricted to certain flags. 

Yes, it allows for the user to use more docker build flags.


wjacquette:
@edvald @eysi09

edvald:
Hey @wjacquette and thanks for the PR!

At first brush my concern is that the arguments may not be portable across different builder implementations. For example, when using [in-cluster building](https://docs.garden.io/using-garden/in-cluster-building) and Kaniko, the command argument wouldn't work. The `container` module type is meant to be (to the extent feasible) a high level abstraction that can be used in various contexts, so we've tried to avoid things like this.

But... There are certainly valid uses for an "escape hatch" like this, and as long as we warn the user (in the key description field) that the arguments may not be portable across implementations, we should be fine.

I'd like to propose a slightly different design though: Instead of `commandArgs` as an object, I'd suggesting making it `extraFlags` and have it be a simple list of strings. Those can then be passed to build commands easily, and there's no awkward key/value stuff to parse. This would also make it easier to change the flags based on the environment you're running, since string values can be templated, but object keys cannot.

wjacquette:
Hello @edvald!
I believe I have covered all the concerns from your comment in my most recently pushed changes. 
now instead of:
```
commandArgs:
    ssh: default
```
We have:
```
extraFlags: 
  - --ssh
  - default
```
Or
```
extraFlags: [--ssh, default]
```

garden-service/src/plugins/container/build.ts
```diff
@@ -55,6 +55,10 @@ export async function buildContainerModule({ module, log }: BuildModuleParams<Co
 
   const cmdOpts = ["build", "-t", identifier, ...getDockerBuildFlags(module)]
 
+  cmdOpts.push(...module.spec.extraFlags ? module.spec.extraFlags : [])
+
+  console.log(cmdOpts)
```
edvald: Oops, left a console.log here :)

garden-service/src/plugins/container/build.ts
```diff
@@ -55,6 +55,10 @@ export async function buildContainerModule({ module, log }: BuildModuleParams<Co
 
   const cmdOpts = ["build", "-t", identifier, ...getDockerBuildFlags(module)]
 
+  cmdOpts.push(...module.spec.extraFlags ? module.spec.extraFlags : [])
```
edvald: No biggie but this can be shortened to `cmdOpts.push(...module.spec.extraFlags || [])`.

garden-service/src/plugins/container/build.ts
```diff
@@ -87,3 +91,16 @@ export function getDockerBuildFlags(module: ContainerModule) {
 
   return args
 }
+
+export function getDockerExtraFlags(module: ContainerModule) {
```
edvald: This function may be unnecessary now that it's just a simple string array.

edvald:
Couple of minor comments but otherwise 👍👍 I'll approve and merge once those are all set. Thank you!

wjacquette:
Thanks! Sorry about some of those mistakes I left in. Appreciate the feedback.

edvald:
Oh no worries, those are exactly the mistakes I do all the time 😬

edvald:
The test failure is because of a conflicting change we just merged. I'm happy take to rebase and take care of that.

wjacquette:
Thanks for fixing it! Looking forward to using it soon

edvald:
Doing the merge from #1065 (had to make a new PR to be able to edit it, turns out)

PR #1052, opened 2019-07-31
feat: automatic port forwarding for deployed services

### LINKED ISSUE #967: Automatic port-forwarding/tunneling
Adding feature request here for visibility of a user.

@edvald please expand your thoughts here!

edvald:
This adds a built-in TCP proxy for long-running Garden processes,
that immediately opens a port for each forwardable service port (as
specified by providers) and creates a tunnel on-demand when the port is
accessed.

This should work for any TCP port, and is implemented here for the
kubernetes provider, but the proxy itself could be implemented for other
providers through the fairly simple `getPortForward` handler interface.

Unlike kubefwd, for example, this implementation avoids any
need to mess with domain names or host files (and by extension running
as root) since a random free port is assigned directly on localhost.
The tunnels are only created when a connection is first made, and are
then kept alive while the Garden agent is running.

Closes #967

### DIFF
--- a/garden-service/src/plugins/kubernetes/hot-reload.ts
+++ b/garden-service/src/plugins/kubernetes/hot-reload.ts
@@ -19,7 +19,7 @@ import { Service } from "../../types/service"
 import { LogEntry } from "../../logger/log-entry"
 import { getResourceContainer } from "./helm/common"
 import { waitForContainerService } from "./container/status"
-import { getPortForward, killPortForward } from "./util"
+import { getPortForward, killPortForward } from "./port-forward"
 import { RSYNC_PORT } from "./constants"
 import { getAppNamespace } from "./namespace"
 import { KubernetesPluginContext } from "./config"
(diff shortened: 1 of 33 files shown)


garden-service/src/plugins/kubernetes/hot-reload.ts
```diff
@@ -241,11 +241,11 @@ export async function syncToService(
   targetName: string,
   log: LogEntry,
 ) {
-  const targetDeployment = `${targetKind.toLowerCase()}/${targetName}`
+  const targetResource = `${targetKind.toLowerCase()}/${targetName}`
```
thsig: Remove the block comment above this function (since we're now relying on a general port-forward cache that's used for more than just hot reloading).

garden-service/src/plugins/kubernetes/port-forward.ts
```diff
@@ -0,0 +1,146 @@
...
+import { getAppNamespace } from "./namespace"
+import { registerCleanupFunction } from "../../util/util"
+import { PluginContext } from "../../plugin-context"
+import { kubectl } from "./kubectl"
+import { KubernetesResource } from "./types"
+import { ForwardablePort } from "../../types/service"
+import { isBuiltIn } from "./util"
+import { LogEntry } from "../../logger/log-entry"
+
+// TODO: implement stopPortForward handler
```
thsig: Should we add a brief high-level comment here? Maybe something like:
> Maintains a simple in-process cache of randomly allocated local ports that have been port-forwarded to a given port on a given Kubernetes resource. This removes the need for the user to manually establish port-forwards for e.g. services defined in Garden config files.

garden-service/src/types/service.ts
```diff
@@ -131,19 +140,48 @@ export const serviceIngressSchema = serviceIngressSpecSchema
...
+    .default("TCP")
+    .description("The protocol of the port."),
+  targetHostname: joi.string()
+    .description("The target hostname of the service (only used for informational purposes)."),
+  targetPort: joi.number()
+    .integer()
+    .required()
+    .description("The target port on the service."),
+  urlProtocol: joi.string()
+    .description("The protocol to use for URLs pointing at the port. This can be valid URI protocol."),
```
thsig: "This can be" -> "This can be any"

PR #1053, opened 2019-07-31
Test built binary in CI

eysi09:

It adds a check for the built binary to our CI pipeline.

We've had issues where the compiled TS version works but the built binary fails.

EDIT: The first commit re-names integ tests to e2e tests throughout the codebase. This is something we've talked about internally and the latter is a more appropriate name.


eysi09:
Was actually expecting the `test-dist` step to fail.

PR #1055, opened 2019-07-31
ci: move garden bin to HOME dir in test-dist job

eysi09:

The previous version ran the Garden bin from the `garden` root dir which is a git repo. Turns out, that bypassed a bug which only crops up if the binary is not in a git repo at all. So now we've moved the binary to the `$HOME/.garden` dir.


eysi09:
Note that this should fail on first try since the original error mentioned above has not been fixed.

eysi09:
And now it should pass, after the rebase.

PR #1062, opened 2019-08-01
chore: update release script due to deps updates

eysi09:
...and make sure release script updates package-lock.json

Had to fix how execa was called due to recent update of deps in the root level package.json. Also, contains a hacky way to update package-lock files via lerna. Here's the issue and workaround: lerna/lerna#1415


eysi09:
Left a commented line in there that I just fixed.

PR #1066, opened 2019-08-01
fix(run): interactive run module/service commands not working

edvald:

Basically running interactive commands via `garden run module` or `garden run service` was busted, after some refactors. Not sure how to best automate tests around this tbh, but it's something we need to look out for.

Please try messing with the run commands in a few different ways, and see if you find any issues, even related issues.

### DIFF
--- a/garden-service/src/commands/run/service.ts
+++ b/garden-service/src/commands/run/service.ts
@@ -78,6 +78,7 @@ export class RunServiceCommand extends Command<Args, Opts> {
       service,
       runtimeContext,
       interactive: true,
+      timeout: 999999,
     })
 
     return { result }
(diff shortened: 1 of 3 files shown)


eysi09:
Fixes #1069

garden-service/src/commands/run/service.ts
```diff
@@ -78,7 +78,7 @@ export class RunServiceCommand extends Command<Args, Opts> {
       service,
       runtimeContext,
       interactive: true,
-      timeout: 9999999,
+      timeout: 999999,
```
thsig: There was one 9 too many here (caused an overflow error).

PR #1071, opened 2019-08-05
Remote sources fix watch

### LINKED ISSUE #965: Watch mode does not work for linked remote sources
### Current Behavior

When running Garden in watch mode and working on a linked remote source, changes to that source are not detected by Garden. Instead you have to kill the process and run the command again.

### Expected behavior

Garden should detect the changes and automatically re-run.

### Reproducible example

1. Use the `remote-sources` example project
2. Clone one of the remotes, say `garden-example-remote-sources-web-services`
3. Link it with `garden link source garden-example-remote-sources-web-services local/path/to/clone`
4. Notice how changes to the cloned code don't trigger a watch event

### Workaround

Re-run the command on changes

eysi09:

First of all, it fixes an issue where Garden wasn't watching linked sources (when you do `garden link my-module /to/some/path`). We fix this by simply including all linked paths when initialising the `Watcher` class.

Second, it refactors the `test-project-ext-module-sources` and `test-project-ext-project-sources` test projects. Before, they would contain `mock-dot-garden` and `mock-local-path` directories. We then did a lot of stubbing to make Garden think these were the `.garden` dir and `/some/local/project/dir`, respectively.

Instead of all this, we now initialise these projects with a proper `.garden` directory that contains the sources and git init them, just like Garden would expect in the "real world". We also remove the `mock-local-path` dirs and instead use `test-project-local-*-sources` projects when testing the linking functionality. 

Fixes #965


garden-service/src/util/fs.ts
```diff
@@ -21,7 +21,8 @@ import { VcsHandler } from "../vcs/vcs"
 const VALID_CONFIG_FILENAMES = ["garden.yml", "garden.yaml"]
 const metadataFilename = "metadata.json"
 export const defaultDotIgnoreFiles = [".gitignore", ".gardenignore"]
-export const fixedExcludes = [".git", ".garden", "debug-info-*"]
+// FIXME: We should use `garden.gardenDirPath` instead of ".garden" since the dir name can be variable
```
edvald: I don't quite understand this comment

garden-service/src/watch.ts
```diff
@@ -20,15 +20,17 @@ import { isConfigFilename } from "./util/fs"
 // IMPORTANT: We must use a single global instance of the watcher, because we may otherwise get
 // segmentation faults on macOS! See https://github.com/fsevents/fsevents/issues/273
 let watcher: FSWatcher | undefined
-let projectRoot: string
 
-// The process hangs after tests if we don't do this
-registerCleanupFunction("stop watcher", () => {
+// Export so that we can clean up the global watcher instance when running tests
+export function cleanUp() {
```
edvald: I'd suggest a more descriptive name in that case

garden-service/src/watch.ts
```diff
@@ -39,13 +41,11 @@ export type ChangeHandler = (module: Module | null, configChanged: boolean) => P
...
-  constructor(private garden: Garden, private log: LogEntry, modules: Module[]) {
-    projectRoot = this.garden.projectRoot
-
-    this.log.debug(`Watcher: Watching ${projectRoot}`)
+  constructor(private garden: Garden, private log: LogEntry, paths: string[], modules: Module[]) {
+    log.debug(`Watcher: Watching paths ${paths.join(", ")}`)
 
     if (watcher === undefined) {
-      watcher = watch(projectRoot, {
+      watcher = watch(paths, {
```
edvald: This basically means that the watcher will not work right if you add a new source. That was okay(ish) when it was just the project root, but may be problematic now. I think it's fairly simple to solve though. We just need to keep track of which paths are currently being watched (or read that from the existing watcher if it exists) and add/remove as necessary.

edvald: Yeah, you can just get a list of currently watched paths (if any), and reconcile that against the paths we want to watch: https://github.com/paulmillr/chokidar#methods--events

eysi09: Do you mean if someone runs the link / unlink commands while Garden is already running in watch mode?

garden-service/src/util/fs.ts
eysi09: We stopped using `.garden` as a global constant and removed all references to it at some point. Instead the path is a property on the Garden class that we set when we initialise it. So you could technically do `new Garden({ gardenDirPath: ".foo" })`. We use this for the system Garden class (although it's nested within `.garden`). 

Btw, this should also be `".garden/**/*`, so currently this doesn't do anything. Thought I had fixed it but maybe I botched some rebase or something.

garden-service/src/watch.ts
edvald: No, more thinking of when you add a source in your project config while watching for changes.

garden-service/src/util/fs.ts
edvald: I see. I'd then rather explicitly add that path to the excludes list in `Garden.factory()` or the constructor, and keep the `fixedExcludes` constant as-is.

garden-service/src/watch.ts
eysi09: Ah, didn't think of that. However, you'd still need to link it (even if it's local), which is a separate command. So the assumption is still that the user would kill the running process, run `garden link` and then start it again.

eysi09: Linking basically updates the `.garden/local-config.yml` file and I'm not sure if a running process will handle changes to that.

edvald: We could also treat that as a config change and watch that file, I suppose

eysi09: Yeah I guess. Wondering if there are any unwanted side effects since Garden updates this file programmatically. And not just for setting/removing linked sources. But I guess in any case you'd want to trigger a config change event.

edvald: We could also special-case the particular variables in the config. But I'd suggest creating a separate issue for that.

eysi09:
Here's the issue for the watcher paths thing: https://github.com/garden-io/garden/issues/1082

eysi09:
Just realised I didn't push my commit for this:

>>> I'd then rather explicitly add that path to the excludes list in Garden.factory() or the constructor, and keep the fixedExcludes constant as-is.

It's there now and I'll squash if you approve.

PR #1086, opened 2019-08-08
fix: exception when retrieving logs from helm resource

### LINKED ISSUE #1068: Unhandled exception viewing logs
### Current Behavior

When clicking "Logs" on the daskboard, it spins forever, and I see this in my console:

Screen Shot 2019-08-02 at 8 51 18 AM

The line of code throwing this error appears to be: https://github.com/garden-io/garden/blob/master/garden-service/src/plugins/kubernetes/util.ts#L67

I am using the stable/mysql helm chart which doesn't have a selector key in the spec. I logged the object to my console:

```JavaScript
{ apiVersion: 'extensions/v1beta1',
  kind: 'Deployment',
  metadata:
   { name: 'mysql',
     namespace: 'local',
     labels:
      { app: 'mysql',
        chart: 'mysql-1.3.0',
        release: 'mysql',
        heritage: 'Tiller' },
     annotations: {} },
  spec: { template: { metadata: [Object], spec: [Object] } } }
```

Here's from my garden.yml:

```yaml
kind: Module
type: helm
name: mysql
description: Mysql deployment via helm
chart: stable/mysql
values:
  mysqlRootPassword: password
  initializationFiles:
    oauth-db.sql: |-
      CREATE DATABASE IF NOT EXISTS oauth;
```

### Expected behavior

Exceptions are caught. Logs can be viewed.

### Reproducible example

```yaml
kind: Project
name: oauth2
environments:
  - name: local
    variables:
      namespace: local
providers:
  - name: local-kubernetes
    environments: [local]
    namespace: local
    
---
# mysql
kind: Module
type: helm
name: mysql
description: Mysql deployment via helm
chart: stable/mysql
values:
  mysqlRootPassword: password
  initializationFiles:
    oauth-db.sql: |-
      CREATE DATABASE IF NOT EXISTS oauth;
```

### Workaround

Get logs via kubectl :(

### Suggested solution(s)

🔜  https://github.com/microsoft/TypeScript/issues/16#issuecomment-515160784

### Your environment

docker-for-desktop mac

`garden version`
`0.10.3`
`kubectl version`
```Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.0", GitCommit:"e8462b5b5dc2584fdcd18e6bcfe9f1e4d970a529", GitTreeState:"clean", BuildDate:"2019-06-20T04:49:16Z", GoVersion:"go1.12.6", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.3", GitCommit:"5e53fd6bc17c0dec8434817e69b04a25d8ae0ff0", GitTreeState:"clean", BuildDate:"2019-06-06T01:36:19Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}```
`docker version`
```
Client: Docker Engine - Community
 Version:           19.03.1
 API version:       1.40
 Go version:        go1.12.5
 Git commit:        74b1e89
 Built:             Thu Jul 25 21:18:17 2019
 OS/Arch:           darwin/amd64
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          19.03.1
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.5
  Git commit:       74b1e89
  Built:            Thu Jul 25 21:17:52 2019
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          v1.2.6
  GitCommit:        894b81a4b802e4eb2a91d1ce216b8817763c29fb
 runc:
  Version:          1.0.0-rc8
  GitCommit:        425e105d5a03fabd737a126ad93d62a9eeede87f
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

### PR DESCRIPTION
**Pull request by 10ko**
Whenever we retrieve logs with `garden logs` we assume the resource has a `selector`. This is not always true when we deal with Helm charts. This PR introduce a new function to determine a selector in case of an Helm chart. 

Please @edvald or @eysi09, do a sanity check to see if it makes sense. I did some tests with `stable/mysql` and some other charts and it "looks" like it's working but yeah, maybe you have some specific scenarios/charts in mind.

Fixes #1068

Well, it fixes an exception.

### DIFF
--- a/garden-service/src/plugins/kubernetes/util.ts
+++ b/garden-service/src/plugins/kubernetes/util.ts
@@ -53,10 +53,42 @@ export async function getAllPodNames(api: KubeApi, namespace: string, resources:
 }
 
 /**
+ * Given a resources, try to retrieve a valid selector or return undefined otherwise.
+ */
+export function getSelectorFromResource(resource: KubernetesWorkload) {
+  // We check if the resource has its own selector
+  if (resource.spec && resource.spec.selector
+    && resource.spec.selector.matchLabels) {
+    return resource.spec.selector.matchLabels
+  }
+  // We check if the pod template has labels
+  if (resource.spec.template
+    && resource.spec.template.metadata
+    && resource.spec.template.metadata.labels) {
+    return resource.spec.template.metadata.labels
+  }
+  // We check if the resource is from an Helm Chart
+  // (as in returned from kubernetes.helm.common.getChartResources(...))
+  if (resource.metadata
+    && resource.metadata.labels
+    && resource.metadata.labels.chart
+    && resource.metadata.labels.app) {
+    return {
+      app: resource.metadata.labels.app,
+    }
+  }
+
+  // No selector found.
+  throw new ConfigurationError(`No selector found for ${resource.metadata.name} while retrieving pods.`, {
+    resource,
+  })
+}
+
+/**
  * Retrieve a list of pods based on the provided label selector.
  */
 export async function getWorkloadPods(api: KubeApi, namespace: string, resource: KubernetesWorkload) {
-  const selector = resource.spec.selector.matchLabels
+  const selector = getSelectorFromResource(resource)
   const pods = await getPods(api, resource.metadata.namespace || namespace, selector)
 
   if (resource.kind === "Deployment") {

### CODE REVIEW

File: garden-service/src/plugins/kubernetes/util.ts [hunk 1]
```diff
@@ -52,11 +52,34 @@ export async function getAllPodNames(api: KubeApi, namespace: string, resources:
   return (await getAllPods(api, namespace, resources)).map(p => p.metadata.name)
 }
 
+/**
+ * Given a resources, try to retrieve a valid selector or return {} otherwise.
+ * NOTE: selector: {} matches all the pods for the resource, when using the k8s APIs.
+ */
+export function getSelectorFromResource(resource: KubernetesWorkload) {
+  let selector = {}
```
eysi09: I guess my only concern would be the case where the function returns `{}`. Wondering if no pods are a better default than all the pods. Might also be useful to add a `log.debug` statement for that case, regardless of how we handle it.

10ko: On point. I am open to both tbh. I am not extremely familiar with selectors and I don't know what's the "expected" default.

10ko: Updated the function to return undefined instead.
Thanks @eysi09

garden-service/src/plugins/kubernetes/util.ts
```diff
@@ -52,11 +53,37 @@ export async function getAllPodNames(api: KubeApi, namespace: string, resources:
   return (await getAllPods(api, namespace, resources)).map(p => p.metadata.name)
 }
 
+/**
+ * Given a resources, try to retrieve a valid selector or return undefined otherwise.
+ * NOTE: { selector: undefined } matches no pods for the resource, when using the k8s APIs.
+ */
+export function getSelectorFromResource(resource: KubernetesWorkload) {
+  const logger = getLogger()
```
edvald: This is not a good idea, we should pass a `LogEntry` instance to the function.

garden-service/src/plugins/kubernetes/util.ts
```diff
@@ -52,11 +53,37 @@ export async function getAllPodNames(api: KubeApi, namespace: string, resources:
...
+      && resource.metadata.labels.chart
+      && resource.metadata.labels.app) {
+      return {
+        app: resource.metadata.labels.app,
+      }
+    }
+  }
+  // No selector found, return.
+  logger.debug(`No selector found for ${resource.metadata.name}: returning empty selector.`)
+  return
```
edvald: Hmm, if we return undefined here, how does that play with the `getPods`? Shouldn't we rather just throw? We don't want to get _all_ of the Pods from a namespace if we're missing a selector.

10ko: So it looks like the k8s javascript library has a class `V1LabelSelector` which is used in various resources and matches everything if it's empty but nothing if it's undefined.
The function `listNamespacedPod(...)` called by `getPods(...)` has a parameter called `labelSelector` which is, of course, not of type `V1LabelSelector` and defaults to "all the pods".

I thought the getPods would follow the same behaviour, hence my decision of either returning empty object or undefined. 

I'd also rather throw.

garden-service/src/plugins/kubernetes/util.ts
10ko: I am going to remove this since we'll throw further down.
I am wondering why this is a bad idea thou?

edvald: It's because this completely ignores any nesting and structure in the logs. With `garden dev`, for example, this would print the log line at the bottom of the terminal and not next to the relevant task log.

eysi09:
@edvald, is this one ready?

garden-service/src/plugins/kubernetes/util.ts
```diff
@@ -52,11 +52,42 @@ export async function getAllPodNames(api: KubeApi, namespace: string, resources:
...
+    // (as in returned from kubernetes.helm.common.getChartResources(...))
+    if (resource.metadata
+      && resource.metadata.labels
+      && resource.metadata.labels.chart
+      && resource.metadata.labels.app) {
+      selector = {
+        app: resource.metadata.labels.app,
+      }
+    }
+  }
```
edvald: Sorry, could have thought of this earlier, but... Instead of doing this special case, wouldn't it make sense to just use the exact labels from the Pod template (`spec.template.metadata.labels`)? And maybe leave the last Helm-specific clause as a 3rd fallback? Because if the Pod spec explicitly specifies labels, those will work fine as a selector.

garden-service/src/plugins/kubernetes/util.ts
```diff
@@ -52,11 +52,40 @@ export async function getAllPodNames(api: KubeApi, namespace: string, resources:
...
+/**
+ * Given a resources, try to retrieve a valid selector or return undefined otherwise.
+ */
+export function getSelectorFromResource(resource: KubernetesWorkload) {
+  // We check if the resource has its own selector
+  if (resource.spec.selector) {
+    return resource.spec.selector.matchLabels
+  }
+  // We check if the pod template has labels
+  if (resource.spec.template.metadata.labels) {
```
edvald: This may error if one of the variables doesn't exist. Actually the same applies above as well, even though it's less likely to occur IRL.

10ko:
Really looking forward to [this](https://github.com/microsoft/TypeScript/issues/16#issuecomment-515160784).

PR #1096, opened 2019-08-12
Terraform provider

### LINKED ISSUE #1095: Allow exposing runtime outputs across services and tasks
Currently we can only reference _module_ outputs, that are statically resolved when resolving configuration files. It would be useful in a number of different ways to be able to reference service outputs (which would be exported by the relevant provider, via the `getServiceStatus` handler) and task outputs (which would be exported through the `runTask`/`getTaskStatus` handlers).

Service outputs could e.g. be used to propagate runtime values such as DNS names, IP addresses, and other deployment/infrastructure outputs. A practical example would be exposing outputs from an IaaC provider, e.g. Terraform.

Task outputs would work in a similar fashion, and could be very useful because providers could export command outputs that are then exposed to other tasks or services.

edvald:

This adds a `terraform` provider, that allows users to incorporate a Terraform stack as part of their Garden project. The `terraform-gke` example should illustrate the usage, and the provided tests and test project should also make the functionality fairly clear.

Please walk through the usage guide, and try out the `terraform-gke` example and see if everything works as expected, and that the instructions are clear. 
(I'm tbh fully expecting something to be not quite right, so please don't hold back.)


examples/terraform-gke/README.md
```diff
@@ -0,0 +1,72 @@
...
+```tfvars
+# terraform.tfvars
+gcp-region     = "europe-west1"
+gcp-zone       = "europe-west1-b"
+gcp-project-id = "my-test-project"
+```
+
+**B)** Configure the variables directly in the project `garden.yml` (see the `terraform` provider section).
+
+### Step 5 - Initialize the cluster
```
eysi09: This should be `Step 4`, same for the steps below.

examples/terraform-gke/garden.yml
```diff
@@ -0,0 +1,17 @@
+kind: Project
+name: terraform-gke
+providers:
+  - name: terraform
+    autoApply: true
+    # This key must be specified if we want to apply a stack as part of the provider initialization.
```
eysi09: Shouldn't this command be above the `autoApply` key?

edvald: I should clarify that in the comment(s), but actually both are needed. Thing is, if you don't specify `root`, the provider by design does nothing on init. The idea being that it will also provide the `terraform` module type (once that's ready), and you may not always want to initialize some stack at the project level.

Perhaps it'll be much clearer if it's a simple boolean toggle though. Just wondering what to call it. Ideas? 😬

eysi09: Yeah, feels kind of weird that the `root` field implicitly triggers an apply on init. Is there a case where you use `root` but don't want to apply on `init`?

garden-service/src/docs/config.ts
```diff
@@ -41,6 +41,7 @@ const moduleTypes = [
   { name: "kubernetes", pluginName: "local-kubernetes" },
   { name: "maven-container" },
   { name: "openfaas", pluginName: "local-kubernetes" },
+  // { name: "terraform" },
```
thsig: Should this be commented out?

thsig: Should this line be removed?

edvald: It'll be uncommented once we enable the `terraform` module type

examples/terraform-gke/garden.yml
edvald: Yes. In some cases you'll want to run Terraform manually, e.g. if you need to manually input variables, or if you need to manually review the execution plan. In that case, the provider will warn or error if the stack is not up-to-date.

edvald:
Note: This will be WIP until #1123 is merged.

docs/reference/module-types/terraform.md
```diff
@@ -0,0 +1,349 @@
+# `terraform` reference
+
+Resolves a Terraform stack and either applies automatically (if `autoApply: true`) or errors when the stack
```
eysi09: Shouldn't it be _applies it_?

docs/reference/providers/terraform.md
```diff
@@ -0,0 +1,104 @@
...
+
+[providers](#providers) > autoApply
+
+If set to true, Garden will automatically run `terraform apply -auto-approve` when a stack is not up-to-date. Otherwise, a warning is logged if the stack is out-of-date, and an error thrown if it is missing entirely.
+
+| Type      | Required | Default |
+| --------- | -------- | ------- |
+| `boolean` | No       | `false` |
+
+### `providers[].initRoot`
```
eysi09: I remember we debated this one. You need to specify this to trigger an apply on init, right? Name still feels a little weird. I would expect something like `configPath` but then again that may not capture the init part.

eysi09: Ah yes, the example below explains. Perhaps that explanation should also be included here. Don't think I have a better name in mind.

eysi09:
Testing the example project and ran into two issues.

1. Still can't use underscore for the terraform variables in the `garden.yml`
2. Garden can't find the context: `Error: could not get Kubernetes config for context "gke": context "gke" does not exist`. This happens even though the kubeconfig path is set to the Terraform provider output, and that config has `gke` as the context.
3. Garden writes `kubeconfig.yaml` to the project root. Is that suppose to happen? It also writes: `garden.tfvars`, `gke.tf` and `terraform.tfstate` but those are git ignored.

edvald:
I'm not sure what's happening but both #1 and #2 are working fine on my side. I'll check again to be sure.

Re the `kubeconfig.yaml` file, yes, that's the intended effect. `gke.tf` is part of the project, `terraform.tfstate` is written by Terraform, and we write `garden.tfvars` from the `variables` declaration.

edvald:
@eysi09 are you sure you're running the most recent code? @thsig could you maybe test the example and see if you run into issues?

docs/reference/providers/terraform.md
edvald: Which example are you referring to? Do you mean the guide? I could add some explanation and perhaps a link to the guide from the reference docs.

eysi09:
My bad, looks like I didn't have the latest commits. (Did `git reset --hard origin/terraform-provider-2` without first doing a `git fetch`).

PR #1099, opened 2019-08-13
feat(helm): add valueFiles field to specify custom value files

### LINKED ISSUE #1098: Permit to get values from files when using helm module
It would be great to add an option which permit to use external yaml file to get helm chart values.

My use case: currently, I don't use garden and I have a helm chart for my API, with default value file and one value file per environment (testing, stage, production...) I use the right value file depending of environment on which I want to deploy. Now, I want to try to add garden in this workflow and use for example my values file `my_api.testing.yaml` in helm deployment. 

As suggest by @edvald on slack: an option `valuesFile` should be ok. But, the possibility to specify a valuesFile depending of the current environment will be even better. For example:

```
kind: Module                            
name: my-api                          
description: ...
type: helm                              
chart: my-chart                               
valuesFiles:
  - environment: testing
    file: my-api.testing.yaml
  - environment: stage
    file: my-api.stage.yaml
  - environment: prod
    file: my-api.prod.yaml   
```

We should add the possibility to declare a `default` value too, in case of user doesn't want to specify one file for each environment. 

One thing annoy me with this implementation: environments names will be used in both files: project and module and modification of them in project file will need the modification on module project too.

edvald:

Previously you could only (easily) provide custom values to charts via the `values` key. Having different value files based on context/environment is a pretty common practice with Helm, so we should enable that.

You could for example have a base values file with defaults, and then a custom one based on environment:

```yaml
kind: Module
type: helm
...
valueFiles:
  - values.common.yaml
  - values.${environment.name}.yaml
```

You can still use the current `values` field, which will take precedence over the files listed under `valueFiles`.

Closes #1098

See if the docs and the schema makes sense, as documented.

Yes


docs/using-garden/using-helm-charts.md
```diff
@@ -108,6 +108,53 @@ tests:
...
+kind: Module
+type: helm
+name: my-helm-module
+...
+values:
+  some:
+    key: some-value
+```
+
+This will effectively create a new YAML with the supplies values and pass it to Helm when rendering/deploying the chart. This is particularly handy when you want to template in the values (see the next section for a good example).
```
eysi09: _supplies_ => _supplied_

docs/using-garden/using-helm-charts.md
```diff
@@ -108,6 +108,53 @@ tests:
...
+  key: other-default
+```
+
+```yaml
+# values.prod.yaml
+some:
+  key: prod-value
+```
+
+In this example, `some.key` is set to `"prod-value"` for the `prod` environment, and `other.key` keep the default value set in `values.default.yaml`.
```
eysi09: _keep_ => _keeps_. Or maybe better, _maintains_.

docs/using-garden/using-helm-charts.md
```diff
@@ -108,6 +108,53 @@ tests:
...
+
+If you were to also set the `values` field in the Module configuration, the values there would take precedence over both of the value files.
```
eysi09: May drop the "hypothetical past tense" (or whatever it's called) and just say: "_If you also set the `values` field [...], the values there take..._"

garden-service/src/plugins/kubernetes/helm/config.ts
```diff
@@ -209,9 +210,23 @@ export const helmModuleSpecSchema = joi.object().keys({
...
-    .description(
-      "Map of values to pass to Helm when rendering the templates. May include arrays and nested objects.",
-    ),
+    .description(deline`
+      Map of values to pass to Helm when rendering the templates. May include arrays and nested objects.
+      When specified, these take precedence over the values in the \`values.yaml\` file (or the files specified
+      in \`valueFiles\`).
+    `),
+  valueFiles: joiArray(joi.string().posixPath({ subPathOnly: true }))
+    .description(dedent`
```
eysi09: Shouldn't this be `deline`? Noticed some extra new lines in the generated doc above (although not visible to when rendered as markdown).

garden-service/src/plugins/kubernetes/helm/common.ts
```diff
@@ -125,8 +126,22 @@ export async function getChartPath(module: HelmModule) {
...
+}
+
+/**
+ * Get the value files arguments that should be applied to any helm install/render command.
+ */
+export async function getValueFileArgs(module: HelmModule) {
+  const chartPath = await getChartPath(module)
+  const gardenValuesPath = getGardenValuesPath(chartPath)
+
+  const valueFiles = module.spec.valueFiles
```
eysi09: Could you add a comment saying that the `gardenValuesPath` takes precedence over the other value files and is therefore appended last.

garden-service/src/plugins/kubernetes/helm/config.ts
edvald: No, that'll bunch up all the paragraphs, right? It's too much text for a single paragraph imo.

edvald:
Addressed the comments, to be squashed on merge.

PR #1106, opened 2019-08-16
Minikube fixes

eysi09:

Contains fixes to two issues I ran into while trying to run Garden on minikube in CI. 

One issue was that Garden just hangs while trying to deploy the `kubernetes-dashboard` helm chart. Minikube already has a `dashboard` addon so now we just use that. This does mean that the link to the K8s dashboard is no longer visible on the Garden dashboard for minikube users.

I'll add a separate issue for that since it's a higher priority to be able to deploy a project to begin with.

### DIFF
--- a/garden-service/src/plugins/kubernetes/local/config.ts
+++ b/garden-service/src/plugins/kubernetes/local/config.ts
@@ -6,6 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+import * as Bluebird from "bluebird"
 import * as execa from "execa"
 import { KubernetesBaseConfig, kubernetesConfigBase, k8sContextSchema } from "../config"
 import { ConfigureProviderParams } from "../../../types/plugin/provider/configureProvider"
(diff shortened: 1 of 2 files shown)


garden-service/src/plugins/kubernetes/local/config.ts
```diff
@@ -84,7 +85,13 @@ export async function configureProvider({ config, log, projectName }: ConfigureP
   }
 
   if (context === "minikube") {
-    await execa("minikube", ["config", "set", "WantUpdateNotification", "false"])
+    const initCmds = [
+      ["config", "set", "WantUpdateNotification", "false"],
+      ["addons", "enable", "dashboard"],
+    ]
+    Bluebird.map(initCmds, async (cmd) => {
```
edvald: Missing `await`. Plus you can simplify the handler to just `cmd => execa("minikube", cmd)`.

thsig: No further comments from me 👍

eysi09: Whoops, good catch.

eysi09:
I've updated the commit wrt the comment above.

PR #1108, opened 2019-08-18
docs: update remote sources guide

eysi09:
Note: We need to update the links to the Using Remote Sources doc once this has been merged. Currently the links to just point to the Using Garden parent section.


docs/using-garden/using-remote-sources.md
```diff
@@ -0,0 +1,160 @@
...
+    ├── Dockerfile
+    ├── garden.yml
+    └── ...
+```
+
+You can imagine that this file tree gets merged into the parent project.
+
+If you now run `garden get status` you will see all the services from the remote source:
+
+```sh
```
thsig: The services will have `state: missing` unless the user has already run `garden deploy` or the like.

Let's either change `state: ready` to `state: missing`, or instruct the user to run `garden deploy` before `garden get status` here.

docs/using-garden/using-remote-sources.md
```diff
@@ -0,0 +1,160 @@
...
+
+```yaml
+repositoryUrl: file:///my/local/project/path#master
+```
+
+As usual, the URL must point to a specific branch or tag.
+
+Local paths work just the same remote URLs and you'll still need to [link the repository](#linking-remote-sources/modules-to-local-code) if you want to edit it locally.
+
+In general we don't recommend using local paths except for testing purposes. The `garden.yml` files should be checked into your version control system and therefore shouldn't contain content specific to a user's setup.
```
thsig: _contain content specific to a user's setup_ -> _contain anything specific to a particular user's setup_

docs/using-garden/using-remote-sources.md
eysi09: Good catch. The idea wasn't really to have the user run `garden get status` but rather just show that Garden does indeed pick up the remote services. Maybe it would be clearer to just do a `garden deploy` though.

eysi09: ...or skip it altogether.

PR #1110, opened 2019-08-19
fix(proxy): handle undefined _remote Socket

### LINKED ISSUE #1088: Proxy fails with: TypeError: Cannot read property 'end' of undefined
### Current Behavior

Sometimes when opening a port forward link, Garden crashes with:

```
/Users/eysi/code/garden-io/garden/garden-service/src/proxy.ts:156
      _remote.end()
              ^
TypeError: Cannot read property 'end' of undefined
    at Socket.local.on (/Users/eysi/code/garden-io/garden/garden-service/src/proxy.ts:156:15)
    at Socket.emit (events.js:182:13)
    at TCP._handle.close (net.js:610:12)
```

### Reproducible example

Run `garden dev` in `demo-project`. Open the port forward link in your browser, close it, open again, etc etc until it crashes.

eysi09:

Circumvents an issue where a `"close"` event is fired for the `local` socket in `proxy.ts` with the `_remote` still undefined.

The event fires when _opening_ a port forward URL in the browser, usually after having opened and closed the tab a few times before. I'm not sure why it fires on open and why `_remote` is undefined.

Fixes #1088 

@edvald Could you please sanity check this? I'm not sure what the underlying issue is TBH.

### DIFF
--- a/garden-service/src/proxy.ts
+++ b/garden-service/src/proxy.ts
@@ -83,6 +83,7 @@ async function createProxy(garden: Garden, log: LogEntry, service: Service, spec
 
   const server = createServer((local) => {
     let _remote: Socket
+    let localDidClose = false
 
     const getRemote = async () => {
       if (!_remote) {
@@ -115,7 +116,13 @@ async function createProxy(garden: Garden, log: LogEntry, service: Service, spec
         _remote.on("error", (err) => {
           log.debug(`Remote socket error: ${err.message}`)
         })
+
+        // Local connection was closed while remote connection was being created
+        if (localDidClose) {
+          _remote.end()
+        }
       }
+
       return _remote
     }
 
@@ -153,7 +160,8 @@ async function createProxy(garden: Garden, log: LogEntry, service: Service, spec
     })
 
     local.on("close", () => {
-      _remote.end()
+      _remote && _remote.end()
+      localDidClose = true
     })
 
     local.on("error", (err) => {


garden-service/src/proxy.ts
```diff
@@ -149,11 +149,11 @@ async function createProxy(garden: Garden, log: LogEntry, service: Service, spec
     })
 
     local.on("drain", () => {
-      _remote.resume()
+      _remote && _remote.resume()
     })
 
     local.on("close", () => {
-      _remote.end()
+      _remote && _remote.end()
```
edvald: This'll fix the reported error, but it's _possible_ this would end up leaking memory or remote connections. What appears to be happening is that the local connection is closing before the remote connection is up. For completeness I'd suggest additionally setting a flag so that the remote connection is either not created or immediately closed after creation, since that whole procedure may keep going after this handler completes.

eysi09: `_remote` is **always** `undefined` on close, even though `_remote` has clearly been set.

edvald: Hmm I don't understand. I've never personally seen this issue, so that can't be right (even though I can definitely see how it _can_ happen). Can you reliably reproduce that state?

eysi09: Yeah, neither do I. And yes, I can reliably reproduce. If I open a tab, I get a `data` event with `_remote` defined. Now when I close it I _sometimes_ get a `close` event, and when I do, `_remote` is undefined, even though it was defined before. Sometimes the `close` event doesn't fire when I close the tab but later when I open it again. Much confuse.

I'll continue to investigate.

edvald: I'd suggest that maybe the `close` event is only firing when the browser has a chance to explicitly cancel a connection attempt, in which case `_remote` will always be undefined.

Also, "was defined before" may simply apply to another connection, so that may be confusing your debugging. The browser will open multiple connections so there will be a number of instances of `_remote` connections. We re-use a single tunnel, but create an outbound connection per every inbound connection.

edvald: To clarify, the handler defined at https://github.com/garden-io/garden/blob/43b414f576943e9aed1ea819a6612da8708b97b2/garden-service/src/proxy.ts#L84 is executed for every inbound connection, and only applies to a single connection.

garden-service/src/proxy.ts
```diff
@@ -149,11 +149,11 @@ async function createProxy(garden: Garden, log: LogEntry, service: Service, spec
     })
 
     local.on("drain", () => {
-      _remote.resume()
+      _remote && _remote.resume()
```
edvald: The `drain` event can only come up if the connection in question (here, the `local` connection) has been paused. We only do that after the remote connection is up, so I think this line change is not necessary.

edvald: Actually more accurately the `drain` event can come up if we've performed a write that hasn't flushed. But same result, no need for this change.

garden-service/src/proxy.ts
eysi09: I updated the PR. This is what you had in mind, right?

PR #1113, opened 2019-08-20
feat(hot-reload): added post-sync callback option

thsig:

Adds a new config option to hot reload sync specs: `callback`.

This is a command that is run inside the container after a hot reload
sync takes place for the source/target pair in question.

Doesn't fix any issues, but implements a feature that's been requested by several users.

If we could think of a more instructive/straightforward example use case of this than the `rebuild-static-assets.sh` I added to the Joi spec, that would be good. When requested by users, the use cases so far have been relatively nontrivial and/or specific to their setup/stack (things like clearing caches or running custom scripts).

Any thoughts on this?

### DIFF
--- a/docs/reference/module-types/container.md
+++ b/docs/reference/module-types/container.md
@@ -294,6 +294,26 @@ hotReload:
     - target: "/app/src"
 ```
 
+### `hotReload.sync[].callback[]`
+
+[hotReload](#hotreload) > [sync](#hotreload.sync[]) > callback
+
+A optional command to run inside the container after syncing. Applies only to files belonging to this source/target pair.
+
+| Type            | Required |
+| --------------- | -------- |
+| `array[string]` | No       |
+
+Example:
+
+```yaml
+hotReload:
+  ...
+  sync:
+    - callback:
+      - rebuild-static-assets.sh
+```
+
 ### `dockerfile`
 
 POSIX-style name of Dockerfile, relative to module root.
@@ -976,6 +996,7 @@ hotReload:
   sync:
     - source: .
       target:
+      callback:
 dockerfile:
 services:
   - name:
--- a/garden-service/src/plugins/kubernetes/hot-reload.ts
+++ b/garden-service/src/plugins/kubernetes/hot-reload.ts
@@ -18,6 +18,7 @@ import { set } from "lodash"
 import { Service } from "../../types/service"
 import { LogEntry } from "../../logger/log-entry"
 import { getResourceContainer } from "./helm/common"
+import { execInDeployment } from "./container/run"
 import { waitForContainerService } from "./container/status"
 import { getPortForward, killPortForward } from "./port-forward"
 import { RSYNC_PORT } from "./constants"
@@ -239,13 +240,29 @@ export async function syncToService(
   const doSync = async () => {
     const portForward = await getPortForward({ ctx, log, namespace, targetResource, port: RSYNC_PORT })
 
-    return Bluebird.map(hotReloadSpec.sync, ({ source, target }) => {
+    return Bluebird.map(hotReloadSpec.sync, async ({ source, target, callback }) => {
       const src = rsyncSourcePath(service.sourceModule.path, source)
       const destination = `rsync://localhost:${portForward.localPort}/volume/${rsyncTargetPath(target)}`
 
       log.debug(`Hot-reloading from ${src} to ${destination}`)
 
-      return execa("rsync", ["-vrpztgo", src, destination])
+      const rsyncResult = await execa("rsync", ["-vrpztgo", src, destination])
+
+      // TODO: Allow DaemonSet and StatefulSet?
+      if (callback && targetKind === "Deployment") {
+        // Run post-sync callback inside the pod
+        const callbackResult = await execInDeployment({
+          log,
+          namespace,
+          command: callback,
+          provider: ctx.provider,
+          deploymentName: targetName,
+          interactive: false,
+        })
+        log.debug(`Running callback "${callback}" for ${targetName}, output: ${callbackResult.output}`)
+      }
+
+      return rsyncResult
     })
   }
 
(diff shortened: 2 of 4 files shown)


docs/reference/module-types/container.md
```diff
@@ -294,6 +294,26 @@ hotReload:
     - target: "/app/src"
 ```
 
+### `hotReload.sync[].callback[]`
```
edvald: First thing I ponder is whether we really want the callback to belong to each sync clause? I have a hard time imagining users wanting different callbacks for each copy command. I'd rather default to fewer indents and fewer executions.

Also, if you do have multiple folders to sync, you need to either repeat the callback command or attach it to only one, in which case you couldn't be sure all syncs have completed when that command is run.

garden-service/src/plugins/kubernetes/hot-reload.ts
```diff
@@ -239,13 +240,29 @@ export async function syncToService(
...
+    return Bluebird.map(hotReloadSpec.sync, async ({ source, target, callback }) => {
       const src = rsyncSourcePath(service.sourceModule.path, source)
       const destination = `rsync://localhost:${portForward.localPort}/volume/${rsyncTargetPath(target)}`
 
       log.debug(`Hot-reloading from ${src} to ${destination}`)
 
-      return execa("rsync", ["-vrpztgo", src, destination])
+      const rsyncResult = await execa("rsync", ["-vrpztgo", src, destination])
+
+      // TODO: Allow DaemonSet and StatefulSet?
```
edvald: Can't see why not. Any workload should apply.

eysi09:
Is `callback` a good name for this? Makes it sound like you're passing a function. What about `hooks`, `postSyncHooks`, `postSyncCmds`, or `postSyncActions`? Or replace `postSync` with `afterSync` or `onSync`?

thsig:
> Is `callback` a good name for this? Makes it sound like you're passing a function. What about `hooks`, `postSyncHooks`, `postSyncCmds`, or `postSyncActions`? Or replace `postSync` with `afterSync` or `onSync`?

Right you are, `callback` isn't the best term for this. How about `afterSyncCmd`?

edvald:
How about `callbackCommand`?

edvald:
(imo the `Cmd` shortening is not helpful/necessary)

edvald:
On reflection I think I'd suggest `postSyncCommand`, and to move it up a level to the `hotReload` clause, instead of having one per synced folder.

docs/reference/module-types/container.md
thsig: > First thing I ponder is whether we really want the callback to belong to each sync clause? I have a hard time imagining users wanting different callbacks for each copy command.

The idea here is that the user may want the callback to run only when files change in a particular directory.

I was discussing this feature with @elliott-davis the other day, maybe he's got some more detailed use-cases to share here? I'm partly posting this PR just to open a discussion on this possible feature, so it's quite possible that I've gotten the requirements/rationale a bit off here.

> Also, if you do have multiple folders to sync, you need to either repeat the callback command or attach it to only one [...]

In the first case, the idea was that the callback would repeated in each relevant sync spec in the configuration. Maybe this would be better served by a "named callback" setup, similar to how we reference ports in service specs (to avoid repeating the callbacks).

> [...] in which case you couldn't be sure all syncs have completed when that command is run

Good point. How about running the callbacks in a separate `Bluebird.map` after the syncs have completed?

thsig:
I like `postSyncCommand`.

If we move it up a level to the `hotReload` clause, should we add optional `only: [list of paths]` and `exclude: [list of paths]`fields to the `postSyncCommand` spec?

docs/reference/module-types/container.md
edvald: My main concern is just that level of complexity and nesting in code + configurations. Thing is, all these sync commands run in parallel and can be expected to finish at roughly the same time. So there's no performance benefit to having separate commands.

We're also not really checking whether files changed in each of the sync targets, we're just running all the sync commands and then all associated post-sync commands. 

If we were to check the result of each sync command and then _conditionally_ run the post-sync command, this nesting would make a bit more sense to me, but frankly it does feel like we're deep in edge-case territory there.

edvald:
Perhaps we could instead defer to the user for the more advanced use-case. How about we supply the list of changed files in an environment variable, e.g. as a JSON-encoded array? Then that can be scripted around. I'm happy to pave the way for the use-case but want to avoid adding complexity to our configuration, and by extension make it harder to use when you _don't_ need that level of control.

docs/reference/module-types/container.md
elliott-davis: 👋 My use-case that @thsig was mentioning is the case of static assets being embedded in with the backend code and run as part of the same container. For this case @thsig's solution would be helpful.

To elaborate a bit more: I have a PHP app that runs behind apache in a container. I don't want to rebuild the UI assets each time the php code changes, only when the UI code changes. So I would like to set up two sync targets, one for `backend` and one for `web` where the web target would have a call back of `npm run build` and the backend code really wouldn't have a callback. 

I have a similar situation with a golang app that is hosted in Heroku. Since I can't have two containers receiving web traffic in the same deployment I am forced to compile the UI assets into the same container. Garden could easy my development here by building the static assets when the web code changes, and building the golang assets when the go assets change.

Personally, I'm not too bothered by the nesting. There are other areas of the config that get nested pretty deep (annotations). @edvald I'm not 100% sure I understand:

> Thing is, all these sync commands run in parallel and can be expected to finish at roughly the same time. So there's no performance benefit to having separate commands.

edvald: That makes sense. It's just that the current implementation would always call every specified callback/command, which makes the extra indentation unnecessary, and is why I thought maybe this was for performance reasons (hence that part of my comment). 

We either need to address that in a more complete way, by conditionally running the relevant command(s), or instead provide a list of changed files as an environment variable to the executed command, which can be used to determine a course of action on the user's side.

thsig: I suggest we go with one (optional) post-sync command, which is provided with a list of changed files as an environment variable.

Does that sound good to everyone?

thsig:
Closed in favour of https://github.com/garden-io/garden/pull/1211.

PR #1122, opened 2019-08-22
chore: auto-generate table of contents

### LINKED ISSUE #988: Automatically generate docs/README.md file
If a link to a doc is missing from our Table of Contents file in `docs/README.md`, the doc won't be available on Gitbook. 

We should generate the file automatically to prevent this from happening by accident.

ellenkorbes:

We often forget to update the table of contents, which causes errors with GitBook. This makes it so that the table of contents is automatically re-generated whenever we update the docs.

Fixes #988 

See comments atop `garden-service/src/docs/tableofcontents.ts`.

As a next step we'll have to go through our docs and add the appropriate `order` and `title` tags to their front matter.

No.


garden-service/src/docs/tableofcontents.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+let indent: string = "  "
+
+export function generateTableOfContents(docsRoot: string) {
+  let tree = <ObjType>dtree(docsRoot, { extensions: /\.md/, exclude: /node_modules/ })
+  attachMetadata(tree)
+  sortTree(tree)
+  let toWrite: string = "# Table of Contents\n\n" + generateTOC(tree, docsRoot, 0)
+  fs.writeFileSync(resolve(docsRoot, "README.md"), toWrite)
```
eysi09: I noticed we always use the sync file methods in the generate docs code. Not sure why though, maybe @edvald can explain. A priori it feels like this should be non-blocking so that we can generate the docs in parallel.

edvald: It's just old code, figured at the time it wouldn't matter because it's only run as a script. We can separately refactor that.

garden-service/src/docs/tableofcontents.ts
```diff
@@ -0,0 +1,92 @@
...
+    })
+    for (let item in obj.children) {
+      sortTree(obj.children[item])
+    }
+  }
+}
+
+function generateTOC(obj: ObjType, docsRoot: string, depth: number) {
+  let path: string = obj.path.replace(docsRoot, ".")
+  let output = repeat(indent, depth) + "* [" + obj.title + "](" + path + ")\n"
```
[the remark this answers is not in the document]
ellenkorbes: This is not a constant, it changes later.

ellenkorbes:
@eysi09 Please re-review. Will write tests and add metadata to current docs afterwards.

garden-service/src/docs/table-of-contents.ts
```diff
@@ -0,0 +1,128 @@
...
+import { readFileSync, writeFile, createFile } from "fs-extra"
+import { resolve } from "path"
+import { cloneDeep } from "lodash"
+
+interface Metadata {
+  order: number
+  title: string
+}
+
+interface fileTree extends dtree.DirectoryTree, Metadata {
```
eysi09: This should be pascal cased.

garden-service/src/docs/table-of-contents.ts
```diff
@@ -0,0 +1,128 @@
...
+    } else {
+      const name = file.match(/^#[^#][\s]*(.+?)#*?$/m)
+      if (name) {
+        tree.title = name[1]
+      } else {
+        tree.title = tree.name
+      }
+    }
+  }
+  catch (e) {
```
eysi09: This should be in the line above ( ` } catch (e) {`)

eysi09: Does the linter not complain?

garden-service/src/docs/table-of-contents.ts
```diff
@@ -0,0 +1,128 @@
...
+    if (e.code != "ENOENT") {
```
eysi09: `!==`

The linter should also catch this.

garden-service/src/docs/table-of-contents.ts
```diff
@@ -0,0 +1,128 @@
...
+  let newTree = cloneDeep(tree)
+  transform(newTree)
+  return newTree
+}
+
+function attachMetadata(tree: fileTree) {
+  if (tree.type === "directory") {
+    tree.path = tree.path + "/README.md"
+  }
+  try {
```
eysi09: I would only have the `readFileSync` function in the `try` statement.

Something like:

```
let file
try {
  file = ...
} catch (e) {
  ...
}

Comment (eysi09) [reply garden-service/src/docs/table-of-contents.ts]: You could also consider making this an `async` function and use just use `file = await readFile...`

File: garden-service/src/docs/table-of-contents.ts [hunk 7]
```diff
@@ -0,0 +1,128 @@
...
+  if (tree.children) {
+    for (let item in tree.children) {
+      attachMetadata(tree.children[item])
+    }
+  }
+}
+
+function sortTree(tree: fileTree) {
+  if (tree.children) {
+    tree.children.sort(function(a, b) {
```
eysi09: We usually use arrow functions for callbacks. So `sort((a, b) => { ... }))`

garden-service/src/docs/table-of-contents.ts
```diff
@@ -0,0 +1,128 @@
...
+      if (a.order === b.order) {
+        if (a.title > b.title) { return 1 } else { return -1 }
```
eysi09: Here you should use a ternary operator:

`a.title > b.title ? return 1 : return - 1`.

Side note, but as a _convention_, we usually don't use single line if statements. But we're not enforcing it at the linter level so I guess you're free to make your own call.

But in any case, I'd suggest the ternary here.

garden-service/src/docs/table-of-contents.ts
```diff
@@ -0,0 +1,128 @@
...
+      }
+      if (a.order > b.order) { return 1 } else { return -1 }
```
eysi09: Also here.

garden-service/src/docs/table-of-contents.ts
```diff
@@ -0,0 +1,128 @@
...
+    })
+    for (let item in tree.children) {
+      sortTree(tree.children[item])
+    }
+  }
+}
+
+function generateTOC(tree: fileTree, docsRoot: string, depth: number) {
+  const path = tree.path.replace(docsRoot, ".")
+  let output = repeat(indent, depth) + "* [" + tree.title + "](" + path + ")\n"
```
eysi09: Again I would just `repeat` from `lodash` here.

garden-service/src/docs/table-of-contents.ts
```diff
@@ -0,0 +1,128 @@
...
+  if (path === "./README.md") {
+    output = ""; depth = -1
```
eysi09: Also as a _convention_, we don't normally assign multiple variables in the same line. But again, the linter has the last word IMO.

garden-service/src/docs/table-of-contents.ts
ellenkorbes: I get no errors on VSCode and `npm run fix-format` left it like that. Am I missing something?

garden-service/src/docs/table-of-contents.ts
eysi09: Whoops, that was actually suppose to be:
```
return a.title > b.title ? 1 : -1
```
My bad.

ellenkorbes:
@eysi09 Okay I think this is ready now. Please re-re-re-re-re-review :)

PR #1123, opened 2019-08-23
feat(core): expose service and task dependency outputs at runtime

### LINKED ISSUE #1095: Allow exposing runtime outputs across services and tasks

edvald:

This is a fairly substantial addition, which allows users to reference
runtime outputs from services and tasks in dependant services and tasks.
It involves adding a couple of new TaskGraph tasks, so that we can
resolve service statuses and task results in dependency order. We also
substantially overhaul the usage and testing of runtime contexts across
the board, since it is now a more fundamental feature.

The usage is documented in a new section in the Configuration Files
guide, please start with that when reviewing.

Closes #1095


docs/using-garden/configuration-files.md
```diff
@@ -457,6 +457,36 @@ services:
 For a full reference of the keys available in template strings, please look at the
 [Template Strings Reference](../reference/template-strings.md).
 
+#### Runtime outputs
```
eysi09: We can do it in a separate PR but we really need to split this doc up.

docs/using-garden/configuration-files.md
```diff
@@ -457,6 +457,36 @@ services:
...
+
+Template keys prefixed with `runtime.` have some special semantics. They are used to expose runtime outputs from services and tasks, and therefore are resolved later than other template strings. _This means that you cannot use them for some fields, such as most identifiers, because those need to be resolved before validating the configuration._
```
eysi09: How does the user know what fields exactly can use `runtime.` template strings?

docs/reference/module-types/container.md
```diff
@@ -1132,3 +1133,18 @@ outputs:
   ...
   deployment-image-name: "my-deployment-registry.io/my-org/my-module"
 ```
+
+
+### Task outputs
```
eysi09: Shouldn't we also add service outputs here?

docs/reference/module-types/exec.md
```diff
@@ -432,3 +433,18 @@ The outputs defined by the module.
...
 | -------- | -------- |
 | `object` | Yes      |
+
+
+### Task outputs
+
+The following keys are available via the `${runtime.tasks.<task-name>}` template string key for `exec` module tasks.
+Note that these are only resolved when deploying/running dependants of the task, so they are not usable for every field.
+
+### `runtime.tasks.<task-name>.outputs.log`
```
eysi09: Are we foreseeing any other output types than log? Also log feels a little off to me, would `stdout` be clearer?

eysi09: Feels like this could just be `runtime.tasks<task-name>.stdout` (or `runtime.tasks<task-name>.output)`, but maybe I'm missing something.

garden-service/src/runtime-context.ts
```diff
@@ -0,0 +1,169 @@
...
+  module: Module
+  dependencies: DependencyRelations
+  serviceStatuses: { [name: string]: ServiceStatus }
+  taskResults: { [name: string]: RunTaskResult }
+}
+
+export async function prepareRuntimeContext(
+  { garden, module, dependencies, serviceStatuses, taskResults }: PrepareRuntimeContextParams,
+): Promise<RuntimeContext> {
+  // const buildDepKeys = module.build.dependencies.map(dep => getModuleKey(dep.name, dep.plugin))
```
eysi09: Remove this?

garden-service/src/runtime-context.ts
```diff
@@ -0,0 +1,169 @@
...
+interface PrepareRuntimeContextParams {
+  garden: Garden
+  graph: ConfigGraph
+  module: Module
+  dependencies: DependencyRelations
+  serviceStatuses: { [name: string]: ServiceStatus }
+  taskResults: { [name: string]: RunTaskResult }
+}
+
+export async function prepareRuntimeContext(
```
eysi09: It would be nice to better document the `runtimeContext` and how we use it.

garden-service/src/actions.ts
```diff
@@ -550,9 +572,28 @@ export class ActionHelper implements TypeGuard {
       defaultHandler,
     })
 
+    // Resolve ${runtime.*} template strings if needed.
+    if (runtimeContext && (await getRuntimeTemplateReferences(module)).length > 0) {
```
eysi09: Feels like this and the code above could be de-duplicated.

garden-service/src/tasks/base.ts
```diff
@@ -64,3 +77,21 @@ export abstract class BaseTask {
...
+  // DeployTask results take precedence over GetServiceStatusTask results, because status changes after deployment
+  const combined = { ...getServiceStatusResults, ...deployResults }
+  const statuses = mapValues(combined, r => r.output as ServiceStatus)
+  return mapKeys(statuses, (_, key) => splitLast(key, ".")[1])
+}
+
+export function getRunTaskResults(dependencyResults: TaskResults): { [name: string]: RunTaskResult } {
+  const storedResults = pickBy(dependencyResults, r => r.type === "get-task-result")
+  const runResults = pickBy(dependencyResults, r => r.type === "task")
+  // TaskTask results take precedence over GetTaskResultTask results
```
eysi09: That's a mouthful :)

docs/reference/module-types/container.md
edvald: We do in the generator, there just aren't any in this case.

docs/using-garden/configuration-files.md
edvald: That's tricky, because it's going to be quite dependent on individual module type schemas. So we can't describe it here, nor can we easily flag this in schemas.

docs/reference/module-types/exec.md
edvald: Not for this module type in particular, but tasks could feasibly have any number of outputs. Say if a provider outputs file paths for artifacts, particular metadata etc.

garden-service/src/actions.ts
edvald: Feels like, yes, but I felt the code is subtly different enough for an abstraction to be unnecessary.

garden-service/src/tasks/base.ts
edvald: Yeah, once we rename the TaskGraph to ActionGraph and *Tasks to *Actions this will be more digestable :P

docs/using-garden/configuration-files.md
edvald: True. But in a separate PR for sure. Maybe an `Advanced configuration` guide would be good.

edvald:
I'll take a look at the `Getting status...` issue, hadn't noticed that when testing.

Re: Test outputs, we could add them as well, but there was no immediately obvious use case—you could only pass test outputs to other tests, and I'm not sure if/how that's useful—so I figured we could fairly easily add those later.

PR #1133, opened 2019-08-29
feat: blue-green deployment

10ko:
- Introduce the ability to specify different deployment-strategies.
- Implement blue-green deployments.
- Adds "experimental" meta field to joi schemas.
- Introduce "Experimental feature" warning message.
- Add new example called `deployment-strategies`

Fixes #

### DIFF
--- a/examples/deployment-strategies/README.md
+++ b/examples/deployment-strategies/README.md
@@ -1,0 +1,33 @@
+# Deployment strategies example
+
+A basic demo project showing different deployment strategies.
+
+It is based on the [examples/demo-project](https://github.com/garden-io/garden/tree/master/examples/demo-project) and it's meant to show how to configure the deployment strategies.
+
+> NOTE: the "Deployment Strategies feature" is still in the experimental phase. This means there might be changes to the properties names/values or to the behaviour in the future. Please be aware of this when using the feature.
+
+
+## Usage
+
+This project doesn't require any specific set up and can be deployed (in your local cluster) in a single step with the `deploy` command:
+
+```sh
+garden deploy --env=local-blue-green
+```
+
+The first deploy on a fresh cluster will always be a normal `rolling-update` deploy. After making some changes to any of the module you will be able to see the Blue/Green strategy in action.
+
+### Example of configuration
+
+For more detailed configuration please check out the `garden.yml` file.
+
+```yaml
+kind: Project
+name: My Project
+environments:
+  # Blue-green deployment strategy on local-kubernetes.
+  - name: local-blue-green
+    providers:
+      - name: local-kubernetes
+        deploymentStrategy: blue-green
+```
--- a/garden-service/src/plugins/kubernetes/container/deployment.ts
+++ b/garden-service/src/plugins/kubernetes/container/deployment.ts
@@ -6,20 +6,22 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+import chalk from "chalk"
+import Bluebird from "bluebird"
 import { V1Container } from "@kubernetes/client-node"
-import { extend, keyBy, set } from "lodash"
 import { Service, ServiceStatus } from "../../../types/service"
+import { extend, find, keyBy, merge, set } from "lodash"
 import { ContainerModule, ContainerService } from "../../container/config"
 import { createIngressResources } from "./ingress"
 import { createServiceResources } from "./service"
-import { waitForResources } from "../status/status"
+import { waitForResources, compareDeployedObjects } from "../status/status"
 import { apply, deleteObjectsByLabel } from "../kubectl"
 import { getAppNamespace } from "../namespace"
 import { PluginContext } from "../../../plugin-context"
 import { KubeApi } from "../api"
 import { KubernetesProvider, KubernetesPluginContext } from "../config"
 import { configureHotReload } from "../hot-reload"
-import { KubernetesResource } from "../types"
+import { KubernetesResource, KubernetesServerResource } from "../types"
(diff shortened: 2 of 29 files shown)


garden-service/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -259,17 +369,27 @@ export async function createDeployment(
...
     service: service.name,
+    version: service.module.version.versionString,
+  }
+
+  let selector: any = {
+    matchLabels: {
+      service: service.name,
+    },
+  }
+  if (service.name !== "default-backend") {
```
edvald: Why this exception?

garden-service/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -57,6 +70,94 @@ export async function deployContainerService(params: DeployServiceParams<Contain
...
+
+      await waitForResources({
+        ctx: k8sCtx,
+        provider: k8sCtx.provider,
+        serviceName: service.name,
+        resources: manifests,
+        log,
+      })
+    }
+  }
```
edvald: Might be good to add some logging to indicate which "phase" we're in. Either as additional log lines, or by updating the parent log line.

10ko: Absolutely, already in the todo list. ;)

garden-service/src/plugins/kubernetes/container/deployment.ts
10ko: It's an old check, at the beginning I wanted to apply something only to the project namespace and we use the same logic to deploy `garden-system` resources. I'll delete it.

10ko:
Added with latest commits:
- better handling of Services: if a service change we now detect that and apply the full manifest instead of patching and we patch only if the version is what changed.
- Cleanup of old deployments: we now get rid (after blue-green is successfull) of any deployment not associated with a Service. This comes handy when we encounter some errror and deployments get created but not used. E.g. imagine deploying the "green" and then k8s hangs/fails on patching the Service. This will take care of any unused deployment on when you run `garden deploy` again.

I will write some tests and if nobody has any other input I would make the PR ready to be reviewed.
E.

garden-service/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -133,6 +274,11 @@ export async function createDeployment(
     valueFrom: { fieldRef: { fieldPath: "spec.serviceAccountName" } },
   })
 
+  env.push({
+    name: "VERSION",
```
edvald: We already have GARDEN_VERSION in there (by way of `runtimeContext.envVars`). Plus it's a good idea in general to prefix our own labels, env vars etc.

garden-service/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -260,17 +406,26 @@ export async function createDeployment(
 
 function deploymentConfig(service: Service, configuredReplicas: number, namespace: string): object {
 
-  const labels = {
+  let labels = {
     module: service.module.name,
     service: service.name,
+    version: service.module.version.versionString,
```
edvald: I think we should probably prefix the labels using `gardenAnnotationKey()`.

garden-service/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -260,17 +406,26 @@ export async function createDeployment(
 
 function deploymentConfig(service: Service, configuredReplicas: number, namespace: string): object {
 
-  const labels = {
+  let labels = {
```
edvald: Should still be const, right?

garden-service/test/unit/src/docs/config.ts
```diff
@@ -159,6 +159,8 @@ describe("config", () => {
   describe("generateConfigReference", () => {
     it("should return the correct markdown", () => {
       const { markdownReference } = renderConfigReference(portSchema)
+      console.log("---------------")
+      console.log(markdownReference)
```
edvald: Stray console logs

examples/deployment-strategies/README.md
```diff
@@ -0,0 +1,3 @@
+# Deployment strategies example
+
+A very basic demo project showing different strategies of deployments.
```
edvald: This could do with more explanation imo. For example, mention that this is to test the experimental (prob best to mention that it's experimental) new blue-green deployment strategy, and that it's based on the `demo-project` etc.

garden-service/src/plugins/kubernetes/container/deployment.ts
10ko: I believe that env is redundant.

garden-service/src/plugins/kubernetes/container/deployment.ts
10ko: Got it!

garden-service/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -58,6 +71,134 @@ export async function deployContainerService(params: DeployServiceParams<Contain
...
+      serviceName: `Deploy ${service.name}`,
+      resources: filteredManifests,
+      log,
+    })
+
+    // Patch for the current service to point to the new Deployment
+    const servicePatchBody = {
+      metadata: {
+        annotations: {
+          "garden.io/version": deploymentManifest.metadata.labels.version,
```
edvald: Better to use `[gardenAnnotationKey("version")]` here as well.

10ko: Of course!

PR #1143, opened 2019-09-02
fix: task/test results outputs are not shown

### LINKED ISSUE #1137: Garden not persisting task results
### Current Behavior

Running `garden get task-result` returns an empty response. This is likely not an issue with the command itself but rather that we're not persisting the task result. You'll also notice that running a task prints `undefined`. This also effects the Dashboard. 

### Expected behavior

Garden should persist task results and users should be able to retrieve them.

### Reproducible example

Go to `vote` example and run `garden get task-result` 

### Workaround

N/A

10ko:
This PR fixes an issue where the results of tasks/tests were not printed out on screen.
The fix itself is [this line](https://github.com/garden-io/garden/blob/fix-persistent-task-results/garden-service/src/logger/renderers.ts#L172) but a bit of refactoring was required.
Biggest difference now is how we store the `data` passed to a logger (it's now stored together with `msg, symbols, etc` in `messageStates[]`. 

Fixes #1137 

Only downside I can think of (which came to my mind while writing the comments above) is we now use the same mechanism for updating data that we use for `msg` etc. A potential issue could be we duplicate too much data (if `data` is a big object), but I didn't find a case where it might create problems.

yes, fixes the issue of not seeing test/task output


garden-service/test/unit/src/logger/log-entry.ts
```diff
@@ -97,6 +98,7 @@ describe("LogEntry", () => {
         section: "caesar",
         symbol: "info",
         status: "done",
+        data: undefined,
```
eysi09: Could you give this some value so that we can assert that it gets set when calling `setState`?

eysi09:
Actually, there's also the issue with `undefined` being printed in the console after running a task:

Screenshot 2019-09-02 at 22 15 15

eysi09:
Regarding the `undefined` thing, we can address that in a separate PR if that makes sense.

PR #1146, opened 2019-09-02
improvement: check static dir exists

### LINKED ISSUE #1127: Better error message when `static` dir is missing
Currently, Garden fails with `spawn git ENOENT` if it can't find the `static` directory. This can happen for users who install Garden manually and remove the `static` dir or otherwise don't co-locate it with the Garden binary.

This has caused confusion a few times. We should explicitly check for the `static` dir and throw a helpful error if it's not found.

dcharbonnier:
close #1127


garden-service/src/cli/cli.ts
```diff
@@ -315,6 +316,9 @@ export class GardenCli {
           // tslint:disable-next-line: no-floating-promises
           checkForUpdates(garden.globalConfigStore, headerLog)
 
+          // tslint:disable-next-line: no-floating-promises
```
edvald: This comment should be unnecessary.

garden-service/src/cli/helpers.ts
```diff
@@ -216,6 +217,15 @@ export function failOnInvalidOptions(argv, ctx) {
   }
 }
 
+export async function checkForStaticDir() {
+  if (!await pathExists(STATIC_DIR)) {
```
edvald: This is a common gotcha (I've been burned many times myself), but you need to put parentheses around `await pathExists(STATIC_DIR)`. Or put the result in a variable and then check that, maybe easier to read.

garden-service/src/cli/helpers.ts
```diff
@@ -216,6 +217,15 @@ export function failOnInvalidOptions(argv, ctx) {
...
+    throw new InternalError(
+      `The Garden static directory is missing, reinstall Garden.`,
```
edvald: The phrasing could be a touch more helpful. Something like "Could not find the static data directory. Garden is packaged with a data directory called 'static', which should be located next to your garden binary. Please try reinstalling, and make sure the release archive is fully extracted to the target directory."

garden-service/src/cli/helpers.ts
dcharbonnier: the precedence of the await and the not are the same with an associativity to the left so this is correct but for clarity I change it

Issue #630, opened 2019-03-14
Output logs while running task via `garden run task`


edvald:
The `run task` command is currently missing the option to run interactively, unlike the other `garden run` commands. We should add that option, especially since tasks are commonly run ad-hoc.

PR #1147, opened 2019-09-03
fix: running task would print undefined

### LINKED ISSUE #1137: Garden not persisting task results

10ko:
The shape of the `TaskResult` output has changed and the logger was printing undefined instead fo the `TaskResult` output

Fixes #1137 

### DIFF
--- a/garden-service/src/commands/run/task.ts
+++ b/garden-service/src/commands/run/task.ts
@@ -64,7 +64,9 @@ export class RunTaskCommand extends Command<Args, Opts> {
 
     if (!result.error) {
       log.info("")
-      log.info(chalk.white(result.output.output))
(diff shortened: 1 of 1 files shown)


edvald:
This is okay for now, although we do later need to update the command to stream logs while running the task. The task outputs may have different shapes for new providers or module types. Might be worth adding a comment to that effect.

See #630

10ko:
👍 Added comment referring to #630.

PR #1153, opened 2019-09-04
feat(config): add a number of new operators for template expressions

edvald:

This adds a variety of interesting possibilities for template string
expressions. The grammar is a reduced JavaScript grammar, with
some added restrictions on what is actually allowed (no insane
cross-type comparisons etc.). 

The whole list of operators is specified in the Configuration Files guide, along with some usage examples.

### DIFF
--- a/docs/using-garden/configuration-files.md
+++ b/docs/using-garden/configuration-files.md
@@ -486,9 +486,25 @@ For a practical use case, you might for example make a task that provisions some
 
 Different module types expose different output keys for their services and tasks. Please refer to the [module type reference docs](https://docs.garden.io/reference/module-types) for details.
 
-#### Conditionals
+#### Operators
 
-You can use conditional expressions in template strings, using the `||` operator. For example:
+You can use a variety of operators in template string expressions:
+
+* Arithmetic: `*`, `/`, `%`, `+`, `-`
+* Numeric comparison: `>=`, `<=`, `>`, `<`
+* Equality: `==`, `!=`
+* Logical: `&&`, `||`, ternary (`<test> ? <value if true> : <value if false>`)
+* Unary: `!` (negation), `typeof` (returns the type of the following value as a string, e.g. `"boolean"` or `"number"`)
+
+The arithmetic and numeric comparison operators can only be used for numeric literals and keys that resolve to numbers. The equality and logical operators work with any term.
+
+Clauses are evaluated in standard precedence order, but you can also use parentheses to control evaluation order (e.g. `${(1 + 2) * (3 + 4)}` evaluates to 21).
+
+These operators can be very handy, and allow you to tailor your configuration depending on different environments and other contextual variables.
+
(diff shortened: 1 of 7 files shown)


garden-service/src/template-string-parser.pegjs
```diff
@@ -13,45 +13,36 @@ TemplateString
...
 
 InvalidFormatString
   = Prefix? FormatStart .* {
-  	throw new options.TemplateStringError("Invalid template string: " + text())
+  	throw new options.TemplateStringError("Unable to parse as valid template string.")
   }
 
 FormatStart
-  = "${" ws
+  = "${" __
```
eysi09: I'm assuming this is intentional

eysi09: ...ah yes, noticed the change below.

garden-service/src/template-string-parser.pegjs
```diff
@@ -64,17 +55,113 @@ Key
...
+        })
+    }
+  / LogicalORExpression
+
+Expression
+  = ConditionalExpression
+
 // Much of the below is based on https://github.com/pegjs/pegjs/blob/master/examples/json.pegjs
-ws "whitespace" = [ \t\n\r]*
+__ "whitespace" = [ \t\n\r]*
```
eysi09: Is there are a reason you changed from `ws` to `__`. The former feels clearer out of context.

edvald: It's to match the examples that I'm borrowing from, and appears to be a convention.

edvald:
Added a few tests. We were already testing logical ORs, added some for ANDs.

edvald:
Realized I'd forgotten to add negation operators, added those just now. Also replaced the JSON lexical grammar we'd used with the full JS grammar.

docs/using-garden/configuration-files.md
```diff
@@ -494,6 +494,7 @@ You can use a variety of operators in template string expressions:
 * Numeric comparison: `>=`, `<=`, `>`, `<`
 * Equality: `==`, `!=`
 * Logical: `&&`, `||`, and ternary (`<test> ? <value if true> : <value if false>`)
+* Unary: `!` (negation), `typeof` (returns the type of the following value as a string, e.g. `"boolean"` or `"number"`)
```
eysi09: Add _and_ before typeof.

eysi09: Also, what use case did you had in mind for `typeof`?

edvald: I suppose I kinda left it in there as I was trimming stuff out of the JS parser, but one case I could think of was a bit forward thinking, for when we'll allow users to do casts. Specifying Kubernetes resource requirements could for example require some conversions, for which you could then use a ternary with a typeof check.

Re the copy, I think I'll actually remove "and" from the line above, since we just use commas in the other lines.

PR #1156, opened 2019-09-04
feat(config): add support for varFiles

### LINKED ISSUE #1093: Allow Garden to read variables from files
docker-compose lets you specify an env_file to decouple variables from the main configuration. Would be handy to let Garden do the same and make it easier for users migrating from docker-compose to adopt.

edvald:

This allows users to include a `garden.env` file for project-global
variables, and `garden.<env-name>.env` files for environment-specific
variables. You may also override the file paths for each of those.

These variables are made available in `${var.<key>}` template keys,
same as variables specified directly in the project garden.yml.

Closes #1093 

A full documentation guide for variables and templating will follow in a
separate PR.


docs/reference/config.md
```diff
@@ -132,6 +132,21 @@ environmentDefaults:
...
+Specify a path (relative to the project root) to a file containing variables, that we apply on top of the
+_environment-specific_ `variables` field. The file should be in a standard "dotenv" format, specified
+[here](https://github.com/motdotla/dotenv#rules).
+
+If you don't set the field and the `garden.<env-name>.env` file does not exist,
+we simply ignore it. If you do override the default value and the file doesn't exist, an error will be thrown.
+
+| Type     | Required | Default                  |
+| -------- | -------- | ------------------------ |
+| `string` | No       | `"garden.undefined.env"` |
```
thsig: Should this `undefined` be here?

edvald: Hmm no. Not sure what causes that... I'll fix it.

Issue #1097, opened 2019-08-13
Garden won't find modules that are git submodules


Decisional:
Hey there.

I have a simple meta-repository: The garden project level configuration resides in the root directory, modules are stored in git submodules that are mounted in this root. 

_meta # meta git repository
├── submodule
│   ├── Dockerfile
│   └── garden.yml
└── not-a-submodule
│   └── garden.yml
└── garden.yml

When adding a garden module configuration in one of the submodules,  garden won't find them with `garden scan` from the root directory. Simply `cp submodule/garden.yml not-a-submodule/` will make garden pick up the configuration from the folder that's not a submodule. Adding the not-a-submodule directory to .gitignore doesn't change anything.

Any thoughts? :confused:

edvald:
Ah, I think that's a bug. Sorry about that!

cc @eysi09. I guess we should be running `ls-files` with `--recurse-submodules`: https://stackoverflow.com/questions/36848544/git-finding-a-list-of-files-e-g-using-git-ls-files-including-submodules

gen1321:
I have same issue

PR #1165, opened 2019-09-06
fix(vcs): overflow error when repo contains large number of files

### LINKED ISSUE #1162: GitHandler.getFiles() crashes on very long file list
### Current Behavior

If your project directory contains a very large list of files, _even if they are marked to be excluded_, the `GitHandler.getFiles()` method will crash.

### Expected behavior

A long list of files in a repo should not cause any problems.

### Reproducible example

I don't have a repo with it, but you can create a bare-bones project and install a huge number of large npm packages...

### Workaround

Sadly the only current workaround is to remove the large directories. In many cases it's okay because those directories will be installed in the build staging folder automatically, but it's not ideal, hence the `priority:high` tag.

### Suggested solution(s)

We need to revisit both how we list those files (ideally we don't have to list ignored files to begin with) and also to stream the result instead of buffering.

### Your environment

Reported on v0.10.8 and reproduced on latest master.

edvald:

We now run fewer git commands to list files in a directory, and avoid
having to list every file before later filtering out ignored files.
We also use streaming to avoid any buffering issues, and for improved
efficiency.

Fixes #1162

### DIFF
--- a/garden-service/src/vcs/git.ts
+++ b/garden-service/src/vcs/git.ts
@@ -8,10 +8,11 @@
 
 import * as execa from "execa"
 import { join, resolve } from "path"
-import { flatten, uniq } from "lodash"
+import { flatten } from "lodash"
 import { ensureDir, pathExists, stat, createReadStream } from "fs-extra"
 import { PassThrough } from "stream"
 import * as hasha from "hasha"
+import split2 = require("split2")
 
 import { VcsHandler, RemoteSourceParams, VcsFile, GetFilesParams } from "./vcs"
 import { ConfigurationError, RuntimeError } from "../exceptions"
@@ -54,7 +55,7 @@ export class GitHandler extends VcsHandler {
   private gitCli(log: LogEntry, cwd: string): GitCli {
     return async (...args: string[]) => {
       log.silly(`Calling git with args '${args.join(" ")}`)
-      const { stdout } = await execa("git", args, { cwd })
+      const { stdout } = await execa("git", args, { cwd, maxBuffer: 10 * 1024 * 1024 })
       return stdout.split("\n").filter(line => line.length > 0)
     }
   }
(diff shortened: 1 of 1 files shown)


eysi09:
Does the change anything for #1097?

garden-service/src/vcs/git.ts
```diff
@@ -74,60 +75,86 @@ export class GitHandler extends VcsHandler {
...
-
-        return [lsIgnoredFiles, lsUntrackedIgnoredFiles]
-      }))
-
-      ignored = uniq(flatten(await Bluebird.map(commands, async (cmd) => git(...cmd, path))))
-    } catch (err) {
-      // if we get 128 we're not in a repo root, so we get no files
-      if (err.exitCode !== 128) {
-        throw err
+    // List modified files, so that we can ensure we have the right hash for them later
```
eysi09: It might be useful to document the sequencing above the function since the control flow is a little hard to follow.

garden-service/src/vcs/git.ts
```diff
@@ -74,60 +75,86 @@ export class GitHandler extends VcsHandler {
...
+    const trackedButIgnored = new Set(flatten(
+      await Promise.all(this.ignoreFiles.map(f =>
+        git("ls-files", "--ignored", "--exclude-per-directory", f),
+      )),
+    ))
+
+    // We run ls-files for each ignoreFile and do a manual set-intersection (by counting elements in an object)
+    // in order to optimize the flow.
+    const paths: { [path: string]: number } = {}
+    const output: VcsFile[] = []
```
eysi09: Rename to `files` or similar?

edvald:
No, this doesn't address that issue.

garden-service/src/vcs/git.ts
eysi09: Regardless, we should at least add a comment saying that this function returns files filtered against ignore files, includes and excludes, since the name is pretty generic.

PR #1169, opened 2019-09-09
chore: use prettier for stricter format checking

edvald:
I'm unsure if we want to do this actually. Take a look and see if the style is to your liking.

We might go for alternatives that also do auto-formatting, but in a slightly less opinionated way.


eysi09:
I can live with the styling tbh. In any case I'd most definitely want to use it for the dashboard.

garden-service/src/commands/link/module.ts
```diff
@@ -64,15 +66,19 @@ export class LinkModuleCommand extends Command<Args> {
...
+        .filter(hasRemoteSource)
+        .sort()
 
       throw new ParameterError(
-        `Expected module(s) ${chalk.underline(moduleName)} to have a remote source.` +
-        ` Did you mean to use the "link source" command?`,
+        `Expected module(s) ${chalk.underline(
+          moduleName
+        )} to have a remote source.` +
+          ` Did you mean to use the "link source" command?`,
```
edvald: This feels very awkward to me, for example.

garden-service/src/commands/link/module.ts
```diff
@@ -53,7 +50,12 @@ export class LinkModuleCommand extends Command<Args> {
         garden link module my-module path/to/my-module # links my-module to its local version at the given path
   `
 
-  async action({ garden, log, headerLog, args }: CommandParams<Args>): Promise<CommandResult<LinkedSource[]>> {
+  async action({
+    garden,
+    log,
+    headerLog,
+    args,
+  }: CommandParams<Args>): Promise<CommandResult<LinkedSource[]>> {
```
edvald: Not a big fan of this, but could live with it.

garden-service/src/plugins/kubernetes/container/build.ts
```diff
@@ -57,16 +71,20 @@ export async function k8sGetContainerBuildStatus(
...
   const { ctx } = params
   const provider = <KubernetesProvider>ctx.provider
   const handler = buildHandlers[provider.config.buildMode]
   return handler(params)
 }
 
-type BuildStatusHandler = (params: GetBuildStatusParams<ContainerModule>) => Promise<BuildStatus>
+type BuildStatusHandler = (
+  params: GetBuildStatusParams<ContainerModule>
+) => Promise<BuildStatus>
```
edvald: This is a bit odd imo.

garden-service/src/commands/link/module.ts
eysi09: Yeah this is bad. I had a hard time just parsing it.

garden-service/src/commands/link/module.ts
eysi09: Same.

eysi09:
@thsig @10ko @ellenkorbes What are your thoughts on this?

Note that if we don't go with Prettier we can still improve our linter settings to the point that unformatted code should fail CI checks.

And _if_ most the lint rules are fixable and we configure our editors to fix on save, we basically end up with the same thing but with way more control over the formatting. Note also that we'll migrate from tslint to eslint since the former is being deprecated.

garden-service/src/cli/helpers.ts
```diff
@@ -79,41 +80,59 @@ export function helpTextMaxWidth() {
 
 // Add platforms/terminals?
 export function envSupportsEmoji() {
-  return process.platform === "darwin"
-    || process.env.TERM_PROGRAM === "Hyper"
-    || process.env.TERM_PROGRAM === "HyperTerm"
+  return (
+    process.platform === "darwin" ||
+    process.env.TERM_PROGRAM === "Hyper" ||
+    process.env.TERM_PROGRAM === "HyperTerm"
```
10ko: This is very annoying

garden-service/src/cli/helpers.ts
```diff
@@ -79,41 +80,59 @@ export function helpTextMaxWidth() {
...
+    {}
+  )
 }
 
 // Sywac specific transformers and helpers
-export function getOptionSynopsis(key: string, { alias }: Parameter<any>): string {
+export function getOptionSynopsis(
+  key: string,
+  { alias }: Parameter<any>
+): string {
```
10ko: Also this, I don't like it.

thsig:
I'd rather first see how far we can get by just adding/changing rules in tslint.

That said, I don't have a strong opinion on this.

ellenkorbes:
I really, really, really prefer Prettier. 

I don't know whether it's because I don't know what I'm talking about or because I don't have any pre-existing biases as to what TS should look like but... To me the way Prettier expands every clump of code and adds negative space to the page makes things so much easier to parse at a glance and understand what's going on without having to squint at the screen.

IMO although some things about it are annoying, they'd fade into background noise in a couple of days, and the upside is totally worth it. 

In addition, we can start using Prettier in its "final state" right away as opposed to having to spend time tweaking eslint and bikeshedding about its configs for who knows how long.

edvald:
I guess it is more frustrating when you've used a certain style for a very long time (I've been using TS for three years and JS for much longer). The bikeshedding point is valid though.

edvald:
@eysi09 I did update that, albeit only for garden-service. Looks a little better now, yes. And I gave eslint formatting a spin and I aborted that whole thing, too messy and complicated. So I could be swayed to go for this... Let's call a vote.

eysi09:
Shouldn't we use the same config for both dashboard and garden-service?

edvald:
Yes we should. There are also other things I need to fix/update.

eysi09:
@edvald did you make the updates mentioned above?

dashboard/.prettierrc.json
```diff
@@ -0,0 +1,8 @@
+{
+  "arrowParens": "always",
+  "printWidth": 120,
+  "quoteProps": "consistent",
+  "semi": false,
+  "trailingComma": "es5",
+  "tabWidth": 2
+}
```
dcharbonnier: this PR was almost good, the only thing was to remove this file...

PR #1187, opened 2019-09-13
chore: fix TOC when no README is present

### LINKED ISSUE #1180: Table of contents generator: Handle dirs with no READMEs
The table of contents generator currently assumes that all directories have a `README.md` file which won't always be the case. We should fix this. See this PR for details: #1179 for more details.

ellenkorbes:

Now when there's no `README.md` inside a subdirectory the link for that directory in the table of contents will:
- Be title cased
- Point to the first file in the directory (before custom/metadata sorting is applied)

Fixes #1180


garden-service/src/docs/table-of-contents.ts
```diff
@@ -55,6 +56,17 @@ function attachMetadata(tree: FileTree) {
     if (e.code !== "ENOENT") {
       throw (e)
     }
+    if (e.code === "ENOENT" && tree.type === "directory") {
+      let checkFile: string | undefined
+      try {
+        checkFile = readFileSync(tree.children[0].path, "utf-8")
```
edvald: Using `pathExists()` here would be much cleaner.

garden-service/package.json
```diff
@@ -103,6 +103,7 @@
     "sywac": "^1.2.2",
     "tar": "^4.4.10",
     "terminal-link": "^2.0.0",
+    "title-case": "^2.1.1",
```
eysi09: We actually also have `humanize` and `titleize` (imported below) and `titleize(humanize(string))` is basically the same as `titleCase(string)`. I suggest we either remove the `title-case` library and add a `titleCase` util function in `util/string` that uses `titleize` and `humanize`. Or alternatively, remove `titleize` and `humanize` and just use `title-case`.

garden-service/src/docs/table-of-contents.ts
ellenkorbes: Turns out that checking is not necessary since dtree does that for me already elsewhere in the code.

garden-service/src/docs/table-of-contents.ts
```diff
@@ -46,14 +49,21 @@ function createNewTree(tree: FileTree, transform: Function): FileTree {
 
 function attachMetadata(tree: FileTree) {
   if (tree.type === "directory") {
-    tree.path = tree.path + "/README.md"
+    if (tree.children.length > 0) {
+      tree.path = tree.path + "/README.md"
```
edvald: Better to use `resolve(tree.path, "README.md")`, otherwise it won't work cross platform.

ellenkorbes:
@edvald @eysi09 Please review again.

garden-service/src/docs/table-of-contents.ts
```diff
@@ -117,6 +141,10 @@ const indent: string = "  "
 
 export function generateTableOfContents(docsRoot: string): string {
   const rawTree = <FileTree>dtree(docsRoot, { extensions: /\.md/ })
+  if (rawTree === null) {
+    throw ("Directory not found.")
```
eysi09: This should be `throw new Error("...")`

eysi09: This should be `throw new Error("...")`

ellenkorbes: @eysi09 Done, please merge :-)

ellenkorbes:
@eysi09 Done, please merge :-)

PR #1212, opened 2019-09-19
improvement(logger): skip fancy rendering when log level > info

eysi09:

With this PR, the fancy writer just prints log messages directly if the log level is above `info`, as opposed to updating nodes or adding spinner.

This is because it's hard to parse the log if it contains a lot of text that gets updated in different places.

I also used to chance removed some old code that was written for uses cases that we've never really needed.


eysi09:
I want to get the Dashboard patch in before we merge this. I've flagged as WIP.

PR #1213, opened 2019-09-19
Fix hot reload issues

edvald:

See the individual commits. I tried a few things and made some logging and type-safety improvements to get to the bottom of the issues, but the actual fix is the last commit. That resolves both (thus far unreported) issues with non-Deployment type workloads, as well issues with exec-ing and hot-reloading after the blue-green PR was merged. So I don't expect users will have run into issues, but at least the code is more robust now than before.

Please test hot-reloading and ad-hoc exec-ing/running commands, for a couple of different module types.

### DIFF
--- a/garden-service/src/plugins/kubernetes/container/deployment.ts
+++ b/garden-service/src/plugins/kubernetes/container/deployment.ts
@@ -7,28 +7,27 @@
  */
 
 import chalk from "chalk"
-import Bluebird from "bluebird"
 import { V1Container } from "@kubernetes/client-node"
-import { Service, ServiceStatus } from "../../../types/service"
+import { Service } from "../../../types/service"
 import { extend, find, keyBy, merge, set } from "lodash"
 import { ContainerModule, ContainerService } from "../../container/config"
 import { createIngressResources } from "./ingress"
 import { createServiceResources } from "./service"
 import { waitForResources, compareDeployedObjects } from "../status/status"
-import { apply, deleteObjectsByLabel } from "../kubectl"
+import { apply, deleteObjectsBySelector } from "../kubectl"
 import { getAppNamespace } from "../namespace"
 import { PluginContext } from "../../../plugin-context"
 import { KubeApi } from "../api"
 import { KubernetesProvider, KubernetesPluginContext } from "../config"
 import { configureHotReload } from "../hot-reload"
-import { KubernetesResource, KubernetesServerResource } from "../types"
+import { KubernetesWorkload } from "../types"
 import { ConfigurationError } from "../../../exceptions"
 import { getContainerServiceStatus, ContainerServiceStatus } from "./status"
 import { containerHelpers } from "../../container/helpers"
 import { LogEntry } from "../../../logger/log-entry"
 import { DeployServiceParams } from "../../../types/plugin/service/deployService"
 import { DeleteServiceParams } from "../../../types/plugin/service/deleteService"
-import { millicpuToString, kilobytesToString, prepareEnvVars } from "../util"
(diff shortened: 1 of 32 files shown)


garden-service/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -236,9 +214,9 @@ interface CreateDeploymentParams {
   log: LogEntry,
 }
 
-export async function createDeployment(
+export async function createWorkloadResource(
```
10ko: Why the change of name?

garden-service/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -175,32 +162,22 @@ export async function deployContainerServiceBlueGreen(
...
+    await deleteObjectsBySelector({
+      log,
+      provider,
+      namespace,
+      objectTypes: workloadTypes,
+      // Find workloads that match this service, but have a different version
+      selector:
+        `${gardenAnnotationKey("service")}=${service.name},` +
+        `${versionKey}!=${newVersion}`,
+    })
```
10ko: We don't `waitForResources` anymore. Is it intentional?

garden-service/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -46,17 +47,16 @@ export async function deployContainerService(params: DeployServiceParams<Contain
...
 
   const namespace = await getAppNamespace(k8sCtx, log, k8sCtx.provider)
 
-  const manifests = await createContainerObjects(k8sCtx, log, service, runtimeContext, hotReload)
+  const { manifests } = await createContainerManifests(k8sCtx, log, service, runtimeContext, hotReload)
 
-  // TODO: use Helm instead of kubectl apply
   const provider = k8sCtx.provider
-  const pruneSelector = "service=" + service.name
+  const pruneSelector = gardenAnnotationKey("service") + "=" + service.name
```
10ko: I not sure if this apply but in [service.ts](https://github.com/garden-io/garden/blob/master/garden-service/src/plugins/kubernetes/container/service.ts#L28) the label service is not prefixed.
Is that what the `--prune --selector` matches against or am I misunderstanding how that works?

garden-service/src/plugins/kubernetes/container/deployment.ts
edvald: Because we create a DaemonSet when you set `daemon: true`, so the name was misleading

garden-service/src/plugins/kubernetes/container/deployment.ts
edvald: The delete function (which just uses `kubectl delete` atm) waits until the resources are deleted.

But wait, I thought the `waitForResources` had something to do with the cleanup, because it was within the `if (unusedDeployments) {` block. I take it that it was supposed to run no matter what?

garden-service/src/plugins/kubernetes/container/deployment.ts
edvald: We should update that function, but it's equivalent because `gardenAnnotationKey` currently returns "module" and "service" unchanged (see comment in there).

garden-service/src/plugins/kubernetes/container/deployment.ts
10ko: No misunderstanding, that was the expected behaviour. I just didn't we were using `kubectl delete` underneath. 👍

10ko:
Looks good to me, I am not sure if the tests are failing for the usual flakiness or because errors.
Once that is cleared, LGTM.

PR #1214, opened 2019-09-19
improvement(config): add linkUrl field

### LINKED ISSUE #1185: Garden displays http URLs even when they aren't applicable
### Current Behavior

On deploying an app, Garden prints a URL to access it. In our case, the app's ingress controller is itself fronted by a load balancer providing (and requiring) HTTPS, so an HTTP URL is incorrect. This has caught a few of our devs by surprise, since loading that HTTP URL makes it look like Garden isn't working. This would also be relevant in cases where the load balancer is rewriting paths, or deploying apps to the same URL but routing to different backends based on cookies.

### Suggested solution(s)

Garden allows overriding/templating the displayed URL, which may be different than the ingress `hostname:` field.

eysi09:
EDIT: Depends on #1234. Please review and merge that first. (This is because that one contains fixes that I hoped resolved the build error we were seeing. They don't. See comment below).

Allows the user to set the display URL for a service ingress that Garden prints in the console, displays in the Dashboard, and uses for the `call` command.

We need this because the URL is different from what Garden constructs from the ingress spec, e.g. because the user has a load balancer in front of the services that re-writes paths or secures them.

Closes #1185 

We've debated the name for the field and might still want to change it. Here are some contenders:

- `ingressUrl`
- `linkUrl`
- `actualUrl`

Basically, the opposite of the inferred URL (cc @thsig).

### DIFF
--- a/garden-service/src/types/service.ts
+++ b/garden-service/src/types/service.ts
@@ -80,6 +80,7 @@ export type ServiceProtocol = "http" | "https"  // | "tcp" | "udp"
 
 export interface ServiceIngressSpec {
   hostname?: string
+  linkUrl?: string
   path: string
   port: number
   protocol: ServiceProtocol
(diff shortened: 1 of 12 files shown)


garden-service/src/types/service.ts
```diff
@@ -98,6 +99,18 @@ export const ingressHostnameSchema = joi.string()
     Note that if you're developing locally you may need to add this hostname to your hosts file.
   `)
 
+export const displayUrlSchema = joi.string()
+  .uri()
+  .description(dedent`
+    The display URL for the ingress to show in the console and on the dashboard.
+    Also used when calling the service with the \`call\` command
```
edvald: Missing period.

eysi09:
TBH I think something `[actual|real|true]Url` conveys the meaning best. But they just feels like bad variable names. But I agree that `linkUrl` is better than `displayUrl`.

eysi09:
I've changed the field name to `linkUrl`.

eysi09:
I've updated the commit. Now we duplicate the `getLinkUrl` function in the dashboard and import it from there. This resolves the build error we were seeing. For a full discussion, see #1240.

PR #1231, opened 2019-09-25
chore: remove gulp task to update Homebrew Formula

10ko:
Since the we tested the github action to update the Homebrew Formula succesfully, we can now safely remove the old `gulp` task we used to do that.


CONTRIBUTING.md
```diff
@@ -230,11 +230,8 @@ To make a new release, set your current working directory to the garden root dir
     * To get a list of all contributors between releases, ordered by count, run: `git log <previous-tag>..<current-tag> --no-merges "$@" | grep ^Author | sort | uniq -c | sort -nr`. Note that authors of squashed commits won't show up so it might be good to do a quick sanity check on Github as well.
 7. Click the **Publish release** button.
 8. Make a pull request for the branch that was pushed by the script.
-9. **Update the Homebrew package: `gulp update-brew`.**
-10. Install the Homebrew package and make sure it works okay:
```
edvald: I think I'd like to keep this one around.

gulpfile.ts
```diff
@@ -6,21 +6,12 @@
...
  */
 
-import { resolve } from "path"
-import execa from "execa"
-import { writeFile, readFile, ensureDir, pathExists, remove } from "fs-extra"
-import { getUrlChecksum } from "./support/support-util"
-import handlebars from "handlebars"
-import { find } from "lodash"
-
-const Octokit = require("@octokit/rest")
```
edvald: We should remove packages we don't need anymore from the package.json.

CONTRIBUTING.md
10ko: Ok but FYI the github action runs on mac and installs/runs `garden --version`.

gulpfile.ts
10ko: Ops, I would have sweared we use Oktokit somewhere else.
Removed

10ko: Ops, I would have sweared we use Oktokit somewhere else.
Removed

eysi09: We could try this: https://www.npmjs.com/package/depcheck

PR #1233, opened 2019-09-25
chore(dashboard): update react scripts to version 3.1.2

eysi09:

Updates the Dashboard react-scripts deps. 

It was necessary to re-configure the lint setup and to address some lint
errors that came up, most notably the react-hooks/exhaustive-deps. To
address this we added the useMountEffect helper function.


tslint.base.json
```diff
@@ -6,6 +6,11 @@
   "rulesDirectory": [
     "node_modules/tslint-microsoft-contrib"
   ],
+  "linterOptions": {
+    "exclude": [
+      "./dashboard/**/*"
+    ]
+  },
```
edvald: I don't think this is necessary, since we don't lint from the root project anymore.

eysi09: I couldn't get linting to work in VSCode until I removed this. I'll went back on forth on some configuration options though so I maybe something else was configured differently when it wasn't working. I'll verify and report back.

eysi09: So yeah, if we don't exclude the dashboard, we get lint warnings in VSCode from `tslint.base.json`. But the dashboard doesn't use those lint rules, in fact in can't. Instead we define our own lint rules for the dashboard. I've made some changes to those rules in #1234.

eysi09: Actually, I've found away around this in #1234. But it requires some other changes so I'll leave this as is and address it there.

PR #1237, opened 2019-09-27
docs: mention that remote builders don't support private registries

eysi09:

This was missing from our docs and has caused confusion among users.

### DIFF
--- a/docs/using-garden/in-cluster-building.md
+++ b/docs/using-garden/in-cluster-building.md
@@ -140,3 +140,19 @@ The command does the following:
 
 There are plans to do this automatically when disk-space runs low, but for now you can run this manually or set up
 your own cron jobs.
+
+## Pulling base images from private registries
+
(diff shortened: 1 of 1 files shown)


docs/using-garden/in-cluster-building.md
```diff
@@ -140,3 +140,11 @@ The command does the following:
 
 There are plans to do this automatically when disk-space runs low, but for now you can run this manually or set up
 your own cron jobs.
+
+## Pulling images from private registries
+
+Currently, only the _Local Docker_ build mode supports images from private registries. If you see an `ImagePullBackOff` error when using the other build modes, it's likely that it's failing because the image in question is in a private registry.
+
+We do plan on supporting private registries for more build modes but it's a non-trivial feature to add. You can track the discussion and progress [in this issue](https://github.com/garden-io/garden/issues/1236).
+
```
edvald: This is not quite accurate. It needs to be clear that _base images_ from private registries, in user's Dockerfiles, are currently not explicitly supported. If you're just referencing private registry images in your container module config (`image: my-registry.example.com/foo/bar`) or in your Helm/Kubernetes modules, the `imagePullSecrets` field works fine.

eysi09: I've updated the commit.

PR #1239, opened 2019-09-28
refactor(core): only set config names once in resolveModuleConfigs

eysi09:

Config names were being set twice when resolving module configs. Assuming this was a rebase error.


edvald:
Feels like we should add a quick test for this (which I should have done when I did that refactor).

eysi09:
How would you test for this? It was literally just setting the values twice, just using a helper function the second time.

PR #1248, opened 2019-10-02
fix(commands): fix regression due to changes to test|task result output

### LINKED ISSUE #1224: Get task|test result commands don't return correct results
### Current Behavior

We've changed the shape of the outputs of `getTestResutl` and `getTaskResult` actions but the command still expects the old shape. This means that the `output` field is always null when the command is called. These results won't load in the dashboard either.

### LINKED ISSUE #1225: Task state changes for test nodes not visible in Dashboard
### Current Behavior

The dashboard is not catching events for graph nodes of type test. Therefore the stack graph won't update for those nodes.

eysi09:

Fixes a regression due to changes we made to shape of the output of task and test results. This caused the `get-task-result` and `get-test-result` commands to return empty output log. This also caused these not to load in the dashboard.

Fixes #1225
Fixes #1224

### DIFF
--- a/dashboard/src/api/actions.ts
+++ b/dashboard/src/api/actions.ts
@@ -11,9 +11,9 @@ import { groupBy } from "lodash"
 
 import { ServiceLogEntry } from "garden-service/build/src/types/plugin/service/getServiceLogs"
 import { StatusCommandResult } from "garden-service/build/src/commands/get/get-status"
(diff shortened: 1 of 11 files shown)


eysi09:
@thsig I've updated the PR. I just forced pushed since the changes only apply to the test files.

PR #1255, opened 2019-10-04
docs: ensure check-docs script checks internal header links

### LINKED ISSUE #1249: docs: Project/Modules links don't work in "Config Files Reference" page
In the first sentence in https://docs.garden.io/reference/config#project-configuration.
>Below is the schema reference for the *Project* and *Module* garden.yml configuration files.

*Project* and *Modules* are anchor links but don't point to anything.

eysi09:

`markdown-check-link` doesn't check internal header links. This is a
known issue (https://github.com/tcort/markdown-link-check/issues/65).

We therefore add a new tool to check for internal links and limit
markdown-check-link to external links.

The new tool is from the remark family of markdown tools and expects
anchor links to not include any special characters so we also had to
make appropriate changes to the docs generator functions.

Finally, this commit fixes broken links discovered because of these
fixes.

Fixes #1249

### DIFF
--- a/bin/check-docs
+++ b/bin/check-docs
@@ -1,20 +1,35 @@
 #!/bin/bash -e
 
-# Linter before git check so we can lint docs even if they are 'dirty'
-
 # Use "|| true" so we don't exit on empty
 modified_docs=$(git diff --name-status master docs README.md) || true
 modified_examples=$(git diff --name-status master examples | grep "examples.*\README.md$") || true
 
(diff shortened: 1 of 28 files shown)


eysi09:
Yeah I looked into switching and tried out `remark-lint-no-dead-url`. However it doesn't support ignores like `markdown-link-check` which we use for pseudo URLs like `http://my-service`.

PR #1256, opened 2019-10-05
refactor(plugins): formalize plugin and module type extension mechanisms

edvald:

This adds framework native abilities to
A) define a plugin from a base plugin, and
B) explicitly create or extend module types.

Previously, there was no built-in notion of "base" plugins, so plugins
that wanted to duplicate another plugin had to do that directly at the
code level. This also meant that Garden had no idea whether two plugins
would be compatible, so it was not possible for a plugin to
depend on a base plugin and work with any plugins that inherit it.
For a practical example, that was why we needed a `local-openfaas`
plugin, that paired with the `local-kubernetes` plugin. Now any plugin
that depends on the `kubernetes` plugin _also_ works with any plugin
that inherits from that plugin, e.g. the `local-kubernetes` plugin.

We ensure this by enforcing that inheriting plugins use a configuration
schema that is compatible with the base plugin, either by making its
schema a superset of the base's schema, or by having its
`configureProvider` handler return a config that matches the base's
schema.

It is also now explicit whether a plugin is creating or extending a
module type, and we add a bunch of validation to
ensure it's all being done correctly.

This is arguably a feature commit, but I'm categorizing it as a refactor
because it doesn't immediately introduce user-facing features, and
rather opens the door for creating more plugins.

For example, we can now readily make more "special-case" versions of the
`kubernetes` provider, e.g. one for GKE, another for EKS and so on.

We can also make a variety of plugins that depend on the `kubernetes`
plugin, e.g. a Linkerd plugin, monitoring plugins etc., that can then
work with any flavor of the `kubernetes` plugin.

A following commit will add the same type of inheritance features to
module types, with analogous benefits.

Also, a couple of smaller/simpler refactors.

I suggest looking at commit by commit to make it easier to parse.
The big one is the 3rd one (formalize plugin refactor).

For reviewing that, I strongly recommend looking at the added tests first,
since they describe the intended behavior. Also look at the changes
made to the `openfaas` example, the `local-kubernetes` plugin and
the (now deprecated) `local-openfaas` plugin to see the changes in
practical context.

A plugin dev guide will follow once the remaining refactors are done.


garden-service/src/types/plugin/plugin.ts
```diff
@@ -207,53 +201,130 @@ export const moduleActionDescriptions:
...
 }
 
-export interface PluginFactory {
-  (params: PluginFactoryParams): GardenPlugin | Promise<GardenPlugin>
-}
-export type RegisterPluginParam = string | PluginFactory
-export interface Plugins {
-  [name: string]: RegisterPluginParam
-}
+export interface GardenPlugin extends GardenPluginSpec { }
```
eysi09: Why do we have both `GardenPlugin` and `GardenPluginSpec`?

garden-service/src/types/plugin/plugin.ts
```diff
@@ -204,64 +260,203 @@ export const moduleActionDescriptions:
...
+    name: joiIdentifier()
+      .required()
+      .description("The name of module type."),
+    handlers: joi.object().keys(mapValues(moduleActionDescriptions, () => joi.func()))
+      .description("A map of module action handlers provided by the plugin."),
+  })
+
+const createModuleTypeSchema = extendModuleTypeSchema
+  .keys({
+    // base: joiIdentifier()
```
eysi09: Assuming this is introduced in the "plugin inheritance" PR, right?

garden-service/src/types/plugin/plugin.ts
edvald: It ended being temporarily superfluous, since `createGardenPlugin()` doesn't really do anything atm, but the distinction was meant to be between the input and output of that function.

garden-service/src/types/plugin/plugin.ts
edvald: The module type inheritance, yes.

eysi09:
Only had a couple of small comments.

garden-service/src/util/util.ts
```diff
@@ -396,3 +396,14 @@ export function hashString(s: string, length: number) {
   urlHash.update(s)
   return urlHash.digest("hex").slice(0, length)
 }
+
+/**
+ * Ensures that `obj` has an array at `key`, creating it if necessary, and then pushes `value` on that array.
+ */
+export function pushToKey(obj: object, key: string, value: any) {
+  if (obj[key]) {
+    obj[key].push(value)
```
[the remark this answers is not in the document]
edvald: I suppose that's not explicitly caught, I can add a quick check and throw if not

garden-service/src/plugins/openfaas/local.ts
```diff
@@ -6,14 +6,12 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import { GardenPlugin } from "../../types/plugin/plugin"
+import { createGardenPlugin } from "../../types/plugin/plugin"
 import { gardenPlugin as o6sPlugin } from "./openfaas"
```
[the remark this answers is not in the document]
edvald: Fair point.

PR #1260, opened 2019-10-10
fix(vcs): no files were found when dotIgnoreFiles was set to empty list

edvald:

If you set `dotIgnoreFiles: []` in your project config, no modules would be found.

Just came across this myself.

### DIFF
--- a/garden-service/src/vcs/git.ts
+++ b/garden-service/src/vcs/git.ts
@@ -60,7 +60,7 @@ export class GitHandler extends VcsHandler {
 
   private gitCli(log: LogEntry, cwd: string): GitCli {
     return async (...args: string[]) => {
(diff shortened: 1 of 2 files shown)


garden-service/src/vcs/git.ts
```diff
@@ -166,16 +166,20 @@ export class GitHandler extends VcsHandler {
       // We push to the output array when all ls-files commands "agree" that it should be included,
       // and it passes through the include/exclude filters.
       if (
-        paths[resolvedPath] === this.ignoreFiles.length
+        paths[resolvedPath] >= this.ignoreFiles.length
```
eysi09: Why this change?

edvald: Otherwise the fix doesn't work when there are no ignore files, since the count (1) is higher than the number of ignore files (0).

PR #1262, opened 2019-10-10
feat(plugins): add local flag to exec module type

eysi09:

When `local` is set to `true` for an `exec` module,  the build command, tests and tasks are run in
the module directory, as opposed in the .garden/build directory.

This can be useful e.g. for Java users that want to build the `.jar` before adding it to the container image.

Check out the example to see how this works in action.


docs/reference/module-types/exec.md
```diff
@@ -7,6 +7,9 @@ title: Exec
 A simple module for executing commands in your shell. This can be a useful escape hatch if no other module
 type fits your needs, and you just need to execute something (as opposed to deploy it, track its status etc.).
 
+By default, the `exec` module type executes the commands in the Garden build directory. By setting `local: true`,
+the commands are executed in the same directory as the module itself.
```
edvald: This phrasing is slightly clumsy ("the same directory as the module itself"). How about "... in the module source directory instead."?

docs/reference/module-types/exec.md
```diff
@@ -7,6 +7,9 @@ title: Exec
 A simple module for executing commands in your shell. This can be a useful escape hatch if no other module
 type fits your needs, and you just need to execute something (as opposed to deploy it, track its status etc.).
 
+By default, the `exec` module type executes the commands in the Garden build directory. By setting `local: true`,
```
edvald: Maybe add to clarify: "... in the Garden build directory (under `.garden/build/<module name>`)

examples/local-exec/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+> Note: You need to have Go installed to run this project.
+
+This example project demonstrates how you can use the `exec` module type to run build commands, tasks and tests in the module directory, by setting `local: true` in the module config. By default the commands are executed in the `.garden/build` directory.
+
+The idea is to use a local exec module to run pre-build commands for a container module.
+
+## Project Structure
+
+The project consists of a `local-executor` module, a `builder` module, and a `backend` module.
```
edvald: I wonder if the `local-executor` module (which is an odd name to begin with) and task is unnecessary for the example? It just confuses, I think.

garden-service/test/unit/data/test-project-exec/module-local/garden.yml
```diff
@@ -0,0 +1,9 @@
+kind: Module
+name: module-local
+type: exec
+local: true
+build:
+  command: [pwd]
```
edvald: I'd suggest having the build command create a file (which is then gitignored) and check for its existence in the right place.

edvald:
Had a few line comments, but I'm also wondering:
1. I'm unsure if removing the `trim()` from command outputs is a good idea. The idea is that people can use outputs in template strings, and having a line break at the end is both tricky in those cases, and also may break compatibility with existing code.
2. Have we considered what happens with modules that copy _from_ local exec modules? Wouldn't we need to special-case that somehow?

eysi09:
> I'm unsure if removing the trim() from command outputs is a good idea. The idea is that people can use outputs in template strings, and having a line break at the end is both tricky in those cases, and also may break compatibility with existing code.

Yeah good point. I'll revert that. I does look like we only do this for the `exec` module type though. So task outputs for other module types will presumably have a new line at the end, as well as all test outputs.

> Have we considered what happens with modules that copy from local exec modules? Wouldn't we need to special-case that somehow?

I considered it, yes. We still stage the build for the local exec module type but of course the build command is executed in the module dir. But mostly I'm just not sure where we'd go about detecting it. Would this belong in `resolveModuleConfigs`? In any case I'd rather special case this at the code level than having ambiguous, user-facing behaviour.

garden-service/test/unit/data/test-project-exec/module-local/garden.yml
eysi09: Yeah I actually did that to begin with but somehow thought this was better because it doesn't need a clean up step.

edvald:
> So task outputs for other module types will presumably have a new line at the end, as well as all test outputs.

It's worth reviewing (unrelated to this PR), but not necessarily because we in most cases use `execa` which trims automatically.

> I considered it, yes. We still stage the build for the local exec module type but of course the build command is executed in the module dir. But mostly I'm just not sure where we'd go about detecting it. Would this belong in resolveModuleConfigs? In any case I'd rather special case this at the code level than having ambiguous, user-facing behaviour.

We would need to special-case this in our code, unless/until we make this `local` flag somehow a framework native concept. I think it's best if we set this special case where `buildPath` is set, since that path is referenced everywhere else, including in the build staging code.

That would be here: https://github.com/garden-io/garden/blob/fb02889d3ffff133cea0e16d07bde48e0e6cfa1d/garden-service/src/build-dir.ts#L92

I'd change that method to accept a full module config, and do the special case if statement there. And make sure to comment and explain.

garden-service/test/unit/data/test-project-exec/module-local/garden.yml
edvald: Yeah I guess that also works, as long as we validate that the `pwd` output is correct.

eysi09: Yeah we do that in the test. That's basically how I verify the `it("should run the build command in the module dir if local true")` statement:

```javascript
expect(res.buildLog).to.eql(join(projectRoot, "module-local"))
```

eysi09:
@edvald I've added a PR to address the comment above, namely that local exec modules should not have their sources synced to the build dir.

examples/local-exec/README.md
```diff
@@ -8,18 +8,24 @@ The idea is to use a local exec module to run pre-build commands for a container
...
-The `backend` and `builder` modules are in the same `garden.yml` file in the `backend` directory.
+The project consists a `builder` module and a `backend` module. Both modules are in the same `garden.yml` file in the `backend` directory.
 
 The `backend` module is a simple `container` module that acts as a web server written in Go. The corresponding Dockerfile expects the web server binary to already be built before adding it to the image (note that this is only for demonstration purposes and not a recommended pattern for Go binaries).
 
 To achieve this, we add a `go build` command to the `builder` module, set `local: true`, and then declare it as a build dependency in the `backend` module.
 
 This ensures that Garden runs `go build` in the module directory before it attempts to build the Docker image for the `backend` module.
 
+Note also that since we git ignore the built binary, we need to tell Garden to **not** use the `.gitignore` file, which it does by default. We do this by setting the `dotIgnoreFiles` directive to `[]` in the project level configuration:
```
edvald: It's slightly odd to say "git ignore" imo... Maybe just go for more verbose, say that we exclude it from source control using the `.gitignore` file.

Also, I'd suggest setting `dotIgnoreFiles: [.gardenignore]` and maybe linking to the relevant docs section for further explanation.

edvald: Actually, maybe the `.gardenignore` business is needless complication for the example, but the link to the ignore docs would still be good.

garden-service/src/build-dir.ts
```diff
@@ -19,10 +19,15 @@ import {
...
 import { ConfigurationError } from "./exceptions"
 import { FileCopySpec, Module, getModuleKey } from "./types/module"
-import { zip } from "lodash"
 import execa from "execa"
 import { normalizeLocalRsyncPath } from "./util/fs"
 import { LogEntry } from "./logger/log-entry"
+import { ModuleConfig } from "./config/module"
+import { ConfigGraph } from "./config-graph"
+
+function isLocalExecModule(moduleConfig: ModuleConfig) {
```
edvald: Maybe drop in a FIXME comment here, since we don't want to keep special-casing this going forward.

examples/local-exec/README.md
```diff
@@ -8,18 +8,24 @@ The idea is to use a local exec module to run pre-build commands for a container
 
 ## Project Structure
 
-The project consists of a `local-executor` module, a `builder` module, and a `backend` module.
-
-The `local-executor` module has a task called `local-task` that just prints the current directory. You'll notice that it's the directory that the module is in, not the Garden build directory.
-
-The `backend` and `builder` modules are in the same `garden.yml` file in the `backend` directory.
+The project consists a `builder` module and a `backend` module. Both modules are in the same `garden.yml` file in the `backend` directory.
```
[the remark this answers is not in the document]
eysi09: This is from an old version.

examples/local-exec/README.md
eysi09: This has also been removed in the final version.

PR #1267, opened 2019-10-15
Terraform fixes

edvald:

One small bug fix, and a small improvement to the `terraform-gke` example.

Look at the individual commits, pretty straightforward.


garden-service/src/plugins/terraform/common.ts
```diff
@@ -177,7 +177,7 @@ export async function applyStack(log: LogEntry, provider: TerraformProvider, roo
  * appropriate arguments to pass to the Terraform CLI, otherwise an empty array.
  */
 export async function prepareVariables(targetDir: string, variables: object): Promise<string[]> {
-  if (Object.entries(variables).length === 0) {
+  if (Object.entries(variables || {}).length === 0) {
```
eysi09: Shouldn't the `variables` param be optional then?

edvald: Yes, or otherwise ensured to be present before calling this function.

Issue #1269, opened 2019-10-15
File watching optimization


matuszeman:
Garden seems to watch all files under project folder (ref: [garden.ts#L287](https://github.com/garden-io/garden/blob/master/garden-service/src/garden.ts#L287)) which are then filtered out depending on e.g. module settings (ref: [watch.ts#L172](https://github.com/garden-io/garden/blob/master/garden-service/src/watch.ts#L172)). This causes an issue with larger projects and the process runs out of memory after a while.

Could we discuss best way how this could be optimized please?

matuszeman:
Here are some options which came to my mind:

- Add new configuration options `watchPaths: string[]` on project/module level.
- Use existing [project module include/exclude](https://docs.garden.io/reference/config#modules) and [module include/exclude](https://docs.garden.io/reference/config#include) configuration.

edvald:
Thanks a lot for filing the issue @matuszeman. I've done a little bit of exploration on the matter.

We _can_ technically introduce an optimization when `modules.include` is specified in the project config, and includes only specific directories (i.e. `some/directory/**/*`). But unfortunately any more open-ended pattern in that list (e.g. `*.js` or anything else without a specific directory path) would require us to look at the whole project and post-filter, like we currently do. So users would need to be quite specific for us to be able to avoid watching the whole project root, and I'm not sure that's reasonable in terms of user experience.

If you specify exclude filters, there's still really no way for us to avoid watching the whole project root, because the list of directories to watch can always change, so we couldn't be aware of new directories if they're not already under the watched directories when we start the Garden process.

So I think your first suggested option is the only viable one to solve your issue (unless I'm missing a better idea, which I won't rule out :P). By default I resist adding config options, but I don't really see a way around it here, since the include/exclude filters can't sufficiently inform us what to watch except in very specific cases.

@matuszeman would you like to have a go at implementing a `watchPaths` option, or would you like to wait until we find time to add it? We'd love to help you contribute if you're up for it :)

matuszeman:
@edvald I played with it yesterday a bit, it's not tested at all, I don't even know if this approach helps the problem. I'd need to run it on the project I had a issue with before. Have a look.

I wanted to stay consistent with other `include` / `exclude` configuration options garden already has and I think I make sense to leverage `chodikar` `ignored` option for `exclude` here.

PR #1270, opened 2019-10-17
Implement `project.watch` configuration options

matuszeman:

TODO

Fixes garden-io/garden#1269

TODO

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -213,6 +213,28 @@ modules:
     - tmp/**/*
 ```
 
+### `watch`
+
+| Type     | Required |
+| -------- | -------- |
+| `object` | No       |
+
+### `watch.include[]`
+
+[watch](#watch) > include
+
+| Type            | Required |
+| --------------- | -------- |
+| `array[string]` | No       |
+
+### `watch.exclude[]`
+
+[watch](#watch) > exclude
+
+| Type            | Required |
+| --------------- | -------- |
+| `array[string]` | No       |
+
 ### `providers`
 
 A list of providers that should be used for this project, and their configuration. Please refer to individual plugins/providers for details on how to configure them.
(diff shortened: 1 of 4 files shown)


eysi09:
Hi @matuszeman!

Let me start by apologising for the late response. And follow up with a big thank you for the pull request!

The person you'd been in contact with is away on vacation so this one somehow fell in between the cracks. I'm really sorry for that. 

I'm away myself until after the weekend but we'll get to this ASAP early next week.

edvald:
Sorry I just got to this, back from vacation now. I think this is a good feature to add, but there's a bit of nuance to consider in the implementation. 

A couple of points I see at first glance:
1) It might actually make sense to just use the `modules.include`/`modules.exclude` fields. We're only really watching for changes to modules, and the separate field is likely superfluous.
2) We can only in some cases apply the `modules.include` field as an optimization, which is when there are only specific directories listed under that field, or directories followed by globs. That's feasible to implement, but would need to be documented.
3) The `exclude` part should work mostly as-is, just needs a bit more work to handle configuration changes. We essentially will need to fully reset the watcher when we reconfigure. Shouldn't be hard though, and likely better to do that either way.
4) We would need to add some tests to make sure this all works as expected, but of course this is just a draft PR :)

@matuszeman would you like to carry on with this, or should I have a go?

matuszeman:
You came and I'm leaving until Sunday :) so to move things forward I'd appreciate any help from your side if you find spare second.

I've been also thinking about implementing `watch` option on module level instead where it might make more sense maybe, and will be closer to the codebase of particular module, thus easier to manage.
`modules` on project level would be used to discover module garden files only and watch for those and module configuration would dictate what to watch on module level only. One module could be with NodeJS codebase where I'd definitely exclude `node_modules` or PHP with its `vendor` folder for example.

edvald:
Closing in favor of #1320, but thanks so much @matuszeman for the contribution!

PR #1271, opened 2019-10-17
go module type preliminary/prototype

l-korbes:
This is very preliminary work. *Do not merge.*

I'll fill out this PR tomorrow.

@edvald I'm currently studying `garden-service/src/types/plugin/service/deployService.ts` and `garden-service/src/plugins/kubernetes/container/deployment.ts` to try and figure out how to do the hot reload part. Any direction there is welcome. (Thanks!)

Fixes #

### DIFF
--- a/garden-service/src/plugins/go-container/go-container.ts
+++ b/garden-service/src/plugins/go-container/go-container.ts
@@ -1,0 +1,269 @@
+/*
+ * Copyright (C) 2018 Garden Technologies, Inc. <info@garden.io>
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+// import { omit, get } from "lodash"
+import { copy, pathExists, readFile, writeFile, ensureDir } from "fs-extra"
+import { createGardenPlugin } from "../../types/plugin/plugin"
+import {
+  ContainerModuleSpec,
+  ContainerServiceSpec,
+  ContainerTestSpec,
+  ContainerModuleConfig,
+  ContainerTaskSpec,
+  ContainerBuildSpec,
+} from "../container/config"
+import { joiArray, joiProviderName, joi } from "../../config/common"
+import { Module } from "../../types/module"
+import { resolve } from "path"
+// import { RuntimeError, ConfigurationError } from "../../exceptions"
+import { RuntimeError } from "../../exceptions"
+// import { containerHelpers } from "../container/helpers"
+import { STATIC_DIR } from "../../constants"
+import { containerModuleSpecSchema } from "../container/config"
+import { providerConfigBaseSchema } from "../../config/provider"
+import { go } from "./go"
+import { dedent } from "../../util/string"
+import { ModuleConfig, baseBuildSpecSchema } from "../../config/module"
+import { ConfigureModuleParams } from "../../types/plugin/module/configure"
+import { GetBuildStatusParams } from "../../types/plugin/module/getBuildStatus"
+import { BuildModuleParams } from "../../types/plugin/module/build"
+// import { binary } from "@hapi/joi"
+
+const defaultDockerfileName = "go-container.Dockerfile"
+const defaultDockerfilePath = resolve(STATIC_DIR, "go-container", defaultDockerfileName)
+
+interface GoContainerBuildSpec extends ContainerBuildSpec {
+  debug: boolean
+  compress: boolean | number
+  run: string
+  buildFlags: string
+  //other fields here
+}
+
+export interface GoContainerModuleSpec extends ContainerModuleSpec {
+  build: GoContainerBuildSpec
+}
+
+export type GoContainerModuleConfig = ModuleConfig<GoContainerModuleSpec>
+
+export interface GoContainerModule<
+  M extends GoContainerModuleSpec = GoContainerModuleSpec,
+  S extends ContainerServiceSpec = ContainerServiceSpec,
+  T extends ContainerTestSpec = ContainerTestSpec,
+  W extends ContainerTaskSpec = ContainerTaskSpec
+  > extends Module<M, S, T, W> { }
+
+const goKeys = {
+  debug: joi.boolean()
+    .description(dedent`
+      Go binaries contain DWARFv4 debugging information that can be used to, for example, attach a debugger to a live process.
+      By default, Garden passes the '-w' flag to the linker to omit this information and make the resulting binaries smaller.
+      If you do want this information, set this option to \`true\`.
+    `)
+    .example("false"),
+  // compress: joi.alternatives().try(
+  //     joi.boolean(),
+  //     joi.number(),
+  //   ).description(dedent`
+  //   "By default, Garden will compress the resulting Go binary with UPX \(upx.github.io\).
+  //   This can be disabled with \`false\`, or adjusted with
+  //   an integer in the 1–9 range, where 1 is fastest and 9 is smaller.
+  //   For this functionality to work with macOS, UPX needs
+  //   to first be installed with Homebrew using \`brew install upx\`.
+  //   `)
+  //   .example("1"),
+  buildFlags: joi.string()
+    .description("Use this to pass custom build flags to the Go compiler."),
+  // run: joiArray(joi.string())
+  //   .description("Use this to add arbitrary an arbitrary RUN command to the default Dockerfile."),
+}
+
+const goContainerModuleSpecSchema = containerModuleSpecSchema.keys({
+  build: baseBuildSpecSchema.keys(goKeys),
+})
+export const goContainerConfigSchema = providerConfigBaseSchema
+  .keys({
+    name: joiProviderName("go-container"),
+  })
+
+export const gardenPlugin = createGardenPlugin({
+  name: "go-container",
+  dependencies: ["container"],
+
+  // TODO
+  createModuleTypes: [{
+    name: "go-container",
+    base: "container",
+    docs: dedent`
+      A specialized version of the [container](https://docs.garden.io/reference/module-types/container) module type
+      that has special semantics for JAR files built with Maven.
+
+      Rather than build the JAR inside the container (or in a multi-stage build) this plugin runs \`mvn package\`
+      ahead of building the container, which tends to be much more performant, especially when building locally
+      with a warm artifact cache.
+
+      A default Dockerfile is also provided for convenience, but you may override it by including one in the module
+      directory.
+
+      To use it, make sure to add the \`maven-container\` provider to your project configuration.
+      The provider will automatically fetch and cache Maven and the appropriate OpenJDK version ahead of building.
+    `,
+    schema: goContainerModuleSpecSchema,
+    handlers: {
+      configure: configureGoContainerModule,
+      getBuildStatus,
+      build,
+    },
+  }],
+})
+
+export async function configureGoContainerModule(params: ConfigureModuleParams<GoContainerModule>) {
+  const { base, moduleConfig } = params
+
+  let containerConfig: ContainerModuleConfig = { ...moduleConfig, type: "container" }
+
+  const configured = await base!({ ...params, moduleConfig: containerConfig })
+
+  return {
+    moduleConfig: {
+      ...configured.moduleConfig,
+      type: "go-container",
+      // exclude: ["vendor"],
+      spec: {
+        ...configured.moduleConfig.spec,
+        build: moduleConfig.spec.build,
+        dockerfile: defaultDockerfileName,
+      },
+    },
+  }
+}
+
+async function getBuildStatus(params: GetBuildStatusParams<GoContainerModule>) {
+  // const { base, module, log } = params
+  const { base } = params
+
+  // await prepareBuild(module, log)
+
+  return base!(params)
+}
+
+async function build(params: BuildModuleParams<GoContainerModule>) {
+  const { base, module, log } = params
+  // let { debug, compress, buildFlags, run } = module.spec
+  let { debug } = module.spec.build
+
+  log.setState(`Starting Go build process...`)
+
+  // Check if any imports have changed
+  console.log(`Checking for new dependencies...`)
+  let tidy
+  try {
+  tidy = await go.exec({
+    args: ["mod", "tidy"],
+    cwd: module.path,
+    log,
+  })
+} catch (err) {
+  console.log(err.stderr)
+}
+
+  // If they did, update the vendor directory
+  console.log(`Downloading new dependencies...`)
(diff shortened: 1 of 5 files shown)


eysi09:
@ellenkorbes FYI: You can flag PRs as drafts if you don't intend to merge them as is.

garden-service/src/plugins/go-container/go.ts
```diff
@@ -0,0 +1,46 @@
...
+    targetPath: ["go", "bin", "go"],
+  },
+}
+
+export const go = new BinaryCmd({
+  name: "go",
+  specs: {
+    darwin: macos,
+    linux: penguin,
+    // Is the "32" here really 32 or is it just the name?
```
edvald: It's just a relic, standard Node.js platform names that we match.

garden-service/src/plugins/go-container/go-container.ts
```diff
@@ -0,0 +1,269 @@
...
+
+  if (!(await pathExists(binaryPath))) {
+    throw new RuntimeError(
+      `Could not find artifact at: ${binaryPath}'`,
+      { binaryPath },
+    )
+  }
+
+  // What was this for?
+  // await copy(resolvedJarPath, resolve(module.buildPath, "app.jar"))
```
edvald: This was needed in the `maven-container` module type, because the Jar file there is generated inside the user's source folder.

garden-service/src/plugins/go-container/go-container.ts
```diff
@@ -0,0 +1,269 @@
...
+    cwd: module.path,
+    log,
+  })
+} catch (err) {
+  console.log(err.stderr)
+}
+
+  // If they did, update the vendor directory
+  console.log(`Downloading new dependencies...`)
+  if (tidy.stdout != "") {
```
edvald: We should always use `!==`, but you can also just use `if (tidy.stdout)` here.

garden-service/src/plugins/go-container/go-container.ts
```diff
@@ -0,0 +1,269 @@
...
+      args: ["mod", "vendor"],
+      cwd: module.path,
+      log,
+    })
+  }
+
+    try {
+    await ensureDir(resolve(module.path, "bin"))
+  } catch (err) {
+    console.error(err)
```
edvald: I don't believe `ensureDir()` should ever error, no need for the try/catch.

garden-service/src/plugins/go-container/go-container.ts
```diff
@@ -0,0 +1,269 @@
...
+  debug: joi.boolean()
+    .description(dedent`
+      Go binaries contain DWARFv4 debugging information that can be used to, for example, attach a debugger to a live process.
+      By default, Garden passes the '-w' flag to the linker to omit this information and make the resulting binaries smaller.
+      If you do want this information, set this option to \`true\`.
+    `)
+    .example("false"),
+  // compress: joi.alternatives().try(
+  //     joi.boolean(),
+  //     joi.number(),
```
edvald: You can further specify the allowed numeric values: `joi.number().integer().min(1).max(9)`

garden-service/src/plugins/go-container/go-container.ts
```diff
@@ -0,0 +1,269 @@
...
+  //   ).description(dedent`
+  //   "By default, Garden will compress the resulting Go binary with UPX \(upx.github.io\).
```
edvald: Extra quote-mark at the start. And the escaping is not needed on the parentheses.

garden-service/src/plugins/go-container/go-container.ts
```diff
@@ -0,0 +1,269 @@
...
+  //   This can be disabled with \`false\`, or adjusted with
+  //   an integer in the 1–9 range, where 1 is fastest and 9 is smaller.
+  //   For this functionality to work with macOS, UPX needs
+  //   to first be installed with Homebrew using \`brew install upx\`.
+  //   `)
+  //   .example("1"),
```
edvald: The example should be an actual value, e.g. `1` and not `"1"`, since the example value is actually validated when the schema is defined. You can also set the default here (`.default(9)`).

garden-service/src/plugins/go-container/go-container.ts
```diff
@@ -0,0 +1,269 @@
...
+  let tidy
+  try {
+  tidy = await go.exec({
+    args: ["mod", "tidy"],
+    cwd: module.path,
+    log,
+  })
+} catch (err) {
+  console.log(err.stderr)
+}
```
edvald: I see these spacing issues here and there. I highly recommend installing the TSLint plugin in VSCode, which will automatically correct these.

garden-service/src/plugins/go-container/go-container.ts
```diff
@@ -0,0 +1,269 @@
...
+
+async function build(params: BuildModuleParams<GoContainerModule>) {
+  const { base, module, log } = params
+  // let { debug, compress, buildFlags, run } = module.spec
+  let { debug } = module.spec.build
+
+  log.setState(`Starting Go build process...`)
+
+  // Check if any imports have changed
+  console.log(`Checking for new dependencies...`)
```
edvald: I recommend adding a status subline for these logs:

```typescript
const status = log.info("Checking for new dependencies...")
// And then for the next status message below
...
status.setState("Downloading new dependencies...")
```

garden-service/src/plugins/go-container/go-container.ts
```diff
@@ -0,0 +1,269 @@
...
+  > extends Module<M, S, T, W> { }
+
+const goKeys = {
+  debug: joi.boolean()
+    .description(dedent`
+      Go binaries contain DWARFv4 debugging information that can be used to, for example, attach a debugger to a live process.
+      By default, Garden passes the '-w' flag to the linker to omit this information and make the resulting binaries smaller.
+      If you do want this information, set this option to \`true\`.
+    `)
+    .example("false"),
```
edvald: Change to `.example(true)` (no quotes)
And add `.default(false)`

garden-service/static/go-container/go-container.Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM alpine
+RUN apk add entr ca-certificates
+RUN mkdir app
+WORKDIR /app
+EXPOSE 8080
```
edvald: The `EXPOSE` directive is not necessary, easier to just omit it.

garden-service/static/go-container/go-container.Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM alpine
```
edvald: We should definitely specify an Alpine version (just go with whatever is most recent).

garden-service/static/go-container/go-container.Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+RUN apk add entr ca-certificates
+RUN mkdir app
+WORKDIR /app
+EXPOSE 8080
+
+COPY bin/ /app/
+
+CMD ls binary | entr -r ./binary
```
edvald: I realize this is more of a placeholder, but we should use `ENTRYPOINT` instead of `CMD` here and make it simply `ENTRYPOINT [/app/bin]`. When we use hot reloading, we'll anyway override the command in the manifest.

garden-service/src/plugins/go-container/go-container.ts
```diff
@@ -0,0 +1,269 @@
...
+
+      Rather than build the JAR inside the container (or in a multi-stage build) this plugin runs \`mvn package\`
+      ahead of building the container, which tends to be much more performant, especially when building locally
+      with a warm artifact cache.
+
+      A default Dockerfile is also provided for convenience, but you may override it by including one in the module
+      directory.
+
+      To use it, make sure to add the \`maven-container\` provider to your project configuration.
+      The provider will automatically fetch and cache Maven and the appropriate OpenJDK version ahead of building.
```
edvald: Let's remember to update this text, and also to add the module type to our reference docs.

garden-service/src/plugins/go-container/go-container.ts
```diff
@@ -0,0 +1,269 @@
...
+  console.log(`Building...`)
+  try {
+  await go.exec({
+    args: buildArgs,
+    cwd: module.buildPath,
+    log,
+    env: {
+      CGO_ENABLED: "0",
+      GOOS: "linux",
+      GOCACHE: resolve(module.buildPath, "cache"),
```
edvald: Why do we override the cache?

garden-service/src/plugins/go-container/go-container.ts
```diff
@@ -0,0 +1,269 @@
...
+// import { binary } from "@hapi/joi"
+
+const defaultDockerfileName = "go-container.Dockerfile"
+const defaultDockerfilePath = resolve(STATIC_DIR, "go-container", defaultDockerfileName)
+
+interface GoContainerBuildSpec extends ContainerBuildSpec {
+  debug: boolean
+  compress: boolean | number
+  run: string
+  buildFlags: string
```
edvald: Both `run` and `buildFlags` should be `string[]`

garden-service/src/plugins/go-container/go-container.ts
```diff
@@ -0,0 +1,269 @@
...
+}
+
+async function getBuildStatus(params: GetBuildStatusParams<GoContainerModule>) {
+  // const { base, module, log } = params
+  const { base } = params
+
+  // await prepareBuild(module, log)
+
+  return base!(params)
+}
```
edvald: I think we can remove this handler.

garden-service/src/plugins/go-container/go-container.ts
edvald: Actually, I think we should omit this option entirely. It adds unnecessary complexity. Compression is anyway applied in container images, and when syncing during hot-reloading.

garden-service/src/plugins/go-container/go-container.ts
```diff
@@ -0,0 +1,269 @@
...
+
+// if (run != "") {
+//   let dockerFile = await readFile(dockerFileBuild, "utf8")
+//   dockerFile = dockerFile.replace(/echo/g, run)
+//   try {
+//   await writeFile(dockerFileBuild, dockerFile, "utf8")
+//   } catch (err) {
+//     console.log(err)
+//   }
+// }
```
edvald: This is quite messy. Better to use the `buildArgs` field on the container module type. See the `maven-container` module type configure handler for how to do it: https://github.com/garden-io/garden/blob/fca6a72aec570ac548b8f0fd6f67f8e4211807ab/garden-service/src/plugins/maven-container/maven-container.ts#L129

You'd do something like

```typescript
  containerConfig.spec.buildArgs = {
    EXTRA_RUN: moduleConfig.spec.build.run.join(" ") || "echo"
  }
```

And then add `RUN ${EXTRA_RUN}` at the bottom of the Dockerfile.

edvald:
Re: Adding hot-reloading, it should actually be reasonably simple. First override the `hotReloadCommand` for each service in the configure handler, and auto-populate the `hotReload.sync` field. Something like (I've not tested the code btw, just illustrating the idea):

```typescript
containerConfig.spec.hotReload = containerConfig.spec.hotReload || {
  sync: [{ source: "./bin", target: "/app/bin" }],
}

for (const service of containerConfig.spec.services || []) {
  service.hotReloadCommand = service.hotReloadCommand || [<the entr base command>]
}
```

Then you need to ensure the binary is built before the hot reloading happens:

```typescript
async function hotReload(params: HotReloadServiceParams<GoContainerModule>) {
  const { base, module, log } = params

  // Need to extract the binary build from the build handler into a separate function
  await buildBinary(module, log)

  // Defer the rest of it to the kubernetes provider (or whatever other provider applies, technically)
  return base!(params)
}
```

The kubernetes provider will then take care of the rest.

garden-service/src/plugins/go-container/go-container.ts
ellenkorbes: I'll leave it there commented out until the whole thing is fully functional and I can benchmark it. I suspect it could make a significant difference for large binaries.

edvald: Might be worth trying as a stretch goal, but I'd be quite surprised if it made a meaningful difference, since data is anyway compressed when going over the wire.

garden-service/src/plugins/go-container/go-container.ts
ellenkorbes: Go's default cache location is `~/.cache/`. Since we're keeping the Go install compartmentalized inside the module, I figured we should do the same for the compiler cache.

Which raises an interesting question. If a user has 10 Go projects, they'll have 10 copies of Go in their machine. Should we keep this in a central location e.g. `~/.garden/go`?

edvald: We download Go once for each _user_. So that won't be an issue. 

And I don't really see a reason to partition the cache by module. If different modules vendor the same libraries it's just less efficient to cache separately. Unless there's some concern specific to Go that I'm not aware of?

ellenkorbes: That refers to the compiler cache, for `.a` files and such. 

Libraries are vendored separately by default. We could come up with a hack to have a central location for all of them, but that'd cause version conflicts e.g. when one project uses version `v1` of a library and the other one uses `v2`.

edvald: I'm not suggesting doing anything unusual, just to leave the default GOCACHE setting. The default is "central" anyway, in that there's a shared cache directory. Unless there are known issues with the default (which would be pretty odd) I still don't see a reason to override it.

garden-service/src/plugins/go-container/go-container.ts
```diff
@@ -0,0 +1,300 @@
...
+    // .description("Use this to pass custom build flags to the Go compiler."),
+  // run: joi.array().items(joi.string())
+  // run: joi.string()
+    // .description("Use this to add arbitrary an arbitrary RUN command to the default Dockerfile."),
+  //   .default([" "]), //.example([["apk", "add", "curl"]]),
+
+}
+
+const goContainerModuleSpecSchema = containerModuleSpecSchema.keys({
+  build: baseBuildSpecSchema.keys(goKeys),
```
eysi09: The module spec should be top level, not under the build key. So basically:  `const goContainerModuleSpecSchema = containerModuleSpecSchema.keys(goKeys)`. Fixes the validation error we were seeing.

edvald: This was intentional, since all those parameters have to do with the build. It's admittedly clumsy to handle the schema this way, but we'll deal with that later at the framework level.

edvald:
You can now rebase this branch off master, since the branch it depended on has been merged. Will also make this easier to review and iterate on.

eysi09:
Closing in favour of #1511.

Issue #1283, opened 2019-10-28
Kubernetes context errors aren't presented to the end user


Eronarn:
### Current Behavior

If someone tries to use a Garden environment with a Kubernetes context that they don't have access to, the error message is indistinguishable from a connectivity problem to the Kubernetes cluster (which we frequently see if someone isn't on our VPN):

```
Failed resolving provider kubernetes. Here is the output:
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Unable to connect to Kubernetes cluster. Please make sure it is running, reachable and that you have the right context configured.
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

Failed resolving one or more provider configurations:
- kubernetes: Unable to connect to Kubernetes cluster. Please make sure it is running, reachable and that you have the right context configured.

See error.log for detailed error message
```

The actual error message is only in the error.log:

```
<snip>
  ingressHttpPort: 80
  ingressHttpsPort: 443
  setupIngressController: false
  path: /home/XXX/quantopian/qexec
  deploymentRegistry:
    hostname: '127.0.0.1:5000'
    namespace: staging
message: "Command failed with exit code 1 (EPERM): /home/XXX/.garden/tools/kubectl/804c71b3e6ad5976/kubectl --context=stg-admin@staging versionClient Version: version.Info{Major:\"1\", Minor:\"16\", GitVersion:\"v1.16.0\", GitCommit:\"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77\", GitTreeState:\"clean\", BuildDate:\"2019-09-18T14:36:53Z\", GoVersion:\"go1.12.9\", Compiler:\"gc\", Platform:\"linux/amd64\"}could not get token: AccessDenied: User: arn:aws:iam::XXX:user/XXX is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::XXX:role/XXX\n\tstatus code: 403, request id: XXX\nUnable to connect to the server: getting credentials: exec: exit status 1"
```

### Expected behavior

The IAM error message gets surfaced to the end user without having to read through error.log

### Suggested solution(s)

[...]

PR #1284, opened 2019-10-28
Standardise errors

### LINKED ISSUE #1282: Better error messages when tasks fails
We print the `error.message` part when tasks fails. This is not always very informative.

Case in point, if we have an exec module that runs the command `[mkdir, foo]` where the dir `foo` already exists, we get the following error:

```
Failed building my-module. Here is the output:
Command failed with exit code 1 (EPERM): mkdir foo

1 build task(s) failed!
```

...which doesn't tell us a lot.

 If we however print `error.all` (assuming the error originated in an `execa` call, `all` is basically `stdout + stderr`), we get:

```
Failed building my-module. Here is the output:
mkdir: foo: File exists

1 build task(s) failed!
```

...which is way more helpful.

The underlying issue is that we use different libraries to run these kind of commands and they return error objects of different shapes. We should aim to standardise these and include the `stderr` part when applicable.

eysi09:

Previously, the error message was not visible when commands were executed via `execa` (see #1282 for an example). Furthermore `execa` and `spawn` would return different looking errors.

The `execa` error message looked like this:

Screenshot 2019-10-28 at 15 04 45

Now both `execa` and `spawn` errors look like this:

Screenshot 2019-10-28 at 15 06 01

In particular, the error itself is visible and the command and error are bold. (EDIT: Text is no longer bold, see comments below).

Fixes #1282, #1283 

I updated the commit to address an issue that just came up (#1283). I explain the changes in the comments below so that we can compare the different screenshots.

I suggest looking at the final changes as opposed to commit by commit. The tests should also paint a good picture.

Also, I'm flagging this as a draft PR until we squash and merge #1262 as it depends on it.

### DIFF
--- a/garden-service/src/plugins/kubernetes/namespace.ts
+++ b/garden-service/src/plugins/kubernetes/namespace.ts
@@ -18,6 +18,7 @@ import { GetEnvironmentStatusParams } from "../../types/plugin/provider/getEnvir
 import { kubectl, KUBECTL_DEFAULT_TIMEOUT } from "./kubectl"
 import { LogEntry } from "../../logger/log-entry"
 import { gardenAnnotationKey } from "../../util/string"
+import dedent from "dedent"
 
 const GARDEN_VERSION = getPackageVersion()
 type CreateNamespaceStatus = "pending" | "created"
@@ -116,21 +117,14 @@ export async function prepareNamespaces({ ctx, log }: GetEnvironmentStatusParams
     // TODO: use API instead of kubectl (I just couldn't find which API call to make)
     await kubectl.exec({ log, provider: k8sCtx.provider, args: ["version"] })
   } catch (err) {
-    let message = err.message
-    if (err.stdout) {
-      message += err.stdout
-    }
-    if (err.stderr) {
-      message += err.stderr
-    }
-    throw new DeploymentError(
-      `Unable to connect to Kubernetes cluster. ` +
-      `Please make sure it is running, reachable and that you have the right context configured.`,
-      {
-        providerConfig: k8sCtx.provider.config,
-        message,
-      },
-    )
+    log.setError("Error")
+
+    throw new DeploymentError(dedent`
+      Unable to connect to Kubernetes cluster. Got error:
+
+      ${err.message}
+    `,
+      { providerConfig: k8sCtx.provider.config })
   }
 
   return Bluebird.props({
--- a/garden-service/test/unit/src/util/util.ts
+++ b/garden-service/test/unit/src/util/util.ts
@@ -1,15 +1,179 @@
 import { expect } from "chai"
+import { describe } from "mocha"
 import {
   pickKeys,
   getEnvVarName,
   deepOmitUndefined,
   deepFilter,
   splitLast,
+  exec,
+  createOutputStream,
+  makeErrorMsg,
+  renderOutputStream,
+  spawn,
 } from "../../../../src/util/util"
 import { expectError } from "../../../helpers"
 import { splitFirst } from "../../../../src/util/util"
+import { getLogger } from "../../../../src/logger/logger"
+import { dedent } from "../../../../src/util/string"
+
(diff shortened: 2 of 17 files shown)


eysi09:
Actually, it might not be a good idea to include ansi characters in the message itself. Although the bold looks pretty good imo. Here's what it looks now:

Screenshot 2019-10-28 at 21 52 48

eysi09:
Updated so that it only prints "Here are the last lines..." part if they are different from the error.

Also addressed the comments in #1283. Results in the following three changes:

1. "Unable to connect to cluster..." is set as the state of the active log line.
2. The "Here's the output:" part now contains the actual error (as opposed to the text from 1.)
3. The "Failed resolving one or more providers..." message now only lists the names of the providers but not the message itself since that's printed above. The error object still contains all the error messages.

So we go from this:

Screenshot 2019-10-28 at 22 37 09

...to this:

Screenshot 2019-10-28 at 22 36 16

garden-service/src/plugins/kubernetes/namespace.ts
```diff
@@ -116,21 +117,11 @@ export async function prepareNamespaces({ ctx, log }: GetEnvironmentStatusParams
...
-      {
-        providerConfig: k8sCtx.provider.config,
-        message,
-      },
-    )
+    log.setError(dedent`
+      Unable to connect to Kubernetes cluster. Please make sure it is running, reachable and that you have the right context configured.
+    `)
+
+    throw new DeploymentError(err.message, { providerConfig: k8sCtx.provider.config })
```
edvald: This seems slightly odd to me. I'd rather suggest prepending the "Unable to connect" string to `err.message`.

edvald: I'd go with 3) 👍

eysi09: Would we not want to set some error message on the active log line though?

garden-service/test/unit/src/util/util.ts
```diff
@@ -1,15 +1,159 @@
...
+          error: "ls: scottiepippen: No such file or directory",
+        }))
+      }
+    })
+  })
+
+  describe("spawn", () => {
+    before(function() {
+      // These tests depend on the underlying OS and are only executed on macOS
+      if (process.platform !== "darwin") {
```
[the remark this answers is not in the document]
eysi09: One way to find out.

garden-service/src/plugins/kubernetes/namespace.ts
edvald: We could simply set it to "Error" imo... Or something like that. "Failed" maybe.

garden-service/test/unit/src/util/util.ts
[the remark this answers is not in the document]
edvald: Didn't work directly, but you can modify the assert to match part of the string. That should be portable.

Another option would be a command that we control more explicitly, e.g. `sh -c "echo foo; exit 1"`. That way we know the exact string and exit code.

garden-service/src/plugins/kubernetes/namespace.ts
eysi09: I set the error in the log line to "Unable to connect to cluster" as well.

Screenshot 2019-11-01 at 08 34 58

This looks fine IMO but happy to change if you have any better suggestions.

eysi09:
@edvald  I've addressed the comments and the tests are passing.

garden-service/src/plugins/kubernetes/namespace.ts
edvald: That looks okay but my concern with the inline error message is that it's provider-specific. Multiple providers might fail, and we should expect users to have more and more providers over time.

eysi09: Yeah I was kind of thinking that other providers would then just overwrite the error. But you're right, I'll use something more generic.

PR #1294, opened 2019-10-30
fix(k8s): handle CronJob resources correctly

### LINKED ISSUE #1275: Unrecognized resource type batch/v1beta1/CronJob
### Current Behavior

Attempting to deploy a Helm chart for a CronJob resource fails with
`Unrecognized resource type batch/v1beta1/CronJob`

### Expected behavior

The resource deploys successfully, as it does with `helm upgrade --install`

### Reproducible example

See https://github.com/garoose/garden-cron-job-test for a small example of the problem

### Your environment

`garden version`
0.10.13

`kubectl version`
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-17T17:16:09Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.4", GitCommit:"67d2fcf276fcd9cf743ad4be9a9ef5828adc082f", GitTreeState:"clean", BuildDate:"2019-09-18T14:41:55Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}

`docker version`
Client:
 Version:           18.09.7
 API version:       1.38 (downgraded from 1.39)
 Go version:        go1.10.1
 Git commit:        2d0083d
 Built:             Fri Aug 16 14:20:06 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          18.06.1-ce
  API version:      1.38 (minimum version 1.12)
  Go version:       go1.10.4
  Git commit:       e68fc7a
  Built:            Tue May  7 17:57:34 2019
  OS/Arch:          linux/amd64
  Experimental:     false

edvald:

This would also have applied in cases where a group had different
resource types under different API versions. Which is not particularly
common, but most likely affected some other resource types.

Fixes #1275

### DIFF
--- a/garden-service/src/plugins/kubernetes/api.ts
+++ b/garden-service/src/plugins/kubernetes/api.ts
@@ -45,23 +45,18 @@ interface ApiGroupMap {
 }
 
 interface ApiResourceMap {
-  [kind: string]: V1APIResource
+  [group: string]: { [kind: string]: V1APIResource }
 }
 
 interface ApiInfo {
   coreApi: V1APIVersions
   groups: V1APIGroup[]
   groupMap: ApiGroupMap
-  resources: { [group: string]: ApiResourceMap }
-}
-
-interface ApiResourceInfo {
-  group: V1APIGroup
-  resource: V1APIResource
 }
 
 const cachedConfigs: { [context: string]: KubeConfig } = {}
 const cachedApiInfo: { [context: string]: ApiInfo } = {}
+const cachedApiResourceInfo: { [context: string]: ApiResourceMap } = {}
 const apiInfoLock = new AsyncLock()
 
 // NOTE: be warned, the API of the client library is very likely to change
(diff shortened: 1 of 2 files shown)


garden-service/src/plugins/kubernetes/status/status.ts
```diff
@@ -178,6 +178,8 @@ export async function waitForResources({ ctx, provider, serviceName, resources,
 
     statuses = await checkResourceStatuses(api, namespace, resources, log)
 
+    console.log(statuses)
```
eysi09: Remove console.log

edvald:
Good catch. I've modified the code to account for this, which reduces slightly the number of API calls we need on init. I kept the methods though, because we may have use for them later.

PR #1296, opened 2019-10-31
feat: allow specifying tolerations for registry-proxy

### LINKED ISSUE #1279: No way to add tolerations to `garden-registry-proxy`
### Current Behavior

If a node with taints is added (with, for example, GPU), the `garden-registry-proxy` daemonset never gets deployed there, since by default it does not have any tolerations

### Expected behavior

I would expect to be able to configure that in project-level `garden.yaml`.

### Workaround

```
kubectl -n garden-system edit daemonsets.apps garden-registry-proxy
```
and then add the needed tolerations to the definition.

eysi09:

Enables users to set tolerations for the registry-proxy DaemonSet. From the docs:

> For setting tolerations on the registry-proxy when using in-cluster building.
The registry-proxy is a DaemonSet that proxies connections to the docker registry service on each node.
>
> Use this only if you're doing in-cluster building and the nodes in your cluster
have [taints](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/).

Fixes #1279

The description for the individual toleration fields and the validation rules are derived from the [K8s API reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#toleration-v1-core).

### DIFF
--- a/garden-service/src/plugins/kubernetes/config.ts
+++ b/garden-service/src/plugins/kubernetes/config.ts
@@ -54,6 +54,14 @@ interface KubernetesStorage {
   sync: KubernetesStorageSpec
 }
 
+export interface Toleration {
+  effect?: "NoSchedule" | "PreferNoSchedule" | "NoExecute"
+  key?: string
+  operator: "Exists" | "Equal"
+  tolerationSeconds?: number
+  value?: string
+}
+
 export type ContainerBuildMode = "local-docker" | "cluster-docker" | "kaniko"
 
 export type DefaultDeploymentStrategy = "rolling"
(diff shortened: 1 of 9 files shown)


garden-service/src/plugins/kubernetes/config.ts
```diff
@@ -360,6 +369,46 @@ export const kubernetesConfigBase = providerConfigBaseSchema
...
+        .description(dedent`
+          Key is the taint key that the toleration applies to. Empty means match all taint keys.
+          If the key is empty, operator must be Exists; this combination means to match all values and all keys.
+        `),
+      operator: joi.string()
+        .allow("Exists", "Equal")
+        .default("Equal")
+        .description(dedent`
+          Operator represents a key's relationship to the value. Valid operators are Exists and Equal. Defaults to
+          Equal. Exists is equivalent to wildcard for value, so that a pod can tolerate all taints of a
```
edvald: Would be nice to have quotes around the value names.

PR #1297, opened 2019-10-31
chore(tests): create separate integ test suite

edvald:

Split some of our tests into a separate `integ` suite. We should use this for tests that require e.g. Docker, Kubernetes or other runtime dependencies to be running on the machine when executing.

These tests currently run in the `test-minikube` job in CI.

Also did some minor refactoring on the e2e tests, and updated the
contributor docs.


10ko:
Once the conflicts are fixed, I'll merge.
👍

edvald:
I figured 30 seconds would be plenty, haven't seen any tests take longer.

10ko:
I stand corrected, I must have had some bad run at some point. I checked some recent run and it looks the max is 17-18 seconds.
Cool, LGTM

PR #1298, opened 2019-10-31
fix(k8s): hostPath is now relative to module source path

edvald:

Previously we implicitly expected an absolute path. Now we explicitly
make it relative to the module path, but also support absolute paths.

Also added some docs, warning against the usage of hostPath.


garden-service/src/plugins/container/config.ts
```diff
@@ -285,10 +286,18 @@ const volumeSchema = joi.object()
...
       .required()
       .description("The path where the volume should be mounted in the container."),
     hostPath: joi.string()
-      .meta({ deprecated: true }),
+      .posixPath()
+      .description(dedent`
+        _NOTE: Usage of hostPath is generally discouraged, since it doesn't work reliably across different platforms
+        and providers. Some providers may not support it at all._
+
+        A local path or path on the node that's running the container, to mount in the container, relative to the
```
eysi09: Should we not note that you can also use absolute paths?

PR #1303, opened 2019-11-01
improvement(k8s): much faster init and status checks

edvald:

Initializing Garden, especially for remote clusters, tends to take way too much time. We had been deferring that problem until we start running a persistent shared instance of Garden per project, but that'll take a while longer, so I decided to look for some quick optimizations.

Specifically, I did the following:

1) Avoid detailed checks on resources in helm module status checks,
   unless hot reloading is enabled. This does have a trade-off, which
   is that status may drift, but users can re-run `garden init --force`
   or `garden cluster kubernetes cluster-init` to re-apply resources
   when needed.
2) Removed all usage of `kubectl diff`, which is super flimsy anyway.
3) Added our own "last-applied-configuration" annotation and use that
   for slightly more accurate resource comparison when applicable.


eysi09:
Haven't started the review but curious if you noticed much difference on your end.

edvald:
The "resolving providers" step went down from approx. 20 to 4 seconds working from my home wi-fi against our dev cluster, but YMMV (network latency, configuration and all kinds of things can change the absolute numbers). The difference is much less pronounced for local clusters, but still appreciable.

PR #1332, opened 2019-11-11
fix(sync): issue with build staging sync procedure on Windows

edvald:

We found rsync rename errors during testing on Windows. This is a workaround, that we only apply on Windows.


edvald:
Yeah good idea. I'll add that and then merge.

PR #1334, opened 2019-11-11
Artifacts dashboard

eysi09:

Displays stored artifacts in the dashboard.


eysi09:
This is how it looks with 10 artifacts:

Screenshot 2019-11-11 at 23 50 39

The artifacts are downloaded when you click them.

When you click the copy button, a banner appears:

Screenshot 2019-11-11 at 23 50 45

Would be nice to have two key value pairs in the same for the first few fields two save space. I'll see about fixing that but regardless, this should be ready for review.

PR #1339, opened 2019-11-12
feat: production flag

10ko:
Implements a `environments[].production` flag which alters the generated manifests for the container modules and put protections on a subset of garden commands.

Fixes #

WIP: Needs refactoring and more tests + the commands protection implementation.
Just putting it out here so we can review it fast.

### DIFF
--- a/docs/reference/commands.md
+++ b/docs/reference/commands.md
@@ -29,6 +29,7 @@ quiet: suppresses all log output, same as --silent.
   | `--log-level` | `-l` | `error` `warn` `info` `verbose` `debug` `silly` `0` `1` `2` `3` `4` `5`  | Set logger level. Values can be either string or numeric and are prioritized from 0 to 5 (highest to lowest) as follows: error: 0, warn: 1, info: 2, verbose: 3, debug: 4, silly: 5.
   | `--output` | `-o` | `json` `yaml`  | Output command result in specified format (note: disables progress logging and interactive functionality).
   | `--emoji` |  | boolean | Enable emoji in output (defaults to true if the environment supports it).
+  | `--yes` | `-y` | boolean | Automatically approve any yes/no prompts during execution.
 
 ### garden build
 
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -429,6 +429,22 @@ The name of the environment.
 | -------- | -------- |
 | `string` | Yes      |
 
+### `environments[].production`
+
+[environments](#environments) > production
+
+Set environment as production.
+
+Setting this flag to `true` will activate the protection on the `deploy`, `test`, `task`, `build`, `call`, `init` and `dev` commands.
+A protected command will ask for a user confirmation every time is run agains an environment marked as production.
+Run the command with the "--yes" flag to skip the check (e.g. when running Garden in CI).
+
+Note: This flag will affect how certain providers behave. For more details please check the documentation for the providers in use.
+
+| Type      | Required | Default |
+| --------- | -------- | ------- |
+| `boolean` | No       | `false` |
+
 
 ## Project YAML schema
 ```yaml
@@ -463,6 +479,7 @@ environments:
     varfile: garden.<env-name>.env
     variables: {}
     name:
+    production: false
 ```
 
 ## Module configuration keys
--- a/garden-service/src/cli/cli.ts
+++ b/garden-service/src/cli/cli.ts
@@ -142,6 +142,11 @@ export const GLOBAL_OPTIONS = {
     help: "Enable emoji in output (defaults to true if the environment supports it).",
     defaultValue: envSupportsEmoji(),
   }),
+  "yes": new BooleanParameter({
+    alias: "y",
+    help: "Automatically approve any yes/no prompts during execution.",
+    defaultValue: false,
+  }),
 }
 
 export type GlobalOptions = typeof GLOBAL_OPTIONS
@@ -304,7 +309,7 @@ export class GardenCli {
           await this.initFileWriters(logger, garden.projectRoot, garden.gardenDirPath)
 
           // Init Analytics, track command if user opted-in
-          const analytics = await new AnalyticsHandler(garden).init()
+          const analytics = await new AnalyticsHandler(garden, parsedOpts).init()
           analytics.trackCommand(command.getFullName())
 
           // tslint:disable-next-line: no-floating-promises
@@ -312,16 +317,23 @@ export class GardenCli {
 
           await checkForStaticDir()
 
-          // TODO: enforce that commands always output DeepPrimitiveMap
-          result = await command.action({
-            garden,
-            log,
-            footerLog,
-            headerLog,
-            args: parsedArgs,
-            opts: parsedOpts,
-          })
+          // Check if the command is protected and ask for confirmation to proceed if production flag is "true".
+          if (await command.isAllowedToRun(garden, log, parsedOpts)) {
+            // TODO: enforce that commands always output DeepPrimitiveMap
 
+            result = await command.action({
+              garden,
+              log,
+              footerLog,
+              headerLog,
+              args: parsedArgs,
+              opts: parsedOpts,
+            })
+          } else {
+            // The command is protected and the user decided to not continue with the exectution.
+            log.setState("\nCommand aborted.")
+            result = {}
+          }
           await garden.close()
         } catch (err) {
           // Generate a basic report in case Garden.factory(...) fails and command is "get debug-info".
--- a/garden-service/src/commands/base.ts
+++ b/garden-service/src/commands/base.ts
@@ -17,6 +17,9 @@ import { LogEntry } from "../logger/log-entry"
 import { printFooter } from "../logger/util"
 import { GlobalOptions } from "../cli/cli"
 import { joi } from "../config/common"
+import inquirer = require("inquirer")
+import dedent = require("dedent")
+import chalk from "chalk"
 
 export interface ParameterConstructor<T> {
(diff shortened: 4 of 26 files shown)


garden-service/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -212,6 +224,14 @@ interface CreateDeploymentParams {
   namespace: string
   enableHotReload: boolean
   log: LogEntry
+  production: boolean
+}
+
+const getNumberOfReplicas = (service, production) => {
+  if (production && service.spec.replicas < MINIMUM_PROD_REPLICAS) {
+    return MINIMUM_PROD_REPLICAS
+  }
```
edvald: This is probably not a good idea. Users may have legit reasons to want a single instance of a container. I'd rather just modify the default.

garden-service/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -348,6 +369,37 @@ export async function createWorkloadResource({
 
   deployment.spec.template.spec.containers = [container]
 
+  if (production) {
+    const affinity: V1Affinity = {
+      podAntiAffinity: {
+        preferredDuringSchedulingIgnoredDuringExecution: [
```
edvald: Can you explain the motivation here?

10ko: The idea is that if you are running multiple nodes you might want to have all your pods evenly distributed between your nodes, to avoid unavailability in case that node goes down.
I set that up with antiaffinity for pods matching those labels and based on the node hostname. This is a soft requirement thou (hence the "preferred...").

I am open to suggestion thou.

garden-service/src/plugins/kubernetes/container/deployment.ts
10ko: Makes sense, I'll revert that.

10ko: I was checking how is the default but it's defaulted to 1 replica in the service schema definition. Meaning we cannot know if the user has set the replica numbers to 1 or it's just the default. I guess I'll leave it configurable by the user.

edvald: You can also remove the schema default, and apply the default elsewhere.

docs/reference/commands.md
```diff
@@ -29,6 +29,7 @@ quiet: suppresses all log output, same as --silent.
   | `--log-level` | `-l` | `error` `warn` `info` `verbose` `debug` `silly` `0` `1` `2` `3` `4` `5`  | Set logger level. Values can be either string or numeric and are prioritized from 0 to 5 (highest to lowest) as follows: error: 0, warn: 1, info: 2, verbose: 3, debug: 4, silly: 5.
   | `--output` | `-o` | `json` `yaml`  | Output command result in specified format (note: disables progress logging and interactive functionality).
   | `--emoji` |  | boolean | Enable emoji in output (defaults to true if the environment supports it).
+  | `--yes` | `-y` | boolean | Accept to proceed running a protected command against a production environment.
```
edvald: The message here is quite specific for a global option. How about something like "Automatically approve any yes/no prompts during execution."?

docs/reference/config.md
```diff
@@ -429,6 +429,16 @@ The name of the environment.
 | -------- | -------- |
 | `string` | Yes      |
 
+### `environments[].production`
+
+[environments](#environments) > production
+
+Set environment as production
```
edvald: We should definitely explain this further, what sort of thing this may affect etc.

garden-service/src/cli/cli.ts
```diff
@@ -312,16 +317,23 @@ export class GardenCli {
...
+              garden,
+              log,
+              footerLog,
+              headerLog,
+              args: parsedArgs,
+              opts: parsedOpts,
+            })
+          } else {
+            // The command is protected and the user decided to not continue with the exectution.
+            log.setState("\nYou chose to not proceed. Exiting now...")
```
edvald: Phrasing is slightly odd. How about simply "Command aborted."? Or is that perhaps too dry? 😛

garden-service/src/commands/base.ts
```diff
@@ -327,6 +332,43 @@ export abstract class Command<T extends Parameters = {}, U extends Parameters =
...
+   *
+   * @param {Garden} garden
+   * @param {LogEntry} log
+   * @param {GlobalOptions} opts
+   * @returns {Promise<Boolean>}
+   * @memberof Command
+   */
+  async isAllowedToRun(garden: Garden, log: LogEntry, opts: GlobalOptions): Promise<Boolean> {
+    log.root.stop()
+    // debugger
```
edvald: Remove debugger comment

garden-service/src/commands/base.ts
```diff
@@ -327,6 +332,43 @@ export abstract class Command<T extends Parameters = {}, U extends Parameters =
...
+    if (!opts.yes && this.protected && garden.production) {
+      const defaultMessage = chalk.yellow(dedent`
+        Warning: you are trying to run "garden ${this.getFullName()}" against the production environment "${
```
edvald: "Warning: you" -> "Warning: You"
"the production environment <name>" -> "a production environment (<name>)"

garden-service/src/cli/cli.ts
10ko: "Command aborted.. Byebye?" :)

Joking. Sounds good. :)

garden-service/src/commands/base.ts
```diff
@@ -327,6 +332,43 @@ export abstract class Command<T extends Parameters = {}, U extends Parameters =
...
+        garden.environmentName
+      }"!
+          Are you sure you want to continue? (use "--yes" to skip this check).
```
edvald: `use "--yes" ...` -> `run the command with the "--yes" flag ...`

garden-service/src/commands/call.ts
```diff
@@ -31,6 +31,7 @@ type Args = typeof callArgs
 export class CallCommand extends Command<Args> {
   name = "call"
   help = "Call a service ingress endpoint."
+  protected = true
```
edvald: Do we want to protect this command? What scenario are we trying to prevent?

garden-service/src/commands/get/get-status.ts
```diff
@@ -42,6 +42,7 @@ export interface StatusCommandResult extends AllEnvironmentStatus {
 export class GetStatusCommand extends Command {
   name = "status"
   help = "Outputs the status of your environment."
+  protected = true
```
edvald: Why do we need to protect this one? Getting the prod env status seems like a legit thing to do.

garden-service/src/plugins/container/config.ts
```diff
@@ -378,8 +378,7 @@ const serviceSchema = baseServiceSpecSchema.keys({
   replicas: joi
     .number()
     .integer()
-    .min(1)
-    .default(1).description(deline`
+    .min(1).description(deline`
       The number of instances of the service to deploy.
```
edvald: Perhaps add a note in the description, like "Defaults to 3 for environments configured with `production: true`, otherwise 1."

garden-service/src/plugins/kubernetes/container/deployment.ts
edvald: That makes sense. Could you add that as a comment above?

garden-service/test/integ/src/plugins/kubernetes/helm/config.ts
```diff
@@ -10,7 +10,7 @@ import { apply } from "json-merge-patch"
 import { getHelmTestGarden } from "./common"
 import { defaultHelmTimeout } from "../../../../../../src/plugins/kubernetes/helm/config"
 
-describe("validateHelmModule", () => {
+describe.only("validateHelmModule", () => {
```
edvald: Left `.only`

garden-service/src/plugins/kubernetes/container/deployment.ts
edvald: Also, could you add a test case for this function with production=true?

garden-service/src/commands/call.ts
10ko: Call endpoint `/delete-mailing-list`?

docs/reference/commands.md
10ko: Ok but then I will also refactor the Analytics prompt to be aware of this flag.

garden-service/src/commands/get/get-status.ts
10ko: That was for testing and I forgot ti there!

garden-service/src/commands/call.ts
edvald: I'd assume that would be blocked from the outside? 🤷‍♂ Or require some form of auth at least. `garden call` is almost stupidly simple, just fires a GET request with no parameters :P

garden-service/src/config/project.ts
```diff
@@ -84,6 +81,15 @@ export const environmentNameSchema = joiUserIdentifier()
 
 export const environmentSchema = environmentConfigSchema.keys({
   name: environmentNameSchema,
+  production: joi.boolean().default(false).description(dedent`
+      Set environment as production.
+
+      Setting this flag to \`true\` will activate the protection on the \`deploy\`, \`test\`, \`task\`, \`build\`, \`call\`, \`init\` and \`dev\` commands.
```
edvald: Remove `call`

PR #1340, opened 2019-11-12
improvement(k8s): update rsync image used for build sync and hot reload (WIP)

edvald:

The new image is lighter and simpler, and does the same job.


eysi09:
Is this still a WIP?

PR #1342, opened 2019-11-12
fix(k8s): helm status check now compares Garden version

edvald:

Turns out `helm status` doesn't do any meaningful comparison checks
on its own (in hindsight I really don't why I thought it did), so
updates to `helm` modules would get ignored without the `--force` flag.

We can rely on our version hash for comparison, however, since it
includes hashes of all the templates and the config passed to the
template.

Perhaps try messing with the `vote-helm` project, triggering reloads etc., and see if everything happens as expected.


eysi09:
I'm having some issues with the `vote-helm` example but I don't think that it's related to this PR. Basically the `vote` page won't load after re-deploy, I just get an empty response. Pod looks healthy though. 

This also appears to happen on master so I'll file it separately and approve this one.

PR #1343, opened 2019-11-12
fix: displaced/wrong logging when setting up cert-manager

10ko:
Fix the cert-manager logger appearing when the certManager.install property
was set to "false".

Bonus: fix triggering the creation of certificates when cert-manager is not installed
even if no certificates managed by cert-manager are present in the config.

Fixes #

### DIFF
--- a/garden-service/src/plugins/kubernetes/integrations/cert-manager.ts
+++ b/garden-service/src/plugins/kubernetes/integrations/cert-manager.ts
@@ -197,81 +197,94 @@ export interface SetupCertManagerParams {
  * @param {SetupCertManagerParams} { ctx, provider, log, status }
  */
 export async function setupCertManager({ ctx, provider, log, status }: SetupCertManagerParams) {
-  const entry = log.info({
-    section: "cert-manager",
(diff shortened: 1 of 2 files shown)


garden-service/src/plugins/kubernetes/integrations/cert-manager.ts
```diff
@@ -251,26 +260,29 @@ export async function setupCertManager({ ctx, provider, log, status }: SetupCert
...
+      certsLog.setState('Certificates created and "Ready"')
+    } else {
+      certsLog.setState("No certificates found...")
     }
-    await waitForResourcesWith(waitForCertificatesParams)
-
-    entry.setState('Certificates created and "Ready"')
   }
-
   entry.setSuccess({ msg: chalk.green(`Done (took ${entry.getDuration(1)} sec)`), append: true })
```
edvald: Tests fail here because `entry` may be undefined.

10ko: I mistaken it for an error with the dev cluster configuration.
Fixed, thanks.

PR #1346, opened 2019-11-13
fix(sync): fix intermittent concurrency issues while syncing directories

edvald:

If two processes would copy at the same time to a target, one could
delete the files from the other. We now set rsync's temp directory
outside of the target directory, which resolves the problem.

### DIFF
--- a/garden-service/src/plugins/kubernetes/container/build.ts
+++ b/garden-service/src/plugins/kubernetes/container/build.ts
@@ -13,7 +13,7 @@ import { containerHelpers } from "../../container/helpers"
 import { buildContainerModule, getContainerBuildStatus, getDockerBuildFlags } from "../../container/build"
 import { GetBuildStatusParams, BuildStatus } from "../../../types/plugin/module/getBuildStatus"
 import { BuildModuleParams, BuildResult } from "../../../types/plugin/module/build"
-import { getPods, millicpuToString, megabytesToString } from "../util"
+import { millicpuToString, megabytesToString, getRunningPodInDeployment } from "../util"
 import { systemNamespace } from "../system"
 import { RSYNC_PORT } from "../constants"
 import { posix, resolve } from "path"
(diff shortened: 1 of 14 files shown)


garden-service/src/plugins/kubernetes/hot-reload.ts
```diff
@@ -296,11 +302,12 @@ export async function syncToService({ ctx, service, hotReloadSpec, namespace, wo
 
     const syncResult = await Bluebird.map(hotReloadSpec.sync, ({ source, target }) => {
       const src = rsyncSourcePath(service.sourceModule.path, source)
-      const destination = `rsync://localhost:${portForward.localPort}/volume/${rsyncTargetPath(target)}`
+      const destination = `rsync://localhost:${portForward.localPort}/volume/root/${rsyncTargetPath(target)}`
+      const tmpDir = `/tmp/${rsyncTargetPath(target)}`.slice(0, -1)
```
eysi09: Could you add a comment saying that the slice is to remove the trailing slash?

garden-service/src/util/string.ts
```diff
@@ -77,3 +77,10 @@ export function naturalList(list: string[]) {
     return list.slice(0, -1).join(", ") + " and " + list[list.length - 1]
   }
 }
+
+/**
+ * Generate a random string of a specified `length`.
+ */
+export function randomString(length = 8) {
```
eysi09: This doesn't appear to be used anywhere.

garden-service/src/plugins/kubernetes/container/build.ts
```diff
@@ -140,13 +140,32 @@ const remoteBuild: BuildHandler = async (params) => {
...
   // https://stackoverflow.com/questions/1636889/rsync-how-can-i-configure-it-to-create-target-directory-on-server
   let src = normalizeLocalRsyncPath(`${buildRoot}`) + `/./${module.name}/`
-
   const destination = `rsync://localhost:${syncFwd.localPort}/volume/${ctx.workingCopyId}/`
+  const syncArgs = ["-vrpztgo", "--relative", "--delete", "--temp-dir", "/tmp", src, destination]
 
   log.debug(`Syncing from ${src} to ${destination}`)
 
+  // TODO: remove this after a few releases (from 0.10.15), since this is only necessary for environments initialized
+  //       with 0.10.14 or earlier.
```
eysi09: Minor comment but the `with 0.10.14...` part is oddly indented.

garden-service/src/util/string.ts
edvald: No, I thought I needed to generate a temp folder per transfer, but then I didn't. But I figure it's fine to keep this, it's a basic utility we'll probably use.

garden-service/src/plugins/kubernetes/container/build.ts
edvald: I guess that's a matter of taste :P I'll change it, don't mind either way.

edvald:
Addressed. And yes, tested quickly on Windows (in a VM). Plus this is of course implicitly tested in our CI.

PR #1348, opened 2019-11-14
Fix logs

### LINKED ISSUE #1246: Logs stop loading after hot reload event
### Current Behavior

Logs stop loading after a hot reload event. This is because Garden looks for the pod with the latest content hash version but the pod hasn't been redeployed in the cluster (because hot reload).

See this comment for details: https://github.com/garden-io/garden/issues/1226#issuecomment-535222652

eysi09:

This fixes an issue where Garden wouldn't find pods because the version in
the cluster didn't match the version Garden expected after a hot reload
event. This would e.g. cause issues for the 'logs' and 'exec' commands.

Fixes #1246 (and an undocumented issue where `garden exec my-module <cmd>` wouldn't work after a hot reload event).

### DIFF
--- a/garden-service/src/plugins/kubernetes/util.ts
+++ b/garden-service/src/plugins/kubernetes/util.ts
@@ -7,7 +7,7 @@
  */
 
 import Bluebird from "bluebird"
-import { get, flatten, uniqBy, sortBy, sample } from "lodash"
+import { get, flatten, uniqBy, sortBy, omit, chain, sample, isEmpty } from "lodash"
 import { V1Pod, V1EnvVar } from "@kubernetes/client-node"
 
 import { KubernetesResource, KubernetesWorkload, KubernetesPod, KubernetesServerResource } from "./types"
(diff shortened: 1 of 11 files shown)


garden-service/src/plugins/kubernetes/util.ts
```diff
@@ -79,10 +79,34 @@ export function getSelectorFromResource(resource: KubernetesWorkload) {
...
+    .uniqBy((pod) => JSON.stringify(pod.metadata.labels))
+    .value()
+  return sortBy([...uniqByLabel, ...noLabel], (pod) => pod.metadata.creationTimestamp)
+}
+
+/**
+ * Retrieve a list of pods based on the resource selector, deduplicated so that only the most recent
+ * pod is returned when multiple pods with the same label are found.
+ */
+export async function getDeduplicatedWorkloadPods(api: KubeApi, namespace: string, resource: KubernetesWorkload) {
```
edvald: "Deduplicated" is maybe slightly odd naming, since the Pods won't be duplicated as such. How about `getMostRecentWorkloadPods` or `getCurrentWorkloadPods`?

PR #1349, opened 2019-11-14
docs: improve environment production flag docs

edvald:


docs/guides/remote-kubernetes.md
```diff
@@ -170,12 +170,17 @@ environments:
     ...
 ```
 
-## Production flag
+## Deploying to production
 
-You can define a remote environment as `production` by setting the [production flag](../reference/config.md#environmentsproduction).
+Depending on your setup and requirements, you may or may not want to use Garden to deploy to your production environment. In either case, if you do configure your production environment in your Garden project configuration, we highly recommend that you set the [production flag](../reference/config.md#environmentsproduction) on it.
```
eysi09: Skip "In either case..." and just start with "If you _do_ configure your..."  (with or without the emphasis on _do_).

PR #1350, opened 2019-11-14
fix(windows): normalize path for sync temp directory

edvald:

Forgot to normalize the path we use for the local build sync temp directory, so we got permission errors in testing.


edvald:
Note: Would be good to figure out how this didn't come up in CI...

PR #1360, opened 2019-11-23
allow plugins to augment the stack graph + hadolint plugin

edvald:

Providers are now able to augment the stack graph using a new `augmentGraph` handler, which is called in dependency order for each provider that specifies it. This allows providers to add modules (and by extension services, tasks, tests), and to add dependency relations between existing modules and services/tasks.

Also made a small refactor to no longer require a build handler for every module type, and to avoid unnecessary build tasks when a module doesn't require building.

The `hadolint` provider is a fairly simple plugin that uses the above framework features. Simply add the `hadolint`  provider to your project config, and a linting test will be created for each `container` module in the project. See the `hadolint` provider and `hadolint` module type reference docs for details and more usage options.

Please try using the `hadolint` provider. The added tests should also explain the intended functionality fairly well.


garden-service/src/config/module.ts
```diff
@@ -74,19 +74,25 @@ export interface BaseBuildSpec {
 
 export interface ModuleSpec {}
 
-export interface BaseModuleSpec {
-  apiVersion: string
+export interface AddModule {
```
eysi09: The interface name is a little confusing out of context.

garden-service/src/plugins.ts
```diff
@@ -129,6 +97,45 @@ export function loadPlugins(log: LogEntry, registeredPlugins: PluginMap, configs
...
+    if (err.cyclePath) {
+      throw new PluginError(`Found a circular dependency between registered plugins: ${err.cyclePath.join(" -> ")}`, {
+        cyclePath: err.cyclePath,
+      })
+    } else {
+      throw err
+    }
+  }
+
+  return configs.concat().sort((a, b) => {
```
eysi09: Comment about mutating arrays from above missing.

garden-service/src/tasks/build.ts
```diff
@@ -122,3 +123,41 @@ export class BuildTask extends BaseTask {
     return result
   }
 }
+
+/**
+ * Use this method to get the build tasks for a module. This is needed to be able to avoid an unnecessary build step
+ * when there is no build handler and no dependency files to copy.
+ */
+export async function getBuildTasks(params: BuildTaskParams): Promise<BuildTask[]> {
```
eysi09: Would it make sense to do this in a more "`TaskGraph` native" way? For example add a `getTasks` function to the `BaseTask` class. Feels a little weird to special case the build tasks like this and perhaps something one might forget about.

garden-service/src/types/plugin/provider/augmentGraph.ts
```diff
@@ -0,0 +1,106 @@
...
+import { Provider, providerSchema } from "../../../config/provider"
+import { moduleSchema, Module } from "../../module"
+
+export interface AugmentGraphParams extends PluginActionParamsBase {
+  modules: Module[]
+  providers: Provider[]
+}
+
+interface AddDependency {
+  by: string
```
eysi09: Is it just me or does `by` sound a little off in this context? "Add a dependency by module-a on module-b"? Would _for_ make more sense here?

garden-service/src/plugins/hadolint/hadolint.ts
```diff
@@ -0,0 +1,223 @@
...
+
+      if (!provider.config.autoInject) {
+        return {}
+      }
+
+      return {
+        addModules: await Bluebird.filter(modules, async (module) => {
+          return module.compatibleTypes.includes("container") && (await containerHelpers.hasDockerfile(module))
+        }).map((module) => {
+          return {
```
eysi09: Don't we need a test spec here? Otherwise we can't get test results via the CLI and/or dashboard.

examples/hadolint/README.md
```diff
@@ -0,0 +1,3 @@
+# hadolint project
+
+A simple variation on the [demo-project](../demo-project/README.md) that adds the [hadolint provider](https://docs.garden.io/reference/providers/hadolint.md). This generates an additional Dockerfile linting test for each `container` module in your project that contains a Dockerfile.
```
eysi09: Should we add some instructions for breaking the lint so that users can see it in action?

garden-service/src/actions.ts
```diff
@@ -139,7 +139,7 @@ export class ActionRouter implements TypeGuard {
     this.moduleActionHandlers = <WrappedModuleActionMap>fromPairs(moduleActionNames.map((n) => [n, {}]))
     this.loadedPlugins = keyBy(loadedPlugins, "name")
 
-    garden.log.silly(`Creating ActionRouter with ${configuredPlugins.length} configured plugins`)
+    garden.log.silly(`Creating ActionRouter with ${configuredPlugins.length} configured provider`)
```
eysi09: Missing _s_ in _provider_ :)

garden-service/src/types/plugin/provider/augmentGraph.ts
eysi09: Or _from - to_.

garden-service/src/garden.ts
```diff
@@ -570,150 +569,109 @@ export class Garden {
...
-          schema: description.schema,
-          name: config.name,
-          path: config.path,
-          projectRoot: this.projectRoot,
-        })
+    // Walk through all plugins in dependency order, and allow them to augment the graph
+    for (const provider of getDependencyOrder(providers, this.registeredPlugins)) {
+      // Skip the routine if the provider doesn't have the handler
+      try {
+        await actions.getActionHandler({ actionType: "augmentGraph", pluginName: provider.name })
```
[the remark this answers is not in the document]
edvald: Yeah perhaps, I was just trying to avoid having to do the inverse, which is to catch a potential null/undefined in every other place we call this. I can look at it to see how much work that is and evaluate based on that.

garden-service/src/garden.ts
```diff
@@ -570,150 +569,109 @@ export class Garden {
...
 
-      if (config.plugin) {
-        for (const serviceConfig of config.serviceConfigs) {
-          serviceConfig.name = getModuleKey(serviceConfig.name, config.plugin)
-        }
-        for (const taskConfig of config.taskConfigs) {
-          taskConfig.name = getModuleKey(taskConfig.name, config.plugin)
+        for (const moduleConfig of moduleConfigs) {
+          for (const serviceConfig of moduleConfig.serviceConfigs) {
+            if (serviceConfig.name == dependency.by) {
```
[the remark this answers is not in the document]
edvald: No, not even sure how that got through. I thought we were linting for this?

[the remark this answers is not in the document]
eysi09: I thought so too.

garden-service/src/types/plugin/provider/augmentGraph.ts
edvald: I thought about this, and felt this was most reasonable, even if it feels slightly odd. Grammatically, _from_ -> _to_ doesn't hold up, since I don't believe you can correctly say something is a dependency _from_ A _to_ B. My reasoning was that B is depended _on_, _by_ A.

We have used `dependency`/`dependant` in our own code, which is imo a whole other kind of confusing, but I remember lots of discussions trying to find alternatives but never had a better result.

garden-service/src/plugins/hadolint/hadolint.ts
edvald: The `configure` handler for the module type adds that.

garden-service/src/tasks/build.ts
edvald: What I'd ultimately like to do is have the ConfigGraph specify these, and to make that the canonical "source of truth" when it comes to generating tasks for specific modules, services etc. But I felt that sort of a refactor would be out of scope.

I can try to think of something more clear in the meantime though, maybe a static method on BuildTask, something like that.

garden-service/src/types/plugin/provider/augmentGraph.ts
edvald: Or I don't know. Opinions welcome, but from/to seems just as odd to me :P

examples/hadolint/README.md
edvald: Added some more instructions

garden-service/src/tasks/build.ts
edvald: I've updated this to place it at `BuildTask.factory()`  and set a guard attribute to avoid accidentally using the constructor directly.

garden-service/src/types/plugin/provider/augmentGraph.ts
eysi09: Yeah, I agree that _from -> to_ isn't good. My reasoning with _for_ was that you're _adding a dependency for A, on B_. Or for that matter, _adding dependency to A, on B_.

But all of these could work imo. I'll leave this up to you :)

PR #1361, opened 2019-11-24
Docs: Openfaas reference states wrong module type

ozburo:
Openfaas module reference documentation states wrong module type `container` should be `openfaas`

Simple documentation typo/error fix

Fixes # N/A

### DIFF
--- a/docs/reference/module-types/openfaas.md
+++ b/docs/reference/module-types/openfaas.md
@@ -41,7 +41,7 @@ The type of this module.
 Example:
 
 ```yaml
-type: "container"
+type: "openfaas"
 ```
 
 ### `name`


eysi09:
Thanks for the PR @ozburo!

Actually, this part of the docs is auto-generated, based on a schema that we define in-code for all our module types. So updates to the markdown file will just be overwritten next time we generate the docs.

To complicate matters further, all modules types (`openfaas` included) extend the base module schema, and it's there where the `type` field is defined. So the module types themselves don't define it but rather inherit it. 

So basically, what it's saying is that all module types have a `type` field and an example of that would be `container`.

Therefore to fix this, we'll have to update the code that generates the docs and overwrite the "Example" value on the "Type" field for modules. This is not going to make the generate docs code any prettier though (https://github.com/garden-io/garden/blob/master/garden-service/src/docs/config.ts).

ozburo:
@eysi09 Awe thank you for your informative and understanding reply! I thought I'd help with a little quickfix and didn't realise your docs are auto-generated (I learned something today!)  👍

PR #1364, opened 2019-11-27
CLI fancy logger improvements

edvald:

This was prompted by issues found when trying to record Garden terminal sessions with asciinema. That previously came out completely wrong, but terminal-kit seems to do the job better. It'll also be useful for other things later.

Also, fixed an issue where hidden verbose+ logs would cause spinners to stutter.

n/a

Please test in a few different scenarios, even platforms. I tried to find issues, nothing came up, but I didn't test all possible setups and commands.


edvald:
Didn't try on Windows, no.

edvald:
Turns out this doesn't play nice with Powershell (but seems fine with the new Windows Terminal). Putting a pin in this until that's sorted out.

edvald:
I've updated this to special-case Windows, so it should work as expected on all platforms now.

PR #1367, opened 2019-11-29
improvement(config): allow provider configs to reference variables

edvald:

We previously were unable to include variables in the templating context for provider configs, but after recent refactors this is now trivial to allow.


eysi09:
Looks like the OpenFaaS project keeps failing.

edvald:
That test has been flaky for a while, we should investigate that.

eysi09:
Like a new baked croissant.

PR #1375, opened 2019-12-02
chore(ci): more optimal CircleCI pipelines

edvald:

We were doing unnecessary work on tags and master merges, this should save some time and cycles.

Also separating e2e tests into individual tests now, instead of running all from one container.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -199,7 +199,8 @@ jobs:
             npm test
           environment:
             CHOKIDAR_USEPOLLING: "1"
-  e2e-test-service:
+  # (not used atm)
+  e2e-all:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -408,134 +437,96 @@ jobs:
...
   commit:
     jobs:
+      ### ALL BRANCHES/TAGS ###
+
       - build
       - build-dist:
-          requires:
-            - build
+          requires: [build]
+      - lint:
```
eysi09: Can't we remove the `requires: [build]`?

edvald: For the lint step? No, unfortunately not, because of the check-docs script.

PR #1378, opened 2019-12-04
improvement: add protection to more commands

eysi09:
Set `protected = true` on delete commands.


edvald:
Could we add tests for this on individual commands? Or at least assertions that the flag is set on them. I'd hate for us to accidentally remove the protection in some refactor.

baspetersdotjpeg:
Fixes the issue 👍

PR #1384, opened 2019-12-05
chore(ci): test against more Kubernetes versions with minikube

edvald:

It's good for us to know which versions work reliably with Garden, and of course to catch any backwards-incompatible changes we make. These all run in parallel, so we shouldn't be extending our testing time.


edvald:
It'll disappear on merge, because the patch will be empty for that commit

PR #1394, opened 2019-12-06
improvement(cli): remove `garden init` command

edvald:

This command is no longer needed and its existence (and name) just
caused confusion.

BREAKING CHANGE:

Any user scripts that run `garden init` will need to be updated to
remove those references.


garden-service/src/commands/delete.ts
```diff
@@ -84,7 +84,7 @@ export class DeleteEnvironmentCommand extends Command {
 
   description = dedent`
     This will delete all services in the specified environment, and trigger providers to clear up any other resources
-    and reset it. When you then run \`garden init\` or \`garden deploy\`, the environment will be reconfigured.
+    and reset it. When you then run \`garden prepare\` or \`garden deploy\`, the environment will be reconfigured.
```
eysi09: We don't have a `prepare` command.

edvald: 👍 I left that there, the original plan had been to rename it. I'll remove the reference.

edvald: Fixed.

Issue #1382, opened 2019-12-04
Helm module validation should be performed only after module's dependencies are built


khaled:
## Bug

I've got a helm module that depends on an exec module which runs a custom tool to generate the chart from a custom artifact:

```
kind: Module
description: Generates helm chart using custom tool
name: my-service-chart-builder
type: exec
build:
  command:
    - chart-builder
    - build
    - o ./build
    - <args>
---
kind: Module
description: Node service helm chart
name: my-service
type: helm
build:
  dependencies:
	- name: backend-image
    - name: my-service-chart-builder
      copy:
        - source: ./build
          target: ./
chartPath: ./build/
```

### Current Behavior

Garden attempts to validate that a chart exists at `chartPath` before the `my-service-chart-builder` dependency completes and files are copied from it.  This validation fails because the chart hasn't been generated and nothing is there yet.

### Expected behavior

Validation doesn't happen until the build dependency is finished doing its thing and the dependent files are copied over.

### Reproducible example

Stick a chart in an exec module, have the exec module execute some dummy command, then have a helm module as a dependency that just copies the chart files from the exec module.

### Workaround

Put a dummy chart in the helm module so that validation passes, later allowing it to get overwritten by `copy` command in the dependency.  This seems to only work the first time; after making a source change (in `garden dev` mode), the dummy chart seems to get written back to the staging area (`.build`), overwriting the dummy chart, and is then deployed :(

### Suggested solution(s)

Implement the expected behavior :) 

### Your environment

```
$ garden --version

0.10.15
```

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.3", GitCommit:"b3cbbae08ec52a7fc73d334838e18d17e8512749", GitTreeState:"clean", BuildDate:"2019-11-18T14:56:51Z", GoVersion:"go1.12.13", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.9", GitCommit:"e09f5c40b55c91f681a46ee17f9bc447eeacee57", GitTreeState:"clean", BuildDate:"2019-05-27T15:58:45Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}

```
```
$ microk8s.docker version
Client:
 Version:           18.09.2
 API version:       1.39
 Go version:        go1.10.4
 Git commit:        6247962
[...]

PR #1400, opened 2019-12-06
refactor(k8s): avoid building Helm chart before getting deploy status

edvald:

This helps avoids an issue when generating charts in other modules.

Relates to #1382

### DIFF
--- a/garden-service/src/plugins/kubernetes/helm/common.ts
+++ b/garden-service/src/plugins/kubernetes/helm/common.ts
@@ -42,7 +42,7 @@ export async function containsSource(config: HelmModuleConfig) {
 /**
  * Render the template in the specified Helm module (locally), and return all the resources in the chart.
  */
-export async function getChartResources(ctx: PluginContext, module: Module, log: LogEntry) {
+export async function getChartResources(ctx: PluginContext, module: Module, hotReload: boolean, log: LogEntry) {
   const chartPath = await getChartPath(module)
   const k8sCtx = <KubernetesPluginContext>ctx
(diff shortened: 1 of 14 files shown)


garden-service/src/plugins/kubernetes/helm/common.ts
```diff
@@ -146,15 +146,21 @@ export function getGardenValuesPath(chartPath: string) {
...
 
   // The garden-values.yml file (which is created from the `values` field in the module config) takes precedence,
   // so it's added to the end of the list.
   const valueFiles = module.spec.valueFiles.map((f) => resolve(module.buildPath, f)).concat([gardenValuesPath])
 
-  return flatten(valueFiles.map((f) => ["--values", f]))
+  const args = flatten(valueFiles.map((f) => ["--values", f]))
+
+  if (hotReload) {
+    args.push("--set", '"\\.garden".hotReload=true')
```
eysi09: Can we test that the `garden-values` file renders as expected?

edvald: This doesn't affect the value file as such, but it might be good to validate that the hotReload flag gets set properly by Helm.

edvald:
This relates to the headaches around builds and build staging clashing for Helm charts because they can occur in an unpredictable order. This caused problems e.g. when generating Helm charts using an exec module before building+deploying it.

edvald:
I've added an integ test for the helm deployment handler @eysi09. I think it's good to go, if tests pass.

PR #1401, opened 2019-12-06
improvement(k8s): switch to Helm 3 and remove Tiller

### LINKED ISSUE #1230: Support for Helm 3 / Tillerless
With Helm 3 currently in beta, maybe we can start looking into optional support for it.  I would like to keep Tiller out of my clusters. :)

edvald:

This removes Tiller and any related resources. Migration is handled
automatically, with some caveats (see below).

BREAKING CHANGE:

Helm 2.x is no longer supported. The migration (both for garden-system
services and your project namespace) is handled automatically
via the `helm 2to3` plugin. It is _possible_ that the automatic
migration fails though (due to any number of potential issues with
Helm or issues exposed with individual charts upon upgrade).

We've tried to cover and test for these  cases as best we can, but can't
rule out issues, so you may need to intervene (by e.g. manually removing
resources or using the helm CLI directly) if migration or upgrades after
deployment throw errors.

If you do run into tricky issues, please don't hesitate to log issues
on GitHub or ping us on Slack and we'll be happy to help.

Closes #1230

The transition from Helm 2 to 3 is probably the most tricky part here, and the most difficult to test to completion. I did my best, but please try the upgrade on any dev cluster you have and see if issues come up. I've already done the upgrade on our CI cluster.


garden-service/src/actions.ts
```diff
@@ -565,12 +565,12 @@ export class ActionRouter implements TypeGuard {
   /**
    * Deletes all services and cleans up the specified environment.
```
eysi09: Update comment since the function appears to only delete the services.

garden-service/src/plugins/kubernetes/helm/tiller.ts
```diff
@@ -11,96 +11,114 @@ import { KubernetesResource } from "../types"
...
-  await apply({ log, provider, manifests: roleResources, namespace })
-  await waitForResources({
-    ctx,
-    provider,
-    serviceName: "tiller",
-    resources: roleResources,
-    log: entry,
-  })
+    log.debug(
+      await helm({
```
eysi09: Does this throw if the conversion fails?

garden-service/src/plugins/kubernetes/kubectl.ts
```diff
@@ -65,13 +65,29 @@ export async function apply({
...
+  const args = [
+    "delete",
+    "--wait=true",
+    "--ignore-not-found=true",
+    ...resources.map((r) => `${r.kind}/${r.metadata.name}`),
+  ]
+
+  includeUninitialized && args.push("--include-uninitialized")
+
+  return kubectl.stdout({ provider, namespace, args, log })
```
eysi09: Can't we use the API for this?

edvald: Not if we want to wait for the resources to be terminated. That is, without re-implementing all that logic. In any case, this is executed so rarely that the inefficiency is fine.

garden-service/src/plugins/kubernetes/helm/tiller.ts
edvald: It does.

PR #1404, opened 2019-12-08
docs: put YAML schema at top of reference docs

eysi09:

Move the YAML schema to the top of the reference docs, above the list of keys and their descriptions. 

This has come up in internal conversations before. This was also pointed out by a user recently who wasn't even aware of the complete YAML schema reference below and found the long list of keys confusing.

Check if you think this makes sense.


edvald:
We used to have field descriptions inlined as comments. Think we could re-introduce that?

eysi09:
Yeah, that shouldn't be too difficult. Just the description or the example as well?

edvald:
Maybe just the description for now, see how it looks. 

I wonder if it's possible to have links in there, to the full reference? Seems like it might be possible on GitHub, unsure about GitBook: https://gist.github.com/jesstelford/cb4dd6fafc18221ce27250e84fd19327

eysi09:
I've added the description string to the YAML. I couldn't get the links to work though.

PR #1406, opened 2019-12-08
docs(examples): add base-image example

eysi09:

Add an example for using base Docker images with Garden.

This gets asked about a lot on our community Slack, and a user specifically asked for us to put an example like this into our examples dir.


examples/base-image/garden.yml
```diff
@@ -0,0 +1,9 @@
+kind: Project
+name: demo-project
+environments:
+  - name: dev
+    providers:
+      - name: local-kubernetes
+  - name: prod
+    providers:
+      - name: local-kubernetes
```
edvald: I'd suggest using the top-level `providers` key here. We're deprecating the `environments.providers[]` key.

eysi09: Ah yes. Good point.

eysi09: I've updated the PR.

examples/base-image/garden.yml
```diff
@@ -0,0 +1,8 @@
+kind: Project
+name: demo-project
+environments:
+  - name: dev
+  - name: prod
+providers:
+  - name: local-kubernetes
+    evnironments: [dev, prod]
```
edvald: This line is not necessary if the provider covers all environments. A matter of taste then, if it should be specified in the example, since it's technically correct, just a touch more verbose.

PR #1410, opened 2019-12-09
improvement(k8s): update bundled nginx controller version

edvald:

We hadn't updated this in a while, which a user pointed out on our Slack.

### DIFF
--- a/garden-service/static/kubernetes/system/ingress-controller/garden.yml
+++ b/garden-service/static/kubernetes/system/ingress-controller/garden.yml
@@ -6,14 +6,20 @@ chart: stable/nginx-ingress
 releaseName: garden-nginx
 dependencies:
   - default-backend
-version: 0.25.1
+version: 1.26.2
 values:
   name: ingress-controller
   controller:
     defaultBackendService: ${var.namespace}/default-backend
     kind: DaemonSet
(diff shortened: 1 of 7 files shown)


garden-service/static/kubernetes/system/ingress-controller/garden.yml
```diff
@@ -6,14 +6,20 @@ chart: stable/nginx-ingress
...
     defaultBackendService: ${var.namespace}/default-backend
     kind: DaemonSet
+    updateStrategy:
+      type: RollingUpdate
+      rollingUpdate:
+        maxUnavailable: 1
     daemonset:
       useHostPort: true
+    service:
+      omitClusterIP: true
```
10ko: From the [README](https://github.com/helm/charts/tree/master/stable/nginx-ingress) looks like `omitClusterIP` is deprecated.
Is this set becuase of the issue `Error: UPGRADE FAILED: Service "?????-controller" is invalid: spec.clusterIP: Invalid value: "": field is immutable`?

edvald: Yes, exactly. I'm unsure why it's deprecated but it seemed necessary to upgrade without issues.

PR #1416, opened 2019-12-10
Use top level providers

eysi09:

Make providers top-level in docs, examples and tests, since the other style is deprecated. That is, this:

```yaml
kind: Project
name: my-project
environments:
  - name: local
    providers:
      - name: local-kubernetes
  - name: remote
    providers:
      - name: kubernetes
```

becomes this:

```yaml
kind: Project
name: my-project
environments:
  - name: local
  - name: remote
providers:
  - name: local-kubernetes
    environments: [local]
  - name: kubernetes
    environments: [remote]
```

I added a specific test for providers nested under environments in `test/unit/src/config/project.ts` since we now default to the other style in our tests. Also removed tests for flat config style since support for the nested style has been removed and all test projects now use the flat style.


examples/deployment-strategies/garden.yml
```diff
@@ -1,28 +1,27 @@
...
+    environments: [local]
+    # Same as above but explicit
+  - name: local-kubernetes
+    environments: [local-default]
+    deploymentStrategy: rolling
+    # Blue-green deployment strategy.
+  - name: local-kubernetes
+    environments: [local-blue-green]
+    deploymentStrategy: blue-green
+    # Testing environment
```
edvald: Would this look cleaner if the deploymentStrategy was set as a variable on the environments?

edvald: Probably would, _if_ we allowed wildcard matching on the environment names, e.g. `[local*]`... Maybe a nice idea to log?

garden-service/test/unit/src/config/base.ts
```diff
@@ -256,57 +259,6 @@ describe("loadConfig", () => {
     ])
   })
 
-  it("should parse a config file using the flat config style", async () => {
```
edvald: Why remove this?

garden-service/test/unit/src/config/base.ts
```diff
@@ -324,15 +276,6 @@ describe("loadConfig", () => {
     ])
   })
 
-  it("should throw an error when parsing a flat-style config using an unknown/invalid kind", async () => {
```
edvald: And this?

garden-service/test/unit/src/config/base.ts
eysi09: Because we no longer support the nested config style, and all configs therefore have the flat style. I mentioned this in the notes above.

garden-service/test/unit/src/config/base.ts
eysi09: See my comment above.

examples/deployment-strategies/garden.yml
eysi09: It's logged. Should we leave this as is then?

garden-service/test/unit/src/config/base.ts
edvald: Right, I just thought the test was still valid, but the description wasn't. So this same test is actually performed elsewhere then?

Actually the PR note says the "flat style" has been removed, which it hasn't. The nested style was removed.

examples/deployment-strategies/garden.yml
edvald: Yeah that's fine.

garden-service/test/unit/src/config/base.ts
eysi09: All test projects now use the flat style so the tests above implicitly test for what this is test is testing for. But we can leave it in as well and re-name the project and update the test description.

eysi09: ...and yeah, the PR note was wrong. My bad.

edvald: If the other tests cover this, that's fine. But we should keep the one you deleted below, right? Or is there another test for that case specifically?

eysi09: I've added a new test for the case below (`should throw if a config specifies an invalid kind`) and added a `variables` field to the `test-project-multiple-modules` project so that the `should load and parse a config file defining a project and a module` test now covers all the cases the test that I removed covered.

edvald: Still waiting to push that change? Or am I losing my mind and not seeing it somewhere? (I won't rule that out 😛)

eysi09: Bah, apparently forgot to push. It's been a long week...

PR #1418, opened 2019-12-10
improvement: switch to Node.js 12

### LINKED ISSUE #1407: Support building garden with node 12.x/13.x
Our current setup for developing garden only supports node `10.x.x`.

Please, migrate dependencies to node 12.x/13.x.

Known issues:
 - `fsevents 2.x.x` is needed as explicit dependency when building on MacOS on `node 12.x.x`.
 - Building with `node 12.x.x` on Linux fails.
 - When explicitely specifiying `fsevents` as `devDependency`, the build will fail on Linux

cc: @solomonope

edvald:

It is time. Node 12 is LTS now, and we don't want to be laggards, do we?

Also added a Debian container image, in order to sanity check our linux-amd64 binary.

Closes #1407 

This should work without any user-facing changes.


edvald:
Unfortunately, it turns out the alpine build doesn't work, so we may need to postpone until that's fixed in some later release (or I stumble across some solution).

edvald:
Alpine issue resolved. This should be good to go, but we should make sure to test it well, and I'd even suggest deferring until 0.11.0 is out.

PR #1419, opened 2019-12-10
fix(build): validate helm modules at build rather than configure phase

khaled:
This addresses #1382 by moving the check for whether or not a helm module
contains sources to the build phase, rather than the configure phase.

This also moves the the logic for syncing dependency products to the build
phase rather than the "stage build" phase.

This corrects the case where a module's build depends on artifacts
generated by and copied over from a dependency.


eysi09:
Thanks a lot for the contribution @khaled! It is very much appreciated.

I'll leave the review to @edvald, who happens to be in transit today.

edvald:
This looks good @khaled! Just one comment: Could you split the exec build status to a separate commit?

khaled:
> This looks good @khaled! Just one comment: Could you split the exec build status to a separate commit?

Done!

PR #1420, opened 2019-12-11
chore: require include/exclude on overlapping modules

eysi09:

From the commit message:

BREAKING CHANGE:

It is now required to set the `include` and/or `exclude` directive on
modules that overlap. This is to prevent subtle errors that can occur
when two modules share the same build context.

Module level `garden.yml` files that have overlapping modules and no
includes/excludes will need to be updated. A common case is when there
are multiple modules in the same `garden.yml` file.

For example, this:

```yaml
kind: Module
name: module-a
type: container

---

kind: Module
name: module-b
type: container
```

becomes:

```yaml
kind: Module
name: module-a
type: container
include: ["src-a/**/*"]

---

kind: Module
name: module-b
type: container
include: ["src-b/**/*"]
```


garden-service/src/garden.ts
```diff
@@ -800,6 +807,24 @@ export class Garden {
...
+      if (overlappingModules.length > 0) {
+        const msg = overlappingModules
+          .map(
+            ({ module, overlaps }) =>
+              `Module ${chalk.bold(module)} overlaps with module(s) ${overlaps.map((o) => chalk.bold(o)).join(", ")}`
+          )
+          .join("\n")
+        const errorMsg = chalk.red(dedent`
+          Missing ${chalk.bold("include")} and/or ${chalk.bold("exclude")} directives on overlapping modules.
+          Setting includes/excludes is required for modules whose paths overlap.
```
edvald: This error message could be a touch nicer imo, and perhaps more explanatory. The meaning of "overlapping modules" may not be clear to the reader, for example. I'd lean towards more explicitly saying something like "when modules have the same path, or when one module is nested within another", which is more verbose but more helpful.

garden-service/test/unit/src/garden.ts
```diff
@@ -2044,6 +2044,25 @@ describe("Garden", () => {
...
       expect(getNames(moduleConfigs).sort()).to.eql(["module-a"])
     })
+
+    it("should throw a nice error if module paths overlap", async () => {
+      const projectRoot = getDataDir("test-projects", "multiple-module-config-bad")
+      const garden = await makeTestGarden(projectRoot)
+      await expectError(
+        () => garden.resolveModuleConfigs(),
+        (err) => {
+          expect(stripAnsi(err.message)).to.equal(deline`
```
edvald: I'd suggest using `dedent` here, to avoid the `\n` characters in the string

garden-service/src/garden.ts
```diff
@@ -800,6 +807,24 @@ export class Garden {
         }
       })
 
+      // Require include/exclude on modules if their paths overlap
+      const overlappingModules = detectModuleOverlap(rawConfigs)
+      if (overlappingModules.length > 0) {
+        const msg = overlappingModules
+          .map(
+            ({ module, overlaps }) =>
+              `Module ${chalk.bold(module)} overlaps with module(s) ${overlaps.map((o) => chalk.bold(o)).join(", ")}`
```
edvald: I think it could be good to differentiate here for the user whether modules overlap by nesting, or if they have the same path.

Also, there's a nice little helper here for more natural lists of strings: https://github.com/garden-io/garden/blob/715abe06c6c275e1a4f9dd9f4548d7c5c7396829/garden-service/src/util/string.ts#L71

eysi09: How would you formulate the error message in that case?

Something like:

```
Module module-a overlaps with modules module-b (same path), module-c (same path), module-d (nested)
```

perhaps?

eysi09: ...I originally thought of something like:
```
Modules a, b, and c all have the same path.
Module a also overlaps with...
Module d overlaps with...
``` 

But it just felt confusing so I opted for listing them like above.

edvald: Yeah I'm not 100% sure how best to do it, but both of your suggestions look like decent improvements. That plus a little more helpful text above should be fine.

eysi09:
Added another commit that addresses the comments.

PR #1422, opened 2019-12-11
fix(vcs): ensure module versions are stable between runtimes

eysi09:

Previously, module version hashes were contaminated with runtime specific content such as the module path and the `outputs` field.

This has now been fixed and tests added.

### DIFF
--- a/garden-service/src/config/module.ts
+++ b/garden-service/src/config/module.ts
@@ -219,7 +219,7 @@ export const moduleConfigSchema = baseModuleSpecSchema
   .description("The configuration for a module.")
   .unknown(false)
 
-export function serializeConfig(moduleConfig: ModuleConfig) {
+export function serializeConfig(moduleConfig: Partial<ModuleConfig>) {
   return stableStringify(moduleConfig)
 }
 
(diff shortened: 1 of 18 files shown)


examples/remote-sources-simple-2/garden.yml
```diff
@@ -0,0 +1,16 @@
+kind: Project
```
edvald: Is this supposed to be here?

garden-service/test/data/test-projects/fixed-modules-1/garden.yml
```diff
@@ -0,0 +1,10 @@
+kind: Project
+name: test-project-fixed-modules-1
```
edvald: Maybe rename `fixed-modules` to `fixed-version-hashes` for clarity?

edvald:
Couple of comments, otherwise 👍

examples/remote-sources-simple-2/garden.yml
eysi09: Nope, my bad.

eysi09:
I've addressed the comments.

PR #1430, opened 2019-12-14
feat(core): use service's port when forwarding to localhost if available

### LINKED ISSUE #1428: feat(core): use service's port when forwarding to localhost if available

mjgallag:

This PR uses the service's port when forwarding to localhost, if it's available, otherwise it uses a random available port. We need it because it's helpful for services to be on the expected localhost ports e.g. 3000 for Create React App, 27017 for MongoDB, 6379 for Redis.

Fixes #1428

### DIFF
--- a/garden-service/src/proxy.ts
+++ b/garden-service/src/proxy.ts
@@ -169,7 +169,7 @@ async function createProxy(garden: Garden, log: LogEntry, service: Service, spec
     })
   })
 
-  const localPort = await getPort()
+  const localPort = await getPort({ port: spec.targetPort })
   const host = `localhost:${localPort}`
   // For convenience, we try to guess a protocol based on the target port, if no URL protocol is specified
   const protocol = spec.urlProtocol || guessProtocol(spec)


mjgallag:
@edvald @eysi09 I don't seem to have permission to add reviewers even though the pull template asks me to. Also, minikube test error `ERROR: (gcloud.auth.activate-service-account) Could not read json file /home/circleci/gcloud-key.json: No JSON object could be decoded` is unrelated to this pull.

edvald:
Hey @mjgallag! Thanks so much for this. I'll take a look. And you're right, adding reviewers shouldn't be suggested in the PR template. The issue with CI is basically that we're presently relying on auth that we don't pass to 3rd-party PRs, but has nothing to do with the PR as such. I'll make sure to run those tests.

edvald:
This looks good, nice and simple. Just an additional note: I've been looking into a somewhat more advanced version of this, inspired by kubefwd, which is to use a different localhost IP (e.g. `127.1.27.1`) when the port isn't available on another local IP. We can leave that as a next step and merge this though.

PR #1437, opened 2019-12-16
fix(log): updated to stern from kubectl

solomonope:
This is to allow for tailing of logs to continue on service restarts
This PR fixes issue #411

Fixes #

### DIFF
--- a/garden-service/src/logger/log-entry.ts
+++ b/garden-service/src/logger/log-entry.ts
@@ -133,7 +133,8 @@ export class LogEntry extends LogNode {
       // Next state does not inherit the append field
       append: updateParams.append,
       timestamp: Date.now(),
-      maxSectionWidth: updateParams.maxSectionWidth || messageState.maxSectionWidth,
+      maxSectionWidth:
+        updateParams.maxSectionWidth !== undefined ? updateParams.maxSectionWidth : messageState.maxSectionWidth,
     }
 
     // Hack to preserve section alignment if spinner disappears
--- a/garden-service/src/logger/log-entry.ts
+++ b/garden-service/src/logger/log-entry.ts
@@ -41,6 +41,7 @@ interface MessageBase {
   symbol?: LogSymbol
   append?: boolean
   data?: any
+  maxSectionWidth?: number
 }
 
 export interface MessageState extends MessageBase {
@@ -106,6 +107,7 @@ export class LogEntry extends LogNode {
         symbol: params.symbol,
         status: params.level === LogLevel.error ? "error" : params.status,
         data: params.data,
+        maxSectionWidth: params.maxSectionWidth,
       })
     }
   }
@@ -131,6 +133,7 @@ export class LogEntry extends LogNode {
       // Next state does not inherit the append field
       append: updateParams.append,
       timestamp: Date.now(),
+      maxSectionWidth: updateParams.maxSectionWidth || messageState.maxSectionWidth,
     }
 
     // Hack to preserve section alignment if spinner disappears
--- a/garden-service/src/plugins/kubernetes/logs.ts
+++ b/garden-service/src/plugins/kubernetes/logs.ts
@@ -143,7 +143,9 @@ async function followLogs(
     "-t",
   ]
 
-  //Getting labels on the pod with no numbers, The Idea is these labels are less likely to change between different deployments of these pods
+  /* Getting labels on the pod with no numbers,
+  The Idea is these labels are less likely to change between different deployments of these pods
+  */
   const labels = getStaticLabelsFromPod(pod)
   if (Object.keys(labels).length > 0) {
     sternArgs.push(`${getSelectorString(labels)}`)
--- a/garden-service/src/plugins/kubernetes/logs.ts
+++ b/garden-service/src/plugins/kubernetes/logs.ts
@@ -11,8 +11,6 @@ import { omit } from "lodash"
 import moment = require("moment")
 
 import { GetServiceLogsResult, ServiceLogEntry } from "../../types/plugin/service/getServiceLogs"
-import { splitFirst } from "../../util/util"
-import { kubectl } from "./kubectl"
 import { KubernetesResource, KubernetesPod } from "./types"
 import { getAllPods } from "./util"
 import { KubeApi } from "./api"
@@ -21,6 +19,7 @@ import Stream from "ts-stream"
 import { LogEntry } from "../../logger/log-entry"
 import Bluebird from "bluebird"
 import { KubernetesProvider } from "./config"
+import { BinaryCmd } from "../../util/ext-tools"
 
 interface GetLogsBaseParams {
   defaultNamespace: string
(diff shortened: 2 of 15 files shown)


garden-service/src/plugins/kubernetes/logs.ts
```diff
@@ -74,38 +92,41 @@ export async function getAllLogs(params: GetAllLogsParams) {
...
+  const sternArgs = [
+    `--context=${provider.config.context}`,
+    `--namespace=${pod.metadata.namespace}`,
+    "--tail",
+    String(tail),
+    "-t",
+  ]
 
-  kubectlArgs.push(`pod/${pod.metadata.name}`)
+  sternArgs.push(`-l module=${service.module.name}`)
```
eysi09: Here's the tricky part. This will only work for the container module type. The `helm` and `kubernetes` module types are completely configured by the user so we don't have that label on them.

So we need to somehow find the pod that corresponds to helm/k8s resource with out relying on the `module` label.

garden-service/test/integ/src/plugins/kubernetes/container/logs.ts
```diff
@@ -59,7 +59,8 @@ describe("kubernetes", () => {
         follow: false,
         tail: -1,
       })
-
+      console.log(entries)
```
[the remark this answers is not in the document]
solomonope: This was to debug the integration test

garden-service/src/plugins/kubernetes/logs.ts
```diff
@@ -74,38 +92,41 @@ export async function getAllLogs(params: GetAllLogsParams) {
...
+      serviceName: service.name.trim(),
       timestamp,
       msg: `${pod.metadata.name} ${msg}`,
     })
   })
 
   return proc
 }
+function parseSternLogMessage(message: string): string[] {
+  const risk = message.split(" ")
```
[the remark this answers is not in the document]
solomonope: Thanks alot, these comments helped me find why the integration tests have been failing

garden-service/src/plugins/kubernetes/logs.ts
```diff
@@ -74,38 +92,41 @@ export async function getAllLogs(params: GetAllLogsParams) {
...
-    kubectlArgs.push("--follow=true")
-  }
+async function getLogs({ log, provider, service, stream, tail, pod }: GetLogsParams) {
+  const sternArgs = [
+    `--context=${provider.config.context}`,
+    `--namespace=${pod.metadata.namespace}`,
+    "--tail",
+    String(tail),
+    "-t",
+  ]
```
edvald: Stern has an option to output JSON (`--output json`). We should almost definitely be using that instead of splitting output by spaces.

solomonope: thanks, going to try that out

garden-service/src/plugins/kubernetes/logs.ts
solomonope: fixed this, to use non numeric labels on the pod as selectors

garden-service/src/plugins/kubernetes/logs.ts
```diff
@@ -74,38 +92,41 @@ export async function getAllLogs(params: GetAllLogsParams) {
...
 
-async function getLogs({ log, provider, service, stream, tail, follow, pod }: GetLogsParams) {
-  // TODO: do this via API instead of kubectl
-  const kubectlArgs = ["logs", "--tail", String(tail), "--timestamps=true", "--all-containers=true"]
-
-  if (follow) {
-    kubectlArgs.push("--follow=true")
-  }
+async function getLogs({ log, provider, service, stream, tail, pod }: GetLogsParams) {
+  const sternArgs = [
```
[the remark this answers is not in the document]
solomonope: I have able to do it with stern, so when you follow logs you won't see logs from the sync containers, I have not figured out to do it with kubectl because with kubectl you read logs from a container or all containers.

garden-service/src/plugins/kubernetes/logs.ts
solomonope: added

garden-service/src/logger/log-entry.ts
```diff
@@ -131,6 +133,7 @@ export class LogEntry extends LogNode {
       // Next state does not inherit the append field
       append: updateParams.append,
       timestamp: Date.now(),
+      maxSectionWidth: updateParams.maxSectionWidth || messageState.maxSectionWidth,
```
eysi09: @solomonope, just realised that this won't work because zero is considered falsey. 

So if you're updating a log entry and setting the `maxSectionWidth` to `0`, it will still be evaluated as `undefined` because:

```javascript
maxSectionWidth: 0 || undefined // returns undefined
```

So we'd need something like `maxSectionWidth: updateParams.maxSectionWidth !== undefined ? updateParams.maxSectionWidth : messageState.maxSectionWidth` or similar.

We should also add a test case for this.

solomonope: updated

garden-service/src/plugins/kubernetes/logs.ts
```diff
@@ -90,22 +121,76 @@ async function getLogs({ log, provider, service, stream, tail, follow, pod }: Ge
...
+  json?: boolean
+) {
   let timestamp: Date
 
   proc.stdout!.pipe(split()).on("data", (s) => {
     if (!s) {
       return
     }
-    const [timestampStr, msg] = splitFirst(s, " ")
+    const [timestampStr, msg] = json === true ? parseSternLogMessage(s) : splitFirst(s, " ")
```
eysi09: Small nitpick, but since `json` variable is a boolean or undefined you can also just write `json ? parseSternLogMessage(s) : splitFirst(s, " ")`.

garden-service/src/plugins/kubernetes/logs.ts
```diff
@@ -90,22 +121,76 @@ async function getLogs({ log, provider, service, stream, tail, follow, pod }: Ge
...
+    `--context=${provider.config.context}`,
+    `--namespace=${pod.metadata.namespace}`,
+    `--exclude-container=garden-*`,
+    "--tail",
+    String(tail),
+    "--output=json",
+    "-t",
+  ]
+
+  const labels = getStaticLabelsFromPod(pod)
```
eysi09: Might be useful to add a comment here, explaining the reasoning behind using "static" labels.

garden-service/src/plugins/kubernetes/util.ts
```diff
@@ -381,3 +382,22 @@ export async function getRunningPodInDeployment(deploymentName: string, provider
 
   return sample(pods)
 }
+
+export function getStaticLabelsFromPod(pod: KubernetesPod): { [key: string]: string } {
+  const labels: { [key: string]: string } = {}
+
+  for (let label in pod.metadata.labels) {
```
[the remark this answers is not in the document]
solomonope: done

garden-service/src/plugins/kubernetes/util.ts
```diff
@@ -381,3 +382,22 @@ export async function getRunningPodInDeployment(deploymentName: string, provider
...
+    if (!pod.metadata.labels[label].match(STATIC_LABEL_REGEX)) {
+      labels[label] = pod.metadata.labels[label]
+    }
+  }
+  return labels
+}
+
+export function getSelectorString(labels: { [key: string]: string }) {
+  let selectorString: string = "-l"
+  for (let label in labels) {
```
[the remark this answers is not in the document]
solomonope: done

garden-service/test/unit/src/logger/log-entry.ts
```diff
@@ -98,6 +100,43 @@ describe("LogEntry", () => {
...
           append: undefined,
           timestamp,
+          maxSectionWidth: 8,
+        },
+      ])
+      expect(entry.getMetadata()).to.eql({ task: taskMetadata })
+    })
+    it("should update maxSectionWidth to zero", () => {
+      const timestamp = freezeTime().valueOf()
+      const taskMetadata: TaskMetadata = {
```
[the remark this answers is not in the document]
solomonope: done

garden-service/test/unit/src/plugins/kubernetes/util.ts
```diff
@@ -259,3 +261,41 @@ describe("flattenResources", () => {
     expect(flattenResources(resources).map((r) => r.metadata.name)).to.eql(["a", "d", "e"])
   })
 })
+
+describe("getStaticLabelsFromPod", () => {
+  it("It should should only select labels without characters", () => {
```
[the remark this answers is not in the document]
solomonope: done

garden-service/src/plugins/kubernetes/logs.ts
solomonope: done

garden-service/test/unit/src/plugins/kubernetes/util.ts
```diff
@@ -259,3 +261,41 @@ describe("flattenResources", () => {
...
+
+    expect(labels).to.eql({
+      module: "a",
+      service: "a",
+    })
+  })
+})
+
+describe("getSelectorString", () => {
+  it("It for format a label map to comma separated key value string ", () => {
```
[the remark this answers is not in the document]
solomonope: done

PR #1441, opened 2019-12-17
feat(providers): add conftest providers

edvald:

`conftest` is a handy provider for applying policies on your config
files. You can use the `conftest` provider and module type directly
to manually configure your tests, or you can use the "generator"
providers `conftest-container` and `conftest-kubernetes` to
automatically generate conftest modules. The former creates one test
for each `container` module in your project (very similar to the
`hadolint` provider), and the latter does the same for each `kubernetes`
and `helm` module in your project.

I also added a couple of small commits related to the hadolint provider.


garden-service/src/plugins/conftest/conftest-container.ts
```diff
@@ -0,0 +1,70 @@
...
+            // Make sure we don't step on an existing custom conftest module
+            !existingConftestModuleDockerfiles.includes(dockerfilePath) &&
+            // Only create for modules with Dockerfiles
+            (await containerHelpers.hasDockerfile(module))
+          )
+        }).map((module) => {
+          const baseName = "conftest-" + module.name
+
+          let name = baseName
+          let i = 2
```
eysi09: Why does it start at `2`?

garden-service/src/plugins/conftest/conftest-kubernetes.ts
```diff
@@ -0,0 +1,67 @@
...
+        addModules: await Bluebird.filter(modules, async (module) => {
+          return (
+            // Pick all kubernetes or helm modules
+            module.compatibleTypes.includes("helm") || module.compatibleTypes.includes("kubernetes")
+          )
+        }).map((module) => {
+          const baseName = "conftest-" + module.name
+
+          let name = baseName
+          let i = 2
```
eysi09: Same as above.

garden-service/src/plugins/conftest/conftest-kubernetes.ts
```diff
@@ -0,0 +1,67 @@
...
+          while (allModuleNames.has(name)) {
+            name = `${baseName}-${i++}`
+          }
+
+          allModuleNames.add(name)
+
+          // For helm modules, we only want to look at the rendered YAML output
+          // (which is output when building Helm modules)
+          const isHelmModule = module.compatibleTypes.includes("helm")
+          const files = isHelmModule ? [".rendered.yaml"] : module.include || ["*.yaml", "**/*.yaml"]
```
eysi09: Could you add parenthesis to make to eval order a little clearer?

Also, what if you have `*.yml` endings?

edvald: Prettier rejected my parentheses :/ But yeah good catch, I should add `*.yml`

garden-service/src/plugins/conftest/conftest-container.ts
edvald: By default it's just the `baseName`. We only add the suffix if there is another module with `baseName`, and then the new name is the 2nd one. I think that makes sense.

docs/reference/module-types/conftest.md
```diff
@@ -0,0 +1,396 @@
...
+          # Defaults to to same as source path.
+          target: <same as source path>
+
+# Specify a module whose sources we want to test.
+sourceModule:
+
+# POSIX-style path to a directory containing the policies to match the config against, or a
+# specific .rego
+# file, relative to the module root.
+# Must be a relative path, and should in most cases be within the project root.
```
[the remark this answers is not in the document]
edvald: For this, yes. Because it isn't actually used from the build context, but rather referenced directly in the project source. Otherwise we couldn't viably have a policy directory shared across a project.

garden-service/src/plugins/conftest/conftest-container.ts
eysi09: Ah yeah makes sense. Missed that.

PR #1442, opened 2019-12-17
improvement: set default include on Helm modules

eysi09:

From commit message:

BREAKING CHANGE:

The `include` field on Helm modules now defaults to:

```javascript
["*", "charts/**/*", "templates/**/*"]
```

if not set manually.

Previously, Helm modules would simply include all content under the
module path. If your Helm modules doesn't have `include` set and depends
on content that's not captured with the default include, you will need
to update the relevant `garden.yml` file and set the includes manually.

### DIFF
--- a/garden-service/src/plugins/kubernetes/helm/config.ts
+++ b/garden-service/src/plugins/kubernetes/helm/config.ts
@@ -352,5 +352,10 @@ export async function configureHelmModule({
     "release-name": getReleaseName(moduleConfig),
   }
 
+  // Automatically set the include if not explicitly set
+  if (!moduleConfig.include) {
+    moduleConfig.include = containsSources ? ["*", "charts/**/*", "templates/**/*"] : []
+  }
+
   return { moduleConfig }
 }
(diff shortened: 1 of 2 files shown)


garden-service/src/plugins/kubernetes/helm/config.ts
```diff
@@ -352,5 +352,10 @@ export async function configureHelmModule({
     "release-name": getReleaseName(moduleConfig),
   }
 
+  // Automatically set the include if not explicitly set
+  if (!moduleConfig.include) {
+    moduleConfig.include = ["*", "charts/**/*", "templates/**/*"]
+  }
```
edvald: On reflection, we should probably also set `include: []` if there are no local chart sources.

edvald:
Also, integ tests need to be updated.

garden-service/src/plugins/kubernetes/helm/config.ts
eysi09: Done.

Issue #1250, opened 2019-10-02
Conditional deployment of modules


baspetersdotjpeg:
We currently have no straightforward way to conditionally deploy modules. This could be useful in cases where a user, for example, would want to deploy a database in their dev environment but use a hosted service when deploying to production.

We only support `skipDeploy` for Helmcharts and openfaas functions. We have the`modules.exclude` and `modules.include` fields available at the project level, but these are limited in functionality because on invocation, Garden has no awareness of, for example, what environment it's deploying to.

Some ideas:
 - have `modules.include|exclude` in the environment config?
 - make `skipDeploy` part of the base module config?
 - all of the above?

eysi09:
Closed with #1515.

PR #1443, opened 2019-12-17
Allow excluding services from a deployment

GavinRay97:
This is a rough draft but appears to work from console-logging the output of `services` after the exclusion filter.

https://github.com/garden-io/garden/issues/1250

### DIFF
--- a/garden-service/src/commands/deploy.ts
+++ b/garden-service/src/commands/deploy.ts
@@ -51,6 +51,11 @@ const deployOpts = {
     `,
     alias: "hot",
   }),
+  "exclude": new StringsParameter({
+    help: deline`The name(s) of the service(s) to exclude from the deployment.
+      Use comma as a separator to specify multiple services.
+    `,
+  }),
 }
 
 type Args = typeof deployArgs
@@ -73,6 +78,7 @@ export class DeployCommand extends Command<Args, Opts> {
         garden deploy                      # deploy all modules in the project
         garden deploy my-service           # only deploy my-service
         garden deploy service-a,service-b  # only deploy service-a and service-b
+        garden deploy --exclude svc1,svc2  # exclude certain services from deployment
         garden deploy --force              # force re-deploy of modules, even if they're already deployed
         garden deploy --watch              # watch for changes to code
         garden deploy --hot=my-service     # deploys all services, with hot reloading enabled for my-service
@@ -103,11 +109,16 @@ export class DeployCommand extends Command<Args, Opts> {
     }
 
     const initGraph = await garden.getConfigGraph(log)
-    const services = await initGraph.getServices(args.services)
+    let services = await initGraph.getServices(args.services)
 
     if (services.length === 0) {
       log.error({ msg: "No services found. Aborting." })
       return { result: {} }
+    }
+
+    // If exclude is specified, omit services whose names match
+    if (opts.exclude) {
+      services = services.filter(service => !opts.exclude!.includes(service.name))
     }
 
     const hotReloadServiceNames = await getHotReloadServiceNames(opts["hot-reload"], initGraph)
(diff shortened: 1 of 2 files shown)


GavinRay97:
Whoops, failing on lint :grimacing:

eysi09:
Thanks for the contribution @GavinRay97!

Sorry for the late reply, the Helm 3 migration has consumed most of our attention for the past week or so.

The PR looks great but it would be good to add unit tests for the `exclude` flag here: https://github.com/garden-io/garden/blob/master/garden-service/test/unit/src/commands/deploy.ts

And the lint is failing because the auto-generated docs are out of date. We auto-generate docs from the command help texts (among other things) and you need to run `npm run generate-docs` from the repo root to update them.

I'd also be happy to take over from here if you'd like.

garden-service/src/commands/deploy.ts
```diff
@@ -103,13 +109,18 @@ export class DeployCommand extends Command<Args, Opts> {
...
+    let services = await initGraph.getServices(args.services)
 
     if (services.length === 0) {
       log.error({ msg: "No services found. Aborting." })
       return { result: {} }
     }
 
+    // If exclude is specified, omit services whose names match
+    if (opts.exclude) {
+      services = services.filter(service => !opts.exclude!.includes(service.name))
```
eysi09: Not quite sure why you need the `!` after `opts.exclude` since you've already checked that it is defined. Also, the linter will likely complain about the missing parenthesis around the `service` arg of the `filter` callback. We use Prettier for code formatting and you can configure it to fix these errors on save. Alternatively, running `npm run fix-format` from root should do the trick.

GavinRay97: Yeah I am unsure either. TS complains that Object is possibly undefined, even if I EXPLICITLY set the `if (options.exclude)` check to be for `!= undefined`:

image

It seems due to the type signature for `StringsParameter`:

image

GavinRay97: Also, this is a *VERY* large and complex project, is this commit enough to properly integrate this exclusion option with everything?

It appears to work, but I am unsure if this is ***all there is***, so to speak -- you know?

GavinRay97:
Fixing that lint error, I saw more in my unstaged changes and realized O forgot to commit the documentation I added for this command :woman_facepalming:.

Going to toss that in there too.

GavinRay97:
Git hook fails with:
```
> garden@ check-docs /home/user/Projects/garden
> ./bin/check-docs

./bin/check-docs: line 36: parallel: command not found
```

eysi09:
Hi again @GavinRay97 

Really sorry for how long this has dragged out. We were kind of bogged down by v0.11 release and the Helm migration business, and then we closed shop for the holidays.

But this is a useful feature so we very much want to get this merged. And to address your questions above:

* Parallel is a dev dependency, we use it to speed up the links checker.
* The changes you made effect our auto-generated docs, so you'd need to run `npm run generate-docs` from the project root to pass CI.

It would also be great to add tests for this, just to verify that the services get excluded. It would basically duplicate the logic of this test: https://github.com/garden-io/garden/blob/master/garden-service/test/unit/src/commands/deploy.ts#L176, except you would call the command with different options. Something like:

```javascript
    const { result, errors } = await command.action({
      garden,
      log,
      headerLog: log,
      footerLog: log,
      args: {},
      opts: withDefaultGlobalOpts({
        "hot-reload": undefined,
        "watch": false,
        "force": false,
        "force-build": true,
        "exclude": "service-a" // <--- Test the exclude option.
      }),
    })
```

I could also take over from here if you want.

garden-service/src/commands/deploy.ts
eysi09: Yeah, in this case it basically is, except it would be great to add tests as well, see my comment below.

edvald:
#1515 relates to this. That deals with disabling modules/services/etc. at the config level though, not at the parameter level. It would imo also be fine to include the CLI parameter, but probably best to merge the other one first and then use the mechanism offered there as a basis.

eysi09:
This probably needs a little more consideration and is not this straightforward to implement. If we add an `exclude` option, we should also add it to other commands for consistency. I suggest as we close this for now since we already have the `disabled` field at the config level.

PR #1445, opened 2019-12-18
fix(k8s): make sure Helm 3 migration is run for project namespaces

edvald:

The Helm 3 migration wasn't being triggered for project namespaces.


baspetersdotjpeg:
This fixes the reported issue 👍

PR #1457, opened 2019-12-20
fix: failing init if remote is not set

10ko:
The `AnalyticsHandler` was wrongly expecting the project to have a `remote.url` set and it was failing when running garden on a local repo. Fixed that.

Fixes #

cc @thsig

### DIFF
--- a/garden-service/src/vcs/git.ts
+++ b/garden-service/src/vcs/git.ts
@@ -389,6 +389,11 @@ export class GitHandler extends VcsHandler {
   async getOriginName(log: LogEntry) {
     const cwd = process.cwd()
     const git = this.gitCli(log, cwd)
-    return (await git("config", "--get", "remote.origin.url"))[0]
+    try {
+      return (await git("config", "--get", "remote.origin.url"))[0]
+    } catch (error) {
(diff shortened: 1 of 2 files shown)


garden-service/src/vcs/git.ts
```diff
@@ -389,6 +389,11 @@ export class GitHandler extends VcsHandler {
   async getOriginName(log: LogEntry) {
     const cwd = process.cwd()
     const git = this.gitCli(log, cwd)
-    return (await git("config", "--get", "remote.origin.url"))[0]
+    try {
+      return (await git("config", "--get", "remote.origin.url"))[0]
+    } catch (error) {
+      log.silly(`Trying to retrieve origin url but encountered an error: ${error}`)
```
eysi09: Could you expand a little on the error. E.g. say that it was trying to get _git remote origin URL_ or similar.

garden-service/src/vcs/git.ts
```diff
@@ -389,6 +389,11 @@ export class GitHandler extends VcsHandler {
...
+    }
+    return ""
```
eysi09: Would it make sense to have it rather return `null` or `undefined` if getting the URL fails?

10ko:
👍

PR #1461, opened 2019-12-20
fix: remove resources only when tiller not deployed

10ko:
Fix the unhandled exception thrown by Garden when running the helm migration against a namespace where either:
- tiller has been deployed in the past but is not present anymore.
- Or it's deployed in the garden-system but not in a new project namespace.

Fixes #


edvald:
Doesn't matter here, but a `fix(k8s)` category would make sense for this one

PR #1473, opened 2020-01-02
refactor: update Joi library

edvald:

This turned out to be quite involved, but it's good to just get it over
with before moving on to any further improvements to our validation.

I had to fork and use the latest master (the fork being to make sure
we have a consistent version) so that we get a recent fix in there. We
should get rid of that once the next release (after 16.1.8) is out.

### DIFF
--- a/garden-service/src/config/validation.ts
+++ b/garden-service/src/config/validation.ts
@@ -14,17 +14,27 @@ import uuid from "uuid"
 
 const joiPathPlaceholder = uuid.v4()
 const joiPathPlaceholderRegex = new RegExp(joiPathPlaceholder, "g")
-const joiOptions = {
-  abortEarly: false,
-  language: {
-    key: `key ${joiPathPlaceholder} `,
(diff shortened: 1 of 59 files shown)


garden-service/src/config/validation.ts
```diff
@@ -104,6 +114,7 @@ export function validate<T>(
       }
 
       // a little hack to always use full key paths instead of just the label
+      e.message = e.message.replace(joiLabelPlaceholderRegex, "key " + chalk.underline(renderedPath || "."))
       e.message = e.message.replace(joiPathPlaceholderRegex, chalk.underline(renderedPath || "."))
```
thsig: This line is a duplicate/no-op replacement due to the line above.

edvald: That's not true. The regexes being replaced are different.

thsig: Ah, I misread that there. Cool.

PR #1474, opened 2020-01-03
chore: Create garden / Azure cli tools image

davestephens:
As requested in the docs, I've added a tools image for my cloud provider (Azure)

None, AFAIK!


garden-service/azure.Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+ARG TAG
+FROM gardendev/garden:${TAG}
+
+RUN apk add bash py-pip \
+  && apk add --virtual=build gcc libffi-dev musl-dev openssl-dev python-dev make \
+  && pip --no-cache-dir install -U pip \
+  && pip --no-cache-dir install azure-cli \
+  && apk del --purge build
+
+COPY --from=lachlanevenson/k8s-kubectl:latest /usr/local/bin/kubectl /usr/local/bin/kubectl
```
edvald: Could we pin the k8s-kubectl version to a specific one (e.g. v1.17.0)?

(sorry about the late review btw!)

davestephens: Yeah, for sure. I didn't bother as your gke example don't have a pinned version either :-)

Is there anywhere in the garden repo where tool versions are defined that I can use?

edvald: Yeah we should actually fix that, and download a specific gcloud CLI archive instead of using that installer script.

There is a specific kubectl version used by Garden itself, which is in fact automatically downloaded when needed, but we don't export the version anywhere. We have talked about some improvements to the tool management we do internally, and making it more explicit in the framework (incl. having a `garden tools` command, with subcommands like `garden tools kubectl` etc.). Until we do that we can just go for the latest available version.

davestephens: OK, so I'm clear - do you want it hard-coded to the latest version now (which feels like it'd be prime for being forgotten about), or are you ok with how I've done it? If it's the latter, we could add a TODO pending your comments above about tooling?

edvald: On reflection I guess it comes down to the same, since we'll replace this anyway before long. Fine to keep it as-is.

edvald:
As it happens, we just made some changes to that build script that conflict, as we moved to Node 12.x. Should be a quick fix though.

edvald:
Hey @davestephens, sorry this one has dragged out so much, we just happened to be updating our build flows around the time you posted this, to update to Node 12.x. Should I just copy your changes to a new branch that fits with our new build flow?

eysi09:
@edvald, would you mind resolving the conflicts on this one?

edvald:
Closing in favor of #1893

PR #1480, opened 2020-01-06
fix(cli): do not init Garden class for commands with no project

eysi09:

Now we don't initialise the Garden class when running commands with `noProject = true`. We need this because `noProject` commands can fail if the project contains template strings (and presumably for other reasons as well as Garden, e.g. as Garden attempts to resolve providers and configs). 

No issue but I noticed this when testing the `migrate` command.

### DIFF
--- a/garden-service/src/commands/base.ts
+++ b/garden-service/src/commands/base.ts
@@ -248,7 +248,7 @@ export interface PrepareParams<T extends Parameters = {}, U extends Parameters =
 }
 
 export interface CommandParams<T extends Parameters = {}, U extends Parameters = {}> extends PrepareParams<T, U> {
-  garden: Garden
+  garden?: Garden
 }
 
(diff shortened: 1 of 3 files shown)


garden-service/src/commands/base.ts
```diff
@@ -248,7 +248,7 @@ export interface PrepareParams<T extends Parameters = {}, U extends Parameters =
 }
 
 export interface CommandParams<T extends Parameters = {}, U extends Parameters = {}> extends PrepareParams<T, U> {
-  garden: Garden
+  garden?: Garden
```
edvald: Doesn't this cause a number of type errors?

eysi09: It does. In fact, I need to refactor this. I'll close this one and submit a new PR.

PR #1483, opened 2020-01-06
chore(ci): get minikube tests working for external contributors

edvald:

We were authenticating with gcloud for just one specific integ test in the minikube tests, and that was breaking test execution for external contributors.


.circleci/config.yml
```diff
@@ -85,17 +85,22 @@ commands:
           paths: [dashboard/node_modules]
           key: dashboard-v4-{{ checksum "dashboard/package-lock.json" }}
 
-  configure_kubectl_context:
+  install_kubectl:
     description: Configure the kubectl context so that we can access our remote cluster. Used for e2e testing.
```
eysi09: Can you update the description.

garden-service/test/integ/src/plugins/kubernetes/container/build.ts
```diff
@@ -55,7 +82,7 @@ describe("k8sBuildContainer", () => {
       await init("local")
     })
 
-    it("should build a simple container", async () => {
+    it("should build a simple container (local only)", async () => {
```
eysi09: Is there a better to specify whether a test is local/remote than just grepping the description?

edvald: Not that I could find, no.

.circleci/config.yml
edvald: Done

PR #1484, opened 2020-01-06
fix(cli): ensure 'noProject' commands run in invalid projects

eysi09:

Adds a `DummyGarden` class to use for commands that run outside the context of a valid Garden project.

Fixes an issue where commands with `noProject=true` fail if the project contains template strings (and presumably for other reasons as well as Garden attempts to resolve providers and configs).

### DIFF
--- a/garden-service/src/cli/cli.ts
+++ b/garden-service/src/cli/cli.ts
@@ -67,21 +67,33 @@ const OUTPUT_RENDERERS = {
 const GLOBAL_OPTIONS_GROUP_NAME = "Global options"
 const DEFAULT_CLI_LOGGER_TYPE = "fancy"
 
-// For initializing garden without a project config
-export const MOCK_CONFIG: ProjectConfig = {
-  path: process.cwd(),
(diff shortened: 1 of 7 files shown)


garden-service/src/cli/cli.ts
```diff
@@ -67,21 +67,37 @@ const OUTPUT_RENDERERS = {
...
+    path: root,
+    apiVersion: DEFAULT_API_VERSION,
+    kind: "Project",
+    name: "no-project",
+    defaultEnvironment: "",
+    dotIgnoreFiles: defaultDotIgnoreFiles,
+    environments: defaultEnvironments,
+    providers: [
+      {
+        name: "local-kubernetes",
```
edvald: Is this relevant/necessary?

eysi09: Good catch. I've updated the PR.

eysi09:
Also re-named the project metadata fields in `analytics.ts` like we talked about (cc @10ko).

Issue #1472, opened 2019-12-30
cleanup-cluster-registry fails with 404 at flagging images as deleted


DanielKillenberger:
### Current Behavior

`garden plugins kubernetes cleanup-cluster-registry` often fails with the following message:

```
Flagging 11 unused images as deleted in cluster registry...
Request failed with status code 404
```

The flagging seems to take place but it doesn't get to garbage collection. When creating a new registry pod and running the command it will now flag 0 images as deleted and run the garbage collection.

### Expected behavior
Images should get flagged and it should continue with the garbage collection.

### Reproducible example
Run:
```garden plugins kubernetes cleanup-cluster-registry```
Expect it to randomly fail.

### Workaround
Scale the garden-docker-registry down to 0 and back up to 1 and rerun the command.

### Your environment

Ubuntu 19.10. Cluster running on google cloud with in-cluster docker building.

`garden version`
0.10.16 - (0.11.0  edge release)
`kubectl version`
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2019-12-07T21:20:10Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15+", GitVersion:"v1.15.4-gke.22", GitCommit:"a6ba43f5a24ac29e631bb627c9b2a719c4e93638", GitTreeState:"clean", BuildDate:"2019-11-26T00:40:25Z", GoVersion:"go1.12.11b4", Compiler:"gc", Platform:"linux/amd64"}
`docker version`
18.06.1-ce

eysi09:
We made some fixes to the cleanup command for our next release but I notice you're on `edge` so presumably this is still an issue.

edvald:
It seems we're missing some exception handling there, we shouldn't be failing on 404s from the registry. I can take a look.

PR #1485, opened 2020-01-06
fix(k8s): handle 404 exception when tagging image for deletion

solomonope:
This PR allows a 404 error when tagging images in the registry for deletion fail silently and log a message

#1472
Fixes #


garden-service/src/plugins/kubernetes/commands/cleanup-cluster-registry.ts
```diff
@@ -169,20 +169,24 @@ async function deleteImagesFromRegistry(ctx: KubernetesPluginContext, log: LogEn
...
-    // Issue the delete request
-    await queryRegistry(ctx, log, `${name}/manifests/${digest}`, {
-      method: "DELETE",
-    })
+      // Issue the delete request
+      await queryRegistry(ctx, log, `${name}/manifests/${digest}`, {
+        method: "DELETE",
+      })
+    } catch (err) {
+      log.debug(`${err.message} happend when deleting ${image}`)
```
edvald: I think we need to more specifically catch the error. If you just wrap the first query and explicitly catch a 404, you can just return without logging any error. Other errors should be thrown as usual.

solomonope: I have made the change, I have updated to only check for 404

garden-service/src/plugins/kubernetes/commands/cleanup-cluster-registry.ts
```diff
@@ -169,20 +169,27 @@ async function deleteImagesFromRegistry(ctx: KubernetesPluginContext, log: LogEn
...
-    // Issue the delete request
-    await queryRegistry(ctx, log, `${name}/manifests/${digest}`, {
-      method: "DELETE",
-    })
+      // Issue the delete request
+      await queryRegistry(ctx, log, `${name}/manifests/${digest}`, {
+        method: "DELETE",
+      })
+    } catch (err) {
+      if (err.response && !(err.response.status === 404)) {
```
edvald: Minor nitpick, but `err.response.status !== 404` is a bit cleaner :)

garden-service/src/plugins/kubernetes/commands/cleanup-cluster-registry.ts
edvald: That works. I'd just remove that log line, don't think it's necessary. Then it's good to go!

solomonope: fixed

garden-service/src/plugins/kubernetes/commands/cleanup-cluster-registry.ts
solomonope: fixed

PR #1495, opened 2020-01-10
improvement: allow to specify release to install

khaled:
This will allow users to do things like:

```bash
curl -sL https://get.garden.io/install.sh | bash -s edge
```

or 

```bash
curl -sL https://get.garden.io/install.sh | bash -s v0.11.0-alpha.4
```

Leaving out the parameter `-s <release>` passed to `bash` will install the latest released version.

### DIFF
--- a/support/install.sh
+++ b/support/install.sh
@@ -15,17 +15,20 @@ echo "[0;40;36m[48;2;35;163;116m  [48;2;30m[48;2;67;136;127m  [36m[48;2;92
 echo "❊ Installing the Garden CLI ❊"
 echo ""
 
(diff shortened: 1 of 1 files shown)


edvald:
Thanks @khaled! The script change looks good, but the curl command specified doesn't work, because the argument is passed to `curl` and not `bash`. I'm not quite sure how to fix that?

khaled:
> Thanks @khaled! The script change looks good, but the curl command specified doesn't work, because the argument is passed to `curl` and not `bash`. I'm not quite sure how to fix that?

Silly me! I corrected the example in the comment.

PR #1504, opened 2020-01-14
fix(helm): only hot reload serviceResource

thsig:

Before this fix, the sync volume would be attached to all containers in the target resource when using hot reloading with Helm modules, which is not desirable e.g. when using user-specified sidecar containers.

Now, the sync volume is only attached to the container identified by `serviceResource.containerName` or the Helm module's name.

We (I) didn't create a GitHub issue for this one, but it was raised by a user in our community Slack channel.

Note that we're now also only attaching the sync volume to the "main" container for other Kubernetes module types (e.g. `container` and `kubernetes`).

### DIFF
--- a/garden-service/src/plugins/kubernetes/helm/common.ts
+++ b/garden-service/src/plugins/kubernetes/helm/common.ts
@@ -183,7 +183,7 @@ export function getReleaseName(config: HelmModuleConfig) {
  *
  * Throws error if no resource spec is configured, or it is empty.
  */
-export function getServiceResourceSpec(module: HelmModule) {
+export function getServiceResourceSpec(module: HelmModule): HelmResourceSpec {
   const baseModule = getBaseModule(module)
   let resourceSpec = module.spec.serviceResource || {}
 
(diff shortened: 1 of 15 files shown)


garden-service/src/plugins/kubernetes/helm/common.ts
```diff
@@ -183,7 +183,7 @@ export function getReleaseName(config: HelmModuleConfig) {
  *
  * Throws error if no resource spec is configured, or it is empty.
  */
-export function getServiceResourceSpec(module: HelmModule) {
+export function getServiceResourceSpec(module: HelmModule): HelmResourceSpec {
```
edvald: That's not really necessary, just as an FYI. The type is inferred.

garden-service/src/plugins/kubernetes/hot-reload.ts
```diff
@@ -58,14 +58,10 @@ export function configureHotReload({
   hotReloadCommand,
   hotReloadArgs,
   containerName,
-}: ConfigureHotReloadParams) {
+}: ConfigureHotReloadParams): void {
```
edvald: Same here. Is this a matter of taste, or did you have a reason to explicitly declare the return type?

thsig: No, just added the types while I was passing by. What do you think, should I remove them?

edvald: No it's all good, just wanted to point this out. It doesn't hurt anything, just redundant.

PR #1506, opened 2020-01-14
fix(analytics): Analytics race condition + Telemetry page

### LINKED ISSUE #1489: Nice to have: Show example of telemetry sent
Not so much a bug, but more a nice to have. 

I'd feel more comfortable having analytics/telemetry enabled if an example of what is sent was shown in the readme. I could go digging through source and work it out but would save time if it was provided.

I'd like to help you guys out with this (as I'm sure many others would) but need to be confident I'm only sending data I'm comfortable with.

10ko:
This PR introduces two changes:
- Detail page about telemetry with samples of data we collect.
- Workaround for a race condition in the segment client which would cause the first event to potentially never been sent to Segment. See [here](https://github.com/segmentio/analytics-node/issues/219).

Fixes #1489


docs/telemetry.md
```diff
@@ -0,0 +1,136 @@
+# Telemetry
+
+Whenever you run a Garden command we collect anonymized telemetry and send it to [Segment.io](https://segment.io).
+It helps us understand how our users use garden and aids our decision process when prioritizing new features, bug fixing, etc.
```
edvald: garden -> Garden

docs/telemetry.md
```diff
@@ -0,0 +1,136 @@
...
+
+We put great effort in making sure we only collect completely anonymized data: we use random generated uuids for identifying users and we hash sensitive information using SHA-256.
```
edvald: : we -> : We
uuids -> UUIDs
users -> users and projects

docs/telemetry.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Types of events
+
+Below you can find examples for each one of the events we currently collect.
```
edvald: Maybe safer to just say "examples of events we collect", unless we commit to always update this when we add new events or properties. If we go for the latter I'd suggest automating this, similar to how we generate reference docs, but imo that would be good for a later iteration.

docs/telemetry.md
```diff
@@ -0,0 +1,136 @@
...
+
+**Note:** events are batched before being sent to our data processor. This means you shouldn't worry about Garden making dozens of HTTP requests and affecting commands performance.
```
edvald: I'm not sure this is necessary to mention. Seems like a table-stakes thing to do in our implementation :)

docs/telemetry.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Identify
+
+Collected the first time a Garden command is run on a machine.
```
edvald: Collected -> Sent (easier to read that way)

docs/telemetry.md
```diff
@@ -0,0 +1,136 @@
...
+    system: {
+      gardenVersion: '0.10.16',
+      platform: 'darwin',
+      platformVersion: '18.7.0'
+    }
+  }
+}
+```
+
+## Update your telemetry preferences
```
edvald: Update -> Updating

docs/telemetry.md
```diff
@@ -0,0 +1,136 @@
...
+
+If you would like to update your analytics settings, please run (from within a project):
```
edvald: Wasn't the "from within a project" requirement removed recently?

eysi09: 'twas

garden-service/src/analytics/analytics.ts
```diff
@@ -319,15 +320,20 @@ export class AnalyticsHandler {
...
+            if (err && this.log) {
+              this.log.debug(`Error sending tracking event: ${err}`)
+            }
+            resolve(true)
+          })
         })
       }
 
-      return trackToRemote(segmentEvent)
+      return await trackToRemote(segmentEvent)
```
eysi09: Should we `debug` (or maybe `silly`) log the data we're sending?

edvald: A silly log would make sense, yes.

Also, `return await` is in most cases redundant. Just noticing that now. No harm, but it means creating an additional Promise that is resolved at the same time as the other one.

docs/telemetry.md
10ko: We hash the projectId so technically we don't use UUIDs to identify projects.

garden-service/src/analytics/analytics.ts
10ko: Will fix and add `silly` log

10ko:
Fixed.

edvald:
There are still a couple of unresolved comments, otherwise looks good.

eysi09:
> There are still a couple of unresolved comments, otherwise looks good.

I'll go ahead and approve this then since those comments have already been resolved. So that we can into this release.

PR #1519, opened 2020-01-16
improvement(core): detect rsync and ensure the version is recent enough

### LINKED ISSUE #1493: Fail nicely (and up front) if rsync isn't available
### Current Behavior

If rsync is not available garden fails with a not-very-pretty error. I appreciate that "you must have rsync is installed" is in the docs but garden could fail faster than it does :-)

```
[2020-01-09T14:53:09.481Z] Error: kubernetes — an error occurred when configuring environment:
Error: Command "rsync --recursive --times --owner --group --perms --links --safe-links --ignore-missing-args --temp-dir /opt/buildagent/work/9a1291be564fcd5c/.garden/kubernetes.garden/build/nfs-provisioner.tmp --prune-empty-dirs --include-from=- --exclude=* --delete-excluded /root/.garden/bin/static/kubernetes/system/nfs-provisioner/ /opt/buildagent/work/9a1291be564fcd5c/.garden/kubernetes.garden/build/nfs-provisioner" failed with code undefined:
    at Object.<anonymous> (/snapshot/dist/build/src/plugins/kubernetes/system.js:0)
    at fulfilled (/snapshot/dist/build/src/plugins/kubernetes/system.js:0)
```

### Expected behavior

Garden should fail at the init stage if rsync is not available. 

### Reproducible example

Run `garden plugins kubernetes cluster-init --logger-type basic` from a box that doesn't have rsync.

### Workaround

Install rsync ;-)

### Suggested solution(s)

Check up front if rsync is available, and fail at the start with a pretty message if its not.

### Your environment

`garden version`: 0.11.0-alpha.3

edvald:

What it says on the tin ^ 

Closes #1493


garden-service/src/build-dir.ts
```diff
@@ -7,34 +7,87 @@
...
+      )
+    }
+
+    if (!version) {
+      throw versionDetectFailure
+    }
+
+    let versionGte = true
+
+    try {
```
eysi09: Just out of curiosity, when does this throw?

eysi09: Is the `try/catch` necessary?

edvald: Yeah, semver will throw an error if the version string is invalid, for some reason. One of the added tests is for that case.

PR #1520, opened 2020-01-16
fix(k8s): fix failing tasks not throwing errors

thsig:

Before this fix, failing tasks (e.g. with an invalid command or args configured) would succeed.

Also added integration tests for the failing case for tests and tasks for our k8s plugins, and fixed/improved logging of test/task logs and errors.

### DIFF
--- a/garden-service/src/plugins/container/build.ts
+++ b/garden-service/src/plugins/container/build.ts
@@ -64,7 +64,7 @@ export async function buildContainerModule({ module, log }: BuildModuleParams<Co
   }
 
   // Stream log to a status line
-  const outputStream = createOutputStream(log.placeholder(LogLevel.debug))
+  const outputStream = createOutputStream(log.placeholder(LogLevel.info))
   const timeout = module.spec.build.timeout
   const buildLog = await containerHelpers.dockerCli(module, [...cmdOpts, buildPath], { outputStream, timeout })
 
--- a/garden-service/src/plugins/kubernetes/container/run.ts
+++ b/garden-service/src/plugins/kubernetes/container/run.ts
@@ -71,15 +71,15 @@ export async function runContainerTask(params: RunTaskParams<ContainerModule>): 
     ignoreError: true, // to ensure results get stored when an error occurs
   })
 
-  const result = {
+  const result: RunTaskResult = {
     ...res,
     taskName: task.name,
     outputs: {
-      log: res.output || "",
+      log: res.log || "",
     },
   }
 
-  await storeTaskResult({
+  return storeTaskResult({
     ctx,
     log,
     module,
@@ -87,6 +87,4 @@ export async function runContainerTask(params: RunTaskParams<ContainerModule>): 
     taskVersion,
     taskName: task.name,
   })
-
-  return result
 }
--- a/garden-service/src/tasks/task.ts
+++ b/garden-service/src/tasks/task.ts
@@ -30,6 +30,12 @@ export interface TaskTaskParams {
   forceBuild: boolean
 }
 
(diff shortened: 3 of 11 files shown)


garden-service/src/plugins/kubernetes/container/run.ts
```diff
@@ -69,24 +73,23 @@ export async function runContainerTask(params: RunTaskParams<ContainerModule>):
...
+    stdout: outputStream,
   })
 
-  const result = {
+  const result: RunTaskResult = {
     ...res,
     taskName: task.name,
     outputs: {
-      log: res.output || "",
+      log: res.log || "",
```
eysi09: Can we also test for this? That is, that the log contains some expected entry. Just to ensure that the output actually gets rendered to the terminal. This broke because we failed to update it when we changed the shape of the `res` object. Feels like it could happen again.

thsig: We are testing this, at least to some extent - see for example:

https://github.com/garden-io/garden/blob/5551b351a46ade5dcc607ed2a166cd75d15cac8a/garden-service/test/integ/src/plugins/kubernetes/container/container.ts#L53-L69

garden-service/src/plugins/kubernetes/container/run.ts
```diff
@@ -58,6 +60,8 @@ export async function runContainerTask(params: RunTaskParams<ContainerModule>):
 
   const image = await containerHelpers.getDeploymentImageId(module, ctx.provider.config.deploymentRegistry)
 
+  const outputStream = createOutputStream(log.placeholder(LogLevel.info))
```
edvald: Do we actually want to print the log output like this?

garden-service/src/plugins/container/build.ts
```diff
@@ -64,7 +64,7 @@ export async function buildContainerModule({ module, log }: BuildModuleParams<Co
   }
 
   // Stream log to a status line
-  const outputStream = createOutputStream(log.placeholder(LogLevel.debug))
+  const outputStream = createOutputStream(log.placeholder(LogLevel.info))
```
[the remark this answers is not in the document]
thsig: Yeah, I brought this in from a branch Eysi had been working on. Maybe @eysi09 could chime in here?

garden-service/src/plugins/kubernetes/container/run.ts
thsig: Same here, @eysi09.

garden-service/src/tasks/task.ts
```diff
@@ -30,6 +30,12 @@ export interface TaskTaskParams {
   forceBuild: boolean
 }
 
+class RunTaskError extends Error {
+  toString() {
+    return this.message
+  }
```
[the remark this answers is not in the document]
thsig: Maybe this new error class is unnecessary - I added it to correspond with the `TestError` class in `TestTask`.

https://github.com/garden-io/garden/blob/fix-task-errors/garden-service/src/tasks/test.ts#L29

garden-service/src/plugins/container/build.ts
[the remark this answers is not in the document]
edvald: Could we make that an explicit commit then? Feels unrelated to the main commit in this PR.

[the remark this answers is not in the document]
thsig: Yeah, makes sense. Will do.

garden-service/src/plugins/kubernetes/container/run.ts
thsig: I'm removing this one.

PR #1521, opened 2020-01-17
refactor(logger): Add dataFormat to LogEntry

10ko:
This PR introduces the ability of passying a `dataFormat` to the logging methods like:
```js
const someData: {
  key: "value",
}
log.info({ data: someData, dataFormat: "json" })
```
This way we can choose if we want to render the data as `json` or `yaml` (the current default).

The second commit removes some js files I found while writing the tests and adds a new tslint rule [tslint-no-focused-test](https://github.com/jasonmendes/tslint-no-focused-test) which fails if you forget `only` in your tests. :)


garden-service/test/unit/src/logger/log-entry.ts
```diff
@@ -66,6 +66,7 @@ describe("LogEntry", () => {
       symbol: undefined,
       status: undefined,
       data: undefined,
+      dataFormat: undefined,
```
eysi09: Could you add a value for `dataFormat` in the `"should update entry state"` test to check that it gets set. That is, set `dataFormat: "json"` in the `setState` call and check that the expected object also has `dataFormat: "json"`.

10ko: Makes sense. Done.

10ko:
Can we merge this for 0.11.1?

10ko:
Can we merge this for 0.11.1?

PR #1526, opened 2020-01-20
fix(test): improve error logging in e2e tests

thsig:


edvald:
@thsig looks good, but linting is failing

PR #1530, opened 2020-01-21
feat(k8s): add test and task support for kubernetes module type

### LINKED ISSUE #1529: [FEATURE]: Allow to specify tests and tasks in kubernetes modules
### Background / Motivation

Few users have been requesting this on our community slack an it should be addressed.

Define `tests` and `tasks` when using a `kubernetes` module the same way as it's currently possible for the `container` [module](https://docs.garden.io/module-types/container#tests)

Some users would like to use the `kubernetes` module to allow for more customization (as opposed to using for example the `container` module) but in doing so they would still need to run tests and tasks.

- [ ] Crucial (Garden is unusable for us without it)
- [x] High (Not having this feature makes using Garden painful)
- [ ] Medium (Using Garden without it is mildly inconvenient)
- [ ] Low (Nice to have someday)

edvald:

I basically re-used the same functionality from the `helm` module type.
Nice and simple.

Note: We may want to revisit that schema at some point, but I figured
the sane approach would be to keep it consistent for now.

Closes #1529


garden-service/src/plugins/kubernetes/config.ts
```diff
@@ -548,3 +559,93 @@ export const configSchema = kubernetesConfigBase
...
+
+export const serviceResourceSchema = joi.object().keys({
+  // TODO: consider allowing a `resource` field, that includes the kind and name (e.g. Deployment/my-deployment).
+  kind: joi
+    .string()
+    .valid(...hotReloadableKinds)
+    .default("Deployment")
+    .description("The type of Kubernetes resource to sync files to."),
+  name: joi.string().description(
+    deline`The name of the resource to sync to. If the chart contains a single resource of the specified Kind,
```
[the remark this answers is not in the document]
eysi09: Actually, just noticed that we're using the same schema for helm and k8s modules. But I guess manifest better applies to both, right? Or some 3rd more generic option.

garden-service/src/plugins/kubernetes/config.ts
```diff
@@ -548,3 +559,93 @@ export const configSchema = kubernetesConfigBase
...
+    ),
+    command: joi
+      .array()
+      .items(joi.string())
+      .description("The command/entrypoint used to run the task inside the container.")
+      .example(commandExample),
+    args: joi
+      .array()
+      .items(joi.string())
+      .description("The arguments to pass to the pod used for execution.")
```
eysi09: Shouldn't this be _container_ instead of _pod_?

garden-service/src/plugins/kubernetes/config.ts
```diff
@@ -548,3 +559,93 @@ export const configSchema = kubernetesConfigBase
...
+    ),
+    command: joi
+      .array()
+      .items(joi.string())
+      .description("The command/entrypoint used to run the test inside the container.")
+      .example(commandExample),
+    args: joi
+      .array()
+      .items(joi.string())
+      .description("The arguments to pass to the pod used for testing.")
```
eysi09: Same as above s/pod/container.

garden-service/src/plugins/kubernetes/kubernetes-module/run.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  const res = await runAndCopy({
+    ...params,
+    container,
+    command,
+    args,
+    artifacts: task.spec.artifacts,
+    envVars: task.spec.env,
+    image: container.image,
+    podName: `task-${module.name}-${task.name}-${Math.round(new Date().getTime())}`,
```
eysi09: Might be useful to have a function the creates the name, we use this same name schema in six places in the codebase for test and task pods.

garden-service/src/plugins/kubernetes/kubernetes-module/run.ts
```diff
@@ -0,0 +1,76 @@
...
+    description: `Task '${task.name}' in container module '${module.name}'`,
+    timeout,
+    ignoreError: true, // to ensure results get stored when an error occurs
+  })
+
+  const result = {
+    ...res,
+    taskName: task.name,
+    outputs: {
+      log: res.output || "",
```
eysi09: Should this be `res.log` , as per @thsig fixes in #1520 (garden-service/src/plugins/kubernetes/container/run.ts:78).

garden-service/src/plugins/kubernetes/util.ts
```diff
@@ -401,3 +407,135 @@ export function getSelectorString(labels: { [key: string]: string }) {
...
+): ServiceResourceSpec {
+  let resourceSpec = module.spec.serviceResource || {}
+
+  if (baseModule) {
+    resourceSpec = jsonMerge(baseModule.spec.serviceResource || {}, resourceSpec)
+  }
+
+  if (isEmpty(resourceSpec)) {
+    throw new ConfigurationError(
+      deline`Helm module '${module.name}' doesn't specify a \`serviceResource\` in its configuration.
```
eysi09: This is now also used by the k8s module so we should remove the reference to Helm or have a way of determining the module type.

garden-service/src/plugins/kubernetes/util.ts
```diff
@@ -401,3 +407,135 @@ export function getSelectorString(labels: { [key: string]: string }) {
...
+      // need to resolve the template string
+      const chartPath = await getChartPath(<HelmModule>module)
+      targetName = await renderHelmTemplateString(ctx, log, module, chartPath, targetName)
+    }
+
+    target = find(<HotReloadableResource[]>manifests, (o) => o.kind === targetKind && o.metadata.name === targetName)!
+
+    if (!target) {
+      throw new ConfigurationError(
+        `Helm module '${module.name}' does not contain specified ${targetKind} '${targetName}'`,
```
eysi09: Same as above where we're referencing the Helm module type specifically. Also in the messages below.

garden-service/src/plugins/kubernetes/config.ts
edvald: Yeah I guess. This didn't change in the PR, just moved to a different file, but I can revise it along the way. Same actually goes for most/all of the other comments.

garden-service/src/plugins/kubernetes/kubernetes-module/run.ts
edvald: I didn't change this, so I can't say if that PR was supposed to update this as well... I can check to make sure.

garden-service/src/plugins/kubernetes/util.ts
edvald: Yep good catch

garden-service/src/plugins/kubernetes/kubernetes-module/run.ts
edvald: @thsig I'm actually not sure if this should change. Can you take a quick look? I basically duplicated this from https://github.com/garden-io/garden/blob/3619db0dbdce58e3778b0bec3afe3f149f2e858d/garden-service/src/plugins/kubernetes/helm/run.ts#L133

garden-service/src/plugins/kubernetes/kubernetes-module/run.ts
edvald: Yeah, plus that function can ensure the maximum length of the pod name. I'll throw that in there.

PR #1538, opened 2020-01-22
fix(build): error in alpine container image builds

edvald:


.circleci/config.yml
```diff
@@ -339,6 +340,27 @@ jobs:
...
+      - setup_remote_docker:
+          docker_layer_caching: true
+      # Need to checkout to run example project
+      - checkout
+      - configure_kubectl_context
+      - *attach-workspace
+      - run:
+          name: Deploy demo-project with container
+          # overriding CIRCLE_BUILD_NUM to avoid conflict with other tests
+          command: CIRCLE_BUILD_NUM=$CIRCLE_BUILD_NUM-docker garden-service/dist/linux-amd64/garden build --root examples/demo-project --env remote --logger-type basic
```
eysi09: Isn't this just running the binary from the attached workspace? Don't we want to run the `garden` binary that's a part of the container image?

edvald: Yeah, there are still several unresolved issues with this PR (should have marked it WIP). Kinda need to optimize this as well, split up the container builds too. The alpine image doesn't actually need to depend on the dist build, for one.

Issue #1477, opened 2020-01-05
Project dependencies


edvald:
## Background

Our current mechanism for importing other repositories as *remote sources* is handy in some cases but also fairly crude. Essentially it allows importing another repository as a new directory of sources, much like a submodule, which is then scanned for modules like any other directory. 

The motivation for that feature was primarily 
A) to remove barriers for adopting Garden when projects are already spread across repositories, and 
B) to (at least half way) support having related projects that can function individually, but also be combined into a single stack.

The main issue with the current implementation (as I see it), is that it completely ignores any project configuration contained within, and is by extension not much more useful than simply using git submodules or subtrees. In fact, the latter (despite their own UX flaws) are probably preferable since those are known and standard ways of nesting repositories.

Another issue is that different projects will need to be aware of each other to some degree, because we currently only have a single namespace for module names, and another for services and tasks combined. This is kinda workable by applying strict conventions, but hardly ideal. And, because all modules are treated equally, they all end up deployed to that same environment, namespace etc. So they must also make sure any included resources don't conflict.

The upside is that our work on that feature can serve as a solid building block for something more useful and Garden-native. 

Also, the related feature of *remote **module** sources* still holds up nicely, and can remain complementary to the reformulated project dependencies feature.

## Motivation and use cases

The motivation for project dependencies **remains largely the same as for remote sources, but goes a step further and makes the relationship more structured, which should improve UX and open the door for interesting new applications. Here are some use cases we want to enable:

1. Allow composing a Garden stack out of code from multiple repositories, in a deterministic fashion.
2. Enable working with portions of a larger stack, without having to build, deploy or even be aware of other parts of the larger stack.
3. Enable sharing an *instance* of a Garden stack between multiple projects. Meaning, supporting cases where a project referenced by a project should not be deployed individually for every instance of the parent project.
4. Enable sharing a Garden stack between multiple different projects. This can be within an organization, or even across organizations (e.g. delivering a stack of services as an open source project, or even a commercial product). Think "Helm chart, except for any supported platform, cloud services, Terraform stacks, whatever... And with structured parameters!"

That last two are new, and the last one is potentially very interesting by itself.

## Design considerations

We obviously want the design to be easy to understand, and easy to implement, both when sharing projects and consuming them.

Some of the more specific questions to consider are:

1. How do we handle the project configuration of a dependant project? Do we merge it with that of the parent project? If so, how?
[...]

PR #1546, opened 2020-01-24
feat(core): add project outputs and `garden get outputs` command

edvald:

This adds a facility for projects to export values, that can be viewed with the CLI and programmatically used by scripts. This is a foundational component of #1477.

### DIFF
--- a/garden-service/src/docs/templates/template-strings.hbs
+++ b/garden-service/src/docs/templates/template-strings.hbs
@@ -4,43 +4,42 @@ title: Template Strings Reference
 
 # Template string reference
 
-Below you'll find the schema of the keys available when interpolating template strings (see our
-[Configuration Files](../guides/configuration-files.md) guide for more information and usage examples).
+Below you'll find the schema of the keys available when interpolating template strings (see our [Configuration Files](../guides/configuration-files.md) guide for more information and usage examples).
 
-Note that there are three sections below, since Project configs and Module configs have different keys available to
-them, and since additional keys are available under `providers` in Project configs.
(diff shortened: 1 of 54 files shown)


garden-service/src/docs/templates/template-strings.hbs
```diff
@@ -4,43 +4,42 @@ title: Template Strings Reference
...
-them, and since additional keys are available under `providers` in Project configs.
-Please make sure to refer to the correct section.
-
-Modules can reference `outputs` defined by other modules, via the `${modules.<module-name>.outputs}` key, as well
-as service and task outputs via the `${runtime.services.<service-name>.outputs}` and
-`${runtime.tasks.<task-name>.outputs}`.
-For details on which outputs are available for a given module type, please refer to the
-[reference](https://docs.garden.io/module-types) docs for the module type in question, and look for the
-_Outputs_ section.
+Note that there are four sections below, since different configuration section have different keys available to them. Please make sure to refer to the correct section.
```
thsig: "section" -> "sections"

garden-service/src/util/string.ts
```diff
@@ -92,3 +93,34 @@ export function randomString(length = 8) {
 export function splitLines(s: string) {
   return s.split(/\r?\n/)
 }
+
+const defaultTableConfig: CliTable.TableConstructorOptions = {
```
thsig: Should we leave these commented lines in here?

edvald: Yeah, they're kind of useful as documentation imo.

PR #1548, opened 2020-01-26
fix(k8s): hash configuration annotation

### LINKED ISSUE #1496: garden plugins kubernetes cleanup-cluster-registry fails with metadata.annotations: Too long: must have at most 262144 characters
### Current Behavior

`garden plugins kubernetes cleanup-cluster-registry `fails with:
`The Deployment "garden-docker-registry" is invalid: metadata.annotations: Too long: must have at most 262144 characters`

`garden.io/last-applied-configuration` is the thing that gets unreasonably long.

### Expected behavior

It should not fail and restart the registry successfully.

### Reproducible example

Run `garden plugins kubernetes cleanup-cluster-registry` multiple times. 

### Workaround

Delete garden.io/last-applied-configuration in the deployment configuration on the cluster.

### Your environment

`garden version`: 0.11.0-alpha.4

eysi09:

Previously we stored the raw manifest under the
garden.io/last-applied-configuration. We now hash it and store it under
garden.io/manifest-hash. This prevents errors when then annotation
string gets too long.

Fixes #1496

### DIFF
--- a/garden-service/src/plugins/kubernetes/util.ts
+++ b/garden-service/src/plugins/kubernetes/util.ts
@@ -12,6 +12,7 @@ import { V1Pod, V1EnvVar } from "@kubernetes/client-node"
 import { apply as jsonMerge } from "json-merge-patch"
 import chalk from "chalk"
 import hasha from "hasha"
+import stableStringify = require("json-stable-stringify")
 
 import { KubernetesResource, KubernetesWorkload, KubernetesPod, KubernetesServerResource } from "./types"
 import { splitLast, serializeValues, findByName } from "../../util/util"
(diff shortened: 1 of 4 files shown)


garden-service/src/plugins/kubernetes/util.ts
```diff
@@ -19,6 +19,8 @@ import { ContainerEnvVars } from "../container/config"
 import { ConfigurationError } from "../../exceptions"
 import { KubernetesProvider } from "./config"
 import { LogEntry } from "../../logger/log-entry"
+import hasha from "hasha"
+import { stringify } from "querystring"
```
edvald: I think that's the wrong import. Previous iteration used a stable JSON serialization.

eysi09: Good catch! Just used the auto-import feature and obviously wasn't paying attention.

garden-service/src/plugins/kubernetes/util.ts
```diff
@@ -19,6 +19,8 @@ import { ContainerEnvVars } from "../container/config"
 import { ConfigurationError } from "../../exceptions"
 import { KubernetesProvider } from "./config"
 import { LogEntry } from "../../logger/log-entry"
+import hasha from "hasha"
+import stringify from "json-stringify-safe"
```
edvald: Still the wrong library. `json-stable-stringify` is the one, to ensure that it's deterministic.

eysi09: 🤦‍♂️

PR #1554, opened 2020-01-27
fix(core): make test task deps hot-reload aware

### LINKED ISSUE #1552: Hot reload will fail the second time if test depends on service being hot reloaded
### Current Behavior

`garden dev --hot=<module name>` will fail to hot reload the second time around if there is a test that depends on service `<module name>`.

### Expected behavior

`garden dev --hot=<module name>` should hot reload every time not just the first time.

### Reproducible example

Have a module with hot reload defined, a service and a test which depends on said service.
Run `garden dev --hot=<module name>` and do two hot reloads triggered by file change. 2nd one should fail.

### Workaround

Remove test?

### Suggested solution(s)

If in watch mode with hot reload enabled make sure that hot reload will be deployed again when a test depends on the service that is being hot reloaded.

### Your environment

`garden version`: 0.11.1

thsig:

Before this fix, the `getDependencies` method of `TestTask` returned deploy tasks without the hot reload flag set appropriately.

This meant that repeated hot reloads of services with test dependants would fail (because the test dependant would redeploy the service without hot reloading enabled) when using the dev command.

Fixes #1552.

Do we think this needs additional test cases? Or is this fine as is?

### DIFF
--- a/garden-service/src/commands/dev.ts
+++ b/garden-service/src/commands/dev.ts
@@ -204,6 +204,7 @@ export class DevCommand extends Command<DevCommandArgs, DevCommandOpts> {
                   module: m,
                   graph: updatedGraph,
                   filterNames,
+                  hotReloadServiceNames,
                 })
               )
             )
--- a/garden-service/src/tasks/test.ts
+++ b/garden-service/src/tasks/test.ts
@@ -8,7 +8,7 @@
 
 import Bluebird from "bluebird"
 import chalk from "chalk"
-import { find } from "lodash"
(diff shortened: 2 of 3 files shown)


edvald:
I approved the PR, but on reflection a test case or two would be good. This is one of those things that easily breaks during refactors.

thsig:
Yeah, I agree. I’ll have a go at writing a test or two around this.

thsig:
Added a test for `getTestTasks` that checks this logic.

PR #1566, opened 2020-02-04
chore: fix log messages

eysi09:

Bumped into this when recording the screencaps for the website.

### DIFF
--- a/garden-service/src/commands/deploy.ts
+++ b/garden-service/src/commands/deploy.ts
@@ -27,6 +27,7 @@ import { getHotReloadServiceNames, validateHotReloadServiceNames } from "./helpe
 import { startServer, GardenServer } from "../server/server"
 import { DeployTask } from "../tasks/deploy"
 import { naturalList } from "../util/string"
+import chalk = require("chalk")
 
 const deployArgs = {
   services: new StringsParameter({
@@ -109,7 +110,9 @@ export class DeployCommand extends Command<Args, Opts> {
     const disabled = services.filter((s) => s.disabled).map((s) => s.name)
 
     if (disabled.length > 0) {
-      log.info({ symbol: "info", msg: `Services ${naturalList(disabled)} are disabled` })
+      const bold = disabled.map((d) => chalk.bold(d))
+      const msg = disabled.length === 1 ? `Service ${bold} is disabled` : `Services ${naturalList(bold)} are disabled`
+      log.info({ symbol: "info", msg: chalk.white(msg) })
     }
 
     services = services.filter((s) => !s.disabled)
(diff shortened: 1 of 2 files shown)


garden-service/src/commands/deploy.ts
```diff
@@ -109,7 +110,9 @@ export class DeployCommand extends Command<Args, Opts> {
     const disabled = services.filter((s) => s.disabled).map((s) => s.name)
 
     if (disabled.length > 0) {
-      log.info({ symbol: "info", msg: `Services ${naturalList(disabled)} are disabled` })
+      const msg =
+        disabled.length === 1 ? `Service ${disabled} is disabled` : `Services ${naturalList(disabled)} are disabled`
```
edvald: Shouldn't this be `${disabled[0]}`?

edvald: Also, might be nice to color the names somehow 🤷‍♂I like that as a visual style, also makes it more clear in case a service name somehow fits strangely into the surrounding sentence, e.g. "Service admin is disabled".

eysi09: > Shouldn't this be ${disabled[0]}?

For some reason the other way works, that's why I didn't notice it:

```
Deploy 🚀

✔ providers                 → Getting status... → Done
ℹ Service frontend is disabled
✔ backend                   → Syncing module sources (2 files)... → Done (took 0 sec)
```

eysi09: Now it looks like this:

Screenshot 2020-02-09 at 10 37 11

and this:

Screenshot 2020-02-09 at 10 37 30

Also added an Oxford comma to the `naturalList` (naturalist?) function.

garden-service/src/util/string.ts
```diff
@@ -68,15 +68,15 @@ export function base64(str: string) {
...
+ * Example: `naturalList(["a", "b", "c"])` -> `"a, b, and c"`
  */
 export function naturalList(list: string[]) {
   if (list.length === 0) {
     return ""
   } else if (list.length === 1) {
     return list[0]
   } else {
-    return list.slice(0, -1).join(", ") + " and " + list[list.length - 1]
+    return list.slice(0, -1).join(", ") + ", and " + list[list.length - 1]
```
edvald: Haha ok. Not be an absolute stickler (although you arguably opened that door up 😜), but if we are going to go with the Oxford comma it should only be used if there are three or more items. I think it looks a little odd in the context to begin with, but won't object if you fix that case.

eysi09: Ah good catch. And for some reason it bothered me when I first saw it without the comma, but on second thought, I agree. Reverting.

PR #1582, opened 2020-02-08
feat(cli): add `garden create project/module` commands

edvald:

See the docs and CLI help texts for details. The generated configs
contain the full project/module config schema (minus deprecated fields),
so users can easily uncomment and/or remove fields depending on their
needs.

Providers are asked to suggest modules, and I implemented that for
`container`, `helm` and `terraform` modules.

This required a bit of refactoring in our docs generation code, so a
bunch of the line changes relate to that. Also needed to make all links
in field descriptions absolute, and tweak the line breaks and formatting
for readability.

### DIFF
--- a/garden-service/src/commands/create/create-module.ts
+++ b/garden-service/src/commands/create/create-module.ts
@@ -56,7 +56,6 @@ const createModuleOpts = {
   type: new StringOption({
     help: "The module type to create. Required if --interactive=false.",
   }),
-  // TODO: add type option
 }
 
 type CreateModuleArgs = typeof createModuleArgs
--- a/garden-service/src/commands/create/create-module.ts
+++ b/garden-service/src/commands/create/create-module.ts
@@ -1,0 +1,252 @@
+/*
+ * Copyright (C) 2018-2020 Garden Technologies, Inc. <info@garden.io>
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import chalk from "chalk"
+import dedent from "dedent"
+import { pathExists } from "fs-extra"
+import inquirer from "inquirer"
+import {
+  Command,
+  CommandResult,
+  CommandParams,
+  PrepareParams,
+  PathParameter,
+  BooleanParameter,
+  StringOption,
+} from "../base"
+import { printHeader } from "../../logger/util"
+import { getConfigFilePath, isDirectory } from "../../util/fs"
+import { loadConfig, findProjectConfig } from "../../config/base"
+import { resolve, basename, relative } from "path"
+import { GardenBaseError, ParameterError } from "../../exceptions"
+import { getModuleTypes } from "../../plugins"
+import { addConfig } from "./helpers"
+import { supportedPlugins } from "../../plugins/plugins"
+import { baseModuleSpecSchema } from "../../config/module"
+import { renderConfigReference } from "../../docs/config"
+import { DOCS_BASE_URL } from "../../constants"
+import { flatten } from "lodash"
+import { fixedPlugins } from "../../config/project"
+import { deline, wordWrap } from "../../util/string"
+import { joi } from "../../config/common"
+import { LoggerType } from "../../logger/logger"
+
+const createModuleArgs = {}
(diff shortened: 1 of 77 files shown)


garden-service/src/commands/create/create-module.ts
```diff
@@ -0,0 +1,316 @@
...
+    help: "Set to false to disable interactive prompts.",
+    defaultValue: true,
+  }),
+  name: new StringOption({
+    help: "Name of the module (defaults to current directory name).",
+  }),
+  type: new StringOption({
+    help: "The module type to create. Required if --interactive=false.",
+  }),
+  // TODO: add type option
```
thsig: Delete comment?

garden-service/src/commands/create/create-project.ts
```diff
@@ -0,0 +1,189 @@
...
+    // This is to avoid `prettier` messing with the string formatting...
+    const formattedIgnoreName = chalk.bold.white(".gardenignore")
+    const configFilesUrl = chalk.cyan.underline("https://docs.garden.io/guides/configuration-files")
+    const referenceUrl = chalk.cyan.underline("https://docs.garden.io/reference/config")
+
+    log.info({
+      symbol: "info",
+      msg: wordWrap(
+        dedent`
+        We recommend reviewing the generated config, uncommenting fields that you'd like to configure, and cleaning up any commented fields that you don't need to use. Also make sure to update the ${formattedIgnoreName} file with any files you'd like to exclude from the Garden projects.
```
thsig: "the Garden projects" -> "the Garden project"

garden-service/src/config/common.ts
```diff
@@ -262,13 +262,9 @@ export const joiIdentifierDescription =
 
 const moduleIncludeDescription = (extraDescription?: string) => {
   const desc = dedent`
-  Specify a list of POSIX-style paths or globs that should be regarded as the source files for this
-  module. Files that do *not* match these paths or globs are excluded when computing the version of the module,
-  when responding to filesystem watch events, and when staging builds.
+  Specify a list of POSIX-style paths or globs that should be regarded as the source files for this   module. Files that do *not* match these paths or globs are excluded when computing the version of the module, when responding to filesystem watch events, and when staging builds.
```
thsig: Delete extra spaces after "source files for this".

garden-service/src/plugin-context.ts
```diff
@@ -66,7 +66,8 @@ export const pluginContextSchema = joi
     production: joi
       .boolean()
       .default(false)
-      .description("Indicate if the current environment is a production environment"),
+      .description("Indicate if the current environment is a production environment")
```
thsig: Add period after "environment".

edvald:
@thsig all resolved. Really no other comments than typos? :P Did you try using it?

thsig:
> @thsig all resolved. Really no other comments than typos? :P Did you try using it?

Just took it for a more proper spin now. Some comments:
* Should we allow running create project outside of a git repo? Maybe that would just be annoying (since it would presumably have to create a directory and run `git init` for you).
* Allow specifying module name on CLI when creating module? One time, I intuitively wrote `garden create module foo` without thinking.
* The create module command allows me to create several modules with the same name if I run the command in different subdirectories of project root.
  * No error is thrown if I:
    * Run `garden create module` in the project root dir and choose the name `foo` for the module;
    * Then create a dir called foo, cd to it and run `garden create module` in it and choose `foo` again as the module name.
    * I can repeat this process with different subdirectories and the module name `foo`.

I like the UI in general, this is a nice feature!

edvald:
All good points. One by one:

- Running outside of a git repo should be fine, just hadn't thought to test that. I can see if it's trivial to add.
- There is an optional `--name` argument for the create module command. I decided against a positional parameter, because there wasn't in mind an obvious choice between making the positional arg the module _type_ or _name_. So I went for explicit. Does that make sense?
- Yeah, we only check that the given _config_ doesn't already include a module with a certain name. We do load the project (if applicable) though, so we could just as well throw if a module with the same name already exists.

thsig:
> There is an optional --name argument for the create module command. I decided against a positional parameter, because there wasn't in mind an obvious choice between making the positional arg the module type or name. So I went for explicit. Does that make sense?

👍

> We do load the project (if applicable) though, so we could just as well throw if a module with the same name already exists.

I think throwing here would be nice (it's easier to run the command again than to find the offending module config file and replace the name later on).

PR #1584, opened 2020-02-10
fix(k8s): fixed handling of timeouts when artifacts are being copied

edvald:

See the added tests. We weren't handling timeouts correctly or consistently depending on whether artifacts were being copied or not.

### DIFF
--- a/garden-service/src/plugins/kubernetes/run.ts
+++ b/garden-service/src/plugins/kubernetes/run.ts
@@ -38,7 +38,6 @@ export async function runAndCopy({
   module,
   args,
   command,
-  ignoreError,
   interactive,
   runtimeContext,
   timeout,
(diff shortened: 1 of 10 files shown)


garden-service/src/plugins/kubernetes/run.ts
```diff
@@ -290,22 +321,20 @@ export async function runAndCopy({
...
-      if (containerLogs[0].log) {
-        result.log = containerLogs[0].log
+      result = await runner.startAndWait({
+        interactive,
+        ignoreError: true,
+        log,
+        remove: false,
+        timeout,
+      })
+      result.log = (await getLogs()) || result.log
```
thsig: Missing `.trim()` after `(await getLogs())`?

PR #885, opened 2019-06-26
improvement(Windows): check for Hyper-V and ask if user wants Docker

edvald:
Allows installation on machines without Hyper-V option, and avoids hard
requirement on Docker.

PR #1586, opened 2020-02-11
fix(Windows): don't always install docker-for-windows

J0sh0nat0r:
Prevents Windows installer from installing `docker-for-windows` without first asking the user. (As implemented in #885)


edvald:
Thanks @J0sh0nat0r! Much appreciated :)

PR #1588, opened 2020-02-11
improvement(maven-container): add useDefaultDockerfile field

### LINKED ISSUE #1570: maven-container should fallback to image pull like container modules do
### Current Behavior

With container modules, when there is no Dockerfile, it reverts to pulling the image instead of building. This is a brilliant feature as it allows my developers to run our suite of microservices from pre-built images and then selectively link the modules they want to develop on and garden will see the Dockerfile and switch to building from code.

With maven-container modules however, it expects that the source is always present and fails if there is no pom.xml

### Expected behavior

Maven-container modules should behave the same way as container modules and revert to image pull if there is no Dockerfile (or perhaps pom.xml).

### Reproducible example

kind: Project
name: vartest
environments:
  - name: local
    providers:
      - name: local-kubernetes
      - name: maven-container

---

kind: Module
type: maven-container
name: maven
jarPath: app.jar
image: hello-world

### Suggested solution(s)

In maven-container.ts you can see that build() immediately tries to locate the pom and will fail if it’s not present:
https://github.com/garden-io/garden/blob/master/garden-service/src/plugins/maven-container/maven-container.ts#L172
Whereas in container/build.ts, you can see it checks if the Dockerfile is present and if not falls back to pulling the image:
https://github.com/garden-io/garden/blob/master/garden-service/src/plugins/container/build.ts#L36

### Additional context

See slack thread here: https://kubernetes.slack.com/archives/CKM7CP8P9/p1580899251183900

### Your environment

`garden version` 0.11.1

`kubectl version`
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.0", GitCommit:"91e7b4fd31fcd3d5f436da26c980becec37ceefe", GitTreeState:"clean", BuildDate:"2018-06-27T22:29:25Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.5", GitCommit:"20c265fef0741dd71a66480e35bd69f18351daea", GitTreeState:"clean", BuildDate:"2019-10-15T19:07:57Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}

`docker version`
19.03.5 / docker for mac

eysi09:

This allows users to fallback to the `image` field by setting
`useDefaultDockerfile` to `false` and not including a Dockerfile of their
own.

Fixes #1570


edvald:
Seems we have another flaky test, after #1584 was merged. I'll look into that.

PR #1589, opened 2020-02-12
fix(core): remove StageBuildTask from GetServiceStatusTask dependencies

### LINKED ISSUE #1583: Helm modules deletes the build folder when trying to access logs through the dashboard.
### Current Behavior

From Slack user Yann Irbah:

> :wave: Hi there. I (finally) started to add Garden to one of our new projects and I have a weird behavior. Everything was working fine with simple container modules. Then I started using Helm modules and every time I try to access logs through the UI, it deletes the `build` folder for the modules in the `.garden` directory and start looping with the following error `.garden/build/blueprints-chart/garden-values.yml: no such file or directory` (which makes sense since the directory is empty at this point).

Link to [discussion](https://kubernetes.slack.com/archives/CKM7CP8P9/p1581085718213500).

eysi09:

We shouldn't be staging the build when getting the service status as it
can remove build artifacts like those generated by the `helm` module. Had
to make adjustments to the `kubernetes` module `getServiceStatus` handler to
account for this change.

Fixes #1583

### DIFF
--- a/garden-service/src/plugins/kubernetes/kubernetes-module/common.ts
+++ b/garden-service/src/plugins/kubernetes/kubernetes-module/common.ts
@@ -22,13 +22,20 @@ import { LogEntry } from "../../../logger/log-entry"
  * Reads the manifests and makes sure each has a namespace set (when applicable) and adds annotations.
  * Use this when applying to the cluster, or comparing against deployed resources.
  */
-export async function getManifests(
-  api: KubeApi,
-  log: LogEntry,
-  module: KubernetesModule,
+export async function getManifests({
+  api,
+  log,
+  module,
+  defaultNamespace,
+  readFromSrcDir = false,
+}: {
+  api: KubeApi
+  log: LogEntry
+  module: KubernetesModule
   defaultNamespace: string
-): Promise<KubernetesResource[]> {
-  const manifests = await readManifests(module)
+  readFromSrcDir?: boolean
+}): Promise<KubernetesResource[]> {
+  const manifests = await readManifests(module, log, readFromSrcDir)
 
   return Bluebird.map(manifests, async (manifest) => {
     // Ensure a namespace is set, if not already set, and if required by the resource type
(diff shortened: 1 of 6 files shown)


garden-service/src/plugins/kubernetes/kubernetes-module/common.ts
```diff
@@ -50,11 +57,21 @@ export async function getManifests(
...
 /**
  * Read the manifests from the module config, as well as any referenced files in the config.
+ *
+ * @param module The kubernetes module to read manifests for.
+ * @param readFromSrcDir Whether or not to read the manifests from the module build dir or from the module source dir.
+ * In general we want to read from the build dir to ensure that manifests added via the `build.dependencies[].copy`
+ * field will be included. However, in some cases, e.g. when getting the service status, we can't be certain that
+ * the build has been staged and we therefore read the manifests from the source.
+ *
+ * FIXME: Always read from build dir.
```
edvald: The fix for this will rather be to check a version hash, when possible, not to always read from the build dir.

edvald: I don't expect we'll ever revert to building/staging before getting service statuses

garden-service/src/types/plugin/service/getServiceStatus.ts
```diff
@@ -27,6 +27,9 @@ export const getServiceStatus = {
 
     Called ahead of any actions that expect a service to be running, as well as the
     \`garden get status\` command.
+
+    NOTE: This handler should not use the build directory since it's not guaranteed
+    that the build will be staged.
```
edvald: "... staged or completed before this handler is called."

garden-service/src/plugins/kubernetes/kubernetes-module/common.ts
eysi09: And in that case, we won't need the manifests in the `getServiceStatus` handler, right?

edvald: Fair point. Depends if we only check the hash, or do both. In any case, all good, we'll revisit at that time.

PR #1592, opened 2020-02-12
chore(dashboard): show disabled entities

eysi09:

We weren't displaying disabled entities in the dashboard, this PR fixes that. Had to make some changes to how we fetch the init dashboard data to facilitate that.

Here's how they look:

Screenshot 2020-02-12 at 23 45 22

Screenshot 2020-02-12 at 23 46 00

This should have been two commits, one for the disabled UI and one for the data fetching. I only realised midway that we had to make changes to the fetching logic and at that point the two were kind of coupled together. My apologies :)


dashboard/src/api/actions.ts
```diff
@@ -28,14 +38,45 @@ import {
 } from "./api"
 import { getTestKey } from "../util/helpers"
 
+// This file contains the API action functions.
+// The actions are responsible for dispatching the appropriate action types and normalising the
```
thsig: "normalising" -> "normalizing"

PR #1594, opened 2020-02-13
fix(terraform): increase init timeout

solomonope:
Increase the exec timeout when calling terraform init

Fixes #


garden-service/src/plugins/terraform/common.ts
```diff
@@ -51,7 +51,7 @@ export async function tfValidate(log: LogEntry, provider: TerraformProvider, roo
     if (reasons.includes("Could not satisfy plugin requirements") || reasons.includes("Module not installed")) {
       // We need to run `terraform init` and retry validation
       log.info("Initializing Terraform")
-      await terraform(tfVersion).exec({ log, args: ["init"], cwd: root })
+      await terraform(tfVersion).exec({ log, args: ["init"], cwd: root, timeout: 120 })
```
edvald: Could be even longer actually, make it 300 sec maybe?

solomonope: ok

solomonope: done

PR #1597, opened 2020-02-13
improvement(k8s): use deploymentRegistry with in-cluster building

edvald:
**What this PR does / why we need it**:

Previously we would always deploy and use the in-cluster registry
when building in-cluster. Now we allow using the configured
`deploymentRegistry`, which is often preferable (and more scalable)
than using the simpler in-cluster registry.

**Which issue(s) this PR fixes**:

Closes #1034

PR #1598, opened 2020-02-14
feat: allow disabling result caching for tasks that support it

edvald:

Certain tasks should be run every time their dependants are run/deployed, rather than getting cached results and passing those down the graph. The `cacheResult` flag allows users to control this behavior.

We had previously talked about a more elaborate multi-value argument for controlling this, but with the `disabled` flag implemented a simple boolean should suffice.

### DIFF
--- a/docs/module-types/container.md
+++ b/docs/module-types/container.md
@@ -358,17 +358,17 @@ tasks:
     # task is executed.
     dependencies: []
 
-    # Set this to `true` to disable the task. You can use this with conditional template strings to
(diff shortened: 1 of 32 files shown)


thsig:
Looks good.

Should we also add integ tests for the `container` module type and the k8s plugins that test their task caching/reading implementations?

edvald:
Yeah, I guess we didn't have existing tests for the `kubernetes` result caching, I can add a couple there.

edvald:
FYI, tests are failing because of an issue that I fixed in #1597

PR #1603, opened 2020-02-14
fix(k8s): fix build-sync Pod crash loop issue

edvald:

Some users have been reporting a crash loop that occurs sporadically in the build-sync pods, which messes with builds and is hard to work around. This fix introduces a simpler version of the rsync image, that hopefully addresses it.

I think there was an open issue but I can't find it atm 😬 

This is WIP while I get feedback from @Eronarn and others.


edvald:
This should be good to go now.

PR #1624, opened 2020-02-20
fix: issues with running HTTPS requests through HTTP proxy

edvald:

This adds and configured `global-agent`
(https://github.com/gajus/global-agent) and replaces Axios with `got` 
(https://github.com/sindresorhus/got) which, since Axios has a
long-standing issue with routing HTTPS requests through HTTPS proxies.

Reported by @khaled on Slack


garden-service/src/plugins/kubernetes/commands/cleanup-cluster-registry.ts
```diff
@@ -126,12 +126,12 @@ async function getImagesInRegistry(ctx: KubernetesPluginContext, log: LogEntry)
 
   while (nextUrl) {
```
thsig: Maybe we should extract this `while` loop(which is mostly repeated below) into a helper function?

thsig:
Should we globally replace Axios with `got`? We're still using it in a couple of places (one in the dashboard, and in `support/support-util.ts`.

garden-service/src/plugins/kubernetes/commands/cleanup-cluster-registry.ts
edvald: This is not related to the PR, per se, but yeah I guess so.

edvald:
We could, sure. It doesn't really matter for those cases, except for consistency, but that has its own merit.

PR #1626, opened 2020-02-20
fix(core): tasks should not implicitly run when getting service status

edvald:
cc @eysi09


garden-service/test/helpers.ts
```diff
@@ -512,3 +513,11 @@ export async function makeTempDir(): Promise<TempDirectory> {
   tmpDir.path = await realpath(tmpDir.path)
   return tmpDir
 }
+
+/**
+ * Retrieves all the child log entries from the given LogEntry and returns a list of all the messages,
+ * stripped of ANSI characters. Useful to check if a particular message was logged.
+ */
+export function getLogMessages(log: LogEntry) {
+  return log.root.getLogEntries().map((l) => stripAnsi(l.getMessageState().msg || ""))
```
eysi09: The `getMessageState` only returns the latest message state. In this case you'd probably want something like `getMessageStates().map(msgState => msgState.msg).join(, )` (or similar).

PR #1633, opened 2020-02-21
docs: update ToC layout

edvald:

This makes the table of contents in our docs site more readable (and attractive), and certain documentation sections more prominent.


docs/README.md
```diff
@@ -1,57 +1,71 @@
...
+* [Container Modules](./guides/container-modules.md)
+* [Development Workflows](./guides/development-workflows.md)
+* [Helm Charts](./guides/using-helm-charts.md)
+* [Hot Reload](./guides/hot-reload.md)
+* [In-Cluster Building](./guides/in-cluster-building.md)
+* [Local Kubernetes](./guides/local-kubernetes.md)
+* [Remote Kubernetes](./guides/remote-kubernetes.md)
+* [Terraform](./guides/terraform.md)
+* [Variables and templating](./guides/variables-and-templating.md)
+* [Advanced](./guides/advanced/README.md)
```
eysi09: Would it make sense to also have the _Advanced_ section top level? So **Guides** and **Advanced Guides**. It's a little hidden like this. I also don't think that _Remote Sources_ should be considered advanced, but perhaps Terraform should.

docs/README.md
```diff
@@ -1,57 +1,71 @@
...
+  * [`helm`](./reference/module-types/helm.md)
+  * [`kubernetes`](./reference/module-types/kubernetes.md)
+  * [`maven-container`](./reference/module-types/maven-container.md)
+  * [`openfaas`](./reference/module-types/openfaas.md)
+  * [`persistentvolumeclaim`](./reference/module-types/persistentvolumeclaim.md)
+  * [`terraform`](./reference/module-types/terraform.md)
+* [Commands](./reference/commands.md)
+* [Config Files](./reference/config.md)
+* [Template Strings](./reference/template-strings.md)
+* [Glossary](./reference/glossary.md)
```
eysi09: Wondering if we should move this one to Basics? At least given its current contents. We could also rename it to something like _Concepts_ or similar. (Not part of the PR per se).

garden-service/src/commands/create/create-module.ts
```diff
@@ -251,8 +252,8 @@ export class CreateModuleCommand extends Command<CreateModuleArgs, CreateModuleO
     }
 
     // This is to avoid `prettier` messing with the string formatting...
-    const moduleTypeUrl = chalk.cyan.underline(`${DOCS_BASE_URL}/module-types/${type}`)
-    const providerUrl = chalk.cyan.underline(`${DOCS_BASE_URL}/providers/${pluginName}`)
+    const moduleTypeUrl = chalk.cyan.underline(getModuleTypeUrl(type))
+    const providerUrl = chalk.cyan.underline(getProviderUrl(pluginName))
     const configFilesUrl = chalk.cyan.underline(`${DOCS_BASE_URL}/guides/configuration-files`)
```
eysi09: Will using the `DOCS_BASE_URL` not break the link checker? This links to the latest release version that only gets updated on releases, right?

garden-service/src/docs/table-of-contents.ts
```diff
@@ -81,7 +81,7 @@ function attachMetadata(tree: FileTree) {
     if (metadata.order) {
       tree.order = metadata.order
     } else {
-      tree.order = Number.MAX_VALUE
+      tree.order = 1000
```
eysi09: Why change this?

garden-service/src/docs/table-of-contents.ts
```diff
@@ -100,6 +100,7 @@ function attachMetadata(tree: FileTree) {
     tree.title = titleize(humanizeString(tree.name))
     tree.order = Number.MAX_VALUE
```
eysi09: Here we're still using `MAX_VALUE`. I suppose it could also be `Infinity`, not sure what's most semantically correct. But if it's just used for ordering, I think either of those convey the intent better than some arbitrary large number.

garden-service/src/plugins/conftest/conftest.ts
```diff
@@ -59,14 +59,18 @@ interface ConftestModuleSpec {
...
 
-    Note that, in many cases, you'll actually want to use more specific providers that can automatically configure your \`conftest\` modules, e.g. the [\`conftest-container\`](${DOCS_BASE_URL}/providers/conftest-container) and/or [\`conftest-kubernetes\`](${DOCS_BASE_URL}/providers/conftest-kubernetes) providers. See the [conftest example project](https://github.com/garden-io/garden/tree/master/examples/conftest) for a simple usage example of the latter.
+    Note that, in many cases, you'll actually want to use more specific providers that can automatically configure your \`conftest\` modules, e.g. the [\`conftest-container\`](${containerModuleTypeUrl}) and/or [\`conftest-kubernetes\`](${kubernetesModuleTypeUrl}) providers. See the [conftest example project](https://github.com/garden-io/garden/tree/master/examples/conftest) for a simple usage example of the latter.
```
eysi09: Not related to the PR per se, but shouldn't the GitHub example links always point to a particular version? We could just read the version from `package.json` instead of transforming this at the docs render level like we talked about, in fact I think that's clearer. So something like: `https://github.com/garden-io/garden/tree/${version}/examples/conftest`

garden-service/src/docs/table-of-contents.ts
edvald: So that we can set certain items to order last, and set an order between those.

garden-service/src/docs/table-of-contents.ts
edvald: See comment above. The arbitrary large number gives us flexibility to order things at the end.

garden-service/src/plugins/conftest/conftest.ts
edvald: Sure, out of scope for the PR but I'd be fine with that. Make a helper function maybe.

garden-service/src/commands/create/create-module.ts
edvald: That's the same as it has been then since we merged the create comands. We've had that problem for a while, that new/updated links don't pass validation until after merge, and I unfortunately don't see a good way around that because we can't use relative URLS in generated YAML. And I really don't want do some complex/brittle URL parsing to address it.

docs/README.md
edvald: I'd be okay with that if we update the guide, could do with a refresh.

docs/README.md
edvald: That's all debatable. I think remote sources are for sure advanced, but maybe Terraform also. That's all a matter of perspective 🤷‍♂️

Matters less, of course, if we move the advanced section higher. I do like having this distinction though.

eysi09: Yeah I definitely agree we should have an _Advanced_ section. And it can of course be debatable what we put in there. But if it's top level it's at least  visible. Do you think we'd have too many top-level sections if we put _Advanced Guides_ there as well?

docs/README.md
eysi09: We can do that in a separate PR.

docs/README.md
edvald: I moved them to a top-level Advanced section, and moved Terraform there too.

PR #1636, opened 2020-02-22
fix(k8s): issues with querying registries for image tags

### LINKED ISSUE #1380: In-cluster builds fail to push to garden registry with microk8s clusters
### Current behavior

Pushes to the in-cluster registry installed by garden fails.

### Expected behavior

Pushes succeed!

### Reproducible example

Any example with in-cluster building.  I used in-cluster docker in my trials.

### Workaround

I suspect this has to do with `hostPort` (which the registry proxy appears to use) not being properly supported by microK8s.

I worked around this by performing the following steps:

* Modified `/var/snap/microk8s/current/args/kube-apiserver` on the cluster node, adding `--service-node-port-range 5000-32767` (it defaults to `30000-32767`).

* Modified `/var/snap/microk8s/current/args/docker-daemon.json` on the cluster node, adding `localhost:5000` to the list of insecure-registries.

* Restarted both the kube apiserver and the docker daemon:

	```bash
	sudo systemctl restart snap.microk8s.daemon-apiserver.service
	sudo systemctl restart snap.microk8s.daemon-docker.service
	```

* Delete the registry proxy DaemonSet:

	```
	kubectl delete ds/garden-registry-proxy -n garden-system
	```
		
* Created a NodePort service on port 5000, pointing to the registry:

	```yaml
	apiVersion: v1
	kind: Service
	metadata:
	  name: garden-docker-registry-nodeport
	  namespace: garden-system
	spec:
	  type: NodePort
	  ports:
	    - name: registry-port
	      nodePort: 5000
	      port: 5000
	      protocol: TCP
	      targetPort: 5000
	  selector:
	    app: docker-registry
	```
	
### Suggested solution(s)

It might simplify things to be able to use the microk8s registry addon, which listens on a host port of 32000, instead of garden installing its own.  MicroK8s is pre-configured with 32000 as an insecure registry.

### Your environment

```
$ garden --version

0.10.15
```

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.3", GitCommit:"b3cbbae08ec52a7fc73d334838e18d17e8512749", GitTreeState:"clean", BuildDate:"2019-11-18T14:56:51Z", GoVersion:"go1.12.13", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.9", GitCommit:"e09f5c40b55c91f681a46ee17f9bc447eeacee57", GitTreeState:"clean", BuildDate:"2019-05-27T15:58:45Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}

```
```
$ microk8s.docker version
Client:
 Version:           18.09.2
 API version:       1.39
 Go version:        go1.10.4
 Git commit:        6247962
 Built:             Tue Feb 26 23:56:24 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          18.09.2
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.4
  Git commit:       6247962
  Built:            Tue Feb 12 22:47:29 2019
  OS/Arch:          linux/amd64
  Experimental:     false
```

edvald:

This affected users using a non-default deploymentRegistry, as well as
microk8s users. The registry is now properly queried when checking
build status in all cases, both when building locally and remotely,
and whether an in-cluster registry is used or not.

Fixes #1380


eysi09:
Could you explain the PR a little better. I noticed we bumped the min Docker client version and that we're now downloading the bin if users don't have it. How and why are we using the new Docker version?

edvald:
Sure. The `docker manifests inspect` command is only available as an experimental command is only available in the latest major docker release. That command provides a simple way to query registries for info, which we can use to check whether an image/tag exists in a registry. The reason we can't just use an API directly is because of auth. Docker supports a number of auth methods, at least locally (credential helpers and such) and we can't reliably replicate that or use libraries that do.

For the in-cluster building scenario I could have gotten around installing the new version because we exec from the in-cluster builder. The issue there had been that I left an assumption in the build status check that we were using the in-cluster registry. The new flow works for either scenario.

For local builds, this had been an open issue for a while. We'd check for the image in the local docker daemon, and if it existed we'd assume the same image was in the deployment registry. But if something had failed or the tag removed from the deployment registry, users would see issues.

I only realized yesterday that this was causing the problems for microk8s because we use the deployment registry flow there, due to how microk8s is set up.

Changing the minimum required docker version would have been a breaking change, so I needed to fetch docker automatically to address this. I added a bit of extra logic to avoid fetching it when a recent version is already installed, since that's a very common case, and this may otherwise come across as a bit unexpected.

garden-service/src/plugins/container/build.ts
```diff
@@ -66,9 +68,12 @@ export async function buildContainerModule({ module, log }: BuildModuleParams<Co
...
   const outputStream = createOutputStream(log.placeholder(LogLevel.debug))
   const timeout = module.spec.build.timeout
-  const buildLog = await containerHelpers.dockerCli(module, [...cmdOpts, buildPath], { outputStream, timeout })
+  const res = await containerHelpers.dockerCli(module.buildPath, [...cmdOpts, buildPath], log, {
+    outputStream,
+    timeout,
+  })
 
-  return { fresh: true, buildLog, details: { identifier } }
+  return { fresh: true, buildLog: res.output || "", details: { identifier } }
```
dcharbonnier:
```suggestion
  return { fresh: true, buildLog: res.output ?? "", details: { identifier } }
```

edvald: Ah yes, I keep forgetting that new operator. Makes no semantic difference in this particular case but it's worth keeping in mind.

PR #1637, opened 2020-02-24
fix(k8s): include value files in default Helm include clause

10ko:
When using the `helm` module type if the user doesn't specify any value in the `include:[]` field we don't automatically include any `values.yaml` that are needed for the chart.

Fixes a request from our community channel.


edvald:
I amended to also include explicitly set `valueFiles` to fix #1638

PR #1639, opened 2020-02-24
fix(cli): create command didn't work from release build

edvald:
Ugh, this is a straight-up facepalm. We really need to work out how
to run our test suite with the release binary.


eysi09:
Actually, we could also add this command to the `bin/test_release.sh` script. I know it's not optimal to only catch these just before a release, but I suppose it's better than not catching them at all.

PR #1644, opened 2020-02-25
fix(container): strip quotes from Dockerfile paths when setting includes

edvald:

A user ran into an error where a copied path was quoted. Quick fix.


garden-service/src/plugins/container/helpers.ts
```diff
@@ -373,8 +374,11 @@ const helpers = {
             return args.slice(0, -1)
           }
         })
+      )
+        // Strip quotes from quoted paths
+        .map(stripQuotes)
```
eysi09: I'm always vary of not expanding the callback function after bumping into a gnarly bug the other day. So `.map((el => fn(el))` instead of `map(el)`. 

If you were to say later add an optional parameter like `stipQuotes(str, exlcudeDoubleQuotes)`, you might get different results for element with index 0, since the index would be the second argument.

edvald: That's fair. Can we lint for that?

eysi09: I tried this but it didn't warn in the above case: https://github.com/buzinas/tslint-eslint-rules/blob/master/src/docs/rules/terPreferArrowCallbackRule.md

Not sure if there's anything else.

PR #1649, opened 2020-02-27
Do not set in-cluster registry namespace to be the K8S namespace

swist:
We would like to share builds across namespaces, but instead we
see that each namespace gets its own repository in the in-cluster
registry. Hopefully this should push everything to its own org
as it won't get overriden by k8s namespace from root k8s provider
config?


swist:
My assumption is that all of the code that verifies the existence of built image for a particular image hash uses the configured container registry to check if the image exists first.

edvald:
That actually makes perfect sense to me. I'm thinking a bit to see if there might be any corner cases to consider and will poke around a little bit, but looks good at first glance.

PR #1665, opened 2020-03-04
docs: update contributing instructions

mitchfriedman:
Updating the contributing instructions based on [this discussion ](https://kubernetes.slack.com/archives/CK7VC402X/p1583328082001000) on slack.

No issue.

@edvald

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,7 +31,7 @@ If you haven't already, please [install Homebrew](https://docs.brew.sh/Installat
 
 Other platforms need to roll their own for now (contributions welcome!). Please have a look at the script for OSX to see what's installed.
 
-**Note:** We recommend using Node 10 when developing Garden.
+**Note:** We recommend using Node 12 when developing Garden.
 
 ### Step 4: Bootstrap project
 
@@ -43,6 +43,13 @@ npm run bootstrap # To bootstrap packages
 ```
 
 from the root directory
+
+You will also need to install Node modules in the `garden-service` package. 
+
+```sh
+cd garden-service
+npm install
+```
 
 ## Developing Garden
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -44,6 +44,13 @@ npm run bootstrap # To bootstrap packages
 
 from the root directory
 
+You will also need to install Node modules in the `garden-service` package.
```
eysi09: Shouldn't the `bootstrap` command cover this? If I

```sh
rm -rf garden-service/node_modules && npm run bootstrap
```

...it'll install the `garden-service/node_modules` agin.

mitchfriedman: Doesn't that run in the root?

I found that neither `npm install` or `npm run bootstrap` at the root installed the required dependencies in the `garden-service` package that were needed

eysi09: That's odd, `npm run bootstrap` (from the root) calls `lerna bootstrap` which in turn runs `npm install` in all packages, that is the `dashboard` and `garden-service`

eysi09: Then again: https://github.com/lerna/lerna/issues/1457

eysi09: So yeah, I suspect `lerna` is failing silently along the way. I tried some of the suggestions in the issue but that didn't work. 

Could we maybe just rephrase this as:

"You may need to install the Node modules in the `garden-service` package manually due to https://github.com/lerna/lerna/issues/1457."

mitchfriedman: Perhaps it was just my machine/environment - do you want to try a clean clone of garden and see if it works without this? If it's not necessary happy to remove this!

mitchfriedman: It _could_ have been a versioning thing since I tried to use node 10 from the instructions then realized it was actually using node 12

PR #1667, opened 2020-03-05
fix(cli): fix empty response handling in call cmd

thsig:

Before this fix, the `call` command would throw an exception when calling the service resulted in an empty response.


garden-service/src/commands/call.ts
```diff
@@ -168,34 +169,39 @@ export class CallCommand extends Command<Args> {
...
 
     try {
       res = await req
       entry.setSuccess()
       statusText = getStatusText(res.statusCode)
       log.info(chalk.green(`${res.statusCode} ${statusText}\n`))
+
+      output = res.body
+
+      if (res.headers["content-type"] === "application/json") {
```
eysi09: Couldn't `"content-type"` be capitalised?

garden-service/src/commands/call.ts
```diff
@@ -204,8 +210,9 @@ export class CallCommand extends Command<Args> {
         url,
         response: {
           data: output,
-          headers: res.headers,
-          status: res.statusCode,
+          error,
+          headers: res ? res.headers : null,
+          status: res ? res.statusCode : 204,
```
eysi09: Is it valid to just default to `204` if there's not status code?

thsig: The idea was to use the status code for "No Content" (or "empty response") here.

I.e. the status code for "we received a response, but it was empty".

garden-service/src/commands/call.ts
eysi09: It could still be lower cased.

edvald: I believe the library lowercases all response headers

eysi09: Ok nice, couldn't find it in their docs.

eysi09: @thsig  ☝️

garden-service/src/commands/call.ts
edvald: Seems a little confusing to me actually, I'd rather just allow a null value here to indicate that there was no status code received.

Issue #1659, opened 2020-03-02
Propagate test return value from `garden run test`


mitchfriedman:
### Current Behavior

Running `garden run test <module> <test>` always returns a 0 exit code, regardless of the return value from the actual test.

### Expected behavior

The `garden run test` command should propagate the status code from the test it ran.

### Reproducible example

Run `garden run test <module> <test>` on a failing test.

### Workaround

Running `garden test` will return the correct status code.

### Suggested solution(s)

Propagate the return value from the test run.

### Additional context

n/a

### Your environment

`garden version`
0.11.5

edvald:
This may be a regression, or an oversight. The expected behavior is indeed the intended behavior. As a temporary workaround, you might be able to use the `--output json` flag and parse the result object, but of course that's not ideal in an automated workflow.

10ko:
Assigned to @mitchfriedman since he mentioned he'd like to give it a try at fixing it.
Let us know if you need support!

PR #1669, opened 2020-03-05
fix: return an error if the task fails

mitchfriedman:
This makes sure that garden doesn't return a 0 exit code if a run test fails.

In our CI, we'd like to run `garden run test` so that we get the streamed output from the test, instead of `garden test` where it suppresses the output as it executes. 

If the test fails, garden needs to fail with a non-zero status code so our CI run fails as well.

Fixes https://github.com/garden-io/garden/issues/1659


mitchfriedman:
In trying to get the tests to pass, I've removed changes to the `task` run since they do this work differently AFAICT. I have left this PR as only fixing the `test` run type.

Also, I had a lot of project changes, so instead I just made a new project and use that in my unit tests.

eysi09:
Whoops, wrong Thor. I was referring to @thsig.

thsig:
Thanks for the PR, @mitchfriedman! A few things I missed when reviewing this earlier:
- Let's throw a `RuntimeError` instead of a `CommandError`.
    - This is for consistency with the "task-based" commands (e.g. `TestCommand`, `DevCommand` etc.) which use `handleTaskResults`.
- We should handle errors the same way in each of the `run` commands.
    - I'd suggest writing a helper function (could e.g. be called `handleActionResult`) that checks `result.success` and returns an error similarly to how you did it here.
    - We could then end the `action` methods of each of the `run` commands with
        ```
        return handleActionResult(result)
        ```
        (maybe adding more parameters to the `handleActionResult` function as needed to generate the appropriate error message text).

mitchfriedman:
@thsig Thanks for the review and feedback.

I've addressed your feedback about using a `RuntimeError` instead of `CommandError`.

I've also added a helper function `handleActionResult` to allow all the action methods to have a common way to handling errors. 

There were 2 issues with this:
- I was unable to use this in the `task` action type since it runs the task action much differently from the rest - it uses this [`TaskResult`](https://github.com/garden-io/garden/blob/master/garden-service/src/task-graph.ts#L28-L38) task result type instead of the [`RunResult`](https://github.com/garden-io/garden/blob/master/garden-service/src/types/plugin/base.ts#L77-L89) that the rest of the actions use. This doesn't have the `success` value, and instead has it's own error type built into it. I'm not very familiar with typescript and this codebase generally, so I didn't do any major refactors to change this. 
- I was unable to create tests for the service/module type since it appears the underlying shell spawn always returns a 0 exit code (perhaps because they actually behave differently i.e. fire and forget?). I'm not sure. I'll leave it to you to decide what to do here.

PR #1670, opened 2020-03-05
improvement: allow setting cred helpers in ImagePullSecrets

swist:
This enables setting credHelper key for docker builders (kaniko/dd)
This means we can use docker repositories which have rolling
credentials such as ECR. We run this on ECR + kaniko


swist:
We were not quite sure how you guys integration test, so added a unit test instead. It turned out to be quite tricky because of mocking. We have added a way to DI api into one of the functions that we're using, we're happy to change it to a mock if you can show us the correct way of doing so

swist:
Some of the whitespace changes seem to be a result of running autofixer, so most likely safe to ignore these too

garden-service/src/plugins/kubernetes/init.ts
```diff
@@ -465,76 +464,90 @@ export function getRegistryHostname(config: KubernetesConfig) {
...
-        { secretRef }
-      )
-    }
+          { secretRef }
+        )
+      }
+      return {
+        ...accumulator,
+        auths: { ...accumulator.auths, ...decoded.auths },
+        credHelpers: { ...accumulator.credHelpers, ...decoded.credHelpers },
```
edvald: Could/should we add the `credsStore` key here too? https://github.com/awslabs/amazon-ecr-credential-helper#docker

I see it used in docs here and there, and even though it's not recommended practice (and would only hold one value across all the used imagePullSecrets) I suspect people may try using that.

PR #1671, opened 2020-03-05
chore: Remove legacy output metadata annotations

elliott-davis:

Annotations were being set with the stdout of tasks run with services. Tasks with large amounts of stdout were causing their dependent services to fail if they hit the character limit for k8s annotations.

**Special notes for reviewer**
I was having one test fail locally but it seems unrelated to my change. If anything stands out and it fails in CI, please let me know and I'll get it fixed up.


garden-service/src/runtime-context.ts
```diff
@@ -164,14 +148,7 @@ export async function prepareRuntimeContext({
...
       version: task.module.version.versionString,
     })
-
-    for (const [key, value] of Object.entries(outputs)) {
-      envVars[`GARDEN_TASK_${envName}__OUTPUT_${getEnvVarName(key)}`] = value
-    }
   }
 
-  // Make the full list of dependencies and outputs available as JSON as well
-  result.envVars.GARDEN_DEPENDENCIES = JSON.stringify(result.dependencies)
```
edvald: I'm actually wondering if we should keep this one, but omit the outputs key on each dependency. I've at least used it once myself, and it's only the outputs that we want to get rid of here.

edvald:
Thanks @elliott-davis :) One comment, otherwise looks good!

garden-service/src/runtime-context.ts
elliott-davis: I left it initially but the service I was running crashed on startup because the log output of the task ended up in the `GARDEN_DEPENDENCIES` envvar. It succeeded on subsequent runs where the task didn't execute. I'm happy to leave it in and maybe add some sanitization around what outputs can pass through?

edvald: I'd suggest removing the `outputs` field from every dependency, keep it simple. Something like `JSON.stringify(result.dependencies.map((d) => { ...d, outputs: undefined })`.

elliott-davis: Shows how long I've been out of the javascript game - thanks @swist

garden-service/test/unit/src/runtime-context.ts
```diff
@@ -180,57 +176,5 @@ describe("prepareRuntimeContext", () => {
...
-      },
-      serviceStatuses: {},
-      taskResults: {},
-    })
-
-    expect(runtimeContext.envVars.GARDEN_MODULE_MODULE_B__OUTPUT_MODULE_OUTPUT_B).to.equal("meep")
-    expect(runtimeContext.envVars.GARDEN_MODULE_MODULE_C__OUTPUT_MODULE_OUTPUT_C).to.equal("moop")
-  })
-
-  it("should output the list of dependencies as an env variable", async () => {
```
edvald: This test should be added back and modified.

edvald:
One last comment + a couple of small linting errors. Otherwise 👍 👍

garden-service/test/unit/src/runtime-context.ts
elliott-davis: tenor-141323956

PR #1674, opened 2020-03-06
feat(cli): add experimental fullscreen logger type

edvald:

This adds a new fullscreen logger type, as an alternative to the current
default "fancy" logger type.

This is still experimental and needs to be tested across different
terminals and platforms, but imo adds a solid usability improvement
to watch mode commands. Enable it by setting `--logger-type=fullscreen`
or `GARDEN_LOG_LEVEL=fullscreen`.

It is disabled by default, so if you're satisfied that it doesn't mess
with normal operation, I think it might be good to merge this and use
internally for a while to work out any kinks and get some usability
ideas.


garden-service/src/logger/writers/fullscreen-terminal-writer.ts
```diff
@@ -0,0 +1,514 @@
...
+  private tickSpinner(key: string): string {
+    if (!this.spinners[key]) {
+      this.spinners[key] = elegantSpinner()
+    }
+    return this.spinners[key]()
+  }
+
+  public onGraphChange(entry: LogEntry, logger: Logger): void {
+    if (entry.level === LogLevel.error) {
+      const out = basicRender(entry, logger)
```
eysi09: Why do we basic render errors?

eysi09: The fancy writer also uses does a basic render if `logger.level > LogLevel.info`, don't we need that here as well?

garden-service/src/cli/cli.ts
```diff
@@ -289,6 +289,7 @@ export class GardenCli {
       // the screen the logs are printed.
       const headerLog = logger.placeholder()
       const log = logger.placeholder()
+      logger.info("")
```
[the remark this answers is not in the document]
edvald: Just for aesthetic reasons, otherwise it looks a little odd imo

garden-service/src/logger/writers/fullscreen-terminal-writer.ts
edvald: The former is so that errors are visible in the terminal after the screen is closed.

edvald: Will see about the other thing, good catch

edvald:
Haven't seen those rendering problems myself, is this on tmux?

eysi09:
Huh, thought I had already replied. The screenshot above is from tmux but I've also noticed this on iTerm, kind of hard to reproduce though. The terminal windows was quite large when I noticed it. 

However, I pulled the latest version and was trying to reproduce, and now the process just exits (with code 0) after I run `garden deploy --env remote --logger-type fullscreen`.

That is, I run the command, Garden deploys the stack and then exits.

edvald:
Yeah that's currently the expected behavior without the watch flag. What would you suggest otherwise?

eysi09:
I thought I was running the dev command 🤦‍♂️.

This is the behaviour I would expect.

garden-service/src/logger/writers/fullscreen-terminal-writer.ts
eysi09: Did you address that other thing? Otherwise looks good. I managed to repro that rendering issue once, but not after that (didn't take a screenshot). Basically the "7" from `localhost:9777` was out of place.

edvald: I did, turned out not to be an issue. The rendering thing may be a blessed issue, but can't say for sure until we manage to reproduce it somehow. That's something we can just keep an eye on imo.

PR #1679, opened 2020-03-09
Updated link to detailed installation instructions

Timtech4u:
This PR updates the link to the detailed installation instructions from a broken link.


edvald:
Oops! We just moved it actually. Thanks for the fix @Timtech4u :)

PR #1680, opened 2020-03-09
fix(k8s): issues with private registry auth and kaniko build status

### LINKED ISSUE #1672: Cluster initialise failed with kaniko and external registry
## Bug

When initialising a fresh cluster with garden services and `kaniko` in cluster building and **an external docker registry (ECR),** we got an error messaging saying garden failed to communicate with the docker daemon.

We tried to get around this by changing our in cluster builder to `cluster-docker` to initialise the docker daemon. The docker daemon failed to be brought up because the proxy failed to resolve the in cluster registry.

### Current Behavior

Garden fails to initialise the cluster with this setup.

### Expected behavior

The proxy failing to talk to the registry should be ignored since we don't need the in cluster registry.

### Reproducible example

Our provider configuration looked like:

```
    buildMode: kaniko
    deploymentRegistry:
      hostname: "${var.AWS_ACCOUNT_ID}.dkr.ecr.${var.AWS_REGION}.amazonaws.com"
```

### Workaround

We changed the buildMode to `cluster-docker` and removed the `deploymentRegistry` temporarily. Then we could add these both back in later.

### Suggested solution(s)

The failing proxy connection should be ignored/not checked, and when using kaniko, we shouldn't need a docker daemon running in the cluster. Unsure why garden failed with this.

### Your environment

0.11.5

edvald:

We now use skopeo (https://github.com/containers/skopeo) to get build
status when using Kaniko. Previously the build status was reported as not
ready without an error, even when auth was missing or failed.

Fixes #1672 

cc @mitchfriedman @swist

### DIFF
--- a/images/skopeo/garden.yml
+++ b/images/skopeo/garden.yml
@@ -1,0 +1,6 @@
+kind: Module
+type: container
+name: skopeo
+description: Used by the kubernetes provider for interacting with container registries within a cluster
+image: gardendev/skopeo:1.41.0
+dockerfile: Dockerfile
--- a/garden-service/src/plugins/kubernetes/container/build.ts
+++ b/garden-service/src/plugins/kubernetes/container/build.ts
@@ -20,7 +20,7 @@ import { KubeApi } from "../api"
 import { kubectl } from "../kubectl"
 import { LogEntry } from "../../../logger/log-entry"
 import { KubernetesProvider, ContainerBuildMode, KubernetesPluginContext, KubernetesConfig } from "../config"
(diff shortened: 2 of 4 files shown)


images/skopeo/garden.yml
```diff
@@ -0,0 +1,6 @@
+kind: Module
+type: container
+name: skopeo
+description: The rsync server image, used for hot reloading and in-cluster build syncs
```
thsig: Forgot to update description.

edvald: Fixed 👍

Issue #1647, opened 2020-02-26
[FEATURE]: Ability to pull images from in-cluster registry


eysi09:
### Background / Motivation

The ability to run pull images from the in-cluster registry. This has come up a few times in conversations for users.

### What should the user be able to do?

Run `garden plugins kubernetes pull [module-names]` 

### Why do they want to do this? What problem does it solve?

E.g. when running a project locally but re-using images built in the remote environment. 

From Slack conversation:

> there’s a good chance that at times, pieces of the stack would be run locally. For example, a frontend engineer might want to pull down the backend image (from the cluster repository) and run that backend service locally and do local development on the frontend service (faster rsync of changes, etc)

### Suggested Implementation(s)

Use the `publish` functionality which basically does the same.

mitchfriedman:
This got added here https://github.com/garden-io/garden/pull/1681

PR #1681, opened 2020-03-09
feat: add pull command

mitchfriedman:
Similarly to the `garden publish` command, this adds a `garden pull` command that works for the kubernetes/container module types. 

A use case: a frontend engineer wants to run a local copy of an image but doesn't want to build the image (building takes longer than pulling down an image).

Fixes # https://github.com/garden-io/garden/issues/1647

This doesn't work at the moment for external registries where auth is required that isn't currently supported (i.e. ecr with `credHelpers`). We could perform a 2 step dance, however:

1. Run skopeo in a pod, pull from ECR -> skopeo pod
2. Port forward skopeo, pull from skopeo pod -> local docker

I'm all open to other ideas that are perhaps simpler though.

@edvald @eysi09


swist:
What we've seen work is that with AWS you can set the cred helper on your local dockerconfig. It does mean extra local configuration step, so perhaps opting into using cluster credentials would be preferable?

garden-service/src/plugins/kubernetes/commands/pull-image.ts
```diff
@@ -0,0 +1,215 @@
...
+    throw new RuntimeError(`Failed pulling image for module ${module.name} with image id ${imageId}`, {
+      err,
+      imageId,
+    })
+  }
+}
+
+async function importImage(module: Module, runner: PodRunner, tarName: string, imageId: string, log: LogEntry) {
+  const sourcePath = `/${tarName}`
+  const getOuputCommand = ["cat", sourcePath]
```
swist: any reason you're doing this over `kubectl cp`?

mitchfriedman: Anecdotally from manual testing, this seemed faster than `kubectl cp`. I think `kubectl cp` tars the file then untars it locally, so maybe it's doing some other things in there as well, or it could have been based on time of day. 

This was also a suggestion from @edvald as they use similar patterns elsewhere in the code

edvald:
Sorry @swist I missed your comment there a while back. I'm not sure we can use cluster credentials in a portable fashion. We should however definitely document the extra step for ECR auth.

garden-service/src/plugins/kubernetes/commands/pull-image.ts
edvald: That is indeed more efficient. It can be made a touch more efficient still, I'll comment with some suggestions.

Also, small thing, typo in `getOuputCommand`.

garden-service/src/plugins/kubernetes/commands/pull-image.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  await runner.spawn({
+    command: getOuputCommand,
+    container: "skopeo",
+    ignoreError: false,
+    log,
+    stdout: writeStream,
+  })
+
+  const args = ["import", tmpFile.name, imageId]
```
edvald: I think we should use `docker load` here, right? https://pspdfkit.com/blog/2019/docker-import-export-vs-load-save/

This would also allow us to pipe the output from the runner command directly to the stdin of the `dockerCli` command and skip the temp file.

garden-service/src/plugins/kubernetes/commands/pull-image.ts
```diff
@@ -0,0 +1,215 @@
...
+  const invalidModules = filter(modules, (module) => {
+    return !module.compatibleTypes.includes("container") || !containerHelpers.hasDockerfile(module)
+  })
+
+  if (invalidModules.length > 0) {
+    const invalidModuleNames = map(invalidModules, (module) => {
+      return module.name
+    })
+
+    throw new ParameterError(chalk.red(`Modules ${chalk.white(invalidModuleNames)} is not a container module.`), {
```
edvald: "is not a container module" -> "are not container modules"

garden-service/src/plugins/kubernetes/run.ts
```diff
@@ -539,6 +539,60 @@ export class PodRunner extends PodRunnerParams {
...
+          })
+        )
+      })
+
+      proc.on("error", (err) => {
+        !proc.killed && proc.kill()
+        throw err
+      })
+
+      stdout && proc.stdout?.pipe(stdout)
```
edvald: Could add the same for `stderr`

garden-service/src/plugins/kubernetes/run.ts
```diff
@@ -539,6 +539,60 @@ export class PodRunner extends PodRunnerParams {
...
+    return new Promise((_resolve, reject) => {
+      proc.on("close", (code) => {
+        if (code === 0) {
+          _resolve({
+            moduleName: this.module.name,
+            command,
+            version: this.module.version.versionString,
+            startedAt,
+            completedAt: new Date(),
+            log: "", // TODO: what here.
```
edvald: Not strictly necessary for the PR but this could mimic this bit of code: https://github.com/garden-io/garden/blob/e98dc05565de6b142ad499c9d8d6a2e832bedc34/garden-service/src/util/util.ts#L235

garden-service/src/plugins/kubernetes/commands/pull-image.ts
mitchfriedman: I tried to use `docker load`, but this doesn't seem to load the name of the image/tags. Not sure if skopeo is outputting it in a different format, but even based on that link, 
```docker import – Import the contents from a tarball to create a filesystem image.```, so I think `docker import` is correct.

File: garden-service/test/integ/src/plugins/kubernetes/commands/pull-image.ts [hunk 6]
```diff
@@ -0,0 +1,58 @@
...
+
+      await k8sBuildContainer({
+        ctx,
+        log: garden.log,
+        module,
+      })
+    })
+
+    it("should pull the image", async () => {
+      await pullFromInClusterRegistry(ctx as KubernetesPluginContext, module, garden.log)
```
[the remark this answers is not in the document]
edvald: Also, the `pullFromExternalRegistry` function should in theory work just the same, right? We could also test that, without actually dealing with an external registry. Another approach could be to test that function by having it pull from a public Docker Hub repo.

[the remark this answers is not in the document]
mitchfriedman: Good point - should be able to use the `cluster-docker-remote-registry` and this should work.

PR #1691, opened 2020-03-12
fix(k8s): error in build status check when using kaniko

edvald:

This wasn't adequately tested, so an error slipped through into the release. We should try to release this swiftly if we can. Also took the opportunity to make build error messages slightly nicer.


eysi09:
I can cut a new release once this passes.

mitchfriedman:
Yeah I tested this out, it works 👍

PR #1692, opened 2020-03-13
fix(core): improved circular dependency detection

thsig:

We now detect circular dependencies in `ConfigGraph` after the graph has been augmented, possibly catching more circular dependencies.

Also added circular task dependency detection to `TaskGraph`.

We also use the new `DependencyValidationGraph` class everywhere we need to detect circular dependencies, e.g. when detecting circular plugin/plugin module dependencies.

Some users have run into difficult-to-debug errors that seem to stem from circular config/task dependencies. Hopefully these changes will help quickly diagnose such errors.

### DIFF
--- a/garden-service/src/task-graph.ts
+++ b/garden-service/src/task-graph.ts
@@ -14,11 +14,12 @@ import { every, flatten, intersection, merge, padEnd, union, uniqWith, without }
 import { BaseTask, TaskDefinitionError, TaskType } from "./tasks/base"
 
 import { LogEntry, LogEntryMetadata, TaskLogStatus } from "./logger/log-entry"
-import { toGardenError, GardenBaseError } from "./exceptions"
+import { toGardenError, GardenBaseError, InternalError } from "./exceptions"
 import { Garden } from "./garden"
 import { dedent } from "./util/string"
 import { defer, relationshipClasses, uuidv4 } from "./util/util"
 import { renderError } from "./logger/renderers"
+import { DependencyValidationGraph } from "./util/validate-dependencies"
 
 class TaskGraphError extends GardenBaseError {
   type = "task-graph"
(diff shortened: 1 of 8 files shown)


garden-service/src/task-graph.ts
```diff
@@ -122,9 +131,28 @@ export class TaskGraph {
...
+  ): Promise<TaskNode[]> {
     return Bluebird.map(tasks, async (task) => {
-      const depNodes = await this.nodesWithDependencies(await task.getDependencies(), unlimitedConcurrency)
+      const depTasks = await task.getDependencies()
+
+      // Detect circular dependencies
+      validationGraph.addNode(task.getKey())
+      for (const depTask of depTasks) {
+        validationGraph.addNode(depTask.getKey())
+        validationGraph.addDependency(task.getKey(), depTask.getKey())
```
edvald: Don't we need to do this recursively? Seems like this flow wouldn't catch a lot of cases. Which begs the question whether it is at all possible to make use of a separate graph structure when we're asynchronously resolving the dependencies. We'd just get stuck while figuring out what the dependencies are. Unless I'm over/under-thinking this somehow...

thsig: The idea is that we incrementally try to detect circular dependencies at each step of the recursion (the next level down gets checked in on line 155 below) so as to detect cycles without having to materialize/compute a full dependency graph (which we can't do with this control flow, since it would loop indefinitely).

It is true that this approach only catches the "first cycle detected" as the dependencies are computed, but I decided that was a reasonable compromise that doesn't over-complicate things.

Or am I missing something here?

edvald: I sort of misread the flow I think, makes sense to me now. Cool, then LGTM!

PR #1693, opened 2020-03-13
Fix/doc

Timtech4u:
Fixed link to Stack Graph

@edvald the href to Stack Graph on the homepage also needs to be updated, see below:

Screenshot from 2020-03-13 11-27-05


Timtech4u:
Closing this with respect to #1700

PR #1702, opened 2020-03-16
Fix plugin commands

edvald:

A couple of issues reported related to plugin commands. See individual commits.


mitchfriedman:
Looks like this should address the issues we're seeing, but I can test it out tomorrow 👍

PR #1704, opened 2020-03-17
fix(cli): configure provider/env correctly in `garden plugins` command

edvald:

I missed a beat in #1702, this should do the trick

 Please test this quickly on your end before merging.


edvald:
That is intentional, since plugin commands are likely to use stdout/stdin and have some expectation that those work as normal.

PR #1705, opened 2020-03-18
docs(examples): add build dependencies example

eysi09:

Adds a dedicated "build dependencies" example. This gets asked about a lot in our community channel and so far this pattern has only been used implicitly in other example without proper explanation.

It'll also be referenced from the FAQ doc we're working on. 

None.

None.

### DIFF
--- a/examples/build-dependencies/README.md
+++ b/examples/build-dependencies/README.md
@@ -1,0 +1,75 @@
+# Build Dependencies
+
+This example project demonstrates how to use build dependencies to include files and directories from outside a module's root with its build context. For example, some .NET and Java projects that are split up into multiple modules need a shared configuration file at build time for each module.
+
+Another common use case is to use build dependencies to include shared libraries with multiple modules. Our [`openfaas` example project](../openfaas) demonstrates this same pattern with an NPM package.
+
+To achieve this, we:
+
+1. wrap the shared files/directories in an `exec` module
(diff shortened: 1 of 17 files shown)


examples/build-dependencies/README.md
```diff
@@ -0,0 +1,75 @@
...
+  dependencies:
+    - name: shared-config
+      copy:
+        - source: "config.env"
+          target: "config/"
+```
+
+This tells Garden to first "build" the `shared-config` module, then to copy the contents from the `source` path *in the `shared-config` module* to the `target` path *in the `frontend` module*.
+
+Note that this takes place inside the `.garden/build` directory. If you for example run `garden build frontend` and then look at the folder structure of the `.garden/build` directory, you'll see:
```
thsig: "If you for example run" -> "For example, if you run"

examples/build-dependencies/README.md
```diff
@@ -0,0 +1,75 @@
...
+├── ...
+```
+
+This is the build context Garden will use when building the `frontend` module.
+
+In this example there's no "build" step for the `exec` module, but you can add it via the `build` field. Check out our [`local-exec` project](../local-exec/backend/garden.yml) for an example of this.
+
+## Usage
+
+Run `garden deploy` to deploy the project. You can verify it works by running:
```
thsig: "verify it works" -> "verify that it works"

PR #1711, opened 2020-03-18
chore: Bump kaniko version to 0.19.0

swist:
This fixes a few hairy things around dockerfiles not
being handled isomorphically to docker daemon.

We have had a few issues since 0.17 release and 
had figured out a few workarounds, but this claims
to fix those issues proper. Means our future modules
will require less oversight


swist:
Reason for this PR is that we've run into this issue:

https://github.com/GoogleContainerTools/kaniko/issues/1076

PR #1713, opened 2020-03-18
improvement(core): make module resolution faster to reduce startup time

edvald:

This makes module template string resolution happen more explicitly
in dependency order, and avoids repeated resolution of the same module
while resolving template strings. Avoids issues with large/complex
projects, such as slow startups.

This is a fairly substantial refactor. Please look for any potentially missing test cases, and test manually if you can think of potential issues.


garden-service/test/unit/src/task-graph.ts
```diff
@@ -542,13 +542,13 @@ describe("task-graph", () => {
       const t2 = new TestTask(garden, "a", false, { uid: "2", versionString: "2", callback: repeatedCallback("2") })
       const t3 = new TestTask(garden, "a", false, { uid: "3", versionString: "3", callback: repeatedCallback("3") })
 
-      const firstProcess = graph.process([t1])
+      const firstProcess = await graph.process([t1])
```
thsig: Is there a reason we're adding `await`s in these test cases? They were meant to test concurrent calls to `graph.process`.

garden-service/src/tasks/resolve-module.ts
```diff
@@ -0,0 +1,218 @@
...
+          chalk.red(
+            `Could not find build dependency ${chalk.white(name)}, configured in module ${chalk.white(
+              this.moduleConfig.name
+            )}`
+          ),
+          { moduleConfig }
+        )
+      }
+
+      if (d[2] === "version") {
```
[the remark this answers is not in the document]
edvald: I agree, I'll add a comment

garden-service/test/unit/src/task-graph.ts
edvald: Oops, that's a refactoring mistake, I'll revert that.

edvald:
@thsig comments addressed, should be good to go

eysi09:
@edvald, we need to resolve conflicts before merging.

edvald:
I also got some feedback on this on the community Slack. I'm gonna test a user project with this branch, it seems like there is still something that needs to be fixed.

edvald:
@thsig this needs additional review, but should be good to go (after rebase + squash). It's been verified by @swist to solve their problem.

PR #1717, opened 2020-03-19
feat: only use socat with incluster registry

mitchfriedman:
This just refactors the container build process for kubernetes slightly. We don't need the socat container when using kaniko + external registries, and if we use that socat container, it just spams error logs consistently. 

This removes the socat container when using the kaniko build mode + external registries.

No issue.

@edvald


edvald:
Merged, but just noticed a super minor thing that I missed: For the commit message `improvement(k8s): ...` would be a better prefix, since this doesn't exactly constitute a new feature. No biggie, just mentioning for future reference .)

mitchfriedman:
Right, good point - will keep it in mind 👍

PR #1718, opened 2020-03-19
improvement(k8s): add default tolerations to system services

### LINKED ISSUE #1600: [FEATURE]: Add nodeSelector and tolerations to system services
## Feature Request

1. Add default tolerations to Garden system services.
2. Add a`systemServiceNodeSelector` field.

### Background / Motivation

It's useful to be able to schedule system services on non-ephemeral nodes. Users with auto-scaling clusters don't want to loose the system cache when it scales down.

Set `nodeSelector` on Garden system services so that you can assign them to specific nodes (and thereby not on e.g. nodes that scale to 0) via a `systemServiceNodeSelector` field.

Set taints on the particular nodes so that only system services get scheduled there. E.g.: `kubectl taint nodes <node-name> garden-system=true:NoSchedule`.

For example enable using preemptible nodes on GKE.

See above.

It can save money so this is at least a 🌵 .

eysi09:

Adds default tolerations and expose the `nodeSelector` field on some Garden system services, namely:

* `build-sync`
* `docker-daemon`
* `docker-registry`
* `ingress-controller`
* `nfs-provisioner`

The tolerations have the following spec:

```yaml
tolerations:
  - key: "garden-system"
    operator: "Equal"
    value: "true"
    effect: "NoSchedule"
```

Fixes #1600 

We use the `conftest` plugin to test whether system services have the right policies. Dog fooding ftw. 

**Note that this does not apply to `cert-manager` and `default-backend`. **

### DIFF
--- a/garden-service/src/plugins/kubernetes/init.ts
+++ b/garden-service/src/plugins/kubernetes/init.ts
@@ -34,7 +34,7 @@ import { ConfigurationError } from "../../exceptions"
 import Bluebird from "bluebird"
 import { readSecret } from "./secrets"
 import { dockerAuthSecretName, dockerAuthSecretKey } from "./constants"
-import { V1Secret } from "@kubernetes/client-node"
+import { V1Secret, V1Toleration } from "@kubernetes/client-node"
 import { KubernetesResource } from "./types"
 import { compareDeployedResources } from "./status/status"
 import { PrimitiveMap } from "../../config/common"
@@ -418,6 +418,15 @@ export function getKubernetesSystemVariables(config: KubernetesConfig) {
   const nfsStorageClass = getNfsStorageClass(config)
   const syncStorageClass = config.storage.sync.storageClass || nfsStorageClass
   const systemNamespace = config.gardenSystemNamespace
+  const systemTolerations: V1Toleration[] = [
+    {
+      key: "garden-system",
+      operator: "Equal",
+      value: "true",
+      effect: "NoSchedule",
+    },
+  ]
+  const registryProxyTolerations = config.registryProxyTolerations || systemTolerations
 
   return {
     "namespace": systemNamespace,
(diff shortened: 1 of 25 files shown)


edvald:
@eysi09 this could be improved using #1725

eysi09:
Nice! I'll follow up with a commit for enabling `nodeSelector`.

garden-service/src/plugins/kubernetes/init.ts
```diff
@@ -454,9 +463,9 @@ export function getKubernetesSystemVariables(config: KubernetesConfig) {
     "sync-storage-class": syncStorageClass,
     "sync-volume-name": `garden-sync-${syncStorageClass}`,
 
-    // Stringifying the tolerations since variable values should be primitives.
-    // Helm handles the decoding automatically.
-    "registry-proxy-tolerations": JSON.stringify(config.registryProxyTolerations),
+    "registry-proxy-tolerations": <PrimitiveMap[]>registryProxyTolerations,
+    "system-tolerations": <PrimitiveMap[]>systemTolerations,
+    "systeme-node-selector": config.systemNodeSelector,
```
edvald: Typo: "systeme"

garden-service/test/integ/src/plugins/kubernetes/system.ts
```diff
@@ -0,0 +1,58 @@
...
+    const root = getDataDir("test-projects", "container")
+    garden = await makeTestGarden(root)
+    provider = (await garden.resolveProvider("local-kubernetes")) as Provider<KubernetesConfig>
+  })
+
+  after(async () => {
+    await garden.close()
+  })
+
+  it("should use conftest to check whether system services have a valid config", async () => {
```
edvald: Would it make sense to have a failure case here? Seems likely we could accidentally not apply the policies after some refactor, and we should catch that.

edvald:
@eysi09 I see a linting error and a valid test error on Windows. Good to go once all the green checkmarks are in.

eysi09:
It looks the windows error is because the config expects a POSIX path but on Windows the actual path is resolved to a Windows style path (in this case `..\policy`). 

I added a separate commit that addresses this by using the [`slash`](https://github.com/sindresorhus/slash) library to convert the path back to POSIX. Could you quickly review?

PR #1725, opened 2020-03-23
feat(config): allow any objects and arrays in project variables

edvald:

This lifts a limitation that is no longer necessary, and adds quite a
bit of flexibility to our variables mechanism. You can now put any
object or array in your project variables (both global and
environment-specific), and template strings can also output any object
or array.

See the updated docs on variables and templating for more details, as
well as the tests.


garden-service/src/template-string.ts
```diff
@@ -49,7 +49,7 @@ export async function resolveTemplateString(
   string: string,
   context: ConfigContext,
   opts: ContextResolveOpts = {}
-): Promise<Primitive | undefined> {
+): Promise<any> {
```
thsig: Could `any` be replaced with `Primitive | DeepPrimitiveMap | Primitive[] | DeepPrimitiveMap[] | undefined` or something of the sort? Or do we need the flexibility of the `any` type here?

garden-service/src/template-string.ts
```diff
@@ -85,7 +85,7 @@ export async function resolveTemplateString(
     }
 
     // Use value directly if there is only one (or no) value in the output.
-    let resolved: Primitive | undefined = outputs[0]?.resolved
+    let resolved: any = outputs[0]?.resolved
```
thsig: (Same here, just a thought)

garden-service/src/template-string.ts
edvald: Well, that's practically equivalent. It's pretty much anything except symbols and such :P So I don't think that really matters.

PR #1727, opened 2020-03-24
feat(platform): add client login to CLI

thsig:
* Added `login` and `logout` commands (not included in the docs for now).

* Auth logic for reading and storing client auth tokens (we use the local SQLite DB for this).

* Auth logic for visiting the platform's login page, and for receiving redirects from that page containing a valid client token.

A couple of open questions:

1. Should we use the presence of a `sessionId` in the options passed to `Garden.factory` as the indicator of whether the login flow should be engaged (this is the current approach)? Or should login be an "outside" operation, performed in the CLI and the login command?

2. Should we hardcode a default value for the platform URL, or keep using the `GARDEN_CLOUD` env var for now? The current setup may be fine for the time being, just wanted to flag this.

### DIFF
--- a/garden-service/src/cli/cli.ts
+++ b/garden-service/src/cli/cli.ts
@@ -13,7 +13,7 @@ import { resolve, join } from "path"
 import { safeDump } from "js-yaml"
 import { coreCommands } from "../commands/commands"
 import { DeepPrimitiveMap } from "../config/common"
-import { shutdown, sleep, getPackageVersion } from "../util/util"
+import { shutdown, sleep, getPackageVersion, uuidv4 } from "../util/util"
 import { deline } from "../util/string"
 import {
   BooleanParameter,
@@ -302,6 +302,8 @@ export class GardenCli {
       logger.info("")
       const footerLog = logger.placeholder()
 
+      const sessionId = uuidv4()
+
       const contextOpts: GardenOpts = {
         commandInfo: {
           name: command.getFullName(),
@@ -310,6 +312,7 @@ export class GardenCli {
         },
         environmentName,
         log,
+        sessionId,
       }
 
       let garden: Garden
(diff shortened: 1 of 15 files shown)


garden-service/src/cli/cli.ts
```diff
@@ -302,6 +302,8 @@ export class GardenCli {
       logger.info("")
       const footerLog = logger.placeholder()
 
+      const sessionId = uuidv4()
```
thsig: Note: This is used by the event & log streaming logic, but since it's also used here in the control flow in `Garden.factory`, I've included it in this PR as well.

garden-service/src/commands/login.ts
```diff
@@ -0,0 +1,32 @@
...
+  name = "login"
+  help = "Log in to Garden Cloud."
+  hidden = true
+
+  description = dedent`
+    Logs you in to Garden Cloud. Subsequent commands will have access to platform features.
+  `
+
+  async action({ garden, log, headerLog }: CommandParams): Promise<CommandResult> {
+    printHeader(headerLog, "Login", "lightning_cloud")
```
edvald: 🌩 seems like an odd choice :P Maybe just a cloud?

thsig: Haha. Thought the lightning was cute as an indicator of "this cloud's connected". But yeah, let's go with a simple cloud.

garden-service/src/platform/auth.ts
```diff
@@ -0,0 +1,181 @@
...
+    this.log = log.placeholder()
+  }
+
+  async start() {
+    if (this.app) {
+      return
+    }
+
+    if (!this.port) {
+      this.port = await getPort({ port: DEFAULT_AUTH_REDIRECT_PORT })
```
edvald: Is there a reason not to just randomize this in every case?

thsig: No, good point.

PR #1728, opened 2020-03-24
fix(k8s): if the output is json, stringify it

ThePixelDeveloper:

I spotted this issue with a task that ran the AWS CLI to create DynamoDB tables. The output of the command is in JSON. When the output is JSON it comes into garden as an object, which doesn't have a trim method. You can see this in [run.ts](https://github.com/garden-io/garden/blob/master/garden-service/src/plugins/kubernetes/run.ts#L341).

You can recreate this issue by having a task output JSON, ala.

``` bash
#!/usr/bin/env bash

# Forced for the sake of the test.
echo '{"hello": "world"}'
exit 0
```
_setup.sh_

``` yaml
kind: Module
name: database-tables
description: Create database tables
type: container
tasks:
 - name: db-migrate-default
   command: ["bash", "/project/setup.sh"]
   dependencies:
     - database-default
```
_garden.yml_

and then a docker file that just copies over the setup.sh file.

Internal issue at the company I work for

### DIFF
--- a/garden-service/src/plugins/kubernetes/status/pod.ts
+++ b/garden-service/src/plugins/kubernetes/status/pod.ts
@@ -113,6 +113,10 @@ export async function getPodLogs({
       }
     }
 
+    if (typeof(log) === "object") {
+      log = JSON.stringify(log)
+    }
+
     // the API returns undefined if no logs have been output, for some reason
     return { containerName, log: log || "" }
   })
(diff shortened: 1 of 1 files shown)


edvald:
Thanks @ThePixelDeveloper! That's odd. It appears the Kubernetes API library does something implicitly there. I'm wondering why and how exactly though. Can you confirm that the log looks exactly as expected, with your fix?

ThePixelDeveloper:
@edvald Yeah I've confirmed this from stepping through the code in Google Chrome. I've the task manually with a higher log level and confirmed it all looks good 👍

PR #1729, opened 2020-03-24
fix(k8s): error when getting build status from microk8s cluster

edvald:

See Slack thread: https://kubernetes.slack.com/archives/CKM7CP8P9/p1585068651025100


.circleci/config.yml
```diff
@@ -447,6 +447,80 @@ jobs:
...
+
+            # Set up kubeconfig and environment
+            mkdir -p /home/circleci/.kube
+            sudo sh -c 'microk8s.kubectl config view --raw > /home/circleci/.kube/config'
+            sudo chown -R circleci:circleci /home/circleci/.kube
+            echo 'export PATH=/snap/bin:$PATH' >> $BASH_ENV
+            echo "********* env prepped! *********"
+      - run:
+          name: Update Node.js
+          command: |
```
eysi09: This takes me back—to that day I spent six hours getting this to work.

edvald: Yeah. I feel ya.

eysi09: Perhaps we should create a re-usable command for this and the steps below. Since it's used here and in for the minikube tests.

garden-service/test/integ/src/plugins/kubernetes/container/build.ts
```diff
@@ -57,12 +57,12 @@ describe("kubernetes build flow", () => {
     })
   })
 
-  context("local-remote-registry mode", () => {
+  grouped("remote-only").context("local-remote-registry mode", () => {
```
eysi09: Do you need this if you're also using `grouped` below?

edvald:
:+1: resolved. I've tested microk8s myself, using multipass, so with that and the tests I feel pretty confident. I'll try and think of any microk8s-specific docs that may be necessary.

PR #1733, opened 2020-03-26
Fix run command exit codes

eysi09:

The first commit fixes an issue with `garden run test` returning exit code 0 if the test fails and artifacts are copied.

As I was working on that, I noticed that `garden run task` would return the wrong exit code for local tasks, and some other inconsistencies in how the outputs were printed. The second commit addresses that.

This came up in a conversation on the community Slack.

I turned the `loggerType` static property into a function so that commands can dynamically decide what the logger type should be. For example, commands with `--interactive=true` should use the basic logger. 

Here are screenshots from how the outputs look now:

Screenshot 2020-03-26 at 14 41 21

Screenshot 2020-03-26 at 14 41 46

Screenshot 2020-03-26 at 14 42 17

### DIFF
--- a/docs/misc/faq.md
+++ b/docs/misc/faq.md
@@ -167,6 +167,16 @@ garden run task <task-name>
 
 This will run the task even if the result is cached.
 
+### What's the difference between `garden test` and `garden run test`
+
+The `garden test` command can run all your tests, or a subset of your tests, and has a `--watch` flag. It won't re-run tests that are cached unless the `--force` flag is set and it won't print the output unless the test fails. [See here](https://docs.garden.io/reference/commands#garden-test) for the synopsis and examples.
+
+The `garden run test` command runs **a single test in interactive mode** regardless of whether or not it's cached. Interactive mode means that the output is streamed to the screen immediately and you can interact with it if applicable.
+
+Note that due to a [known limitation](https://github.com/garden-io/garden/issues/1739), Garden can't copy artifacts for tests in interactive mode. You can disable it by setting `--interactive false`. [See here](https://docs.garden.io/reference/commands#garden-run-test) for the full synopsis.
+
+We plan on making `--interactive=false` the default with our next major release.
+
 ## Secrets
 
 ### How do I pass secrets to container modules?
--- a/garden-service/src/commands/base.ts
+++ b/garden-service/src/commands/base.ts
@@ -18,10 +18,11 @@ import { GardenError, InternalError, ParameterError, RuntimeError } from "../exc
 import { Garden } from "../garden"
 import { LogEntry } from "../logger/log-entry"
 import { LoggerType } from "../logger/logger"
-import { printFooter } from "../logger/util"
+import { printFooter, renderMessageWithDivider } from "../logger/util"
 import { ProcessResults } from "../process"
-import { TaskResults } from "../task-graph"
+import { TaskResults, TaskResult } from "../task-graph"
 import { RunResult } from "../types/plugin/base"
+import { capitalize } from "lodash"
 
 export interface ParameterConstructor<T> {
   help: string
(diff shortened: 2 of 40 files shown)


garden-service/src/plugins/kubernetes/run.ts
```diff
@@ -245,7 +245,7 @@ export async function runAndCopy({
         result = await runner.exec({
           // Pipe the output from the command to the /tmp/output pipe, including stderr. Some shell voodoo happening
```
eysi09: This comment still applies 😛

garden-service/src/commands/base.ts
```diff
@@ -372,21 +376,80 @@ export abstract class Command<T extends Parameters = {}, U extends Parameters =
...
+  success,
+  actionDescription,
+}: {
+  log: LogEntry
+  result: string
+  success: boolean
+  actionDescription: string
+}) {
+  const prefix = success
+    ? `${capitalize(actionDescription)} result:`
```
edvald: Wondering about the phrasing here. "Run task result" and "Run test result" feel a little odd to me. How about using "output" instead of "result", and just using "Task" and "Test" for the action description? So it'd be simply "Task output" and "Test output". That feels more natural to me.

edvald: Which btw also works for the error variant, e.g. "Test failed with error".

docs/misc/faq.md
```diff
@@ -167,6 +167,16 @@ garden run task <task-name>
 
 This will run the task even if the result is cached.
 
+### What's the difference between `garden test` and `garden run test`
+
+The `garden test` command can run all your tests, or a subset of your tests, and has a `--watch` flag. It won't re-run tests that are cached unless the `--force` flag is set and it won't print the output unless the test fails. [See here](https://docs.garden.io/reference/commands#garden-test) for the synopsis and examples.
+
+The `garden run test` command runs **a single test in interactive mode** regardless of whether or not it's cached. Interactive mode means that the output is streamed to the screen immediately and you can interact with it if applicable.
+
+Note that due to a current limitation, Garden can't copy artifacts for tests in interactive mode. You can disable it by setting `--interactive false`. [See here](https://docs.garden.io/reference/commands#garden-run-test) for the full synopsis.
```
edvald: Is there an open issue for that? Good to link that here.

garden-service/src/commands/run/task.ts
```diff
@@ -78,14 +77,6 @@ export class RunTaskCommand extends Command<Args, Opts> {
...
-    if (result && !result.error) {
-      log.info("")
-      // TODO: The command will need to be updated to stream logs: see https://github.com/garden-io/garden/issues/630.
-      // It's ok with the current providers but the shape might change in the future.
-      log.info(chalk.white(result.output.log.trim()))
-      printFooter(footerLog)
-    }
-
-    return { result }
+    return handleTaskResult({ log, actionDescription: "run task", result: result! })
```
edvald: Should this be `handleRunResult`?

garden-service/src/commands/run/test.ts
```diff
@@ -58,13 +61,20 @@ export class RunTestCommand extends Command<Args, Opts> {
...
 
   arguments = runArgs
   options = runOpts
 
+  getLoggerType({ opts }): LoggerType {
+    if (opts.interactive) {
+      return "basic"
+    }
+    return "fancy"
+  }
```
edvald: Is this preferable to stopping the logger before running the interactive part?

docs/misc/faq.md
eysi09: No but we should add it.

garden-service/src/commands/run/task.ts
eysi09: It should not. This `handleTaskResult` function handles, as the name suggests, a result of type `TaskResult`, from a `garden.processTasks` call. The `handleRunResult` handles results of type `RunResult`, from calling something like `garden.testModule`. 

I opted for two functions since the types, and especially return types, are different and instead enforcing correctness at the test level, should they drift

edvald: Gotcha. Maybe add comments on top of the handler functions to explain the difference?

eysi09: Good point!

garden-service/src/commands/run/test.ts
eysi09: It felt more robust. Didn't really try the alternative.

eysi09: Actually it seems to work just fine.

Issue #1708, opened 2020-03-18
Setting the Kubernetes context in local development


eysi09:
When developing locally, the `local-kubernetes` plugin will attempt to guess how the user is running Kubernetes locally by looking at the available contexts. As of writing, the supported contexts are `docker-for-desktop`, `minikube`, and `microk8s`. 

This however is very brittle. For example, users can re-name these contexts (at least in some scenarios) and Garden could also "guess" the wrong context if it finds more then one of these.

For context (pun not intended), I noticed this when trying out Kind for local development. You can easily have multiple Kind clusters with different contexts so it's pretty hard to guess which one to use.

You can get around this by setting the `context` field in the `local-kubernetes` provider config but that of course doesn't work well for teams where people use different clusters for local development.

So I'm wondering how to best address this. The `context` field isn't a framework level primitive but just a configuration option for a plugin so it doesn't really make sense to add it as a command option or include it in something like a user level configuration. Unless of course we design those in such a way that they can overwrite plugin configurations which I wouldn't think is a good idea. 

cc @solomonope

PR #1751, opened 2020-03-30
fix(k8s): fix some issues with KinD and add integ test

edvald:

Found a few issues when setting up and using KinD. For one, installing nginx was not working, and some things needed explanation in the docs. 

Plus I've been wanting to set up an integ test for a while, having a crack at that now.

### DIFF
--- a/docs/guides/local-kubernetes.md
+++ b/docs/guides/local-kubernetes.md
@@ -51,6 +51,14 @@ And then adding this to your `.bashrc`/`.zshrc`:
 export KUBECONFIG=$HOME/.kube/microk8s.config:${KUBECONFIG:-$HOME/.kube/config}
 ```
 
+You also need to ensure microk8s commands can be run by the user that's running Garden, so that Garden can get its status and enable required extensions if necessary. To do this, add your user to the `microk8s` group:
+
+```sh
+sudo usermod -a -G microk8s $USER   # or replace $USER with the desired user, if it's not the current user
+```
+
+Note that in-cluster building is currently not supported with microk8s clusters.
+
 ### Minikube
 
 For Minikube installation instructions, please see the [official guide](https://github.com/kubernetes/minikube#installation).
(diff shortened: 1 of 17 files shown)


garden-service/src/plugins/kubernetes/local/config.ts
```diff
@@ -18,12 +18,15 @@ import { getNfsStorageClass } from "../init"
 import { isClusterKind } from "./kind"
 
 // TODO: split this into separate plugins to handle Docker for Mac and Minikube
-
 // note: this is in order of preference, in case neither is set as the current kubectl context
 // and none is explicitly configured in the garden.yml
-const supportedContexts = ["docker-for-desktop", "microk8s", "minikube"]
+const supportedContexts = ["docker-for-desktop", "docker-desktop", "microk8s", "minikube", "kind-kind"]
```
eysi09: Is there a better way around this? The order of preference feels very arbitrary. At the very least it feels like we should log it at the `info` level if there are multiple supported contexts found and Garden picks the first one. You can e.g. have multiple `kind` contexts. 

See also this issue: https://github.com/garden-io/garden/issues/1708

This issue acknowledges that it's not easy to get around this so perhaps logging and documenting this is our best bet for now.

garden-service/src/plugins/kubernetes/local/config.ts
```diff
@@ -18,12 +18,15 @@ import { getNfsStorageClass } from "../init"
...
 const nginxServices = ["ingress-controller", "default-backend"]
 
+function isSupportedContext(context: string) {
+  return supportedContexts.includes(context) || context.startsWith("kind-")
```
eysi09: Same for this one. Users can easily rename their contexts and it isn't super obvious that this is what Garden is doing. Perhaps we could include this in the docs for these tools. For example that Garden expects Kind contexts to begin with `kind-` and that it'll pick the first one it finds.

garden-service/src/plugins/kubernetes/local/config.ts
eysi09: One simple idea for teams with different local clusters (and that can't depend on the order of contexts above) would be to just use local env vars: `context: ${local.env.LOCAL_CONTEXT}`  or similar. Perhaps we could document that as a workaround.

garden-service/src/plugins/kubernetes/local/config.ts
```diff
@@ -18,12 +18,15 @@ import { getNfsStorageClass } from "../init"
 import { isClusterKind } from "./kind"
```
eysi09: Not part of the PR per se but it would be nice to refactor the `configureProvider` function at some point. A lot of ifs and buts.

garden-service/src/plugins/kubernetes/local/config.ts
```diff
@@ -90,15 +93,21 @@ export async function configureProvider(params: ConfigureProviderParams<LocalKub
...
   }
 
   if (await isClusterKind(provider, log)) {
     config.clusterType = "kind"
-  }
-  if (config.context === "minikube") {
+
+    if (config.setupIngressController === "nginx") {
+      log.debug("Using nginx-kind service for ingress")
+      remove(_systemServices, (s) => nginxServices.includes(s))
```
eysi09: Would it make sense to refactor this into its own function since it's used in three places?

edvald: I don't think it matters tbh.

garden-service/src/plugins/kubernetes/local/config.ts
edvald: Yes, I actually put in a TODO the code below. Thing is, any behavior change we make here is essentially a breaking change. So all I did here was to augment the behavior to explicitly account for kind, with a mind to address this better in a breaking release.

Agreed on documentation though, perhaps coupled with making the `clusterType` config variable public and documented.

eysi09:
LGTM. Notice that the integ test is failing and the PR is still flagged as WIP.

edvald:
@eysi09 this is now good to go.

docs/guides/local-kubernetes.md
```diff
@@ -68,6 +76,39 @@ Once Minikube and the appropriate driver for your OS are installed, you can star
 minikube start --vm-driver=<your vm driver>  # e.g. hyperkit on macOS
 ```
 
+### kind
+
+For kind installation instruction, see the [official docs](https://kind.sigs.k8s.io/docs/user/quick-start/).
```
thsig, line 81: "instruction" -> "instructions"

thsig (changes requested):
Noticed a typo there, otherwise I think this is good to go.

edvald:
Rebased and updated @thsig

thsig (approved):

PR #1752, opened 2020-03-30
feat(core): provide secrets to template strings

thsig:
For projects connected with Garden Cloud, any secrets for the project in the given environment can now be used in template strings.


examples/build-dependencies/backend/main.go
```diff
@@ -3,10 +3,11 @@ package main
 import (
 	"fmt"
 	"net/http"
+	"os"
 )
 
 func handler(w http.ResponseWriter, r *http.Request) {
-	fmt.Fprint(w, "Hello from Go!")
+	fmt.Fprint(w, os.Getenv("THE_SEC"))
```
eysi09, line 10: Should this be here?

examples/demo-project/backend/garden.yml
```diff
@@ -4,6 +4,8 @@ description: Backend service container
 type: container
 services:
   - name: backend
+    env:
```
eysi09, line 7: Same as above.

garden-service/src/secrets/gardencloud/getSecret.ts
```diff
@@ -0,0 +1,30 @@
+/*
```
eysi09, line 1: File name is camel cased whereas we usually kebab case file names.

garden-service/package.json
```diff
@@ -108,6 +108,7 @@
     "request": "^2.88.2",
     "request-promise": "^4.2.5",
     "slash": "^3.0.0",
+    "request-promise-native": "^1.0.8",
```
eysi09, line 111: Is this used anywhere?

garden-service/src/garden.ts
```diff
@@ -294,6 +299,13 @@ export class Garden {
         throw new InternalError(errMsg, {})
       } else {
         await login(platformUrl, log)
+        secrets = await getSecrets({
```
eysi09, line 302: What happens if `secrets` don't contain all the secrets the config expects. Say you have something like:

```yaml
services:
# ...
  env:
    API_KEY: ${secrets.apiKey}
```

and `secrets.apiKey` is undefined. 

Does it just deploy the service with an empty `API_KEY`? Or is this handled somewhere?

solomonope: I think this is a valid question, from a fail early point of view, I think garden should fail that secrets were not resolved. I think that is what is currently happening

thsig: An error will be thrown if the key is not defined (see https://github.com/garden-io/garden/blob/secrets-in-template-strings/garden-service/test/unit/src/config/config-context.ts#L343).

This leans on our generic error handling for undefined keys in template strings.

We are not, however, validating that the value is truthy (so a secret with a `null`/`undefined` value for a valid/present key would not result in an error).

garden-service/src/cloud/secrets/garden-cloud/get-secret.ts
```diff
@@ -0,0 +1,31 @@
...
+  try {
+    const url = `${platformUrl}/secrets/project/name/${projectConfig.name}/env/${environmentName}`
+    const headers = { "x-access-auth-token": clientAuthToken }
+    const res = await got(url, { headers }).json<GotResponse<any>>()
+    if (res && res["status"] === "success") {
+      return res["data"]
+    }
+    return {}
+  } catch (err) {
+    log.error("An error occurred while fetching secrets for the project.")
```
eysi09, line 28: Feels like this should be a clearer error message. Could this e.g. be an auth error? In which case we should instruct the user to log in.

garden-service/src/garden.ts
eysi09: All right. Wondering how we should in general handle platforrm related errors. For example if the secret is missing from the backend you want a better error than just: "Failed resolving template string..." (or whatever it is).

thsig: I've added some error handling around empty secret values (in a TBS commit).

garden-service/src/cloud/secrets/garden-cloud/get-secret.ts
thsig: Shall we address this in a separate PR?

garden-service/src/template-string.ts
```diff
@@ -131,6 +131,10 @@ export function getModuleTemplateReferences<T extends object>(obj: T) {
   return refs.filter((ref) => ref[0] === "modules" && ref.length > 1)
 }
 
+export function detectMissingSecrets<T extends object>(obj: T, secrets: PrimitiveMap) {
```
eysi09, line 134: What's this for?

garden-service/src/garden.ts
```diff
@@ -663,6 +663,10 @@ export class Garden {
 
     this.log.silly(`Resolving module configs`)
 
+    if (Object.keys(this.secrets).length > 0) {
+      // Gather secret references from project config and raw module configs
```
eysi09, line 667: Something missing here?

garden-service/src/template-string.ts
thsig: Ah, my bad I had some uncommited on a separate branch. Will update tomorrow.

PR #1756, opened 2020-03-31
fix(core): fix watch task logic for sourceModules

thsig:

Before this fix, if a service with a `sourceModule` was deployed with hot reloading enabled, a build task for its source module would be added when its sources change (in addition to the hot reload task).

This was not the desired behavior, and would (e.g. for `helm` modules using a `container` module as a `sourceModule`) result in unnecessary rebuilds of the source module.

This issue was reported by a user on our community Slack.

### DIFF
--- a/garden-service/src/tasks/helpers.ts
+++ b/garden-service/src/tasks/helpers.ts
@@ -7,7 +7,7 @@
  */
 
 import Bluebird from "bluebird"
-import { intersection, flatten, uniqBy } from "lodash"
+import { includes, intersection, flatten, uniqBy } from "lodash"
 import { DeployTask } from "./deploy"
 import { Garden } from "../garden"
 import { Module } from "../types/module"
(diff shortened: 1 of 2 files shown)


garden-service/src/tasks/helpers.ts
```diff
@@ -47,9 +69,11 @@ export async function getModuleWatchTasks({
     })
   }
 
+  const dependantSourceModuleNames = dependantSourceModules.map((m) => m.name)
+
   const dependantBuildTasks = flatten(
     await Bluebird.map(
-      dependants.build.filter((m) => !m.disabled),
+      dependants.build.filter((m) => !m.disabled && !includes(dependantSourceModuleNames, m.name)),
```
edvald: There's a native `.includes()` method on arrays that you can use.

thsig: Aha. I keep underestimating modern JS.

garden-service/src/tasks/helpers.ts
```diff
@@ -38,7 +38,29 @@ export async function getModuleWatchTasks({
...
+    depModule.serviceConfigs.find((s) => s.sourceModuleName === module.name)
+  )
+
+  const dependantSourceModuleServiceNames = flatten(
+    dependantSourceModules.map((depModule) => {
+      return depModule.serviceConfigs.filter((s) => s.sourceModuleName === module.name).map((s) => s.name)
+    })
+  )
+
+  const serviceNamesForModule = [...module.serviceNames, ...dependantSourceModuleServiceNames]
```
edvald: Is this correct? Should we count the services from a source module along with the services for this module?

thsig: The context here is "services for which `module` has the sources" (which are `module`'s services, and services that have `module` as their `sourceModule`). Or am I missing something?

thsig: Maybe the variable name should be `serviceNamesUsingModule`?

thsig: Ok, cool. I've renamed the variable—the long comment above was supposed to explain the rationale here, do you think this still needs additional clarification?

edvald: No this is good. Approved and merged!

PR #1760, opened 2020-04-01
fix(k8s): automatic include for helm modules should include chartPath

edvald:

Automatic includes weren't factoring in the configured `chartPath` (if any), which could cause errors when using that field.

I just noticed this, rather incidentally.


edvald:
@eysi09 fixed

PR #1772, opened 2020-04-03
fix(k8s): error in cleanup-cluster-registry command and added test

edvald:

One step of cleaning up the cluster registry was failing, which also prevented following steps to complete. Added a test against our cluster as well, so we should catch these issues from now on.


.circleci/config.yml
```diff
@@ -374,10 +374,27 @@ jobs:
...
       - run:
           name: Cleanup
           command: CIRCLE_BUILD_NUM=$CIRCLE_BUILD_NUM-docker kubectl delete --wait=false $(kubectl get ns -o name | grep testing-$CIRCLE_BUILD_NUM) || true
           when: always
+  cleanup-cluster-registry:
+    docker:
+      - image: gardendev/garden-gcloud:${CIRCLE_SHA1}
+        environment:
+          <<: *shared-env-config
+          GARDEN_TASK_CONCURRENCY_LIMIT: "10"
```
eysi09, line 389: Why do we need this here?

.circleci/config.yml
```diff
@@ -374,10 +374,27 @@ jobs:
...
           # overriding CIRCLE_BUILD_NUM to avoid conflict with other tests
           command: CIRCLE_BUILD_NUM=$CIRCLE_BUILD_NUM-docker /garden/garden build --root examples/demo-project --env remote --logger-type basic
+      - run:
+          name: Run cluster cleanup
+          command: CIRCLE_BUILD_NUM=$CIRCLE_BUILD_NUM-docker /garden/garden plugins kubernetes cleanup-cluster-registry --root examples/demo-project --env remote --logger-type basic
       - run:
           name: Cleanup
           command: CIRCLE_BUILD_NUM=$CIRCLE_BUILD_NUM-docker kubectl delete --wait=false $(kubectl get ns -o name | grep testing-$CIRCLE_BUILD_NUM) || true
           when: always
+  cleanup-cluster-registry:
```
eysi09, line 384: Will this slow down our e2e tests in a meaningful way? Not the cleanup step itself but rather the next run on a "clean" registry. We of course don't delete images _in use_, so I'm assuming it won't really have an affect, but wanted to confirm.

.circleci/config.yml
edvald: We don't, it was just in the copy-paste, now removed

.circleci/config.yml
edvald: Not by a lot, but there are other issues with running the test like this. I've instead made it a part of the minikube tests, which is in a more closed environment and has no potential of conflicting with another test.

.circleci/config.yml
```diff
@@ -374,6 +374,9 @@ jobs:
           name: Deploy demo-project with container
           # overriding CIRCLE_BUILD_NUM to avoid conflict with other tests
           command: CIRCLE_BUILD_NUM=$CIRCLE_BUILD_NUM-docker /garden/garden build --root examples/demo-project --env remote --logger-type basic
+      - run:
```
eysi09, line 377: Doesn't this run against the CI cluster?

garden-service/static/kubernetes/system/docker-registry/garden.yml
```diff
@@ -4,7 +4,7 @@ name: docker-registry
 type: helm
 chart: stable/docker-registry
 releaseName: garden-docker-registry
-version: 1.8.3
+version: 1.9.2
```
eysi09, line 7: Is this version already running in our cluster (to make sure the update doesn't break anything)?

eysi09:
@edvald, had two outstanding questions :)

.circleci/config.yml
edvald: Indeed. I'll fix.

garden-service/static/kubernetes/system/docker-registry/garden.yml
edvald: It is, yes.

edvald:
@eysi09 should be good to go now

PR #1775, opened 2020-04-05
improvement: prepareImagePullSecrets() on namespace creation via garden

### LINKED ISSUE #1774: [FEATURE]: Ensure secret for image pulling from private registry is copied to project namespace as well on namespace creation
### Background / Motivation
Currently the configuration for `imagePullSecrets` is supported for in-cluster-building, and as soon as you define it, you can see a secret was created under `garden-system` namespace, which is expected according to the documentation. However, an extension of this feature would be really nice, which is to also create this secret under the app namespace, so whenever a namespace is created via garden deploy, it will include the private registry secret also, and images can be normally pulled.

In cases where I do not want to build an image to push to my local kubernetes cluster, but rather just specify a helm chart as a dependency a scenario like this would be possible:

- I'm using garden module to build and deploy a frontend docker image to my local kubernetes cluster
- My frontend has a dependency with the backend, which is in a different repository and I don't want to build the image, so I could just define a given version in a values configuration in a garden module of type helm
- `garden deploy` would build an image for the frontend, and for the backend would just deploy the helm chart and download the image from a private registry

For example to define a group of applications for deployment via a unique garden.yaml file using modules of type helm, so different engineers could use `garden deploy` to create a namespace for testing, which is one of the awesome ways garden can be really helpful

The function that creates namespaces on https://github.com/garden-io/garden/blob/839da74741c0607a81f6939bc18796ec1ffdaccd/garden-service/src/plugins/kubernetes/namespace.ts#L49, could have some additional lines in order to add the extra functionality similar to what is done with the system namespace at https://github.com/garden-io/garden/blob/839da74741c0607a81f6939bc18796ec1ffdaccd/garden-service/src/plugins/kubernetes/init.ts#L549.

Seems to be quite straightforward and would be quite helpful.

This is a really cool feature to make it easier to manage namespaces with garden.
Mainly because it's easy to just destroy everything under your namespace and create later again in order to save resources. 

I would appreciate that.

jose-fmeyer:
Preparing imagePullSecrets on app namespace creation in addition to what was already done for `garden-system` namespace and for deployment spec created via garden. This way I can delete and re-create environments + namespaces on local and remote kubernetes, and  don't need to worry with the secrets creation for pulling images from a private registry.

Fixes #1774 

Below the test scenario:

Snippet of my garden.yaml

```
kind: Project
name: spring-petclinic
environments:
  - name: local
providers:
  - name: local-kubernetes
    environments: ["local"]
    imagePullSecrets:
      - name: docker-reg-cred

---
kind: Module
description: Helm chart for my module
type: helm
include:
  - kube/spring-petclinic-backend/*
  - kube/spring-petclinic-backend/templates/*
name: spring-petclinic-backend
chartPath: ./kube/spring-petclinic-backend
releaseName: spring-petclinic-backend
```

Make sure you have a secret called `docker-reg-cred` in your target testing cluster.
As soon as garden creates the namespace, you should be able to list the secret copied from the `default` namespace.

### DIFF
--- a/docs/guides/in-cluster-building.md
+++ b/docs/guides/in-cluster-building.md
@@ -65,6 +65,13 @@ In this mode, builds are executed as follows:
 
 After enabling this mode (we currently still default to the `local-docker` mode), you will need to run `garden plugins kubernetes cluster-init --env=<env-name>` for each applicable environment, in order to install the required cluster-wide services. Those services include the Docker daemon itself, as well as an image registry, a sync service for receiving build contexts, two persistent volumes, an NFS volume provisioner for one of those volumes, and a couple of small utility services.
 
+Optionally, you can also enable [BuildKit]((https://github.com/moby/buildkit)). In most cases, this should work well and be more performant, but remains optional for now. If you have `cluster-docker` set as your `buildMode` you can enable BuildKit for an environment as follows:
+
+```yaml
+clusterDocker:
+  enableBuildKit: false
+```
+
 Make sure your cluster has enough resources and storage to support the required services, and keep in mind that these
 services are shared across all users of the cluster. Please look at the [resources](../reference/providers/kubernetes.md#providersresources) and [storage](../reference/providers/kubernetes.md#providersstorage) sections in the provider reference for
 details.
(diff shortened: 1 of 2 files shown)


eysi09:
Thanks for the pull request @jfernandomeyer!

My apologies for not getting back to you sooner. But I believe this one is slightly trickier than just preparing the secrets when creating the namespace. The person responsible for this part of the codebase is away on holiday but we can give you a more constructive feedback once he's back.

jfernandomeyer:
Glad to discuss what can be improved or the edge cases I'm not seeing here @eysi09. 
I've been using my development version since the PR was opened and faced no any issue so far if this helps somehow. 

Will be waiting for further feedback and will adapt whatever comes as a suggestion. 👍

Thanks a lot!!

thsig:
Implemented in #2714.

PR #1776, opened 2020-04-06
fix(terraform): fix stdin handling in commands

### LINKED ISSUE #1709: Terraform apply-root | apply-module commands can't receive input
### Current Behavior

If you run `garden plugins terraform apply-root | apply-module`, the command will ask for input from the user. However, the logger won't register the input and the command eventually times out.

### Expected behavior

The user should be able to type in the needed input.

### Reproducible example

Run `garden plugins terraform apply-root` in the `terraform-gke` example.

### Workaround

Run the command with `-auto-approve`:

```
garden plugins terraform apply-root|apply-module -auto-approve
```

or call Terraform directly with:

```
~/.garden/tools/terraform-<VERSION>/<HASH>/terraform apply -var-file garden.tfvars.json
```

in the directory where the Terraform resources are defined.

### Suggested solution(s)

Calling `logger.stop` should prevent this.

thsig:

Before this fix, `stdin` was not being handled correctly for the terraform plugin's commands. They were unresponsive to input (including `CTRL-C`), which prevented the commands from being run to completion in some cases.

This was fixed by not setting `setRawMode` to `true` on the child process' `stdin`.

We added the `rawMode` option to `SpawnParams` to facilitate other use cases that need interactive input, but where `rawMode` doesn't play nicely with `stdin` for the underlying child process.

Fixes #1709.

Is there a decent way to test this for future regressions?

I thought about using the input stream's `write` method (see https://nodejs.org/api/net.html#net_socket_write_data_encoding_callback) for this (e.g. for testing that `CTRL-C` is handled/respected).

Would that be a valid approach, or would it be functionally different from receiving inherited input (and/or be problematic in a CI environment)?

If this approach does make sense, then I'd like to write a couple of unit tests for this too.

### DIFF
--- a/garden-service/src/plugins/terraform/commands.ts
+++ b/garden-service/src/plugins/terraform/commands.ts
@@ -46,7 +46,7 @@ function makeRootCommand(commandName: string) {
       await tfValidate(log, provider, root, provider.config.variables)
 
       args = [commandName, ...(await prepareVariables(root, provider.config.variables)), ...args]
-      await terraform(provider.config.version).spawnAndWait({ log, args, cwd: root, tty: true })
+      await terraform(provider.config.version).spawnAndWait({ log, args, cwd: root, rawMode: false, tty: true })
 
       return { result: {} }
     },
@@ -73,7 +73,7 @@ function makeModuleCommand(commandName: string) {
       await tfValidate(log, provider, root, provider.config.variables)
 
       args = [commandName, ...(await prepareVariables(root, module.spec.variables)), ...args.slice(1)]
-      await terraform(module.spec.version).spawnAndWait({ log, args, cwd: root, tty: true })
+      await terraform(module.spec.version).spawnAndWait({ log, args, cwd: root, rawMode: false, tty: true })
 
       return { result: {} }
     },
--- a/garden-service/src/util/util.ts
+++ b/garden-service/src/util/util.ts
@@ -190,6 +190,7 @@ export interface SpawnOpts {
   data?: Buffer
   ignoreError?: boolean
   env?: { [key: string]: string | undefined }
+  rawMode?: boolean // Only used if tty = true. See also: https://nodejs.org/api/tty.html#tty_readstream_setrawmode_mode
   stdout?: Writable
   stderr?: Writable
   tty?: boolean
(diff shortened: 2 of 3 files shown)


edvald:
Looks good at first glance. Any idea why disabling raw mode is necessary for this particular scenario?

thsig:
Not 100% sure, but quoting the docs here: https://nodejs.org/api/tty.html#tty_readstream_setrawmode_mode

> Additionally, all special processing of characters by the terminal is disabled, including echoing input characters. CTRL+C will no longer cause a SIGINT when in this mode.

This explains CTRL-C not SIGINT-ing.

Not echoing back input characters might explain the input not working on the prompts (maybe typing "yes" and then return would work, I don't think it did when we were testing this out).

thsig:
I guess the question is "why does it work for `kubectl`" and certain other commands with `setRawMode = true`.

thsig:
Maybe there's something special about the way the input is piped back and forth when the child process creates a shell (and possibly handles input/output in a different way from a "raw command that handles stdin/stdout)?

garden-service/src/plugins/terraform/commands.ts
```diff
@@ -46,7 +46,7 @@ function makeRootCommand(commandName: string) {
       await tfValidate(log, provider, root, provider.config.variables)
 
       args = [commandName, ...(await prepareVariables(root, provider.config.variables)), ...args]
-      await terraform(provider.config.version).spawnAndWait({ log, args, cwd: root, tty: true })
+      await terraform(provider.config.version).spawnAndWait({ log, args, cwd: root, rawMode: false, tty: true })
```
eysi09, line 49: Can we test for this? Feels like something could easily get lost during a refactor. Would also be good to add an explanatory comment.

garden-service/src/plugins/terraform/commands.ts
```diff
@@ -73,7 +73,7 @@ function makeModuleCommand(commandName: string) {
       await tfValidate(log, provider, root, provider.config.variables)
 
       args = [commandName, ...(await prepareVariables(root, module.spec.variables)), ...args.slice(1)]
-      await terraform(module.spec.version).spawnAndWait({ log, args, cwd: root, tty: true })
+      await terraform(module.spec.version).spawnAndWait({ log, args, cwd: root, rawMode: false, tty: true })
```
eysi09, line 76: Same as above.

garden-service/src/util/util.ts
```diff
@@ -206,7 +207,7 @@ export interface SpawnOutput {
 
 // TODO Dump output to a log file if it exceeds the MAX_BUFFER_SIZE
 export function spawn(cmd: string, args: string[], opts: SpawnOpts = {}) {
-  const { timeout = 0, cwd, data, ignoreError = false, env, stdout, stderr, tty, wait = true } = opts
+  const { timeout = 0, cwd, data, ignoreError = false, env, rawMode = true, stdout, stderr, tty, wait = true } = opts
```
eysi09, line 210: We throw if you have `tty: true` and `data` defined. Should we do the same if you have `tty: false` and `rawMode: true`?

eysi09:
> I.e. maybe kubectl, for example, is specifically listening for and handling raw keyboard events when it's running a shell, which interactive commands in general may not do?

If that's the case, shouldn't the default be `rawMode = false`?

edvald:
> If that's the case, shouldn't the default be rawMode = false?

Possibly, but that's changing existing behavior, so we'd need to walk through all relevant usages. I think this is fine for now.

PR #1788, opened 2020-04-09
fix: use tail -f

mitchfriedman:
We need to follow the /tmp/output file in order to get all the logs, otherwise we will only get the last 10 (default by tail).

https://github.com/garden-io/garden/issues

Fixes #

@edvald


edvald:
Thank you @mitchfriedman! That was subtle, good catch. LGTM except, would you mind rephrasing the commit message? Something like `fix(k8s): don't truncate logs in task/test runs`, just so that it's more clear in our changelog and history.

mitchfriedman:
Yep definitely 👍

edvald:
And a shoutout to @lucas-sonnabend too :) Some nice teamwork on the community Slack there.

PR #1796, opened 2020-04-17
fix(platform): fix login/logout control flow

thsig:

Before this fix, the logout command would require the user to be logged in, which prevented the user from logging out when they had an invalid/expired auth token persisted in their local store.

### DIFF
--- a/garden-service/src/garden.ts
+++ b/garden-service/src/garden.ts
@@ -61,7 +61,7 @@ import { deline, naturalList } from "./util/string"
 import { ensureConnected } from "./db/connection"
 import { DependencyValidationGraph } from "./util/validate-dependencies"
 import { Profile } from "./util/profiling"
-import { readAuthToken, login } from "./cloud/auth"
+import { readAuthToken, checkClientAuthToken } from "./cloud/auth"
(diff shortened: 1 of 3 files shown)


garden-service/src/garden.ts
```diff
@@ -291,15 +292,21 @@ export class Garden {
...
           backend endpoint (e.g. myusername-cloud-api.cloud.dev.garden.io, without an http/https
           prefix).`
         throw new InternalError(errMsg, {})
       } else {
-        await login(platformUrl, log)
+        const tokenIsValid = await checkClientAuthToken(clientAuthToken, platformUrl, log)
+        if (!tokenIsValid) {
+          log.warn(deline`
+            You were previously logged in to the platform, but your session has expired or is invalid. Please run
+            garden login to continue using platform features, or run garden logout to suppress this message.
```
eysi09, line 307: Could you add quotes around the commands (i.e. garden login/logout) or highlight them with something like `chalk.bold("garden login")`'

thsig: Added `chalk.bold` around the command invocations in a separate commit (to be squashed).

PR #1797, opened 2020-04-17
feat(k8s): run a persistent skopeo daemon

mitchfriedman:
This adds a persistent skopeo daemon deployed to the cluster. This is to avoid having to spin up a new skopeo container per build status check for container. We end up seeing some failures randomly in our EKS cluster, most likely to do with failures/timeouts mounting volumes, and this will resolve that issue for us.

@edvald

No issue.

### DIFF
--- a/garden-service/src/plugins/kubernetes/constants.ts
+++ b/garden-service/src/plugins/kubernetes/constants.ts
@@ -17,3 +17,9 @@ export const MAX_RUN_RESULT_LOG_LENGTH = 250 * 1024
 export const dockerAuthSecretName = "builder-docker-config"
 export const dockerAuthSecretKey = ".dockerconfigjson"
 export const inClusterRegistryHostname = "127.0.0.1:5000"
+
+export const gardenUtilDaemonDeploymentName = "garden-util-daemon"
+export const dockerDaemonDeploymentName = "garden-docker-daemon"
+
+export const dockerDaemonContainerName = "docker-daemon"
+export const skopeoDaemonContainerName = "util"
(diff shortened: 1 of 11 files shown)


garden-service/src/plugins/kubernetes/constants.ts
```diff
@@ -17,3 +17,9 @@ export const MAX_RUN_RESULT_LOG_LENGTH = 250 * 1024
 export const dockerAuthSecretName = "builder-docker-config"
 export const dockerAuthSecretKey = ".dockerconfigjson"
 export const inClusterRegistryHostname = "127.0.0.1:5000"
+
+export const skopeoDaemonDeploymentName = "garden-skopeo-daemon"
```
edvald, line 21: One thought: I think it might be good to call the service something like `garden-util`. This way we could over time add more utilities to the image without deploying more Pods unnecessarily.

garden-service/src/plugins/kubernetes/container/build.ts
```diff
@@ -464,13 +437,13 @@ export async function execInBuilder({
   })
 }
 
-export async function getBuilderPodName(provider: KubernetesProvider, log: LogEntry) {
-  const pod = await getRunningPodInDeployment(dockerDaemonDeploymentName, provider, log)
+export async function getBuilderPodName(deployment: string, provider: KubernetesProvider, log: LogEntry) {
```
edvald, line 440: It would be good to rename this function, as well as `execInBuilder` since they are now more generic, e.g. `getDeploymentPodName` and `execInPod`.

garden-service/src/plugins/kubernetes/constants.ts
mitchfriedman: yeah seems sensible

PR #1802, opened 2020-04-20
chore(docker): add image that contains gcloud and aws

solomonope:
<!--  Thanks for sending a pull request!  Here are some tips for you:

This PR creates an image that include gcloud, awscli and gardencli in one image
It is meant to be used for gardencloud staging deployment
Fixes #

### DIFF
--- a/garden-service/bin/build-containers.sh
+++ b/garden-service/bin/build-containers.sh
@@ -10,6 +10,7 @@ version=${args[0]:-$(git rev-parse --short HEAD)}
 base_tag=gardendev/garden:${version}
 aws_tag=gardendev/garden-aws:${version}
 gcloud_tag=gardendev/garden-gcloud:${version}
+aws_gcloud_tag=gardendev/garden-aws-gcloud:${version}
 buster_tag=gardendev/garden:${version}-buster
 
 echo "Building version ${version}"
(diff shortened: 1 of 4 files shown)


garden-service/bin/build-containers.sh
```diff
@@ -10,6 +10,7 @@ version=${args[0]:-$(git rev-parse --short HEAD)}
 base_tag=gardendev/garden:${version}
 aws_tag=gardendev/garden-aws:${version}
 gcloud_tag=gardendev/garden-gcloud:${version}
+all_tag=gardendev/garden-all:${version}
```
eysi09, line 13: Can we find a better name than `garden-all`? It's not clear to the end user what `all` means in this context? What about `garden-all-providers`, `garden-all-clouds`, or simply `garden-aws-gcloud`. 

I think I actually like the last one best (if we don't foresee adding more credential helpers) but open to suggestions.

solomonope: I have updated it, I generally don't like length names, for me a short name and a bit of context works.

eysi09: Yeah, that's valid point as well. I'll approve this for now and we can later make the changes mentioned internally.

edvald:
Merging this, but wanted to mention that we do plan on changing our docker image structure, and plan on merging all the individual provider-specific ones into a single one and/or making different version tags within a single repo, instead of all the specific repos.

PR #1803, opened 2020-04-20
fix(dashboard): make sure we can display disabled modules

swist:
Turns out that if module is disabled, `modules[node.name]`
is undefined (perhaps because it got filtered out). Since
we don't care about disabled modules other than saying:
"here's the name and it's disabled", I have fixed the
method for getting the entity from API response and made
it more defensive in general case (better to miss results
than to display react exception :))

Fixes (As described above)

I am not sure if `getTestKey` is still needed. It worked fine without it, but my eyballs could have been deceiving me.

### DIFF
--- a/dashboard/src/containers/graph.tsx
+++ b/dashboard/src/containers/graph.tsx
@@ -8,14 +8,15 @@
 
 import React, { useEffect } from "react"
 import styled from "@emotion/styled"
+
 import Graph from "../components/graph"
 import PageError from "../components/page-error"
-import { TaskState, useApi } from "../contexts/api"
+import { TestEntity, TaskEntity, ModuleEntity, ServiceEntity, Entities, TaskState, useApi } from "../contexts/api"
 import { StackGraphSupportedFilterKeys, EntityResultSupportedTypes, useUiState } from "../contexts/ui"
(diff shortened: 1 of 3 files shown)


dashboard/src/containers/graph.tsx
```diff
@@ -25,6 +25,24 @@ const Wrapper = styled.div`
   padding-left: 0.75rem;
 `
 
+const getEntity = (
+  node: RenderedNode,
+  entities: Entities
+): TestEntity | TaskEntity | ModuleEntity | ServiceEntity | undefined => {
+  switch (node.type) {
+    case "deploy":
+      return entities.services[node.name]
```
swist, line 34: This seems to have passed the typecheck, but perhaps you'd prefer an explicit typecast?

eysi09:
Hi @swist and thanks for the PR!

On close inspection, the underlying issue is that the `get config` command (which the dashboard relies on) was no longer returning disabled module configs. I fixed that in #1806.

If that works as expected, we shouldn't need any changes to the dashboard. Any chance you'd be able to verify that it looks good on your end?

swist:
closing since this was fixed in #1806

PR #1806, opened 2020-04-22
fix(commands): ensure get config returns disabled configs (by default)

eysi09:

Due to a regression, the `get config` command was no longer returning disabled module configs. This caused issues with the dashboard.

This PR ensures that `get config` returns disabled configs by default and adds tests.

EDIT: Also added a `disabled-configs` example project and a manual release test.  

Addresses the same problem as this PR: #1803 

### DIFF
--- a/bin/test-release.sh
+++ b/bin/test-release.sh
@@ -82,12 +82,18 @@ test_release() {
   echo ""
   ${garden_release} create module
   echo ""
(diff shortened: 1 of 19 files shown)


swist:
Re #1803 yep this also fixes that bug. Do you still want the PR that aligns frontend types?

PR #1807, opened 2020-04-22
feat(platform): added id and domain to projects

thsig:

This replaces the temporary setup we had based on project names (using ids instead) and the `GARDEN_CLOUD` env variable.

### DIFF
--- a/garden-service/src/config/project.ts
+++ b/garden-service/src/config/project.ts
@@ -150,6 +150,8 @@ export interface ProjectConfig {
   kind: "Project"
   name: string
   path: string
+  id?: string
+  domain?: string
   configPath?: string
   defaultEnvironment: string
   dotIgnoreFiles: string[]
@@ -251,8 +253,7 @@ const projectOutputSchema = () =>
       .description(
         dedent`
         The value for the output. Must be a primitive (string, number, boolean or null). May also be any valid template
-        string.
-      `
+        string.`
       )
       .example("${modules.my-module.outputs.some-output}"),
   })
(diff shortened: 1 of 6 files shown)


garden-service/src/config/project.ts
```diff
@@ -277,6 +278,12 @@ export const projectDocsSchema = () =>
         .meta({ internal: true })
         .description("The path to the project config file."),
       name: projectNameSchema(),
+      // TODO: Refer to platform documentation for more details.
+      id: joi.string().description("The project's ID in Garden Cloud."),
```
edvald, line 282: I suggest marking these with `.meta({ internal: true })` for the time being.

garden-service/src/config/project.ts
```diff
@@ -277,6 +278,12 @@ export const projectDocsSchema = () =>
...
+      // TODO: Refer to platform documentation for more details.
+      domain: joi
+        .string()
+        .description("The domain to use for cloud features. Should point to the API/backend endpoint."),
```
edvald, line 286: endpoint -> hostname?

garden-service/src/garden.ts
```diff
@@ -287,20 +289,38 @@ export class Garden {
...
+          errorMessages.push(deline`
+            ${chalk.bold("project.domain")} is not set in your project-level ${chalk.bold("garden.yml")}. Make sure it
+            is set to the appropriate API backend endpoint (e.g. myusername-cloud-api.cloud.dev.garden.io, without an
+            http/https prefix).
+          `)
+        }
+        if (!projectId) {
+          errorMessages.push(deline`
+            ${chalk.bold("project.id")} is not set in your project-level ${chalk.bold("garden.yml")}. Please use the
+            ${chalk.bold("garden set project")} command to set this value, or paste it from the platform's web UI
```
eysi09, line 309: Is it weird to suggest the command before adding it? But I suppose you can say that about the other stuff as well.

thsig: Yeah, agreed. I'll remove the reference to the command in here (was planning to implement the command at the time I wrote that description).

PR #1810, opened 2020-04-23
chore(analytics): stop collecting task graph events

eysi09:

The task graph emits a lot of events that are mostly noise. Instead we
want to track individual task events at the task level. Such as a build
event or a deploy event.

Fixes #

### DIFF
--- a/garden-service/src/analytics/analytics.ts
+++ b/garden-service/src/analytics/analytics.ts
@@ -16,7 +16,6 @@ import { SEGMENT_PROD_API_KEY, SEGMENT_DEV_API_KEY } from "../constants"
 import { LogEntry } from "../logger/log-entry"
 import hasha = require("hasha")
 import { Garden } from "../garden"
-import { Events, EventName } from "../events"
 import { AnalyticsType } from "./analytics-types"
 import dedent from "dedent"
 import { getGitHubUrl } from "../docs/common"
@@ -38,7 +37,7 @@ export function getUserId(globalConfig: GlobalConfig) {
   }
 }
 
-export interface SystemInfo {
+interface SystemInfo {
   gardenVersion: string
   platform: string
   platformVersion: string
@@ -54,7 +53,7 @@ interface ProjectMetadata {
   moduleTypes: string[]
 }
 
-export interface AnalyticsEventProperties {
+interface AnalyticsEventProperties {
   projectId: string
   projectName: string
   ciName: string | null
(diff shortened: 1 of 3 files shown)


garden-service/src/analytics/analytics.ts
```diff
@@ -244,25 +240,15 @@ export class AnalyticsHandler {
...
-  }
-
-  /**
-   * Typeguard to check wether we can process or not an event
-   */
-  static isSupportedEvent(name: EventName, _event: Events[EventName]): _event is SupportedEvents {
-    const supportedEventsKeys = ["taskPending", "taskProcessing", "taskComplete", "taskError"]
-    return supportedEventsKeys.includes(name)
+    // TODO: (@eysi) Re-implement once we've implemented events at the task level
+    // AnalyticsHandler.instance.garden.events.onAny((name, payload) =>
```
edvald, line 249: I don't think we will re-implement this. I'd favor more explicit tracking anyway, as opposed to implicitly converting from the `Garden` event bus.

eysi09: Right, so just call something like `analytics.trackBuild` from the `Task` class as opposed to actually emitting a build event?

garden-service/src/analytics/analytics.ts
```diff
@@ -233,36 +224,14 @@ export class AnalyticsHandler {
...
-  /**
-   * Handler used internally to process the TaskGraph events.
-   */
-  private async processEvent<T extends EventName>(name: T, payload: Events[T]) {
-    if (AnalyticsHandler.isSupportedEvent(name, payload)) {
-      await this.trackTask(payload.batchId, payload.name, payload.type, name)
-    }
-  }
-
   static async refreshGarden(garden: Garden) {
```
10ko, line 233: I don't think this function is needed at all anymore if you remove the Tasks event handler.
The main reason I created it was to refresh the EventsBuffer.
The `this.garden`instance is only accessed for hashing the project name and to retrieve the project metadata, so I believe it should be safe to remove the refresh functionality.
Double-check thou. :)

garden-service/src/analytics/analytics.ts
edvald: Yeah. It does feel slightly odd to me to put a specific method on the AnalyticsHandler actually, just in terms of separation of concerns. But functionally, yes, that's what I had in mind.

eysi09: So where would you put the method?

edvald: This will only be called in a single location. So imo we'd just use a more generic tracking method and set the shape of the event at the call location.

eysi09: Ahhhh, gotcha. Yeah absolutely, this was just pseudo code to make the intent clear. I thought you were referring to moving the handler somewhere else.

PR #1820, opened 2020-05-06
chore(k8s): bump kaniko

mitchfriedman:
There appears to have been a bug in the v0.20 kaniko release, so they released a quick update:

https://github.com/GoogleContainerTools/kaniko/releases/tag/v0.21.0

@edvald


mitchfriedman:
I think wait on merging this one: there is some discussion if the build went through or not: https://kubernetes.slack.com/archives/CQDCHGX7Y/p1588768957094500

mitchfriedman:
I think this is fine actually based on https://kubernetes.slack.com/archives/CQDCHGX7Y/p1588790020098100

PR #1826, opened 2020-05-12
chore(k8s): upgrade helm

mitchfriedman:
This upgrades helm to the latest versions.

@edvald

### DIFF
--- a/garden-service/src/plugins/kubernetes/helm/build.ts
+++ b/garden-service/src/plugins/kubernetes/helm/build.ts
@@ -7,7 +7,7 @@
  */
 
 import { join } from "path"
-import { writeFile } from "fs-extra"
+import { remove, writeFile } from "fs-extra"
 import { HelmModule } from "./config"
 import { containsBuildSource, getChartPath, getGardenValuesPath, getBaseModule, renderTemplates } from "./common"
 import { helm } from "./helm-cli"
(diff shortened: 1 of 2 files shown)


edvald:
Looks like there is a breaking behavior change in `helm fetch` that's causing issues.

```
Error: Command "/home/circleci/.garden/tools/helm3/dff97bc0c693fd1b/linux-amd64/helm --kube-context minikube fetch stable/postgresql --destination /home/circleci/project/garden-service/test/data/test-projects/helm/.garden/build/postgres --untar --version 3.9.2" failed with code 1:

Error: failed to untar: a file or directory with the name /home/circleci/project/garden-service/test/data/test-projects/helm/.garden/build/postgres/postgresql already exists
```

It seems like there isn't an option to revert back to the old behavior, so I guess we need to check and remove the existing directory if it already exists, in this function: https://github.com/garden-io/garden/blob/b23eeaaf391610ea8c2b9a93afdc778cadd6d020/garden-service/src/plugins/kubernetes/helm/build.ts#L78

garden-service/src/plugins/kubernetes/helm/build.ts
```diff
@@ -78,6 +78,8 @@ export async function buildHelmModule({ ctx, module, log }: BuildModuleParams<He
 async function fetchChart(ctx: KubernetesPluginContext, log: LogEntry, module: HelmModule) {
   const buildPath = module.buildPath
 
+  await remove(buildPath)
```
edvald, line 81: Ah, small problem, should have flagged this. We should not remove the whole build directory, just the chart directory. This would otherwise mess with other generated/built/copied files. The `getChartPath()` function returns the appropriate path. Then it should be good to go.

mitchfriedman: ah yeah okay makes sense - updated

PR #1828, opened 2020-05-12
fix(container): explicit task & test timeouts

thsig:

Before this fix, tasks and tests for `container`, `helm`, `kubernetes` and `maven-container` modules implicitly used a default timeout of 600 seconds, but the reference docs listed the default timeout as `null`, which has confused some users.

Here, we make default test and task timeouts explicit for these module types (while leaving the default timeouts for the base test and task specs, used e.g. by the `exec` module type, as `null`).

### DIFF
--- a/docs/reference/module-types/container.md
+++ b/docs/reference/module-types/container.md
@@ -337,9 +337,6 @@ tests:
     # specific environments, e.g. only during CI.
     disabled: false
 
-    # Maximum duration (in seconds) of the test run.
-    timeout: null
-
     # The arguments used to run the test inside the container.
     args:
 
@@ -390,6 +387,9 @@ tests:
         # services at the same time. Refer to the documentation of the module type in question to learn more.
         module:
 
+    # Maximum duration (in seconds) of the test run.
+    timeout: 600
+
 # A list of tasks that can be run from this container module. These can be used as dependencies for services (executed
 # before the service is deployed) or for other tasks.
 tasks:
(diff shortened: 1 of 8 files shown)


edvald:
My instinct would be to correct this by removing the implicit timeout. Why did you lean in this direction instead?

thsig:
Hmm, I was thinking of this mostly as a change to clarify the current behavior—the task and test handlers for these module types end up calling `kubectl`, and using the `defaultCommandTimeoutSecs` value (600 secs) if no timeout is provided:

I figured it might be a breaking change if we were to change this.

Btw, just noticed that this breaks a few test cases (ones that hash the contents of configs using these default values). If we're in favour of the changes in this commit, I'll go ahead and fix those.

thsig:
Or do we want to go ahead and remove the implicit timeout, as you say?

edvald:
Gotcha. That does make sense, but in my mind, having a default timeout of 10 minutes is slightly odd and arbitrary behavior. Plus it wasn't the intent. So I'd rather just make sure the default `null` timeout works as expected.

edvald:
Is this still in progress @thsig?

thsig:
Haven't looked at it since our comment thread above the other day, but I think it's time to finish it along the lines we discussed.

thsig:
I'll look into this.

vvagaytsev:
@thsig can we close this? In `0.13` we plan to set non-null default values for all timeout fields, and require a number.

thsig:
Yeah, let's close this.

PR #1830, opened 2020-05-13
feat(k8s): support immutable build success

mitchfriedman:
When using immutable tags on ECR, it gives an error that looks like the one in the test and causes the kaniko build to fail. This gracefully handles that as a success.

 @edvald

### DIFF
--- a/garden-service/src/plugins/kubernetes/container/build.ts
+++ b/garden-service/src/plugins/kubernetes/container/build.ts
@@ -42,6 +42,7 @@ import { getSystemNamespace, getAppNamespace } from "../namespace"
 import { dedent } from "../../../util/string"
 import chalk = require("chalk")
 import { loadImageToMicrok8s, getMicrok8sImageStatus } from "../local/microk8s"
+import { RunResult } from "../../../types/plugin/base"
 
 const kanikoImage = "gcr.io/kaniko-project/executor:debug-v0.21.0"
 
@@ -352,7 +353,7 @@ const remoteBuild: BuildHandler = async (params) => {
     const buildRes = await runKaniko({ provider, namespace, log, module, args, outputStream: stdout })
     buildLog = buildRes.log
 
-    if (!buildRes.success) {
+    if (kanikoBuildFailed(buildRes)) {
       throw new BuildError(`Failed building module ${chalk.bold(module.name)}:\n\n${buildLog}`, { buildLog })
     }
   }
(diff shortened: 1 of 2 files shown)


garden-service/test/unit/src/plugins/kubernetes/container/build.ts
```diff
@@ -0,0 +1,28 @@
...
+
+describe("kaniko build", () => {
+  it("logs containing immutable errors are success", () => {
+    const errorMessage = `
+    error pushing image: failed to push to destination
+  dockerhub.com/garden/backend:v-1234567: TAG_INVALID: The
+  image tag 'v-1234567' already exists in the 'garden/backend' repository and cannot be
+  overwritten because the repository is immutable.`
+
+    expect(kanikoBuildFailed({ succes: false, buidlLog: errorMessage })).to.be.false
```
edvald, line 20: Typo in both parameter names here

garden-service/test/unit/src/plugins/kubernetes/container/build.ts
```diff
@@ -0,0 +1,28 @@
...
+  })
+
+  it("logs containing immutable errors are success", () => {
```
edvald, line 23: Description here is duplicated

garden-service/test/unit/src/plugins/kubernetes/container/build.ts
```diff
@@ -0,0 +1,28 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { kanikoBuildFailed } from "../../../../../../src/plugins/kubernetes/container/build"
+import { expect } from "chai"
+
+describe("kaniko build", () => {
+  it("logs containing immutable errors are success", () => {
```
edvald, line 13: Test description is a little confusing. Maybe something like `"should return as successful when immutable tag already exists in destination"` would be clearer.

garden-service/test/unit/src/plugins/kubernetes/container/build.ts
```diff
@@ -0,0 +1,28 @@
...
+  image tag 'v-1234567' already exists in the 'garden/backend' repository and cannot be
+  overwritten because the repository is immutable.`
+
+    expect(kanikoBuildFailed({ succes: false, buidlLog: errorMessage })).to.be.false
+  })
+
+  it("logs containing immutable errors are success", () => {
+    const errorMessage = "error pushing"
+
+    expect(kanikoBuildFailed({ succes: false, buidlLog: errorMessage })).to.be.true
```
edvald, line 26: Parameter typos also present here. Also, could we add a simple test case for when `success: true`?

garden-service/src/plugins/kubernetes/container/build.ts
```diff
@@ -380,6 +380,13 @@ export interface BuilderExecParams {
   stderr?: Writable
 }
 
+export function kanikoBuildFailed(buildRes) {
```
edvald, line 383: Adding a type declaration here would be good, and would prevent typos like in the tests below. The correct type for `buildRes` would be `RunResult`.

PR #1836, opened 2020-05-22
feat(core): workflows for sequencing commands

thsig:

Workflows are a new framework primitive. They specify a list of `steps`, each of which specifies a Garden command to be run.

They also specify `triggers`, which facilitate automatic triggering of workflows based on VCS events.

`Workflow` configs are a new top-level config type, along with `Project` and `Module`.

The `run workflow` command can be used to run a workflow by name.


garden-service/src/cli/cli.ts
```diff
@@ -77,7 +77,7 @@ class DummyGarden extends Garden {
   async resolveProviders() {
     return []
   }
-  async scanModules() {}
+  async scanAndAddConfigs() {}
```
eysi09, line 80: 👍

examples/workflows/garden.yml
```diff
@@ -0,0 +1,33 @@
+kind: Project
+name: tasks
```
eysi09, line 2: Change name to `worfklows`

examples/workflows/garden.yml
```diff
@@ -0,0 +1,33 @@
...
+  - name: testing
+providers:
+  - name: local-kubernetes
+    environments: [local]
+  - name: kubernetes
+    environments: [testing]
+    context: gke_garden-platform-dev-267510_europe-west1-b_gke-cluster-dev-f29377c0
+    # context: gke_garden-dev-200012_europe-west1-b_garden-dev-1
+    namespace: workflows-testing-${local.env.CIRCLE_BUILD_NUM || local.username}
+    defaultHostname: tasks-testing.dev-1.sys.garden
```
eysi09, line 14: Change hostname to `workflows-testing.dev-1.sys.garden`

examples/workflows/hello/garden.yml
```diff
@@ -0,0 +1,33 @@
...
+      - init-db
+tests:
+  - name: unit
+    args: [npm, test]
+  - name: integ
+    args: [npm, test]
+    # args: [npm, run, test:integ]
+    dependencies:
+      - populate-db
+tasks:
```
eysi09, line 24: I sometimes get the following error when running the task:

```
Failed running task populate-db in module user. Here is the output:
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

PG::UndefinedTable: ERROR:  relation "users" does not exist
LINE 8:  WHERE a.attrelid = '"users"'::regclass
                            ^
```

eysi09: This happens if the task result is already cached. You can test this by running the `garden run workflows` twice.

garden-service/src/commands/run/workflow.ts
```diff
@@ -0,0 +1,151 @@
...
+    help: "The name of the workflow to be run.",
+    required: true,
+  }),
+}
+
+type Args = typeof runWorkflowArgs
+
+export class RunWorkflowCommand extends Command<Args, {}> {
+  name = "workflow"
+  alias = "w"
```
eysi09, line 30: I'm a little on the fence with this alias, `garden run w` seems a little odd.

garden-service/src/commands/run/workflow.ts
```diff
@@ -0,0 +1,151 @@
...
+
+        garden run workflow my-workflow    # run my-workflow
+  `
+
+  arguments = runWorkflowArgs
+
+  async action({ garden, log, headerLog, args, opts }: CommandParams<Args, {}>): Promise<CommandResult<null>> {
+    const workflow = await garden.getWorkflowConfig(args.workflow)
+    const steps = workflow.steps
+    if (steps.length === 0) {
```
eysi09, line 46: Should this not be validated at the config level?

garden-service/src/commands/run/workflow.ts
```diff
@@ -0,0 +1,151 @@
...
+      throw new CommandError(
+        chalk.red(`Workflow ${workflow.name} has 0 steps defined. A workflow must define at least one step to be run.`),
+        { workflow }
+      )
+    }
+
+    printHeader(headerLog, `Running workflow ${chalk.white(workflow.name)}`, "runner")
+
+    const inheritedOpts = { ...opts, "logger-type": "basic", "interactive": false }
```
eysi09, line 55: Why do we set the `logger-type` and `interactive` here?

garden-service/src/garden.ts
```diff
@@ -637,6 +646,22 @@ export class Garden {
     return providers
   }
 
+  async getWorkflowConfig(name: string): Promise<WorkflowConfig> {
+    return (await this.getWorkflowConfigs([name]))[0]
+  }
+
+  async getWorkflowConfigs(names?: string[]): Promise<WorkflowConfig[]> {
+    const providers = await this.resolveProviders()
+    throwOnMissingSecretKeys(this.workflowConfigs, this.secrets, "Workflow")
```
eysi09, line 655: Feels a little brittle to have to do this in several places. Is there a way to consolidate it somehow?

docs/reference/commands.md
```diff
@@ -728,6 +728,26 @@ Examples:
   | `--force` |  | boolean | Run the test even if it&#x27;s disabled for the environment.
   | `--force-build` |  | boolean | Force rebuild of module before running.
 
+### garden run workflow
```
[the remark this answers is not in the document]
thsig: Yeah, would be good to write a short guide on this.

We should beef up and document the `workflows` example project at the same time we do that, so we're able to reference it in the guide. 

That example is quite bare-bones and non-shiny right now, not really something intended for users to try out—just wrote it to test out the `run workflow` command as I was writing it, and left it in there for the reviewer of this PR to kick the tyres. I was sort of wondering if we should remove it before we merge. If not, it needs some extra work.

docs/reference/config.md
```diff
@@ -846,3 +846,260 @@ Defaults to to same as source path.
...
+
+    # A description of the workflow step.
+    description:
+
+# A list of triggers that determine when the workflow should be run, and which environment should be used.
+triggers:
+  - # The environment name (from your project configuration) to use for the workflow when matched by this trigger.
+    environment:
+
+    # A list of GitHub events that should trigger this workflow.
```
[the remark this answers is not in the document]
thsig: Yeah, I wondered that too. "GitHub or GitLab events" or "VCS provider" both sounded a bit off.

Then again, we're now omitting this from the generated docs, so we could also revisit this later.

garden-service/src/commands/run/workflow.ts
thsig: Yeah, agreed. I'll just get rid of the alias for now.

garden-service/src/commands/run/workflow.ts
thsig: Agreed.

My thinking at the time was that the user might want to write a placeholder workflow (e.g. starting with the trigger specs) without things failing to validate, but on second thought, we don't allow non-functional/abstract configs for anything else.

I'll move this (and the test case) to the config validation piece.

garden-service/src/commands/run/workflow.ts
thsig: I'll get rid of those—these are leftovers from when I was fiddling around with the log output, seem to have forgotten to get rid of them

garden-service/src/garden.ts
thsig: Hmm, we could definitely do the missing secrets check for modules and workflows in the same place (e.g. in `Garden#scanAndAddConfigs`).

Provider resolution is a "mandatory" operation that's done in `Garden.factory`.

`scanAndAddConfigs`, however, is ultimately called by `garden.getConfigGraph` which is called on an as-needed basis by e.g. commands. So we'd need to do a bit of restructuring of the control flow around this to only validate missing secrets in one place.

So how about we call `throwOnMissingSecretKeys` in two places for now: Once in the provider resolution logic, and once after we scan & add module and workflow configs?

This reminds me that I forgot to call `getConfigGraph` in the `run workflow` command's `action` method (which might result in errors down the line if we support commands that don't call `getConfigGraph`). I'll add that call there now.

examples/workflows/hello/garden.yml
thsig: Yeah. This isn't a problem with the `run workflow` command per se, just the way this particular project is configured.

I'll add a commit that cleans up and improves this example project (which will also address the other comments here).

garden-service/src/garden.ts
eysi09: Yeah, that makes sense.

Issue #1824, opened 2020-05-12
garden exec is slow


mnpenner:
### Current Behavior

When I run `garden exec php bash`, garden spends an awful lot of time "Getting status". Even after it's done there's a very long pause before dropping me into my shell. Pretty sure docker on its own was way quicker, unless kubernetes adds some kind of overhead?

### Expected behavior

Much quicker to get an interactive shell.

### Reproducible example

For my test I'm using a docker image built on `php:7.1-fpm`. 

### Workaround

*Maybe* there's a way to bypass garden to get into the container? I don't know what the exact command is though, that's why I'm using garden 😋 

### Suggested solution(s)

Make it fassttrrrr. Maybe spend less time probing the current state of things and just try executing the command? If that fails, *then* run some diagnosis?

### Your environment

`garden version` 0.11.13

`kubectl version` Client Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.6-beta.0", GitCommit:"e7f962ba86f4ce7033828210ca3556393c377bcc", GitTreeState:"clean", BuildDate:"2020-01-15T08:26:26Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.6-beta.0", GitCommit:"e7f962ba86f4ce7033828210ca3556393c377bcc", GitTreeState:"clean", BuildDate:"2020-01-15T08:18:29Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}

`docker version` 19.03.8

mnpenner:
When I click this little button in Docker for Windows it opens a shell just about instantly:

![](https://mpen.xyz/share/2020/05/phpstorm64_2020-05-12_20-36-35.png)

So I'm pretty sure Garden is adding overhead.

eysi09:
Hi @mnpenner and thanks for the issue!

You're right that Garden adds some overhead. For most commands, Garden has to scan for configuration files, resolve the configuration and template strings and get the status of the providers. In the context of the Kubernetes provider, this means e.g. checking that system services are available. For commands like `build` and `deploy`, this may not be as noticeable, but it certainly is for commands like `exec`. 

We've been exploring ways to sidestep this initial overhead, for example by having Garden run as a daemon or a long running process that can execute actions. However, these are not on our immediate roadmap.

[...]

PR #1838, opened 2020-05-23
perf(analytics): don't wait for event tracks

eysi09:

Previously, we would wait for the `analytics.trackCommand` and `analytics.trackApi` calls to finish. This could cause a delay after Garden is finished resolving the providers and before it executes a given command.

Now we just call the track handlers synchronously and wait for pending events when Garden exits (generally there should be no pending events because the first event is flushed immediately and subsequent events every 300 ms).

It also adds some tests for the `AnalyticsHandler` class and one for the `CLI` class to verify that events are flushed before Garden exits.

Partially addresses #1824 

- I changed the `generateProjectMetadata` function to only use raw module configs. This way it doesn't have to resolve all the modules which is unneeded overhead.

- The unit tests are to a degree verifying the behaviour of a 3rd party library. That certainly feels like scope creep but I'd rather test this properly and risk not going by the book.

- We talked about forking the `analytics-node` lib to prevent that "flush on first event" issue. That could still lead to dropped events so I instead went for the `waitForPending` approach. The code comments should explain what that's all about.

### DIFF
--- a/garden-service/src/analytics/analytics.ts
+++ b/garden-service/src/analytics/analytics.ts
@@ -9,9 +9,9 @@
 import segmentClient = require("analytics-node")
 import { platform, release } from "os"
 import ci = require("ci-info")
-import { flatten } from "lodash"
+import { uniq } from "lodash"
 import { globalConfigKeys, AnalyticsGlobalConfig, GlobalConfigStore, GlobalConfig } from "../config-store"
-import { getPackageVersion, uuidv4 } from "../util/util"
+import { getPackageVersion, uuidv4, sleep } from "../util/util"
 import { SEGMENT_PROD_API_KEY, SEGMENT_DEV_API_KEY } from "../constants"
 import { LogEntry } from "../logger/log-entry"
 import hasha = require("hasha")
@@ -94,7 +94,7 @@ interface AnalyticsEvent {
   properties: AnalyticsEventProperties
 }
 
-interface SegmentEvent {
+export interface SegmentEvent {
   userId: string
   event: AnalyticsType
   properties: AnalyticsEventProperties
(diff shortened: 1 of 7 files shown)


garden-service/src/analytics/analytics.ts
```diff
@@ -314,24 +318,19 @@ export class AnalyticsHandler {
...
-            resolve(true)
-          })
-        })
-      }
-
-      return trackToRemote(segmentEvent)
+      const eventUid = uuidv4()
+      this.pendingEvents.set(eventUid, segmentEvent)
+      this.segment.track(segmentEvent, (err: any) => {
+        this.pendingEvents.delete(eventUid)
```
10ko, line 324: Why do you delete the event from `pendingEvents` before checking if any erorr occurred?

garden-service/src/analytics/analytics.ts
```diff
@@ -431,19 +430,57 @@ export class AnalyticsHandler {
...
+        }
+        return
+      } else {
+        return waitForPending(retry + 1)
+      }
+    }
+
+    await this.segmentFlush()
+
+    if (this.pendingEvents.size === 0) {
```
10ko, line 467: Just to double check: the `segmentFlush()` internally will call the callback which will delete the events from our `pendingEvents` lists right?

eysi09: Yup, it calls all of them.

garden-service/src/analytics/analytics.ts
eysi09: Because we want to remove the event regardless. If we don't, and Segment errors, we'll have pending events when Garden exits and we'll wait for them to clear which will never happen (since they're no longer in the Segment queue).

That being said, we could make a new map of error events and attempt to flush those at exit, but that feels a little out of scope for this PR. At least that wasn't the initial behaviour. 

Would be useful though when we start storing pending events in the DB on exit.

eysi09:
@10ko, is anything unresolved?

10ko:
Thanks for replying to my comments @eysi09.
LGTM after you solved the conflicts!
❤️

PR #1847, opened 2020-05-27
fix(k8s): issue with GCR auth when running in-cluster builds on GKE

edvald:
fix(k8s): issues with GCR auth when running in-cluster builds on GKE

This adds a new `gke` example project, with instructions on how to set
up GCR auth with Kaniko. The procedure is unfortunately a bit involved,
but it does follow the suggested best practice in the Kaniko docs:
https://github.com/GoogleContainerTools/kaniko#pushing-to-gcr-using-workload-identity

Once #2216 is merged, we should fold the relevant parts of the README in the new `gke` example project into the in-cluster building guide.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -700,6 +700,7 @@ workflows:
           <<: *only-internal-prs
           context: docker
           requires: [build-dist]
+
       - test-docker-gcloud:
           <<: *only-internal-prs
           context: docker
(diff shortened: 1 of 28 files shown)


RetWolf:
Is there any issue to track for this, or should I go ahead and create one? Currently running into this issue as well, would also like to test if Artifact Registry (GCRs eventual replacement) works out of the box or if there might need to be extra configuration done to support that as well.

PR #1848, opened 2020-05-27
chore: update packages, and update TypeScript to 3.9

edvald:


edvald:
@thsig this should fix the build failures for the other PRs. Not sure why this came up, but updating packages seems to have done the trick.

PR #1849, opened 2020-05-28
feat(platform): support non-interactive auth

thsig:

Added support for non-interactive authentication with the platform via a new environment variable: `GARDEN_AUTH_TOKEN`.

In some contexts where interactive login is not feasible (e.g. CI), a valid client auth token may still be available before logging in,

Furthermore, storing the client auth token in the local config DB is not always desirable in these contexts.

To handle these situations, a client auth token can now be provided to garden commands making use of platform features (such as secrets) by populating the `GARDEN_AUTH_TOKEN` environment variable with the client auth token before running the command.


eysi09:
Perhaps out of scope for this one but it'd be great to have a centralised place where we validate and import all env vars from.

edvald:
@thsig tests are failing. I can take a look once that is sorted.

thsig:
Ah, I see. Will have a look.

thsig:
@edvald: Tests are passing now.

PR #1850, opened 2020-05-28
fix: properly connect to remote in proxy

stevenmatthewt:

Due to [this behavior in Node](https://github.com/nodejs/node-v0.x-archive/commit/c11612026fa28f7aedc60c577120f87d86fc15bf), the `'connect'` event will not be emitted on the `local` socket when the tcp connection is first established. Instead, the connection to the remote won't occur until `'data'` is received on the local socket. This generally happens to work, but won't work in the situation where the remote actually proceeds to send data immediately following a connection, and the client does not. I noticed this when debugging issues with MySQL connections through the proxy.

This change simply frontloads the `getRemote()` work to occur immediately after the `net.Server`s `connectionListener` is executed. The connection has already been established at this point, so it's appropriate to contact the remote.

N/A

**Steps to reproduce original issue**:

@edvald @eysi09 

Using the following minimal `garden.yml`:
```
kind: Project
name: test-remote
environments:
  - name: local
    providers:
    - name: local-kubernetes
      defaultHostname: localhost.com
      environments: [local]
---

kind: Module
description: MySQL container for testing
type: container
name: mysql
image: mysql:latest
services:
  - name: mysql
    args:
      - "mysqld"
      - "--bind-address=0.0.0.0"
    ports:  # Expose the default MySQL port
      - name: db
        containerPort: 3306
    env:
      MYSQL_ROOT_PASSWORD: test
```

Running latest master's `garden dev` will result in a non-responsive mysql connection:
```
❯ mysql -h127.0.0.1 -uroot -ptest
mysql: [Warning] Using a password on the command line interface can be insecure.
```

Running this branch's version of `garden dev`:
```
❯ mysql -h127.0.0.1 -uroot -ptest
mysql: [Warning] Using a password on the command line interface can be insecure.
Welcome to the MySQL monitor.  Commands end with ; or \g.
Your MySQL connection id is 8
Server version: 8.0.20 MySQL Community Server - GPL

Copyright (c) 2000, 2020, Oracle and/or its affiliates. All rights reserved.

Oracle is a registered trademark of Oracle Corporation and/or its
affiliates. Other names may be trademarks of their respective
owners.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

mysql>
```


eysi09:
Thank you for this PR @stevenmatthewt!

We have a bank holiday this weekend in Germany but we'll get to reviewing this as soon we're back at the (virtual) office!

PR #1860, opened 2020-06-04
improvement(core): don't respect .gitignore files by default

edvald:

The previous default behavior has been known to cause confusion, and
sometimes makes it difficult to correctly control which files should
be used for build contexts. We now default to only respect
`.gardenignore` files.

You can override the default behavior by explicitly setting the
`dotIgnoreFiles` field in your project configuration.

The `garden create project` command now also copies a .gitignore file
next to the project config file, if one is found, and instructs the
user accordingly.

I also added a bunch of .gardenignore files (and sometimes missing
.gitignore files) to our examples.

BREAKING CHANGE:

Garden no longer respects `.gitignore` files by default now. If you'd
like to retain the previous default behavior, you can explicitly set
`dotIgnoreFiles: [.gitignore, .gardenignore]` in your project configs.
If you already have `dotIgnoreFiles` set in your config, no change is
necessary.

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -94,14 +94,14 @@ providers:
 defaultEnvironment: ''
 
 # Specify a list of filenames that should be used as ".ignore" files across the project, using the same syntax and
-# semantics as `.gitignore` files. By default, patterns matched in `.gitignore` and `.gardenignore` files, found
-# anywhere in the project, are ignored when scanning for modules and module sources.
+# semantics as `.gitignore` files. By default, patterns matched in `.gardenignore` files, found anywhere in the
+# project, are ignored when scanning for modules and module sources (Note: prior to version 0.12.0, `.gitignore` files
+# were also used by default).
 # Note that these take precedence over the project `module.include` field, and module `include` fields, so any paths
 # matched by the .ignore files will be ignored even if they are explicitly specified in those fields.
 # See the [Configuration Files
 # guide](https://docs.garden.io/guides/configuration-files#including-excluding-files-and-directories) for details.
 dotIgnoreFiles:
-  - .gitignore
   - .gardenignore
 
 # Control where to scan for modules in the project.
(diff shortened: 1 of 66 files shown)


docs/reference/config.md
```diff
@@ -94,14 +94,14 @@ providers:
 defaultEnvironment: ''
 
 # Specify a list of filenames that should be used as ".ignore" files across the project, using the same syntax and
-# semantics as `.gitignore` files. By default, patterns matched in `.gitignore` and `.gardenignore` files, found
-# anywhere in the project, are ignored when scanning for modules and module sources.
+# semantics as `.gitignore` files. By default, patterns matched in `.gardenignore` files, found anywhere in the
+# project, are ignored when scanning for  modules and module sources (Note: prior to version 0.12.0, `.gitignore`
```
thsig, line 98: There's a accidental double space between "for" and "modules" (in "for modules and module sources").

garden-service/src/commands/create/create-project.ts
```diff
@@ -159,10 +159,25 @@ export class CreateProjectCommand extends Command<CreateProjectArgs, CreateProje
...
-        chalk.green(`-> Created .gardenignore file at ${chalk.bold.white(relative(process.cwd(), ignoreFilePath))}`)
-      )
+      const gitIgnorePath = resolve(configDir, ".gitignore")
+
+      if (await pathExists(gitIgnorePath)) {
+        await copyFile(gitIgnorePath, ignoreFilePath)
+        const gitIgnoreRelPath = chalk.bold.white(relative(process.cwd(), ignoreFilePath))
+        log.info(
+          chalk.green(
+            `-> Copied the .gitignore file at ${gitIgnoreRelPath} to a new .gardenignore in the same directory. Please edit that if you'd like Garden to include or ignore different files.`
```
thsig, line 169: I'd suggest replacing "Please edit that" with "Please edit the .gardenignore file" to make this even clearer (since this command is somewhat directed at the novice user).

edvald: Oops, wrong reaction, haha

edvald:
Thanks! I'm not sure we'll have or need a migration command, but if we need it for something else we can look into adding this.

garden-service/src/commands/create/create-project.ts
thsig: Hahaha

edvald: Doesn't happen much IRL, putting up the wrong facial expression, or pointing your thumb the wrong way.

Anyway, this has been updated  ☹️ ✨

PR #1864, opened 2020-06-09
feat(core): allow custom local db path

thsig:

A custom path to Garden's config DB can now be set via the `GARDEN_CONFIG_DB_PATH` variable.

There are no tests for this functionality (since I couldn't think of a way to test this automatically without getting into a lot of gymnastics).

But the semantics are very simple, and any errors should be obvious to the user during usage.

### DIFF
--- a/garden-service/src/db/connection.ts
+++ b/garden-service/src/db/connection.ts
@@ -6,11 +6,13 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import { Connection, getConnectionManager, ConnectionOptions } from "typeorm"
 import { join } from "path"
+import { Connection, getConnectionManager, ConnectionOptions } from "typeorm"
 import { GARDEN_GLOBAL_PATH } from "../constants"
 
 let connection: Connection
+
+const databasePath = join(process.env.GARDEN_DB_DIR || GARDEN_GLOBAL_PATH, "db")
 
 // Note: This function needs to be synchronous to work with the typeorm Active Record pattern (see ./base-entity.ts)
 export function getConnection(): Connection {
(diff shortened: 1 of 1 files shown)


garden-service/src/db/connection.ts
```diff
@@ -6,12 +6,15 @@
...
 
-import { Connection, getConnectionManager, ConnectionOptions } from "typeorm"
 import { join } from "path"
+import { Connection, getConnectionManager, ConnectionOptions } from "typeorm"
 import { GARDEN_GLOBAL_PATH } from "../constants"
 
 let connection: Connection
 
+// Note: TypeORM / the JS SQLite library will create any intermediate directories needed.
+const databasePath = process.env.GARDEN_CONFIG_DB_PATH || join(GARDEN_GLOBAL_PATH, "db")
```
edvald, line 16: I think "config db" is a bit of a misnomer, since the db is meant for a broader set of uses. I think just `GARDEN_DB_PATH` would be better. Also, the join should also be applied to the provided env var value imo. `db` is a file, not a directory.

thsig: Aha, gotcha. Will make that change.

edvald: The comment added is also unnecessary/misleading then btw

thsig: I've updated the PR now.

Btw, the original idea was to also allow the user to also the name of the db file (SQLite doesn't seem to care if it's actually called `db`), so that they could e.g. have a temp dir with various db files (e.g. `/my/temp/dir/` containing `db-123` + `db-456`).

But I agree that it's better to stick with database files with the same name (`db`) and just provide for custom directory paths instead.

edvald: Yeah, that also makes sense of course, but I anticipate we might have a set of files at some point instead of just the one db file, so this looks good :+1:

PR #1867, opened 2020-06-10
fix(workflows): abort if step cmd returns errors

thsig:

This fixes an oversight in the initial implementation, where succeeding steps were only aborted if the step threw an exception, but not if it had an an error in its command result.

We now abort subsequent steps if a step command has errors in its result, and log the error details directly in the log (since looking into `error.log` is an undesirable extra step e.g. in a CI context).

### DIFF
--- a/garden-service/src/commands/run/workflow.ts
+++ b/garden-service/src/commands/run/workflow.ts
@@ -8,12 +8,13 @@
 
 import chalk from "chalk"
 import { cloneDeep, isEqual, merge, repeat, take } from "lodash"
-import { printHeader, getTerminalWidth } from "../../logger/util"
+import { printHeader, getTerminalWidth, formatGardenError } from "../../logger/util"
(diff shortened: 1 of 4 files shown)


garden-service/src/commands/run/workflow.ts
```diff
@@ -65,6 +67,11 @@ export class RunWorkflowCommand extends Command<Args, {}> {
       } catch (err) {
         throw err
       }
+      if (result.errors) {
+        logErrors(log, result.errors, index, steps.length, step.description)
+        // We don't include errors in the result, since we're handling the detailed logging ourselves in this command.
+        return {}
```
edvald, line 73: Doesn't this mean that the workflows command will return with code 0, even if it fails?

thsig: Aha, good point. I'll fix that.

thsig: Done.

PR #1869, opened 2020-06-10
feat(workflows): allow writing files ahead of workflow execution

edvald:

This allows writing files from either string data, or referencing secrets when using Garden Enterprise, ahead of running workflows. 

The files are resolved and written before resolving providers, so they can reference the files. A common use case would e.g. be to write authentication files like kubeconfigs, which providers can then use.

### DIFF
--- a/garden-service/test/unit/src/commands/run/workflow.ts
+++ b/garden-service/test/unit/src/commands/run/workflow.ts
@@ -9,27 +9,35 @@
 import execa from "execa"
 import tmp from "tmp-promise"
 import { expect } from "chai"
-import { TestGarden, makeTestGardenA, withDefaultGlobalOpts } from "../../../../helpers"
+import { TestGarden, makeTestGardenA, withDefaultGlobalOpts, expectError } from "../../../../helpers"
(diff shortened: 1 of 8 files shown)


garden-service/test/unit/src/commands/run/workflow.ts
```diff
@@ -178,7 +179,138 @@ describe("RunWorkflowCommand", () => {
...
+        path: garden.projectRoot,
+        files: [{ path: "garden.yml/foo.txt", data: "foo" }],
+        steps: [{ command: ["get", "outputs"] }],
+      },
+    ])
+
+    await expectError(
+      () => cmd.action({ ...defaultParams, args: { workflow: "workflow-a" } }),
+      (err) =>
+        expect(err.message).to.equal("Unable to write file 'garden.yml/foo.txt': EEXIST: file already exists, mkdir")
```
thsig, line 292: This assertion is failing in CI, we should be good to merge this PR once that's fixed (either by changing this to `.to.match` or by templating in the expected absolute path).

PR #1875, opened 2020-06-13
Feature/configurable kaniko image

swist:
This PR lets us override the image used for the kaniko builder in builder config

Fixes #

### DIFF
--- a/garden-service/src/plugins/kubernetes/config.ts
+++ b/garden-service/src/plugins/kubernetes/config.ts
@@ -27,6 +27,7 @@ import { baseTestSpecSchema, BaseTestSpec } from "../../config/test"
 import { ArtifactSpec } from "../../config/validation"
 import { V1Toleration } from "@kubernetes/client-node"
 
+export const DEFAULT_KANIKO_IMAGE = "gcr.io/kaniko-project/executor:debug-v0.23.0"
 export interface ProviderSecretRef {
   name: string
   namespace: string
@@ -90,6 +91,9 @@ export interface KubernetesConfig extends ProviderConfig {
   buildMode: ContainerBuildMode
   clusterDocker?: {
     enableBuildKit?: boolean
+  }
+  kaniko?: {
+    image?: string
   }
   context: string
   defaultHostname?: string
@@ -327,6 +331,20 @@ export const kubernetesConfigBase = providerConfigBaseSchema().keys({
     })
     .default(() => {})
     .description("Configuration options for the `cluster-docker` build mode."),
+  kaniko: joi
+    .object()
+    .keys({
+      image: joi
+        .string()
+        .default(DEFAULT_KANIKO_IMAGE)
+        .description(
+          deline`
+            Change the kaniko image (repository/image:tag) to use when building in kaniko mode.
+          `
+        ),
+    })
+    .default(() => {})
+    .description("Configuration options for the `kaniko` build mode."),
   defaultHostname: joi
     .string()
     .description("A default hostname to use when no hostname is explicitly configured for a service.")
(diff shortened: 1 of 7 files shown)


garden-service/src/plugins/container/config.ts
```diff
@@ -546,6 +546,7 @@ export interface ContainerBuildSpec extends BaseBuildSpec {
 export interface ContainerModuleSpec extends ModuleSpec {
   build: ContainerBuildSpec
   buildArgs: PrimitiveMap
+  builderArgs?: PrimitiveMap
```
edvald, line 549: I think this would confuse users quite a bit. I'd prefer something more explicit, like `kaniko.extraArgs`, and making `kaniko` an object. That's going to scale better, and it's good to make it so specific because flags won't generally be portable across different builders.

I think it should also be a simple string array, and not an object. The object makes sense for `buildArgs` because that's intended as a key-value map in Docker.

garden-service/src/plugins/kubernetes/config.ts
```diff
@@ -327,6 +331,20 @@ export const kubernetesConfigBase = providerConfigBaseSchema().keys({
...
     .description("Configuration options for the `cluster-docker` build mode."),
+  kaniko: joi
+    .object()
+    .keys({
+      image: joi
+        .string()
+        .default(DEFAULT_KANIKO_IMAGE)
+        .description(
+          deline`
+            Change the kaniko image version to use when building in kaniko mode.
```
edvald, line 342: Maybe "image version" is confusing here, because that would indicate that you're setting a version or tag, as opposed to the full image ID.

garden-service/src/plugins/container/config.ts
swist: Wired `extraFlags` instead

garden-service/src/plugins/kubernetes/config.ts
swist: Fixed and added an example

garden-service/test/integ/src/plugins/kubernetes/container/build.ts
```diff
@@ -428,4 +428,21 @@ describe("kubernetes build flow", () => {
       expect(status.ready).to.be.false
     })
   })
+
+  grouped("kaniko", "image-override").context("kaniko-image-override mode", () => {
+    before(async () => {
+      await init("kaniko-remote-registry")
```
mitchfriedman, line 434: should this be `kaniko-image-override`?

swist: Good spot!

garden-service/src/plugins/kubernetes/container/build.ts
```diff
@@ -336,7 +337,16 @@ const remoteBuild: BuildHandler = async (params) => {
       dockerfile,
       "--destination",
       deploymentImageId,
-      "--cache=true",
+      ...uniqBy(["--cache=true", ...module.spec.extraFlags], (flag) => {
```
edvald, line 340: Need to add a conditional here because extraFlags may be undefined: 
`...uniqBy(["--cache=true", ...module.spec.extraFlags || []], (flag) => {`

That aside, this kinda feels like it could use a quick unit test, to make sure the deduplication works as expected. But the code seems fine to me.

garden-service/test/integ/src/plugins/kubernetes/container/build.ts
```diff
@@ -428,4 +428,21 @@ describe("kubernetes build flow", () => {
       expect(status.ready).to.be.false
     })
   })
+
+  grouped("kaniko", "image-override").context("kaniko-image-override mode", () => {
```
edvald, line 432: Ah, that's where the error comes from: This should also include a `"remote-only"` group because it relies on some particular data that isn't available for external contributors. We should have documented that btw, so that's on us.

swist: Cool, updated!

PR #1880, opened 2020-06-17
feat(workflows): improved log & event streaming

thsig:

Added `workflowStep` metadata to log entries generated by step commands and scripts.

Added step status events, which are emitted when a workflow step starts processing, completes or fails.

Added support for an optional `GARDEN_WORKFLOW_UID` environment variable that is sent along with streamed events & log entries (if it's defined).


eysi09:
Just one question, why do we need the env var? Also, we should setup a fixed entry point to all the env vars at some point and use something like the `env-var` lib.

thsig:
We need the env var for the backend to know which workflow run the log entries & events belong to.

I definitely agree about the fixed entrypoint / import location for env vars.

PR #1883, opened 2020-06-17
fix(helm): allow runtime templates in helm module values field

### LINKED ISSUE #1881: [FEATURE]: Support runtime outputs for Helm chart values
### Background / Motivation

Today, we use a jungle of bash scripts to pass Cloudformation and Terraform outputs to our Helm charts. Typically, these values are indeterminate until deploy time (e.g. AWS RDS endpoint URLs). We were looking to use Garden to streamline handling these outputs.

A Garden user should be able to reference runtime outputs in their Helm chart values. e.g.

Terraform that creates an SNS topic
```
----
kind: Module
type: terraform
name: tf-infra
```

```
---
kind: Module
type: helm
.
.
.
serviceResource:
    containerModule: docker-image
dependencies:
    - tf-infra
values:
    microservice:
        appConfig:
            environment:
                SNS_TOPIC_ARN: ${runtime.services.tf-infra.outputs.sns_topic_arn}
              # or
              # SNS_TOPIC_ARN: ${providers.terraform.outputs.sns_topic_arn}
        image:
            repository: ${modules.docker-image.outputs.deployment-image-name}
            tag: ${modules.docker-image.version}
```

It allows runtime outputs from Terraform (or other modules such as exec) to be resolvable at deploy time.

* Allow the Helm module `values:` stanza to resolve runtime outputs.
* Add an additional `setValues:` stanza to the Helm module to allow runtime variables to be passed to Helm via `--set` at deploy time.

We want to take a machete to our jungle of bash. Please help! :)

edvald:

We now render the values we feed into Helm on every render, instead of
only at build-time. This unblocks the usage of `${runtime.*}` template
strings, which in turns unblocks several common use-cases, including
feeding in outputs from Terraform modules.

This required making a specific conftest module type that is Helm aware.
Once we've implemented a render/plan phase in our framework, we should
be able to get rid of that specific implementation.

Fixes #1881

cc @namron86


garden-service/src/plugins/conftest/conftest.ts
```diff
@@ -161,76 +164,85 @@ export const gardenPlugin = createGardenPlugin({
...
+      docs: dedent`
+        Special module type for validating helm modules with conftest. This is necessary in addition to the \`conftest\` module type in order to be able to properly render the Helm chart ahead of validation, including all runtime values.
 
-          const allFailures = parsed.filter((p: any) => p.Failures?.length > 0)
-          const allWarnings = parsed.filter((p: any) => p.Warnings?.length > 0)
+        If the helm module requires runtime outputs from other modules, you must list the corresponding dependencies with the \`runtimeDependencies\` field.
 
-          const resultCategories: string[] = []
-          let formattedResult = "OK"
+        > Note: In most cases, you'll let specific the [\`conftest-kubernetes\`](${kubernetesProviderUrl}) provider create this module type automatically, but you may in some cases want or need to manually specify files to test.
```
eysi09, line 194: Typo: _you'll let specific the_

PR #1884, opened 2020-06-18
feat(workflows): support referencing outputs from previous steps

edvald:

Workflow steps can now reference outputs from previous steps. This
allows for much more flexible workflows, especially when using script steps.

This turned out to be fairly involved, largely due to the previous lack
of solid and defined command output structure. This has now been defined
better, and some output structures improved. I also made a bunch more
commands available for workflows.

The templating for workflow steps is now resolved just-in-time, such
that previous steps can be referenced. I also removed providers from
the workflow template contexts, because that was getting in the way of
using scripts to prepare e.g. auth for provider configuration.

BREAKING CHANGE:

The JSON/YAML output from the build, deploy and test commands has been
modified. The prior format is now nested under the `graphResults` key,
and new more structured fields have been added, which we recommend using
when possible.

Similarly, the outputs from the `run` commands have been modified. The
`result` key now contains the data from the previous format, along with
some additional metadata.

### DIFF
--- a/garden-service/src/commands/base.ts
+++ b/garden-service/src/commands/base.ts
@@ -277,7 +277,7 @@ export abstract class Command<T extends Parameters = {}, U extends Parameters = 
   arguments?: T
   options?: U
 
-  outputsSchema?: () => Joi.Schema
+  outputsSchema?: () => Joi.ObjectSchema
 
   cliOnly: boolean = false
   hidden: boolean = false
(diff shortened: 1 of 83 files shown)


garden-service/src/commands/base.ts
```diff
@@ -458,9 +484,59 @@ export async function handleTaskResult({
...
+      "A map of all services that were deployed (or deployment scheduled/attempted for) and the service status."
+    ),
+    tests: joiStringMap(testResultSchema().keys(resultMetadataKeys())).description(
+      "A map of all tests that were run (or scheduled/attempted) and the test results."
+    ),
+    graphResults: graphResultsSchema(),
+  })
+
 /**
  * Handles the command result and logging for commands the return results of type ProcessResults.
```
thsig, line 541: Not introduced in this PR, but I just noticed a typo here. Should be "that return".

edvald:
Should be good to go now. I'll do docs in a separate branch, along with other doc updates.

PR #1886, opened 2020-06-18
fix(build-sync): randomly choose pod for rsync

swist:
Turns out it was not random enough in our cluster - it always
picked the same pod for us. This ensures that we do not pick
the same pod over and over again if k8s api ends up returning
the pods in the same order

This might seem a bit odd at first (i.e. we'd also expect that portForward gives you whatever node on a deployment) but we have verified this behaviour by checking the logs on each of the builder pods - only one ever had any activity!


garden-service/src/plugins/kubernetes/container/build.ts
```diff
@@ -221,14 +221,25 @@ const remoteBuild: BuildHandler = async (params) => {
...
 
+  const buildSyncPod = await getRunningPodInDeployment(buildSyncDeploymentName, provider, log)
+
+  // TODO: remove this after a few releases (from 0.10.15), since this is only necessary for environments initialized
+  // with 0.10.14 or earlier.
+  if (!buildSyncPod) {
+    throw new PluginError(`Could not find running build sync Pod`, {
+      deploymentName: buildSyncDeploymentName,
+      systemNamespace,
+    })
```
swist, line 232: I suspect that this check would actually make more sense inside getRunningPodInDeployment (it makes sense for it to only return running pods or throw if it can't get one)

swist: Ended up refactoring this - turned out that all the places that were getting a pod in a running deployment, required it to come back truthy

garden-service/src/plugins/kubernetes/util.ts
```diff
@@ -395,14 +395,21 @@ export function convertDeprecatedManifestVersion(manifest: KubernetesResource):
...
-export async function getRunningPodInDeployment(deploymentName: string, provider: KubernetesProvider, log: LogEntry) {
+export async function getDeploymentPodName(deploymentName: string, provider: KubernetesProvider, log: LogEntry) {
   const api = await KubeApi.factory(log, provider)
   const systemNamespace = await getSystemNamespace(provider, log)
 
   const status = await api.apps.readNamespacedDeployment(deploymentName, systemNamespace)
   const pods = await getPods(api, systemNamespace, status.spec.selector.matchLabels)
+  const pod = sample(pods)
+  if (!pod) {
+    throw new PluginError(`Could not a running pod in a deployment`, {
```
edvald, line 406: Would be good to put the deployment name in the error message here.

PR #1887, opened 2020-06-19
feat: add dataFile option to conftest module

lucas-sonnabend:
you can add a data path that will convert it into the pass that is
passed as --data into conftest

We've started using the conftest capabilities of garden more, and are missing some features that plain conftest provides. One of them is to pass in extra data via the --data flag

### DIFF
--- a/garden-service/src/plugins/conftest/conftest.ts
+++ b/garden-service/src/plugins/conftest/conftest.ts
@@ -55,6 +55,7 @@ interface ConftestModuleSpec {
   policyPath: string
   namespace: string
   files: string[]
+  data: string
   sourceModule: string
 }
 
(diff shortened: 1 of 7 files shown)


garden-service/src/plugins/conftest/conftest.ts
```diff
@@ -120,6 +121,14 @@ export const gardenPlugin = createGardenPlugin({
               A list of files to test with the given policy. Must be POSIX-style paths, and may include wildcards.
             `
           ),
+        data: joi
```
edvald, line 124: Isn't it possible to feed multiple `--data` options to conftest? If so, it'd be good to make this an array. Also, to allow wildcards you need to add `.allowGlobs()` to the schema.

That aside, PR looks good, thanks @lucas-sonnabend :)

lucas-sonnabend: as far as I understand it it's just a single file, but I can try multiple one, too!

lucas-sonnabend:
I might shelf this branch for now:
It turns out I don't really need the --data option at the moment, and I'm having trouble running tests locally on ubuntu. :/

edvald:
I'm happy to help wrapping this up, since it's pretty close to being ready.

PR #1888, opened 2020-06-19
feat(kaniko): when using remote registry do not run the proxy

swist:
We currently run a bunch of noop containers that proxy kaniko to in-cluster registry, even if in-cluster registry is switched off. 

This means when running heavier loads, you're less likely to
starve out of PIDs. Also means the builds should start faster

We know that these are safe to remove because in our clusters we have manually deleted garden-registry and these just busyloop

### DIFF
--- a/garden-service/src/plugins/kubernetes/container/build.ts
+++ b/garden-service/src/plugins/kubernetes/container/build.ts
@@ -9,6 +9,7 @@
 import pRetry from "p-retry"
 import split2 = require("split2")
 import { differenceBy } from "lodash"
+import { V1PodSpec } from "@kubernetes/client-node"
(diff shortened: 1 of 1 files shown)


edvald:
Ah, actually, looks like there's a type error in there. The change otherwise looks fine, just something missing to set TypeScript at ease.

swist:
Is there a way I can disable the husky check for making sure links don't 404? My internet is too slow for this to ever finish so I find myself pushing with `no-verify` more often than I'd like to

edvald:
Ah yes, I've actually been wanting to rethink those pre-push checks. I find myself doing the same, just have a habit of running builds continuously with `npm run watch` in `garden-service` and `npm run lint` ahead of pushing. I'm not aware of a way to selectively disable husky checks.

PR #1889, opened 2020-06-19
Feature/lucas conftest combine option

lucas-sonnabend:

Turns what I really want is the --combine option. It makes it a lot easier to run conftest on several files and make assertions across those files. Without this you can only make assertions that are true for all the files passed in individually

### DIFF
--- a/docs/reference/module-types/conftest.md
+++ b/docs/reference/module-types/conftest.md
@@ -117,6 +117,9 @@ policyPath:
 # The policy namespace in which to find _deny_ and _warn_ rules.
 namespace: main
 
+# Set to true to use the conftest --combine flag
+combine: false
+
 # A list of files to test with the given policy. Must be POSIX-style paths, and may include wildcards.
 files:
 ```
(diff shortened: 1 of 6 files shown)


docs/reference/module-types/conftest.md
```diff
@@ -352,6 +356,14 @@ A list of files to test with the given policy. Must be POSIX-style paths, and ma
 | ------------------ | -------- |
 | `array[posixPath]` | Yes      |
 
+### `data`
```
edvald, line 359: This snuck in there from the other branch. Think you just need an `npm run generate-docs`. Otherwise looks good!

lucas-sonnabend: ah, good spot, I'll update it

lucas-sonnabend: I've run it again, but it didn't remove it.
Is there an intermediate result or step that has been cached, and that is leftover from my other branch?

edvald: You probably just need a rebuild as well. `npm run build` from the repo root should do, including the docs rebuild.

lucas-sonnabend:
@thsig I've got the integration tests passing, the remaining failures are on deploy openfaas. Do you think that is due to flakes?

PR #1891, opened 2020-06-22
fix(enterprise): use correct body param when streaming events

eysi09:

We were using the wrong param name for the event and log entry endpoint. Highlights the need for e2e tests. But this fixes it for now. 

N/A

Should we be referring to the project ID as project UID internally? It leaks the implementation details but it's still just internal, and I can imagine this happening again.


thsig:
We should be referring to the project UID as the project ID only in user-facing strings, IMHO.

PR #1895, opened 2020-06-23
feat(workflows): add namespacing support

thsig:

Added an optional `namespace` field to trigger specs.

We apply the same validation and default namespace rules for trigger namespaces as we do for environment specs.


edvald:
Linting is failing, but LGTM

PR #1899, opened 2020-06-24
feat(core): emit runtime status events

thsig:

We now emit `serviceStatus`, `testStatus` and `taskStatus` events in the appropriate `ActionRouter` methods.


garden-service/src/types/plugin/base.ts
```diff
@@ -124,3 +124,24 @@ export const artifactsPathSchema = () =>
...
+
+export type RunState = "outdated" | "succeeded" | "failed" | "not-implemented"
+
+export interface RunStatus {
+  state: RunState
+  startedAt?: Date
+  completedAt?: Date
+}
+
+export function runStatus<R extends RunResult>(result: R | null | undefined): RunStatus {
```
edvald, line 136: Kinda confusing function name. Maybe `getRunStatus`? Otherwise LGTM

thsig: Agreed—I've renamed it to `getRunStatus` now.

PR #1900, opened 2020-06-24
improvement(core): always require a namespace and simplify env config

edvald:
We now always require a namespace for all environments, and make the
default simply "default". You can override the default namespace with
`environments[].defaultNamespace` as before.

This reduces code branching significantly, especially in the enterprise
UI, and simplifies our configuration as well.

BREAKING CHANGE:

The default namespace in the `kubernetes` provider is now
`<project name>-<environment namespace>` (previously it was just the
project name). Users need to override this to `"${project.name}"` if
they would like to revert to the previous default.

### DIFF
--- a/docs/example-projects/using-garden-in-ci.md
+++ b/docs/example-projects/using-garden-in-ci.md
@@ -44,7 +44,7 @@ name: ci-demo-project
 environments:
   ...
   - name: preview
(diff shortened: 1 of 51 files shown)


edvald:
Yeah I see it, forgot to make a change in the K8s provider

eysi09:
@edvald, anything I can help out with. Kind of blocked on the release atm.

edvald:
Just updated, let's see if tests pass.

PR #1902, opened 2020-06-24
feat(template): allow integer keys in templates to reference array items

edvald:

Came across this while experimenting with step outputs. We had a
limitation where integer keys were not handled properly, so we could
not retrieve values from arrays in template strings. This simple fix
addresses that.


edvald:
Closing in favor of #1903

PR #1904, opened 2020-06-25
fix(hot-reload): fix path handling on Windows

thsig:

(Hopefully) fixes a regression in hot reloading on Windows.

This needs to be tested on a Windows machine.


thsig:
Nope, not yet—don't have a Windows machine handy. Was hoping @eysi09 could give it a spin when he's at the office.

eysi09:
On closer inspection, it looks the issue was the example not hot reloading per se. The original version works on the `hot-reload` example but not in the `gatsby-hot-reload` example. We'll need to look into the example separate but I'll close this for now.

PR #1910, opened 2020-06-26
feat(config): allow environment[].variables to reference top-level vars

### LINKED ISSUE #1814: Garden resolves template strings for environments that aren't in use
### Current Behavior

Garden resolves template strings for environments other than the one in use. This can cause issues, for example when some environment expects an env var that is not set in the _current_ environment. 

For example, a `ci` environment might have the following config:

```yaml
environments:
  - name: ci
     variables: ${local.env.SOME_VAL}
```

This will cause Garden to throw:

```console
Invalid template string ${local.env.SOME_VAL}: Could not find key SOME_VAL under local.env.
```  

for all envs that don't export `SOME_VAL`.

### Expected behavior

Garden shouldn't resolve template strings for environments other than the one in use.

### Reproducible example

In for example the `demo-project`, add the following to the project level `garden.yml`:

```yaml
kind: Project
name: demo-project
environments:
  - name: local
  - name: remote
  - name: bad-env
    variables:
      not-set: ${local.env.NOT_SET}
providers:
  # ...
```

and run a Garden command with `--env local|remote`. 

### Workaround

Use a fallback value, for example:

```
not-set: ${local.env.NOT_SET || ""}
```

or make it optional:

```
not-set: ${local.env.NOT_SET}?
```

### Suggested solution(s)

N/A

### Additional context

N/A

### Your environment

Latest `master`

edvald:
This makes it easier to keep your project config neat and concise, allowing for some deduplication when environment configs repeat a lot of variables.

Also fixes the issue where template strings were resolved in all
environments, instead of just the selected one.

Fixes #1814

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -65,7 +65,8 @@ environments:
     varfile:
 
     # A key/value map of variables that modules can reference when using this environment. These take precedence over
-    # variables defined in the top-level `variables` field.
+    # variables defined in the top-level `variables` field, but may also reference the top-level variables in template
+    # strings.
     variables: {}
 
 # A list of providers that should be used for this project, and their configuration. Please refer to individual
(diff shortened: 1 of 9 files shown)


edvald:
Small bug that's failing the tests there, easy fix though, will push that in a bit

PR #1915, opened 2020-06-26
bugfix(kaniko): correctly set container command when using kaniko

swist:
When running without socat we can't wait for socat to start.
I'm not sure how this passed the garden-io CI, it also magically
passed ours when we merged the update to the cli. This is to
quickly unbreak the master branch, I'll investigate further

Fixes #

### DIFF
--- a/garden-service/src/plugins/kubernetes/container/build.ts
+++ b/garden-service/src/plugins/kubernetes/container/build.ts
@@ -459,21 +459,26 @@ async function runKaniko({ provider, namespace, log, module, args, outputStream 
   // Escape the args so that we can safely interpolate them into the kaniko command
   const argsStr = args.map((arg) => JSON.stringify(arg)).join(" ")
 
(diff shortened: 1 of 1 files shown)


garden-service/src/plugins/kubernetes/container/build.ts
```diff
@@ -459,21 +459,31 @@ async function runKaniko({ provider, namespace, log, module, args, outputStream
...
+          export exitcode=$?;
+          touch ${commsMountPath}/done;
+          exit $exitcode;
+        else
+          sleep 0.3;
+        fi
+      done
+    `
+  } else {
+    commandStr = dedent`
```
edvald, line 480: Could define this variable above, and then interpolate into the socat block, just to avoid the duplication. Otherwise LGTM, and good catch!

Also a tiny detail on the commit message: The prefix should be `fix`, otherwise it won't get automatically picked up by our changelog generator

PR #1923, opened 2020-07-01
improvement(kaniko): allow configuring builder flags alongside image

swist:
This is helpful when you want to set a certain flag globally for
your entire deployment e.g. snapshotMode

Fixes #


garden-service/src/plugins/kubernetes/config.ts
```diff
@@ -341,6 +342,9 @@ export const kubernetesConfigBase = providerConfigBaseSchema().keys({
             Change the kaniko image (repository/image:tag) to use when building in kaniko mode.
           `
         ),
+      extraFlags: joi.array().items(joi.string()).description(deline`
+        Specify extra flags to use when building the container image with kaniko.
+        Flags set on container module take precedenece over these.`),
```
edvald, line 347: Typo: precedenece -> precedence

swist: Fixed!

PR #1925, opened 2020-07-02
docs: add supported github events

baspetersdotjpeg:

Documents the currently supported GitHub events that can be used as workflow triggers, as these currently aren't documented anywhere. We should probably do this automatically at some point.

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -986,7 +986,15 @@ triggers:
     # this trigger's environment, as defined in your project's environment configs.
     namespace:
 
-    # A list of GitHub events that should trigger this workflow.
+    # A list of [GitHub events](https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads)
+    # that should trigger this workflow.
+    #
+    # Supported events:
+    #
+    # `create`, `pull-request`, `pull-request-created`, `pull-request-updated`, `push`, `release`, `release-created`,
+    # `release-deleted`, `release-edited`, `release-prereleased`, `release-published`, `release-unpublished`
+    #
+    #
     events:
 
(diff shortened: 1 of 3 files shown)


docs/reference/config.md
```diff
@@ -1249,7 +1249,21 @@ The namespace to use for the workflow when matched by this trigger. Follows the
...
+  `[pull-request]`
+  `[pull-request-created]`
+  `[pull-request-updated]`
+  `[release]`
+  `[release-published]`
+  `[release-unpublished]`
+  `[release-created]`
+  `[release-edited]`
+  `[release-deleted]`
+  `[release-prereleased]`
```
edvald, line 1266: This file is generated automatically from schemas. So this should be added where the field is defined, and I'd also suggest using the `triggerTypes` constant to render the list in the field docs, rather than manually typing it up.

baspetersdotjpeg: Done.

garden-service/src/config/workflow.ts
```diff
@@ -213,8 +213,13 @@ export interface TriggerSpec {
   ignoreTags?: string[]
 }
 
-export const triggerSchema = () =>
-  joi.object().keys({
+export const triggerSchema = () => {
+  const eventDescriptions = triggerEvents
+    .sort()
+    .map((event) => `\`[${event}]\``)
```
edvald, line 219: Point of preference, but I prefer to skip this map with the brackets and backticks, since it's quite normal to use multiple event names for one trigger.

Also, need to run `npm run generate-docs` to update the generated markdown. Sorry, I know that's annoying, but it's an artifact of us using Gitbook + the generated markdown.

PR #1930, opened 2020-07-06
perf(git): cache git exec results

mitchfriedman:
Cache results of the `git exec calls`.

Pretty anecdotal/not-profile-driven perf improvements, but intuitively it makes sense anyways.

For example, on a project i'm working on I get over 200 of:
```
Calling git with args 'rev-parse --show-toplevel'
```
with > 700 git commands executing

before:
```
❯ time gfs validate
garden validate 42.44s user 63.68s system 237% cpu 44.720 total
```

after:
```
❯ time gfs validate                   
garden validate  40.08s user 52.55s system 279% cpu 33.179 total
```

Faster commands :)

### DIFF
--- a/garden-service/src/vcs/git.ts
+++ b/garden-service/src/vcs/git.ts
@@ -58,10 +58,11 @@ interface Submodule {
 @Profile()
 export class GitHandler extends VcsHandler {
   name = "git"
+  repoRoots = new Map()
 
   private gitCli(log: LogEntry, cwd: string): GitCli {
     return async (...args: string[]) => {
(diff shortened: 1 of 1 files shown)


garden-service/src/vcs/git.ts
```diff
@@ -81,10 +82,16 @@ export class GitHandler extends VcsHandler {
   }
 
   async getRepoRoot(log: LogEntry, path: string) {
+    if (this.repoRoot) {
+      return this.repoRoot
```
edvald, line 86: This needs to be cached based on the `path` value, because there may be multiple repo roots, e.g. with remote sources, submodules etc

PR #1936, opened 2020-07-09
chore: consolidate environment variables

thsig:

We now read and export most environment variables used by the framework in `constants.ts`, using the `env-var` library.

This should simplify tracking the environment variables used by the framework (compared to ad-hoc reading of `process.env.*`).

### DIFF
--- a/garden-service/src/cli/cli.ts
+++ b/garden-service/src/cli/cli.ts
@@ -48,7 +48,13 @@ import {
   checkForStaticDir,
 } from "./helpers"
 import { defaultEnvironments, ProjectConfig, defaultNamespace } from "../config/project"
-import { ERROR_LOG_FILENAME, DEFAULT_API_VERSION, DEFAULT_GARDEN_DIR_NAME, LOGS_DIR_NAME } from "../constants"
+import {
+  ERROR_LOG_FILENAME,
+  DEFAULT_API_VERSION,
+  DEFAULT_GARDEN_DIR_NAME,
+  LOGS_DIR_NAME,
+  gardenEnv,
+} from "../constants"
 import stringify = require("json-stringify-safe")
 import { generateBasicDebugInfoReport } from "../commands/get/get-debug-info"
 import { AnalyticsHandler } from "../analytics/analytics"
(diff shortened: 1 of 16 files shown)


garden-service/src/constants.ts
```diff
@@ -42,3 +43,66 @@ export const SEGMENT_PROD_API_KEY = "b6ovUD9A0YjQqT3ZWetWUbuZ9OmGxKMa"
...
+    .asString(),
+  GARDEN_DISABLE_ANALYTICS: env
+    .get("GARDEN_DISABLE_ANALYTICS")
+    .required(false)
+    .asBool(),
+  GARDEN_DISABLE_VERSION_CHECK: env
+    .get("GARDEN_DISABLE_VERSION_CHECK")
+    .required(false)
+    .asString(),
+  GARDEN_ENABLE_PROFILING: env
```
eysi09, line 78: Should we use `asBool` here? It coerces `"1"` and `"0"` to  `true` and `false` respectively.

garden-service/src/cli/cli.ts
```diff
@@ -526,7 +532,7 @@ export async function run(): Promise<void> {
     console.log(err)
     code = 1
   } finally {
-    if (process.env.GARDEN_ENABLE_PROFILING === "1") {
+    if (gardenEnv.GARDEN_ENABLE_PROFILING === "1") {
```
eysi09, line 535: We can skip the comparison if we treat this as bool, see my comment below.

garden-service/src/constants.ts
```diff
@@ -42,3 +43,66 @@ export const SEGMENT_PROD_API_KEY = "b6ovUD9A0YjQqT3ZWetWUbuZ9OmGxKMa"
...
+    .asString(),
+  GARDEN_LOG_LEVEL: env
+    .get("GARDEN_LOG_LEVEL")
+    .required(false)
+    .asString(),
+  GARDEN_LOGGER_TYPE: env
+    .get("GARDEN_LOGGER_TYPE")
+    .required(false)
+    .asString(),
+  GARDEN_SERVER_PORT: env
```
eysi09, line 90: We can use `asPortNumber` here.

garden-service/src/server/server.ts
```diff
@@ -45,7 +45,7 @@ export async function startServer(log: LogEntry, port?: number) {
   // Start HTTP API and dashboard server.
   // allow overriding automatic port picking
   if (!port) {
-    port = Number(process.env.GARDEN_SERVER_PORT) || undefined
+    port = Number(gardenEnv.GARDEN_SERVER_PORT) || undefined
```
eysi09, line 48: We can skip `Number()` if we coerce this to port number, see my comment above.

thsig:
I think I've addressed all the comments now. If this is good to go, I'll squash before merging.

PR #1938, opened 2020-07-10
fix(k8s): correct kaniko immutability check

mitchfriedman:
**Description**
Slight issue in the kaniko immutable repo check - there can be log lines in the build that indicate that _layers_ couldn't be pushed, but these shouldn't be treated as the source of truth for if the _image_ was pushed. 

This has been causing us problems because we might see that some layers fail to push for immutability reasons (which is fine, kaniko treats this as a warning), sometime later the build fails for other reasons (legitimate failures), but the logic here is catching those previous warnings and considering the build as passing when it shouldn't.

This makes this check more specific to ensure that only when the actual image failed to upload (the last step in the kaniko build) we consider the "failing" build to have passed. 
   
@edvald


garden-service/test/unit/src/plugins/kubernetes/container/build.ts
```diff
@@ -30,8 +30,8 @@ describe("kaniko build", () => {
     ).to.be.false
   })
 
-  it("should return as failure when other error messages are present", () => {
-    const errorMessage = "error pushing"
+  it.only("should return as failure when other error messages are present", () => {
```
thsig, line 33: Left an `.only` in there : )

PR #1940, opened 2020-07-13
feat(enterprise): register workflow runs

thsig:

When the user is logged in to Garden Enterprise, the `run workflow` command will now register a workflow run with the backend, and include the workflow run's UID in streamed events and log entries.

This enables `run workflow` command invocations from e.g. a local machine or an external CI system to be represented in Garden Enterprise.

Depends on #1936, so we should merge that first (I'll rebase on `master` and mark this PR as ready for review once that's done).

The `_workflowRunRegistered` control event seemed like the least hacky way to communicate the workflow run UID from the `run workflow` command to `BufferedEventStream`, but I'm open to other suggestions.

### DIFF
--- a/garden-service/src/cli/cli.ts
+++ b/garden-service/src/cli/cli.ts
@@ -341,21 +341,16 @@ export class GardenCli {
             garden = await Garden.factory(root, contextOpts)
           }
 
-          if (garden.clientAuthToken && garden.enterpriseDomain && garden.projectId) {
+          if (garden.enterpriseContext) {
             log.silly(`Connecting Garden instance to BufferedEventStream`)
             bufferedEventStream.connect({
               eventBus: garden.events,
-              clientAuthToken: garden.clientAuthToken,
-              enterpriseDomain: garden.enterpriseDomain,
-              projectId: garden.projectId,
+              enterpriseContext: garden.enterpriseContext,
               environmentName: garden.environmentName,
               namespace: garden.namespace,
             })
           } else {
             log.silly(`Skip connecting Garden instance to BufferedEventStream`)
-            log.silly(`clientAuthToken present: ${!!garden.clientAuthToken}`)
-            log.silly(`enterpriseDomain: ${garden.enterpriseDomain}`)
-            log.silly(`projectId: ${garden.projectId}`)
           }
 
           // Register log file writers. We need to do this after the Garden class is initialised because
--- a/garden-service/src/commands/run/workflow.ts
+++ b/garden-service/src/commands/run/workflow.ts
@@ -25,6 +25,8 @@ import { getDurationMsec } from "../../util/util"
 import { runScript } from "../../util/util"
 import { ExecaError } from "execa"
 import { LogLevel } from "../../logger/log-node"
+import { gardenEnv } from "../../constants"
+import { registerWorkflowRun } from "../../enterprise/workflow-lifecycle"
 
 const runWorkflowArgs = {
   workflow: new StringParameter({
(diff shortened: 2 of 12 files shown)


garden-service/src/commands/run/workflow.ts
```diff
@@ -63,6 +65,10 @@ export class RunWorkflowCommand extends Command<Args, {}> {
     const outerLog = log.placeholder()
     // Prepare any configured files before continuing
     const workflow = garden.getWorkflowConfig(args.workflow)
+    if (garden.useEnterprise && !gardenEnv.GARDEN_PLATFORM_SCHEDULED) {
```
eysi09, line 68: What does `GARDEN_PLATFORM_SCHEDULED` do?

garden-service/src/cli/cli.ts
```diff
@@ -341,13 +341,15 @@ export class GardenCli {
             garden = await Garden.factory(root, contextOpts)
           }
 
-          if (garden.clientAuthToken && garden.enterpriseDomain && garden.projectId) {
+          if (garden.useEnterprise) {
             log.silly(`Connecting Garden instance to BufferedEventStream`)
+            // If useEnterprise = true, we know that clientAuthToken, enterpriseDomain and projectId are defined.
```
eysi09, line 346: This feels a little brittle if we're not testing for it.

eysi09: Should we perhaps merge them into a object that has these fields required. Something like `garden.enterpriseConfig` that's either undefined or the all the things. Then you can for the existence of that above.

garden-service/src/enterprise/workflow-lifecycle.ts
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright (C) 2018-2020 Garden Technologies, Inc. <info@garden.io>
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { got, GotResponse } from "../util/http"
```
eysi09, line 9: Ahhh the temptation

garden-service/src/enterprise/workflow-lifecycle.ts
```diff
@@ -0,0 +1,61 @@
...
+export async function registerWorkflowRun({
+  workflowConfig,
+  clientAuthToken,
+  projectId,
+  environment,
+  namespace,
+  enterpriseDomain,
+  log,
+}: RegisterWorkflowRunParams): Promise<string> {
+  log.info(`Registering workflow run for ${workflowConfig.name}...`)
```
eysi09, line 37: Should this be info level? Not a strong opinion, just wanted to check if it was by accident.

garden-service/src/garden.ts
```diff
@@ -160,11 +161,15 @@ export class Garden {
...
-  // Platform-related instance variables
+  /**
+   * Platform-related instance variables
+   */
   public clientAuthToken: string | null
   public projectId: string | null
   public enterpriseDomain: string | null
   public sessionId: string | null
+  // True if the user is logged in and the project config defines a project id/uid and domain. False otherwise.
+  public useEnterprise: boolean
```
eysi09, line 172: See my comment above about bundling these together.

garden-service/src/cli/cli.ts
thsig: Yeah, that's a good idea.

garden-service/src/enterprise/workflow-lifecycle.ts
thsig: Hmm, on second thought this should probably be `debug`. I'll make that change.

garden-service/src/cli/cli.ts
```diff
@@ -341,21 +341,18 @@ export class GardenCli {
             garden = await Garden.factory(root, contextOpts)
           }
 
-          if (garden.clientAuthToken && garden.enterpriseDomain && garden.projectId) {
+          if (garden.enterpriseContext) {
             log.silly(`Connecting Garden instance to BufferedEventStream`)
+            // If useEnterprise = true, we know that clientAuthToken, enterpriseDomain and projectId are defined.
```
eysi09, line 346: We can remove this comment.

PR #1953, opened 2020-07-17
#1951 Added helm dependency update in case of requirements.yaml or Chart dependencies

### LINKED ISSUE #1951: Error while using helm with dependencies chart
### Current Behavior

When I use a helm chart with dependencies (requirements.yaml or Chart.yaml+dependencies) it is not downloading the chart dependencies anymore

Tried with version `0.11.14` and it works fine
It looks like on this change the template used to add `--dependency-update` but the new version with dry-run doesn't support it 

https://github.com/garden-io/garden/commit/26a87b9f5b07d1bbaccc39076d8e349eeec38b48#diff-22c65fc3306e692e526134d89fd8bc17L121 i

```

chart-db                 → Deploying version v-27d4bebaf9...

Failed deploying service 'chart-db' (from module 'chart-db'). Here is the output:
Command "/home/janario/.garden/tools/helm/dff97bc0c693fd1b/linux-amd64/helm 
--kube-context dev --namespace my-ns upgrade db /home/janario/Projects/my-project/.garden/build/chart-db 
--dry-run --namespace my-ns --output json --timeout 300s 
--values /home/janario/Projects/my-project/.garden/build/chart-db/values.yaml 
--values /home/janario/Projects/my-project/.garden/build/chart-db/garden-values.yml" failed with code 1:

Error: found in Chart.yaml, but missing in charts/ directory: mysql

Error: found in Chart.yaml, but missing in charts/ directory: mysql

```

### Expected behavior

It should automatically download helm dependencies 

### Reproducible example

Define a chart with dependencies
```
apiVersion: v2
appVersion: "1.0"
name: chart-db
version: 0.0.1
type: application
dependencies:
  - name: mysql
    version: ~1.6.0
    repository: "@stable"
    condition: mysql.enabled,global.mysql.enabled
```

and run `garden dev`

### Workaround

Could not find a way to workaround it

### Suggested solution(s)

`renderTemplates` should run `helm dependency update` when there is a requirements.yaml or a Chart with dependencies 

https://github.com/garden-io/garden/commit/26a87b9f5b07d1bbaccc39076d8e349eeec38b48#diff-22c65fc3306e692e526134d89fd8bc17L121

### Your environment

`garden version`

0.12.0

`kubectl version`

`docker version`

janario:

As described at https://github.com/garden-io/garden/issues/1951

When a helm chart has dependencies it needs to resolve them before executing the helm install/upgrade command

Fixes #1951

 @edvald, @thsig

### DIFF
--- a/garden-service/src/plugins/kubernetes/types.ts
+++ b/garden-service/src/plugins/kubernetes/types.ts
@@ -83,13 +83,10 @@ export type KubernetesWorkload =
   | KubernetesResource<V1ReplicaSet>
   | KubernetesResource<V1StatefulSet>
 
-
-export type Chart =
-  {
-    apiVersion: string
-    dependencies: Array<ChartDependency>
-  }
-export type ChartDependency =
-  {
-    name: string
-  }
+export type Chart = {
+  apiVersion: string
+  dependencies: Array<ChartDependency>
+}
+export type ChartDependency = {
+  name: string
+}
--- a/garden-service/src/plugins/kubernetes/types.ts
+++ b/garden-service/src/plugins/kubernetes/types.ts
@@ -82,11 +82,3 @@ export type KubernetesWorkload =
   | KubernetesResource<V1Deployment>
   | KubernetesResource<V1ReplicaSet>
   | KubernetesResource<V1StatefulSet>
-
-export type Chart = {
-  apiVersion: string
-  dependencies: Array<ChartDependency>
-}
-export type ChartDependency = {
-  name: string
-}
--- a/garden-service/src/plugins/kubernetes/types.ts
+++ b/garden-service/src/plugins/kubernetes/types.ts
@@ -82,3 +82,14 @@ export type KubernetesWorkload =
   | KubernetesResource<V1Deployment>
   | KubernetesResource<V1ReplicaSet>
   | KubernetesResource<V1StatefulSet>
+
+
+export type Chart =
+  {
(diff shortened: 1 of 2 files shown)


janario:
@edvald, @thsig any updates here?

we've had to rollback using the 0.11.14 because of this issue

garden-service/src/plugins/kubernetes/types.ts
```diff
@@ -82,3 +82,11 @@ export type KubernetesWorkload =
   | KubernetesResource<V1Deployment>
   | KubernetesResource<V1ReplicaSet>
   | KubernetesResource<V1StatefulSet>
+
+export type Chart = {
```
eysi09, line 86: I think this would make more sense as an interface. We could also inline `ChartDependency` since I don't think it'l be used elsewhere. Also, isn't `dependencies` optional?

In that case we'd get:

```
export interface Chart {
  apiVersion: string
  dependencies?: { name: string }[]
}
```

eysi09: I'd also suggest just moving this to the file where it's used, `helm/common.ts`, since this file is mostly used for types that are extensions of the types provided by the K8s npm package.

janario: done,

thanks for the suggestions :+1:

janario:
Thanks @eysi09 looking forward to using it soon ;-)

PR #1958, opened 2020-07-24
fix(local-k8s): add 'docker-desktop' as supported context

eysi09:

Adds `docker-desktop` as supported context for the `local-kubernetes` provider. Bumped into this after updating Docker for Mac.

Fixes #


eysi09:
Assigning @edvald and @thsig but just need a review from either of you. So whoever gets here first.

PR #1970, opened 2020-07-30
feat(cli): add --var flag for setting individual variable values

edvald:

From the added docs:

For tasks that are often run ad-hoc from the CLI, you can use variables and the `--var` CLI flag to pass in values to the task.
Here for example, we have a simple container task that can receive an argument via a variable:

```yaml
kind: Module
type: container
...
tasks:
  - name: my-task
    command: ["echo", "${var.my-task-arg || ''}"]
    ...
```

You can run this task and override the argument variable like this:

```sh
garden run task my-task --var my-task-arg="hello"!
```

### DIFF
--- a/docs/using-garden/tasks.md
+++ b/docs/using-garden/tasks.md
@@ -129,6 +129,27 @@ tasks:
 
 Tasks are also implicitly disabled when the parent module is disabled.
 
+### Running tasks with arguments from the CLI
+
+For tasks that are often run ad-hoc from the CLI, you can use variables and the `--var` CLI flag to pass in values to the task.
+Here for example, we have a simple container task that can receive an argument via a variable:
+
+```yaml
+kind: Module
+type: container
+...
+tasks:
+  - name: my-task
+    command: ["echo", "${var.my-task-arg || ''}"]
+    ...
+```
+
+You can run this task and override the argument variable like this:
+
+```sh
+garden run task my-task --var my-task-arg="hello!"
+```
+
 ### Kubernetes Provider
 
 The Kubernetes providers execute each task in its own Pod inside the project namespace. The Pod is removed once the task has finished running.
--- a/docs/using-garden/variables-and-templating.md
+++ b/docs/using-garden/variables-and-templating.md
@@ -214,14 +214,28 @@ services:
 
 ### Variable files (varfiles)
 
-You can also provide variables using "variable files" or _varfiles_. These work mostly like "dotenv" files or envfiles. However, they don't implicitly affect the environment of the Garden process and the configured services, but rather are added on top of the `variables` you define in your project `garden.yml`.
+You can also provide variables using "variable files" or _varfiles_. These work mostly like "dotenv" files or envfiles. However, they don't implicitly affect the environment of the Garden process and the configured services, but rather are added on top of the `variables` you define in your project configuration.
 
 This can be very useful when you need to provide secrets and other contextual values to your stack. You could add your varfiles to your `.gitignore` file to keep them out of your repository, or use e.g. [git-crypt](https://github.com/AGWA/git-crypt), [BlackBox](https://github.com/StackExchange/blackbox) or [git-secret](https://git-secret.io/) to securely store the files in your Git repo.
 
 By default, Garden will look for a `garden.env` file in your project root for project-wide variables, and a `garden.<env-name>.env` file for environment-specific variables. You can override the filename for each as well. The format of the files is the one supported by [dotenv](https://github.com/motdotla/dotenv#rules).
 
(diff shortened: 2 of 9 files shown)


docs/using-garden/tasks.md
```diff
@@ -129,6 +129,27 @@ tasks:
...
+tasks:
+  - name: my-task
+    command: ["echo", "${var.my-task-arg || ''}"]
+    ...
+```
+
+You can run this task and override the argument variable like this:
+
+```sh
+garden run task my-task --var my-task-arg="hello"!
```
edvald, line 150: Misplaced the exclamation point there

garden-service/src/cli/cli.ts
```diff
@@ -158,6 +160,10 @@ export const GLOBAL_OPTIONS = {
     help: "Force refresh of any caches, e.g. cached provider statuses.",
     defaultValue: false,
   }),
+  "var": new StringsParameter({
+    help:
+      'Set a specific variable value, using the format <key>=<value>, e.g. `--var some-key=custom-value`. This will override any value set in your project configuration. You can specify multiple variables by separating with a comma, e.g. `--var key-a=foo,key-b="value,with,commas"`.',
```
edvald, line 165: Turns out, due to a sywac issue, the commas within quotes don't work. I think that's okayish for the PR, but a nudge to accelerate getting rid of sywac.

docs/using-garden/variables-and-templating.md
```diff
@@ -214,14 +214,26 @@ services:
...
 
 This can be very useful when you need to provide secrets and other contextual values to your stack. You could add your varfiles to your `.gitignore` file to keep them out of your repository, or use e.g. [git-crypt](https://github.com/AGWA/git-crypt), [BlackBox](https://github.com/StackExchange/blackbox) or [git-secret](https://git-secret.io/) to securely store the files in your Git repo.
 
 By default, Garden will look for a `garden.env` file in your project root for project-wide variables, and a `garden.<env-name>.env` file for environment-specific variables. You can override the filename for each as well. The format of the files is the one supported by [dotenv](https://github.com/motdotla/dotenv#rules).
 
+You can also set variables on the command line, with `--var` flags. Note that while this is handy for ad-hoc invocations, we don't generally recommend relying on this for normal operations, since you lose a bit visibility within your configuration. But here's one practical example:
```
eysi09, line 223: lose a bit visibility > lose a bit of visibility

docs/using-garden/variables-and-templating.md
```diff
@@ -214,14 +214,26 @@ services:
...
+# Override one specific variable value and run a task
+garden run task my-task --var my-task-arg=foo
+```
+
 The order of precedence across the varfiles and project config fields is as follows (from highest to lowest):
-_`garden.<env-name>.env` file_ > _`environment[].variables` field_ > _`garden.env` file_ > _`variables` field
+
+1. Individual variables set with `--var` flags.
+2. The environment-specific varfile (defaults to `garden.<env-name>.env`).
+3. The environment-specific variables set in _`environment[].variables`.
```
eysi09, line 234: Remove `_`

garden-service/src/cli/cli.ts
```diff
@@ -6,8 +6,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import sywac from "sywac"
+import dotenv = require("dotenv")
 import chalk from "chalk"
+import sywac from "sywac"
```
eysi09, line 11: How was this missing?

garden-service/src/commands/base.ts
```diff
@@ -116,12 +116,12 @@ export interface StringsConstructor extends ParameterConstructor<string[]> {
...
   type = "array:string"
   schema = joi.array().items(joi.string())
-  delimiter: string
+  delimiter: string | RegExp
 
   constructor(args: StringsConstructor) {
     super(args)
 
-    this.delimiter = args.delimiter || ","
+    this.delimiter = args.delimiter || /,(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)/
```
eysi09, line 124: Would you mind adding a comment explaining what the regex is for. (It's for handling quotes, right?) Also, do we need `/g` here?

edvald: Sure, I'll add a comment. It's for dealing with commas within quotes.

garden-service/src/cli/cli.ts
edvald: Probably some mixup when I started splitting out the sywac-b-gone branch

edvald:
Updated, should be good to go now

PR #1973, opened 2020-08-03
Fix hot reloading issues

edvald:

See commits for details. We had a few path handling issues after change made in v0.12.0, in particular on Windows and when hot-reloading `helm` modules.

Fixes #

Would be good to do a sanity check on this. I've tested on both Windows and mac, for container and helm+container syncs, but ya never know.


edvald:
Gotten a couple of 👍 from our users on Slack, so this seems to do the trick.

PR #1975, opened 2020-08-03
refactor(cli): get rid of sywac dependency and improve CLI-code tests

edvald:
**What this PR does / why we need it**:

Some of our oldest tech-debt there... It'll be good to have this over
with. This will unblock a few small improvements but I figure it's best
to do just the refactor first.

**Special notes for your reviewer**:

Probably good to hold off on merging until after our next release. Best let this simmer and test it well, see if we find any unexpected issues.

PR #1978, opened 2020-08-04
Dashboard command

edvald:

The new `garden dashboard` command is designed to run standalone, and
receive events from other Garden processes. Any other process running
with the same project, environment and namespace will pipe events to
this process when it's running.

All active Garden processes are now cataloged in a table in our sqlite
database, and this table is polled for active processes during all
CLI execution. This will no doubt be handy for other stuff later.

The dashboard server adds an authenticated `/events` endpoint, which
uses the same API as the same endpoint in GE, and much of the same
event streaming code is used.

This should make the dashboard much more usable in day-to-day dev work,
and I aim to continue work in that direction.

There are some further improvements needed to make the dashboard better.
One of which is to fix reconnection issues, and we also need to make it
more generally useful by leaning on its pluggability, making more handy
providers which embed pages in the dashboard that will be useful for
development.

Notably, this improvement is also an important step in embedding
dashboards in the GE UI. This should be fairly easy now, since the
dashboard can receive events in the exact same shape as the GE event
store holds.

Once we feel this is stable, and we have a few more steps covered, I
will make a separate PR with a detailed usage guide.

The PR is based on #1975, let's wrap that up before merging. I'll leave this as a draft until then but you can go ahead and review + test if you like.


core/src/commands/dashboard.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  arguments = dashboardArgs
+  options = dashboardOpts
+
+  getLoggerType(): LoggerType {
+    return "basic"
+  }
+
+  async prepare({ footerLog, opts }: PrepareParams<Args, Opts>) {
+    this.server = await startServer({ log: footerLog, port: opts.port })
```
thsig, line 54: If a `dashboard` command is already running in project A when I run a `dashboard` command in project B, I get an "address already in use" error for the second invocation.

This can be worked around by providing a custom port (via the `--port` option).

We should probably display a more helpful error message that recommends picking another port to the user.

edvald: Updated, see two new TBS commits.

thsig:
This one seems good to go now. We just need to fix the linting errors and squash the TBS commits before merging.

PR #1984, opened 2020-08-07
feat(config): add $merge key for merging maps together in configs

edvald:
Any object or mapping field supports a special `$merge` key, which
allows you to merge two objects together. This can be used to avoid
repeating a set of commonly repeated values.

Example:

```yaml
kind: Project
...
variables:
  - commonEnvVars:
      LOG_LEVEL: info
      SOME_API_KEY: abcdefg
      EXTERNAL_API_URL: http://api.example.com
  ...
```

```yaml
kind: Module
type: container
name: service-a
...
services:
  env:
    $merge: ${var.commonEnvVars}
    OTHER_ENV_VAR: something
    # <- This overrides the value set in commonEnvVars, because it is
    #    below the $merge key
    LOG_LEVEL: debug
  ...
```

```yaml
kind: Module
type: container
name: service-b
...
services:
  env:
    # <- Because this is above the $merge key, the API_KEY from
    #    commonEnvVars will override this
    SOME_API_KEY: default
    $merge: ${var.commonEnvVars}
  ...
```

### DIFF
--- a/docs/using-garden/variables-and-templating.md
+++ b/docs/using-garden/variables-and-templating.md
@@ -141,6 +141,50 @@ When the nested expression is a simple key lookup like above, you can also just 
 
 You can even use one variable to index another variable, e.g. `${var.a[var.b]}`.
 
+### Merging maps
+
+Any object or mapping field supports a special `$merge` key, which allows you to merge two objects together. This can be used to avoid repeating a set of commonly repeated values.
+
+Here's an example where we share a common set of environment variables for two services:
+
+```yaml
+kind: Project
+...
+variables:
+  - commonEnvVars:
+      LOG_LEVEL: info
+      SOME_API_KEY: abcdefg
+      EXTERNAL_API_URL: http://api.example.com
+  ...
+```
+
+```yaml
+kind: Module
+type: container
+name: service-a
+...
+services:
+  env:
+    $merge: ${var.commonEnvVars}
+    OTHER_ENV_VAR: something
+    LOG_LEVEL: debug  # <- This overrides the value set in commonEnvVars, because it is below the $merge key
+  ...
+```
+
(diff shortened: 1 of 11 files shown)


docs/using-garden/variables-and-templating.md
```diff
@@ -141,6 +141,50 @@ When the nested expression is a simple key lookup like above, you can also just
 
 You can even use one variable to index another variable, e.g. `${var.a[var.b]}`.
 
+### Merging maps
+
+Any object or mapping field supports a special `$merge` key, which allows you to merge two objects together. This can be used to avoid repeating a set of commonly repeated values.
```
thsig, line 146: We should probably refer to objects as dictionaries in the context of YAML syntax (to stick with YAML's terminology).

garden-service/src/config/common.ts
```diff
@@ -278,22 +281,29 @@ joi = joi.extend({
 })
 
 /**
- * Add a joi.customObject() type, which includes additional methods, including one for validating with a
+ * Extend the joi.object() type with additional methods and minor customization, including one for validating with a
```
thsig, line 284: "customization" -> "customizations"

docs/using-garden/variables-and-templating.md
edvald: YAML seems to mostly use "mapping" in the spec: https://yaml.org/spec/1.2/spec.html#mapping//
Although I agree that "object" can be swapped out here. Feels quite subjective though, so I don't feel strongly about it.

thsig: Yeah, agreed. Let's just leave it as it is.

edvald:
Should be good to go now @thsig

PR #1990, opened 2020-08-11
fix(enterprise): fix login flow

thsig:

Fixes a recent regression in the login flow, and stops throwing enterprise-related errors for non-enterprise projects when the user is logged in.

Also added unit tests for the `login` command.

### DIFF
--- a/garden-service/src/garden.ts
+++ b/garden-service/src/garden.ts
@@ -116,17 +116,12 @@ export interface GardenOpts {
   variables?: PrimitiveMap
 }
 
-export interface GardenEnterpriseContext {
-  clientAuthToken: string
-  projectId: string
-  enterpriseDomain: string
-}
-
 export interface GardenParams {
   artifactsPath: string
   buildDir: BuildDir
+  clientAuthToken: string | null
+  enterpriseDomain: string | null
   projectId: string | null
-  enterpriseContext: GardenEnterpriseContext | null
   dotIgnoreFiles: string[]
   environmentName: string
   environmentConfigs: EnvironmentConfig[]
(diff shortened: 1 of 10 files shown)


garden-service/src/garden.ts
```diff
@@ -329,25 +326,26 @@ export class Garden {
...
+    const enterpriseDomain = config.domain || null
     if (!opts.noEnterprise) {
       const enterpriseInitResult = await enterpriseInit({ log, projectConfig: config, environmentName })
       secrets = enterpriseInitResult.secrets
-      const clientAuthToken = enterpriseInitResult.clientAuthToken
-      if (clientAuthToken && projectId && enterpriseDomain) {
-        enterpriseContext = { clientAuthToken, projectId, enterpriseDomain }
-      }
+      clientAuthToken = enterpriseInitResult.clientAuthToken
+      // enterpriseContext = { clientAuthToken, projectId, enterpriseDomain: enterpriseDomain || null }
```
eysi09, line 337: Remove this?

garden-service/test/unit/src/commands/login.ts
```diff
@@ -0,0 +1,84 @@
...
+    footerLog: log,
+    args: {},
+    opts: withDefaultGlobalOpts({}),
+  }
+}
+
+describe("LoginCommand", () => {
+  let tmpDir: tmp.DirectoryResult
+  let projectConfig: ProjectConfig
+  // let sandbox = sinon.sandbox.create()
```
eysi09, line 35: Remove this.

garden-service/package.json
```diff
@@ -215,6 +215,7 @@
     "prettier": "^1.19.1",
     "replace-in-file": "^6.0.0",
     "shx": "^0.3.2",
+    "sinon": "^9.0.3",
```
eysi09, line 218: Do we need to install `sinon` for this? Why not just use `testdouble`?

thsig: Good point—hadn't thought of that. These test cases now use `testdouble` (and I've gotten rid of `sinon`).

PR #1991, opened 2020-08-11
Dashboard fix ws

eysi09:

Attempt to reconnect to Garden core API server from dashboard when websocket connection is lost. 

Fixes #


edvald:
Works nicely! Only comment is that that the indicator is a little easy to miss because it's black text on a white background. Perhaps white on red would be better, seeing as it is an error state?

eysi09:
Yeah good point. So white font on red background?

eysi09:
Realised that we could also easily pause the connection retries if the windows isn't focused, I'll add that as well.

PR #2001, opened 2020-08-13
feat(terraform): add allowDestroy flags to automatically destroy stacks

### LINKED ISSUE #1928: [FEATURE]: Delete service handler for Terraform modules
### Background / Motivation

Currently Garden can provision Terraform resources, but it cannot de-provision them. We currently use the Terraform provider to create per-environment cloud resources, which then get deleted when an environment is deleted.

Destroy resources created by a Terraform module using the `garden delete` command.

Currently when deleting an environment with Garden, an error will be printed that Terraform resources will not be deleted:
```
      ✖    tf                     → No delete service handler available for module type terraform
      ✔    tf                     → No delete service handler available for module type terraform → Done (took 4.9 sec)
```

We have a workaround (an example for the service named `tf`), and it involves several steps:
1.  Overwriting the Terraform configuration files with "empty" files. These files are valid Terraform files that define no resources.
2.  Deploy `tf` again. This causes Garden to destroy all managed resources.
3.  Proceeding to do a `garden delete environment`

Implement a delete service handler that invokes `terraform destroy`.

edvald:
You can now set `allowDestroy: true` in both the provider config and
`terraform` module configs, which if set allow Garden to automatically
destroy stacks when calling `garden delete env` and/or
`garden delete service <module name>`.

Also added are new `garden plugins terraform destroy-root` and
`garden plugins terraform destroy-module` commands.

Closes #1928

cc @noizwaves

### DIFF
--- a/core/src/plugins/terraform/commands.ts
+++ b/core/src/plugins/terraform/commands.ts
@@ -19,7 +19,7 @@ import { join } from "path"
 import { remove } from "fs-extra"
 import { getProviderStatusCachePath } from "../../tasks/resolve-provider"
 
-const commandsToWrap = ["apply", "plan"]
+const commandsToWrap = ["apply", "plan", "destroy"]
 const initCommand = chalk.bold("terraform init")
 
 export const terraformCommands: PluginCommand[] = commandsToWrap.flatMap((commandName) => [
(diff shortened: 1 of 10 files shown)


noizwaves:
Woohoo, super excited for this, thanks!!

thsig (approved):
Looks good to me 👍

PR #2006, opened 2020-08-14
feat(dashboard): add octant provider and dashboard integration

edvald:

This adds an `octant` provider, that you can drop into any Garden
project that includes a `kubernetes`/`local-kubernetes` provider:

```
...
providers
  - name: octant
...
```

This will add an Octant tab to your dashboard, which will be loaded
on-demand, pointed at the context and namespace of your project.
You also get a `garden tools octant` command if you'd like to run it
standalone.

Most of the work in the PR is to do with the mechanism for providers to
add dashboard pages, which is now easy to do and can be done in a
dynamic fashion.

I also added a toggle for provider tools, to indicate whether they
should be pre-fetched as part of our container builds.

Please test the dashboard itself, see if it all works as expected.


core/src/commands/util/fetch-tools.ts
```diff
@@ -6,30 +6,28 @@
...
+import { printHeader, printFooter } from "../../logger/util"
+import { BooleanParameter } from "../../cli/params"
 
 const fetchToolsOpts = {
-  all: new BooleanParameter({
+  "all": new BooleanParameter({
     help: "Fetch all tools for registered plugins, instead of just ones in the current env/project.",
     required: false,
   }),
+  "prefetch-only": new BooleanParameter({
```
eysi09, line 26: Wouldn't `--only-prefetch` be more idiomatic?

I actually think either option name is a little confusing in this context (although it's mainly used internally). I would expect `garden util fetch-tools --all` to basically prefetch tools, so it's not clear what `--prefetch-only | --only-prefetch` adds to it.

Can we think of a better flag than `prefetch=true` for these tools? Perhaps `builtIn`? Or something more explicit like `includeInImage` / `addToImage` (or some version of that).

core/src/plugins/octant/octant.ts
```diff
@@ -0,0 +1,122 @@
...
+          // Wait for dashboard to be available
+          octantProc.stderr?.on("data", (data) => {
+            if (data.toString().includes("Dashboard is available")) {
+              resolved = true
+              resolve({ url: "http://" + host })
+            }
+          })
+
+          octantProc.on("error", (err) => {
+            !resolved && reject(err)
```
eysi09, line 69: Why do we need to check for `!resolved` here?

core/src/config-store.ts
```diff
@@ -185,20 +185,21 @@ const analyticsLocalConfigSchema = () =>
     .meta({ internal: true })
 
 const localConfigSchemaKeys = {
-  linkedModuleSources: joiArray(linkedSourceSchema()),
-  linkedProjectSources: joiArray(linkedSourceSchema()),
-  analytics: analyticsLocalConfigSchema(),
+  linkedModuleSources: () => joiArray(linkedSourceSchema()),
```
eysi09, line 188: Why are we making the value a function?

core/test/unit/src/server/server.ts
```diff
@@ -67,7 +67,10 @@ describe("GardenServer", () => {
   })
 
   describe("GET /", () => {
-    // TODO: test dashboard endpoint
+    // await request(server)
```
eysi09, line 70: Should this be commented out?

dashboard/src/api/actions.ts
```diff
@@ -198,7 +212,7 @@ function processStatusInitResult(entities: Entities, status: StatusCommandResult
       draft.tests[testName] = entities.tests[testName] || {}
       draft.tests[testName].status = testStatus
     }
-    draft.providers = status.providers
+    draft.environmentStatuses = status.providers
```
eysi09, line 215: Why this change?

dashboard/src/components/provider.tsx
```diff
@@ -6,24 +6,55 @@
...
-    </div>
-  )
+const ProviderPageWrapper = styled.div`
+  flex: 0 auto;
+  border: 0;
+  width: 100%;
+  height: 100%;
+`
+
+class ProviderPageFrame extends React.Component<ProviderPageProps, ProviderPageState> {
```
eysi09, line 30: Any reason you didn't go for a functional component here? It would probably be a bit simpler and more consistent with the rest of the components.

dashboard/src/components/provider.tsx
```diff
@@ -6,24 +6,55 @@
...
+  constructor(props: ProviderPageProps) {
+    super(props)
+    this.state = {
```
eysi09, line 33: For some reason the spinner isn't visible when fetching Octant, during the very first load which is also the longest one.

eysi09 (changes requested):
Had a few question / comments, otherwise looks good.

core/src/plugins/octant/octant.ts
edvald: It's possible that the proc emits an error after the promise is resolved, in which case calling reject is basically an error.

core/src/commands/util/fetch-tools.ts
edvald: Yeah I don't feel strongly about this. It's _only_ used internally and is hidden from any docs. The flag will also not be a part of any external APIs. That said, it's fine to rename them for clarity, make it a little more verbose and obvious.

core/src/config-store.ts
edvald: To avoid circular import issues, which tend to come up if the values are resolved at module resolution time. Most schemas are now defined like this, and otherwise I tend to fix it if I come across it.

core/test/unit/src/server/server.ts
edvald: Nope, will fix.

dashboard/src/components/provider.tsx
edvald: Yes. I could not find a way to make functional components persist, i.e. not unmount, on route change. The slight added complexity means that the iframe is hidden/shown on route changes after it's first created.

dashboard/src/components/provider.tsx
edvald: Hmm. But only in that specific first case?

core/src/commands/util/fetch-tools.ts
edvald: Changed in added TBS commit

dashboard/src/components/provider.tsx
eysi09: I'd guess it's because we're not mounting this component until the API call returns. And it's only really noticeable if Garden has to actually fetch the tool as part of the API call.

dashboard/src/components/provider.tsx
eysi09: Hmm ok.

edvald: Maybe I'm out of my depth here in React land, open to a different solution if you have one.

dashboard/src/api/actions.ts
eysi09: Ok makes sense. So it was basically broken before since we were overriding the providers with environment statuses. But doesn't appear to be used anywhere.

dashboard/src/components/provider.tsx
```diff
@@ -6,24 +6,55 @@
...
+      loading: true,
+    }
+  }
+  // componentDidUpdate(prevProps, prevState) {
```
eysi09, line 37: Also forgot mention this in my first pass. Do we need that?

dashboard/src/components/provider.tsx
eysi09: This is fine IMO. But perhaps add a comment explaining why, so some wise-guy/gal won't come along and refactor this to a functional component.

dashboard/src/components/provider.tsx
edvald: Gotcha. I can take another look, and test it with a sleep to make sure it works as expected.

dashboard/src/components/provider.tsx
edvald: Nope, just needs a cleanup.

dashboard/src/components/provider.tsx
edvald: > I'd guess it's because we're not mounting this component until the API call returns.

That's not correct though. We mount the component immediately, the iframe just gets the URL that eventually redirects (no separate API call performed), in this particular case after the octant tool is fetched. 

And it looks to work fine for me, I get the pink spinner immediately 🤷‍♂️. Can you give another... _spin_?

dashboard/src/api/actions.ts
edvald: I think it worked as it was, but we have since moved the `dashboardPages` key from the environment status to the provider definition. And yeah I suppose it's no longer used then, but I'm quite sure we will have some use for this later, so I'll just leave it.

dashboard/src/components/provider.tsx
edvald: Turns out I could use a function component. Indeed, much more elegant. The problem I previously had turned out to be solved in the router setup. See the updated TBS commit.

dashboard/src/components/provider.tsx
eysi09: Hehe. But yeah ok, this only happens if you navigate to the Octant tab from another tab. So if I go straight to the Octant page via `/providers/octant` I see this spinner. But if I start at the Overview page and navigate my way towards the Octant page, I don't see the spinner.

edvald: For whatever reason, I don't get the same result :/ Works in either case for me.

edvald: Updated with your suggested fix. Can you give it a ~spin~ another go to confirm that it works?

eysi09: Works now. But boy, did we have to go round and round on this one.

eysi09 (approved):

PR #2010, opened 2020-08-19
feat(config): allow explicitly declaring provider dependencies

edvald:
This adds a `providers[].dependencies` field, which can be used to
explicitly declare dependencies at config time. Previously you could
only create implicit dependencies by referencing another provider's
outputs in template strings.

This should be handy when e.g. using the `exec` provider with an
`initScript`, which should run before resolving other providers.


eysi09 (approved):
LGTM

PR #2017, opened 2020-08-25
chore(docker): upgrade node version in alpine and buster

solomonope:
This PR upgrades the version of Nodejs in use in the alpine and buster docker images.

Node version 12.18.3, is the minimum node version that supports CLS, Garden Enterprise runs an image that is based on Gardendev images.

Fixes #


thsig:
Looks good to me, but the `build-docker-aws-gcloud`, `build-docker-azure` and `build-docker-aws` steps are failing here.

edvald:
I suspect the switch between alpine versions is the cause for the problem. Is there any particular reason to do that?

edvald:
Another potential issue is that it's the Node.js version that _pkg_ uses that needs to be updated. I'm not quite sure if this particular version is available there.

solomonope:
> Looks good to me, but the `build-docker-aws-gcloud`, `build-docker-azure` and `build-docker-aws` steps are failing here.

I would check it

solomonope:
> I suspect the switch between alpine versions is the cause for the problem. Is there any particular reason to do that?

There is no specific reason for upgrading the alpine version, I would look for a lower alpine version that works

thsig (approved):

PR #2018, opened 2020-08-25
feat(enterprise): enable secrets in more contexts

thsig:

We now make secrets available in environment configs, and in top-level project configs and variables.

Previously, secrets were only available in provider, module and workflow configs.


thsig:
There are a few failing tests there, will check those out now.

thsig:
Ok, this is ready for review now.

edvald (approved):

PR #2019, opened 2020-08-26
refactor: re-organize repo, split some plugins out of core

edvald:

This continues some work I had already started, basically breaking up the `core` package, so that we can untangle plugins from that package and create new ones in separate packages (even if it's within the same repo for now).

You can see the individual commit messages for more on what I actually did, but here's a quick rundown:

* **Switched from npm to yarn across the board**. This solves a few headaches we had with package resolution, and is faster as well.
* Added a root level `yarn run dev` command, which watches all the different packages in the repo.
* Moved the `static` directory to the top level.
* Created a new `cli` package, which we build the final binary from. This depends on the `core` package, where most of the code remains, but the benefit is that the `cli` package can then take care of bundling plugins that are outside of `core`.
* Created a `@garden-io/sdk` package, which exports some functions and types that plugins need. The eventual goal being that plugins never need to directly reference the `@garden-io/core` package.
* Moved the `conftest*` plugins into the `plugins` directory, outside of the core package, along with their tests.
* Added `yarn run test:plugins` script to quickly test the plugins, and `yarn run test:framework:unit` which unit tests everything else basically. Note that the plugin tests are generally _integration_ tests, and are run in CI right after the core integ test suite.
* Rewrote the `scripts/build-pkg.sh` script in TypeScript, made it both faster and more robust (and of course easier to develop than friggin bash). The build flow is also different now, and avoids some interop issues between `pkg` and our internal package linking.
* Added a section to CONTRIBUTING.md describing the top-level directory structure.
* And lastly some misc small tweaks and improvements, stuff I came across along the way.


thsig (approved):
Looks good to me 👍

thsig:
Maybe we'll need to update our CircleCI config to accommodate these changes?

dashboard/src/components/graph/index.tsx
```diff
@@ -100,10 +100,9 @@ function drawChart(graph: Graph, width: number, height: number, onGraphNodeSelec
     // Round the corners of the nodes
     node.rx = node.ry = 4
     // Remove node padding
-    node.paddingBottom = 0
-    node.paddingTop = 0
-    node.paddingLeft = 0
-    node.paddingRight = 0
+    node.padding = 0
```
thsig, line 103: I'm not familiar with the rendering context here, but would it be enough to just set `node.padding = 0` here?

thsig (approved):
Looks good to me 👍

dashboard/src/components/graph/index.tsx
edvald: I don't know either, so I just went with all of the above 🤷‍♂️

PR #2021, opened 2020-08-27
chore: fixes to release process

thsig:

In a recent refactor, some of the release-related scripts were changed and moved around, causing problems in the release pipeline.

Please comment if any of the changes feel hacky or incomplete with respect to the larger setup.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -164,7 +164,7 @@ commands:
       - *attach-workspace
       - run:
           name: Run dist command with the appropriate argument
-          command: npm run dist -- -- garden_version=<<parameters.version>>
+          command: npm run dist -- -- <<parameters.version>>
       - persist_to_workspace:
           root: ./
           paths:
--- a/scripts/release.ts
+++ b/scripts/release.ts
@@ -47,7 +47,7 @@ async function release() {
   }
   // Update package.json versions
   await execa("node_modules/.bin/lerna", [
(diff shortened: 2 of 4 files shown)


scripts/release.ts
```diff
@@ -47,7 +47,7 @@ async function release() {
   }
   // Update package.json versions
   await execa("node_modules/.bin/lerna", [
-    "version", "--no-git-tag-version", "--yes", releaseType,
+    "version", "--no-git-tag-version", "--force-publish", "--yes", releaseType,
```
thsig, line 50: Lerna wasn't updating the versions in the `package.json`s without the `--force-publish` flag here.

.circleci/config.yml
```diff
@@ -164,7 +164,7 @@ commands:
       - *attach-workspace
       - run:
           name: Run dist command with the appropriate argument
-          command: npm run dist -- -- garden_version=<<parameters.version>>
+          command: npm run dist -- -- <<parameters.version>>
```
thsig, line 167: Before this change, the version was coming through as `garden_version.` instead of e.g. `v0.12.3-alpha.6` in the asset artifact names on the prerelease.

It looks like it may have to do with `parameters.version` being blank, which resulted in a non-blank value missing the version suffix being passed to the subscripts, hence overriding the default calculation, which I think reads the version directly from `core/package.json` (which works).

edvald (approved):

PR #2026, opened 2020-09-01
improvement: add plugin alias for plugins command

edvald:
I kept mistyping it and would see users do the same, so why not :P


edvald:
The test failure appears to be unrelated to the PR, will look into that.

Issue #2003, opened 2020-08-13
TypeError: Cannot read property 'split' of undefined


ig-5g:
### Current Behavior

`garden delete env` or `garden delete service <service-name>` result in:

```
[2020-08-12T17:49:38.810Z] TypeError: Cannot read property 'split' of undefined
    at makeErrorMsg (/snapshot/project/garden-service/tmp/dist/build/src/util/util.js:0)
    at Object.<anonymous> (/snapshot/project/garden-service/tmp/dist/build/src/util/util.js:0)
    at Generator.throw (<anonymous>)
    at rejected (/snapshot/project/garden-service/tmp/dist/build/src/util/util.js:0)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

This is a regression in garden `0.12.x`. We don't see this error in `0.11.x` 

The error is not platform-specific. We get the same error on MacOS, CentOS 7 and Windows 10.

### Expected behavior

`garden delete env` and `garden delete service` work cleanly in `0.12.x`

### Reproducible example

We are still working on creating a minimal reproducible `garden.yaml`.  We need help with the error message -- see below.

### Workaround

We could not find one. We pinned our garden version to `0.11.14`

### Suggested solution(s)

There are two issues here. The `split` error comes from this line in `util.ts`

```
export function makeErrorMsg({
   ...
   const lines = output.split("\n")
   ...
}
```

The code dereferences undefined `output`. This bug masks the original error.

It'd be nice to fix the undefined `output` issue in the upcoming `0.12.x` release. Once it's fixed, we hope to see more clues about the root cause of the problem.

### Your environment

```
$ garden version
0.12.2
```

Note: we started getting the error in `0.12.0`. The regression happened between `0.11.14` and `0.12.0`

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-15T16:58:53Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #2027, opened 2020-09-01
improvement(cli): catch OOM errors and exit with helpful error message

edvald:
This is one of a series of improvements I'm working on, to warn users
about very large projects that probably need better configuration of
includes/excludes/ignores.

This catches the "nuclear" case of running out of heap space, exiting
with a standard code of 137, and printing an error with a link to docs.

See #2003 for a reported case.


eysi09 (approved):
LGTM!

PR #2028, opened 2020-09-01
improvement: explicitly catch EMFILE errors with better error message

edvald:
Another one in the series of potential errors when the project file
count is too high.

See issue #2003 for one reported case of this.

### DIFF
--- a/core/src/util/util.ts
+++ b/core/src/util/util.ts
@@ -142,7 +142,7 @@ export function makeErrorMsg({
 
     ${trimEnd(error, "\n")}
   `
-  if (output !== error) {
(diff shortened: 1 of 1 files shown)


core/src/util/util.ts
```diff
@@ -142,7 +142,7 @@ export function makeErrorMsg({
 
     ${trimEnd(error, "\n")}
   `
-  if (output !== error) {
+  if (out && output !== error) {
```
eysi09, line 145: Shouldn't this be if `output && ...` since `out` is just `output` that's been formatted.

eysi09 (approved):

PR #2029, opened 2020-09-01
improvement(core): warn on large file count in modules

edvald:
Yet another enhancement geared towards alerting users about potential
misconfiguration.

This includes a new command and mechanism to suppress individual
warnings. We can use that to emit more warnings of this nature, where
the warning may or may not be relevant and could be annoying to see on
every invocation.

### DIFF
--- a/core/src/commands/util/hide-warning.ts
+++ b/core/src/commands/util/hide-warning.ts
@@ -1,0 +1,41 @@
+/*
+ * Copyright (C) 2018-2020 Garden Technologies, Inc. <info@garden.io>
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { Command, CommandParams } from "../base"
+import dedent from "dedent"
+import { StringParameter } from "../../cli/params"
+import { Warning } from "../../db/entities/warning"
+
+const hideWarningArgs = {
+  key: new StringParameter({
+    help: "The key of the warning to hide (this will be shown along with relevant warning messages).",
+    required: true,
+  }),
+}
+
+type Args = typeof hideWarningArgs
+
+export class HideWarningCommand extends Command<Args, {}> {
+  name = "hide-warning"
+  help = "Hide a specific warning message."
+  cliOnly = true
+
+  noProject = true
+
+  description = dedent`
+    Hides the specified warning message. The command and key is generally provided along with displayed warning messages.
+  `
+
+  arguments = hideWarningArgs
+
+  async action({ args }: CommandParams<Args, {}>) {
+    await Warning.hide(args.key)
+
+    return {}
+  }
+}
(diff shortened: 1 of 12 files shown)


core/src/commands/util/hide-warning.ts
```diff
@@ -0,0 +1,41 @@
...
+const hideWarningArgs = {
+  key: new StringParameter({
+    help: "The key of the warning to hide (this will be shown along with relevant warning messages).",
+    required: true,
+  }),
+}
+
+type Args = typeof hideWarningArgs
+
+export class HideWarningCommand extends Command<Args, {}> {
```
eysi09, line 23: Should you be able to un-hide warnings?

eysi09: Also, out of scope for this PR, but feels like we should start grouping the commands in the help text by purpose, usability or such. For example basic commands for beginners vs advanced commands. The current list is a little unwieldy:

```
  build          Build your modules.
  call           Call a service ingress endpoint.
  config         Configure user and project settings.
  create         Create new project or module.
  dashboard      Starts the Garden dashboard for the current project and environment.
  delete         Delete configuration or objects.
  deploy         Deploy service(s) to your environment.
  dev            Starts the garden development console.
  exec           Executes a command (such as an interactive shell) in a running service.
  get            Retrieve and output data and objects, e.g. secrets, status info etc.
  link           Link a remote source or module to a local path.
  login          Log in to Garden Enterprise.
  logout         Log out of Garden Enterprise.
  logs           Retrieves the most recent logs for the specified service(s).
  migrate        Migrate `garden.yml` configuration files to version v0.11.x
  options        Print global options.
  plugins        Plugin-specific commands.
  publish        Build and publish module(s) to a remote registry.
  run            Run ad-hoc instances of your modules, services, tests, tasks or workflows.
  scan           Scans your project and outputs an overview of all modules.
  set            Set or modify data, e.g. secrets.
  test           Test all or specified modules.
  tools          Access tools included by providers.
  unlink         Unlink a remote source or module from its local path.
  update-remote  Pulls the latest version of remote sources or modules from their repository.
  util           Misc utility commands.
  validate       Check your garden configuration for errors.
```

core/src/db/entities/warning.ts
```diff
@@ -0,0 +1,42 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { Entity, Column, Index } from "typeorm"
+import { GardenEntity } from "../base-entity"
+import { LogEntry } from "../../logger/log-entry"
+import chalk from "chalk"
+
+/**
```
eysi09, line 14: This comment doesn't apply here.

eysi09 (approved):
LGTM! Had a few questions/comments but approving regardless.

eysi09:
Ah whoops, spoke too soon. There's that comment that needs to be removed.

core/src/commands/util/hide-warning.ts
edvald: I considered that (un-hiding warnings) but I just couldn't imagine a scenario when someone would actually do that. How would one even think to do it, or know which key to un-hide, etc.

core/src/db/entities/warning.ts
edvald: Ah yes. Copy paste fail.

eysi09 (approved):

PR #2034, opened 2020-09-03
docs: add path to artifacts dir to artifacts field description

eysi09:

The location of the artifacts dir was missing from the field description.

Fixes #

### DIFF
--- a/docs/reference/module-types/container.md
+++ b/docs/reference/module-types/container.md
@@ -346,14 +346,17 @@ tests:
     # The arguments used to run the test inside the container.
     args:
 
-    # Specify artifacts to copy out of the container after the run.
+    # Specify artifacts to copy out of the container after the run. The artifacts are stored locally under the
+    # `.garden/artifacts` directory.
+    #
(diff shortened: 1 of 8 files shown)


docs/reference/module-types/container.md
```diff
@@ -346,14 +346,19 @@ tests:
...
+    #
     # Note: Depending on the provider, this may require the container image to include `sh` `tar`, in order to enable
     # the file transfer.
     artifacts:
       - # A POSIX-style path or glob to copy. Must be an absolute path. May contain wildcards.
         source:
 
-        # A POSIX-style path to copy the artifacts to, relative to the project artifacts directory.
+        #
+        #   A POSIX-style path to copy the artifacts to, relative to the project artifacts directory at
```
edvald, line 359: There's some extra indent here that we should fix. That aside, looks good. The failing tests are something I'll fix separately.

edvald (changes requested):

docs/reference/module-types/container.md
eysi09: I've removed the extra indent.

PR #2036, opened 2020-09-04
fix(k8s): filter out undefined environment variable values

solomonope:
When we have environment variables  in nature below 
```
VAULT_APPROLE_SECRET_ID: ${var.vault-approle-secret-id}?
VAULT_APPROLE_ROLE_ID: ${var.vault-approle-role-id}?
```
They are conditional and can be `undefined`, `null` check involving `===` does not handle the case of it being `undefined`

This PR adds an explicit `undefined` check

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/util.ts
+++ b/core/src/plugins/kubernetes/util.ts
@@ -333,29 +333,31 @@ export function flattenResources(resources: KubernetesResource[]) {
  * Maps an array of env vars, as specified on a container module, to a list of Kubernetes `V1EnvVar`s.
  */
 export function prepareEnvVars(env: ContainerEnvVars): V1EnvVar[] {
-  return Object.entries(env).map(([name, value]) => {
-    if (value === null) {
-      return { name, value: "null" }
-    } else if (typeof value === "object") {
-      if (!value.secretRef.key) {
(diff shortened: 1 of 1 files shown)


core/src/plugins/kubernetes/util.ts
```diff
@@ -334,7 +334,7 @@ export function flattenResources(resources: KubernetesResource[]) {
  */
 export function prepareEnvVars(env: ContainerEnvVars): V1EnvVar[] {
   return Object.entries(env).map(([name, value]) => {
-    if (value === null) {
+    if (value === null || value === undefined) {
```
edvald, line 337: Good catch! But I'd prefer to do this differently, by basically filtering out undefined values in the object, instead of implicitly casting to null. So basically adding a `.filter(...)` before the `.map(...)` instead.

edvald (changes requested):

core/src/plugins/kubernetes/util.ts
edvald: I'd also suggest updating the commit message accordingly :)

solomonope: I have done the update

edvald (approved):

PR #2046, opened 2020-09-09
feat(cli): add `get module(s)` command

### LINKED ISSUE #2038: [FEATURE]: easy way to get version of one and/or all modules
### Background / Motivation

At the moment there seems to be no easy way to get just the module name and version for either one module or all modules.
running `garden get outputs` will give you this, but it might run a lot of builds/tasks that you might not want

Have a command `garden get module <module-name>` that returns the name and the version of the module.
Optionally it can return more information, as long as no actual builds or tasks have to be run

Usecase 1:
I'm in my project and I want to pull the current image of a build file, for that I need the tag, which contains the version. Running the build step is slow, especially when I know that the image has already been build.

We have one workaround for this: tagging every published image with the git-hash, but that can result into more tags for an image than our image registry supports

Usecase 2:
Have a quick way to get the version to get an idea whether module source has changed

edvald:
Closes #2038

### DIFF
--- a/core/src/commands/get/get-modules.ts
+++ b/core/src/commands/get/get-modules.ts
@@ -1,0 +1,89 @@
+/*
+ * Copyright (C) 2018-2020 Garden Technologies, Inc. <info@garden.io>
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { Command, CommandParams, PrepareParams } from "../base"
+import { StringsParameter, BooleanParameter } from "../../cli/params"
+import { moduleSchema, GardenModule } from "../../types/module"
+import { keyBy, omit, sortBy } from "lodash"
+import { joiIdentifierMap, joi } from "../../config/common"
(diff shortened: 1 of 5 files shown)


core/src/commands/get/get-modules.ts
```diff
@@ -0,0 +1,88 @@
...
+
+type OutputModule = Omit<GardenModule, "_config">
+
+export class GetModulesCommand extends Command {
+  name = "modules"
+  alias = "module"
+  help = dedent`
+    Outputs all or specified modules. Use with --output=json and jq to extract specific fields.
+
+    Examples:
```
thsig, line 43: The examples string should be put under the `description` field (e.g. to avoid cluttering up the subcommand listing that shows up when you run `garden get`).

thsig (changes requested):
Looks good! Only had one comment (relating to the docstrings and generated docs).

core/src/commands/get/get-modules.ts
edvald: Ah yes, will fix.

edvald: Fixed

lucas-sonnabend:
Thanks a lot for the quick turnaround!

PR #2047, opened 2020-09-09
fix(enterprise): enforce stream batch size

thsig:

Changed `BufferedEventStream`'s batching logic to fill batches with events / log entries up to a constant maximum byte count.

Compared to the previous batching logic, the new implementation is fundamentally robust with respect to request size, and generally drains its buffer much faster.

Also streamlined some event payloads to omit unnecessary and potentially very large fields.


edvald (approved):

thsig:
I'll take care of merging this one, just want to check a couple of things I just noticed.

PR #2049, opened 2020-09-11
chore: fixes to release process after yarn switch

edvald:


.circleci/config.yml
```diff
@@ -45,8 +45,7 @@ orbs:
       branches:
         ignore: /.*/
       tags:
-        # Technically could be a better regex but for now just starting with v feels good enough
-        only: /v.*/
+        only: /[0-9].*/
```
edvald, line 48: Does this make sense? This translates to any string starting with a number, basically.

edvald: I guess it's kinda fine, we don't use tags much for anything else.

scripts/release.ts
```diff
@@ -129,8 +135,14 @@ async function release() {
...
-    "core/package.json", "core/yarn.lock",
-    "dashboard/package.json", "dashboard/yarn.lock",
+    "package.json",
+    "core/package.json",
+    "cli/package.json",
+    "dashboard/package.json",
+    "sdk/package.json",
+    "plugins/conftest-container/package.json",
+    "plugins/conftest-kubernetes/package.json",
+    "plugins/conftest/package.json"
```
edvald, line 145: We should automatically create this list from the `packages` variable.

.circleci/config.yml
thsig: Yeah. This really should be `/[0-9]+\.[0-9]+\.\[0-9]+(-[0-9]+)?` or the like.

thsig:
Closing in favor of the release branch (which includes the changes made here).

PR #2051, opened 2020-09-14
fix(core): fix default environment resolution

thsig:

This fixes a regression in the resolution flow for template strings in `project.defaultEnvironment`.


core/test/unit/src/config/project.ts
```diff
@@ -81,13 +99,14 @@ describe("resolveProjectConfig", () => {
 
   it("should resolve template strings on fields other than environment and provider configs", async () => {
     const repositoryUrl = "git://github.com/foo/bar.git#boo"
+    const defaultEnvironment = "defautl"
```
edvald, line 102: Typo in the value

core/test/data/test-project-templated/garden.yml
```diff
@@ -1,5 +1,6 @@
 kind: Project
 name: test-project-templated
+defaultEnvironment: "${local.env.TEST_VARIABLE == 'banana' ? 'local' : 'local'}"
```
edvald, line 3: Where and how is this actually tested?

core/test/unit/src/config/project.ts
thsig: Fixed.

core/test/data/test-project-templated/garden.yml
thsig: This test project is used in the `"should resolve templated env variables in project config"` case in the unit tests for the `Garden` class.

Before the fix, it failed with the error reported by Mitchell (and now no longer fails).

thsig: It would definitely have been cleaner to do this inline in the test case, but `Garden.factory` reads the file from the FS.

Maybe I should add a comment to this config referring to the test case where it's used?

edvald: Nah it's alright for this instance, but yeah going forward we should inline the configuration where possible.

edvald (approved):

PR #2061, opened 2020-09-18
(improvement): show explicit warn when include path cannot be resolved from Dockerfile ARG and ENV variables

ivanbabenko:
Added explicit warn that include path for container module cannot be resolved automatically from Dockerfile ARG and ENV variables.

Currently, auto-resolve of include path for container modules is not supported when Dockerfile contains path defined as ARG or ENV variable, which is used then as argument in COPY command. This is confirmed in the code with a comment and according TODO, but there is no clear message on what is actually wrong when one faces such case running garden build.
Without clear warn message garden just throws a generic overlap error, which doesn't provide any hint that it is actually related to ARG being used in COPY command.
It cost me some time to figure this out and I hope this tiny improvement can save time for someone else as well.

No ticket, just a general improvement for a specific described use case.


edvald (approved):
Looks great and a good catch, thank you!

edvald:
Ah just noticed the linting error, `prettier` complaining:
```
core ┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄  /home/circleci/project/core/src/plugins/container/helpers.ts:389:18
core ┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄  ERROR: 389:18   prettier                Insert `⏎··········`
core ┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄  ERROR: 390:109  no-trailing-whitespace  trailing whitespace
```
You can either fix manually or run `yarn fix-format` from the core directory, and then I'll merge!

ivanbabenko:
Linting issues are fixed, thanks for pointing out @edvald! And thanks for the review ;)

edvald:
Merged! Thanks for the PR @ivanbabenko :)

PR #2065, opened 2020-09-20
feat(cli): add --show-timestamps flag to CLI commands

eysi09:

Enables users to toggle timestamps on/off with log output

Before this PR, I was having a hard time figuring out how long some things were taking, in particular in CI and with a higher log level set. For example, builds would time out but it was hard to figure out what was taking long. Increasing the log level gives better visibility into what's going on, but it was still impossible to see exactly which steps were slow. 

This addresses that. 

I think we can probably still do a better job of giving visibility into builds natively, but `garden build -l 4 --show-timestamps` feels like a decent workaround to begin with. 

N/A

Unfortunately the spinner doesn't play well with strings that contain ANSI codes so it has to be displayed before those. I thought it made most sense to put the timestamp first, then spinner, but that means the timestamps can't have ANSI codes and look something like this:

Screenshot 2020-09-20 at 09 39 52

Alternatively we could use ANSI colours and skip the spinner (i.e. do a "basic" render) when `--show-timestamps` is set.

For example:

Screenshot 2020-09-20 at 09 46 52

This gives us more options in terms of style obviously but it feels pretty arbitrary to just disable the spinner like that (from a user's perspective). At the same time, this will likely mostly be used in a context where you don't care about the spinner. So not a a strong opinion either way. Also happy to hear suggestions on format and position in general, and style if we go that route.


edvald (approved):
This seems solid to me. I wouldn't bother with color coding and the added complexity.

PR #2068, opened 2020-09-22
Fix timestamps

eysi09:

Previously we would print the timestamp on the log entry proper, as opposed to on the latest message state. That meant that for basic log entries, the message state was being rendered would have the older timestamp of the entry itself. This fixes that.

Also, previously we'd show timestamps with the fancy logger as well but that doesn't really make sense because fancy entries update in-place and it's not clear what timestamp should mean in that context. So now we automatically set the logger to basic when timestamps are enabled. 

Fixes #

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -186,14 +186,16 @@ ${renderCommands(commands)}
     // Parse command line --var input
     const parsedCliVars = cliVars ? dotenv.parse(cliVars.join("\n")) : {}
 
+    // Init logger
+    const level = parseLogLevel(logLevel)
     let loggerType = <LoggerType>loggerTypeOpt || command.getLoggerType({ opts: parsedOpts, args: parsedArgs })
 
     if (silent || output) {
       loggerType = "quiet"
+    } else if (loggerType === "fancy" && (level > LogLevel.info || showTimestamps)) {
+      loggerType = "basic"
     }
 
-    // Init logger
-    const level = parseLogLevel(logLevel)
     const logger = initLogger({ level, loggerType, emoji, showTimestamps })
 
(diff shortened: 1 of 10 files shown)


core/src/cli/cli.ts
```diff
@@ -186,14 +186,18 @@ ${renderCommands(commands)}
...
+    const level = parseLogLevel(logLevel)
     let loggerType = <LoggerType>loggerTypeOpt || command.getLoggerType({ opts: parsedOpts, args: parsedArgs })
 
     if (silent || output) {
       loggerType = "quiet"
     }
 
-    // Init logger
-    const level = parseLogLevel(logLevel)
+    if (loggerType === "fancy" && (level > LogLevel.info || showTimestamps)) {
```
thsig, line 197: To make this more explicit, could we make this an `else if` (following the `if (silent || output)` above)?

core/src/cli/params.ts
```diff
@@ -313,7 +313,10 @@ export const globalOptions = {
     defaultValue: envSupportsEmoji(),
   }),
   "show-timestamps": new BooleanParameter({
-    help: "Show timestamps with log output.",
+    help: deline`
+      Show timestamps with log output. When enabled, Garden will use the ${chalk.bold(
+        "basic"
+      )} logger. I.e., log status changes are rendered as new lines as opposed to updated in-place.`,
```
thsig, line 319: "as opposed to updated in place" -> "instead of being updated in place"

thsig (changes requested):
Looks good! Only had a couple of minor comments there.

eysi09:
I've updated the PR.

thsig (approved):

PR #2069, opened 2020-09-22
feat(terraform): add v0.13.3 version and allow use of global terraform

### LINKED ISSUE #1988: [FEATURE]: expand terraform support
### Background / Motivation

Terraform .13 is out! But also being locked to one version of terraform is pretty limiting :(

Specify versions of Terraform besides .12.26.

Use new features / bug fixes in Terraform. `for_each` for modules!!

Since the CLI interaction with Terraform is pretty stable, I think this would mostly amount to expanding [this](https://github.com/garden-io/garden/blob/a783adc02a31cd9dac89f945aa3115033f7ed488/garden-service/src/plugins/terraform/cli.ts#L27-L65) and doing integration tests. The json output had some changes in .13 but I think it's just extra fields.

edvald:

Terraform 0.13 is nice and stable, sorry I hadn't added it sooner.
Also now allowing `null` for the version, which allows users to use a `terraform` version which is already on their path.

Closes #1988


thsig (approved):
Looks good to me 👍

PR #2073, opened 2020-09-23
fix(k8s): fix IO handling for exec command

### LINKED ISSUE #2064: The exec command doesn't work with the K8s provider
### Current Behavior

Garden appears to hang when running something like `garden exec api /bin/sh`. The reasons is that the output from the pod that's being attached to is never rendered in the users terminal.

More precisely, the `message` events for the websocket connection here are not handled:

https://github.com/garden-io/garden/blob/master/core/src/plugins/kubernetes/api.ts#L600

It's not clear to me whether the K8s library should be doing that or whether it needs to be handled explicitly. 

### Expected behavior

Output should be visible.

### Reproducible example

Run `garden exec backend /bin/sh --env remote` in the `demo-project`. 

### Workaround

Use `kubectl` directly, for example:

```
kubectl exec <pod-name> -n <namespace-name> -it -- /bin/sh
``` 

### Your environment

Lastest `master`.

thsig:

This fixes a regression that was introduced when we started using the k8s API to run commands in pods.

Fixes #2064.

### DIFF
--- a/core/src/plugins/kubernetes/api.ts
+++ b/core/src/plugins/kubernetes/api.ts
@@ -9,6 +9,7 @@
 // No idea why tslint complains over this line
 // tslint:disable-next-line:no-unused
 import { IncomingMessage } from "http"
+import { ReadStream } from "tty"
 import {
   KubeConfig,
   V1Secret,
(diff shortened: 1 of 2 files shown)


eysi09:
Might be useful to add a sanity check for this in the `test-release.sh` script.

core/src/plugins/kubernetes/api.ts
```diff
@@ -574,9 +575,29 @@ export class KubeApi {
...
+      }
+      if (stdin) {
+        /**
+         * We use raw mode for stdin to ensure that control characters aren't intercepted by the terminal and that
+         * input isn't echoed back (among other things).
+         *
+         * See https://nodejs.org/api/tty.html#tty_readstream_setrawmode_mode for more.
+         */
+        const ttyStdin = stdin as ReadStream
+        ttyStdin.setRawMode(true)
```
edvald, line 594: This can fail if something other than a TTY is passed in. I'd instead check for the `setRawMode` attribute, and set it if available (like in the `spawn()` utility function).

edvald (changes requested):

core/src/plugins/kubernetes/api.ts
thsig: Aha, good point.

thsig: Updated. Or did you also want the `rawMode` parameter added here (and, by extension, to `ExecOpts`)?

edvald: No this looks good 👍

edvald (approved):

eysi09 (changes requested):
I would really like to see a simple test for this. It’s core functionality that broke without warning and we didn’t notice for weeks.

thsig:
I'm adding a test case for this to the pre-release tests, for what it's worth. Testing the interactive shell functionality from a separate process might be a bit tricky (or so I thought).

eysi09:
> I'm adding a test case for this to the pre-release tests, for what it's worth.

Yes that's exactly what I had in mind, as per my comment above. Just requested a change because this had already been approved.

thsig:
Ok, I've added a "test case" for the exec command to `scripts/test-release.sh`.

edvald (approved):

PR #2076, opened 2020-09-24
chore: fix test-release script

thsig:

This script stopped working after the yarn migration, due to the changed version handling in our `package.json` files.

The prerelease suffix is not included in the version used for core, and also not in the file name of the release archive on GitHub. We've now added some logic to handle that.


scripts/test-release.sh
```diff
@@ -106,7 +109,7 @@ test_release() {
   echo ""
   cd ..
   cd vote-helm
-  timeout 2m ${garden_release} dev
+  timeout 5m ${garden_release} dev
```
eysi09, line 112: Does it have to be this long? It's kind of annoying to have to wait 5 minutes.

eysi09: I'd rather suggest just running the script again if it times out (second time the project will likely be deployed so it'll be faster).

thsig: Yeah, good point.

eysi09 (approved):

PR #2082, opened 2020-09-28
fix(workflows): forbid use of global options

thsig:

Global options (such as `--env` or `-l`) have no effect when used in workflow step commands, since the step commands share a Garden instance with the enclosing `run workflow` command.

To prevent user confusion, we now validate that global options are not used in workflow step command specs, and throw an error if any are detected.

For example, the following workflow config
```yaml
kind: Workflow
name: test-workflow
steps:
  - command: [test, "--env=foo", "-l", "4"
```
Would fail validation with the following error message:
```
$ garden validate

Invalid step command options for workflow test-workflow:

In command [test, --env=foo, -l, 4]: --env, -l

```

### DIFF
--- a/core/src/config/workflow.ts
+++ b/core/src/config/workflow.ts
@@ -6,7 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import { cloneDeep, isEqual, omit, take } from "lodash"
+import { cloneDeep, isEqual, merge, omit, take } from "lodash"
 import { joi, joiUserIdentifier, joiVariableName, joiIdentifier } from "./common"
 import { DEFAULT_API_VERSION } from "../constants"
 import { deline, dedent } from "../util/string"
@@ -19,6 +19,9 @@ import { ConfigurationError } from "../exceptions"
 import { coreCommands } from "../commands/commands"
 import { CommandGroup } from "../commands/base"
 import { EnvironmentConfig, getNamespace } from "./project"
+import { globalOptions } from "../cli/params"
+import { isTruthy } from "../util/util"
+import { parseCliArgs, pickCommand } from "../cli/helpers"
 
 export interface WorkflowConfig {
   apiVersion: string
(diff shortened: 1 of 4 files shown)


edvald:
I think it'd be good to add further explanation in the error message. It's not at all clear _why_ these parameters might be disallowed. If only to say "certain global option flags are currently not supported for workflow command steps", or something along those lines.

thsig:
Yeah, good point. I'll add some informational text to that effect.

thsig:
Ok, made some changes here. Now the output would be:
```
$ garden validate
Invalid step command options for workflow test-workflow:

In command [test, --env=foo, -l, 4]: --env, -l
(available options: --name, --force, --force-build, --watch)

Global options (such as --env or --log-level) are not available in workflow step commands.

```
Does that look good?

core/src/config/workflow.ts
```diff
@@ -346,7 +369,52 @@ function validateSteps(config: WorkflowConfig) {
...
+
+function findInvalidOptions(commandSpec: string[]) {
+  const stripOptionValue = (option: string) => option.replace(/=.*/, "")
+  const invalidOptions = commandSpec
+    .filter((c) => {
+      let opt = c.split("--")[1] || c.split("-")[1]
+      opt = opt && stripOptionValue(opt)
+      return globalOptionNamesSet.has(opt)
+    })
+    .map((option) => stripOptionValue(option))
```
edvald, line 386: Shouldn't we be using proper argument parsing here? We're already doing that properly in our CLI code. This looks quite brittle and untested.

thsig: Yeah, good point.

thsig: I've updated this PR to use `parseCliArgs` instead. How does it look now?

core/src/config/workflow.ts
```diff
@@ -319,23 +322,38 @@ function getStepCommands() {
     .filter((cmd) => cmd.workflows)
 }
 
+const globalOptionNames = Object.keys(globalOptions).sort()
+const globalOptionNamesSet = new Set<string>(globalOptionNames)
```
edvald, line 326: This isn't really necessary imo. It takes almost no time to scan through a short list.

core/src/config/workflow.ts
```diff
@@ -319,23 +322,38 @@ function getStepCommands() {
...
-  if (invalidSteps.length > 0) {
+  if (stepsWithInvalidPrefix.length > 0) {
     const msgPrefix =
-      invalidSteps.length === 1
-        ? `Invalid step command for workflow ${config.name}:`
-        : `Invalid step commands for workflow ${config.name}:`
-    const descriptions = invalidSteps.map((step) => `[${step.command!.join(", ")}]`)
+      stepsWithInvalidPrefix.length === 1
+        ? `Invalid step command prefix for workflow ${config.name}:`
+        : `Invalid step command prefixes for workflow ${config.name}:`
```
edvald, line 355: Is "command prefix" a sensible user-facing term?

core/src/config/workflow.ts
thsig: Yeah, true that.

core/src/config/workflow.ts
```diff
@@ -346,7 +364,63 @@ function validateSteps(config: WorkflowConfig) {
...
+      availableOptions.length > 0 ? `(available options: ${availableOptions.map((opt) => `--${opt}`).join(", ")})` : ""
+    const errMsg = dedent`
+      Invalid options in step command [${step.command!.join(", ")}]: ${usedGlobalOptions.join(", ")}
+      ${availableDescription}
+    `
+    error = new ParameterError(errMsg, {
+      availableOptions,
+      invalidOptions: usedGlobalOptions,
+    })
+    return { step, error }
```
edvald, line 402: I don't believe this error is thrown anywhere directly. Might as well make it just a string, right?

thsig:
I think I've addressed the outstanding PR comments now.

edvald (approved):

PR #2089, opened 2020-10-02
Module Templates

edvald:
This adds a powerful new templating feature, for generating entire modules.

Here are the draft docs: https://github.com/garden-io/garden/blob/14ddee2648e7db8c9a026bf207a79fd382c96524/docs/using-garden/module-templates.md
And the included example: https://github.com/garden-io/garden/tree/14ddee2648e7db8c9a026bf207a79fd382c96524/examples/templated-k8s-container

Please review both from a code and design perspective. See if the terminology makes sense, and that the docs feel complete and understandable.


core/src/garden.ts
```diff
@@ -1015,21 +1026,41 @@ export class Garden {
...
 
-      throwOnMissingSecretKeys(Object.fromEntries(rawModuleConfigs.map((c) => [c.name, c])), this.secrets, "Module")
-      throwOnMissingSecretKeys(Object.fromEntries(rawWorkflowConfigs.map((c) => [c.name, c])), this.secrets, "Workflow")
+      // Resolve bundles
+      const templatesByKey = keyBy(bundleTemplates, "name")
+      const resolvedBundled = await Bluebird.map(rawBundleResources, (r) => resolveBundle(this, r, templatesByKey))
+      rawModuleConfigs.push(...resolvedBundled.flatMap((c) => c.modules))
 
       await Bluebird.all([
         Bluebird.map(rawModuleConfigs, async (config) => this.addModuleConfig(config)),
```
thsig, line 1066: How about including the source bundle template name in the error message for duplicate module names (in `addModuleConfig` below)? Might help users pinpoint the problem in their config more quickly.

Then again, it might already be obvious enough from the generated module path.

Just a thought.

thsig (changes requested):
Looks great! Had a few minor comments / suggestions.

core/src/garden.ts
edvald: Fair point, but I think it's clear enough from the config paths

thsig (approved):

PR #2090, opened 2020-10-05
fix(core): omit dependencyResults from events

thsig:

We've now removed `dependencyResults` from the payload types of the `taskComplete` and `taskError` events (which were previously `GraphResult`).

Since `dependencyResults` is a recursive data structure, it can grow very large. And since `dependencyResults` in event payloads is not currently being used anywhere, it can safely be removed.

We also omit `result.output.log` and `result.output.buildLog` when present, since those strings can grow arbitrarily long (and since we're not currently using those payload fields anywhere).

### DIFF
--- a/core/src/events.ts
+++ b/core/src/events.ts
@@ -6,6 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+import { omit } from "lodash"
 import { EventEmitter2 } from "eventemitter2"
 import { GraphResult } from "./task-graph"
 import { LogEntryEvent } from "./enterprise/buffered-event-stream"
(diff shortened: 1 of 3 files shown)


core/src/events.ts
```diff
@@ -59,6 +60,16 @@ export interface LoggerEvents {
 
 export type LoggerEventName = keyof LoggerEvents
 
+export type GraphResultEventPayload = Omit<GraphResult, "dependencyResults">
+
+export function toGraphResultEventPayload(result: GraphResult): GraphResultEventPayload {
+  const payload = { ...omit(result, "dependencyResults") }
```
eysi09, line 66: You can skip the braces + spread here, right?

eysi09 (approved):
Looks good, but proposed one syntax change.

Also, this doesn't fix the underlying issue of lugging this potentially huge object around and for example dumping it as YAML on workflow errors. Just rendering the YAML string could be very expensive for some projects and regardless, the size of the output makes the logs pretty much unusable. 

But this is a good first step towards unblocking our CI pipeline.

PR #2091, opened 2020-10-05
fix(enterprise): fix batch sizing logic

thsig:

We now correctly handle the case where a single event or log entry payload exceeds `MAX_BATCH_BYTES` (dropping the event / log entry and logging an error).

Also wrote a couple of unit tests for the batching logic.

In addition, the log level for error messages generated when posting to targets failed was raised, to avoid cluttering the log.

### DIFF
--- a/core/test/unit/src/enterprise/buffered-event-stream.ts
+++ b/core/test/unit/src/enterprise/buffered-event-stream.ts
@@ -11,7 +11,11 @@ import { StreamEvent, LogEntryEvent, BufferedEventStream } from "../../../../src
 import { getLogger } from "../../../../src/logger/logger"
 import { Garden } from "../../../../src/garden"
 import { makeTestGardenA } from "../../../helpers"
-import { find, isMatch } from "lodash"
+import { find, isMatch, range, repeat } from "lodash"
+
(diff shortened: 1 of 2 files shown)


core/test/unit/src/enterprise/buffered-event-stream.ts
```diff
@@ -89,4 +98,40 @@ describe("BufferedEventStream", () => {
...
+      const batch = bufferedEventStream.makeBatch(records)
+      const batchSize = Buffer.from(JSON.stringify(batch)).length
+      expect(batch.length).to.be.lte(records.length)
+      expect(batch.length).to.be.lte(MAX_BATCH_BYTES / (recordSizeKb * 1024))
+      expect(batchSize).to.be.lte(MAX_BATCH_BYTES)
+    })
+
+    it("should drop individual records whose payload size exceeds MAX_BATCH_BYTES", async () => {
+      const recordSizeKb = 30
+      // const recordSizeKb = MAX_BATCH_BYTES * 1024 + 3
```
eysi09, line 118: Remove.

eysi09 (approved):

PR #2097, opened 2020-10-07
improvement(core): resolve remote sources in parallel

edvald:

We were unnecessarily resolving remote sources sequentially, which would slow down initialization quite a bit with multiple remote sources.

cc @mitchfriedman


mitchfriedman (approved):
Great feature, thanks for getting this in!

thsig (approved):

thsig:
Note: We'll need to squash before merging.

PR #2099, opened 2020-10-08
feat(k8s): enable hot reloading for kubernetes modules

eysi09:

Enables hot reloading for the `kubernetes` module type.

Fixes #

I lumped the hot reload handler for the `kubernetes` and for the `helm` module into a single handler and as result had to move some files around. Concretely, we now have all the K8s related hot reload logic in `kubernetes/hot-reload/hot-reload.ts` and `kubernetes/hot-reload/hot-reload.ts` instead of being split up between module types.


core/src/plugins/kubernetes/kubernetes-module/handlers.ts
```diff
@@ -106,44 +111,60 @@ export async function deployKubernetesService(
...
+      })
+
+      // Replace the original hot reload resource with the modified spec
+      otherManifests = otherManifests
+        .filter((m) => !(m.kind === hotReloadTarget!.kind && hotReloadTarget?.metadata.name === m.metadata.name))
+        .concat(<KubernetesResource<BaseResource>>hotReloadTarget)
+    }
+
     await apply({ log, ctx, provider: k8sCtx.provider, manifests: otherManifests, pruneSelector })
-    await waitForResources({
```
eysi09, line 131: This was being called after applying namespaces, again here, and finally once more below, for good measure I suppose.

Now it's only called once after all the `apply` calls. Does that make sense or where we doing all the waiting for resources for a reason?

core/test/integ/src/plugins/kubernetes/hot-reload.ts
```diff
@@ -8,16 +8,19 @@
...
+import { PluginContext } from "../../../../../src/plugin-context"
+import { KubernetesProvider } from "../../../../../src/plugins/kubernetes/config"
+import { findServiceResource, getServiceResourceSpec } from "../../../../../src/plugins/kubernetes/util"
+import {
+  getHotReloadSpec,
+  getHotReloadContainerName,
+  configureHotReload,
+} from "../../../../../src/plugins/kubernetes/hot-reload/helpers"
 
 describe("getHotReloadSpec", () => {
```
eysi09, line 25: This tests the functions in `kubernetes/hot-reload` for the a helm module. I could duplicate them for a K8s module but the code paths are almost identical so I figured it wasn't needed.

core/test/setup.ts
```diff
@@ -28,7 +28,7 @@ before(async () => {
 
 after(() => {
   // tslint:disable-next-line: no-console
-  console.log(getDefaultProfiler().report())
+  // console.log(getDefaultProfiler().report())
```
thsig, line 31: Remove comment.

examples/hot-reload/README.md
```diff
@@ -13,25 +13,24 @@ This project contains a single service called `node-service`. When running, the
 In the `garden.yml` file of the `node-service` module we first enable hot-reloading and specify the target directory it should hot-reload changed sourcefiles into:
 
 ```yaml
-...
+
+---
```
thsig, line 17: Just curious—is there a reason we're changing from `...` to `---` here to indicate "more above/below"?

thsig (changes requested):
Looks good! Only had a couple of minor comments.

examples/hot-reload/README.md
eysi09: Nope, not sure how that happened. I'll revert it.

core/src/plugins/kubernetes/kubernetes-module/handlers.ts
eysi09: Could I get a second opinion from @edvald on this one? In the previous implementation we had three `waitForResources` calls, was that just a mistake? I removed two of them but just want to make sure I'm not missing something.

thsig: Ah, this one slipped past me. It was a bit late when I read through it.

The thing is, the non-namespace resources could very well belong inside the namespaces being created. So we should be waiting for the namespaces to be up first (like we were doing before this change), to ensure that the namespaces exist before applying the non-namespace resources.

The third `waitForResources` call was probably unnecessary though. I think it's a leftover from this PR: https://github.com/garden-io/garden/pull/1942/files.

thsig: One other question: Have we verified that deploying a `kubernetes` service with/without hot reloading and then redeploying it with the opposite setting (without/with hot reloading) results in a redeploy, as expected?

examples/hot-reload/README.md
eysi09: ...I think this was my YAML formatter fixing it on autosave.

eysi09:
@thsig, I addressed the PR comments in a separate commit.

core/src/plugins/kubernetes/kubernetes-module/handlers.ts
eysi09: Good catch! It does not. How are we doing this for Helm?

eysi09:
@thsig, I've addressed the issue with redeploys not working in a new commit on top of the others.

thsig (approved):

PR #2104, opened 2020-10-12
fix(core): fix circular dependency detection

thsig:

Before this fix, circular build dependencies between modules would result in an internal error (instead of an informative error message).

### DIFF
--- a/core/test/unit/src/garden.ts
+++ b/core/test/unit/src/garden.ts
@@ -2784,6 +2784,63 @@ describe("Garden", () => {
     })
   })
 
+  describe("getConfigGraph", () => {
+    it("should throw an error if modules have circular build dependencies", async () => {
+      const garden = await TestGarden.factory(pathFoo, {
+        config: {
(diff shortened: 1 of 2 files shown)


core/test/unit/src/garden.ts
```diff
@@ -2784,6 +2784,58 @@ describe("Garden", () => {
...
+          serviceConfigs: [],
+          taskConfigs: [],
+          testConfigs: [],
+          spec: {},
+        },
+      ])
+
+      await expectError(
+        () => garden.getConfigGraph(garden.log),
+        (err) => expect(err.message).to.match(/Detected circular dependencies between module configurations/)
```
edvald, line 2834: Might as well test the whole string, no? Especially since the code that generates it is a little opaque.

thsig: Yeah, good point. Will do.

thsig: Ok, should be good now.

edvald (approved):

PR #2105, opened 2020-10-13
fix(core): create artifact placeholder in a world-readable place

wyrie:

Fixes #2105 


core/src/plugins/kubernetes/run.ts
```diff
@@ -367,7 +367,7 @@ export async function runAndCopy({
             // Tarball the requested files and stream to the above extractor.
             runner
               .exec({
-                command: ["sh", "-c", "cd / && touch .DS_Store && " + tarCmd.join(" ")],
+                command: ["sh", "-c", "cd /tmp && touch .DS_Store && " + tarCmd.join(" ")],
```
edvald, line 370: This approach should work, but you do need to `cd` back to the root before the tar command. After that it should be fine, as far as I can tell.

edvald (changes requested):

core/src/plugins/kubernetes/run.ts
edvald: Or just change it to `touch /tmp/.DS_Store`. I'll later modify the code to make it a bit more obvious what's going on there, instead of using this particular filename that might be a source of confusion.

wyrie: That works now, thanks.

wyrie (commented):
https://github.com/garden-io/garden/pull/2105/commits/ee236aa75114ccee159e7bd6325b6cea7529425d

edvald (approved):

PR #2116, opened 2020-10-20
feat(workflows): implemented when modifier

thsig:

Workflow steps can now use the `when: always | never | onError` modifier.

Steps with `when: always` can be used e.g. for running scripts or commands at the end of the workflow, regardless of whether the workflow failed.

Steps with `when: onError` can facilitate custom error handling and cleanup after failed steps or sequences of steps.

See the updated workflows guide (under Using Garden) for more details.

### DIFF
--- a/core/src/commands/run/workflow.ts
+++ b/core/src/commands/run/workflow.ts
@@ -7,14 +7,14 @@
  */
 
 import chalk from "chalk"
-import { cloneDeep, merge, repeat } from "lodash"
+import { cloneDeep, flatten, last, merge, repeat, size } from "lodash"
 import { printHeader, getTerminalWidth, formatGardenError, renderMessageWithDivider } from "../../logger/util"
 import { Command, CommandParams, CommandResult } from "../base"
 import { dedent, wordWrap, deline } from "../../util/string"
 import { Garden } from "../../garden"
 import { WorkflowStepSpec, WorkflowConfig, WorkflowFileSpec } from "../../config/workflow"
 import { LogEntry } from "../../logger/log-entry"
-import { GardenError, GardenBaseError } from "../../exceptions"
+import { GardenError, WorkflowScriptError } from "../../exceptions"
 import { WorkflowConfigContext, WorkflowStepConfigContext, WorkflowStepResult } from "../../config/config-context"
 import { resolveTemplateStrings, resolveTemplateString } from "../../template-string"
 import { ConfigurationError, FilesystemError } from "../../exceptions"
@@ -87,7 +87,12 @@ export class RunWorkflowCommand extends Command<Args, {}> {
       steps: {},
     }
 
+    let stepErrors: StepErrors = {}
+
     for (const [index, step] of steps.entries()) {
+      if (shouldBeDropped(index, steps, stepErrors)) {
+        continue
+      }
       printStepHeader(outerLog, index, steps.length, step.description)
 
       const stepName = getStepName(index, step.name)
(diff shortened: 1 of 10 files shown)


eysi09:
Nice! Looking forward to adding this our own workflows.

docs/using-garden/workflows.md
```diff
@@ -77,6 +77,48 @@ steps:
...
+
+The `skip` field is a boolean. If its value is `true`, the step will be skipped, and the next step will be run as if the skipped step succeeded.
+
+Note that skipped steps don't produce any outputs (see the [step outputs](#step-outputs) section below for more). However, skipped steps are shown in the command log.
+
+The `when` field can be used with the following values:
+* `always`: The step will always be run, regardless of whether any previous steps have failed.
+* `never`: The step will always be ignored, even if all previous steps succeeded. Note: Ignored steps don't show up in the command logs.
+* `onError`: This step will be run if a preceding step failed, or if its preceding step has `when: onError`.
+ If the next step has `when: onError`, it will also be run. Otherwise, all subsequent steps are ignored.
```
edvald, line 92: This feels a little odd or at least unclear to me: "... or if its preceding step has `when: onError`". Can you elaborate on what that means exactly?

docs/using-garden/workflows.md
```diff
@@ -77,6 +77,48 @@ steps:
...
+      echo "Run if deploy step failed"
+    when: onError
+  - script: |
+      echo "Also run if deploy step failed"
+    when: onError
+  - command: [build]
+    when: never # This is never run
+  - command: [test]
+  - script: |
+      echo "Run if test step failed, but not if deploy step failed"
```
edvald, line 115: Ok I _think_ I get this now... So, `onError` only applies to when the step directly preceding fails? I get the basic idea, but I worry a bit that we're either not explaining this well enough, or that the semantics are too complex.

Also feels a bit like the one big example with a bunch of conditions is a little hard to digest.

thsig: Gotcha. What do you think, @eysi09?

eysi09: I think the error handling makes sense as is, but perhaps we could explain it a little better. Did you have something else in mind @edvald?

core/src/commands/run/workflow.ts
```diff
@@ -314,15 +317,92 @@ export async function runStepCommand({
...
+      }
+    }
+    /**
+     * If true, then there is one or more `onError` step between this step and the step that threw the error,  and
+     * there's also a non-`onError`/`never` step in between. That means that it's not up to this sequence of `onError`
+     * steps to "handle" that error.
+     *
+     * Example: Here, steps a, b and c don't have a `when` modifier, and e1, e2 and e3 have `when: onError`.
+     *   [a, b, e1, e2, c, e3]
+     * If a throws an error, we run e1 and e2, but drop e3.
```
eysi09, line 391: `c` is dropped as well, right?

docs/using-garden/workflows.md
eysi09: The example in the [code comment above](https://github.com/garden-io/garden/pull/2116/files#diff-537f3b689f163c87bda525793a193217f76fb7ef4582c41eb5cb29063dd2c7e8R389) actually does a good job of explaining this IMO. Perhaps we could just include that inline here.

core/src/commands/run/workflow.ts
```diff
@@ -314,15 +317,92 @@ export async function runStepCommand({
...
   return result
 }
 
+class WorkflowScriptError extends GardenBaseError {
+  type = "workflow-script"
+}
+
+export async function runStepScript({ garden, bodyLog, step }: RunStepParams): Promise<CommandResult<any>> {
+  try {
+    await runScript(bodyLog, garden.projectRoot, step.script!)
```
eysi09, line 335: FYI, this not platform agnostic and assumes bash is available on the local host. See also #2124. We need a platform agnostic way to run this, although out of scope for this PR.

eysi09 (commented):
Had a few comments but otherwise looks good.

core/src/commands/run/workflow.ts
thsig: Yeah, that's right. The comment should explicitly note that.

core/src/commands/run/workflow.ts
```diff
@@ -87,7 +87,12 @@ export class RunWorkflowCommand extends Command<Args, {}> {
       steps: {},
     }
 
+    let stepErrors: StepErrors = {}
+
     for (const [index, step] of steps.entries()) {
+      if (shouldBeDropped(index, steps, stepErrors)) {
```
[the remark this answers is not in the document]
thsig: We may want to handle skipped and dropped steps differently in the UI for clarity, so I think we should keep this as is (but make the documentation clearer in this PR).

thsig:
I believe I've addressed all the comments. Is the documentation clear enough now?

eysi09 (approved):
LGTM!

PR #2117, opened 2020-10-21
fix(kubernetes-module): fix regression

thsig:

This fixes a regression introduced when we added hot reloading support for the `kubernetes-module` module type.

Also added an integ test case to guard against this regression in the future.


edvald (commented):
Changes look good, but tests are failing.

PR #2126, opened 2020-11-01
docs: small fixes

eysi09:


thsig:
Looks like the docs aren't up to date (and when I run `generate-docs`, the lines that were added to `docs/reference/providers/README.md` are removed).

Maybe something's up with the logic that generates the list of provider reference links there?

eysi09:
Should be fixed now.

thsig (approved):

PR #2127, opened 2020-11-04
fix(k8s): containerModule refs in helm modules not creating build deps

edvald:
This resolves an internal issue where module configure handlers were
not able to reliably add a build dependency. This caused problems with
another PR that I have in progress, but could also cause issues when
using helm and kubernetes modules referencing container modules.

The solution was to no longer use the TaskGraph for module resolution,
and instead use a more bespoke flow that could handle dynamic addition
of dependencies during resolution.

Existing tests should mostly handle this. I just added one test to
validate the specific case of adding a build dependency in a configure
handler.


thsig (approved):
Looks good to me 👍

PR #2128, opened 2020-11-04
fix(enterprise): don't resolve config on login

thsig:

Fixes an issue where using secrets outside of providers in project configuration would break the login command.

### DIFF
--- a/core/src/commands/login.ts
+++ b/core/src/commands/login.ts
@@ -10,12 +10,19 @@ import { Command, CommandParams, CommandResult } from "./base"
 import { printHeader } from "../logger/util"
 import dedent = require("dedent")
 import { login } from "../enterprise/auth"
-import { ConfigurationError } from "../exceptions"
+import { CommandError, ConfigurationError } from "../exceptions"
+import { findProjectConfig } from "../config/base"
 
 export class LoginCommand extends Command {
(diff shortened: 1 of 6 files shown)


core/src/commands/login.ts
```diff
@@ -10,20 +10,34 @@ import { Command, CommandParams, CommandResult } from "./base"
...
+   * only available after we've logged in).
+   */
+  noProject = true
+
   description = dedent`
     Logs you in to Garden Enterprise. Subsequent commands will have access to enterprise features.
   `
 
-  async action({ garden, log, headerLog }: CommandParams): Promise<CommandResult> {
+  async action({ garden: dummyGarden, log, headerLog }: CommandParams): Promise<CommandResult> {
```
10ko, line 31: I am not sure `dummyGarden` is the best name, because we use the following line:
`const projectConfig = await findProjectConfig(dummyGarden.projectRoot)`
makes me think we are extracting the `projectRoot` from a dummy project.

thsig: Aha yeah, good point.

thsig: Should be fine now.

10ko (approved):
Thanks @thsig!

PR #2130, opened 2020-11-05
feat(core): provide git branch to template strings

### LINKED ISSUE #2086: [FEATURE]: Make current git branch available as variable
## Feature Request
Make the current git branch available as a pre-defined garden variable.

### Background / Motivation
We would like to use this as part of our defaultNamespace value. Currently we either have to deploy by manually specifying the branch name with --env or ensure the branch name is available as an evironment variable. 

Set, using say local.git-branch, the value of the current branch anywhere in the garden project

It creates an easy and automated way to work with multiple feature branches by linking them to garden environments.

thsig:

The current Git branch can now be used in template strings under `${git.branch}`.

Closes #2086.

### DIFF
--- a/core/src/config/config-context.ts
+++ b/core/src/config/config-context.ts
@@ -296,6 +296,33 @@ class ProjectContext extends ConfigContext {
   }
 }
 
+class GitContext extends ConfigContext {
+  @schema(
+    joi
+      .string()
+      .description(
+        dedent`
+          The current Git branch, if available. Resolves to an empty string if HEAD is in a detached state
+          (e.g. when rebasing), or if the repository has no commits.
+
+          When using remote sources, the branch used is that of the project/top-level repository (the one that contains
+          the project configuration).
+
+          The branch is computed at the start of the Garden command's execution, and is not updated if the current
+          branch changes during the command's execution (which could happen, for example, when using watch-mode
+          commands).
+        `
+      )
+      .example("my-feature-branch")
+  )
+  public branch: string
+
+  constructor(root: ConfigContext, branch: string) {
+    super(root)
+    this.branch = branch
+  }
+}
+
 /**
  * This context is available for template strings in the `defaultEnvironment` field in project configs.
  */
@@ -310,17 +337,25 @@ export class DefaultEnvironmentContext extends ConfigContext {
   @schema(ProjectContext.getSchema().description("Information about the Garden project."))
   public project: ProjectContext
 
+  @schema(
+    GitContext.getSchema().description("Information about the current state of the project's local git repository.")
+  )
+  public git: GitContext
+
   constructor({
     projectName,
     artifactsPath,
+    branch,
     username,
   }: {
     projectName: string
     artifactsPath: string
+    branch: string
     username?: string
   }) {
     super()
     this.local = new LocalContext(this, artifactsPath, username)
+    this.git = new GitContext(this, branch)
     this.project = new ProjectContext(this, projectName)
   }
 }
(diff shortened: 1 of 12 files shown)


thsig:
Ah, looks like I used an option that's not supported by the Git version used in the test suite. Will have a look.

edvald:
I wonder if it makes more sense to make this `${git.branch}` instead? Or `${local.git.branch}` perhaps.

thsig:
I've changed this template variable to `${git.branch}` now, so I think we should be good here.

core/src/config/config-context.ts
```diff
@@ -340,16 +364,6 @@ export interface ProjectConfigContextParams {
...
 export class ProjectConfigContext extends DefaultEnvironmentContext {
-  @schema(
-    LocalContext.getSchema().description(
-      "Context variables that are specific to the currently running environment/machine."
-    )
-  )
-  public local: LocalContext
-
-  @schema(ProjectContext.getSchema().description("Information about the Garden project."))
-  public project: ProjectContext
```
edvald, line 351: I'm assuming this is unintentional?

thsig: It's intentional, actually—`ProjectConfigContext` extends `DefaultEnvironmentContext`, which has the `local` and `project` keys, so these declarations were essentially duplicates.

The relevant tests pass (and the generated docs haven't changed when it comes to these variables).

thsig:
Tests are still failing in CI (but pass locally). Maybe the git versions are different.

core/src/config/config-context.ts
edvald: Right, got it.

core/src/vcs/git.ts
```diff
@@ -417,4 +409,30 @@ export class GitHandler extends VcsHandler {
...
+      return (await git("rev-parse", "--abbrev-ref", "HEAD"))[0]
+    } catch (err) {
+      if (err.exitCode === 128) {
+        try {
+          // If this doesn't throw, then we're in a repo with no commits.
+          await git("rev-list", "-n", "1", "--all")
+          return undefined
+        } catch (notInRepoError) {
+          // Throw nice error when we detect that we're not in a repo root
+          throw new RuntimeError(notInRepoRootErrorMessage(path), { path })
```
edvald, line 425: We need a more specific error check here (the git command could be failing in any number of ways other than not being in a repo root). That should bring to light what's actually failing in the tests.

edvald, line 425: (Oops, meant to edit and not delete what I previously wrote)

It seems we also need a more specific error check for this nested git command. Or perhaps we should generally fail silently here, and just document that `${git.branch}` may be undefined in certain cases? We might for example be on a detached HEAD (e.g. rebasing).

thsig:
I think this should be good now. Are the comments etc. clear enough?

thsig:
Are there any outstanding comments on this one?

edvald:
One thought: If running in watch mode, the branch won't update unless Garden configuration is changed. I wonder if it makes sense to at least document that? The alternative would be to watch something in the `.git` folder but I'm not sure we want to get into such things now.

Also, it's not defined in documentation what to expect when using remote sources. I think it makes sense to always use the project repo branch (although I suppose it is kind of debatable), but we should at least clarify that in docs.

thsig:
> One thought: If running in watch mode, the branch won't update unless Garden configuration is changed. I wonder if it makes sense to at least document that?

Yeah, agreed. I'll add a description of that to the docstring.

> The alternative would be to watch something in the .git folder but I'm not sure we want to get into such things now.

Aye. That was my thought initially, but I think we should leave that for another PR if that feature request comes up from users.

> Also, it's not defined in documentation what to expect when using remote sources. I think it makes sense to always use the project repo branch (although I suppose it is kind of debatable), but we should at least clarify that in docs.

Good point. I'll describe the current behavior more fully in the docstring.

thsig:
I've updated the docstring.

edvald (approved):

PR #2132, opened 2020-11-09
fix(core): ensure primitive values are not cast on schema validation

eysi09:

Fixes issue where primitive values would get cast on schema validation. For example, something like `DB_PORT: "8080"` would be resolved as `DB_PORT: 8080`.

Fixes #


thsig:
Looks good to me.

thsig (approved):

PR #2133, opened 2020-11-09
improvement(enterprise): log secrets fetching

thsig:

Added a log line for fetching secrets.

Before, the command would seem to hang for a bit while secrets were were being fetched. Now, it's clearer what Garden is waiting on during that time.

### DIFF
--- a/core/src/enterprise/secrets/garden/get-secrets.ts
+++ b/core/src/enterprise/secrets/garden/get-secrets.ts
@@ -6,28 +6,36 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+import chalk from "chalk"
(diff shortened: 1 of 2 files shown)


eysi09 (changes requested):
We need to solve this differently. Anything we log before the command header like this jumps around when we render the actual command header (when not in watch mode).

We could either hoist the command header logging to the `cli/cli.ts` file or ensure all commands have a `prepare` function that does the logging early.

eysi09:
@thsig, would be great to close this one soon.

eysi09:
Also, open question whether we should explicitly log "Fetching secrets" or rather something like "Connecting to Garden Enterprise..." or similar. Thoughts?

eysi09:
Closing in favour of #2196

PR #2134, opened 2020-11-09
fix(enterprise): whitelist commands for streaming

thsig:

We now only stream events and log entries to GE for whitelisted commands (those with `streamEvents = true` and/or `streamLogEntries = true`).

I couldn't think of a decent way to test this that didn't rely on counting invocations to `BufferedEventStream#connect`. Is this fine as is?

### DIFF
--- a/core/src/commands/call.ts
+++ b/core/src/commands/call.ts
@@ -46,6 +46,8 @@ export class CallCommand extends Command<Args> {
   name = "call"
   help = "Call a service ingress endpoint."
 
+  streamEvents = true
+
   description = dedent`
     Resolves the deployed ingress endpoint for the given service and path, calls the given endpoint and
     outputs the result.
--- a/core/src/commands/get/get-status.ts
+++ b/core/src/commands/get/get-status.ts
@@ -50,6 +50,7 @@ export class GetStatusCommand extends Command {
   help = "Outputs the full status of your environment."
 
   workflows = true
+  streamEvents = true
 
   outputsSchema = () =>
     joi.object().keys({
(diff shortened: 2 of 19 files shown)


core/src/commands/call.ts
```diff
@@ -46,6 +46,8 @@ export class CallCommand extends Command<Args> {
   name = "call"
   help = "Call a service ingress endpoint."
 
+  streamEvents = true
```
eysi09, line 49: Do we need to stream this?

core/src/commands/get/get-status.ts
```diff
@@ -50,6 +50,7 @@ export class GetStatusCommand extends Command {
   help = "Outputs the full status of your environment."
 
   workflows = true
+  streamEvents = true
```
eysi09, line 53: Or this?

core/src/commands/get/get-task-result.ts
```diff
@@ -37,6 +37,7 @@ export class GetTaskResultCommand extends Command<Args> {
   help = "Outputs the latest execution result of a provided task."
 
   workflows = true
+  streamEvents = true
```
eysi09, line 40: Or this?

core/src/commands/get/get-test-result.ts
```diff
@@ -41,6 +41,7 @@ export class GetTestResultCommand extends Command<Args> {
   help = "Outputs the latest execution result of a provided test."
 
   workflows = true
+  streamEvents = true
```
eysi09, line 44: Or this?

eysi09 (commented):
Wondering if we should be streaming some of these commands. Does GE need to know if a user calls the `get status` command?

core/src/commands/call.ts
thsig: The reason we're streaming from this command is that it calls `getServiceStatus` (which may result in Garden emitting an updated service status).

Same applies to the `get status` command below.

That said, I think it makes more sense to stream only events (and not log entries) for most commands. There are many cases where we're interested in the runtime status events but not the log output.

eysi09 (approved):

PR #2137, opened 2020-11-13
improvement(core): allow wildcard in first label in hostname

eysi09:

Allow wildcards in the first DNS label of a hostname. I.e., we allow `*.foo.bar`, but not `foo.*.bar`. This is consistent with how Kubernetes treats hostnames.

Fixes #

Only allowing wildcards in the first label in ingress hostnames is pretty contested in the K8s community, see for example this issue: https://github.com/kubernetes/kubernetes/issues/41881.

However, I figured it'd make sense to start with the same restrictions as K8s since that's where the field is most used.


edvald (approved):

ArnaudRinquin:
This was required in our setup and everything works just fine now, thanks a lot for this! 🙏

Issue #2098, opened 2020-10-08
cert-manager integration ignores secretRef's namespace


LeoDupont:
### Current Behavior

I have this Garden project, leveraging the [cert-manager integration](https://docs.garden.io/advanced/cert-manager-integration):

```yaml
kind: Project
name: my-cloud

defaultEnvironment: dev
environments:
  - name: dev
    defaultNamespace: dev

providers:
  - name: kubernetes
    environments: [dev]
    context: gke_redacted_context
    buildMode: cluster-docker
    setupIngressController: false # using Istio
    certManager:
      install: false # installed manually
      email: my@redacted-email.com
      issuer: acme
      acmeChallengeType: HTTP-01
      acmeServer: letsencrypt-staging
    tlsCertificates:
      - name: dev-cert-01
        managedBy: cert-manager
        hostnames:
          - my.redacted.domain.com
        secretRef:
          name: dev-tls
          namespace: istio-system # <= this is the ignored bit
```

> (Not related to Garden, but for the records and for other Istio users, I'd like to mention the fact that the acme-solver ingress created didn't play well with my Istio gateway, but I managed to work around that by adding a route in my Istio VirtualService for `/.well-known` to the acme-solver service... this isn't good for automatic certificate renewal, but that's another topic).

The bug is that the `dev-tls` secret is not created in the `istio-system` namespace, but rather in the Garden project's namespace, that is `my-cloud-dev`.

Incidentally, the HTTPS traffic was not working at all and it took me a while to realize the secret was simply not in the right Kubernetes namespace. 

### Expected behavior

I was expecting the `dev-tls` secret to be created in the `istio-system` namespace, and not the current project's namespace.

### Workaround

After [exporting the secret](https://www.revsys.com/tidbits/copying-kubernetes-secrets-between-namespaces/) from `my-cloud-dev` to `istio-system`, the HTTPS traffic worked out. 

But that's not great for automatic certificates renewal.

### Suggested solution(s)

It seems that the `secretRef.namespace` property is ignored somehow. But I don't know in which situations.

### Your environment

* **OS:** Windows 10 Pro 2004
* **How I'm running Kubernetes:** GKE
* `garden version`: 0.12.6

Please tell me if I can provide more info.

eysi09:
Hi @LeoDupont and apologies for the late reply.

We've noticed this on our end as well and we're looking into this. Is this currently a blocker for you?

solomonope:
I have investigated this.

[...]

PR #2138, opened 2020-11-14
fix(cert): remove namespace for certificate secrets

solomonope:
This issue was https://github.com/garden-io/garden/issues/2098 was reported,  the investigation revealed that it is misleading to specify a namespace because cert-manager would create the Certificate Secret in the same namespace as the secret.

This PR removes the namespace parameter for the Certificate secret and only request the name. An alternative solution would be to copy that secret to the specified namespace

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/config.ts
+++ b/core/src/plugins/kubernetes/config.ts
@@ -270,9 +270,10 @@ const tlsCertificateSchema = () =>
           "If you don't specify these, they will be automatically read from the certificate."
       )
       .example(["www.mydomain.com"]),
-    secretRef: secretRef
+    secretRef: joiIdentifier()
+      .required()
       .description("A reference to the Kubernetes secret that contains the TLS certificate and key for the domain.")
-      .example({ name: "my-tls-secret", namespace: "default" }),
+      .example("my-tls-secret"),
     managedBy: joi
       .string()
       .description(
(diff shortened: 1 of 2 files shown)


edvald:
I'm not quite sure about this one. This is effectively a breaking API change, even though it never worked as expected. Another option here would be to copy the secret to the specified namespace—do we see any potential issue with that approach?

solomonope:
> I'm not quite sure about this one. This is effectively a breaking API change, even though it never worked as expected. Another option here would be to copy the secret to the specified namespace—do we see any potential issue with that approach?

The potential issue with copying the certificate is when cert-manager updates/renews the certificate there is no process running to copy it over to the new namespace

edvald:
I see, that makes sense. In that case, I'd prefer to update the docs for the field instead of removing it, and emitting a log warning when it's being used, just stating that it's not currently supported by cert-manager.

solomonope:
> I see, that makes sense. In that case, I'd prefer to update the docs for the field instead of removing it, and emitting a log warning when it's being used, just stating that it's not currently supported by cert-manager.

Yes, I can take that approach, I would update the PR, thanks

vvagaytsev:
The cert-manager is deprecated now, and we're going to remove it in `0.13`. Closing this as "Won't do".

vvagaytsev:
GARDEN_LEGACY_FANCY_LOG_RENDER

PR #2147, opened 2020-11-20
improvement(core): experimental build stage implementation without rsync

edvald:
This implements the one-way build staging process entirely without
rsync. It's not trivial to get right, turns out, so for now this has
to be explicitly enabled by setting GARDEN_EXPERIMENTAL_BUILD_STAGE=true.

We'll keep it that way until the next breaking release. Users who have
issues with the rsync-based implementation should give this a try and
by all means file issues and provide feedback.


thsig (approved):
Looks great! Only had a couple of minor comment-related suggestions there.

ArnaudRinquin:
👋 

I tried this using the latest edge build: it's day and night in our monorepo. What used to take a hard 70+ seconds now takes 0.5seconds 😻 

Thanks you so much for this!

PR #2148, opened 2020-11-21
fix(core): auto-exclude git and .garden dirs for modules in project root

edvald:
Because we require an `include` or `exclude` to be set on modules if
there are multiple modules in the same directory, this would only have
come up  as an issue if there's a single module in the project root
_or_ if users had used `exclude` fields and not `include` fields to
specify the module context.

Which is likely why this passed by for so long, but certainly this
change is the more sane default behavior.


thsig (approved):

thsig:
Looks good to me—the remote sources tests under `GitHandler` are failing, though.

Issue #1608, opened 2020-02-15
Garden falsely identifies remote sources as an  overlapping module


eysi09:
### Current Behavior

If a project has a module at the project root, and uses remote sources, Garden thinks that the sources (under `./garden/sources/...`) overlap with the module at the root. See also this error message:

```
[2020-02-13T10:52:05.702Z] Error: Missing include and/or exclude directives on modules with overlapping paths.
Setting includes/excludes is required when modules have the same path (i.e. are in the same garden.yml file),
or when one module is nested within another.
Module organization overlaps with module(s) authorization-group (nested), postgres (nested) and redis (nested).
    at Garden.<anonymous> (/snapshot/project/garden-service/tmp/dist/build/src/garden.js:0)
    at Generator.next (<anonymous>)
    at fulfilled (/snapshot/project/garden-service/tmp/dist/build/src/garden.js:0)
Error Details:
overlappingModules:
  - module:
      name: organization
      path: /Users/balazsrostas/organization
    overlaps:
      - name: authorization-group
        path: >-
          /Users/balazsrostas/organization/.garden/sources/project/authorization-group--b3c5ccadd0
      - name: postgres
        path: >-
          /Users/balazsrostas/organization/.garden/sources/project/postgres-redis--f1c1fa90b5/postgres
      - name: redis
        path: >-
          /Users/balazsrostas/organization/.garden/sources/project/postgres-redis--f1c1fa90b5/redis
```

### Expected behavior

Garden should not consider remote sources when detecting overlapping modules.

### Reproducible example

N/A

### Workaround

Set `exclude: [.garden]` in the module config for the module at the root.

### Additional context

A user who ran into this has the following setup:

`A` (1 project, 1 module) repo uses `B` and `C` as remotes and has `.garden` in `.gitignore`
`B` (1 project, 1 module) uses `C` as remote and has `.garden` in `.gitignore`
`C` (two modules, no project) does not have `.garden` in `.gitignore`

PR #2151, opened 2020-11-25
fix(core): don't flag remote modules as overlapping with modules in root

edvald:
Came across this while debugging another issue. Simple fix.


eysi09:
Closes #1608.

thsig (approved):

PR #2158, opened 2020-12-01
fix(core): prohibit templates in workflow name

thsig:

We don't want to allow template strings in workflow names or trigger specs.

### DIFF
--- a/core/src/config/workflow.ts
+++ b/core/src/config/workflow.ts
@@ -6,7 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import { cloneDeep, isEqual, merge, omit, take } from "lodash"
+import { isEqual, merge, omit, take } from "lodash"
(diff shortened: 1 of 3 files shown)


core/src/config/workflow.ts
```diff
@@ -307,7 +307,18 @@ export function resolveWorkflowConfig(garden: Garden, config: WorkflowConfig) {
   const log = garden.log
   const context = new WorkflowConfigContext(garden)
   log.silly(`Resolving template strings for workflow ${config.name}`)
-  let resolvedConfig = resolveTemplateStrings(cloneDeep(config), context, { allowPartial: true })
+  // {
```
eysi09, line 310: Remove comment?

eysi09 (approved):

eysi09 (changes requested):
One comment, otherwise looks good.

core/src/config/workflow.ts
thsig: Done!

eysi09 (approved):

PR #2162, opened 2020-12-01
refactor: reduce work at module resolution time

edvald:
Just a bit of debt being cleaned up here, but it does have a small
impact on init time.

The primary change was in the plugin interface, which is now a callback
that is resolved later in the init process.


thsig (approved):
LGTM 👍

PR #2165, opened 2020-12-10
feat(k8s): use pod spec fields in tasks and tests

### LINKED ISSUE #2118: [FEATURE]: expose service account in tests
## Feature Request
Expose service account in tests instead of using default.

### Background / Motivation
Using the default service account is not the best practice and is rather limiting when testing complex stacks with differing permissions...

Allow setting a desired service account for tests (and possibly tasks).

Isolation of service accounts with varying permissions is critical. One such example is having a test (without mocks of course) that deploys a resource in other namespaces. It should be easy to run tests with any service account.

Expose service account when copying params for the test container.

https://github.com/garden-io/garden/blob/e2f0d8dff8c06f81502ae064f8a88183eeea164e/core/src/plugins/kubernetes/container/test.ts#L30

Somewhere between painful and critical.

We either add a RoleBinding to the default service account (yuck) or ignore garden tests and run "tests" inside of a kubernetes resource with an explicit service account (also yuck).

thsig:

For `helm` and `kubernetes` modules, we now use all pod spec fields from the service resource when running the task/test (excluding a fixed blacklist of pod spec fields).

This allows tasks/tests to e.g. use custom service accounts, node selectors etc.

Also refactored / broke up the `runAndCopy` function.

Fixes #2118.

**Special notes for reviewer**:

What do we think of the blacklist as it is here? Do we want to add/remove any fields?

### DIFF
--- a/core/test/data/test-projects/helm/artifacts/values.yaml
+++ b/core/test/data/test-projects/helm/artifacts/values.yaml
@@ -41,7 +41,8 @@ resources: {}
   #  cpu: 100m
   #  memory: 128Mi
 
-nodeSelector: {}
+# This field is whitelisted in `runPodSpecWhitelist`, so it should be included when running tests/tasks
+shareProcessNamespace: true
 
 tolerations: []
 
--- a/core/src/plugins/kubernetes/config.ts
+++ b/core/src/plugins/kubernetes/config.ts
@@ -27,6 +27,7 @@ import { baseTaskSpecSchema, BaseTaskSpec, cacheResultSchema } from "../../confi
 import { baseTestSpecSchema, BaseTestSpec } from "../../config/test"
 import { ArtifactSpec } from "../../config/validation"
 import { V1Toleration } from "@kubernetes/client-node"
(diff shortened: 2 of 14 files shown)


core/test/data/test-projects/helm/artifacts/values.yaml
```diff
@@ -41,7 +41,7 @@ resources: {}
   #  cpu: 100m
   #  memory: 128Mi
 
-nodeSelector: {}
+nodeSelector: { diskType: ssd }
```
edvald, line 44: This may be why the tests are failing. The Pod cannot be scheduled.

edvald:
This all seems to make sense @thsig, but there is that test timing out. I had suspicion as to why, see comment above.

thsig:
This should be good to go now.

I've replaced the blacklist with a whitelist of pod spec fields, and it also looks like the tests are now fine (since I used a different field than `nodeSelector` to test the pod spec logic).

edvald (approved):

PR #2166, opened 2020-12-10
Add imageModule field on container modules

edvald:


vvagaytsev:
This was implemented in some past PRs.

PR #2167, opened 2020-12-10
feat(enterprise): implement auth token refresh

10ko:
This PR implements the GE auth token refresh mechanism in core.
It follow roughly the same logic as the GE frontend except it stores the token(s) in the the local db instead of cookies.
Refactoring has been carried on to make sure the only entry point for calling GE is a new class called `EnterpriseApi`.
Once `EnterpriseApi.init()` is called an interval is started and the token is refreshed at regular intervals. 

Other notable changes include:
- We now assume the GE domain is the `frontend` domain and not the api anymore.
- Once a cli login has succeded we redirect to a new url: currently it doesn't work because it still needs to be merged in the GE repo.

Fixes #

Some tests are failing and I would love to pair on figuring out why.


core/src/enterprise/api.ts
```diff
@@ -0,0 +1,287 @@
...
+          refreshToken: rt.value || "",
+          tokenValidity: res.body.data.jwtValidity,
+        }
+        await this.saveAuthToken(tokenObj)
+      } catch (err) {
+        const res = err.response
+
+        if (res && res.statusCode === 401) {
+          this.log.debug({ msg: `Failed to refresh the token.` })
+          // await clearAuthToken(this.log)
```
thsig, line 127: Should this be commented out?

core/src/enterprise/api.ts
```diff
@@ -0,0 +1,287 @@
...
+        } else {
+          throw new RuntimeError(
+            `Generic = An error occurred while verifying client auth token with platform: ${err.message}`,
```
thsig, line 130: Remove "Generic ="?

core/src/enterprise/init.ts
```diff
@@ -28,17 +27,17 @@ export interface GardenEnterpriseContext {
...
   if (clientAuthToken) {
-    if (!enterpriseDomain || !projectId) {
+    if (!enterpriseApi.getDomain() || !projectId) {
       const errorMessages: string[] = []
-      if (!enterpriseDomain) {
+      if (!enterpriseApi.getDomain()) {
         errorMessages.push(deline`
           ${chalk.bold("project.domain")} is not set in your project-level ${chalk.bold("garden.yml")}. Make sure it
-          is set to the appropriate API backend endpoint (e.g. http://myusername-cloud-api.cloud.dev.garden.io,
+          is set to the appropriate Garden Enterprise url (e.g. https://garden-enterprise.example.com,
```
thsig, line 40: "url" -> "URL"

core/src/enterprise/workflow-lifecycle.ts
```diff
@@ -43,20 +40,26 @@ export async function registerWorkflowRun({
...
+      throw new EnterpriseApiError(
+        `Error while registering workflow run: Request failed with status ${res["status"]}`,
+        {
+          status: res["status"],
+          workflowRunUid: res["workflowRunUid"],
+        }
+      )
+    }
   }
+  throw new EnterpriseApiError("Error while registering workflow run: Couldn't initialize api.", {})
```
thsig, line 64: "api" -> "API"

thsig (commented):
Looks good! Had a few comments there.

core/src/enterprise/api.ts
10ko: Good catch, I think I commented it out while trying to debug something.
I am now also throwing a `RuntimeError` since there is nothing else to do at that point and the user's required to log back in.

10ko:
@thsig thanks for the review, I addressed all your comments! ✌️

edvald:
@10ko can you rebase this one?

thsig (approved):

PR #2169, opened 2020-12-11
improvement(dashboard): better Stack Graph layout

edvald:
This replaces the previous dagre-based layout with Reaflow, which uses
ELK for the algorithm. This makes the layout stable for every run,
and uses a layered horizontal layout which scales much better with a
large number of nodes in the graph.

Reaflow also makes the code better fitted to React, so this should be
easier to work with as we do further iterations and improvements.

Please test this manually on different projects, poke at it, and see if you find any subtle issues.


eysi09 (changes requested):
Thanks for your work on this, this is a great improvement!

As future work though, I'd love to be able to use proper DOM elements instead of just SVGs to really style this thing. 

See e.g. the new GitHub Actions workflow graph:

GitHub Actions Graph

A lot of nice ideas there. (Here you can see it in action: https://github.com/garden-io/garden/actions/runs/423337266)

But anyway, back to this PR. I noticed a handful of UI bugs, see below.

1. Stack Graph does not zoom out to fit page, rather scrollbars are added. (It doesn't zoom fully out for very large graphs on current master either, but at least there are no scrollbars and it's straightforward to zoom further out. 
2. Disabled modules don't look disabled (see e.g. the version-check module). They also kind of look like they're a part of the main graph, not sure if it's possible to configure that somehow.

Screenshot 2020-12-22 at 09 42 17

Compare the two points above to current master:

Screenshot 2020-12-22 at 09 23 46

3. The spinning "active" border is not visible.

Screenshot 2020-12-22 at 09 10 17

4. The legend overflows (also the title of my autobiography)

Screenshot 2020-12-22 at 09 09 09

5. The x-scrollbar becomes visible after opening the side menu.

Screenshot 2020-12-22 at 09 29 08

edvald:
I'll take a look at these. Many of these I don't see myself (e.g. scrollbars and missing active border), so I wonder what the difference might be. I'll try some different scenarios, one that occurs to me is that a non-default zoom level could be causing it, any chance that was the case? Should still fix it ofc, just wondering where that comes from.

Regarding the behavior when the sidebar opens, I sorta prefer the graph not re-arranging and fitting the resized viewport. You feel strongly that it should?

As for using pure DOM elements, I mean yea sure, that's just pretty far outside of my abilities atm, would mean not using D3 or other visual libraries etc.

eysi09:
> one that occurs to me is that a non-default zoom level could be causing it, any chance that was the case?

I had a normal zoom level. On my laptop screen the scrollbars are still there but only become visible when I start scrolling.

> Regarding the behavior when the sidebar opens, I sorta prefer the graph not re-arranging and fitting the resized viewport. You feel strongly that it should?

No, I don't think it should rearrange, but figured the user should either be able to zoom out or drag the graph around, as opposed to introducing the scrollbars.

edvald:
@eysi09 can you give this another spin? I've fixed the "in progress" animation (turned out to be an SCSS syntax error I added just before pushing), and hidden the scrollbars.

edvald:
Nudge @eysi09

eysi09:
@edvald any update on that pan situation?

edvald:
Yea did you try it again today?

eysi09:
Yup, LGTM! Great job!

eysi09 (approved):

PR #2178, opened 2021-01-04
improvement(enterprise): include message metadata with log entries

eysi09:

The log entries we were sending over didn't include all the message metadata to properly render them in GE. We were also including previous message states with the message sent which we shouldn't do. 

Fixes #

I kept the previous fields in the payload to maintain compatibility. We need to discuss how we do this moving forward. 

I also renamed some fields for clarity.

### DIFF
--- a/core/src/logger/log-entry.ts
+++ b/core/src/logger/log-entry.ts
@@ -53,8 +53,8 @@ interface MessageBase {
   maxSectionWidth?: number
 }
 
-export interface MessageState extends MessageBase {
-  timestamp: number
+export interface LogEntryMessage extends MessageBase {
+  timestamp: Date
 }
 
 export interface UpdateLogEntryParams extends MessageBase {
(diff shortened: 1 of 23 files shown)


core/src/logger/log-entry.ts
```diff
@@ -210,23 +208,23 @@ export class LogEntry extends LogNode {
...
-    return this.messageStates
+  getMessages() {
+    return this.messages
   }
 
   /**
-   * Returns a deep copy of the latest message state, if availble.
-   * Otherwise return an empty object of type MessageState for convenience.
+   * Returns a deep copy of the latest message, if availble.
+   * Otherwise return an empty object of type LogEntryMessage for convenience.
```
thsig, line 217: _Otherwise return_ -> _Otherwise returns_

eysi09: Fixed.

thsig (approved):

thsig (approved):

PR #2185, opened 2021-01-06
chore: fix disabled-configs example project

thsig:

This example project was broken when using the `remote` environment (since the `defaultNamespace` was set at the provider rather than at the environment level).

We usually run this project using the `local` environment, so this only surfaced today.


edvald:
Hmm. This should have come up as a configuration error, right?

edvald:
Ah I guess the validation error would only surface when picking that environment, nm.

eysi09 (approved):

PR #2195, opened 2021-01-15
fix(enterprise): lower workflow resource minima

thsig:

The minimum resource request values for workflows were too high. We lower them to more sensible levels here.


core/src/config/workflow.ts
```diff
@@ -81,12 +81,12 @@ export const workflowConfigSchema = () =>
           cpu: joi
             .number()
             .default(defaultContainerLimits.cpu)
-            .min(1000)
+            .min(500)
```
eysi09, line 84: Why not just 100?

eysi09 (commented):
Had one question.

core/src/config/workflow.ts
thsig: I think Kubernetes' minimum is 200, but we could just go for that.

My thinking was that we should have a floor that ensures that Garden has enough resources to run reasonably fast and give a nice experience.

eysi09 (approved):

PR #2206, opened 2021-01-18
improve(config): allow most template strings in templated module inputs

edvald:
Basically we now defer the full resolution on input template strings,
with the caveat that input values needed for module identifiers need
to be resolvable immediately.

Previously the values passed to the `inputs` field on templated modules
only had a very limited templating context available.

This involved quite a bit of added complexity, which is prompting me to
seriously consider a refactor in our template string resolution logic. This 
does the job for now, but I think we'll soon want to implement at least
partial lazy+recursive resolution of template strings, instead of all
this fine-grained logic to control what's resolved when.

### DIFF
--- a/core/src/config/config-context.ts
+++ b/core/src/config/config-context.ts
@@ -225,6 +225,20 @@ export abstract class ConfigContext {
   }
 }
 
+/**
+ * A generic context that just wraps an object.
+ */
+export class GenericContext extends ConfigContext {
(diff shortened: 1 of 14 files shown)


thsig (changes requested):
Looks good! Had a few text-related comments there.

edvald:
👍 all comments addressed

thsig (approved):

PR #2211, opened 2021-01-19
fix(enterprise): fix start interval when using access tokens

10ko:
This PR fixes a problem we introduced recently where we would start the auth token refresh interval when using an Access Token. 
Additionally, I refactored the `logout` function, replacing old unused code with the actual implementation that was in `EnterpriseApi`

Fixes #

### DIFF
--- a/core/src/enterprise/api.ts
+++ b/core/src/enterprise/api.ts
@@ -81,11 +81,10 @@ export class EnterpriseApi {
       if (!tokenIsValid) {
         // If the token is an Access Token and it's invalid we return.
         if (gardenEnv.GARDEN_AUTH_TOKEN) {
-          this.log.error({
-            msg:
-              "The provided access token is expired or revoked, please create a new one from the Garden Enterprise UI.",
-          })
-          return
+          throw new RuntimeError(
+            "The provided access token is expired or has been revoked, please create a new one from the Garden Enterprise UI.",
+            {}
+          )
         } else {
           // Try to refresh an expired JWT
           // This will throw if it fails to refresh
--- a/core/src/enterprise/api.ts
+++ b/core/src/enterprise/api.ts
@@ -70,19 +70,38 @@ export class EnterpriseApi {
       return
     }
     this.enterpriseDomain = enterpriseDomain
+
+    // Retrieve an authentication token
     const authToken = await this.readAuthToken()
     if (authToken && commandAllowed) {
-      this.log.debug({ msg: `Refreshing auth token and starting refresh interval.` })
+      // Verify a valid token is present
+      this.log.debug({ msg: `Refreshing auth token and trying to start refresh interval.` })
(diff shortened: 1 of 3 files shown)


core/src/enterprise/api.ts
```diff
@@ -70,27 +70,47 @@ export class EnterpriseApi {
...
     if (authToken && commandAllowed) {
-      this.log.debug({ msg: `Refreshing auth token and starting refresh interval.` })
+      // Verify a valid token is present
+      this.log.debug({ msg: `Refreshing auth token and trying to start refresh interval.` })
       const tokenIsValid = await this.checkClientAuthToken(this.log)
+
       if (!tokenIsValid) {
-        await this.refreshToken()
+        // If the token is an Access Token and it's invalid we return.
+        if (gardenEnv.GARDEN_AUTH_TOKEN) {
```
eysi09, line 83: Shouldn't this throw? If you're explicitly setting an access token that doesn't work you probably want an error.

core/src/enterprise/api.ts
```diff
@@ -70,27 +70,47 @@ export class EnterpriseApi {
...
+          this.log.error({
+            msg:
+              "The provided access token is expired or revoked, please create a new one from the Garden Enterprise UI.",
```
eysi09, line 86: is expired or revoked, > is expired or has been revoked,

core/src/enterprise/auth.ts
```diff
@@ -55,13 +57,22 @@ export async function login(enterpriseApi: EnterpriseApi, log: LogEntry): Promis
 }
 
 export async function logout(enterpriseApi: EnterpriseApi, log: LogEntry): Promise<void> {
```
eysi09, line 59: This function isn't used anywhere.

core/src/enterprise/api.ts
```diff
@@ -207,26 +229,6 @@ export class EnterpriseApi {
     return token
   }
 
-  async logout() {
```
eysi09, line 210: Why are we removing this? Note that the "other" logout function isn't used.

eysi09 (changes requested):

core/src/enterprise/api.ts
10ko: I moved the logic to be out of the `EnterpriseApi` class to maintain consistency with the `login` function

core/src/enterprise/auth.ts
10ko: I updated the PR, I missed the `LogOutCommand`

core/src/enterprise/api.ts
10ko: Possibly. I wanted to try a more conservative approach since it fails later, but I agree we can throw here.
I'll add it.

10ko:
Thanks @eysi09, I addressed all your comments.

eysi09 (approved):

PR #2216, opened 2021-01-19
feat(k8s): add cluster-buildkit buildMode

edvald:
This adds a new `cluster-buildkit` buildMode to the kubernetes provider,
which uses the [buildkit](https://github.com/moby/buildkit) project.

It also works differently to the existing `cluster-docker` and `kaniko`
build modes, in that it doesn't require any cluster-wide components,
such as the ever-frustrating NFS provisioner.

When configured, a `garden-buildkit` Deployment with an ephemeral
volume is deployed on-demand to _the project namespace_. This greatly
simplifies administration and operation, while still offering a
performance benefit compared to `kaniko`, and it still uses the
deployment registry (in-cluster or otherwise) as a shared cache between
users/namespaces.

It can also be configured to run in "rootless" mode, i.e. run without
elevated privileges. This has some caveats (see the buildkit docs for
details) but offers a much improved security profile compared to the
privileged Docker and Kaniko builders.


thsig (approved):
Looks good! Had one comment there (which is outside the scope of this PR anyway).

PR #2232, opened 2021-01-29
test(workflows): add workflow smoke-test in pre-release script

10ko:
Add a sanity check for the `run workflow` command in order to avoid future regressions.

Fixes #

### DIFF
--- a/examples/vote/README.md
+++ b/examples/vote/README.md
@@ -49,3 +49,15 @@ garden dev --hot=vote
 Then try making a change to one of the source files in the `vote` service, to see it synchronize into the
 running container, instead of the normal build+deploy flow. Note that changing the file will _also_ trigger a
 build and some tests, but the hot-reloading should complete almost instantly while those take longer to complete.
+
+### Try out Workflows
+
+This example includes a usage example for Garden workflows. The `workflows.garden.yml` file contains the configuration for a sample workflow called `full-test` which initializes the `db`, runs the tests and cleans up the database afterwards. An example of how to execute custom scripts is also included.
+
+To run the workflow:
+
+```sh
+garden run workflow full-test
+```
+
+For more complex use-cases and additional configuration options please refer to the [docs](https://docs.garden.io/using-garden/workflows).
--- a/examples/vote/workflows.garden.yml
+++ b/examples/vote/workflows.garden.yml
@@ -1,0 +1,7 @@
+kind: Workflow
+name: full-test
+description: "An example workflow showing how to run garden commands and scripts"
+steps:
+  - command: [test]
+  - command: [run, task, db-clear]
+  - script: echo "Greetings ${local.username}, your full-test script run succesfully."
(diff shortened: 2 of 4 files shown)


edvald:
Ugh, I think GitBook messed up our docs folder (again). I can fix that.

10ko:
Uh, I was wondering what was the issue. Thanks @edvald

thsig:
Looks good to me, do you have any comments @edvald?

The main point is to test the "outer" part, the CLI/command init etc. (similarly to how we're testing the other commands here), and for that, I think the test case is sufficient as written.

The unit tests should cover the details of the command's execution pretty well.

edvald:
It occurs to me that we should do this in a specific testing project, outside of our example projects. Either that or make the workflow somehow more organically included in the example, it's kinda just lumped in there without a purpose or further explanation at the moment, which we could address in the example README or comments in the added workflow config.

10ko:
Makes sense, I will make some changes to make it look more "useful" under that example and add a bit of explanation in the README of the project.
The reason is there is because we use this project in demos and it would have been handy to have an example of a workflow.

thsig (dismissed):

10ko:
Added few lines to the README of the example, made the workflow a "full-test" and fixed the `clear-db` task which was broken.

examples/vote/workflows.garden.yml
```diff
@@ -0,0 +1,8 @@
+kind: Workflow
+name: full-test
+description: "An example workflow showing how to run garden commands and scripts"
+steps:
+  - command: [run, task, db-init]
```
edvald, line 5: This task is implicitly run in the test command, so that seems sorta superfluous. But it makes sense to clear the db after.

examples/vote/workflows.garden.yml
```diff
@@ -0,0 +1,8 @@
...
+  - command: [test]
+  - command: [run, task, db-clear]
+  - script: echo "Greetings $USER, your full-test script run succesfully."
```
edvald, line 8: Could use `${local.username}` just to illustrate the templating, just a thought.

examples/vote/README.md
```diff
@@ -49,3 +49,15 @@ garden dev --hot=vote
 Then try making a change to one of the source files in the `vote` service, to see it synchronize into the
 running container, instead of the normal build+deploy flow. Note that changing the file will _also_ trigger a
 build and some tests, but the hot-reloading should complete almost instantly while those take longer to complete.
+
+### Try out Garden Workflows
+
+This example includes a usage example for Garden workflows. The `workflows.garden.yml` file contains the configuration for a sample workflow called `full-test` which initilizes the `db`, runs the tests and cleanups at the end. An example of how to execute custom scripts is also included.
```
edvald, line 55: "initilizes" -> "initializes"
"cleanups at the end" -> "and cleans up the database afterwards"

edvald (changes requested):
Super minor stuff, mostly just the db-init step that felt superfluous.

10ko:
Thanks @edvald, addressed all the last requests.

edvald (approved):

PR #2233, opened 2021-01-29
Revert "GitBook: [master] 130 pages and 21 assets modified"

edvald:
This reverts commit 05a069a0ec55d135441a6df9ed032772ab4cad5e.


edvald:
FYI @wints, we run into problems if we let GitBook update our repo directly. We may need to make whatever changes you were making differently.

PR #2239, opened 2021-02-02
feat(templates): allow concatenating arrays with + operator

edvald:
The example from the docs:

```yaml
kind: Project
# ...
variables:
  some-values: ["a", "b"]
  other-values: ["c", "d"]
---
kind: Module
type: helm
# ...
values:
  some-array: ${var.some-values + var.other-values}
  ...
```

### DIFF
--- a/docs/using-garden/variables-and-templating.md
+++ b/docs/using-garden/variables-and-templating.md
@@ -46,8 +46,9 @@ You can use a variety of operators in template string expressions:
 * Logical: `&&`, `||`, ternary (`<test> ? <value if true> : <value if false>`)
 * Unary: `!` (negation), `typeof` (returns the type of the following value as a string, e.g. `"boolean"` or `"number"`)
 * Relational: `contains` (to see if an array contains a value, an object contains a key, or a string contains a substring)
+* Arrays: `+`
 
-The arithmetic and numeric comparison operators can only be used for numeric literals and keys that resolve to numbers. The equality and logical operators work with any term.
+The arithmetic and numeric comparison operators can only be used for numeric literals and keys that resolve to numbers, except the `+` operator which can be used to concatenate two array references. The equality and logical operators work with any term (but be warned that arrays and complex objects aren't currently compared in-depth).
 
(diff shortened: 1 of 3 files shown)


docs/using-garden/variables-and-templating.md
```diff
@@ -46,8 +46,9 @@ You can use a variety of operators in template string expressions:
 * Logical: `&&`, `||`, ternary (`<test> ? <value if true> : <value if false>`)
 * Unary: `!` (negation), `typeof` (returns the type of the following value as a string, e.g. `"boolean"` or `"number"`)
 * Relational: `contains` (to see if an array contains a value, an object contains a key, or a string contains a substring)
+* Arrays: `+`
 
-The arithmetic and numeric comparison operators can only be used for numeric literals and keys that resolve to numbers. The equality and logical operators work with any term.
+The arithmetic and numeric comparison operators can only be used for numeric literals and keys that resolve to numbers, except the `+` operator which can be used to join two array references. The equality and logical operators work with any term.
```
thsig, line 51: "any term" -> "any atomic term (numbers, strings and booleans)"

... since we're not implementing an element-wise array equality comparison here.

docs/using-garden/variables-and-templating.md
```diff
@@ -108,6 +109,32 @@ services:
...
+```yaml
+kind: Module
+type: container
+...
+services:
+  replicas: ${var.default-replicas + 1}
+  ...
+```
+
+And the `+` operator can also be used to concatenate two array:
```
thsig, line 121: "array" -> "arrays"

thsig (changes requested):
Looks good! Had a couple of doc-related comments there.

docs/using-garden/variables-and-templating.md
edvald: Well, we do still _allow_ those for any term, so I'm not sure we want to change this phrasing.

edvald:
@thsig updated and should be good to go

PR #2250, opened 2021-02-09
Publish command fixes/improvements

edvald:
It was time to give the publish command a bit of ❤️. Couple of commits:

**fix(k8s): garden publish command now works with any deploymentRegistry**
We had been making some assumptions that the in-cluster registry was
being used. We already had the code to pull images from other
registries, so this wasn't actually to hard to address. Also added
tests to validate.

**feat(publish): add template-able --tag parameter to publish command**

This should generally make the `garden publish` command a lot more
usable. You can now override the default tag at publish-time, and
even apply template strings on the tags.

From the updated in-cluster building docs:

-------

You can also set the `--tag` option on the `garden publish` command to override the tag used for images. You can both set a specific tag or you can _use template strings for the tag_. For example, you can

- Set a specific tag on all published modules: `garden publish --tag "v1.2.3"`
- Set a custom prefix on tags but include the Garden version hash: `garden publish --tag "v0.1-${module.hash}"`
- Set a custom prefix on tags with the current git branch: `garden publish --tag "v0.1-${git.branch}"`

Generally, you can use any template strings available for module configs for the tags, with the addition of the following:

- `${module.name}` — the name of module being tagged
- `${module.version}` — the full Garden version of module being tagged, e.g. `v-abcdef1234`
- `${module.hash}` — the Garden version hash of module being tagged, e.g. `abcdef1234` (i.e. without the `v-` prefix)

-------

When reviewing, please give the publish command a spin, and try messing about with the `--tag` parameter.

### DIFF
--- a/core/src/plugins/container/publish.ts
+++ b/core/src/plugins/container/publish.ts
@@ -10,14 +10,14 @@ import { ContainerModule } from "./config"
 import { PublishModuleParams } from "../../types/plugin/module/publishModule"
 import { containerHelpers } from "./helpers"
 
-export async function publishContainerModule({ ctx, module, log }: PublishModuleParams<ContainerModule>) {
+export async function publishContainerModule({ ctx, module, log, tag }: PublishModuleParams<ContainerModule>) {
   if (!containerHelpers.hasDockerfile(module, module.version)) {
     log.setState({ msg: `Nothing to publish` })
     return { published: false }
   }
 
   const localId = containerHelpers.getLocalImageId(module, module.version)
-  const remoteId = containerHelpers.getPublicImageId(module)
+  const remoteId = containerHelpers.getPublicImageId(module, tag)
 
   log.setState({ msg: `Publishing image ${remoteId}...` })
 
--- a/docs/guides/in-cluster-building.md
+++ b/docs/guides/in-cluster-building.md
@@ -314,19 +314,34 @@ providers:
 
 You can publish images that have been built in your cluster, using the `garden publish` command.
 
-The only caveat is that you currently need to have Docker running locally, and you need to have authenticated with the
-target registry. When publishing, we pull the image from the in-cluster registry to the local Docker daemon, and then
-go on to push it from there. We do this to avoid having to (re-)implement all the various authentication methods (and
-by extension key management) involved in pushing directly from the cluster.
+The only caveat is that you currently need to have Docker running locally, and you need to have authenticated with the target registry. When publishing, we pull the image from the remote registry to the local Docker daemon, and then go on to push it from there. We do this to avoid having to (re-)implement all the various authentication methods (and by extension key management) involved in pushing directly from the cluster, and because it's often not desired to give clusters access to directly push to production registries.
 
-As usual, you need to specify the `image` field on the `container` module in question. For example:
+Unless you're publishing to your configured deployment registry, you need to specify the `image` field on the `container` module in question, to indicate where the image should be published. For example:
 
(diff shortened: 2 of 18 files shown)


docs/guides/in-cluster-building.md
```diff
@@ -314,20 +314,31 @@ providers:
...
-target registry. When publishing, we pull the image from the in-cluster registry to the local Docker daemon, and then
-go on to push it from there. We do this to avoid having to (re-)implement all the various authentication methods (and
-by extension key management) involved in pushing directly from the cluster.
+The only caveat is that you currently need to have Docker running locally, and you need to have authenticated with the target registry. When publishing, we pull the image from the remote registry to the local Docker daemon, and then go on to push it from there. We do this to avoid having to (re-)implement all the various authentication methods (and by extension key management) involved in pushing directly from the cluster, and because it's often not desired to give clusters access to directly push to production registries.
 
-As usual, you need to specify the `image` field on the `container` module in question. For example:
+You'll generally need to specify the `image` field on the `container` module in question, to indicate where the image should be published. For example:
```
eysi09, line 319: Why _generally_? Doesn't this always apply? If not, can we mention when it doesn't.

core/src/plugins/container/publish.ts
```diff
@@ -10,14 +10,14 @@ import { ContainerModule } from "./config"
...
   if (!containerHelpers.hasDockerfile(module, module.version)) {
     log.setState({ msg: `Nothing to publish` })
     return { published: false }
   }
 
   const localId = containerHelpers.getLocalImageId(module, module.version)
-  const remoteId = containerHelpers.getPublicImageId(module)
+  const remoteId = containerHelpers.getPublicImageId(module, tag)
 
   log.setState({ msg: `Publishing image ${remoteId}...` })
```
eysi09, line 22: Perhaps out of scope for this one but would be nice to stream the logs from the publish command when at `verbose` log level.

docs/guides/in-cluster-building.md
```diff
@@ -314,20 +314,31 @@ providers:
...
+image: my-repo/my-image:v1.2.3   # <- if you omit the tag here, the Garden module version will be used by default
 ...
 ```
 
+By default we use the tag specified in the `container` module `image` field, if any. If none is set there, we default to the Garden module version.
+
+You can also set the `--tag` option on the `garden publish` command to override the tag used for images. You can both set a specific tag or you can _use template strings for the tag_. For example, you can
+
+- Set a specific tag on all published modules: `garden publish --tag "v1.2.3"`
+- Set a custom prefix on tags but include the Garden version hash: `garden publish --tag "v0.1-${module.hash}"`
```
eysi09, line 333: I need to escape the `$` for this to work, otherwise I get a `zsh: bad substitution` error. We should probably mention that here.

eysi09 (commented):
Had a couple of comments but otherwise looks great!

docs/guides/in-cluster-building.md
edvald: You don't need that if you're publishing to your deployment registry, for example.

edvald: But I'll clarify that in the docs.

docs/guides/in-cluster-building.md
edvald: I updated the example to use single-quotes, which solves that matter, and documented that as well.

edvald:
Updated @eysi09

eysi09 (approved):

PR #2254, opened 2021-02-17
feat(core): lower case username for templates

thsig:

In addition to the `local.username` template variable, we now provide `local.usernameLowerCase`.

This is useful e.g. when templating in the local username in namespaces or hostnames.

### DIFF
--- a/core/src/config/config-context.ts
+++ b/core/src/config/config-context.ts
@@ -289,10 +289,23 @@ class LocalContext extends ConfigContext {
   @schema(
     joi
       .string()
-      .description("The current username (as resolved by https://github.com/sindresorhus/username)")
+      .description("The current username (as resolved by https://github.com/sindresorhus/username).")
       .example("tenzing_norgay")
   )
   public username?: string
+
+  @schema(
+    joi
+      .string()
+      .description(
+        deline`
+          The current username (as resolved by https://github.com/sindresorhus/username), with any upper case
+          characters converted to lower case.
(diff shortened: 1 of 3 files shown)


edvald:
I'm trying to decide if we consider this a breaking change. It _technically_ is, but it's not particularly likely that this is being used in a manner that expects the prior behavior. @eysi09 what do you think?

Chipcius:
If it is a breaking change, adding a new variable that is lowercase or some modifier that can turn it lowercase would also solve the problem with using the username in namespaces

edvald:
That's actually a good idea. I'd be in favor of adding a `${local.username-lowercase}` or something like that. We have designs on functions/methods on template strings (i.e. modifiers) but that'll take a bit more doing.

thsig:
Good idea—I'll add the lowercase username as a separate variable as you suggest.

thsig:
I've updated the PR—instead of lowercasing `local.username`, I added ` local.usernameLowerCase`.

Does that sound like a decent name for this field?

edvald (approved):
Looks good, aside from that linting error

edvald:
Also the commit message is now a little odd :)

edvald:
I'll separately work out the failing tests, seems unrelated

thsig:
I've fixed the linting error. Having a hard time coming up with a non-awkward commit message for this one, but I did my best. Haha.

PR #2257, opened 2021-02-21
feat(template): allow escaping template strings for generated files

edvald:
This addresses an issue where template strings couldn't easily be
escaped, such that Garden _wouldn't_ attempt to resolve them.

From the docs:

Sometimes you may want to pass template strings through when generating
files, instead of having Garden resolve them. This could for example be
handy when templating a Terraform configuration file which uses a
similar templating syntax.

To do this, simply add an additional `$` in front of the template
string, e.g. `$${var.dont-resolve-me}`.

### DIFF
--- a/core/src/config/config-context.ts
+++ b/core/src/config/config-context.ts
@@ -43,6 +43,8 @@ export interface ContextResolveOpts {
   allowPartial?: boolean
   // a list of previously resolved paths, used to detect circular references
   stack?: string[]
+  // Unescape escaped template strings
+  unescape?: boolean
 }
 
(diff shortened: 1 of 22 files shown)


thsig (approved):

thsig:
Looks like we need to re-generate the docs here (probably because of the merge of the namespace annotations stuff).

Should we rebase this PR's commit on `master` instead and regenerate the docs there, so that the updated docs get folded into that commit? @edvald

edvald:
Yeah this needs to be rebased. I'm afk for a little bit, can do that after lunch

PR #2259, opened 2021-02-22
feat(config): add ${command.name} and ${command.params} for templating

edvald:
This allows users to template by the current command context in some
interesting ways. A simple example would be templating based on whether
the `test` command is being run, things like that.

A more complex example would be something like
`${command.params contains 'hot-reload' && command.params.hot-reload contains 'my-service'}`,
basically checking for a specific parameter.

Later I'd like to make this sort of thing more elegant in terms of syntax (maybe throwing in optional chaining), but it does already work nicely if used with a bit of care.

The first commit is just a refactor to split up the `config-context.ts` module. Probably best to review commits separately.


thsig (approved):
Looks great! The refactors are very welcome too.

We should probably rebase this one on `master`, regenerate the docs and force-push (like we did with that other branch earlier, since there are doc generation changes on `master`).

PR #2264, opened 2021-02-22
feat(config): allow sparse arrays where appropriate in config schemas 

edvald:
Users can now place undefined or null values in place of real values in many array configuration values, in order to conditionally place values there.

For example, you might do something like `dependencies: ["service-a", ${var.some-condition ? 'service-b' : null}]` to conditionally include a service dependency.

Generally, this applies for any array schema that wouldn't naturally expect null values (which would make the intent ambiguous).

### DIFF
--- a/core/src/config/common.ts
+++ b/core/src/config/common.ts
@@ -119,6 +119,7 @@ export interface Schema extends Joi.Root {
   gitUrl: () => GitUrlSchema
   posixPath: () => PosixPathSchema
   hostname: () => Joi.StringSchema
+  sparseArray: () => Joi.ArraySchema
 }
 
 export let joi: Schema = Joi.extend({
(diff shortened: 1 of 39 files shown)


core/src/config/common.ts
```diff
@@ -393,6 +394,19 @@ joi = joi.extend({
   },
 })
 
+/**
+ * Add a joi.sparseArray() type, that both allows sparse arrays _and_ filters the falsy values out.
```
thsig, line 398: Some of the new test cases also mention falsy values, but the `coerce` method doesn't filter out empty strings.

Was the intention to not filter out empty strings, or should we change the corresponding test descriptions?

edvald: Right, I originally figured we'd filter out empty strings, but decided against it because it doesn't serve a particular purpose and the intent with an empty string is more likely to be ambiguous. I'll update the cases and copy where needed.

edvald: I've updated the test description, should be all set.

thsig (approved):

thsig:
There's a failing unit test case there:
```text
1) exec plugin
     should correctly parse exec modules:
    AssertionError: expected [ Array(1) ] to deeply equal [ Array(1) ]
    + expected - actual
         "dependencies": []
         "disabled": false
         "name": "unit"
         "spec": {
    -      "artifacts": []
           "command": [
             "echo"
             "OK"
           ]
    
    at /home/circleci/project/core/test/unit/src/plugins/exec.ts:160:36
    at /home/circleci/project/core/build/test/unit/src/plugins/exec.js:15:71
    at new Promise (<anonymous>)
    at __awaiter (build/test/unit/src/plugins/exec.js:11:12)
    at Context.<anonymous> (test/unit/src/plugins/exec.ts:85:56)
```
Should the test case be fixed, or did this catch a regression?

PR #2272, opened 2021-02-24
chore(enterprise): update supported trigger events

thsig:


core/src/config/workflow.ts
```diff
@@ -226,20 +226,13 @@ export const workflowStepSchema = () => {
 export type workflowStepModifier = "onSuccess" | "onError" | "always" | "never"
 
 export const triggerEvents = [
-  "create",
-  "push",
   "pull-request",
-  "pull-request-created",
-  "pull-request-updated",
   "pull-request-opened",
```
edvald, line 230: What is the difference between "pull-request" and "pull-request-opened"?

thsig: I'll add an explanation to the docstring.

"pull-request" is a shorthand for "pull-request-created" or "pull-request-updated" or "pull-request-opened".

thsig: "pull-request" is a shorthand for "pull-request-created" or "pull-request-updated" or "pull-request-opened".

This is documented in the enterprise docs, but we should probably include it here as well.

eysi09 (approved):

PR #2278, opened 2021-02-26
fix(enterprise): handle requests that fail before reaching the server

eysi09:

We were assuming that all Got request errors had a response field, which is [not the case](https://github.com/sindresorhus/got#errors).

Fixes #


core/src/enterprise/secrets/index.ts
```diff
@@ -1,42 +0,0 @@
-/*
```
edvald, line 1: Was this just left over from a refactoring?

edvald: Nm, saw the note in the PR description!

core/src/enterprise/api.ts
```diff
@@ -289,10 +291,7 @@ export class EnterpriseApi {
       await this.get(log, "token/verify")
       valid = true
     } catch (err) {
-      const res = err.response
-      if (res.statusCode === 401) {
-        valid = false
-      } else {
+      if (!is401Error(err)) {
```
10ko, line 294: Why aren't we setting `valid = false` anymore?

10ko (commented):
Got a quick question, otherswise looks good.

core/src/enterprise/api.ts
eysi09: It's set to `false` at the top of the function.

eysi09:
@10ko , any further comments on this one?

thsig (approved):

PR #2281, opened 2021-02-27
improvement(core): more granular version hashes

edvald:
This change addresses a long-standing issue where changes to runtime
configuration (e.g. services, test and tasks) would also impact build
versions of container images.

We now tackle versioning more specifically for each entity within any
given module. Module types can control which parts of a configuration
should affect the build version, which is applied in different ways by
different module type handlers.

This change does mean that service, task and test versions are _always_
different from module (i.e. build) versions. This is because the
versions are now computed with the service/test/task configuration
hashed on top of the module configuration, even if there is no natural
distinction between the build version and a runtime version for that
particular module.

This generally has no practical downsides except that users might be
expecting versions to match, which they would previously _sometimes_ do,
e.g. a container service would have the same version as its module—which
is in fact the problem we're solving here.

I've added a documentation section and an FAQ entry, but we should also
mention this in release notes and in the community when announcing the
change.

This could use some manual testing, even though I did my best to write functional tests to validate the behavior. This touches on a lot of critical paths, so we should be careful when reviewing and evaluating this.

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -241,6 +241,7 @@
     "build": "gulp pegjs",
     "check-package-lock": "git diff-index --quiet HEAD -- yarn.lock || (echo 'yarn.lock is dirty!' && exit 1)",
     "clean": "shx rm -rf build",
+    "dev": "gulp pegjs-watch",
     "fix-format": "prettier --write \"{src,test}/**/*.ts\"",
     "lint": "tslint -p .",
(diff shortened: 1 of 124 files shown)


thsig (approved):

thsig:
There's a conflict here now (after another PR was merged today). Once that's been resolved, I'll go ahead and merge this.

PR #2296, opened 2021-03-09
Fix publish issues

edvald:
Three issues fixed:

1. Resource contention due to sharing a container for image pulls. We now start a Pod for each pull.
2. Incorrect command used for loading the extracted image archive. This would _sometimes_ work correctly, but not reliably.
3. This is tangential to the PR, but I noticed we were buffering data that was also being directly streamed when fetching artifacts, which might have caused issues and was definitely unnecessary + slow. I needed to fix this along the way because we'd otherwise have the same problem when fetching image archives.


thsig:
Looks good!

Some of the integ tests are failing though (the error messages indicate a missing `tar` binary in the image).

Do we need to update the underlying test projects?

edvald:
Hmm I think it's more likely to be an actual error, I'll take a look.

thsig (approved):

PR #2297, opened 2021-03-09
Dashboard styling fixes

edvald:
Various fixes based on feedback on the new dashboard layout. Please check out the individual commits for review, and check each individually to see that it all works as desired.


thsig:
Good stuff!


Couple of points: I'm still getting a visible scrollbar in the overview cards. (and the spacing between the `Build | Run | Deploy | Test` select boxes and the vertical scrollbar on the right is still a bit small):

Screenshot 2021-03-10 at 11 00 00

edvald:
I'm not able to reproduce those horizontal scrollbars. Can you maybe poke around the CSS there and debug?

thsig (approved):

PR #2304, opened 2021-03-12
improvement(cli): allow `garden delete services` with no arguments

edvald:
Basically this makes it easy to delete all services in a namespace.
Sort of a silly omission to begin with.


djowinz:
NICE

thsig (approved):

PR #2306, opened 2021-03-14
Refactor Enterprise API class

eysi09:

This is the first part of a larger change where we add some enterprise utility commands. I decided to split this up so that we can start testing out the refactor right away since it touches on some core logic.

In particular:

- We now initialise the Enterprise API in a single factory function. If the user is logged in we, return the api instance, otherwise null. The actual login function is now a static method on the Enterprise API class that we can call without initialising it (since we can't initialise it without being logged in).
- Added some more tests for the login / logout flow.

Fixes #

It would be good to get this merged as soon as possible so that we can test this properly internally before releasing.


core/src/commands/login.ts
```diff
@@ -33,27 +35,46 @@ export class LoginCommand extends Command {
...
+    await EnterpriseApi.saveAuthToken(log, tokenResponse)
     log.info({ msg: `Successfully logged in to Garden Enteprise.` })
-
     return {}
   }
 }
+
+export async function login(log: LogEntry, enterpriseDomain: string, events: EventBus) {
+  // Start auth redirect server and wait for its redirect handler to receive the redirect and finish running.
+  const server = new AuthRedirectServer(enterpriseDomain, events, log)
```
eysi09, line 64: Previously we were creating a specific instance of the events class for this logic but now we're using the events bus that's attached to the Garden class. Any objections to that? (cc @thsig)

The main reason I made this change is that it makes it a lot easier to test this logic, which we weren't really doing before.

tsconfig.base.json
```diff
@@ -8,6 +8,7 @@
     "emitDecoratorMetadata": true,
     "esModuleInterop": true,
     "experimentalDecorators": true,
+    "skipLibCheck": true,
```
eysi09, line 11: We should remove this check before merging, but without it, building the code fails with:

```
cli ┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄  ../plugins/conftest-container/node_modules/@types/node/http.d.ts(1,1): error TS6200:     ┄┄
                             Definitions of the following identifiers conflict with those in another file:            ┄┄
                             RequestListener, Server, OutgoingMessage, ServerResponse, ClientRequest,                 ┄┄
                             IncomingMessage, Agent, METHODS, STATUS_CODES, globalAgent, maxHeaderSize                ┄┄
cli ┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄  ../node_modules/@types/node/http.d.ts(1,1): error TS6200: Definitions of the following   ┄┄
                             identifiers conflict with those in another file: RequestListener, Server,                ┄┄
                             OutgoingMessage, ServerResponse, ClientRequest, IncomingMessage, Agent, METHODS,         ┄┄
                             STATUS_CODES, globalAgent, maxHeaderSize                                                 ┄┄
cli ┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄  ../node_modules/@types/node/http.d.ts(100,9): error TS2717: Subsequent property          ┄┄
                             declarations must have the same type.  Property 'ServerResponse' must be of type         ┄┄
                             'typeof ServerResponse | undefined', but here has type 'typeof ServerResponse |          ┄┄
                             undefined'.  
```

I'm completely blocked on this and would love some pointers.

eysi09: Spoke to @edvald about this one and decided it's be okay to merge this with `skipLibCheck` and revisit.

core/src/commands/login.ts
thsig: Yeah, that makes sense. Can't think of any problems with it, at any rate.

core/src/commands/logout.ts
```diff
@@ -26,19 +26,23 @@ export class LogOutCommand extends Command {
   }
 
   async action({ garden, log }: CommandParams): Promise<CommandResult> {
-    if (!garden.enterpriseApi?.getDomain()) {
-      // If no domain is found or enterpriseApi is null, this is a noop
+    // Then enterprise API is not on the Garden class because this command has noProject=true
```
thsig, line 29: Let's reuse the comment you wrote for the login command here (the wording is better there imho).

core/src/enterprise/api.ts
```diff
@@ -32,146 +34,122 @@ const refreshThreshold = 10 // Threshold (in seconds) subtracted to jwt validity
...
+      if (!tokenIsValid) {
+        enterpriseLog.debug({ msg: `Current auth token is invalid, refreshing` })
+        try {
+          // We can assert the token exsists since we're not using GARDEN_AUTH_TOKEN
+          await api.refreshToken(token!)
+        } catch (err) {
+          enterpriseLog.setError({ msg: `Invalid session`, append: true })
+          enterpriseLog.warn(deline`
+          Your session is invalid and could not be refreshed. If you were previously logged
+          into another instance of Garden Enterprise, please log out first and then
```
thsig, line 136: "logged into" -> "logged in to"? Fine either way, I guess.

core/src/enterprise/api.ts
```diff
@@ -185,117 +163,205 @@ export class EnterpriseApi {
...
-
+  static async getClientAuthTokenFromDb(log: LogEntry) {
     const [tokens, tokenCount] = await ClientAuthToken.findAndCount()
 
-    const token = tokens[0] ? tokens[0].token : null
+    const token = tokens[0] ? tokens[0] : undefined
 
     if (tokenCount > 1) {
-      this.log.debug("More than one client auth tokens found, clearing up...")
+      log.debug("More than one client auth tokens found, clearing up...")
```
thsig, line 184: "tokens" -> "token"

thsig (approved):
Good stuff! Had a few minor text-related comments there.

PR #2313, opened 2021-03-17
feat(enterprise): add requests to workflows

thsig:

Resource requests for workflow runner pods can now be set via `workflow.requests`.

### DIFF
--- a/core/src/config/workflow.ts
+++ b/core/src/config/workflow.ts
@@ -18,7 +18,7 @@ import {
 } from "./common"
 import { DEFAULT_API_VERSION } from "../constants"
 import { deline, dedent } from "../util/string"
-import { defaultContainerLimits, ServiceLimitSpec } from "../plugins/container/config"
+import { ServiceLimitSpec } from "../plugins/container/config"
 import { Garden } from "../garden"
 import { WorkflowConfigContext } from "./template-contexts/workflow"
 import { resolveTemplateStrings } from "../template-string/template-string"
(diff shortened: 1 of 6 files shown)


core/src/config/workflow.ts
```diff
@@ -87,21 +107,30 @@ export const workflowConfigSchema = () =>
         .number()
         .default(48)
         .description("The number of hours to keep the workflow pod running after completion."),
+      requests: joi
```
edvald, line 110: I'm not a fan of this schema. Could we put this and `limits` under a `resources` key instead? And of course keep the `limits` key at the root for compatibility.

thsig: Yeah, I like that idea—will do.

thsig: Should be good now. Any other comments/suggestions for this PR?

core/src/config/workflow.ts
```diff
@@ -87,21 +133,15 @@ export const workflowConfigSchema = () =>
...
-          memory: joi
-            .number()
-            .default(defaultContainerLimits.memory)
-            .min(256)
-            .description("The maximum amount of RAM the workflow pod can use, in megabytes (i.e. 1024 = 1 GB)"),
+          requests: workflowResourceRequestsSchema(),
+          limits: workflowResourceLimitsSchema(),
         })
-        .default(defaultContainerLimits),
+        .default(() => ({}))
```
edvald, line 142: I think we should still have some defaults. Maybe default to the minimum for requests, and keep the prior default for the limits?

thsig: Yeah, I agree.

In that case, I think we'd have to apply the defaults in `resolveWorkflowConfig` (since otherwise, it's hard to tell whether the value in `workflow.resources.limits` comes from the joi default or whether the user wrote it in).

edvald: Hmm, why is that? As in, why do we need to know that?

thsig: I.e. if the user provides a value in `workflow.limits` but not in `workflow.resources.limits`, the default value will be populated by joi in `workflow.resources.limits`.

In that case, we'd want to override the auto-populated default in `workflow.resources.limits` with the user-provided value in `workflow.limits`.

Maybe a compromise would be for `workflow.limits` to always override `workflow.resources.limits` if `workflow.limits` is provided? That bypasses this ambiguity altogether.

edvald: Yeah I think that should be fine.

thsig: I think this should be good now.

core/src/config/workflow.ts
```diff
@@ -31,13 +31,39 @@ import { globalOptions } from "../cli/params"
...
+
+export const minimumWorkflowLimits = {
+  cpu: 100, // 100 millicpu
+  memory: 64, // 64MB
+}
+
+export const defaultWorkflowLimits = {
+  cpu: 300, // 300 millicpu
+  memory: 256, // 256MB
+}
```
edvald, line 49: This changes the previous default, which is sort of a breaking change (and I'm not sure a sensible default anyway). Why not just keep the previous default?

thsig: Ok makes sense, I'll go back to the previous default.

thsig: Done!

edvald (approved):

PR #2314, opened 2021-03-17
feat(k8s): apply container service annotations to Pod templates as well

edvald:
When using the Kubernetes provider, the annotations set on
`services[].annotations` in container modules are now applied to both
Service and Pod resources. You can generally specify the annotations
intended for both Pods or Services here, and the ones that don't apply
on either side will be ignored (i.e. if you put a Service annotation
here, it'll also appear on Pod specs but will be safely ignored there,
and vice versa).


thsig (approved):

thsig:
Looks good! Some of the tests are still failing there, once those are fixed we can get this merged.

PR #2322, opened 2021-03-24
improvement(enterprise): better error message on login 401 errors

eysi09:

I think this only happens for developers who may be logging into different instances of GE during testing. But who's to say we shouldn't get good error messages as well. 

Before:

```
Login ☁️

An error occurred while verifying client auth token with platform: Response code 401 (Unauthorized)

```

After:

```
Login ☁️

⚠ Looks like your session token is invalid. If you were previously logged into a different instance
of Garden Enterprise, log out first before logging in.

An error occurred while verifying client auth token with platform: Response code 401 (Unauthorized)

```

Fixes #

Feel free to add suggestions to the error message.

### DIFF
--- a/core/src/enterprise/api.ts
+++ b/core/src/enterprise/api.ts
@@ -10,7 +10,7 @@ import { IncomingHttpHeaders } from "http"
 
 import { got, GotHeaders, GotHttpError } from "../util/http"
 import { findProjectConfig } from "../config/base"
-import { CommandError, EnterpriseApiError, RuntimeError } from "../exceptions"
+import { CommandError, EnterpriseApiError } from "../exceptions"
 import { LogEntry } from "../logger/log-entry"
 import { gardenEnv } from "../constants"
 import { ClientAuthToken } from "../db/entities/client-auth-token"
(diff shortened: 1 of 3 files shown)


core/src/enterprise/api.ts
```diff
@@ -290,7 +291,11 @@ export class EnterpriseApi {
       await EnterpriseApi.saveAuthToken(this.log, tokenObj)
     } catch (err) {
       this.log.debug({ msg: `Failed to refresh the token.` })
-      throw new RuntimeError(`An error occurred while verifying client auth token with platform: ${err.message}`, {})
+      const detail = is401Error(err) ? { statusCode: err.response.statusCode } : {}
+      throw new EnterpriseApiError(
+        `An error occurred while verifying client auth token with platform: ${err.message}`,
```
thsig, line 296: Maybe we should use _Garden Enterprise_ instead of _platform_ for these messages?

thsig (approved):
Looks good! Had one minor comment/suggestion there.

core/src/enterprise/api.ts
eysi09: I've updated the PR.

PR #2323, opened 2021-03-25
improvement(core): increase liveness probes when in hot-reload mode

eysi09:

During a hot reload event, a pod might not respond to liveness probes, e.g. because the server is restarting. This causes issues where K8s terminates the pod mid restart and the process needs start from the scratch. This PR fixes that be increasing the liveness probes on manifests for the container module type when in hot reload mode. 

The values are what we've seen work for GE but happy to receive other suggestions.

Fixes #

### DIFF
--- a/core/test/integ/src/plugins/kubernetes/container/deployment.ts
+++ b/core/test/integ/src/plugins/kubernetes/container/deployment.ts
@@ -137,6 +137,37 @@ describe("kubernetes container deployment handlers", () => {
       expect(resource.spec.template?.metadata?.annotations).to.eql(service.spec.annotations)
     })
 
+    it("should increase liveness probes when in hot-reload mode", async () => {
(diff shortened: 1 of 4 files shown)


edvald (commented):
Looks good but one or more integ tests need to be updated.

core/test/integ/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -137,6 +137,31 @@ describe("kubernetes container deployment handlers", () => {
       expect(resource.spec.template?.metadata?.annotations).to.eql(service.spec.annotations)
     })
 
+    it("should increase liveness propves when in hot-reload mode", async () => {
```
thsig, line 140: "propves" -> "probes"

edvald (approved):

PR #2325, opened 2021-03-26
chore: add aws-gcloud-azure Docker build

thsig:


edvald:
I think this is okay-ish for now, but I think going forward we have to reconsider the naming scheme and Docker repo layout. This should all be in the same repo with different tags, and I really feel like we should have a `slim` and a `full` tag or something like that.

twelvemo (approved):
looks good to me 👍

PR #2337, opened 2021-04-02
WIP: add arm64 to the cli

### LINKED ISSUE #2333: [FEATURE]: ARM64 support for the garden CLI
## Feature Request

Hi! I would love to be able to use garden from my low power ARM64 based laptop.

### Background / Motivation

I work on a fairly heavyweight project, a web service backed by multiple docker containers running java apps, but I'd love to be able to work from a low power machine without being tied to a large workstation all day.

Install and use garden CLI on ARM64 hardware, connected to a cluster of any architecture (most likely x86_64)

See motivation above.

I have tried to simply remove the check in `src/core/garden.ts` and building according to the instructions in CONTRIBUTING.md. This results in a usable `garden` executable being built. When I try to initialize a remote cluster to use with garden however, I get an error

Sample garden.yml, just the basic example. k8s cluster is running on my x86_64 workstation, and kubectl is connected correctly.
```
kind: Project
name: phr
environments:
- name: dev
  providers:
  - name: kubernetes
    context: k3d-k3s-default
defaultEnvironment: dev
```

Sample error
```
[martin@martin-pinebook phr]$ garden --env=dev plugins kubernetes cluster-init
Plugins

Initializing/updating cluster-wide services for dev environment

ℹ providers                 → Getting status...

Failed resolving provider container. Here is the output:
Command docker doesn't have a spec for this platform/architecture (linux-arm64)
Failed resolving one or more providers:
- container

 providers                 → Getting status...
```

MFAshby:

Adds support for using the garden CLI on arm64 based computers.

Fixes #2333

### DIFF
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -13,14 +13,21 @@ import { GardenProcess } from "@garden-io/core/build/src/db/entities/garden-proc
 import { ensureConnected } from "@garden-io/core/build/src/db/connection"
 import { GardenCli, RunOutput } from "@garden-io/core/build/src/cli/cli"
 import { GardenPluginCallback } from "@garden-io/core/build/src/types/plugin/plugin"
+import { arch } from "os"
 
 // These plugins are always registered
-export const getBundledPlugins = (): GardenPluginCallback[] =>
-  [
+export const getBundledPlugins = (): GardenPluginCallback[] => {
+  if (arch() === "arm64") {
+    // conftest has no arm64 installation candidate.
+    // https://github.com/open-policy-agent/conftest/issues/516
+    return []
+  }
+  return [
     require("@garden-io/garden-conftest"),
     require("@garden-io/garden-conftest-container"),
     require("@garden-io/garden-conftest-kubernetes"),
   ].map((m) => () => m.gardenPlugin())
+}
 
 export async function runCli({
   args,
--- a/core/src/garden.ts
+++ b/core/src/garden.ts
@@ -10,7 +10,7 @@ import Bluebird from "bluebird"
 import chalk from "chalk"
 import { ensureDir } from "fs-extra"
 import dedent from "dedent"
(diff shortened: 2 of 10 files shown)


cli/src/cli.ts
```diff
@@ -13,14 +13,21 @@ import { GardenProcess } from "@garden-io/core/build/src/db/entities/garden-proc
...
+import { arch } from "os"
 
 // These plugins are always registered
-export const getBundledPlugins = (): GardenPluginCallback[] =>
-  [
+export const getBundledPlugins = (): GardenPluginCallback[] => {
+  if (arch() === "arm64") {
+    // conftest has no arm64 installation candidate.
+    // https://github.com/open-policy-agent/conftest/issues/516
+    return []
```
edvald, line 23: Hey @MFAshby, thanks for the PR! I think we should be able to get this done, specifically adding support for linux/arm64.

I do see a linux arm64 build for conftest here: https://github.com/open-policy-agent/conftest/releases/tag/v0.23.0
I think we just need to update the conftest version we use. Unsure if there are API changes to consider there, but generally shouldn't be a big deal to update.

That aside, I'd much prefer to handle this case (missing individual platform/architectures for tools) in a generalized manner. I wonder if this is already enough for the context: https://github.com/garden-io/garden/blob/1033e59f2ab413f201764ef3d93068e345171e1d/core/src/util/ext-tools.ts#L178

edvald:
Quick update on this: The upstream project we rely on to build the Garden binary now has support for ARM64, so we could resume work on this soon.

thsig:
Have we had a chance to look at this one again?

MFAshby:
Unfortunately not, work has taken me in other directions at the moment.

MFAshby:
Practically I'm not going to get back to this unfortunately. Fork remains up in case anyone else wants a go.

PR #2345, opened 2021-04-14
feat(workflows): only resolve config being run

thsig:

In the `run workflow` command, we now only resolve the workflow config being run.

This adds flexibility when some workflows reference e.g. secrets that are created by other workflows.

### DIFF
--- a/core/test/unit/src/garden.ts
+++ b/core/test/unit/src/garden.ts
@@ -2261,7 +2261,7 @@ describe("Garden", () => {
       const garden = await makeTestGarden(resolve(dataDir, "test-projects", "custom-config-names"))
       await garden.scanAndAddConfigs()
 
-      const workflows = garden.getWorkflowConfigs()
+      const workflows = await garden.getRawWorkflowConfigs()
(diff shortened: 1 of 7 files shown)


thsig:
This should be good to go now (the `validate` command resolves all workflow configs).

core/test/unit/src/garden.ts
```diff
@@ -2476,7 +2476,7 @@ describe("Garden", () => {
       )
     })
 
-    it("should throw an error if an invalid workflow config is present", async () => {
+    it.skip("should throw an error if an invalid workflow config is present", async () => {
```
eysi09, line 2479: Did you forget this?

eysi09 (changes requested):
A test is being skipped, otherwise looks good.

core/test/unit/src/garden.ts
thsig: No, actually—we don't want to throw an error here if an invalid workflow config is present (since that would imply resolving all workflow configs during the scan).

I've removed this test case now, that should be clearer.

eysi09 (approved):

PR #2359, opened 2021-04-21
Fixing typo for ingress annotations in container module docs

tvollstaedt:


eysi09 (approved):
Thanks!

eysi09:
....whoops, merged too soon. This is actually from our generated docs, so it looks like there's an issue with the code that spits out the doc string. I can take a look at that. In the mean time I've reverted this because otherwise our CI pipelines break.

PR #2362, opened 2021-04-25
fix(core): include deps in test version

thsig:

This ensures that tests are re-run when a downstream dependency changes that is more than one hop away in the config graph.


eysi09 (approved):
LGTM

PR #2363, opened 2021-04-26
feat(commands): skip-dependants opt for test cmd

thsig:

The current behavior of the test command when called with a list of module names is to run all tests in those modules, and all tests in modules with dependencies on those modules.

This is not the desired behavior in all cases, so here we add a `--skip-dependants` option to the test command. When used, only tests in the requested modules will be run.


eysi09 (approved):
LGTM. Do you think that failing e2e vote test is related to this change though? Perhaps worth verifying before we merge.

PR #2364, opened 2021-04-26
feat(template): add liveness and readiness timeout definitions for co…

tvollstaedt:

The `container` module abstracts many kubernetes internals by providing a simple template syntax. It allows to define health checks to make sure our service are in a proper state when deployed. Garden sets sensible defaults for timeout values for those health checks. 
However, sometimes it may be necessary to adjust these timeouts because services need more time for initial startup or similar. It's currently not possible to adjust these values within the container module, one would need to switch over to the `kubernetes`-module to gain full control over deployment, but also has to have deeper knowledge about Kubernetes for that.

This PR adds control over liveness and readiness timouts through templating to avoid the mentioned workaround.


core/src/plugins/container/config.ts
```diff
@@ -240,6 +242,16 @@ const healthCheckSchema = () =>
       tcpPort: joi
         .string()
         .description("Set this to check the service's health by checking if this TCP port is accepting connections."),
+      readinessTimeout: joi
```
thsig, line 245: Let's rename these to `readinessTimeoutSeconds` and `livenessTimeoutSeconds`, just to be extra obvious to the end-user.

core/src/plugins/container/config.ts
```diff
@@ -240,6 +242,16 @@ const healthCheckSchema = () =>
...
+        .number()
+        .min(1)
+        .default(3)
+        .description("The maximum amount of seconds to wait until the readiness check counts as failed."),
```
thsig, line 249: "amount of" -> "number of"

thsig (changes requested):
Many thanks for the contribution here! Had a couple of minor comments.

Once we've addressed those squashed everything into one commit, we should be ready to merge this.

tvollstaedt:
Not quite sure what happened to the docs, just re-ran the script 😨

tvollstaedt:
Should be ready now.

tvollstaedt:
@thsig Is there anything else you would like me to change?

thsig (approved):

PR #2372, opened 2021-05-03
improvement(k8s): get rid of NFS when using kaniko build mode

### LINKED ISSUE #1798: Get rid of NFS dependency for in-cluster building
## Background

One of the most frustrating issues with our remote building feature is the reliance on shareable (ReadWriteMany) volumes, which currently works through NFS at the moment, and optionally other RWX-capable volume provisioners such as EFS. This has turned out to cost a lot of maintenance burden and operational issues for our users.

The reason we currently have this requirement is that in order to sync code from the user's local build staging directory, we rsync to an in-cluster volume, that needs to be mountable by the in-cluster builder, whether that is an in-cluster docker daemon, or kaniko.

At the time we didn't see an alternative, that would be reasonably performant. We also had a performance goal in mind that may simply not be as important anymore. The raw efficiency of rsync is appealing, but in many (even most) cases the sync of code is only a small part of the overall build time.

## Proposed solution

We can avoid this requirement altogether, through some added code complexity in the actual build flows, but in turn avoiding the complexity relating to managing finicky storage providers.

We still rsync over to the cluster, but instead of directly mounting the sync volume, we modify the flow to allow using a simpler RWO volume.

### Build flow

Depending on the build mode, we do the following:

- cluster-docker
    - We add an rsync container to the cluster-docker Pod spec, and instead of referencing a shared PVC for the build sync volume, we have an RWO PVC that the main docker container and the rsync container share.
    - When building, we do the exact same thing as previously, but instead of syncing to a separate build sync Pod, we sync to the rsync container in the cluster-docker Pod, before executing the build.
- Kaniko
    - We still create a build sync Deployment and a PVC, but the PVC can now be RWO, and we only have one sync Pod, much like the in-cluster registry.
    - We still sync to the build-sync Pod ahead of the build.
    - In the Kaniko build Pod, we create an init container, that rsyncs *from* the build-sync Pod to the container filesystem, replacing the shared build-sync PVC reference with a shared emptyDir volume within the Pod.
    - The Kaniko build Pod does what it did previously.

### Migration

We change the name of the `build-sync` service to `build-sync-v2`, and `docker-daemon` to `docker-daemon-v2`(and the helm release names accordingly). This is to avoid conflicts during rollout since they both function differently from the prior versions, and cannot cover both client versions simultaneously.

Users need to be instructed to remove the old `build-sync` and `docker-daemon` deployments and volumes manually, as well as the NFS provisioner, when their team has updated to the new version. Or uninstall and re-init completely, of course. We can print out a message to this effect in the cluster-init command.

The current `storage.sync` parameters still apply to the new `build-sync-v2` volume without any changes.

We still always install `build-sync-v2`, even though it isn't necessary when using the `cluster-docker` build mode, in order to avoid headaches around using cluster-docker and kaniko in different scenarios on the same cluster.

If it is installed, the `cleanup-cluster-registry` script now also executes through the rsync container in the docker daemon Pod, in addition to the `build-sync-v2` Pod.

## Benefits

1. Far simpler requirements for in-cluster building.
2. Fewer bugs and support headaches.
3. Less operation overhead for customers.
4. New Kaniko flow can be made to work without even re-running cluster-init (falling back to older build-sync volume if it's available).

## Drawbacks

1. The two build modes change a bit and have somewhat different flows after the transition.
2. There is a slightly cumbersome transition, having to instruct users to clean up older system services manually, and potentially in parallel for a bit.

## Prioritization

I'd say sooner is better for this, since this causes annoying operational issues for users, and support issues for the Garden team by extension.

edvald:
I'm sure this'll be a big relief for many users. This is a drop-in upgrade that removes the need for NFS, `build-sync` and other cluster-wide services when using the `kaniko` build mode.

Additionally, users can set `kaniko.namespace: null` in their `kubernetes` provider configs to run the Kaniko pods in the project namespace instead of `garden-system`, which will be helpful in scenarios where it's problematic to allow users to run pods in a shared namespace.

It's also no longer necessary to run `garden plugins kubernetes cluster-init` when using Kaniko, except to install unrelated services (like an ingress controller etc.).

I've updated the documentation and examples to default to Kaniko, and updated the in-cluster building guide and recommendations therein to reflect the changes.

Closes #1798

### DIFF
--- a/core/src/plugins/kubernetes/container/build/buildkit.ts
+++ b/core/src/plugins/kubernetes/container/build/buildkit.ts
@@ -9,22 +9,25 @@
 import AsyncLock from "async-lock"
 import chalk from "chalk"
 import split2 = require("split2")
-import { cloneDeep, isEmpty } from "lodash"
-import {
-  buildSyncVolumeName,
-  dockerAuthSecretKey,
-  inClusterRegistryHostname,
-  k8sUtilImageName,
-  rsyncPortName,
-} from "../../constants"
+import { isEmpty } from "lodash"
+import { buildSyncVolumeName, dockerAuthSecretKey, inClusterRegistryHostname } from "../../constants"
 import { KubeApi } from "../../api"
 import { KubernetesDeployment } from "../../types"
 import { LogEntry } from "../../../../logger/log-entry"
 import { waitForResources, compareDeployedResources } from "../../status/status"
 import { KubernetesProvider, KubernetesPluginContext } from "../../config"
 import { PluginContext } from "../../../../plugin-context"
-import { prepareDockerAuth } from "../../init"
-import { BuildStatusHandler, skopeoBuildStatus, BuildHandler, syncToBuildSync, getSocatContainer } from "./common"
+import {
+  BuildStatusHandler,
+  skopeoBuildStatus,
+  BuildHandler,
+  syncToBuildSync,
+  getSocatContainer,
+  getUtilContainer,
+  utilRsyncPort,
+  ensureBuilderSecret,
+  builderToleration,
+} from "./common"
 import { getNamespaceStatus } from "../../namespace"
 import { containerHelpers } from "../../../container/helpers"
 import { LogLevel } from "../../../../logger/log-node"
@@ -33,13 +36,10 @@ import { ContainerModule } from "../../../container/config"
 import { getDockerBuildArgs } from "../../../container/build"
 import { getDeploymentPod, millicpuToString, megabytesToString } from "../../util"
 import { PodRunner } from "../../run"
-import { V1Container } from "@kubernetes/client-node"
 
 export const buildkitImageName = "gardendev/buildkit:v0.8.1-4"
 export const buildkitDeploymentName = "garden-buildkit"
-export const buildkitAuthSecretName = "garden-docker-auth"
 const buildkitContainerName = "buildkitd"
-const utilRsyncPort = 8730
 
 const deployLock = new AsyncLock()
 
@@ -51,7 +51,7 @@ export const getBuildkitBuildStatus: BuildStatusHandler = async (params) => {
   const api = await KubeApi.factory(log, ctx, provider)
   const namespace = (await getNamespaceStatus({ log, ctx, provider })).namespaceName
 
-  await ensureBuildkit({
+  const { authSecret } = await ensureBuildkit({
     ctx,
     provider,
     log,
@@ -62,7 +62,7 @@ export const getBuildkitBuildStatus: BuildStatusHandler = async (params) => {
   return skopeoBuildStatus({
     namespace,
     deploymentName: buildkitDeploymentName,
-    containerName: utilContainer.name,
+    containerName: getUtilContainer(authSecret.metadata.name).name,
     log,
     api,
     ctx,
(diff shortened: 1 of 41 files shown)


edvald:
Looking at the failing tests, will ping you once it's cleaned up @thsig

thsig:
Cool, thanks!

eysi09:
Just out of curiosity, is there a trade-off here? Like, was there any particular benefit to NFS to begin with that we'll now need to do without it?

edvald:
Good question. There is a slight efficiency loss in that the build context does need to be copied to the Kaniko Pod, instead of just mounted. But I think that's outweighed by the operational simplicity and the instability (and sometimes slow performance) of NFS.

The reason NFS was picked in the first place was for performance but at the time we didn't suspect it would have the myriad operational issues we've found along the way.

docs/guides/in-cluster-building.md
```diff
@@ -47,35 +78,35 @@ The other modes—which are why you're reading this guide—all build your image
...
+- [**`cluster-docker`**](#cluster-docker) was the first implementation included with Garden. It's pretty quick and efficient for small team setups, but relies on a single Docker daemon for all users of a cluster, and also requires supporting services in `garden-system` and some operations to keep it from filling its data volume. It is **no longer recommended** and we may remove it in future releases.
+
+Generally we recommend picking either `kaniko` or `cluster-buildkit`, based on your usage patterns and scalability requirements. For ephemeral namespaces, `kaniko` is generally the better option, since the persistent BuildKit deployment won't have a warm cache anyway. For long-running namespaces, like the ones a developer uses while working, `cluster-buildkit` may be a more performant option.
```
thsig, line 85: "long-running namespaces" -> "long-lived namespaces"

thsig (changes requested):
Looks good! Only had a few text-related comments.

edvald:
Running into issues with CI atm (not related to this PR), but this should be all set for final review.

edvald:
Actually I have found one legitimate issue, with different Kaniko build flows stepping over each other in shared namespaces. I'll need to fix that before we wrap this up.

edvald:
@thsig this should be all good to go now. I'd suggest doing a second code review, since I had to shuffle a couple of things around.

core/src/plugins/kubernetes/container/build/buildkit.ts
```diff
@@ -190,23 +190,27 @@ export async function ensureBuildkit({
...
 
+    // Check status of the buildkit deployment
+    const manifest = getBuildkitDeployment(provider, authSecret.metadata.name)
+    const status = await compareDeployedResources(ctx as KubernetesPluginContext, api, namespace, [manifest], deployLog)
+
     if (status.state === "ready") {
-      return false
+      // Need to wait a little to ensure the secret is updated in the deployment
+      if (secretUpdated) {
+        await sleep(5)
```
thsig, line 211: Is 5ms enough here? Or was this meant to be 5 seconds?

core/src/plugins/kubernetes/container/build/kaniko.ts
```diff
@@ -227,22 +472,27 @@ export async function runKaniko({
...
     pod,
     provider,
-    namespace,
+    namespace: kanikoNamespace,
   })
 
   const result = await runner.runAndWait({
     log,
-    remove: true,
+    remove: false, // TODO
```
thsig, line 495: Should this be left here as is?

thsig (changes requested):
Looks good! Had a couple of minor comments there.

edvald:
@thsig comments resolved!

thsig (approved):

marcelloromani:
Great news! Looking forward to download the next release which includes these changes :)

ArnaudRinquin:
Congrats! Has this been shiped with 0.15?

PR #2383, opened 2021-05-09
fix(core): ensure pod runner throws when container is OOMKilled

eysi09:

We've seen issues were a container exceeds its memory limits and is OOMKilled by Kubernetes, but the `terminted` state under the `containerStatus` field has exit code 0. However, the exit reason is still `"OOMKilled"`. This PR ensures that the pod runner throws when that is the case.  

Fixes #


edvald (approved):

edvald:
The e2e-vote test is still failing but that's unrelated to the PR. Merging as-is.

PR #2386, opened 2021-05-12
fix(k8s): automatic retry for failed API requests

### LINKED ISSUE #2255: Intermittent failure of DNS resolution when running garden agains domain based k8s
## Bug

Garden intermittently fails to resolve the DNS server name of my K8s cluster within my `kubernetes` provider

### Current Behavior

During CI/CD in drone using docker runner
```
kind: pipeline
type: docker
name: garden

steps:
  - name: build
    image: gardendev/garden:0.12.16
    pull: if-not-exists
    commands:
      - ./ci/kubectl-init-1.sh
      - garden -e staging build
    environment:
      GARDEN_LOG_LEVEL: silly # set the log level to your preference here
      GARDEN_LOGGER_TYPE: basic # this is important, since the default logger doesn't play nice with CI :)
      GARDEN_DEV_KUBECTL:
        from_secret: GARDEN_DEV_KUBECTL
```

I get the following output intermittently. At this point it has already build 4+ other services in the k8s cluster using kaniko on the cluster `rancher.<redacted>`

Example 1:
```
...
2007 | ✖ web                       → Building image web:v-1b8a684d8a... | 182s
2008 |   | 182s
2009 | Failed building web. Here is the output: | 182s
2010 | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ | 182s
2011 | getaddrinfo ENOTFOUND rancher.<redacted> | 182s
2012 | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ | 182s
2013 | getaddrinfo ENOTFOUND rancher.<redacted> | 182s
2014 |   | 182s
2015 | Error Details: | 182s
2016 |   | 182s
2017 | errno: ENOTFOUND | 182s
2018 | code: ENOTFOUND | 182s
2019 | syscall: getaddrinfo | 182s
2020 | hostname: rancher.<redacted>
...
```

Example 2:

```
...
611 | Failed getting status for service 'mysql-dotnet' (from module 'mysql-dotnet'). Here is the output: | 14s
612 | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ | 14s
613 | Error: getaddrinfo ENOTFOUND rancher.<redacted> | 14s
614 | ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ | 14s
615 | Error: getaddrinfo ENOTFOUND rancher.<redacted> | 14s
616 |   | 14s
617 | Error Details: | 14s
618 |   | 14s
619 | name: RequestError | 14s
620 | options: | 14s
621 | url: >- | 14s
622 | https://rancher.<redacted>/k8s/clusters/c-snrh4/api/v1/namespaces/<redacted>/services/mysql-dotnet | 14s
623 | method: GET | 14s
624 | json: true | 14s
625 | resolveWithFullResponse: true | 14s
626 | headers: | 14s
627 | Authorization: >- | 14s
628 | Bearer | 14s
629 | kubeconfig-u-udzztfyxtx:9h<redacted> | 14s
630 | simple: true | 14s
631 | transform2xxOnly: false
...
```

### Expected behavior

As the code has already resolved the DNS name correctly before in the code it should not fail to resolve it after the 1st time

### Reproducible example

Intermittently means it is not easily replaceable.
This only happened on our CI/CD system and not our workstations so resource limitations might be at play

### Workaround

~~This started to working more consistently after we disabled all but 1 kubernetes provider in our project.garden.yaml file but we had 5 active provders with different configs before~~

(Updated)
Adding the host record into `/etc/hosts` works as a hotfix.
```
echo "123.123.123.123 myhost.mydomain.com" |sudo tee -a /etc/hosts
```

### Suggested solution(s)

Take a look at re-using the DNS resolutions better in the codebase

### Additional context

- Full logs of CI/CD output is availalable by personal request with silly log level. Not going to put it in the public domain

### Your environment

* **OS:** Linux, Drone docker runner using official image from Garden.io (see drone config above)
* **How I'm running Kubernetes:** EKS (we are using other but the EKS is the cluster our CI/CD is deploying to and failling)
* **Garden version:** 0.12.16

thsig:

We now automatically retry failed Kubernetes API requests if the reason for the failure matches certain conditions.

For example, timeouts or DNS-related errors will result in retries, but not 404/not found errors (and so forth), which will be thrown without retrying.

We can easily add more error codes and/or conditions to this logic if we discover further error cases that should result in retries.

Fixes #2255.

Should also fix the sporadic API timeouts that some users have been encountering in day-to-day usage.

Do we want to include any more status codes and/or error messages in the `shouldRetry` helper? Are there other kinds of conditions for which we want to retry?

Also, feel free to comment on the general approach here.

### DIFF
--- a/core/src/plugins/kubernetes/api.ts
+++ b/core/src/plugins/kubernetes/api.ts
@@ -10,6 +10,8 @@
 // tslint:disable-next-line:no-unused
 import { IncomingMessage } from "http"
 import { Agent } from "https"
+import chalk from "chalk"
+import httpStatusCodes from "http-status-codes"
 import { ReadStream } from "tty"
 import {
   KubeConfig,
@@ -51,7 +53,7 @@ import {
 } from "./types"
 import { LogEntry } from "../../logger/log-entry"
 import { kubectl } from "./kubectl"
-import { urlJoin } from "../../util/string"
+import { deline, urlJoin } from "../../util/string"
 import { KubernetesProvider } from "./config"
 import { StringMap } from "../../config/common"
 import { PluginContext } from "../../plugin-context"
@@ -178,7 +180,7 @@ export class KubeApi {
   public policy: WrappedApi<PolicyV1beta1Api>
   public rbac: WrappedApi<RbacAuthorizationV1Api>
 
-  constructor(public context: string, private config: KubeConfig) {
+  constructor(public log: LogEntry, public context: string, private config: KubeConfig) {
     const cluster = this.config.getCurrentCluster()
 
     if (!cluster) {
@@ -190,13 +192,13 @@ export class KubeApi {
 
     for (const [name, cls] of Object.entries(apiTypes)) {
       const api = new cls(cluster.server)
-      this[name] = this.wrapApi(api, this.config)
+      this[name] = this.wrapApi(log, api, this.config)
     }
   }
 
   static async factory(log: LogEntry, ctx: PluginContext, provider: KubernetesProvider) {
     const config = await getContextConfig(log, ctx, provider)
-    return new KubeApi(provider.config.context, config)
+    return new KubeApi(log, provider.config.context, config)
   }
 
   async getApiInfo(): Promise<ApiInfo> {
@@ -341,12 +343,14 @@ export class KubeApi {
     // apply auth
     await this.config.applyToRequest(requestOpts)
 
-    try {
-      log.silly(`${requestOpts.method.toUpperCase()} ${url}`)
-      return await request(requestOpts)
-    } catch (err) {
-      throw handleRequestPromiseError(err)
-    }
+    return await requestWithRetry(log, path, async () => {
+      try {
+        log.silly(`${requestOpts.method.toUpperCase()} ${url}`)
+        return await request(requestOpts)
+      } catch (err) {
+        throw handleRequestPromiseError(err)
+      }
+    })
   }
 
   /**
@@ -593,7 +597,7 @@ export class KubeApi {
   /**
    * Wrapping the API objects to deal with bugs.
    */
-  private wrapApi<T extends K8sApi>(api: T, config: KubeConfig): T {
+  private wrapApi<T extends K8sApi>(log: LogEntry, api: T, config: KubeConfig): T {
     api.setDefaultAuthentication(config)
 
     api.addInterceptor((opts) => {
@@ -615,28 +619,30 @@ export class KubeApi {
             target["defaultHeaders"] = { ...defaultHeaders, "content-type": "application/merge-patch+json" }
           }
 
-          const output = target[name](...args)
-          target["defaultHeaders"] = defaultHeaders
+          return requestWithRetry(log, name, () => {
+            const output = target[name](...args)
+            target["defaultHeaders"] = defaultHeaders
 
-          if (typeof output.then === "function") {
-            return (
-              output
-                // return the result body directly if applicable
-                .then((res: any) => {
-                  if (isPlainObject(res) && res.hasOwnProperty("body")) {
-                    return res["body"]
-                  } else {
-                    return res
-                  }
-                })
-                // the API errors are not properly formed Error objects
-                .catch((err: Error) => {
-                  throw wrapError(err)
-                })
-            )
-          }
+            if (typeof output.then === "function") {
+              return (
+                output
+                  // return the result body directly if applicable
+                  .then((res: any) => {
+                    if (isPlainObject(res) && res.hasOwnProperty("body")) {
+                      return res["body"]
+                    } else {
+                      return res
+                    }
+                  })
+                  // the API errors are not properly formed Error objects
+                  .catch((err: Error) => {
+                    throw wrapError(err)
+                  })
+              )
+            }
 
-          return output
+            return output
+          })
         }
       },
     })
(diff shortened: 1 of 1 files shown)


thsig:
Note: I didn't use `p-retry` here, since it doesn't play nice with our custom error classes (which include extra fields in addition to the message).

eysi09:
I think @edvald is best suited to review this one. I'll take a look but un-assign myself.

core/src/plugins/kubernetes/api.ts
```diff
@@ -881,3 +886,72 @@ function handleRequestPromiseError(err: Error) {
...
+): Promise<R> {
+  const maxRetries = opts?.maxRetries || 5
+  const minTimeoutMs = opts?.minTimeoutMs || 2000
+  const retry = async (usedRetries: number): Promise<R> => {
+    try {
+      return await req()
+    } catch (err) {
+      if (shouldRetry(err) && usedRetries < maxRetries) {
+        const sleepMsec = minTimeoutMs + usedRetries * minTimeoutMs
+        if (log) {
```
edvald, line 915: I'd prefer to go directly for the root logger as a fallback. Also, I think this should be a warn log. This is a bit too hidden atm.

core/src/plugins/kubernetes/api.ts
```diff
@@ -881,3 +886,72 @@ function handleRequestPromiseError(err: Error) {
...
+              (#${usedRetries}/${maxRetries})
+            `
+          )
+        }
+        await sleep(sleepMsec)
+        return await retry(usedRetries + 1)
+      } else {
+        if (usedRetries === maxRetries) {
+          if (log) {
+            log.debug(`Kubernetes API: Maximum retry count exceeded, throwing error`)
```
edvald, line 928: "throwing error" is maybe a bit odd for a user-facing message :)

core/src/plugins/kubernetes/api.ts
```diff
@@ -881,3 +886,72 @@ function handleRequestPromiseError(err: Error) {
...
+}
+
+const statusCodesForRetry: number[] = [
+  httpStatusCodes.BAD_GATEWAY,
+  httpStatusCodes.GATEWAY_TIMEOUT,
+  httpStatusCodes.REQUEST_TIMEOUT,
+  httpStatusCodes.SERVICE_UNAVAILABLE,
+]
+
+const errorMessageRegexesForRetry = [/getaddrinfo ENOTFOUND/, /getaddrinfo EAI_AGAIN/]
```
edvald, line 957: Isn't this available on the error object in a more structured form?

edvald (changes requested):
Couple of small comments there, otherwise seems solid!

core/src/plugins/kubernetes/api.ts
thsig: Haha yeah, good point—I've removed that text.

core/src/plugins/kubernetes/api.ts
thsig: I refactored this so that the wrapped API has access to the log entry passed to the `KubeApi.factory` method.

I also added a description parameter to the `requestWithRetry` function for including in the warning messages, which should add a bit of clarity to the log output.

core/src/plugins/kubernetes/api.ts
thsig: It might be, but the `wrapError` helper we're using doesn't include those fields (since they're not part of the `KubernetesError` class).

I thought matching against the error message was neater than adding several optional fields to the `KubernetesError` class, since we'd probably end up adding more and more of those fields going forward as we handle more of these cases.

I haven't reproduced it on my end, but here's an example of these error objects ([from here](https://github.com/nodejs/node/issues/5436)):
``` 
{ [Error: getaddrinfo ENOTFOUND host]
  code: 'ENOTFOUND',
  errno: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'host' }
```
That said, I'm totally open to adding more fields to the wrapped error (or maybe putting them all in an `inheritedFields` field with type `any`).

edvald: Let's not overthink it, no worries. I take it you tested this to make sure it works?

thsig: Yeah, I did—I had the logic throw errors and verified that the logging/retrying worked as expected.

The logging could be a bit nicer, but I think addressing that would require a bit of a refactor.

Here's a brief example snippet (where I hardcoded a `throw` with a `getaddrinfo ENOTFOUND` error, which matches the retry criteria):
```
✔ backend                   → Syncing module sources (2 files)... → Done (took 0 sec)
⠦ frontend                  → Getting build status for v-a57012fdb5...
⠼ backend                   → Getting build status for v-6fbe386fc0...
Kubernetes API: listNamespace failed with error test error: getaddrinfo ENOTFOUND, sleeping for 4000ms and retrying (#1/5)
Kubernetes API: /api/v1 failed with error test error: getaddrinfo ENOTFOUND, sleeping for 4000ms and retrying (#1/5)
Kubernetes API: /api/v1/namespaces/demo-project-testing-ths/secrets/garden-docker-auth-41e2da failed with error test error: getaddrinfo ENOTFOUND, sleeping
for 4000ms and retrying (#1/5)
Kubernetes API: patchNamespace failed with error test error: getaddrinfo ENOTFOUND, sleeping for 4000ms and retrying (#1/5)
```

edvald: Ok cool. I'd suggest "sleeping for 4000ms and retrying" -> "retrying in 4000ms", otherwise looks good.

Also, is 4000ms the value for the first retry? Seems a touch high, no?

thsig: > Ok cool. I'd suggest "sleeping for 4000ms and retrying" -> "retrying in 4000ms", otherwise looks good.

Yeah, that's better—will make that change.

> Also, is 4000ms the value for the first retry? Seems a touch high, no?

It's 2000ms for the first retry, actually (`2000 + 2000 * retries`)—this should say "#2/5".

Should we go for `500 + 500 * retries` instead?

edvald (approved):

PR #2387, opened 2021-05-12
feat(k8s): add exec to kubernetes and helm modules

### LINKED ISSUE #2373: [FEATURE]: Add support for Kubernetes modules with exec (to open an interactive shell)
### Background / Motivation

I am not entirely sure this should be a feature request or a bug. When running `garden exec` against a service deployed with the kubernetes module on a remote EKS cluster, I get the following error:
```
No 'execInService' handler configured for module type 'kubernetes' in environment 'remote'. 
Are you missing a provider configuration?
```
Running `garden plugins kubernetes cluster-init` results in:
```
All services already initialized!
```

This would be extremely useful when deploying django and node applications. 

Open an shell or execute a command on a running service without relying on kubectl.

When working in remote development setup, this is extremely necessary for debugging purposes.

Extend the existing Garden Exec command functionality to support services deployed with the Kubernetes module.

thsig:

Implemented `execInService` handlers for the `kubernetes` and `helm` module types.

These are very similar to the one used for `container` modules, differing mostly in the way the target workload is located.

Closes #2373.


eysi09 (approved):
LGTM but this code (and some other plugin code) is quite repetitive, down to the literal code commends.

PR #2390, opened 2021-05-13
improvement(core): better logs command

eysi09:

This PR makes some improvements to the logs stream. E.g.:

- More resilient logs follow mode
- Color code logs
- Add --since flag
- Interleave logs when not in follow mode

Kapture 2021-05-13 at 15 17 49

Fixes #

- The reason I did not use the K8s core API to follow logs is that it doesn't give you control over the request object. Without it, the connection tends to drop and there are some complaints about this on the K8s JS client repo. At the same time, I decided to continue using the core API in other places were we fetch logs as before, since in those case we just want the logs and the not the request object.
- When streaming logs that are more than ~100 lines, the request abruptly ends and the done callback is called. Thanks to the new implementation we reconnect after this, but some logs are dropped. I forked the K8s JS client library and tried getting to the bottom of this but without luck. I tried pausing and draining the request, and using the request implementation from the [Watch class](https://github.com/kubernetes-client/javascript/blob/2aa448fc5dc80ee2ecdda650ccf9c43537c81d0f/src/watch.ts#L57) but to no avail. Essentially [this line](https://github.com/kubernetes-client/javascript/blob/2aa448fc5dc80ee2ecdda650ccf9c43537c81d0f/src/log.ts#L86) gets called and the connection is closed. I honestly don't know how to fix this, but when streaming you generally don't care about past logs. So perhaps we should just set a hard limit on the number of lines that are fetched before streaming starts (currently the limit is 10 minutes). 
 - When in follow mode, we don't interleave the first logs that are fetched (because we stream from each service at a time) but subsequent logs are of course interleaved. This is only pronounced if a lot of logs are fetched before streaming starts.


core/src/commands/logs.ts
```diff
@@ -29,35 +32,66 @@ const logsArgs = {
...
 
-        garden logs               # prints latest logs from all services
-        garden logs my-service    # prints latest logs for my-service
-        garden logs -t            # keeps running and streams all incoming logs to the console
+        garden logs                       # interleaves color coded logs from all services (up to a certain limit)
+        garden logs --since 2d            # interleaves color coded logs from all services from the last 2 days
+        garden logs --tail 100            # interleaves the last 100 log lines from all services
+        garden logs service-a,service-b   # interleaves color coded logs for service-a and service-b
+        garden logs --follow              # keeps running and streams all incoming logs to the console
+        garden logs --original-color      # interaleaves logs from all services and prints the original output color
```
edvald, line 94: Typo: interaleaves

core/src/commands/logs.ts
```diff
@@ -29,35 +32,66 @@ const logsArgs = {
...
+    Outputs logs for all or specified services, and optionally waits for news logs to come in. Defaults
+    to getting logs from the last 10 minutes when in \`--follow\` mode. You can change this with the \`--since\` option.
 
     Examples:
 
-        garden logs               # prints latest logs from all services
-        garden logs my-service    # prints latest logs for my-service
-        garden logs -t            # keeps running and streams all incoming logs to the console
+        garden logs                       # interleaves color coded logs from all services (up to a certain limit)
+        garden logs --since 2d            # interleaves color coded logs from all services from the last 2 days
```
edvald, line 90: Typo: "color coded" -> "color-coded"

core/src/commands/logs.ts
```diff
@@ -72,61 +106,109 @@ export class LogsCommand extends Command<Args, Opts> {
...
+      if (showContainer && entry.containerName) {
+        container = entry.containerName
+      }
+
+      let out = [
+        // Render section "manually" so that we can control the styles
+        hideService ? "" : sectionStyle(formatSection(entry.serviceName, maxServiceName)) + " → ",
+        container ? sectionStyle(container) + " → " : "",
+        timestamp ? style(timestamp) + " → " : "",
+        originalColor ? (hasAnsi(serviceLog) ? serviceLog : chalk.white(serviceLog)) : style(serviceLog),
```
edvald, line 175: This is very difficult to parse imo

core/src/cli/params.ts
```diff
@@ -32,6 +32,28 @@ export const OUTPUT_RENDERERS = {
...
+export const validDurationUnits = ["d", "h", "m", "s"]
+
+function splitDuration(duration: string) {
+  return duration
+    .trim()
+    .split(/([0-9]+)/)
+    .filter(Boolean)
+}
+
+export function convertDurationToSeconds(duration: string) {
```
edvald, line 44: This could really use some unit tests.

edvald: Or, just use something like this: https://www.npmjs.com/package/parse-duration

core/src/commands/logs.ts
```diff
@@ -29,35 +32,66 @@ const logsArgs = {
...
+  "tail": new IntegerParameter({
+    help: dedent`
+      Number of lines to show for each service. Defaults to showing all log lines (up to a certain limit). Takes precedence over
+      the \`--since\` flag if both are set.
+    `,
     alias: "t",
-    defaultValue: -1,
   }),
-  // TODO
-  // since: new MomentParameter({ help: "Retrieve logs from the specified point onwards" }),
```
edvald, line 43: This should just be removed, right?

core/src/commands/logs.ts
```diff
@@ -72,61 +106,109 @@ export class LogsCommand extends Command<Args, Opts> {
...
 
-    void stream.forEach((entry) => {
-      // TODO: color each service differently for easier visual parsing
-      let timestamp = "                        "
-
-      // bad timestamp values can cause crash if not caught
-      if (entry.timestamp) {
+    log.info("")
+    log.info(chalk.white("Service logs:"))
+    log.info(chalk.white("-------------------------------------"))
```
edvald, line 135: This doesn't quite look nice imo. I suggest using this: https://github.com/garden-io/garden/blob/65303d41b0f132393e344d080f0fa2b8dfb088b7/core/src/logger/util.ts#L210

Also, maybe make the divider and the text above it bold?

core/src/plugins/kubernetes/logs.ts
```diff
@@ -112,86 +122,320 @@ async function readLogsFromApi({
...
+    this.connections[connectionId] = {
+      ...conn,
+      status,
     }
-    let timestamp: Date | undefined = undefined
-    let msg: string
+
+    // There's no need to log the closed event that happens after an error event
+    if (!(prevStatus === "error" && status === "closed")) {
+      this.write({
```
edvald, line 228: Somehow this state didn't come up in my testing. When re-deploying I just got this as part of the logs:

`api    → error: code = Unknown desc = Error: No such container: 390fd1ba5565f66189a4e31d5f0872cd1979d57bfdfa6652abf64c2b1deca234`

Which is sort of fine. Do you know if that's something we're printing ourselves from our error handling, or if this just comes over the wire from k8s?

edvald (changes requested):
This is great! Solid improvement, and much needed. Had a few comments but overall looks like a good implementation.

core/src/plugins/kubernetes/logs.ts
eysi09: This came from over the wire.

core/src/commands/logs.ts
eysi09: The last part with the double ternary, or just the whole thing?

core/src/commands/logs.ts
eysi09: It is removed, it's red :)

core/src/commands/logs.ts
eysi09: Good suggestion, looks good imo:
Screenshot 2021-05-13 at 18 04 27

core/src/commands/logs.ts
edvald: image

core/src/plugins/kubernetes/logs.ts
eysi09: I've seen this happen on occasion, and on other occasions it's more graceful. For example here I delete a service mid log stream:

```
result     → Checking for updates { }...
result     → <Lost connection to container "result" in Pod "result-58fb4d74cb-c82nt". Reason: Request closed. Will retry in background... >
result     → <No running containers found for service. W {ill retry in 10s
```

eysi09:
Thanks for the review! I've addressed all the comments. 

There's one outstanding question: This thing with the initial request "ending" when streaming 100+ log lines in follow mode (see the 3rd bullet in "notes for reviewer" above). 

If for example I log a tick every second and then run `garden logs --follow --since 500s`, I get something like:

```
tick 1
tick 2
...
tick 100
tick 101 <--- Around here the connection ends. The logs follower attempts to reconnect but doesn't know the correct position so just gets the logs from the last 10 seconds
tick 489
tick 490
tick 491
...
```

Currently we default to `--since 10m` in follow mode. Perhaps we should rather default to `--tail 10` or similar. Or maybe even enforce it?

edvald:
One other suggestion: Could we bold the service name, and maybe make the timestamp gray (or something like that, haven't actually tried it myself) to make the different parts more visually distinct? Feels a little hard to parse the output atm.

eysi09:
Making the service name bold makes the rest of text feel a little greyed out somehow (it was italic previously):

Screenshot 2021-05-15 at 11 42 32

One thing that's a little annoying also is that we don't know the lengths of the container names. So we can't easily align them like we do services. Although we can adjust the alignment as new logs come in.

eysi09:
We could also pad the log lines so that they don't break under the service names and timestamps, but then copying them wouldn't be nice.

core/src/commands/logs.ts
```diff
@@ -72,61 +111,129 @@ export class LogsCommand extends Command<Args, Opts> {
...
+    const originalColor = opts["original-color"]
+    const showContainer = opts["show-container"]
+    const hideService = opts["hide-service"]
+
+    if (tail) {
+      // Tail takes precedence over since...
+      since = undefined
+    } else if (follow && !since) {
+      // ...but if tail is not set and we're in follow mode, we default to getting the most recent logs.
+      since = "10m"
```
edvald, line 126: I was just testing this and noticed I could get a _lot_ of logs since this is uncapped. Would it maybe make more sense to default to tail?

eysi09: Ah whoops, this is suppose to be "1m". It was originally "10m" but I decided to change it because of this very reason. 

I kind of like the idea of using a duration here, since the idea is you get new logs and just a little bit of context with it if applicable. As opposed to always getting, say, the last 10 lines which may be way old. 

But of course `--since 1m` can still turn up a lot of logs. So no strong opinions really either way.

edvald (approved):

Issue #2309, opened 2021-03-16
[FEATURE]: Configurable resource requests for container type services


noizwaves:
### Background / Motivation

We are currently running into issues when deploying preview environments to an EKS cluster that I think could be solved with [configurable resource requests](https://github.com/garden-io/garden/blob/master/core/src/plugins/kubernetes/container/deployment.ts#L348-L353).

### What should the user be able to do?

Specify resource requests per service in a container module type.

### Why do they want to do this? What problem does it solve?

Our deployments consist of 10s of pods, which when started can consume hundreds (up to a thousand) of MBs of memory. Currently, when Garden creates the kubernetes manifests, it assigns each pod a fixed resource request limit. For 15 pods, that amounts to ~1.5gb (15 * 100) of memory, and these pods are allocated. However, once the pods start, they will grow to consume ~ 4.5gb (15 * ~300) of memory.

We would like to be able to capture these known memory requirements as custom resource requests, so that kubernetes can distribute our pods across nodes and also trigger cluster auto scaling (when the request resources cannot be fulfilled by existing nodes).

The current values cause our deployments to cause a significant number of OOMs and pod relocations that are disruptive to users. Being able to specify custom values would help us prevent this.

### Suggested Implementation(s)

For a container module type, a service should have configurable resource requests (just like they can for [resource limits](https://docs.garden.io/reference/module-types/container#services-limits)):

```
services:
  requests:
    cpu: 1000
    memory: 1024
```

These values would override the defaults [used here](https://github.com/garden-io/garden/blob/master/core/src/plugins/kubernetes/container/deployment.ts#L348-L353).

### How important is this feature for you/your team?

🌵 Not having this feature makes using Garden painful

eysi09:
Hi @noizwaves 

You can of course work around this by using the `kubernetes` module type. And with module templates, you can prevent duplicating the config. [Here's our guide](https://docs.garden.io/using-garden/module-templates) on the topic. Do you think that would make sense for your use case?

That being said, the `container` module type is pretty popular and almost exclusively used with the K8s provider, so it's certainly tempting to expose more K8s primitives (and Docker Compose has a notion of resource reservations, so it does apply more broadly). I'll take this up with the team internally and circle back. In any case it's simple to implement.

eysi09:
Just to follow up, we've decided to add resource requests to the container module type. Thanks for the suggestion!

noizwaves:
Great, thanks for the update @eysi09 !

[...]

PR #2392, opened 2021-05-13
feat(container): custom min/max resources

### LINKED ISSUE #2317: [FEATURE]: Specify test and tasks mem/cpu resouces
### Background / Motivation

Improving PODs stability and performance across the whole cluster by tweaking POD Container resources more aligned to their actual usage to avoid incredibly high CPU load and the OS (not K8s mem limit) OOMKilling processes 

Set container resources on tasks, tests or any other container that runs on K8s as part of a garden command

Helps to autoscale cluster the required resources and capacity without causing the applications on the "developers gardens" to slow down crash, due to liveness probes failing or even taking down the `kubelet`, `containerd`, `coredns`... processes

Be able to configure the `resources` for `tests` and `tasks`... at least these are the ones we have identified so far

thsig:

Resource requests and limits can now be set for `container` services, tasks and tests via the new `cpu` and `memory` fields.

The `service.limits` field has now been deprecated, and should be removed in a future breaking release. Values set in `service.limits` are still applied when deploying `container` services, and take precedence over the max values set in the `cpu` and `memory` fields.

Closes #2317 and #2309.

### DIFF
--- a/core/src/plugins/container/config.ts
+++ b/core/src/plugins/container/config.ts
@@ -34,6 +34,17 @@ export const defaultContainerLimits: ServiceLimitSpec = {
   memory: 1024, // = 1024MB = 1GB
 }
 
+export const defaultContainerResources: ContainerResourcesSpec = {
+  cpu: {
+    min: 10,
+    max: 1000,
+  },
+  memory: {
+    min: 90, // This is the minimum in some clusters.
+    max: 1024,
+  },
+}
+
 export interface ContainerIngressSpec {
   annotations: Annotations
   linkUrl?: string
@@ -73,9 +84,23 @@ export interface ServiceHealthCheckSpec {
   livenessTimeoutSeconds?: number
 }
 
+/**
+ * DEPRECATED: Use `ContainerResourcesSpec` instead.
+ */
 export interface ServiceLimitSpec {
   cpu: number
   memory: number
+}
+
+export interface ContainerResourcesSpec {
+  cpu: {
+    min: number
+    max: number
+  }
+  memory: {
+    min: number
+    max: number
+  }
 }
 
 interface Annotations {
@@ -93,7 +118,9 @@ export interface ContainerServiceSpec extends CommonServiceSpec {
   healthCheck?: ServiceHealthCheckSpec
   hotReloadCommand?: string[]
   hotReloadArgs?: string[]
-  limits: ServiceLimitSpec
+  limits?: ServiceLimitSpec
+  cpu: ContainerResourcesSpec["cpu"]
+  memory: ContainerResourcesSpec["memory"]
   ports: ServicePortSpec[]
   replicas?: number
   volumes: ContainerVolumeSpec[]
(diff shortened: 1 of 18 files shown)


core/src/plugins/container/config.ts
```diff
@@ -34,6 +34,17 @@ export const defaultContainerLimits: ServiceLimitSpec = {
   memory: 1024, // = 1024MB = 1GB
 }
 
+export const defaultContainerResources: ContainerResourcesSpec = {
```
eysi09, line 37: Nice. Happy we decided to go for this spec (as opposed to K8s style).

eysi09: One question though, the workflow config still contains the `ServiceLimitSpec`, feels like we should be using the `ContainerResourcesSpec` there as well for consistency.

core/src/plugins/container/config.ts
```diff
@@ -73,11 +84,25 @@ export interface ServiceHealthCheckSpec {
...
 
+/**
+ * DEPRECATED: Use `ContainerResourcesSpec` instead.
+ */
 export interface ServiceLimitSpec {
   cpu: number
   memory: number
 }
 
+export interface ContainerResourcesSpec {
```
eysi09, line 95: Wondering if this should have a more generic name in case we want to use if for other workloads defined by other plugins in the future. But perhaps just cross that bridge when we get there.

eysi09 (approved):
Approving but left a couple of questions.

thsig:
One design question I wanted to raise here:

If specified, should we use `service.resources` as the defaults for `test.resources` / `task.resources` when they're not specified?

Currently, we're not setting resource requests/limits for the pod runner for `container` tests/tasks unless explicitly specified (before this PR, they weren't being set at all).

core/src/plugins/container/config.ts
thsig: Yeah, definitely.

I'd like to do that in a separate PR so that we can first make the required changes on the Cloud backend to ensure that the changed workflow config schema isn't rejected during registration.

core/src/plugins/container/config.ts
thsig: Yeah, I thought about that. `ResourcesSpec` sounded too generic, but "container" can be interpreted to refer to `container` modules or to actual containers (to which these resource settings are applied).

I think we should stick with this name for now. It's only used internally, so changing it would be a simple refactor without user-facing consequences.

edvald:
I don't think we should use the service resource specs as a default for tests and tasks, mostly because you can specify multiple services, and the resource profile of a running service is generally completely different to tests and tasks.

Did we consider skipping the `resources` key and just placing the fields one step higher? I'm a little worried about all the indentation here.

thsig:
> I don't think we should use the service resource specs as a default for tests and tasks, mostly because you can specify multiple services, and the resource profile of a running service is generally completely different to tests and tasks.

Yeah, agreed.

> Did we consider skipping the resources key and just placing the fields one step higher? I'm a little worried about all the indentation here.

Good question. I just went with the format from our informal internal discussion, but we can definitely change it here.

Here are a few options:
```yaml
# current implementation
resources:
  cpu:
    min: x
    max: x
  memory:
    min: x
    max: x

# option 1
resources:
  maxCpu: x # or cpuMax etc.
  minCpu: x
  maxMemory: x
  minMemory: x

# option 2
cpu: # At top level of service/test/task spec
  min: x
  max: x
memory:
  min: x
  max: x

# option 3 - same as 1, but at top level
maxCpu: x # or cpuMax etc.
minCpu: x
maxMemory: x
minMemory: x
```
... and, of course, feel free to make alternate suggestions.

edvald:
I quite like option 2 there, personally. We could even take it a notch further at some point and allow shorthands, e.g.

```yaml
cpu: "200-500"  # -> min: 200, max: 500
cpu: 500  # -> min: undefined/default, max: 500
```

And the same deal for memory.

thsig:
This has now been updated to use the following config format:
```yaml
cpu: # At top level of service/test/task spec
  min: x
  max: x
memory:
  min: x
  max: x
```

10ko:
Btw, I am very excited for this! 🥳

docs/reference/module-types/container.md
```diff
@@ -308,13 +308,27 @@ services:
     # deployed with hot reloading enabled.
     hotReloadArgs:
 
-    # Specify resource limits for the service.
-    limits:
       # The maximum amount of CPU the service can use, in millicpus (i.e. 1000 = 1 CPU)
-      cpu: 1000
+      cpu:
```
edvald, line 312: This looks wrong... right? :P I think we might need to set the deprecated flag here on the nested field (and on the `memory` field) as well as on the `limits` key itself.

thsig: Ah, good catch! I'll have a look.

PR #2398, opened 2021-05-18
fix(cloud): retry streaming on network error

thsig:

We now retry the `POST` requests made by `BufferedEventStream` if the request fails with a network- or timeout-related status code.

This should make log & event streaming more robust to unstable network conditions or during periods of high load.

Also added a few more status codes to the retry logic for the k8s API.

### DIFF
--- a/core/src/plugins/kubernetes/api.ts
+++ b/core/src/plugins/kubernetes/api.ts
@@ -10,9 +10,9 @@
 // tslint:disable-next-line:no-unused
 import { IncomingMessage } from "http"
 import { Agent } from "https"
+import { ReadStream } from "tty"
 import chalk from "chalk"
 import httpStatusCodes from "http-status-codes"
-import { ReadStream } from "tty"
 import {
   KubeConfig,
   V1Secret,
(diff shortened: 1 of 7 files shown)


core/src/plugins/kubernetes/api.ts
```diff
@@ -889,15 +889,17 @@ function handleRequestPromiseError(err: Error) {
 }
 
 /**
- * Helper function for retrying failed k8s API requests, using exponential backoff.
+ * Helper function for retrying failed HTTP requests, using exponential backoff.
```
eysi09, line 892: Wonder if there's much value in making this generic. Ideally we only have a single way to make HTTP requests (i.e. with got) and just use this in the special case of dealing with the K8s API. That way developers don't have to choose between different ways to do the same thing. 

Or are there other uses cases you had in mind for this function?

eysi09 (commented):
Had one comment but otherwise looks good!

core/src/plugins/kubernetes/api.ts
thsig: Haha yeah, good point. I made this generic in order to reuse it for retries in `EnterpriseApi`, not realising that `got` does its own retries until you pointed it out to me.

Since I'd already made it generic, I thought I'd leave it as is (in case we need to use retries for `request`-based flows for some reason).

Do you think I should revert this helper to the way it was?

eysi09: I honestly do. And to your point `request` is deprecated so we really shouldn't tempt our future selves with this otherwise nice functionality.

thsig: Ok, should be good now.

eysi09 (approved):

PR #2399, opened 2021-05-19
feat(k8s): resolve template strings in kubernetes module manifest files 

edvald:
Previously, you could only use template strings when inlining
manifests using the `manifests` field. Now, template strings are also
resolved in the files referenced in the `files` key.

Internally, we now make a `ctx.resolveTemplateStrings()` for all
plugin handlers, so this will be easy to do in other scenarios as well.

Also added a commit to allow referencing service or task versions via e.g.
`${runtime.services.my-service.version}` or
`${runtime.tasks.my-task.version}`.

### DIFF
--- a/core/src/plugins/kubernetes/kubernetes-module/config.ts
+++ b/core/src/plugins/kubernetes/kubernetes-module/config.ts
@@ -74,7 +74,7 @@ export const kubernetesModuleSpecSchema = () =>
           resolve template strings in any of the manifests.`
     ),
     files: joiSparseArray(joi.posixPath().subPathOnly()).description(
-      "POSIX-style paths to YAML files to load manifests from. Each can contain multiple manifests."
+      "POSIX-style paths to YAML files to load manifests from. Each can contain multiple manifests, and can include any Garden template strings, which will be resolved before applying the manifests."
     ),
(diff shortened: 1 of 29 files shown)


eysi09:
@tvollstaedt, this might be of interest to you.

core/src/plugins/kubernetes/kubernetes-module/config.ts
```diff
@@ -74,7 +74,7 @@ export const kubernetesModuleSpecSchema = () =>
           resolve template strings in any of the manifests.`
     ),
     files: joiSparseArray(joi.posixPath().subPathOnly()).description(
-      "POSIX-style paths to YAML files to load manifests from. Each can contain multiple manifests."
+      "POSIX-style paths to YAML files to load manifests from. Each can contain multiple manifests, and can include any Garden template string, which will be resolved before applying the manifests."
```
thsig, line 77: "template string" -> "template strings"

thsig (changes requested):
Looks great! Had one text-related comment there.

edvald:
Rebased and ready @thsig!

thsig (approved):

PR #2404, opened 2021-05-25
feat(cloud): add cloud workflow template context

thsig:

Added a `workflow` config context that is available when using triggered Garden Cloud / Garden Enterprise workflows,

These values will be undefined outside of our workflow runners, so they can be used in conjunction with conditionals in templates to work across different environments.

### DIFF
--- a/docs/reference/template-strings.md
+++ b/docs/reference/template-strings.md
@@ -271,6 +271,70 @@ Example:
 my-variable: ${local.usernameLowerCase}
 ```
 
+### `${workflow.*}`
+
+Context variables that are specific to the currently running Garden Cloud workflow. These values are only available for workflows triggered by Garden Cloud in response to VCS event (e.g. from GitHub or GitLab).
+
+| Type     |
+| -------- |
+| `object` |
+
+### `${workflow.runUid}`
+
+The globally unique UID of the workflow run.
+
+| Type     |
+| -------- |
+| `string` |
+
+Example:
+
+```yaml
+my-variable: ${workflow.runUid}
+```
+
+### `${workflow.runNumber}`
+
+The run number of this workflow. This is incremented across all workflows within a given project.
+
+| Type     |
+| -------- |
+| `number` |
+
+Example:
+
+```yaml
+my-variable: ${workflow.runNumber}
+```
+
+### `${workflow.repoName}`
+
+The name of the repo that was cloned to run this workflow
+
+| Type     |
+| -------- |
+| `string` |
+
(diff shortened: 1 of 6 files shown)


docs/reference/template-strings.md
```diff
@@ -251,6 +251,70 @@ Example:
 my-variable: ${local.usernameLowerCase}
 ```
 
+### `${workflow.*}`
```
eysi09, line 254: Couldn't this be misinterpreted as referring to workflows in general?

eysi09 (commented):
This great but I had one question on the context name.

docs/reference/template-strings.md
thsig: Yeah, good point—how about `${triggeredWorkflow}`’?

edvald: That feels a bit clumsy imo. Workflow singular seems alright to me tbh. Where we allow referencing e.g. modules, it's always with a plural prefix. Just as long as the workflow context is always available from within any workflow, just with fewer values set when not triggered by cloud/enterprise.

thsig: As things stand, no values are available/relevant here outside of triggered workflows.

Can we think of any that we want to include here? `${workflow.name}` maybe?

thsig: How do we want to proceed here?

edvald: Sorry completely missed the following discussion. Sure, add `${workflow.name}`, nice and easy.

eysi09: Yeah my apologies, the radio silence. I suppose the context name is fine but I think it's a little troublesome that we have this template context that is only available in certain environments. Kind of goes against the idea of being able to run workflows locally and not having to "wait for CI". Didn't really think of that part when first reviewing this. 

Here's a (lame) example:

```
kind: Workflow
name: create-ns
steps:
  - script: >
     kubectl create ns test-${local.env.PR_NUMBER}
```

You can trivially run this locally with `PR_NUMBER=999 garden run workflow create-ns` if it fails in CI. Whereas if you  used `${worfklow.pr-number}` you'd have to change the code before running the command locally.

Of course you can always have a fallback like `${workflow.pr-number || local.env.PR_NUMER}` but that kind of defeats the point of the template context and makes the whole thing more verbose.

Perhaps I'm being too pedantic here but it feels like we're kind of creating a footgun here for users and they'll end up having a harder time replicating CI results then before.

PR #2407, opened 2021-05-26
fix(container): issue with out-of-sync deployment-image-id template

edvald:
WIP, still getting feedback from user to validate

See https://kubernetes.slack.com/archives/CKM7CP8P9/p1621628538039900

### DIFF
--- a/core/src/plugins/container/config.ts
+++ b/core/src/plugins/container/config.ts
@@ -28,6 +28,7 @@ import { baseTestSpecSchema, BaseTestSpec } from "../../config/test"
 import { joiStringMap } from "../../config/common"
 import { dedent, deline } from "../../util/string"
 import { getModuleTypeUrl } from "../../docs/common"
+import { ContainerModuleOutputs } from "./container"
 
 export const defaultContainerLimits: ServiceLimitSpec = {
   cpu: 1000, // = 1000 millicpu = 1 CPU
@@ -729,5 +730,6 @@ export interface ContainerModule<
   M extends ContainerModuleSpec = ContainerModuleSpec,
   S extends ContainerServiceSpec = ContainerServiceSpec,
   T extends ContainerTestSpec = ContainerTestSpec,
-  W extends ContainerTaskSpec = ContainerTaskSpec
-> extends GardenModule<M, S, T, W> {}
+  W extends ContainerTaskSpec = ContainerTaskSpec,
+  O extends ContainerModuleOutputs = ContainerModuleOutputs
+> extends GardenModule<M, S, T, W, O> {}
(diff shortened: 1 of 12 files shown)


JoshDrake-minted:
I'm the original reporter of this issue, here's the full bug report template if you want to hoist it up to the issue description:

## Bug

### Current Behavior

Modules depending on `deployment-image-id` or `local-image-id` from container module outputs are not working since 0.12.20. The value for these outputs does not match that which was used to tag the image in the container module `build` stage.

### Expected behavior

Modules, such as `kubernetes`, should be able to reference the image tag from `container` modules they depend on, for example, to generate manifests that specify the `image`.

### Reproducible example

https://github.com/JoshDrake-minted/k8s-module-outputs-bug/

I've provided a very minimal repro case that uses remote clusters with in-cluster building, but I have observed this behavior with other build modes and `local-kubernetes` as well.

Simply run `garden dev` to observe the described behavior.

### Workaround

Using `extraFlags` in the container module to tag the image with a static tag name (eg: `dev`) is a potential workaround, though this is far from ideal.

### Suggested solution(s)

Pin the `deployment-image-id` and `local-image-id` container module outputs when the image is tagged during `build`. 

### Additional context

From what I can tell after having done quite a bit of debugging on my own now, is that the is basically a result of the changes in https://github.com/garden-io/garden/pull/2281, which introduced more granular versioning of modules. The behavior I'm seeing is that the module outputs for the container module (specifically, the `deployment-image-id`) change multiple times during module execution. This means (again, from what I have observed) that the image tag used during the container build, which is derived from the module name and version, simply uses the module version _at the time the build stage occurs_. However, after this build stage occurs, the module version is still changing, and indeed will likely be **different** than the version used to tag the container image.

This results in any other modules that are depending on the `<deployment/local>-image-id` output from a given container module, (eg: `kubernetes` modules), to potentially end up referencing a module version that was not used to tag the image. This obviously results in the pods being unable to pull the correct image, because the specified tag does not exist.

### Your environment


* Linux, Manjaro
* GKE, but also observed with minikube


`garden version`

0.12.21 (but also observed on latest master built locally, and with this patch applied)

edvald:
I think I've figured this out @JoshDrake-minted! The issue is that the `manifests.yml` file in your kubernetes module is being included in the container module. So when the k8s module is resolved, which happens _after_ the container module is first resolved, it changes the version of the container module in the process.

I'm not quite sure if there's a sane way to outright prevent this at the framework level, but I'll look into it. In any case, the fix for your particular issue is to make sure the generated file is excluded in the container module.

edvald:
This btw also explains why my own experiments didn't reproduce this, because I wasn't including all the files in the directory, just the specific files that were supposed to go in the container.

PR #2408, opened 2021-05-26
Logger refactor

eysi09:

With this change, the logger only stores log entries on the logger
instance when the FancyWriter is enabled or when running tests. This
prevents large log data accumulating in-memory.

Also refactored the logger a litle bit so that it no longer extends the
log node class but rather implements it. It simplifies the relationship
between the root logger and the log entries.

Fixes #


thsig (changes requested):
Looks great! Only had one minor text-related comment there.

thsig (approved):

PR #2413, opened 2021-05-27
improvement(cloud): better secrets/API errors

thsig:

We now log more informative error messages when a missing secret is referenced, and when trying to fetch secrets for an environment that doesn't exist on the backend.

We also don't throw errors when fetching secrets for an environment that doesn't exist (but we do log additional information around this at the `debug` log level).

In addition to better handling of secret-related errors, this PR improves on the following:
* We now throw an error when a null client auth token / jwt is received from the backend during login. Previously, we were only logging an error, without indicating that the command had failed. The error message was also a bit cryptic.
* When workflow run registration fails due to CLI/backend API incompatibility (i.e. if the shape of the registration request isn't what the API expects), throw a better error, and include the failed request payload in the error detail.

### DIFF
--- a/core/src/config/template-contexts/project.ts
+++ b/core/src/config/template-contexts/project.ts
@@ -6,10 +6,12 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+import { last, isEmpty } from "lodash"
+import chalk from "chalk"
 import { PrimitiveMap, joiIdentifierMap, joiStringMap, joiPrimitive, DeepPrimitiveMap, joiVariables } from "../common"
 import { joi } from "../common"
 import { deline, dedent } from "../../util/string"
-import { schema, ConfigContext } from "./base"
+import { schema, ConfigContext, ContextKeySegment } from "./base"
 import { CommandInfo } from "../../plugin-context"
 
 class LocalContext extends ConfigContext {
@@ -200,7 +202,9 @@ export class DefaultEnvironmentContext extends ConfigContext {
 }
 
 export interface ProjectConfigContextParams extends DefaultEnvironmentContextParams {
+  loggedIn: boolean
   secrets: PrimitiveMap
+  enterpriseDomain: string | undefined
 }
 
 /**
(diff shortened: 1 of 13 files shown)


core/src/config/template-contexts/project.ts
```diff
@@ -218,9 +221,37 @@ export class ProjectConfigContext extends DefaultEnvironmentContext {
...
+      return dedent`
+        You are not logged in to Garden Enterprise, but one or more secrets are referenced in template strings in your Garden configuration files.
+
+        Please log in via the ${chalk.green("garden login")} command to use Garden with secrets.
+      `
+    }
+
+    if (size(this.secrets) === 0) {
+      return deline`
+        Note: An empty set of secrets was fetched. If you have defined secrets for the current project and environment in Garden
```
eysi09, line 241: Isn't it most likely that this happens simply when the secrets haven't been created at all? If that's the case, shouldn't the suggestion be similar to below? Maybe with a bit more colour since it looks the user hasn't created secrets at all.

Something like: "Looks like no secrets have been created for this environment in Garden Cloud. To create secrets click the following link ${linkToProjectSecretsPageForGivenEnterpriseDomain}`.

core/src/enterprise/api.ts
```diff
@@ -182,6 +186,12 @@ export class EnterpriseApi {
   }
 
   static async saveAuthToken(log: LogEntry, tokenResponse: AuthTokenResponse) {
+    if (!tokenResponse.token) {
+      throw new EnterpriseApiError(
+        `Received a null/empty client auth token while logging in. Please contact your system administrator.`,
```
eysi09, line 191: "Please contact your system administrator" is probably my least favourite error message. Especially when there's a chance that the person seeing this message is the system administrator. 

When is this likely to happen? Can give more actionable suggestions based on that?

eysi09 (changes requested):
Had few comments/questions/suggestions on the text but overall looks great.

core/src/enterprise/api.ts
thsig: > "Please contact your system administrator" is probably my least favourite error message. Especially when there's a chance that the person seeing this message is the system administrator.

Haha yeah, true that.

> When is this likely to happen? Can give more actionable suggestions based on that?

How about: "This may indicate a problem with your VCS provider setup."?

thsig:
I think this should be good to go now, @eysi09.

If we're OK with this, I'll fold the TBS commit into the first one before merging.

PR #2435, opened 2021-06-11
fix(k8s): rsync with destination-default perms

thsig:

This PR is an attempt to fix an issue some Windows users have encountered, where the build sync process would run into `rsync` errors because of permission errors.

### DIFF
--- a/core/src/plugins/kubernetes/container/build/common.ts
+++ b/core/src/plugins/kubernetes/container/build/common.ts
@@ -44,11 +44,10 @@ export const commonSyncArgs = [
   "--recursive",
   // Copy symlinks (Note: These are sanitized while syncing to the build staging dir)
   "--links",
-  // Preserve permissions
-  "--perms",
   // Preserve modification times
   "--times",
   "--compress",
+  "--executability",
 ]
 
 export const builderToleration = {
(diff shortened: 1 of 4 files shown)


core/src/plugins/kubernetes/container/build/common.ts
```diff
@@ -44,8 +44,9 @@ export const commonSyncArgs = [
   "--recursive",
   // Copy symlinks (Note: These are sanitized while syncing to the build staging dir)
   "--links",
-  // Preserve permissions
-  "--perms",
+  // Use destination-default permissions
+  "--no-perms",
+  "--chown=username:group",
```
edvald, line 49: This doesn't look right... Doesn't this need to match what's in the rsync container?

core/src/plugins/kubernetes/hot-reload/helpers.ts
```diff
@@ -297,8 +298,7 @@ export async function syncToService({ ctx, service, hotReloadSpec, namespace, wo
         // Preserve owner + group
         "--owner",
         "--group",
```
edvald, line 300: This is incompatible with the `--chown` flag

PR #2438, opened 2021-06-15
feat(core): allow variables in remote sources

### LINKED ISSUE #2436: [FEATURE]: Allow Remote Sources/Modules to use CLI, project variables and env file variables to construct repositoryUrl
### Background / Motivation

We need to construct the repositoryUrl property when using remote sources/modules based on values provided at runtime. Garden currently only reads in variables set using environment variables. it would be nice to set these variables using the CLI or from a file.

set variables using either project variables, env file, or CLI to set and have them available in  repositoryUrl

The need to create variables using environment variables.

thsig:

Template strings in remote sources now have access to the same variables as provider configs.

This adds flexibility when using remote sources in conjunction with automation (e.g. when using dynamic Git branches / tags for remote sources, depending on project- and environment-level variables).

Fixes #2436.

Previously, referencing `${var.*}` in remote sources would result in a template string error, since remote sources were being resolved at the same time as the top-level project config (which is before project- and environment-level variables are resolved).

We now only resolve template strings in remote sources on demand, when `getProjectSources` are called. Is this a good approach, or should we do this as part of the main init flow of the `Garden` class?

### DIFF
--- a/static/docs/templates/template-strings.hbs
+++ b/static/docs/templates/template-strings.hbs
@@ -29,6 +29,11 @@ Examples:
 The following keys are available in any template strings within project definitions in `garden.yml` config files, except the `name` field (which cannot be templated). See the [Environment](#environment-configuration-context) and [Provider](#provider-configuration-context) sections below for additional keys available when configuring `environments` and `providers`, respectively.
 {{{projectContext}}}
 
+## Remote source configuration context
+
+The following keys are available in template strings under the `sources` key in project configs.
+{{{remoteSourceContext}}}
+
 ## Environment configuration context
 
 The following keys are available in template strings under the `environments` key in project configs. Additional keys are available for the `environments[].providers` field, see the [Provider](#provider-configuration-context) section below for those.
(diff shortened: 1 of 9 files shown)


thsig:
Any thoughts on this one, @edvald?

static/docs/templates/template-strings.hbs
```diff
@@ -29,14 +29,19 @@ Examples:
 The following keys are available in any template strings within project definitions in `garden.yml` config files, except the `name` field (which cannot be templated). See the [Environment](#environment-configuration-context) and [Provider](#provider-configuration-context) sections below for additional keys available when configuring `environments` and `providers`, respectively.
 {{{projectContext}}}
 
+## Remote source configuration context
```
edvald, line 32: This file is getting quite large, and it's getting increasingly hard to navigate what's supported where. I wonder how we might best address that?

thsig: Hmm, maybe we need to break it up into a few pages?

That said, it's relatively straightforward to find the supported variables for a given context via the nav bar on the right.

What do you think? CC @eysi09

thsig: Most of the size/length of this page comes from the repetition of the keys under each template variable.

One approach would be to "normalize" this, only listing the contexts that are available but not all keys for all contexts.

For example, only list the keys available for `local` in one place, and then simply list `local`, `project`, `git` and `secrets` for the project configuration context.

Each of those (`local`, `project`, `git` and `secrets`) would then be within-page hyperlinks to the section that documents the keys and descriptions.

How does that sound?

edvald (approved):

PR #2440, opened 2021-06-16
chore: add human readable strings to anonymized IDs

eysi09:

This PR introduces a v2 version of anonymized analytics properties. This
version includes a human readable string in the ID so that they are
easier to recognize.

Fixes #

### DIFF
--- a/core/src/analytics/analytics.ts
+++ b/core/src/analytics/analytics.ts
@@ -6,6 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+import codenamize = require("@codenamize/codenamize")
 import segmentClient = require("analytics-node")
 import { platform, release } from "os"
 import ci = require("ci-info")
@@ -55,9 +56,13 @@ interface ProjectMetadata {
 
 interface AnalyticsEventProperties {
   projectId: string
+  projectIdV2: string
   projectName: string
+  projectNameV2: string
   enterpriseProjectId?: string
+  enterpriseProjectIdV2?: string
   enterpriseDomain?: string
+  enterpriseDomainV2?: string
   isLoggedIn: boolean
   ciName: string | null
   system: SystemInfo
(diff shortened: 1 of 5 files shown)


core/src/analytics/analytics.ts
```diff
@@ -249,6 +273,20 @@ export class AnalyticsHandler {
...
+   * This readable string consists of two adjectives, followed by a noun.
+   *
+   * Also truncates the hash part to 32 char (from 128).
+   *
+   * Example: dysfunctional-graceful-request_433c84996726070996f369dfc00dd202
+   */
+  public hashV2(val: string) {
+    const readable = codenamize({ seed: val, adjectiveCount: 2 })
+    const hash = this.hash(val).slice(0, 32)
+    return `${readable}_${hash}`
```
eysi09, line 287: I picked underscore as a delimiter since codenamize uses hyphens.

edvald:
Feature request: Do something similar for Garden version hashes :)

eysi09:
@edvald, I updated the PR to also add a readable user ID as "trait" on the user: https://github.com/garden-io/garden/pull/2440/files#diff-8278b2fcc5bc5afb7cf477c1b2067e9861738d0e44d81bc1334ab3b3e849d299R264

core/src/analytics/analytics.ts
```diff
@@ -294,9 +335,13 @@ export class AnalyticsHandler {
   private getBasicAnalyticsProperties(): AnalyticsEventProperties {
     return {
       projectId: this.projectId,
+      projectIdV2: this.projectIdV2,
       projectName: this.projectName,
+      projectNameV2: this.projectNameV2,
       enterpriseProjectId: this.enterpriseProjectId,
+      enterpriseProjectIdV2: this.enterpriseProjectIdV2,
```
edvald, line 342: Do we still want to use the "enterprise" prefix here, since we're changing the field anyway?

edvald (approved):
Had one (open-ended) question, but LGTM 👍

core/src/analytics/analytics.ts
eysi09: Yeah, that's a good question. Don't have a strong opinion. Would we then go for `cloudProjectId` or `cloudProjectIdV2`?

edvald: I'd say V2 just for consistency, even if it is technically redundant.

edvald: @eysi09 reminder on ☝️

PR #2442, opened 2021-06-17
improvement(core): add get tests command

eysi09:

I noticed this command was missing when I needed it so I added it. Uses the same output format as the `get tasks` command.

Fixes #

### DIFF
--- a/core/src/commands/get/get-tasks.ts
+++ b/core/src/commands/get/get-tasks.ts
@@ -6,13 +6,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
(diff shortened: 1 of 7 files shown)


thsig:
Nice and simple, good stuff.

I do think that the output format looks rather weird (which is my bad—I'm responsible for how the `get tasks` output looks).

For now, I suggest we make the colons after the keys the same dark-grey color as the keys themselves (as is, the `:` sticks out too much visually).

eysi09:
That's good feedback, I'll update the PR. I played around with table mode when writing this but each cell would just contain too much content so I stuck with the original.

thsig (approved):

PR #2444, opened 2021-06-18
feat(core): local targets for build dependencies

thsig:

Added an optional `copyToSourceDir` flag to `build.dependencies[].copy[]`.

When used, build dependencies are copied to the module (source) directory of the dependant module instead of the build directory.

This is useful for many flows, especially when using the `exec` module for custom build flows, and has been indirectly requested by many of our users.

### DIFF
--- a/core/src/build-staging/build-staging.ts
+++ b/core/src/build-staging/build-staging.ts
@@ -10,10 +10,10 @@ import Bluebird from "bluebird"
 import { isAbsolute, join, resolve, relative, parse, basename } from "path"
 import { emptyDir, ensureDir, remove } from "fs-extra"
 import { ConfigurationError, InternalError } from "../exceptions"
-import { FileCopySpec, GardenModule, getModuleKey } from "../types/module"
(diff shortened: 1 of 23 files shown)


edvald:
Seems good at a glance, although I'm unsure about the `localTarget` name. When seeing it next to `target`, it doesn't feel clear that this will be a boolean flag, and feels opaque. How about something a little more verbose (and helpful), like `copyToSourceDir`?

thsig:
Yeah, good point—`copyToSourceDir` is clearer. Should be good to go now.

thsig:
How do we feel about this one? Any further comments?

edvald:
I think we need more complete tests for this. What we need to test for is whether the files are copied _before_ the module version is computed. If it isn't, this won't have the desired effect, and may trigger some awkward version re-computes mid-process (which is a big part of why we stage builds in the first place).

edvald (changes requested):
See comment above.

PR #2448, opened 2021-06-22
fix(k8s): omit probes in runner pod spec

thsig:

We now no longer include liveness or readiness probes on the pod spec used when running e.g. tests or tasks for `kubernetes` or `helm` modules.

In certain situations, these probes could result in test pods being evicted before the test suite was able to finish. This is fixed here.

### DIFF
--- a/core/src/plugins/kubernetes/run.ts
+++ b/core/src/plugins/kubernetes/run.ts
@@ -9,6 +9,7 @@
 import { resolve } from "path"
 import tar from "tar"
 import tmp from "tmp-promise"
+import { cloneDeep, omit, pick } from "lodash"
 import { V1PodSpec, V1Pod, V1Container } from "@kubernetes/client-node"
 import { RunResult } from "../../types/plugin/base"
 import { GardenModule } from "../../types/module"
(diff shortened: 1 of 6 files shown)


core/src/plugins/kubernetes/run.ts
```diff
@@ -88,6 +91,8 @@ export const runPodSpecWhitelist: (keyof V1PodSpec)[] = [
   "volumes",
 ]
 
+export const runContainerBlacklist: (keyof V1Container)[] = ["readinessProbe", "livenessProbe"]
```
edvald, line 94: I figure while we're at it, we might consider this: https://english.stackexchange.com/questions/51088/alternative-terms-to-blacklist-and-whitelist

I'd say for the context, "include fields" and "exclude fields" would be good for the naming.

thsig: Ah, good point. Hadn't thought of that.

thsig: This should be good to go now.

edvald: 👍✨

And yeah of course, it's just a subtle thing. Once pointed out to me, my eye can't help catch it :P

edvald (approved):

PR #2449, opened 2021-06-23
fix(k8s): logs command wouldn't get logs from all pods

edvald:
Ran into this when working on a different feature. Simple fix, but a somewhat embarrassing old bug 😬


eysi09 (approved):

eysi09:
Approved. But I feel like there's a reason it's there in the first place.

PR #2451, opened 2021-06-23
feat(core): emit live logs from k8s tasks and tests (WIP)

edvald:


core/package.json
```diff
@@ -66,6 +66,7 @@
     "env-var": "^6.2.0",
     "escape-string-regexp": "^4.0.0",
     "eventemitter2": "^6.4.1",
+    "eventemitter3": "^4.0.7",
```
eysi09, line 69: Should we remove the dependencies on `eventemitter2` then so that we always use the same one?

edvald: It'd make sense imo, just didn't have time to do it myself.

thsig:
Closing in favor of #2472.

PR #2454, opened 2021-06-24
feat(config): add support for module-level variables

edvald:
This introduces _module variables_ to help make configurations cleaner
and easier to manage, especially for larger projects.

From the added docs:

--

## Module variables

Each Garden module can specify its own set of variables, that can be
re-used within the module, as well as referenced by other dependant
modules.

Simply specify the `variables` field on the module, same as in the
project configuration. For example:

```yaml
# my-service/garden.yml
kind: Module
name: my-service
variables:
  # This overrides the project-level hostname variable
  hostname: my-service.${var.hostname}
  # You can specify maps or lists as variables
  envVars:
    LOG_LEVEL: debug
    DATABASE_PASSWORD: ${var.database-password}
services:
  - name: my-service
    ...
    ingresses:
      - path: /
        port: http
        # This resolved to the hostname variable set above,
        # not the project-level one
        hostname: ${var.hostname}
    # Referencing the above envVar module variable
    env: ${var.envVars}
tests:
  - name: my-test
    ...
    # Re-using the envVar module variable
    env: ${var.envVars}
```

Notice that you can override variables defined at the project-level,
and even reference project-scoped variables when defining the module
variables.

Also notice the generally handy use-case of re-using a common value (in
this case a map of environment variables) in multiple spots in the
module configuration.

### Referencing module variables

On top of that, you can reference the resolved module variables in
other modules. With the above example, another module might for example
reference `${modules.my-service.var.hostname}`. For larger projects
this can be much cleaner than, say, hoisting a lot of variables up to
the project level.

Please give it a quick spin, see if it works alright outside of unit/functional tests.

### DIFF
--- a/core/src/config/module.ts
+++ b/core/src/config/module.ts
@@ -86,6 +86,7 @@ interface ModuleSpecCommon {
   path?: string
   repositoryUrl?: string
   type: string
+  variables?: DeepPrimitiveMap
 }
 
 export interface AddModuleSpec extends ModuleSpecCommon {
@@ -209,6 +210,9 @@ export const baseModuleSpecKeys = () => ({
     .description("When false, disables pushing this module to remote registries."),
   generateFiles: joiSparseArray(generatedFileSchema()).description(dedent`
     A list of files to write to the module directory when resolving this module. This is useful to automatically generate (and template) any supporting files needed for the module.
+  `),
+  variables: joiVariables().default(() => undefined).description(dedent`
+    A map of variables scoped to this particular module. These are resolved before any other parts of the module configuration and take precedence over project-scoped variables. They may reference project-scoped variables, and generally use any template strings normally allowed when resolving modules.
   `),
 })
 
--- a/docs/using-garden/variables-and-templating.md
+++ b/docs/using-garden/variables-and-templating.md
@@ -355,6 +355,10 @@ The order of precedence across the varfiles and project config fields is as foll
 4. Configured project-wide varfile (defaults to `garden.env`).
 5. The project-wide `variables` field.
 
+{% hint style="warning" %}
+Note that [Module variables](#module-variables) always take precedence over any of the above, in the context of the module being resolved.
+{% endhint %}
+
 When you specify variables in multiple places, we merge the different objects and files using a [JSON Merge Patch](https://tools.ietf.org/html/rfc7396).
 
 Here's an example, where we have some project variables defined in our project config, and environment-specific values—including secret data—in varfiles:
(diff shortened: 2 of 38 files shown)


core/src/config/module.ts
```diff
@@ -210,6 +211,9 @@ export const baseModuleSpecKeys = () => ({
   generateFiles: joiSparseArray(generatedFileSchema()).description(dedent`
     A list of files to write to the module directory when resolving this module. This is useful to automatically generate (and template) any supporting files needed for the module.
   `),
+  variables: joiVariables().default(() => undefined).description(dedent`
```
thsig, line 214: Minor question: Is there a reason we don't default to `{}` here?

docs/using-garden/variables-and-templating.md
```diff
@@ -390,6 +394,48 @@ services:
       DATABASE_PASSWORD: ${var.database-password}
 ```
 
+## Module variables
+
+Each Garden module can specify its own set of variables, that can be re-used within the module, as well as referenced by other dependant modules.
```
thsig, line 399: "set of variables, that" -> "set of variables that"

Also, for added clarity, I think we should expand this: "as well as referenced by other dependant modules"

to something like "The variables defined in a module can be reused in modules that have build or runtime dependencies on it."

Btw, are those the semantics? Just making sure I read the resolution flow right.

Also, if a user makes a mistake there (e.g. expects a variable to be available, but didn't get the dependency structure right), will they get an informative error message (or will the origin of the mistake generally be obvious from the usage context)?

docs/using-garden/variables-and-templating.md
```diff
@@ -390,6 +394,48 @@ services:
...
+  envVars:
+    LOG_LEVEL: debug
+    DATABASE_PASSWORD: ${var.database-password}
+services:
+  - name: my-service
+    ...
+    ingresses:
+      - path: /
+        port: http
+        # This resolved to the hostname variable set above, not the project-level one
```
thsig, line 420: "resolved" -> "resolves"

"not the project level one" -> "not the project-level hostname variable"

thsig (changes requested):
Looks great! This will be a very welcome feature for our users.

Only had a few test/documentation-related comments there.

core/src/config/module.ts
edvald: This is to avoid introducing unnecessary changes to version hashes, and breaking a bunch of tests along the way.

docs/using-garden/variables-and-templating.md
edvald: Actually that's not the case—a module can implicitly depend on another by way of template string reference without including an explicit build/runtime reference. So I think this copy (save for the typo) is fine.

thsig (approved):

PR #2465, opened 2021-07-01
fix(core): inconsistency in include/exclude handling for submodules

edvald:
Normally, specifying a directory path (without a trailing `**/*`) in an
`include` or `exclude` statement would include/exclude the whole
directory contents. This was not the case for submodule directories (or
directories within submodules), which could cause issues and confusion.
We now explicitly make sure that the behavior is consistent.

Also made a small performance improvement by adding `micromatch`
which is optimized for the type of glob matching we're doing.


core/src/vcs/git.ts
```diff
@@ -107,12 +107,21 @@ export class GitHandler extends VcsHandler {
...
+    include,
+    exclude,
+    filter,
+  }: GetFilesParams): Promise<VcsFile[]> {
     if (include && include.length === 0) {
       // No need to proceed, nothing should be included
       return []
     }
 
+    log.debug(`Scanning ${pathDescription} at ${path}\nIncludes: ${include}\nExcludes:${exclude}`)
```
thsig, line 123: Maybe this should be at the `silly` level (like the other new log statement at line 160 below)?

core/src/vcs/git.ts
```diff
@@ -510,3 +531,21 @@ const notInRepoRootErrorMessage = (path: string) => deline`
     Please run \`git init\` if you're starting a new project and repository, or move the project to an
     existing repository, and try again.
   `
+
+/**
+ * Given a list of POSIX-style globs
```
thsig, line 536: Unfinished comment?

Also, maybe it would be cleaner to explicitly return `undefined` on line 540 below.

core/test/unit/src/vcs/git.ts
```diff
@@ -262,6 +262,19 @@ describe("GitHandler", () => {
       expect(await handler.getFiles({ path: tmpPath, include: ["foo.*"], exclude: [], log })).to.eql([{ path, hash }])
     })
 
+    it("should include a directory that's explicitly included its exact name", async () => {
```
thsig, line 265: "included its" -> "included by its"

thsig (changes requested):
Looks good! Had a few minor comments there.

core/src/vcs/git.ts
edvald: Felt like this could be at a higher level, because it's not called as frequently and is broadly useful to know when debugging.

edvald:
@thsig comments addressed (and linting errors fixed)

thsig (approved):

edvald:
The failing unit test seems to be down to a behavior difference in git versions. Which is annoying. I'm seeing about how best to fix it.

PR #2470, opened 2021-07-07
feat(core): stream build statuses

thsig:

We now stream build status events, similarly to how we've been streaming service, test and task status events.

We've also added a `running` state to the `RunState` type, and we include a `building` state in the new `BuildState` type to represent builds/tests/tasks that are in progress.

Additionally, we now emit a `serviceStatus` event with the `deploying` state in the `deployService` action just before the deploy handler is called.

### DIFF
--- a/core/src/actions.ts
+++ b/core/src/actions.ts
@@ -32,7 +32,7 @@ import {
   RunResult,
   runStatus,
 } from "./types/plugin/base"
-import { BuildModuleParams, BuildResult } from "./types/plugin/module/build"
+import { BuildModuleParams, BuildResult, BuildState } from "./types/plugin/module/build"
 import { BuildStatus, GetBuildStatusParams } from "./types/plugin/module/getBuildStatus"
 import { GetTestResultParams, TestResult } from "./types/plugin/module/getTestResult"
 import { RunModuleParams } from "./types/plugin/module/runModule"
@@ -82,7 +82,7 @@ import { RunServiceParams } from "./types/plugin/service/runService"
 import { GetTaskResultParams } from "./types/plugin/task/getTaskResult"
 import { RunTaskParams, RunTaskResult } from "./types/plugin/task/runTask"
 import { ServiceStatus, ServiceStatusMap, ServiceState, GardenService } from "./types/service"
-import { Omit, getNames } from "./util/util"
+import { Omit, getNames, uuidv4 } from "./util/util"
 import { DebugInfoMap } from "./types/plugin/provider/getDebugInfo"
 import { PrepareEnvironmentParams, PrepareEnvironmentResult } from "./types/plugin/provider/prepareEnvironment"
 import { GetPortForwardParams } from "./types/plugin/service/getPortForward"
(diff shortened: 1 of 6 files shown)


edvald (approved):

core/src/types/plugin/module/build.ts
```diff
@@ -13,6 +13,13 @@ import { joi } from "../../../config/common"
 
 export interface BuildModuleParams<T extends GardenModule = GardenModule> extends PluginModuleActionParamsBase<T> {}
 
+/**
+ * - `fetched`: The build was fetched from a remote repository instead of building.
+ * - `built`: The module was built.
+ * - `error`: Fetching or building the module failed.
+ */
+export type BuildState = "fetched" | "building" | "built" | "error"
```
edvald, line 21: Maybe "failed" is more consistent in terms of phrasing, rather than "error"?

thsig: Yeah, that's better.

thsig:
@edvald: I've made the changes we talked about, could you have another look?

core/src/actions.ts
```diff
@@ -386,12 +427,28 @@ export class ActionRouter implements TypeGuard {
...
+    const actionUid = uuidv4()
+    const testName = params.test.name
+    const testConfigVersion = params.test.version
+    const moduleName = params.module.name
+    const moduleVersion = params.module.version.versionString
+    this.garden.events.emit("testStatus", {
+      testName,
+      moduleName,
+      moduleVersion,
+      testConfigVersion,
```
edvald, line 439: I think it's clearer to just call this `testVersion`. The test version isn't _just_ the test config version, but rather the whole chain including the module version, dependencies etc.

core/src/actions.ts
```diff
@@ -444,9 +506,23 @@ export class ActionRouter implements TypeGuard {
...
 
   async deployService(params: ServiceActionRouterParams<DeployServiceParams>): Promise<ServiceStatus> {
+    const actionUid = uuidv4()
+    const serviceName = params.service.name
+    const moduleVersion = params.service.module.version.versionString
+    const serviceConfigVersion = params.service.version
+    this.garden.events.emit("serviceStatus", {
+      serviceName,
+      moduleVersion,
+      serviceConfigVersion,
```
edvald, line 516: Same goes here, just call this `serviceVersion` imo.

core/src/actions.ts
```diff
@@ -550,13 +626,27 @@ export class ActionRouter implements TypeGuard {
...
+    const actionUid = uuidv4()
     const tmpDir = await tmp.dir({ unsafeCleanup: true })
     const artifactsPath = normalizePath(await realpath(tmpDir.path))
+    const taskName = params.task.name
+    const taskConfigVersion = params.task.version
+    const moduleVersion = params.task.module.version.versionString
+    this.garden.events.emit("taskStatus", {
+      taskName,
+      moduleVersion,
+      taskConfigVersion,
```
edvald, line 638: And here as well, of course .)

edvald (changes requested):
Had some field naming comments, otherwise looks good!

thsig:
Thanks for the suggestions! The fields have been renamed now.

edvald (approved):

thsig:
@10ko: I've added `moduleName` to the `serviceStatus` and `taskStatus` events as you requested. Will merge this once CI is passing.

Issue #2478, opened 2021-07-12
[FEATURE]: Smarter logs alignment


eysi09:
### Background / Motivation

Long section names are truncated and hard to ready with the log output. We could probably fix this by aligning the section dynamically based on width. This is what we already do for service logs when running the `garden logs` command.

Something like:

```
✔ short → Getting status 
✔ short → Getting status
✔ somewhat-longer → Getting status
✔ short           → Getting status
✔ very-very-very-long → Getting status
✔ short               → Getting status 
✔ somewhat-longer     → Getting status
```

We could still have some sensible max width and perhaps make that configureable.

eysi09:
Thoughts @edvald?

edvald:
See my comment on #2479. I think it's fine to pad to a certain length and then not trim from there. But sure, a dynamic section width might be a nice improvement if you're feeling like it 🤷

eysi09:
Partially addressed with #2496.

PR #2479, opened 2021-07-12
Section width

eysi09:

Quick fix to allow users to configure the logger section width to prevent long module and service names from being truncated. As a more long term solution, we could look into implementing #2478.


edvald:
Might I suggest a (much) simpler approach? We could simply stop capping the section length, while still padding to fill the 25 characters. Adding this obscure and super-specific environment variable feels a bit smelly to me :P

eysi09:
Closing in favour of #2496

PR #2480, opened 2021-07-13
Fix: resolve issue with build kit not caching

bdbelevate:
Adds a tag to the buildkit output so that it is pushing to latest.

Buildkit doesn't actually cache anything because the import-cache is looking for an untagged image (ie. latest) but all the builds are going to the garden module versions.

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/container/build/buildkit.ts
+++ b/core/src/plugins/kubernetes/container/build/buildkit.ts
@@ -117,7 +117,7 @@ export const buildkitBuildHandler: BuildHandler = async (params) => {
 
   // Prepare the build command (this thing, while an otherwise excellent piece of software, is clearly is not meant for
   // everyday human usage)
-  let outputSpec = `type=image,name=${deploymentImageId},push=true`
+  let outputSpec = `type=image,"name=${deploymentImageId},${deploymentImageName}:latest",push=true`
 
   if (provider.config.deploymentRegistry?.hostname === inClusterRegistryHostname) {
     // The in-cluster registry is not exposed, so we don't configure TLS on it.
--- a/core/src/plugins/kubernetes/container/build/buildkit.ts
+++ b/core/src/plugins/kubernetes/container/build/buildkit.ts
@@ -139,7 +139,7 @@ export const buildkitBuildHandler: BuildHandler = async (params) => {
     "--export-cache",
     "type=inline",
     "--import-cache",
-    `type=registry,ref=${deploymentImageName}`,
+    `type=registry,ref=${deploymentImageName}:latest`,
     ...getBuildkitFlags(module),
   ]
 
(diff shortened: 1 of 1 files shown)


edvald:
Hey @bdbelevate. I might have misunderstood the BuildKit docs, but I didn't take this to be the correct parameters when using the inline caching method: https://github.com/moby/buildkit#export-cache

It is perhaps debatable whether to use the `inline` or `registry` caching modes, and we did consider making the latter an option, but AFAICT from the docs we were using the inline cache correctly. Did you test this to work?

bdbelevate:
Hi @edvald -Fro what I gather from the docs and testing is that `inline` means to use the image that was outputted in the `--out` command. Here's an example buildctl command from garden...

```
Execing command in cosm-services-user-briandeboer/Pod/garden-buildkit-5ff5d84cf6-lgpph/buildkitd: buildctl build --frontend=dockerfile.v0 --local context=/garden-build/cfae442f-1f9d-478f-ac62-86112502772f/rust-builder/ --local dockerfile=/garden-build/cfae442f-1f9d-478f-ac62-86112502772f/rust-builder/ --opt filename=docker/builder/Dockerfile --output type=image,name=XXX.dkr.ecr.us-west-2.amazonaws.com/cosm-services/rust-builder:v-32f7224389,push=true --export-cache 
type=inline --import-cache type=registry,ref=XXX.dkr.ecr.us-west-2.amazonaws.com/cosm-services/rust-builder --opt build-arg:GARDEN_MODULE_VERSION=v-32f7224389
```

And here is the logs from buildkit in trying to build the module. Notice: that it's looking for `latest`. 
```
 → #5 importing cache manifest from XXX.dkr.ecr.us-west-2.amazonaws.com/cosm-services/rust-builder
  → #5 sha256:cc9a8fe7b6611f98bf0453e9b88a433a9c7e3faf9aa5597bce45d280070ee8f2
  → #5 ERROR: XXX.dkr.ecr.us-west-2.amazonaws.com/cosm-services/rust-builder:latest: not found
```

The other option is to use export-cache registry, but in my testing that didn't work because the manifest format was not working properly with ECR due to https://github.com/aws/containers-roadmap/issues/876

An ideal solution other than the quick fix I suggested would be to make these options for cluster buildkit. But currently it will fail if you add them as extra flags because buildkit will detect that it has two --import-cache and --export-cache parameters and will throw an error.

core/src/plugins/kubernetes/container/build/buildkit.ts
```diff
@@ -117,7 +117,7 @@ export const buildkitBuildHandler: BuildHandler = async (params) => {
 
   // Prepare the build command (this thing, while an otherwise excellent piece of software, is clearly is not meant for
   // everyday human usage)
-  let outputSpec = `type=image,name=${deploymentImageId},push=true`
+  let outputSpec = `type=image,"name=${deploymentImageId},${deploymentImageName}:latest",push=true`
```
edvald, line 120: Tested and this works nicely 👍  Could we change the tag from `latest` to something like `_buildcache`?

bdbelevate: Sure...I'll update accordingly and push the change.

edvald (approved):

edvald:
Thank you so much @bdbelevate!

PR #2485, opened 2021-07-14
feat(cloud): stream command metadata

thsig:

When logged in, we now stream a command metadata event.


thsig:
Any comments on this one?

edvald (approved):

PR #2491, opened 2021-07-16
Mutagen (dev mode) improvements

### LINKED ISSUE #2519: [FEATURE]: Set correct permissions on files synced with devMode.sync
### Background / Motivation

The synced files will be owned by the container's user with read and write permissions only `-rw-------`.
In our case, this doesn't work well for development. We are working with PHP/FPM and FPM (www) doesn't have permission to read the synced files (now owned by root). We can run the containers with the www user, but that introduces some pain as we lose the ability to make quick experimental changes in the container without having to rebuild it.

This wasn't an issue with docker volumes.

I couldn't find a way to execute postSync commands when using devMode.sync, that could work as a workaround, otherwise, configurable permissions and owner on source/target pair basis would be pretty neat.

This would make the synced files accessible for processes that are running with a user different from the container's main user. 

I'd leave this to the professionals.

(inconvenient would be a better word)

edvald:
Combining a few related commits, around the mutagen sync mechanism:

**improvement(k8s): better process mgmt and logging for dev mode sync** 
We now poll for more information from the mutagen daemon, and we've overall improved the process management around it.

Besides the improved runtime information, this unblocks more development around this feature, such as running a postSyncCommand (like with the old hot reloading mechanism) and also makes the use of mutagen easier for other contexts like syncing build contexts for in-cluster building.

Also closes #2519 through a few added option flags.

**feat(k8s): allow owner/perm tweaks on dev mode syncs**
Also updated our docs a bit.

Closes #2519

**feat(k8s): add experimental mutagen-based build sync mode**
You can now set `GARDEN_K8S_BUILD_SYNC_MODE=mutagen` to enable a new,
as yet experimental build sync mode, which replaces rsync with mutagen
for synchronizing build contexts ahead of `kaniko` and
`cluster-buildkit` builds.

Please give it a spin, since we want to battle-test this a bit before
removing the older rsync method.

### DIFF
--- a/core/src/plugins/kubernetes/dev-mode.ts
+++ b/core/src/plugins/kubernetes/dev-mode.ts
@@ -16,13 +16,11 @@ import { joinWithPosix } from "../../util/fs"
 import chalk from "chalk"
 import { PluginContext } from "../../plugin-context"
 import { ConfigurationError } from "../../exceptions"
-import { ensureMutagenSync, mutagenConfigLock } from "./mutagen"
+import { ensureMutagenSync, getKubectlExecDestination, mutagenAgentPath, mutagenConfigLock } from "./mutagen"
 import { joiIdentifier } from "../../config/common"
 import { KubernetesPluginContext } from "./config"
-import { prepareConnectionOpts } from "./kubectl"
 
 const syncUtilImageName = "gardendev/k8s-sync:0.1.1"
-const mutagenAgentPath = "/.garden/mutagen-agent"
 
 interface ConfigureDevModeParams {
   target: HotReloadableResource
@@ -153,8 +151,6 @@ export async function startDevModeSync({
       })
     }
 
-    const kubectl = ctx.tools["kubernetes.kubectl"]
-    const kubectlPath = await kubectl.getPath(log)
     const k8sCtx = <KubernetesPluginContext>ctx
 
     let i = 0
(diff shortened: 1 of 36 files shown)


thsig:
Good stuff!

One thing I noticed: If I run `garden deploy --dev vote --env remote` in the `vote` example project, I get the following error:
```
Failed deploying service 'vote' (from module 'vote'). Here is the output:
Command "/Users/ths/.garden/tools/mutagen/62a411c726894601/mutagen sync create /Users/ths/src/garden/examples/vote/vote/src
exec:'/Users/ths/.garden/tools/kubectl/fd63e19492c91ef3/kubectl exec -i --context=gke_garden-dev-200012_europe-west1-b_garden-dev-1
--namespace=vote-ths --container vote Deployment/vote -- /.garden/mutagen-agent synchronizer':/app/src --name Deployment--vote-ths--vote-0
 --auto-start=false --mode two-way-safe" failed with code 1:

Error: unknown flag: --mode
```

edvald:
Ah, seems the mode flag is put in the wrong place. Will fix.

thsig:
Another thing I noticed: The Joi validation of `devMode.sync[].mode` for `container` modules isn't validating that `mode` is one of the allowed values.

I was able to set `mode: bongo` without getting a validation error when re-running the above command.

core/src/plugins/kubernetes/dev-mode.ts
```diff
@@ -171,118 +149,52 @@ export async function startDevModeSync({
...
-          log,
-          env: {
-            MUTAGEN_DATA_DIRECTORY: dataDir,
-          },
-        })
-        break
-      } catch (err) {
-        const unableToConnect = err.message.match(/unable to connect to beta/)
-        const alreadyRunning = err.message.match(/project already running/)
-        if ((unableToConnect || alreadyRunning) && loops < 10) {
```
thsig, line 263: Nitpick: Replace `10` with `maxRetries`.

thsig (commented):
Looks great! Only had one (minor) additional comment there.

core/src/plugins/kubernetes/dev-mode.ts
thsig: Haha

thsig (approved):

thsig:
Yeah, I've poked at this from many different directions—looks good so far 👍

PR #2494, opened 2021-07-17
fix(core): allow empty replacement strings in 'replace' helper function

### LINKED ISSUE #2493: 'replace' helper function does not allow empty replacement strings
### Current Behavior

The replacement string provided to the `replace` function cannot be an empty string.

### Expected behavior

The `replace` function can be used to replace substrings with empty strings.

### Reproducible example

```
defaultNamespace: user-${replace("josh.drake", ".", "")}
```

```
Invalid template string (user-${replace("josh.drake", ".", ""…): Error validating argument 'replacement' for replace helper function: value is not allowed to be empty
```

### Workaround

N/A

### Suggested solution(s)

Fix `joi` validation for replacement argument.

### Additional context

N/A

### Your environment

* **OS:** Manjaro

`garden version`

0.12.23

JoshDrake-minted:

Fixes `joi` validation for `replacement` argument of `replace` helper function to allow empty strings.

Fixes #2493 

### DIFF
--- a/core/src/template-string/functions.ts
+++ b/core/src/template-string/functions.ts
@@ -114,7 +114,11 @@ const helperFunctionSpecs: TemplateHelperFunction[] = [
     arguments: {
       string: joi.string().required().description("The string to convert."),
       substring: joi.string().required().description("The substring to replace."),
-      replacement: joi.string().required().description("The replacement for each instance found of the `substring`."),
+      replacement: joi
+        .string()
+        .required()
+        .allow("")
+        .description("The replacement for each instance found of the `substring`."),
     },
     outputSchema: joi.string(),
     exampleArguments: [


eysi09:
Hi @JoshDrake-minted!

Thanks a lot for the PR. This looks good. Tests are failing but looks like that's unrelated to your PR. We'll fix those on our end.

But I think you might need to run `yarn run generate-docs` to ensure the auto generated docs are up to date.

JoshDrake-minted:
Re-generating the docs doesn't generate any changes as the parameters and examples are all still the same. Indeed, an example invocation of the `replace` function indicated this behavior was intended originally but simply did not work as expected:

```
### replace

Replaces all occurrences of a given substring in a string.

Usage: `replace(string, substring, replacement)`

Examples:
* `${replace("string_with_underscores", "_", "-")}` -> `"string-with-underscores"`
* `${replace("remove.these.dots", ".", "")}` -> `"removethesedots"`

```

edvald (approved):

PR #2510, opened 2021-07-27
docs: clarify image publishing docs a bit

edvald:
The placement was a little confusing.


thsig (changes requested):
Good stuff! Had a few text-related comments there.

PR #2511, opened 2021-07-28
feat: add jib provider and jib-container module type

### LINKED ISSUE #2015: [FEATURE]: jib-container support
### Background / Motivation

[Jib](https://github.com/GoogleContainerTools/jib) is a tool used to help on building optimized docker images

We highly use it in our environments to make the process of creating new services easy and reliable

The `container` is expected to work with a Docker file and the `maven-container` is meant to build an image using jar files 

Build docker images with Jib

Jib is super helpful and builds consistent light images with a small effort

A new module to integrate with Jib (maven, cli, gradle), or some support in the other modules to extends build command where the user could specify a `mvn jib:build` when needed

Super important, all service uses an standard Jib config

edvald:
By popular demand, this adds a proper integration for
[Jib](https://github.com/GoogleContainerTools/jib), which makes building
Java container images much more efficient.

The user-facing implementation is described in the added docs, but here
are some technical points for the reviewer(s):

1. We followed Skaffold's example in most ways, in terms of how to
   interact with Jib.
2. The `jib-container` module extends the `container` module type using
   our handy inheritance mechanisms. It behaves the exact same way
   outside of the build flow, and adds just a handful of fields to
   control the image build.
3. The `kubernetes` provider has explicit support for the module type,
   in order to avoid requiring a local Docker instance when using remote
   building. This actually prompted the most complexity in the
   integration, but I figure that makes the feature much more usable for
   many users. Even then it just comes down to one plugin handler.

The included `jib-container` example project is a good place to start
in terms of testing and poking around. Also check out the added
reference docs.

Closes #2015 

TODO:
- [x]  Write tests


thsig (approved):

jondeandres:
hey there, this looks quite amazing 😍 , any estimation of when this will be part of a new release?

PR #2513, opened 2021-07-28
feat: add ephemeralStorage limit and request configuration parameters for kubernetes builder, registry and sync

### LINKED ISSUE #2469: [FEATURE]: Allow configuring ephemeral-storage request and limit for kaniko pods
### Background / Motivation

When I was trying to configure kaniko build mode for a remote kubertenes on GKE autopilot, I hit the ephemeral storage limit which is only 1Gi there:
```
{
  "message": "Pod ephemeral local storage usage exceeds the total limit of containers 1Gi. ",
  "phase": "Failed",
  "reason": "Evicted",
  "startTime": "2021-07-07T13:34:26.000Z"
}
```

I think you should be able to specify ephemeral-storage request and limit for the kaniko pod created by garden:
https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#setting-requests-and-limits-for-local-ephemeral-storage

This will allow you to calibrate the ephemeral storage size for larger docker images.

I think one should be able to specify `ephemeralStorage` just like `cpu` and `memory` for the [requests](https://docs.garden.io/reference/providers/kubernetes#providers-resources-builder-requests) and [limits](https://docs.garden.io/reference/providers/kubernetes#providers-resources-builder-limits).

zyulyaev:
Fixes #2469


edvald:
Thanks so much @zyulyaev! I think this makes good sense. 

My only concern is that we're adding a default limit and request on the ephemeral storage. I think it's a sensible thing to do (in hindsight) but that's kind of a breaking change. I think we have to have no default for now to avoid unexpected issues, and maybe we could set a default in a breaking release.

Btw, those test failures are unrelated to your PR, they're fixed in a separate PR.

zyulyaev:
> My only concern is that we're adding a default limit and request on the ephemeral storage.

Yes, that was my concern as well. But I didn't find an example of that, so just replicated the way it works for cpu and memory.

I will try to fix that now.

edvald:
Great! I'll drop in some comments to help guide the change, in case it's helpful.

core/src/plugins/kubernetes/config.ts
```diff
@@ -68,10 +68,12 @@ interface KubernetesResourceSpec {
   limits: {
     cpu: number
     memory: number
+    ephemeralStorage: number
```
edvald, line 71: Making this `ephemeralStorage?: number` and dropping the relevant `.default(...)` lines should do the trick. I'll take a quick look at where this is referenced as well.

core/src/plugins/kubernetes/init.ts
```diff
@@ -396,15 +396,19 @@ export function getKubernetesSystemVariables(config: KubernetesConfig) {
 
     "registry-limits-cpu": millicpuToString(config.resources.registry.limits.cpu),
     "registry-limits-memory": megabytesToString(config.resources.registry.limits.memory),
+    "registry-limits-ephemeralStorage": megabytesToString(config.resources.registry.limits.ephemeralStorage),
     "registry-requests-cpu": millicpuToString(config.resources.registry.requests.cpu),
     "registry-requests-memory": megabytesToString(config.resources.registry.requests.memory),
+    "registry-requests-ephemeralStorage": megabytesToString(config.resources.registry.requests.ephemeralStorage),
```
edvald, line 402: You'll need something like `config.resources.registry.requests.ephemeralStorage && megabytesToString(config.resources.registry.requests.ephemeralStorage)`, which will resolve to undefined if there's no value there.

static/kubernetes/system/build-sync/garden.yml
```diff
@@ -12,9 +12,11 @@ values:
     limits:
       cpu: ${var.sync-limits-cpu}
       memory: ${var.sync-limits-memory}
+      ephemeral-storage: ${var.sync-limits-ephemeralStorage}
```
edvald, line 15: You'll need a ? at the end of these lines to allow the undefined value: `ephemeral-storage: ${var.sync-limits-ephemeralStorage}?`
Otherwise the template parser will complain.

core/src/plugins/kubernetes/init.ts
zyulyaev: undefined cannot be used here, so I used a conditional spread operator here

edvald (approved):

PR #2514, opened 2021-07-28
feat(cli): add self-update command

edvald:
This adds a `garden self-update` command, that updates the Garden CLI
in place.

You can also specify a specific version to install (including _edge_),
and install to a specified directory.


core/src/commands/self-update.ts
```diff
@@ -0,0 +1,291 @@
...
+  name = "self-update"
+  help = "Update the Garden CLI."
+
+  cliOnly = true
+  noProject = true
+
+  description = dedent`
+    Updates your Garden CLI in-place.
+
+    Defaults to the latest release version, but you can also request a specific release version as an argument.
```
thsig, line 65: "latest release version" -> "latest release"

thsig (changes requested):
This will be much appreciated by many of our users! Only had one text-related comment there.

PR #2540, opened 2021-08-11
feat: allow passing additional tolerations to kaniko pods

to266:
Allows specifying additional tolerations that get passed to kaniko pods (both util and the builder pods).

A nice to have if stuff is run on, say, AKS - where any spot instance automatically has a taint that you need to tolerate in order to run anything.

### DIFF
--- a/core/src/plugins/kubernetes/config.ts
+++ b/core/src/plugins/kubernetes/config.ts
@@ -120,6 +120,7 @@ export interface KubernetesConfig extends BaseProviderConfig {
     extraFlags?: string[]
     namespace?: string | null
     nodeSelector?: StringMap
+    tolerations?: V1Toleration[]
   }
   context: string
   defaultHostname?: string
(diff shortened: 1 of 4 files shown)


edvald (changes requested):

core/src/plugins/kubernetes/config.ts
```diff
@@ -420,6 +421,7 @@ export const kubernetesConfigBase = () =>
             [See here](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/) for the official Kubernetes guide to assigning Pods to nodes.
           `
         ),
+        tolerations: joiSparseArray(tolerationSchema()),
```
edvald, line 424: Would be good to add a `.description(...)` here to describe what the field does. Something like `.description("Specify tolerations to apply to each Kaniko Pod. Useful to control which nodes in a cluster can run builds.")`

edvald (approved):

PR #2559, opened 2021-08-26
Prepare imagePullSecrets on helm module deployments

jondeandres:

Prepares the `imagePullSecrets` in the deployed namespace.

When deploying our own charts through the helm module we are having issues pulling the image, I think because the image pull secret is not being copied to the deployment namespace.

We've set `imagePullSecrets` in the helm chart and the container image seems to be correct:

```
# deployment yaml
# ..
    spec:          
      containers:                                                                       
      - image: 346163573116.dkr.ecr.us-east-2.amazonaws.com/uniphore/tenants:v-596c9dd183

# ...
      imagePullSecrets:               
      - name: ecr-config

```

I didn't check if this fix works or not but looking at garden code I think is the approach to take.

Our intention is using garden to develop our services while we build the Helm charts since we think preparing Helm charts for our service is a good idea for business reasons. At same time, it seems to be proper the way inject garden variables. We know we can define the manifests in the `kubernetes` module itself but we using Helm would be our preference.

We'd appreciate some insights about the good practices when deploying with Helm charts.

Thanks.


jondeandres:
Closing since I think the issue was due to some bad scenario in our side.

PR #2560, opened 2021-08-27
chore(enterprise): add Core version to Cloud API http request headers

10ko:

Adds a `x-core-version` header to all the Garden Cloud HTTP requests.

Fixes #

### DIFF
--- a/core/src/enterprise/api.ts
+++ b/core/src/enterprise/api.ts
@@ -20,6 +20,10 @@ import { isObject } from "lodash"
 import { deline } from "../util/string"
 import chalk from "chalk"
 import { GetProjectResponse } from "@garden-io/platform-api-types"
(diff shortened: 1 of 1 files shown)


10ko:
I actually think we should probably add two headers instead of one, I was thinking of `x-client-version` and `x-client-name`, so you'll have something:
```
x-client-version: "0.12.25"
x-client-name: "garden-core"
```
In this way we could differentiate between different consumers of the `api`. If that's the case we'd go for, would you still add the `x-garden-*` prefix?

thsig:
Makes sense—I think we should go for `x-garden-client-version` and `x-garden-client-name`, just to make things extra clear (and prevent any potential collisions with other tools down the line).

10ko:
@thsig changed as requested.

thsig (approved):

PR #2562, opened 2021-08-30
docs: add section for cluster-init issues

s-chand:
This PR updates the troubleshooting section of the docs with instructions for how to handle service unreachable errors that ocassionally result from updating an existing garden installation in a kubernetes cluster. 

Fixes #


docs/misc/troubleshooting.md
```diff
@@ -112,4 +112,10 @@ This is a bug in Docker CE (i.e. Docker for Desktop), version `2.4.x.y`. See thi
 
 In some container repositories, you may need to create the cache repo manually.
 
-See [this section](https://docs.garden.io/guides/in-cluster-building#kaniko) of our docs and this [GitHub comment](https://github.com/GoogleContainerTools/kaniko/issues/410#issuecomment-433229841) for more details.
\ No newline at end of file
+See [this section](https://docs.garden.io/guides/in-cluster-building#kaniko) of our docs and this [GitHub comment](https://github.com/GoogleContainerTools/kaniko/issues/410#issuecomment-433229841) for more details.
+
+### Can't reach my services on existing ingress urls after updating with `garden cluster-init`
```
eysi09, line 117: It might be more accurate to say: Can't reach my services on existing ingress URLs after re-installing Garden system services.

docs/misc/troubleshooting.md
```diff
@@ -112,4 +112,10 @@ This is a bug in Docker CE (i.e. Docker for Desktop), version `2.4.x.y`. See thi
...
+
+This can occur when `garden --env=<environment-name> plugins kubernetes cluster-init` is run to update the garden installation version of a cluster. The external IP address for the LoadBalancer service can be changed as a result of the update.
```
eysi09, line 119: Maybe it would be slightly more accurate to phrase it like this:

This can occur if you re-install the Garden Nginx Ingress Controller. For example because you ran `garden plugins kubernetes uninstall-garden-services` and then `garden plugins kubernetes cluster-init` when upgrading the system services. When the Ingress Controller gets re-installed, it may be assigned a new IP address by your cloud provider, meaning that hostnames pointing to the previous one will no longer work.

To fix this, run `kubectl get svc -n garden-system` and look for the `EXTERNAL-IP` of the `garden-nginx-nginx-ingress-controller` service and update your DNS records with this new value.

eysi09 (commented):
Thanks for adding this! I suggested some changes to make it slightly more accurate. It's not the `cluster-init` command per se, but rather the fact that nginx gets re-installed. Usually you should be able to run the `cluster-init` command without getting assgined a new IP. It's only when it doesn't work, and you try first deleting the services and then installing them again that this happens.

docs/misc/troubleshooting.md
s-chand: Indeed, this does sound more accurate. Thanks

eysi09 (approved):

PR #2574, opened 2021-09-10
improvement(core): don't watch dev-enabled modules

thsig:

This is a performance optimization that lowers resource use when using dev mode. We now don't watch files inside the module roots of modules having one or more services deployed with dev mode, except for the modules' Garden config path.

The rationale here is that FS watches can be resource intensive, especially when there's a lot of churn (and especially on Linux). Since Mutagen takes care of the dev-mode sync (and doesn't go through the framework-native watch logic), this is usually unwanted overhead.

Also fixed a UI bug for watch-mode commands where an additional "Waiting for code changes..." footer log line would be rendered if a Garden config was changed.

This PR is an attempt to fix OOM issues that some of our Linux users have encountered during long dev-mode sessions.

### DIFF
--- a/core/src/commands/deploy.ts
+++ b/core/src/commands/deploy.ts
@@ -8,6 +8,7 @@
 
 import deline = require("deline")
 import dedent = require("dedent")
+import chalk = require("chalk")
 
 import {
   Command,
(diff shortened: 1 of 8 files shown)


eysi09:
> Also fixed a UI bug for watch-mode commands where an additional "Waiting for code changes..." footer log line would be rendered if a Garden config was changed.

❤️

eysi09 (approved):

PR #2575, opened 2021-09-13
improvement: log namespace and cloud URL

eysi09:

Prints the namespace Garden is running in in the header and the URL to the project in Garden Cloud if applicable.

Screenshot 2021-09-13 at 09 20 36

Fixes #

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -39,6 +39,7 @@ import { DashboardEventStream } from "../server/dashboard-event-stream"
 import { GardenPluginCallback } from "../types/plugin/plugin"
 import { renderError } from "../logger/renderers"
 import { EnterpriseApi } from "../enterprise/api"
+import chalk = require("chalk")
 
 export async function makeDummyGarden(root: string, gardenOpts: GardenOpts) {
   const environments = gardenOpts.environmentName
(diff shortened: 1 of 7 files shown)


core/src/cli/cli.ts
```diff
@@ -236,13 +237,18 @@ ${renderCommands(commands)}
 
     contextOpts.persistent = persistent
     const { streamEvents, streamLogEntries } = command
+    // Print header log before we know the namespace to prevent content from
+    // jumping.
+    // TODO: Link to Cloud namespace page here.
+    const nsLog = headerLog.info(chalk.white(`Running in namespace ...\n`))
```
thsig, line 243: Good stuff!

I think a darker / gray colour "Running in namespace" would be cleaner (while keeping the white for the actual namespace name).

eysi09: Like this?

thsig: Yeah, that's good! I think I'd prefer an even darker gray (like in "Authorizing..."), but either works for me.

eysi09: It's the same color, `chalk.gray`.

thsig (approved):

thsig (approved):

PR #2579, opened 2021-09-14
chore(tests): remove .only call

thsig:


core/test/integ/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -29,7 +29,7 @@ import { pathExists, readFile, remove, writeFile } from "fs-extra"
 import { execInWorkload, kilobytesToString, millicpuToString } from "../../../../../../src/plugins/kubernetes/util"
 import { getResourceRequirements } from "../../../../../../src/plugins/kubernetes/container/util"
 
-describe.only("kubernetes container deployment handlers", () => {
+describe("kubernetes container deployment handlers", () => {
```
eysi09, line 32: When was this introduced and why didn't the linter catch it?

eysi09 (approved):

core/test/integ/src/plugins/kubernetes/container/deployment.ts
thsig: This was introduced 5 days ago (in between releases). I was fixing a broken integ test on the container security context branch, and forgot to remove this after the fact.

Since we introduced new docs in the `jib` PR, the lint step has been failing (because of references to missing doc links, which will get fixed in the next release). So maybe I didn't notice that the lint was failing for that reason.

Anyway, we haven't made any releases without these tests being applied.

PR #2582, opened 2021-09-15
feat(core): implement project-level dev excludes

thsig:

Exclusion rules for dev-mode syncs can now also be set via a new project-level field, `devModeExclude`.

Exclusion rules defined `devModeExclude` are included before any rules specified in `devMode.sync[].exclude`.

This is useful for reducing repetition in sync excludes, e.g. to always exclude `node_modules` or `tmp`.


eysi09:
Looks like you did npm install instead of yarn.

core/src/config/project.ts
```diff
@@ -317,45 +320,54 @@ export const projectDocsSchema = () =>
...
+            Note that these take precedence over the project \`module.include\` field, and module \`include\` fields, so any paths matched by the .ignore files will be ignored even if they are explicitly specified in those fields.
 
-        See the [Configuration Files guide](${DOCS_BASE_URL}/using-garden/configuration-overview#including-excluding-files-and-directories) for details.
-      `
+            See the [Configuration Files guide](${DOCS_BASE_URL}/using-garden/configuration-overview#including-excluding-files-and-directories) for details.
+        `
         )
         .example([".gardenignore", ".gitignore"]),
+      devModeExclude: joiSparseArray(joi.string())
+        .default(defaultDevModeIgnores)
```
eysi09, line 333: If I need to set my own excludes, will I need to remember to include `.garden`, `.gardenignore` and `.git`? It that's the case, should we perhaps just hard code these values? Or at the least thoroughly document it.

core/src/config/project.ts
```diff
@@ -43,6 +43,8 @@ import { CommandInfo } from "../plugin-context"
 export const defaultVarfilePath = "garden.env"
 export const defaultEnvVarfilePath = (environmentName: string) => `garden.${environmentName}.env`
 
+const defaultDevModeIgnores = ["**/*.garden", "**/.gardenignore", "**/*.git"]
```
eysi09, line 46: Why `.gardenignore`?

core/test/integ/src/plugins/kubernetes/dev-mode.ts
```diff
@@ -0,0 +1,192 @@
...
+import { getContainerServiceStatus } from "../../../../../src/plugins/kubernetes/container/status"
+import { KubernetesWorkload } from "../../../../../src/plugins/kubernetes/types"
+import { execInWorkload } from "../../../../../src/plugins/kubernetes/util"
+import { emptyRuntimeContext } from "../../../../../src/runtime-context"
+import { DeployTask } from "../../../../../src/tasks/deploy"
+import { dedent } from "../../../../../src/util/string"
+import { sleep } from "../../../../../src/util/util"
+import { getContainerTestGarden } from "./container/container"
+
+describe.only("dev mode deployments and sync behavior", () => {
```
eysi09, line 26: Remove `only`

core/test/unit/src/actions.ts
```diff
@@ -623,6 +652,7 @@ describe("ActionRouter", () => {
           runtimeContext,
           force: true,
           devMode: false,
+          devModeExcludes: [],
```
eysi09, line 655: Can't we have a helper function that produces the default values so that we don't have to update all these test files every time we change these parameters? (Out of scope for the PR, jus' sayin')

eysi09 (commented):
Had some questions / comments. Also looks like you did a npm install.

core/src/config/project.ts
thsig: Yes, unfortunately. There may be use cases where the user wants to sync the `.git` directory.

The defaults do show up in the generated reference documentation, but we'll need to make sure this is thoroughly covered in the additions we're about to make to the dev mode guide.

core/src/config/project.ts
thsig: Good point, I'll remove that.

core/test/unit/src/actions.ts
thsig: Yeah, I thought about just that while writing this PR. We should do that soon.

core/src/config/project.ts
thsig: The `.garden` directories should always be excluded, though. I'll make that automatic at least.

thsig:
Any further comments on this one, @eysi09?

thsig:
Closing in favor of #2590.

PR #2585, opened 2021-09-16
fix(core): fix handling of empty option values

thsig:

To ensure that `command.params` behaves intuitively in template strings, we don't add option keys with `null`/`undefined` values to the command info object (which gets passed to `command.params` in template strings).

For example, we don't want `${command.params contains 'dev-mode'}` to be `true` when running `garden deploy` unless the `--dev` flag was actually passed (since the user would expect the option value to be an array if present).


10ko (approved):
Looks good, thanks for being the fastest! :D 
There are some tests failing but not sure why if it's related or it's the cluster acting up.

PR #2592, opened 2021-09-23
fix(workflows): defer provider resolution

thsig:

This fixes a regression introduced in `6382d3d` (emit stack graph & task log events), where we began emitting `stackGraph` events before running the workflow's steps.

This caused problems for workflows that rely on script steps to prepare authentication for providers (e.g. k8s credentials).

We now simply emit the `stackGraph` event from the relevant step commands.


10ko (approved):
Looks good to me! 
I tested it locally and works as expected.

PR #2599, opened 2021-09-28
chore(examples): update node base images

thsig:

The node base images we were using in our example projects were getting quite old (mostly using node 9), so it was time for an update.

### DIFF
--- a/examples/remote-sources/garden.yml
+++ b/examples/remote-sources/garden.yml
@@ -2,7 +2,7 @@ kind: Project
 name: remote-sources
 sources:
   - name: web-services
-    repositoryUrl: https://github.com/garden-io/garden-example-remote-sources-web-services.git#v0.2.0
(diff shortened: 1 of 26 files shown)


examples/remote-sources/garden.yml
```diff
@@ -2,7 +2,7 @@ kind: Project
 name: remote-sources
 sources:
   - name: web-services
-    repositoryUrl: https://github.com/garden-io/garden-example-remote-sources-web-services.git#v0.2.0
+    repositoryUrl: https://github.com/garden-io/garden-example-remote-sources-web-services.git#v0.3.0
```
thsig, line 5: Note: The `v0.3.0` tag for the web services here points to a commit I pushed in that repo that updates the node image for the `result` service.

eysi09 (approved):

PR #2609, opened 2021-10-03
fix(cli): ensure help exits with code 0

eysi09:

Previously running `garden --help` and `garden -h` would exit with code
`0` but running `garden help` would exit with code `1`. This fixes that.

Fixes #

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -399,7 +399,7 @@ ${renderCommands(commands)}
     const { command } = pickCommand(Object.values(this.commands), argv._)
 
     if (!command) {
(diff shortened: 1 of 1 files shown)


core/src/cli/cli.ts
```diff
@@ -456,6 +456,7 @@ ${renderCommands(commands)}
     }
 
     const gardenErrors: GardenBaseError[] = errors.map(toGardenError)
+    console.log("lenght", gardenErrors.length)
```
edvald, line 459: Stray console log

edvald (approved):

PR #2611, opened 2021-10-04
fixes docs mentioning one-way-replicated sync mode

guhcampos:
The devMode docs reference `one-way-replicated` as one of mutagen's synchronization methods, while the actual parameter is `one-way-replica` instead, using the formerresults in the following error:

```
...
key .services[0][devMode][sync][0][mode] must 
be one of [one-way, one-way-replica, one-way-reverse, one-way-replica-reverse, two-way], key .services[1][name]
```

This PR simply replaces all typos in the docs, no code changes required.


thsig (approved):
Good catch!

PR #2616, opened 2021-10-12
Helm: Avoid unneccesary builds

thsig:
**Avoid unnecessary Helm builds**

Previously, Helm modules would get rebuilt ahead of deployment, even if the service version hadn't changed (thus leading deploy tasks to not redeploy the service, making the build step unnecessary).

We address that here by making the Helm build handler a no-op, and instead running its logic inside the deployment handler.


eysi09 (dismissed):

edvald:
Wanna revive this one @thsig?

core/test/integ/src/plugins/kubernetes/helm/common.ts
```diff
@@ -46,26 +48,30 @@ export async function getHelmTestGarden() {
...
       })
   )
-  const results = await garden.processTasks(tasks)
-
-  const err = first(Object.values(results).map((r) => r && r.error))
-
-  if (err) {
-    throw err
-  }
+  await garden.processTasks(buildTasks)
```
edvald, line 71: May not really matter here but FYI, you can make this `await garden.processTasks(buildTasks, { throwOnError: true })` to make sure errors are thrown.

core/src/plugins/kubernetes/helm/deployment.ts
```diff
@@ -211,3 +227,108 @@ export async function deleteService(params: DeleteServiceParams): Promise<HelmSe
...
+        log.debug("Fetching chart (after updating)...")
+        await pullChart(k8sCtx, log, module)
+      }
+    }
+  })
+}
+
+async function pullChart(ctx: KubernetesPluginContext, log: LogEntry, module: HelmModule) {
+  if (await chartsFetched(log, module)) {
+    log.debug(`Charts already fetched for module ${module.name}`)
```
edvald, line 273: _Super_ minor nitpick, but it's always a single chart, not plural.

edvald (approved):
LGTM 👍 Very minor comments, barely worth actioning

edvald:
@thsig Tests seem to be legitimately failing. Please make sure those are okay before merging.

PR #2618, opened 2021-10-13
fix: do not modify baseModule.spec.serviceResource and baseModule.spec.values by the values from the dependent modules

### LINKED ISSUE #2589: Helm values from a chart applied to dependent charts
### Current Behavior

I've create [simple project](https://github.com/pstetsuk-exadel/garden-helm-test) to reproduce this bug.  
There are 3 modules:
-  base-chart - base template
- chart1
- chart2 - depend on char1

In base-chart defined values.yaml
```
addPorts: {}
```

In chart1 garden.yml looks like
```
values:
  addPorts:
    - port: 443
      targetPort: https
      protocol: TCP
      name: https
```

In chart2 value for `addPorts` is not defined but defined dependency to chart1
```
dependencies:
  - chart1
```

### Expected behavior

Values defined for chart1 should not apply to chart2

### Reproducible example

After run `garden deploy` check `.garden\build\chart1\garden-values.yml` and `.garden\build\chart2\garden-values.yml`  
Both these files looks almost the same and contain
```
addPorts:
  - port: 443
    targetPort: https
    protocol: TCP
    name: https
```

### Workaround

If deploy modules one by one, `garden deploy chart1` and then `garden deploy chart2` all works as expected

### Your environment

* **OS:** Windows

`garden version` `0.12.26`

pstetsuk-exadel:

Fixes #2589


pstetsuk-exadel:
@edvald, I tested it in more detail, `cloneDeep()` works in all cases so I modified the code to use it.  

Note about `apply(target, patch)`, it returns `null` if `patch` parameter is `null`.  
I.e. for helm modules if values specified in base module and no values defined in dependent module than values from the base module will be lost. Not common use case but anyway I fixed it.

pstetsuk-exadel:
@thsig, will it be included in the next release?

thsig:
@pstetsuk-exadel: Yes indeed. Thanks so much for the contribution!

PR #2620, opened 2021-10-14
fix(k8s): uncaught error when trying to patch namespace resource

edvald:
Fixes the issue reported here: https://kubernetes.slack.com/archives/CKM7CP8P9/p1634222486155200?thread_ts=1634222429.155100&cid=CKM7CP8P9

PSA: Returning a Promise in a try/catch will not achieve what you want :) The Promise is returned after the function returns, so the try/catch is gone by the time the error is thrown.

### DIFF
--- a/core/src/plugins/kubernetes/namespace.ts
+++ b/core/src/plugins/kubernetes/namespace.ts
@@ -24,35 +24,54 @@ import { V1Namespace } from "@kubernetes/client-node"
 import { isSubset } from "../../util/is-subset"
 import chalk from "chalk"
 import { NamespaceStatus } from "../../types/plugin/base"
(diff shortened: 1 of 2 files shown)


thsig:
The integ tests for kind and Minikube are failing on this test case:
```
  1) ensureNamespace
       should do nothing if the namespace has already been configured:
     AssertionError: expected { Object (apiVersion, kind, ...) } to be null
      at /home/circleci/project/core/test/integ/src/plugins/kubernetes/namespace.ts:134:25
      at fulfilled (build/test/integ/src/plugins/kubernetes/namespace.js:12:58)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

PR #2621, opened 2021-10-14
improve(jib): faster builds for most common scenarios

edvald:
This changes the default behavior to use the `jib:build` / `jibBuild`
target instead of building a tarball. This is quite a bit faster for
most scenarios.

Importantly this means you need to have push authorization to the
deployment registry where you're running Garden, instead of just in the
cluster.

The previous behavior can be forced by setting
`jib.pushViaCluster: true` in the `kubernetes` provider config.

Also added is a `build.buildInDocker` flag, to optionally use a local
Docker daemon to perform the build.


edvald:
@thsig alright, this is now ready for review. I'll make sure the tests pass but I've tested manually in various different modes.

thsig (approved):
Looks good to me!

plugins/jib/util.ts
```diff
@@ -69,25 +94,31 @@ export function getBuildFlags(module: JibContainerModule, projectType: JibModule
   const tarPath = resolve(module.path, targetDir, tarFilename)
```
carolinapc, line 94: Hi, I just like to suggest to add a feature where we can set the tar path, cause we are currently working with multiple modules and we need this flexibility. 
Thank you

PR #2635, opened 2021-10-28
fix: remove deployment as default serviceresouce kind

solomonope:
Currently `serviceResourceSchema.Kind`  has a default value of `Deployment`  set. This violates the  `oxor("podSelector", "kind")`  validation condition. This leads to the following error message `storage: Error validating Module 'storage': key .serviceResource contains a conflict between optional exclusive peers [podSelector, kind]`  when values are set for `serviceResource.podSelector`. This PR removes the default value.

Also `Kubernetes` and `Helm` Module types  has  `namespace`  field,  current the  `devMode` ignores this field if it is set. This PR give preference to this  property.

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/config.ts
+++ b/core/src/plugins/kubernetes/config.ts
@@ -753,7 +753,6 @@ export const serviceResourceSchema = () =>
       kind: joi
         .string()
         .valid(...hotReloadableKinds)
-        .default("Deployment")
         .description("The type of Kubernetes resource to sync files to."),
       name: joi.string().description(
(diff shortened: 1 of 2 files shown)


eysi09:
Thanks for the contribution @solomonope! But I think it's actually easier better to go with the approach in #2637 so I'll go ahead and close this one.

solomonope:
thanks

> Thanks for the contribution @solomonope! But I think it's actually easier better to go with the approach in #2637 so I'll go ahead and close this one.

thanks

PR #2638, opened 2021-10-31
improve(cli): clearer message stating the namespace/environment name

edvald:
I found the log message slightly confusing and I think we should include the full environment name.

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -258,7 +258,8 @@ ${renderCommands(commands)}
           garden = await makeDummyGarden(root, contextOpts)
         } else {
           garden = await Garden.factory(root, contextOpts)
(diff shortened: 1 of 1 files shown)


core/src/cli/cli.ts
```diff
@@ -258,7 +258,11 @@ ${renderCommands(commands)}
           garden = await makeDummyGarden(root, contextOpts)
         } else {
           garden = await Garden.factory(root, contextOpts)
-          nsLog.setState(`${chalk.gray(`Running in namespace ${chalk.white.bold(garden.namespace)}\n`)}`)
+          nsLog.setState(
+            `${chalk.gray(
+              `Using environment ${chalk.white.bold(garden.namespace)}.${chalk.white.bold(garden.environmentName)}\n`
```
thsig, line 263: Good stuff! Only comment is that I think this would look a bit clearer clearer if the `.` were also in white.

thsig (approved):

PR #2641, opened 2021-11-01
fix(k8s): use module ns in getServiceResource

thsig:

When provided, we now use the custom namespace defined in `module.spec.namespace` for `helm` and `kubernetes` modules when getting the module's `serviceResource`.


thsig:
CC @solomonope

eysi09 (approved):

PR #2652, opened 2021-11-06
feat(core): render module command

thsig:

Added a new command: `garden render module`.

This command resolves all template strings in the module, and outputs the full module configuration (which includes the resolved module version).

This is useful for debugging/writing template strings for modules, and for reasoning about which files/variables factor into the  module's version calculation.

Note: Garden Cloud/Enterprise secrets are filtered from the rendered output.

### DIFF
--- a/core/src/commands/commands.ts
+++ b/core/src/commands/commands.ts
@@ -9,56 +9,58 @@
 import { Command, CommandGroup } from "./base"
 import { BuildCommand } from "./build"
 import { CallCommand } from "./call"
+import { ConfigCommand } from "./config/config"
 import { CreateCommand } from "./create/create"
+import { DashboardCommand } from "./dashboard"
 import { DeleteCommand } from "./delete"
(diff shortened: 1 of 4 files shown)


thsig:
@eysi09 @edvald: What do we think about the fields included in the output here?

I thought it made sense to omit `serviceConfigs`, `testConfigs` and `taskConfigs`, since the services/tests/tasks also appear in `module.spec`.

tedchang77:
@thsig we were going to make a request for this very feature to help us with debugging...  this command would be combining the functionality of the garden get modules and the garden get config command, correct?

edvald:
@thsig I think we should fold this into the `get modules` command, and basically make the visual output there nicer. The planned `garden render` commands will have different semantics and this clashes with those plans.

thsig:
Yeah, agreed.

thsig:
Closing in favor of #2982.

PR #2658, opened 2021-11-10
feat(cloud): register session with API

thsig:

This is done in preparation for rolling out Stack Streams.

### DIFF
--- a/core/src/enterprise/session-lifecycle.ts
+++ b/core/src/enterprise/session-lifecycle.ts
@@ -1,0 +1,73 @@
+/*
+ * Copyright (C) 2018-2021 Garden Technologies, Inc. <info@garden.io>
+ *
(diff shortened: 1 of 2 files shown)


core/src/enterprise/session-lifecycle.ts
```diff
@@ -0,0 +1,72 @@
...
+  sessionId: string
+  commandInfo: CommandInfo
+  localServerPort?: number
+  projectId: string
+  environment: string
+  namespace: string
+  log: LogEntry
+}
+
+export interface RegisterSessionResponse {
```
eysi09, line 25: Perhaps set a TODO here to read this from the `api-types` npm package once we release this on Cloud side.

eysi09 (approved):
Looks great! One tiny comment but I'll approve regardless.

PR #2660, opened 2021-11-15
fix(core): improve project config validation

### LINKED ISSUE #2653: Strengthen YAML validation and error reporting
### Current Behavior

Right now the YAML parsing in the Garden CLI is very fragile, and unfortunately the error messages leave a lot to be desired.

For example, using the standard boilerplate project configuration you can start with this:
```yaml
providers:
  - # The name of the provider plugin to use.
    name: local-kubernetes
```

I was attempting to clean up the vast amounts of comments in this file and accidentally deleted the comment in the middle of that block, also removing the array notation that was needed for the following line, giving me this:

```yaml
providers:
    name: local-kubernetes
```

Running `garden validate` (or really, any garden command) yields this output:

```shell
$ garden validate
Validate ✔️


undefined is not a function
```

### Expected behavior

Some error that indicates that the configuration is malformed, and an approximate location of where the problem is.

### Reproducible example

An example project.garden.yml:

```yaml
kind: Project
name: example
environments:
  - name: default
providers:
    name: local-kubernetes
```

### Workaround

No workaround.

### Suggested solution(s)

Better error output for malformed configuration files.

### Additional context

N/A

### Your environment

* **OS:** MacOS 11.6

`garden version`
0.12.30

thsig:

We now validate that `project.environments` is an array and throw an informative error if it isn't.

Previously, if `project.environments` was mistakenly formatted as a map instead of an array, a cryptic type error would be printed (e.g. "Cannot read property name of undefined"). We now clearly indicate the problem
to the user.

Additionally, the `joiSparseArray` schema can now properly handle non-array inputs without failing cryptically.

Fixes #2653.

You'll note that I removed support for defining an environment with a string instead of an object. For example something like the following was allowed by `environmentsSchema`
```
environments
- someEnvName
- name: someOtherEnv
  variables:
   ...
```
but would have thrown an error during the resolution process. We could also keep the `- environmentName` format as an option, but fix it so that it's actually usable (i.e. doesn't just result in an error).

### DIFF
--- a/core/src/config/project.ts
+++ b/core/src/config/project.ts
@@ -133,13 +133,7 @@ export const environmentSchema = () =>
   })
 
 export const environmentsSchema = () =>
-  joi
-    .alternatives(
-      joiSparseArray(environmentSchema()).unique("name"),
-      // Allow a string as a shorthand for { name: foo }
-      joiSparseArray(joiUserIdentifier())
-    )
-    .description("A list of environments to configure for the project.")
+  joiSparseArray(environmentSchema()).unique("name").description("A list of environments to configure for the project.")
 
 export interface SourceConfig {
   name: string
--- a/core/src/config/common.ts
+++ b/core/src/config/common.ts
@@ -10,7 +10,7 @@ import Joi from "@hapi/joi"
 import Ajv from "ajv"
 import { splitLast } from "../util/util"
 import { deline, dedent } from "../util/string"
-import { cloneDeep } from "lodash"
+import { cloneDeep, isArray } from "lodash"
 import { joiPathPlaceholder } from "./validation"
 import { DEFAULT_API_VERSION } from "../constants"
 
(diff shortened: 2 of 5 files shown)


core/src/config/project.ts
```diff
@@ -133,13 +133,7 @@ export const environmentSchema = () =>
...
 
 export const environmentsSchema = () =>
-  joi
-    .alternatives(
-      joiSparseArray(environmentSchema()).unique("name"),
-      // Allow a string as a shorthand for { name: foo }
-      joiSparseArray(joiUserIdentifier())
-    )
-    .description("A list of environments to configure for the project.")
+  joiSparseArray(environmentSchema()).unique("name").description("A list of environments to configure for the project.")
```
edvald, line 136: This is a breaking change, right? We were allowing a list of strings previously.

edvald: Sorry, missed your note on that. I guess if it was previously broken, that's all good.

edvald (approved):

PR #2668, opened 2021-11-24
chore: always emit stack graph event in workflows

thsig:

This ensures that we also emit the stack graph event when the workflow only has script steps, or only has command steps that don't emit the stack graph event.


10ko (approved):
Just a comment, I am assuming we are already emitting the graph multiple times in case of a workflow with, for example, a `garden deploy` and a `garden test` in separate steps right?

thsig:
Yes we are.

10ko:
LGTM!

PR #2689, opened 2021-12-02
fix(workflows): fix provider resolution issue

thsig:

This fixes a regression introduced in the last release where providers would be resolved ahead of any script steps in a workflow, which caused problems when using a script to e.g. prepare auth credentials that were subsequently used when resolving providers.

We now don't emit a stack graph event for workflows that don't have a command step that resolves the config graph.


10ko (approved):

10ko (approved):
Thanks @thsig 🚀

PR #2701, opened 2021-12-14
feat(core): module varfiles

### LINKED ISSUE #2676: cannot override module variable with cli variable argument in the command of the exec module
### Current Behavior

setting a cli variable argument and passing it the command doesn't override the variable if it's defined at the module level.

 ===========

### Expected behavior

the cli variable should have precedence and override the variable set at the module level.

### LINKED ISSUE #2687: [FEATURE]: module varfiles
### Background / Motivation

we are using module templates to generate our jib modules and each generated module has a different set of build dependencies.  we would like to be able to use a varfile that is scoped to each module so we can reference it in the dependency property. this is a subset of: https://github.com/garden-io/garden/issues/2506 which describes a more fulsome variable hierarchy.

create a module varfile and have it available as variables in the module. 

```
kind: ModuleTemplate
name: common-rest-module-template
modules:
  - type: jib-container
      name: ${parent.name}-image
      description: Backend service
      include: [./*]
      exclude: [app/target/*]
      image: ${parent.name}
      varfile: dependencies.yaml
      build: 
        jdkVersion: 8
        projectType: maven
        dependencies: 
          ${yamlEncode(var.dependencies)}
```
```
# dependencies.yaml
dependencies:
  - dependency1
  - dependency2
```

we need a way to pass variables to a module from outside the garden application.  passing through files is the easiest way to do that.

this module varfile should have higher precedence that module variables.

thsig:

Added support for module varfiles. These are located at or under the module root, and take precedence over module variables (thus having the second-highest precedence behind CLI vars).

The path to the module varfile is set via the optional `module.varfile` field.

Also fixed an issue where CLI variables were overridden by module variables. CLI variables now correctly have the highest precedence.

Fixes #2687.
Fixes #2676.

### DIFF
--- a/core/src/config/project.ts
+++ b/core/src/config/project.ts
@@ -63,7 +63,7 @@ export interface EnvironmentConfig {
   production?: boolean
 }
 
-const varfileDescription = `
+export const varfileDescription = `
 The format of the files is determined by the configured file's extension:
 
 * \`.env\` - Standard "dotenv" format, as defined by [dotenv](https://github.com/motdotla/dotenv#rules).
@@ -492,7 +492,11 @@ export function resolveProjectConfig({
  * For project variables, we apply the variables specified to the selected environment on the global variables
  * specified on the top-level `variables` key using a JSON Merge Patch (https://tools.ietf.org/html/rfc7396).
  * We also attempt to load the configured varfiles, and include those in the merge. The precedence order is as follows:
+ *
  *   environment.varfile > environment.variables > project.varfile > project.variables
+ *
+ * Variables passed through the `--var` CLI option have the highest precedence, and are merged in later in the flow
+ * (see `resolveGardenParams`).
  *
  * For provider configuration, we filter down to the providers that are enabled for all environments (no `environments`
  * key specified) and those that explicitly list the specified environments. Then we merge any provider configs with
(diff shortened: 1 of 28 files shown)


core/src/config/project.ts
```diff
@@ -653,8 +657,16 @@ export function parseEnvironment(env: string): ParsedEnvironment {
   }
 }
 
-async function loadVarfile(projectRoot: string, path: string | undefined, defaultPath: string): Promise<PrimitiveMap> {
-  const resolvedPath = resolve(projectRoot, path || defaultPath)
+export async function loadVarfile(
+  // project root (when resolving project config) or module root (when resolving module config)
+  configRoot: string,
+  path: string | undefined,
+  defaultPath: string | undefined
```
edvald, line 664: I'd suggest deconstructing this, best to avoid consecutive optional args.

core/src/resolve-module.ts
```diff
@@ -82,6 +84,7 @@ export class ModuleResolver {
     const processingGraph = new DependencyValidationGraph()
 
     for (const rawConfig of this.rawConfigs) {
+      // console.log(`rawConfig: ${JSON.tringify(rawConfig, null, 2)}`)
```
edvald, line 87: 🧹

edvald (changes requested):
Looks good! Just some super minor things I noticed, otherwise seems solid 👌

thsig:
@edvald: This should be good to go now.

PR #2706, opened 2021-12-21
chore(tasks): remove startupProbes from task and test pods

twelvemo:
Exclude `startupProbes` from task and test pods.

Fixes #


thsig:
Looks good! Let's change the commit message from `chore(tasks)` to `fix(k8s)`, since this change applies to the k8s plugin (and not the framework-level task primitive).

thsig (approved):

PR #2714, opened 2022-01-04
feat(k8s): arbitrary shared secrets

mattpolzin:
This PR surfaces a new configuration field that can be used to specify arbitrary secrets that should be copied to each deployed environment. This is very similar to `imagePullSecrets` but it can be very useful to copy arbitrary secrets into namespaces  so that, for example, all developers can share a secret needed to authenticate with AWS S3 or Azure Blob Storage.

### DIFF
--- a/core/src/plugins/kubernetes/config.ts
+++ b/core/src/plugins/kubernetes/config.ts
@@ -136,6 +136,7 @@ export interface KubernetesConfig extends BaseProviderConfig {
   }
   forceSsl: boolean
   imagePullSecrets: ProviderSecretRef[]
+  copySecrets: ProviderSecretRef[]
   ingressHttpPort: number
   ingressHttpsPort: number
   ingressClass?: string
@@ -314,6 +315,12 @@ const imagePullSecretsSchema = () =>
     References to \`docker-registry\` secrets to use for authenticating with remote registries when pulling
     images. This is necessary if you reference private images in your module configuration, and is required
     when configuring a remote Kubernetes environment with buildMode=local.
+  `)
+
+const copySecretsSchema = () =>
+  joiSparseArray(secretRef).description(dedent`
+    References to secrets you need to have copied into all namespaces deployed to. These secrets will be
+    ensured to exist in the namespace before deploying any service.
   `)
 
 const tlsCertificateSchema = () =>
@@ -496,6 +503,7 @@ export const kubernetesConfigBase = () =>
       )
       .meta({ internal: true }),
     imagePullSecrets: imagePullSecretsSchema(),
+    copySecrets: copySecretsSchema(),
     // TODO: invert the resources and storage config schemas
     resources: joi
       .object()
--- a/core/src/plugins/kubernetes/secrets.ts
+++ b/core/src/plugins/kubernetes/secrets.ts
@@ -7,7 +7,7 @@
  */
 
 import { KubeApi } from "./api"
-import { ProviderSecretRef, KubernetesPluginContext, KubernetesProvider } from "./config"
+import { ProviderSecretRef, KubernetesPluginContext } from "./config"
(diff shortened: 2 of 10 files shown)


core/src/plugins/kubernetes/config.ts
```diff
@@ -136,6 +136,7 @@ export interface KubernetesConfig extends BaseProviderConfig {
   }
   forceSsl: boolean
   imagePullSecrets: ProviderSecretRef[]
+  sharedSecrets: ProviderSecretRef[]
```
edvald, line 139: I'd suggest naming this `copySecrets` instead, to make it more obvious and explicit what this does.

core/src/plugins/kubernetes/secrets.ts
```diff
@@ -124,10 +124,10 @@ export async function ensureSecret(api: KubeApi, secretRef: ProviderSecretRef, t
 }
 
 /**
- * Prepare references to imagePullSecrets for use in Pod specs, and ensure they have been copied to the target
- * namespace.
+ * Prepare references to imagePullSecrets and other shared secrets for use in Pod specs,
+ * and ensure they have been copied to the target namespace.
  */
-export async function prepareImagePullSecrets({
+export async function prepareSecrets({
```
edvald, line 130: Instead of making this function aware of both types of secrets, I'd suggest having this accept a list of secrets as a parameter and then calling it twice.

mattpolzin:
Any other documentation I should add to this PR other than the auto-generated stuff?

The lint failure points out some links I didn't go anywhere near so I assume it is just by chance that it is the first time it has been triggered to check for dead links since those links died.

core/src/plugins/kubernetes/config.ts
```diff
@@ -316,6 +317,11 @@ const imagePullSecretsSchema = () =>
     when configuring a remote Kubernetes environment with buildMode=local.
   `)
 
+const copySecretsSchema = () =>
+  joiSparseArray(secretRef).description(dedent`
+    References to secrets you need to have copied into all namespaces deployed to.
```
edvald, line 322: I'd maybe add something to the effect of _"These secrets will be ensured to exist in the namespace before deploying any service."_ just to make it quite clear what happens. 

Beyond that, I think this looks good! I might go ahead and add an integration test later, but can do that alongside some other stuff I'm working on (the integ test setup is a little annoying atm, so I won't put you through that).

core/test/unit/src/garden.ts
```diff
@@ -2551,7 +2551,7 @@ describe("Garden", () => {
       const garden = await makeTestGarden(projectRoot)
 
       const module = await garden.resolveModule("module-a")
-      const moduleCDep = { name: "module-c", copy: [] }
+      const moduleCDep = { name: "module-c", copy: [] }
```
mattpolzin, line 2554: The linter was pretty sure I needed to make this change even though it was unrelated to this PR. 🤷

core/src/plugins/kubernetes/config.ts
mattpolzin: I've added your suggestion verbatim because it sounds about right to me too.

> I might go ahead and add an integration test later [...] the integ test setup is a little annoying atm, so I won't put you through that

Thanks! I tend to be much more eager to write a test when there's an analogous test to base mine off of but I noticed `imagePullSecrets` isn't tested so I will definitely let you take on the task of adding test coverage since you are willing.

mattpolzin:
Is this PR likely pegged for the next Garden release or the one after that?

edvald:
We'll get it into tomorrow's release!

edvald:
@mattpolzin would you mind rebasing off the current master branch? Seems GitHub doesn't allow me to do it myself :)

edvald:
There we go! Thanks again @mattpolzin, much appreciated 🙏

PR #2715, opened 2022-01-04
chore: add garden image with ibmcloud cli tool

twelvemo:
It adds a container image that also includes the ibmcloud cli tool. Some users are using ibmcloud and need this image for running workflows in Garden Cloud. Usually we also add an image just with Garden and the respective cloud provider tool, in this case i am not sure if there is demand for that. But I can easily add that, if it makes sense.

Fixes #

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -365,6 +365,20 @@ jobs:
             TAG=gardendev/garden-aws-gcloud-azure:${CIRCLE_SHA1}
             docker build -t ${TAG} --build-arg TAG=${CIRCLE_SHA1} -f support/aws-gcloud-azure.Dockerfile dist/alpine-amd64
             docker push ${TAG}
+  build-docker-full:
+    <<: *node-config
+    steps:
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - checkout
+      - *attach-workspace
+      - docker_login
+      - run:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -365,6 +365,20 @@ jobs:
             TAG=gardendev/garden-aws-gcloud-azure:${CIRCLE_SHA1}
             docker build -t ${TAG} --build-arg TAG=${CIRCLE_SHA1} -f support/aws-gcloud-azure.Dockerfile dist/alpine-amd64
             docker push ${TAG}
+  build-docker-aws-gcloud-azure-ibmcloud:
```
edvald, line 368: 😬 

Could we instead call this something like "full"?

twelvemo: I agree that the name is too long, however it has the advantage that we know what is in there. I am a bit afraid that if we just call it `garden-full` it is not self-explanatory what this image includes on dockerhub.

twelvemo: I renamed it to `garden-full` in the circleci config and dockerhub, but kept the dockerfile as `aws-gcloud-azure-ibmcloud.Dockerfile` so we have some reference as to what is in that container image.

thsig (approved):

thsig (approved):

PR #2719, opened 2022-01-11
fix(core): properly handle sparse build deps

### LINKED ISSUE #2685: Conditional statement not working with null for jib-container
## Bug
Trying to use ```dependencies: ["${var.buildSharedLibs ? 'shared-libs' : null}"]``` in a jib-container module for **build.dependencies** property inside of a template but is throwing this error: **key .build.dependencies[0][name] must be a string**

### Current Behavior

Garden is expecting a string and does not handle null or undefined.

```
kind: ModuleTemplate
name: my-template
modules:
  - type: jib-container
    name: ${parent.name}-image
    description: Backend service
    include: [./*]
    exclude: [app/target/*]
    image: ${parent.name}
    extraFlags: [-Djib.useOnlyProjectCache=false, -Djib.to.tags=14.0.0-GCP-SNAPSHOT, -DskipTests=true, -Djib.from.image=gcr.io/distroless/java:8-debug, package]
    build: 
      jdkVersion: 8
      projectType: maven
      dependencies: ["${var.myCustomVar ? 'my-dependency-module' : null}"]
```

### Expected behavior

Garden should handle the null or undefined and not append the value to the dependency list.

### Suggested solution(s)

Do for jib-container module what was done in this ticket as well: https://github.com/garden-io/garden/pull/2264

### Your environment

* **OS:** linux distro

`garden version: 0.12.31-edge-a5649c83`

thsig:

Previously, adding an optional build dependency via a template string that sometimes resolves to null would throw an error when the string resolved to null.

The build dependencies array was always intended to be a sparse array, but due to the special handling we have to support both the shorthand (module name only) and the full (a map with the module name and other fields) formats, we needed to add a bit of extra logic to account for this use case.

Fixes #2685.

It's a bit awkward that we have to normalize / clean up the shape of the build dependencies array in two separate places, but it's perhaps acceptable given the incremental nature of the module resolution flow.

### DIFF
--- a/core/src/config/base.ts
+++ b/core/src/config/base.ts
@@ -20,6 +20,7 @@ import { WorkflowResource } from "./workflow"
 import { listDirectory } from "../util/fs"
 import { isConfigFilename } from "../util/fs"
 import { TemplateKind, templateKind, ModuleTemplateResource } from "./module-template"
+import { isTruthy } from "../util/util"
 
 export interface GardenResource {
   apiVersion: string
@@ -141,16 +142,17 @@ function prepareProjectConfig(spec: any, configPath: string): ProjectResource {
 }
 
 export function prepareModuleResource(spec: any, configPath: string, projectRoot: string): ModuleResource {
-  /**
-   * We allow specifying modules by name only as a shorthand:
-   *   dependencies:
-   *   - foo-module
-   *   - name: foo-module // same as the above
-   */
+  // We allow specifying modules by name only as a shorthand:
+  //   dependencies:
+  //   - foo-module
+  //   - name: foo-module // same as the above
+  // Empty strings and nulls are omitted from the array.
   let dependencies: BuildDependencyConfig[] = spec.build?.dependencies || []
 
   if (spec.build && spec.build.dependencies && isArray(spec.build.dependencies)) {
-    dependencies = spec.build.dependencies.map((dep: any) => (typeof dep === "string" ? { name: dep, copy: [] } : dep))
+    // We call `prepareBuildDependencies` on `spec.build.dependencies` again in `resolveModuleConfig` to ensure that
+    // any dependency configs whose module names resolved to null get filtered out.
+    dependencies = prepareBuildDependencies(spec.build.dependencies)
   }
 
   const cleanedSpec = {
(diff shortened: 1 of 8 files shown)


core/src/config/base.ts
```diff
@@ -202,6 +204,23 @@ export function prepareModuleResource(spec: any, configPath: string, projectRoot
...
+export function prepareBuildDependencies(buildDependencies: any[]): BuildDependencyConfig[] {
+  return buildDependencies
+    .map((dep) => {
+      const isString = typeof dep === "string"
+      if (!dep || (!isString && !dep.name)) {
+        return null
+      } else {
+        return isString ? { name: dep, copy: [] } : dep
+      }
+    })
```
edvald, line 220: This is quite difficult to parse. Would it not be simpler to first map to the canonical form and then filter based on whether the name is empty or not..?

thsig: Yeah, good point—I'll make that change.

thsig: How about now?

edvald: Hmm, let's see. If `dep` is a template string that resolves later to null, this won't work as expected. I think the filtering may need to happen quite a bit later, on reflection. That is, when the list is actually used.

thsig: It's not quite obvious, but it does work because of the re-filtering I added in line 335 in `core/src/resolve-module.ts`. I could add another test case to explicitly check for that (I manually verified that part).

edvald: That would be good to verify, I could totally see that break at a later point.

thsig: I've added a test case for this now (good thing that I did, it revealed a bug that I fixed). So I think we should be good.

core/test/unit/src/garden.ts
```diff
@@ -2546,6 +2546,16 @@ describe("Garden", () => {
       expect(module.allowPublish).to.equal(false)
     })
 
+    it.only("should filter out null build dependencies after resolving template strings", async () => {
```
edvald, line 2549: Left `it.only` in there. Seems fine otherwise.

edvald (changes requested):

core/test/unit/src/garden.ts
thsig: Thanks! I've removed it now.

PR #2725, opened 2022-01-13
feat(template): add indent helper function

### LINKED ISSUE #2659: [FEATURE]: Support resolution as objects of YAML template files
### Background / Motivation

I was writing a template kubernetes deployment manifest in the following format:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: web
  annotations:
    garden.io/generated: "false"
    garden.io/version: ${modules.build-web.version}
  labels:
    app.kubernetes.io/managed-by: garden
spec:
  selector:
    matchLabels:
      app: web
  template:
    metadata:
      labels:
        app: web
    spec:
      containers:
        - name: app
          image: ${modules.build-web.outputs.deployment-image-id}
          env:
            - name: APP_ENV
              value: ${environment.name}
          ports:
            - containerPort: 5555
          resources: ${var.webResources}
```
where `resources` can be customized based on the env. So it's a map object, not a string variable.

However, garden resolves the `webResources` as plain string, even though it's written as an object and thus it resolves it as `[object Object]` in the generated manifest.

User should be able to resolve objects and arrays in templated manifests the same way one can do it in garden-specific config yaml.

This will allow proper segregation of code and to not mix the kubernetes manifest along with the garden module config.

Going through the code, I understand that these non-garden files are read as plaintext and hence resolved like plain strings ([how `generateFiles` option works](https://github.com/garden-io/garden/blob/e84066909e73118e8adc58324945282e2e0ea385/core/src/resolve-module.ts#L445)). I propose we can try reading the files with `.yaml/.yml` extension as objects and then resolve them with `resolveTemplateStrings` ([like used with garden configs](https://github.com/garden-io/garden/blob/e84066909e73118e8adc58324945282e2e0ea385/core/src/resolve-module.ts#L302)).
If the file can't be read as object, then garden can smartly fallback to the existing approach of reading file as string.

edvald:
This is useful for indenting YAML strings, for example:

`${indent("some: multiline\nyaml: document", 2)}`

 ->

```
  some: multiline
  yaml: document
```

Closes #2659

### DIFF
--- a/core/src/template-string/functions.ts
+++ b/core/src/template-string/functions.ts
@@ -14,14 +14,20 @@ import Joi from "@hapi/joi"
 import { validateSchema } from "../config/validation"
 import { safeLoad, safeLoadAll } from "js-yaml"
 import { safeDumpYaml } from "../util/util"
+import indentString from "indent-string"
+
+interface ExampleArgument {
+  input: any[]
+  output: any // Used to validate expected output
+  skipTest?: boolean
+}
 
 interface TemplateHelperFunction {
   name: string
   description: string
   arguments: { [name: string]: Joi.Schema }
   outputSchema: Joi.Schema
-  exampleArguments: any[][]
-  exampleOutput?: any
+  exampleArguments: ExampleArgument[]
   fn: Function
 }
 
@@ -33,7 +39,7 @@ const helperFunctionSpecs: TemplateHelperFunction[] = [
       string: joi.string().required().description("The base64-encoded string to decode."),
     },
     outputSchema: joi.string(),
-    exampleArguments: [["bXkgdmFsdWU="]],
+    exampleArguments: [{ input: ["bXkgdmFsdWU="], output: "my value" }],
     fn: (str: string) => Buffer.from(str, "base64").toString(),
   },
   {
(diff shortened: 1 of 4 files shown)


core/src/template-string/functions.ts
```diff
@@ -57,6 +58,17 @@ const helperFunctionSpecs: TemplateHelperFunction[] = [
...
+  {
+    name: "indent",
+    description: "Indents each line in the given string with the specified number of spaces.",
+    arguments: {
+      string: joi.string().required().description("The string to indent."),
+      spaces: joi.number().required().integer().description("How many spaces to add on each line."),
+    },
+    outputSchema: joi.string(),
+    exampleArguments: [["some: multiline\nyaml: document", 2]],
+    fn: (str: string) => indentString(str, 2),
```
thsig, line 70: Shouldn't the indent be a variable here?

thsig (changes requested):
Looks good! Had one comment there (there's a constant that should probably be a variable/argument).

core/src/template-string/functions.ts
edvald: Indeed!

edvald: Fixed + improved the testing setup for all the helpers.

thsig (approved):

Issue #2717, opened 2022-01-05
dev mode unaware of pod re-initialization


jalper:
### Current Behavior

When I run garden dev and there's been a change to our garden config that will necessitate re-initializing our main kubernetes pod, garden will finish doing everything and get to "Waiting for code changes..." before the remote pod actually terminates and re-initializes.

### Expected behavior

I would expect garden to know that due to a config change, a pod re-initialization is needed, and to stay at "Processing" until that has been completed.

### Your environment

* **OS:** macOS
* **How I'm running Kubernetes:** GKE
`garden version` 0.12.33

PR #2742, opened 2022-01-24
fix(k8s): fix rollout status check for Recreate

### LINKED ISSUE #2630: Incorrect status handling for deployment strategy "recreate"
### Current Behavior

When Garden deploys a Kubernetes module with the deployment strategy set to `recreate`, it will report the `Deployment` resource as ready immediately, even though in reality the initial Deployment is still being deleted and the new one not even created yet.

The `kubectl rollout status` command report the correct statuses.

### Expected behavior

Garden should wait for the deployment to be ready and in general behave like the `kubectl rollout status` command.

### Reproducible example

Deploy a Kubernetes module with the deployment strategy set to `recreate`. 

### Suggested solution(s)

In general our "wait for resources" handler is modelled after the `kubectl rollout status` source code, but I suppose there's some case there that we're not handling correctly.

### Additional context

Not sure if relevant but this particular resource also didn't have probes set.

### Your environment

* **OS:** 

`garden version`

thsig:

@twelvemo and I paired on this.

Before this fix, the rollout status of deployments using the `Recreate` strategy would be reported as ready before the new pods were fully ready.

This was fixed by copying/following the latest logic from kubectl's rollout status checking logic, comparing the requested number of replicas from the deployment spec with the actual updated pod count.

See the [corresponding logic in kubectl](https://github.com/kubernetes/kubectl/blob/24d21a0ee42ecb5e5bed731f36b2d2c9c0244c35/pkg/polymorphichelpers/rollout_status.go).

Fixes #2630 and #2717.

### DIFF
--- a/core/src/plugins/kubernetes/status/workload.ts
+++ b/core/src/plugins/kubernetes/status/workload.ts
@@ -18,6 +18,7 @@ import {
   V1StatefulSetSpec,
   V1DeploymentStatus,
   CoreV1Event,
+  V1DeploymentSpec,
 } from "@kubernetes/client-node"
 import dedent = require("dedent")
 import { getCurrentWorkloadPods } from "../util"
@@ -224,8 +225,9 @@ async function getRolloutStatus(workload: Workload) {
     }
   } else {
     const status = <V1DeploymentStatus>workload.status
+    const deploymentSpec = <V1DeploymentSpec>workload.spec
 
-    const desired = status.replicas || 0
+    const desired = deploymentSpec.replicas || 1
     const updated = status.updatedReplicas || 0
     const replicas = status.replicas || 0
     const available = status.availableReplicas || 0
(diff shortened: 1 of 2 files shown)


core/src/plugins/kubernetes/status/workload.ts
```diff
@@ -224,8 +225,9 @@ async function getRolloutStatus(workload: Workload) {
     }
   } else {
     const status = <V1DeploymentStatus>workload.status
+    const deploymentSpec = <V1DeploymentSpec>workload.spec
 
-    const desired = status.replicas || 0
+    const desired = deploymentSpec.replicas || 0
```
edvald, line 230: I suspect the default value here should be 1

thsig: Hmm, what it the deployment is being scaled down to 0?

thsig: Should be good to go now.

edvald (approved):

PR #2743, opened 2022-01-25
feat(core): add --skip-dependencies CLI option

thsig:

Added a `--skip-dependencies` CLI option for the `deploy` and `test` commands.

When used with the `deploy` command, the specified services are deployed, but any additional services or tasks that they depend on are skipped.

This can be useful e.g. when your stack has already been deployed, and you want to deploy a subset of services in dev mode without redeploying any service dependencies that may have changed since you last deployed.

Similarly, for the `test` command, this option skips deploying/running any runtime dependencies of the requested test suites (not recommended for CI, but can be useful during inner-loop development in certain scenarios).

### DIFF
--- a/core/src/commands/deploy.ts
+++ b/core/src/commands/deploy.ts
@@ -70,6 +70,14 @@ export const deployOpts = {
   "skip": new StringsParameter({
     help: "The name(s) of services you'd like to skip when deploying.",
   }),
+  "skip-dependencies": new BooleanParameter({
+    help: deline`Deploy the specified services, but don't deploy any additional services that they depend on or run
+    any tasks that they depend on. This option can only be used when a list of service names is passed as CLI arguments.
+    This can be useful e.g. when your stack has already been deployed, and you want to deploy a subset of services in
+    dev mode without redeploying any service dependencies that may have changed since you last deployed.
+    `,
+    alias: "no-deps",
+  }),
   "forward": new BooleanParameter({
     help: deline`Create port forwards and leave process running without watching
     for changes. Ignored if --watch/-w flag is set or when in dev or hot-reload mode.`,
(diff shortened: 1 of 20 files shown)


core/src/tasks/build.ts
```diff
@@ -30,9 +30,8 @@ export interface BuildTaskParams {
 export class BuildTask extends BaseTask {
   type: TaskType = "build"
   concurrencyLimit = 5
-
-  private graph: ConfigGraph
-  private module: GardenModule
+  graph: ConfigGraph
```
thsig, line 33: Since I'm reading these previously private attributes in the new dependency helpers, I made these fields public for all the task classes (for consistency).

core/src/commands/deploy.ts
```diff
@@ -70,6 +70,14 @@ export const deployOpts = {
   "skip": new StringsParameter({
     help: "The name(s) of services you'd like to skip when deploying.",
   }),
+  "skip-dependencies": new BooleanParameter({
+    help: deline`Deploy the specified services, but don't deploy any additional services that they depend on or run
+    any tasks that they depend on. This option can only be used when a list of service names is passed as CLI arguments.
+    This can be useful e.g. when your stack has already been deployed, and you want to deploy a subset of services in
+    dev mode without redeploying any service dependencies that may have changed since you last deployed.
+    `,
+    alias: "nodeps",
```
thsig, line 79: How do we feel about this alias name? I'm definitely open to alternate suggestions.

edvald: I'd maybe go for `no-deps` instead, personally. Otherwise it parses as `node-ps` in my head :P

edvald (approved):

edvald:
Approved, just that one comment I had on the flag alias.

core/src/commands/deploy.ts
thsig: Yeah, agreed—I'll make that change now.

PR #2752, opened 2022-01-31
feat(core): debug logfiles

thsig:

We now output all debug-level and silly-level logs into logfiles in the project's `.garden/logs` directory.

Debug-level logs are logged to a debug logfile using standard text formatting.

Silly-level logs are logged to a JSONL-formatted logfile (where each line is a JSON string). These can later be used to provide dynamic rendering of command logs via the CLI (e.g. allowing the user to select the log level, and then filtering the JSON entries before rendering them to the terminal).

The names of these logfiles include the command name, log level and the time the command was executed (as an ISO timestamp).

Logfiles older than 7 days are also automatically cleaned up by this logic.

This will be useful for gathering debug information when something goes wrong in a Garden command that was run at a lower (less verbose) log level.

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -9,6 +9,8 @@
 import dotenv = require("dotenv")
 import { intersection, sortBy } from "lodash"
 import { resolve, join } from "path"
+import chalk from "chalk"
+import { pathExists } from "fs-extra"
 import { getBuiltinCommands } from "../commands/commands"
 import { shutdown, sleep, getPackageVersion, uuidv4, registerCleanupFunction } from "../util/util"
 import { Command, CommandResult, CommandGroup, BuiltinArgs } from "../commands/base"
@@ -45,12 +47,13 @@ import { DashboardEventStream } from "../server/dashboard-event-stream"
 import { GardenPluginReference } from "../types/plugin/plugin"
 import { renderError } from "../logger/renderers"
 import { CloudApi } from "../cloud/api"
-import chalk from "chalk"
 import { findProjectConfig } from "../config/base"
 import { pMemoizeDecorator } from "../lib/p-memoize"
 import { getCustomCommands } from "../commands/custom"
-import { pathExists } from "fs-extra"
 import { Profile } from "../util/profiling"
+import { prepareDebugLogfiles } from "./debug-logs"
+import { LogEntry } from "../logger/log-entry"
+import { JsonFileWriter } from "../logger/writers/json-file-writer"
 
 export async function makeDummyGarden(root: string, gardenOpts: GardenOpts) {
   const environments = gardenOpts.environmentName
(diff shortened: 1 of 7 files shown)


core/src/cli/cli.ts
```diff
@@ -99,11 +101,36 @@ ${renderCommands(commands)}
...
+      join(gardenDirPath, LOGS_DIR_NAME),
+      commandFullName
+    )
     const logConfigs: FileWriterConfig[] = [
+      {
+        logFilePath: join(gardenDirPath, LOGS_DIR_NAME, debugLogfileName),
+        truncatePrevious: true,
+        level: LogLevel.debug,
+      },
+      {
```
eysi09, line 129: Why not just have a single silly log file and a built-in Garden command for getting debug logs, filtering on the correct level and either write to the terminal or to a file. 

So you can essentially "replay" a Garden command at any log level. (We could even store output from the last "n" commands and have a drop down list in the TUI with a command + timestamp that you can pick from). 

So as not to block the this PR though, we could skip the debug file for now and implement the built-in command in a follow up PR.

eysi09 (commented):
I had one suggestion (that we can mostly implement in a follow up PR).

core/src/cli/cli.ts
eysi09: In any case it feels odd to have two create two files where once is a strict subset of the other.

edvald: This is pretty hidden debug "feature", I don't exactly mind a slightly-less-than-ideal implementation and just getting it done. We can make this nicer later.

Does the log level on each line get printed as part of the logfile output?

thsig: > a built-in Garden command for getting debug logs, filtering on the correct level and either write to the terminal or to a file.
>
> So you can essentially "replay" a Garden command at any log level. (We could even store output from the last "n" commands and have a drop down list in the TUI with a command + timestamp that you can pick from).

I like this feature idea. In the meantime, I think the debug-level logfile would be good to have in there, since the silly-level one is uncomfortably verbose for the user.

Once we get rid of the debug-level logfiles, they'll be automatically cleaned up by the pruning logic anyway.

What do you think, @edvald? Should we merge this one as is or should we get rid of the debug-level logfile first?

thsig: > Does the log level on each line get printed as part of the logfile output?

Not currently, no.

thsig: I replaced the silly-level logfile with a JSONL-formatted logfile (using a new `JsonFileWriter` for the log rendering), so I think we should be good to go here now.

core/src/cli/debug-logs.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  return {
+    debugLogfileName: makeLogfileName("debug", commandFullName),
+    jsonLogfileName: makeLogfileName("json", commandFullName),
+  }
+}
+
+// Example: build.debug.2022-01-31T09:33:55.001Z.log
+function makeLogfileName(logLevel: string, commandFullName: string): string {
+  return `${commandFullName.replace(" ", "-")}.${logLevel}.${new Date().toISOString()}.log`
```
eysi09, line 55: For the JSON file the log level should still be silly and the file extension should be `jsonl`.

eysi09 (changes requested):
One tiny comment, otherwise looks good.

core/src/cli/debug-logs.ts
thsig: Yeah, agreed. I'll make that change.

thsig: Ok, should be good to go now. Any further comments here?

eysi09 (approved):

PR #2754, opened 2022-01-31
improvement(core): better workflow error logging

thsig:

We now log any errors that come up in a workflow step command in the body log. The `run workflow` command's error logging was also improved and cleaned up.

The optional `skipCliErrorSummary` field isn't ideal, but we need to assume a bit more manual control over the error output/formatting for the `run workflow` command (because we don't want to log the errors twice).

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -579,15 +579,17 @@ ${renderCommands(commands)}
 
     let code = 0
     if (gardenErrors.length > 0) {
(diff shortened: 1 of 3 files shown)


eysi09 (approved):
LGTM. I agree that `skipCliErrorSummary` field is a bit awkward but the `run workflows` command is a bit of a special breed and the normal command behaviour doesn't really apply to it.

PR #2755, opened 2022-02-01
feat(core): --with-dependants option for build cmd

### LINKED ISSUE #2749: [FEATURE]: Garden should build all modules that have another module as a build dependency
### Background / Motivation

we currently have shared libraries that are configured as build dependencies. if a shared library is updated, we want to be able to trigger new builds of modules that are using this shared library. 

Garden should build all modules that have module-1 , module-2 or module-3 as dependencies (and so on, recursively “up” the dependency graph)..

we need to rebuild modules when their dependencies change

introduce --with-dependants flag, where Garden would build all modules that have module-1 , module-2 or module-3 as dependencies (and so on, recursively “up” the dependency graph).

thsig:

Adds a `--with-dependants` option to the `build` command.

When used, build dependants of the modules provided as CLI arguments (and their build dependants, recursively) will also be rebuilt.

Closes #2749.

### DIFF
--- a/core/src/commands/build.ts
+++ b/core/src/commands/build.ts
@@ -24,6 +24,9 @@ import { flatten } from "lodash"
 import { BuildTask } from "../tasks/build"
 import { StringsParameter, BooleanParameter } from "../cli/params"
 import { Garden } from "../garden"
+import { GardenModule } from "../types/module"
+import { uniqByName } from "../util/util"
+import { deline } from "../util/string"
 
 const buildArgs = {
   modules: new StringsParameter({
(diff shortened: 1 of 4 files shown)


core/src/commands/build.ts
```diff
@@ -32,12 +34,15 @@ const buildArgs = {
...
-  force: new BooleanParameter({ help: "Force rebuild of module(s).", alias: "f" }),
-  watch: new BooleanParameter({
+  "force": new BooleanParameter({ help: "Force rebuild of module(s).", alias: "f" }),
+  "watch": new BooleanParameter({
     help: "Watch for changes in module(s) and auto-build.",
     alias: "w",
     cliOnly: true,
   }),
+  "with-dependants": new BooleanParameter({
+    help: "Also rebuild modules that have build dependencies on one of the provided modules (recursively)",
```
edvald, line 44: Maybe s/provided/specified? Also, maybe worth noting that it's ignored if no modules are specified in arguments,

core/src/commands/build.ts
```diff
@@ -104,7 +109,18 @@ export class BuildCommand extends Command<Args, Opts> {
     await garden.clearBuilds()
 
     const graph = await garden.getConfigGraph({ log, emit: true })
-    const modules = graph.getModules({ names: args.modules })
+    const modulesFromArgs = graph.getModules({ names: args.modules })
+    let modules: GardenModule[]
```
edvald, line 113: Not that it matters a lot, but you could make this more concise with `let modules = graph.getModules({ names: args.modules })` and then reassigning if the option is set.

thsig:
I've addressed the comments above, should be good to go now.

edvald (approved):

PR #2756, opened 2022-02-01
chore(k8s): update to Mutagen v0.13.0

xenoscopic:

This PR updates Garden to use Mutagen v0.13.0.  While this still uses a custom fork of the Mutagen CLI, it now switches to using the official Mutagen agent release binaries.  These changes are important for (a) bringing in fixes and enhancements added after Mutagen v0.12.0-beta3 (on which Garden's fork was previously based) and (b) moving Garden towards using official Mutagen release binaries.

This PR also fixes an incorrect command invocation in the Mutagen plugin and switches to using the more idiomatic `MUTAGEN_DISABLE_AUTOSTART` environment variable to prevent autostart of the Mutagen daemon when invoking commands (allowing for removal of the `--auto-start` flag in Garden's fork of Mutagen).

CC @edvald @thsig

### DIFF
--- a/core/src/plugins/kubernetes/constants.ts
+++ b/core/src/plugins/kubernetes/constants.ts
@@ -24,7 +24,7 @@ export const inClusterRegistryHostname = "127.0.0.1:5000"
 export const gardenUtilDaemonDeploymentName = "garden-util-daemon"
 export const dockerDaemonDeploymentName = "garden-docker-daemon"
 
-export const k8sUtilImageName = "gardendev/k8s-util:0.5.1"
+export const k8sUtilImageName = "gardendev/k8s-util:0.5.2"
 export const dockerDaemonContainerName = "docker-daemon"
 export const skopeoDaemonContainerName = "util"
 
--- a/core/src/plugins/kubernetes/dev-mode.ts
+++ b/core/src/plugins/kubernetes/dev-mode.ts
@@ -36,7 +36,7 @@ import { joi, joiIdentifier } from "../../config/common"
 import { KubernetesPluginContext, KubernetesProvider } from "./config"
 import { isConfiguredForDevMode } from "./status/status"
 
-const syncUtilImageName = "gardendev/k8s-sync:0.1.1"
+const syncUtilImageName = "gardendev/k8s-sync:0.1.3"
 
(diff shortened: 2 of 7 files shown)


thsig:
Looks good to me (also did some manual testing to be sure).

Any comments on this one, @edvald?

xenoscopic:
I'm not sure what's going on with the `test-minikube-*` jobs.  It doesn't seem to be the same issue as previously.  [Here](https://app.circleci.com/pipelines/github/garden-io/garden/10729/workflows/b7c71897-d5b4-4ba6-9083-807407ff9ddc/jobs/158686?invite=true#step-110-4399) for example I see `Error: unable to connect to beta: unable to connect to endpoint: unable to dial agent endpoint: version handshake error: version mismatch`.  Is it possible the CI system is caching the `gardendev/k8s-sync` image on a `latest` tag that's become outdated?

I did update the sidecar image reference in several places where it was still referencing `gardendev/k8s-sync:0.1.1`.  I figured this had been inadvertently left outdated with the previous update to `gardendev/k8s-sync:0.1.2`, but now I'm wondering if it was intentional?

edvald:
I _think_ I see the issue. We also need to update the `k8s-util` image, which is based off the `k8s-sync` image. I've gone ahead and published a 0.5.2 version of that, but it does also need to be updated accordingly in the PR ([here](https://github.com/garden-io/garden/blob/1d3e30729c5cb40f4d2730fc0337cc16d1fea37b/core/src/plugins/kubernetes/constants.ts#L27) and [here](https://github.com/garden-io/garden/blob/1d3e30729c5cb40f4d2730fc0337cc16d1fea37b/images/k8s-util/garden.yml#L5)).

xenoscopic:
Ah, okay.  Sorry I missed that.  Let me go ahead and update the PR and let's see if those tests go through.

I'm not seeing the 0.5.2 tag on Docker Hub yet though - is it still pending in build pipeline somewhere?

edvald:
Oops, seems my connection dropped when i was pushing the image, didn't notice. It's there now.

xenoscopic:
Okay, updated and CI has now run. Seems like it fixed that particular issue, though the previous CI errors (seen elsewhere) are still present.

edvald (approved):

PR #2757, opened 2022-02-02
improve(k8s): update kubectl to v1.23.3

edvald:
This is to get in an upstream bug fix. Should work the same as before.


pabbott-starbucks:
Thank you, Jón! 🙏🏻

thsig (approved):

PR #2758, opened 2022-02-03
improvement(cloud): handle AEC for Helm services

10ko:
This PR changes the status check for the Helm plugin to mark resources that have been deployed via Helm and paused by Garden Cloud AEC functionality as `outdated`.
This is needed because when we scale down Helm workload resources from the CleanupRunners, Helm stays unaware of those changes and always returns the `ready` status.

Note: I might need help with testing this @thsig and we were also wondering how to pass a `gardenCloudContext` to plugin handlers in order to call this logic only if the user is logged in Garden Cloud and avoid unnecessary checks otherwise.

Fixes #

cc @twelvemo

### DIFF
--- a/core/src/plugins/kubernetes/helm/deployment.ts
+++ b/core/src/plugins/kubernetes/helm/deployment.ts
@@ -6,11 +6,18 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+import Bluebird from "bluebird"
 import { waitForResources } from "../status/status"
 import { helm } from "./helm-cli"
 import { HelmModule } from "./config"
 import { getChartPath, getReleaseName, getChartResources, getValueArgs, getBaseModule } from "./common"
-import { getReleaseStatus, HelmServiceStatus, getDeployedResources } from "./status"
+import {
+  getReleaseStatus,
+  HelmServiceStatus,
+  getRenderedResources,
+  getPausedResources,
+  gardenCloudAECPauseAnnotation,
+} from "./status"
 import { HotReloadableResource } from "../hot-reload/hot-reload"
 import { apply, deleteResources } from "../kubectl"
 import { KubernetesPluginContext, ServiceResourceSpec } from "../config"
@@ -70,7 +77,7 @@ export async function deployHelmService({
   const chartPath = await getChartPath(module)
 
   const releaseName = getReleaseName(module)
-  const releaseStatus = await getReleaseStatus({ ctx: k8sCtx, service, releaseName, log, devMode, hotReload })
+  const releaseStatus = await getReleaseStatus({ ctx: k8sCtx, module, service, releaseName, log, devMode, hotReload })
 
   const commonArgs = [
     "--namespace",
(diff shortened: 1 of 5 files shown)


core/src/plugins/kubernetes/helm/deployment.ts
```diff
@@ -96,6 +103,24 @@ export async function deployHelmService({
     log.silly(`Upgrading Helm release ${releaseName}`)
     const upgradeArgs = ["upgrade", releaseName, chartPath, "--install", ...commonArgs]
     await helm({ ctx: k8sCtx, namespace, log, args: [...upgradeArgs] })
+    try {
+      // TODO: check "Garden Cloud Context" and only run this if user is logged in
+      const pausedResources = await getPausedResources({ ctx: k8sCtx, module, namespace, releaseName, log })
+      Bluebird.all(
```
thsig, line 109: Aren't we missing an `await` here?

thsig, line 109: We should either `await` here or make sure to handle promise rejections (if the idea is to not wait for the `annotateResource` calls to finish, for performance reasons).

10ko: Oops, this is a mistake, we should `await` there. Good catch!

core/src/plugins/kubernetes/helm/deployment.ts
```diff
@@ -96,6 +103,29 @@ export async function deployHelmService({
...
     const upgradeArgs = ["upgrade", releaseName, chartPath, "--install", ...commonArgs]
     await helm({ ctx: k8sCtx, namespace, log, args: [...upgradeArgs] })
+
+    // If ctx.cloudApi is defined, the user is logged in and they might be trying to deploy to an environment
+    // that could have been paused by Garden Cloud AEC functionality. We therefore make sure to cleanup any
+    // dangling annotation created by Garden Cloud.
+    if (ctx.cloudApi) {
+      try {
+        const pausedResources = await getPausedResources({ ctx: k8sCtx, module, namespace, releaseName, log })
+        Bluebird.all(
```
thsig, line 113: Missing `await` here.

core/src/plugins/kubernetes/helm/deployment.ts
```diff
@@ -96,6 +103,29 @@ export async function deployHelmService({
     log.silly(`Upgrading Helm release ${releaseName}`)
     const upgradeArgs = ["upgrade", releaseName, chartPath, "--install", ...commonArgs]
     await helm({ ctx: k8sCtx, namespace, log, args: [...upgradeArgs] })
+
+    // If ctx.cloudApi is defined, the user is logged in and they might be trying to deploy to an environment
+    // that could have been paused by Garden Cloud AEC functionality. We therefore make sure to cleanup any
```
thsig, line 108: "by Garden Cloud" -> "by Garden Cloud's"

"cleanup" -> "clean up"

"annotation" -> "annotations"

core/src/plugins/kubernetes/helm/deployment.ts
```diff
@@ -96,6 +103,29 @@ export async function deployHelmService({
...
+            const { annotations } = resource.metadata
+            if (annotations) {
+              delete annotations[gardenCloudAECPauseAnnotation]
+              return api.annotateResource({ log, resource, annotations })
+            }
+            return
+          })
+        )
+      } catch (error) {
+        const errorMsg = `Failed to remove Garden Cloud AEC annotations for module: ${module.name}.`
```
thsig, line 124: While the service name and module name are the same for Helm modules, we should use the service name here (and say "annotations for service" in the error message), since services are the thing that's deployed (in Garden parlance).

core/src/plugins/kubernetes/helm/status.ts
```diff
@@ -59,8 +63,10 @@ export async function getServiceStatus({
 
   let deployedWithDevModeOrHotReloading: boolean | undefined
 
+  log.info("checking status")
```
thsig, line 66: I think this should be at the verbose or debug level.

core/src/plugins/kubernetes/helm/status.ts
```diff
@@ -206,6 +215,13 @@ export async function getReleaseStatus({
       ) {
         state = "outdated"
       }
+
+      // If ctx.cloudApi is defined, the user is logged in and they might be trying to deploy to an environment
+      // that could have been paused by Garden Cloud AEC functionality. We therefore make sure to check for
```
thsig, line 220: "Garden Cloud" -> "Garden Cloud's"

core/src/plugins/kubernetes/helm/status.ts
```diff
@@ -206,6 +215,13 @@ export async function getReleaseStatus({
...
+      // the annotations Garden Cloud adds to Helm deployments and statefulsets when pausing an environment.
```
thsig, line 221: "deployments and statefulsets" -> "Deployments and StatefulSets"

core/src/plugins/kubernetes/helm/status.ts
```diff
@@ -221,3 +237,48 @@ export async function getReleaseStatus({
     }
   }
 }
+
+/**
+ *  Returns Helm workload resources that have been marked as "paused" by the Garden Cloud AEC functionality
```
thsig, line 242: "by the Garden Cloud" -> "by Garden Cloud's"

thsig (changes requested):
Had a few minor comments there, looks great otherwise!

core/src/plugins/kubernetes/helm/status.ts
10ko: I think this should be removed altogether, I added it while debugging.

10ko:
@thsig thanks, I made the changes you requested.

thsig (approved):

PR #2761, opened 2022-02-03
Custom commands

edvald:
This adds the ability to define custom commands as part of any Garden project. See the added guide for details: https://github.com/garden-io/garden/blob/63b93ec60e62286b4fe7ecfd082e0065435dac7b/docs/advanced/custom-commands.md

When reviewing, please try making your own and see if you bump into any issues or get confused!

I'm flagging this as experimental just for the moment, since this sort of thing invites a fair number of edge cases and use cases we didn't consider up front, and getting real-world feedback will be important.

**Some additional notes on the PR:**

* As part of the process I started implementing a "spread" option for positional arguments, but went another way in the implementation, so finishing that will wait but I left those initial changes in.
* I had to remove the preemptive validation for workflow commands, and along the way I allowed more commands to run in workflows by making the criteria whether they're persistent or not.
* There are two separate smaller commits there, that I found necessary to get this done.

### DIFF
--- a/core/src/commands/custom.ts
+++ b/core/src/commands/custom.ts
@@ -1,0 +1,271 @@
+/*
+ * Copyright (C) 2018-2021 Garden Technologies, Inc. <info@garden.io>
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import Bluebird from "bluebird"
+import chalk from "chalk"
+import execa from "execa"
+import { apply as jsonMerge } from "json-merge-patch"
+import { cloneDeep, keyBy, mapValues, flatten } from "lodash"
+import { TreeCache } from "../cache"
+import { parseCliArgs, prepareMinimistOpts } from "../cli/helpers"
+import { BooleanParameter, globalOptions, IntegerParameter, Parameter, StringParameter } from "../cli/params"
+import { loadConfigResources } from "../config/base"
+import {
+  CommandResource,
+  customCommandExecSchema,
+  customCommandGardenCommandSchema,
+  CustomCommandOption,
+  customCommandSchema,
+} from "../config/command"
+import { joi } from "../config/common"
+import { CustomCommandContext } from "../config/template-contexts/custom-command"
+import { validateWithPath } from "../config/validation"
+import { ConfigurationError, GardenBaseError, InternalError, RuntimeError, toGardenError } from "../exceptions"
+import { VoidLogger } from "../logger/logger"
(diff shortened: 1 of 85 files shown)


core/src/commands/custom.ts
```diff
@@ -0,0 +1,271 @@
...
+    .filter((r) => {
+      if (r.kind !== "Command") {
+        return false
+      }
+
+      if (builtinNames.includes(r.name)) {
+        // tslint:disable-next-line: no-console
+        console.log(
+          chalk.yellow(
+            `Ignoring custom command ${r.name} because it conflicts with built-in command with the same name`
```
thsig, line 248: "conflicts with built-in command" -> "conflicts with a built-in command"

core/src/config/command.ts
```diff
@@ -0,0 +1,132 @@
...
+    env?: StringMap
+  }
+  gardenCommand?: string[]
+
+  variables: DeepPrimitiveMap
+}
+
+const argumentSchema = () =>
+  joi.object().keys({
+    name: joiIdentifier().required().description("Short name for the paramater."),
```
thsig, line 56: "paramater" -> "parameter"

core/src/commands/run/workflow.ts
```diff
@@ -298,40 +302,84 @@ function printResult({
...
   step,
 }: RunStepCommandParams): Promise<CommandResult<any>> {
-  const { command, rest } = pickCommand(getAllCommands(), step.command!)
+  let rawArgs = step.command!
 
-  if (!command) {
-    throw new ConfigurationError(`Could not find Garden command '${step.command!}`, {
-      step,
-    })
+  const builtinCommands = getAllCommands()
```
thsig, line 315: We should probably rename `getAllCommands` to `getBuiltinCommands` as of this PR.

docs/advanced/custom-commands.md
```diff
@@ -0,0 +1,151 @@
...
+  - --dev
+  - api,worker
+  - --log-level
+  - debug
+  - $concat: ${args.$all}  # Allow any arguments/options on top of the fixed ones above
+```
+
+Here we imagine a basic day-to-day workflow for a certain group of developers. The user simply runs `garden api-dev`. Before the Garden command starts, we update the submodules in the repo, and then we start `garden deploy` with some parameters that we tend to use or prefer.
+
+Of course this is just an example, but no doubt you can imagine some commands, parameters etc. that you use a lot and it would be nice to codify for you and your team. And this example only uses a fraction of what's possible! Read on for more and see what ideas come up.
```
thsig, line 32: "and it would" -> "and which would"

thsig (changes requested):
Looks great! Had a few minor text-related comments there, but the implementation makes sense.

thsig (approved):

PR #2772, opened 2022-02-05
improve(cli): faster initialization

edvald:
I did a bunch of profiling to work out where time was spent on basic
initialization, before actually doing anything meaningful. The changes
combined shave about 50% off the time to run e.g. `garden help` in my
testing.

The only semantic change is the removal of the fullscreen writer, which
was never really made usable and imported some heavy libraries.

Beyond that, it's a lot of changes to lazy-load libraries when needed,
and in particular to load plugin code later, and avoid loading sqlite
until actually necessary.

There was also a seemingly unnecessary call to clean up the Mutagen
daemon directly from the Garden.close() method. I'm not exactly sure
why that was added (cc @thsig), but if it's actually necessary we should find
another way to do it, instead of bleeding plugin code to the core Garden
class.


thsig:
I'm not sure why I added the Mutagen daemon cleanup to the `Garden.close()` method, but I think we should be OK to remove it, since we're also registering a cleanup handler that does basically the same thing (kills the daemon process and cleans up the temp directory).

core/src/vcs/git.ts
```diff
@@ -377,66 +393,78 @@ export class GitHandler extends VcsHandler {
...
-      } catch (err) {
-        // 128 = File no longer exists
-        if (err.exitCode === 128 || err.code === "ENOENT") {
-          // If the file is gone, we filter it out below
-          return { path: resolvedPath, hash: "" }
-        } else {
-          throw err
+    function ensureHash(entry: FileEntry, stats: Stats, cb: (err: Error | null, entry?: FileEntry) => void) {
+      if (entry.hash === "" || modified.has(entry.path)) {
+        // console.log("hash", entry.path, modified.has(entry.path), entry)
```
thsig, line 400: Remove `console.log`

thsig (changes requested):
Looks great! There's a commented-out line that needs removing, but otherwise I think we're good to go here.

core/src/vcs/git.ts
thsig: I've removed this line now. We'll merge once CI is green.

thsig (approved):

PR #2776, opened 2022-02-08
chore: update license headers

thsig:

It's now 2022, so we needed to update the license headers to reflect that.

Added an ignore rule for the `core/src/lib` directory, since we don't want to include copyright headers for library code we copy over.


edvald (approved):

edvald:
Spotting a linting error, that may have slipped through in another PR. Maybe worth fixing along the way?

thsig:
Ok, fixed—maybe it slipped in with the custom commands PR (since we reviewed that one in a hurry).

PR #2780, opened 2022-02-10
Fix slack link

sixhobbits:

* I couldn't sign up for the Slack from the existing link. First I had to sign up on k8s.slack.io and then find the channel at the kubernetes.slack.com link.

Fixes #


edvald (approved):
Thank you @sixhobbits  🙏

PR #2808, opened 2022-02-20
feat(core): support persistent local processes

eysi09:

This commit adds proper support for persistent local processes.

It introduces the 'persistent' flag to the service spec of the local exec
module type.

When the 'persistent' flag is set to true and Garden is in watch mode,
Garden does not wait for the local process to exit before considering
the service deployed.

Users can however set a status command that checks whether the process is
actually 'ready'.

It also adds support for streaming logs from local processes. This is
achieved by streaming the output of local procs to a file and then
tailing the file when the logs command is run. Turns out there's no
proper (i.e. efficient) npm library for this so the bulk of the logic is
actually around file tailing.

I wouldn't be surprised if we run into edge cases or platform specific
gotchas here but they should be easy to fix when they come up.

Fixes #


core/src/commands/logs.ts
```diff
@@ -273,12 +274,8 @@ export class LogsCommand extends Command<Args, Opts> {
...
         out += chalk.gray("[" + tags + "] ")
       }
-      if (originalColor) {
-        // If the line doesn't have ansi encoding, we color it white to prevent logger from applying styles.
-        out += hasAnsi(serviceLog) ? serviceLog : chalk.white(serviceLog)
-      } else {
-        out += style(serviceLog)
-      }
+      // If the line doesn't have ansi encoding, we color it white to prevent logger from applying styles.
+      out += hasAnsi(serviceLog) ? serviceLog : chalk.white(serviceLog)
```
eysi09, line 278: IMO this is not a breaking change, we're simply colouring the terminal output differently. The decision to colour it in the first place was a bad one—my bad :).

core/src/plugins/exec/exec.ts
```diff
@@ -471,24 +566,111 @@ export const getExecServiceStatus: ServiceActionHandlers["getServiceStatus"] = a
...
 
-  return { state: "ready", detail: { deployCommandOutput: result.all } }
+  return {}
+}
+
+export const deployExecService: ServiceActionHandlers["deployService"] = async (params) => {
+  const { module, service, log, ctx } = params
+
+  // TODO: Check if it's in watch mode in general
+  const watchMode = params.devMode || params.hotReload || params.ctx.command.opts["watch"]
```
eysi09, line 592: @thsig, how can I just check if Garden is in persistent mode in general? Is there some variable I can look up so that I don't need to piece it together from command line flags?

core/src/plugins/exec/exec.ts
```diff
@@ -471,24 +566,111 @@ export const getExecServiceStatus: ServiceActionHandlers["getServiceStatus"] = a
...
+  if (watchMode && service.spec.persistent) {
+    // FIXME: Not sure if this emitted anywhere.
+    ctx.events.on("abort", () => {
```
eysi09, line 595: @thsig, thoughts on how to best handle this. That is, kill the child process if Garden is "shutting down".

core/src/plugins/exec/logs.ts
```diff
@@ -0,0 +1,392 @@
...
+
+    let readStream: ReadStream
+    try {
+      readStream = createReadStream(this.logFilePath, { start: cursorStartPos })
+      const splitStream = split2()
+      const sinceSeconds = (since && parseDuration(since, "s")) || undefined
+      let currentLine = -1
+
+      splitStream.on("data", (line: string) => {
+        // TODO: Verify whether this works cross platform (and if not, handle different platforms).
```
eysi09, line 311: I'd love feedback on this one.

core/src/plugins/exec/exec.ts
edvald: There isn't anything like that passed to plugin handlers atm. It would make sense to add it to `ctx.command` though, by adding to this interface (and wherever the `commandInfo` variable is defined): https://github.com/garden-io/garden/blob/3326c3c50b7f28fd25e17f3d0f66e938f79e97cb/core/src/plugin-context.ts#L34

core/src/plugins/exec/exec.ts
edvald: For the moment, I suggest handling both this event and setting this flag: https://github.com/sindresorhus/execa#cleanup

The abort event is emitted here btw: https://github.com/garden-io/garden/blob/87a4bd30ee1926ac48904de5747cfab123296f01/core/src/commands/logs.ts#L138
Currently that is only called by the websocket server.

core/src/plugins/exec/exec.ts
```diff
@@ -102,6 +126,19 @@ export const execServiceSchema = () =>
...
+
+        If a \`statusCommand\` is specified, Garden will wait until until it returns a 0 exit code before
+        considering the service ready.
+      `),
+      // TODO: Set a default in v0.13.
+      timeout: joi.number().description(dedent`
+        The maximum duration (in seconds) to wait for a local script to exit. In the case of a persistent
+        local process, this is the maximum duration to wait for the \`statusCommand\` to return a 0
+        exit code, if it's set.
+      `),
```
edvald, line 141: I had previously imagined this to work a bit differently tbh, but I can see pros and cons.

My line of thinking was adding a `devMode` stanza with a `devMode.command` which is run when the service is in, well, dev mode (which of course implies a persistent process). That way the code path can be cleanly separated and we don't have to have as many if/then/else/that branches in both code and docs.

core/test/unit/src/logger/log-entry.ts
```diff
@@ -16,7 +16,7 @@ const logger: Logger = getLogger()
 
 beforeEach(() => {
   // tslint:disable-next-line: prettier
-  (logger["children"] as any) = []
+  ;(logger["children"] as any) = []
```
edvald, line 19: This probably breaks parsing (same in a few other files).

package-lock.json
```diff
@@ -0,0 +1,8834 @@
+{
```
edvald: Included by mistake, I'm assuming?

core/src/plugins/exec/logs.ts
```diff
@@ -0,0 +1,392 @@
...
+      if (fileSize > warnOnFileSize) {
+        this.log.warn(dedent`
+          Detected unusually large local log file for local service ${this.serviceName} at path ${this.logFilePath}.
+          Size is ${Math.floor(fileSize / (1024 * 1024))}MB. This can slow down local log streaming.
+
+          We recommend clearing the file by either deleting it or restarting the Garden process that started the ${
+            this.serviceName
+          } service.
+
+          If you see this message frequently, please open a GitHub issue at: ${getGitHubIssueLink(
```
edvald, line 114: Nice 👍

core/src/plugins/exec/logs.ts
```diff
@@ -0,0 +1,392 @@
...
+    this.logFilePath = logFilePath
+    this.resolve = null
+    this.retryIntervalMs = retryIntervalMs || defaultRetryIntervalMs
+    this.modified = false
+    this.isTailing = false
+    this.lastStreamedEntry = null
+    this.bytesRead = 0
+    this.watcher = null
+    // TODO: Use a typed event bus
+    this.events = new EventEmitter2()
```
edvald, line 96: We use EventEmitter3 in a couple of places I think, which does just that.

core/src/plugins/exec/logs.ts
edvald: I'll confess I'm not 100% clear on the flow here, but I wonder if we can't avoid some complexity by using a library for this type of thing? e.g. https://github.com/lucagrulla/node-tail

eysi09: I started using it. But decided to read the source for good measure and turns out it reads the entire file into memory. Synchronously :) 

Hence the hand-woven implementation.

core/src/plugins/exec/logs.ts
eysi09: Ooh nice.

eysi09: I actually updated the code a bit before noticing your comment. But I'll switch to EE3.

core/test/unit/src/logger/log-entry.ts
eysi09: Ah yes, my linter went rogue. Forgot to fix.

eysi09: Actually I think this just happened when I ran `yarn lint`. Not in my IDE.

core/src/plugins/exec/logs.ts
```diff
@@ -0,0 +1,446 @@
...
+          const dateSince = new Date()
+          dateSince.setSeconds(dateSince.getSeconds() - sinceSeconds)
+
+          if (entry.timestamp < dateSince) {
+            return
+          }
+        }
+
+        lastStreamedEntry = entry
+        void this.stream.write(entry)
```
edvald, line 389: Note: Unrelated to the PR but I think we should move away from this style of emitting logs at some point.

core/src/plugins/exec/logs.ts
edvald: I think the implementation seems fine, good stuff. Hard to anticipate some edgy cases maybe, so might need an iteration or two to sort out fully after using it in the wild, but that should be alright.

eysi09: Yup, that's what I was anticipating. But my request for feedback wasn't super clear. 

I was basically referring to this line `bytesRead += Buffer.byteLength(line + "\n")` and wondering whether I could safely assume that the new line character is a single byte across platforms. 

That is how we write the file. I.e. `JSON.stringify(data) + "\n"`.

core/src/plugins/exec/logs.ts
eysi09: Yeah, it's a bit awkward to test. 

How would you suggest doing it instead?

eysi09:
I think I've addressed all the PR comments.

core/src/plugins/exec/exec.ts
```diff
@@ -286,6 +336,62 @@ function getDefaultEnvVars(module: ExecModule) {
   }
 }
 
+async function rotateLogFile(logFilePath: string) {
```
edvald, line 339: I guess "rotate" is loosely defined here :) Maybe add a TODO if the intent is to actually back up the old file.

core/src/plugins/exec/exec.ts
```diff
@@ -471,27 +579,122 @@ export const getExecServiceStatus: ServiceActionHandlers["getServiceStatus"] = a
...
+    const startedAt = new Date()
+
+    if (service.spec.statusCommand) {
+      let ready = false
+
+      while (!ready) {
+        const now = new Date()
+        const timeElapsedSec = (now.getTime() - startedAt.getTime()) / 1000
+        // TODO: Remove once we set a default value that the timeout spec in v0.13.
+        const timeout = service.spec.timeout || persistentLocalProcDefaultTimeoutSec
```
edvald, line 650: Isn't this in effect a default timeout?

core/src/plugins/exec/exec.ts
```diff
@@ -471,27 +579,122 @@ export const getExecServiceStatus: ServiceActionHandlers["getServiceStatus"] = a
...
+    })
+    localProcs[key] = {
+      proc,
+      key,
+      service,
+    }
+
+    const startedAt = new Date()
+
+    if (service.spec.statusCommand) {
```
edvald, line 643: I wonder if it makes sense to use the same status command for devMode. I'd imagine it to work completely differently in most scenarios. I'd actually be okay with omitting this for the moment, but since you've already rigged it up, I'd at least suggest making a separate field under devMode.

core/src/plugins/exec/logs.ts
edvald: I see. I think this should be fine, especially since we control the file that we're tailing.

core/src/plugins/exec/logs.ts
edvald: I'd suggest using `ctx.events.emit(...)` instead, and doing appropriate validation on that emitter. But as I said, we can change that interface later.

core/src/types/plugin/service/getServiceLogs.ts
```diff
@@ -38,12 +38,19 @@ export const serviceLogEntrySchema = () =>
       serviceName: joi.string().required().description("The name of the service the log entry originated from."),
       timestamp: joi.date().required().description("The time when the log entry was generated by the service."),
       msg: joi.string().required().description("The content of the log entry."),
-      level: joi.string().description(
-        dedent`
-          The log level of the entry. The 'info' level should be reserved for logs from the service proper.
```
edvald, line 43: Curious, what's the motivation to change this to a number?

core/src/plugins/exec/exec.ts
eysi09: Yeah I can do that. Should do the same for the timeout then. 

But out of curiosity, do you have a use case in mind for a module that has a deploy command (and a corresponding status command) that exits when not in dev mode but is persistent in dev mode?

The only use case I had in mind was just a local version of a module that's typically remote. In which case the config would usually just be:

```yaml
kind: Module
type: my-module-local
services:
  - name: my-module-local
     deployCommand: [] # empty on purpose
     devMode: 
       statusCommand: ["./check-status.sh"]
       deployCommand: ["yarn", "run", "dev"]
```

(Which is why I went with `persistent` to begin with, so we could re-use the `deployCommand` and `statusCommand` fields.)

core/src/plugins/exec/exec.ts
eysi09: The idea was that this would only apply when `persistent: true` and therefore wouldn't be a breaking change. I think a better solution would be to move the `timeout` field to the `devMode` as per my comment above.

core/src/types/plugin/service/getServiceLogs.ts
eysi09: Ugh, yeah this because the `LogLevel` is an enum. 

So when writing the log entry it wouldn't have the correct shape if the level was just a string like "info" and you had to workaround the compiler. And the same applies when reading the entries. 

So you couldn't lean on types in either case and I figured it'd be easier to just change the Joi spec. 

Of course the "correct" solution would be to change the level into a union type but that's kind of out of scope.

eysi09:
I updated the PR again (see latest commit).

core/src/plugins/exec/exec.ts
edvald: I think it's actually okay to make the no-op deploy command implicit, and allow skipping it. In that scenario just log a gray line saying there's no deploy command or something. What do you think?

core/src/types/plugin/service/getServiceLogs.ts
edvald: That's all good, might revert it or change it at a later date when we make the public SDK though.

core/src/plugins/exec/exec.ts
eysi09: Yeah I agree. I actually meant to fix it but it just slipped my mind.

edvald (approved):

PR #2809, opened 2022-02-21
fix: updated brew command in dependency install script

mkhq:

The brew command has changed format, this means `brew cask ...` as used in the installation script does not work anymore.

Fixes #


edvald (approved):

edvald:
Just noticed after merging @mkhq and no big deal at all, but for these types of things we tend to use `chore(something): ` as the commit message prefix, because that doesn't arrive in the release changelog automatically.

PR #2810, opened 2022-02-22
Ensure the type field contains 'templated' in a template invocation.

ohookins:
Small missing text from a module invocation in the docs.

N/A

I understand that the docs are auto-generated from some sources but can't see the source that drives this particular file... please let me know if I edited the wrong one.


edvald (approved):
Thanks @ohookins, good catch! This is the right doc, indeed. Generally only the stuff in the Reference section is auto-generated.

PR #2813, opened 2022-02-23
improve(build): update pkg version and sign mac executables

edvald:
This is a precursor to updating to Node.js 14 and adding proper support for ARM64 builds.

No rush to merge this ahead of tomorrow's release btw @thsig.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ orbs:
   # Configuration for our node jobs
   node-config: &node-config
     docker:
-      - image: gardendev/circleci-runner:12.22.2-0
+      - image: gardendev/circleci-runner:12.22.10-0
         environment:
           <<: *shared-env-config
     resource_class: medium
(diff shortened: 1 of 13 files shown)


.circleci/config.yml
```diff
@@ -775,6 +775,35 @@ jobs:
           command: |
             sudo -E ./bin/garden plugins local-kubernetes cleanup-cluster-registry --root core/test/data/test-projects/container --env cluster-docker
 
+  test-macos:
+    macos:
+      xcode: "12.5.1"
+    steps:
+      - checkout
+      - *attach-workspace
+      # - run:
```
thsig, line 784: Remove comment?

cli/bin/garden
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env node --max-old-space-size=4096
```
thsig, line 1: Is this a deliberate reversion of the changes introduced in https://github.com/garden-io/garden/pull/2812/files?

edvald: It's not a reversion, it moves the option flag to `bin/garden` instead. This didn't work for some reason when calling that (which btw does not affect the dist build).

.circleci/config.yml
edvald: Doesn't really matter, figure we might need this later.

thsig (approved):

PR #2821, opened 2022-02-24
feat(core): add template string sha256 function

ohookins:

We have the need for stable but unguessable identifiers for some aspects of our Garden stacks. Unfortunately uuidv4 as a template string function satisfies the latter requirement, but would presumably change on every deploy so we'll end up with a continually changing stack. Using a hashing function would satisfy both requirements.

This PR just adds a template string `sha256` based on the crypto package in the standard library.

N/A


edvald (approved):
Excellent, thank you @ohookins!

PR #2831, opened 2022-02-28
chore(docs): fix GitBook configuration

edvald:
It was pointed out to me by GitBook support that we hadn't set the `root` field in the config file. That might explain the timeout.


edvald:
Note: The test errors seem to be some CircleCI problem.

thsig (approved):

PR #2835, opened 2022-03-03
feat: pulumi plugin

thsig:

The pulumi plugin wraps the pulumi CLI for deploying and destroying pulumi stacks.

The pulumi module type supports merging variables and varfiles with pulumi configs (via the `pulumiVariables` and `pulumiVarfiles` config fields). Pulumi modules that deploy sensitive resources can be protected by setting `autoApply: false` and/or `allowDestroy: false`.

The `cacheStatus` config flag uses stack tags to implement fast and lightweight service status checking for pulumi stacks.

Also included are plugin commands for cancelling, refreshing, and reimporting stacks (in dependency order, when called for more than one service).

See the usage guide and added reference docs for more details on how to use and configure the pulumi plugin.

Finally, we've also moved the terraform plugin out of the Core package into the `plugins` directory.


docs/advanced/pulumi.md
```diff
@@ -0,0 +1,118 @@
...
+# If the pulumi stack doesn't exist already when deploying, create it
+createStack: true 
+# Cache deploys based on the Garden service version (see the section below)
+cacheStatus: true
+# These variables will be merged into the stack config before deploying or previewing
+pulumiVariables:
+  my-variable: pineapple
+# Variables defined in varfiles will also be merged into the stack config in declaration
+# order (and take precedence over variables defined in this module's pulumiVariables).
+pulumiVarfiles: [my-default-varfile.yaml, dev.yaml]
```
tedchang77, line 53: For simple pulumi programs, you many not need or want the additional flexibility (and complexity) that pulumiVarfiles and pulumiVars provide. You should have the ability to opt out and use a hand crafted Pulumi.<stack>.yaml that is not generated. if both of these directories are not set, we shouldn't generate an empty Pulumi.<stack>.yaml.

vvagaytsev (changes requested):
LGTM. I left a couple of comments. By the way, it would be nice to add `tslint.json` to `pulumi` and `terraform` plugins now and apply the linter's suggested changes in this PR.

If it's too time consuming now, let's then do it in a separate PR. I can do that.

plugins/pulumi/config.ts
```diff
@@ -0,0 +1,168 @@
...
+
+
+export interface PulumiModuleSpec {
+  allowDestroy: boolean
+  autoApply: boolean
+  createStack: boolean
+  dependencies: string[]
+  pulumiVariables: DeepPrimitiveMap
+  pulumiVarfiles: string[]
+  orgName?: string | null
```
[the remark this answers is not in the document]
thsig: Yeah, it has different semantics than `undefined` (i.e. not providing a value at all). See:
https://github.com/garden-io/garden/blob/pulumi/plugins/pulumi/config.ts#L125-L126

vvagaytsev (dismissed):

plugins/pulumi/handlers.ts
```diff
@@ -0,0 +1,222 @@
...
+export const deployPulumiService: ServiceActionHandlers["deployService"] = async ({
+  ctx,
+  log,
+  module,
+  service,
+}) => {
+  const provider = ctx.provider as PulumiProvider
+  const pulumiModule: PulumiModule = module
+  const pulumiParams = { log, ctx, provider, module: pulumiModule }
+  const { autoApply, deployFromPreview, cacheStatus } = pulumiModule.spec
```
vvagaytsev, line 122: Variable `cacheStatus` is unused here.

vvagaytsev (approved):

plugins/pulumi/handlers.ts
vvagaytsev: I'll fix it in another PR with linters config.

PR #2838, opened 2022-03-07
chore: fixed build errors for arm64 arch

vvagaytsev:
Co-authored-by: Thorarinn Sigurdsson <thorarinnsigurdsson@gmail.com>

This PR fixes build errors on arm64 machines.


edvald:
This is all good, with the caveat that this only really solves an issue with internal development on macOS M1s and doesn't actually implement proper ARM builds. That will take a much more comprehensive update, but should be viable now that we've updated pkg and taken some steps in our CI to support that.

edvald (approved):

PR #2839, opened 2022-03-07
fix(core): add missing await for user prompt

### LINKED ISSUE #2826: Manually deploying to a production environment fails with Command aborted
### Current Behavior

We don't currently automate the deployment of our production systems, this is done by hand each time by running garden build --env=staging followed by garden deploy --env=staging giving us a chance to review what is built / deployed to ensure we are doing it correctly.

When running the build command now with the latest version we get the following:

image

### Expected behaviour

I expect to be prompted for a Y or N answer, and the process will continue if Y is pressed.

### Reproducible example

No reproduceable example as yet, but happy to show you our process over a call.

### Workaround

using garden build --env=staging --yes allows the build to function, but we prefer having the human input at this point

### Your environment

* **OS:** Windows 11 Pro
* **How I'm running Kubernetes:** Azure Kubernetes Service

`garden version`  0.12.35

vvagaytsev:
Added missing `await` for user prompt.

Fixes #2826

### DIFF
--- a/core/src/commands/base.ts
+++ b/core/src/commands/base.ts
@@ -276,7 +276,7 @@ export abstract class Command<T extends Parameters = {}, U extends Parameters = 
           Are you sure you want to continue? (run the command with the "--yes" flag to skip this check).
 
       `)
-      const answer: any = userPrompt({
+      const answer = await userPrompt({
         name: "continue",
         message: defaultMessage,
         type: "confirm",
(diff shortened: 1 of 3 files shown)


edvald (dismissed):

vvagaytsev:
> Looks good, I think that one was on me! Just one note on the 2nd commit message. These we generally file as `chore(...)` so that they don't pop up in user-facing changelogs.

Thanks @edvald! Now it's fixed.

thsig (approved):

thsig (approved):

PR #2844, opened 2022-03-10
fix(k8s): sync error with space in user home directory name

edvald:


thsig:
Looks good! Do we need to do the same escaping for the local sync paths, or is that unnecessary?

thsig (approved):

PR #2852, opened 2022-03-16
fix: maven download for windows

### LINKED ISSUE #2849: Fetching Maven on Windows returns invalid signature
### Current Behavior

When fetching maven in Windows with Garden, an error is returned reporting that the zip could not be extracted. 

Screenshot 2022-03-15 at 15 54 05

From the command output it is apparent Garden is expecting a zip file, but receiving a tarball. The mismatch seems to be stemming from this configuration: https://github.com/garden-io/garden/blob/master/plugins/jib/maven.ts#L47

### Expected behavior

Garden should fetch Maven on Windows.
### Workaround

None.
### Your environment

* **OS:** 

`garden version`

twelvemo:
It fixes an issue where garden tries to extract a downloaded maven tarball with zip on windows.

Fixes #2849 

### DIFF
--- a/plugins/jib/maven.ts
+++ b/plugins/jib/maven.ts
@@ -43,9 +43,11 @@ export const mavenSpec: PluginToolSpec = {
       platform: "windows",
       architecture: "amd64",
       ...spec,
+      url: "https://archive.apache.org/dist/maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.zip",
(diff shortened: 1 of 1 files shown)


twelvemo:
@thsig the maven download now works, however when i run the downloaded maven cli from the .garden directory i get a warning about a missing `$JAVA_HOME` environment variable. But i assume we are taking care of that somewhere?
Screenshot 2022-03-18 at 18 12 13

edvald (approved):
LGTM 👍  We do indeed set JAVA_HOME automatically in the plugin.

PR #2855, opened 2022-03-18
improvement(core): add 'string' template function

eysi09:

This template function converts the given value to a string. Useful when
using boolean or numeric values as env vars in K8s Pod specs since
K8s only allows string values.

For e.g.:

```
container:
  env: 
    name: PORT
    value: ${string(var.port)}
```

Fixes #


edvald:
Looks good, although I wonder if we should decide on a naming convention for casting functions. I'd have suggested `toString`, what do you think?

edvald:
Also, this would be a `feat(template)` commit imo :)

eysi09:
Yeah `toString` was my first reaction but I looked around and saw `lower` and `upper` and kind of thought it was consistent with that. 

Happy to change though. 

Also updated the commit message.

thsig (dismissed):
I'm fine with `string` here (since we've already got `lower` and `upper`). If we later decide to standardize on one or the other, we can always add a bunch of aliases.

Also, given that terseness is very valuable in this context (i.e. people constructing concise one-liner expressions with template strings), shorter function names are generally preferable to longer ones.

thsig (approved):

PR #2856, opened 2022-03-20
fix(exec): properly handle empty lines in local service logs

eysi09:

Before this fix, log streaming would simply stop if a line was empty.
This could e.g. happen on double new-lines.

This commit fixes the issue and adds a unit test.

Fixes #

### DIFF
--- a/core/test/unit/src/plugins/exec/exec.ts
+++ b/core/test/unit/src/plugins/exec/exec.ts
@@ -725,6 +725,76 @@ describe("exec plugin", () => {
             },
           ])
         })
+        it("should handle empty log lines", async () => {
+          // This services just echos a string N times before exiting.
+          const service = graph.getService("dev-mode-with-empty-log-lines")
+          const actions = await garden.getActionRouter()
+          const res = await actions.deployService({
+            devMode: true,
+            force: false,
+            hotReload: false,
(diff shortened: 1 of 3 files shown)


core/test/unit/src/plugins/exec/exec.ts
```diff
@@ -725,6 +725,76 @@ describe("exec plugin", () => {
             },
           ])
         })
+        it("should handle empty log lines", async () => {
```
eysi09, line 728: Behind every unit test is a story.

examples/local-service/frontend/app.js
```diff
@@ -2,8 +2,23 @@ const express = require('express');
 const request = require('request-promise')
 const app = express();
 
+const longAssStr = `No issues found\n\n`
```
edvald, line 5: Want to clean this up a bit or leave in? :)

thsig: Haha yeah. Your call. I'm all for sprinkling a bit of slam poetry around the codebase if the spirit moves someone. So long as that's the deliberate intent.

thsig: I went ahead and removed these.

thsig (approved):

PR #2877, opened 2022-03-29
fix(cli): bad error message when dependency fails in run task/test cmd

edvald:
This would mask failures in task/test dependencies and create an opaque error message.


thsig (approved):

thsig (changes requested):
Looks good! There are a few failing unit tests in there.

edvald:
Working on those, will fix and get a fresh review after.

PR #2879, opened 2022-03-30
chore: test release script corrections

vvagaytsev:
@thsig recommended disabling some steps in the test script, those steps have been commented out.
Also, the dev mode test was modified to use the `vote` project instead of the `demo-project`, because the demo project doesn't have a dev mode sync configuration, and the vote does. It allows testing code changes syncing in dev mode.

### DIFF
--- a/scripts/test-release.sh
+++ b/scripts/test-release.sh
@@ -89,8 +89,12 @@ test_release() {
   echo "→ Respond to the prompts to see if the create command works"
   echo ""
   ${garden_release} create module
+
+  cd ..
+  cd vote
   echo ""
-  echo "→ Running 'garden dev' in demo project - exits after 1 minute"
+  echo "→ Running 'garden dev' in vote project - exits after 1 minute"
+  echo "→ Try e.g. to update this file: ${garden_root}/examples/vote/vote/src/views/Home.vue"
   echo ""
   timeout 1m ${garden_release} dev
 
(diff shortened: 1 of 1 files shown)


scripts/test-release.sh
```diff
@@ -102,14 +106,14 @@ test_release() {
...
 
-  echo ""
-  echo "→ Running 'garden dev' in vote-helm project - exits after 2 minutes. Use the chance to change services and test the dashboard."
-  echo "→ Try e.g. to change the status in the POST method in this file: ${garden_root}/examples/vote-helm/api-image/app.py"
-  echo "→ It should break the integ test"
-  echo ""
-  cd ..
-  cd vote-helm
-  timeout 2m ${garden_release} dev
+#  echo ""
```
thsig, line 109: Let's remove these commented-out lines.

scripts/test-release.sh
```diff
@@ -84,13 +84,17 @@ test_release() {
   echo "→ Run a command in the prompt (ls, for example) and see if the TTY behaves as expected."
   echo ""
   ${garden_release} exec backend /bin/sh
+#  echo ""
```
thsig, line 87: We commented this test out because it messes with the "run `garden dev` in `demo-project`" test case below.

Now that we're running `garden dev` in the `vote` project instead, we can uncomment this again (since the created module shouldn't interfere with subsequent tests anymore).

thsig (changes requested):
Looks good! Had a couple of comments there.

scripts/test-release.sh
vvagaytsev: Done

thsig (approved):

PR #2881, opened 2022-03-30
fix(k8s): handle spaces in kubectl path provided to Mutagen

xenoscopic:

This PR fixes an issue where a space in the `kubectl` executable path (e.g. in a home directory name) would cause Garden's Mutagen exec transport to fail.  It reverts a previous fix that performed escaping on the path and updates the Mutagen fork to parse around this issue.

I can't find an assigned issue; the problem was reported via Slack.

The corresponding fix in Garden's Mutagen fork is garden-io/mutagen@1c930de.  This is a temporary fix.  Once we add custom transports in Mutagen v0.15, we can abolish the exec transport and provide the `kubectl` path more robustly.

CC @edvald @thsig


thsig (approved):
This fixed the issue—thanks so much for this!

PR #2885, opened 2022-04-05
fix(template): error in && operator during partial initial resolution

edvald:
We would prematurely resolve a template string as false, where referenced
values would be resolvable later.

### DIFF
--- a/core/src/template-string/template-string.ts
+++ b/core/src/template-string/template-string.ts
@@ -557,7 +557,7 @@ function buildBinaryExpression(head: any, tail: any) {
   }, head)
 }
 
-function buildLogicalExpression(head: any, tail: any) {
+function buildLogicalExpression(head: any, tail: any, opts: ContextResolveOpts) {
   return tail.reduce((result: any, element: any) => {
     const operator = element[1]
     const leftRes = result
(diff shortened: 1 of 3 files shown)


eysi09:
Bumping into this when testing this branch:

```
Failed resolving one or more modules:

logs-maintenance-cronjob: Invalid template string (${ command.params["dev-mode"] && t…): Cannot read property 'allowPartial' of undefined
partition-maintenance-cronjob: Invalid template string (${ command.params["dev-mode"] && t…): Cannot read property 'allowPartial' of undefined
cleanup-cronjobs: Invalid template string (${ command.params["dev-mode"] && t…): Cannot read property 'allowPartial' of undefined
```

core/src/template-string/template-string.ts
```diff
@@ -566,24 +566,28 @@ function buildLogicalExpression(head: any, tail: any) {
     switch (operator) {
       case "&&":
         if (leftRes && leftRes._error) {
-          if (leftRes._error.type === missingKeyExceptionType) {
+          if (!opts.allowPartial && leftRes._error.type === missingKeyExceptionType) {
```
eysi09, line 569: Looks like `opts` isn't always defined. Seeing `Cannot read property partial of undefined` when running Garden from this branch.

eysi09 (changes requested):

eysi09:
Assigning myself.

core/src/template-string/template-string.ts
edvald: Have you rebuilt the parser? :P

eysi09: Ah whoops. I thought it happened automatically when running `yarn dev`. Did a full rebuild and now it works.

eysi09 (approved):

core/src/template-string/template-string.ts
edvald: It should actually, does for me... That's odd.

PR #2896, opened 2022-04-08
chore(release): use Node 14 in npm release action

thsig:

The GitHub action that updates our Core NPM package has been failing for a bit (probably since we upgraded to Node 14). This PR fixes that.


thsig:
Note: I tried running this through (by temporarily changing the triggers to apply to any push) and it went through successfully. So this should work as is.

twelvemo (approved):

PR #2902, opened 2022-04-14
fix: update nginx ingress controller for kind

twelvemo:
The NGINX ingress controller deployed for Kind clusters was out of date and contained several incompatible API types to Kubernetes versions >= 1.22. 
This PR updates the resource to the [latest manifests](https://kind.sigs.k8s.io/docs/user/ingress/#ingress-nginx) as proposed by the kind team. 
The updated manifests work for Kind clusters >= 1.21. In case we are looking for backward compatibility, we can look into adjusting the API versions of `batch/v1` to `batch/v1beta1`.

Fixes #

### DIFF
--- a/static/kubernetes/system/nginx-kind/garden.yml
+++ b/static/kubernetes/system/nginx-kind/garden.yml
@@ -2,317 +2,634 @@ kind: Module
 name: nginx-kind
 description: Special manifests for installing nginx on KinD clusters
 type: kubernetes
+variables:
+  namespace: ingress-nginx
 manifests:
   - apiVersion: v1
     kind: Service
     metadata:
-      name: ingress-nginx
-      namespace: ${var.namespace}
       labels:
+        app.kubernetes.io/component: controller
+        app.kubernetes.io/instance: ingress-nginx
         app.kubernetes.io/name: ingress-nginx
         app.kubernetes.io/part-of: ingress-nginx
+        app.kubernetes.io/version: 1.1.3
+      name: ingress-nginx-controller
+      namespace: ${var.namespace}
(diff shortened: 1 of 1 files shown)


twelvemo:
The tests for kind-15 and kind-16 are failing, which makes sense because the api versions the updated ingress controller is using are not yet supported in these versions. Should we maybe do a bit of housekeeping and get rid of the test suites for very old kubernetes versions  or are those still needed @eysi09 @thsig ?

eysi09:
[We're committed to supporting the latest six stable versions of K8s.](https://docs.garden.io/misc/faq#what-versions-and-variations-of-kubernetes-does-garden-support). So we can safely remove tests for 1.15 and 1.16. 

But we need to be backwards compatible up to that point. We can't just break things for existing Kind users. Didn't we have something in place the converts the manifests at "apply time" (cc @thsig)?

thsig:
I don't think we do, actually.

What we have (to my knowledge, at least) is a backwards-compatible reimplementation of `kubectl`'s `--apply-prune` option (see [e.g. here](https://github.com/garden-io/garden/blob/2e0f20c95be220ce597431930168947805bb0c73/core/src/plugins/kubernetes/kubectl.ts#L25-L44) and the code below that makes use of `versionedPruneKinds`).

So that logic is about ensuring that we're able to delete resources in a backwards-compatible way, but doesn't auto-convert between API versions during the apply.

Or did you have some other piece of logic in the codebase in mind?

jamesmeneghello:
Will this be progressed? Using garden with new versions of Kind is currently broken, as the ingress controller can't be installed.

PR #2904, opened 2022-04-19
feat: add skip-comments option to cli create command

Orzelius:
This pr adds a `--skip-comments` flag to create commands which removes all commented out sections from the generated files.
Motivation for the change is that the [getting started guide](https://docs.garden.io/getting-started/2-initialize-a-project) has all the comments removed so it would make sense to add that option to the cli as well.

Peek 2022-04-19 14-37

* Please note if I need to update any documentation or add a note about the flag to the [getting started guide](https://docs.garden.io/getting-started/2-initialize-a-project)
* I gave up on adding this as an interactive option as current logic doesn't differentiate between boolean parameters being undefined or false. That change would require a bit of a bigger rewrite

### DIFF
--- a/core/src/commands/create/helpers.ts
+++ b/core/src/commands/create/helpers.ts
@@ -7,6 +7,15 @@
  */
 
 import { pathExists, readFile, writeFile } from "fs-extra"
+import { BooleanParameter } from "../../cli/params"
+
+export const createBaseOpts = {
+  noComments: new BooleanParameter({
+    alias: "no-comments",
+    help: "Set to true to disable comment generation.",
+    defaultValue: false,
+  }),
+}
 
 export async function addConfig(configPath: string, yaml: string) {
   let output = yaml
--- a/core/src/commands/create/helpers.ts
+++ b/core/src/commands/create/helpers.ts
@@ -10,7 +10,7 @@ import { pathExists, readFile, writeFile } from "fs-extra"
 import { BooleanParameter } from "../../cli/params"
 
 export const createBaseOpts = {
-  noComments: new BooleanParameter({
+  "skip-comments": new BooleanParameter({
     help: "Set to true to disable comment generation.",
     defaultValue: false,
   }),
--- a/core/src/commands/create/helpers.ts
+++ b/core/src/commands/create/helpers.ts
@@ -11,7 +11,6 @@ import { BooleanParameter } from "../../cli/params"
 
 export const createBaseOpts = {
   noComments: new BooleanParameter({
-    alias: "no-comments",
     help: "Set to true to disable comment generation.",
     defaultValue: false,
   }),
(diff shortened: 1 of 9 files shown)


core/src/commands/create/helpers.ts
```diff
@@ -7,6 +7,15 @@
  */
 
 import { pathExists, readFile, writeFile } from "fs-extra"
+import { BooleanParameter } from "../../cli/params"
+
+export const createBaseOpts = {
+  noComments: new BooleanParameter({
+    alias: "no-comments",
```
vvagaytsev, line 14: Here `no-comments` is not a valid alias. The aliases should be single characters. Running a command with this alias causes an error:

```
> g create project -no-comments
Unrecognized option flag -n
Invalid value for option --output: "-comments" is not a valid argument (should be any of "json", "yaml")
```

I guess we can omit the alias for this option, wdyt?

vvagaytsev: We already have some broken aliases in our codebase, and we're going to remove them in the next minor release, likely in 0.13 because it's a breaking change, see https://www.notion.so/gardendev/Breaking-change-plan-wishlist-v0-11-625c6261e15542acbe916a73aa8b1505

core/src/commands/create/helpers.ts
```diff
@@ -7,6 +7,15 @@
  */
 
 import { pathExists, readFile, writeFile } from "fs-extra"
+import { BooleanParameter } from "../../cli/params"
+
+export const createBaseOpts = {
+  noComments: new BooleanParameter({
```
vvagaytsev, line 13: It seems like we prefer `kebab-lower-case` for command line opts and args, so we might need to rename this option to `no-comments` or `nocomments`. The former might be trickier to implement.

vvagaytsev:
Great job and nice feature! Congrats on the first PR! 🎉
I left some comments, please check.

vvagaytsev (changes requested):

vvagaytsev:
> Please note if I need to update any documentation or add a note about the flag to the [getting started guide](https://docs.garden.io/getting-started/2-initialize-a-project)

Yes, that would be nice to have, see https://github.com/garden-io/garden/blob/master/docs/getting-started/2-initialize-a-project.md

core/src/commands/create/helpers.ts
Orzelius: good point, removed in [81c65c4](https://github.com/garden-io/garden/pull/2904/commits/81c65c4ecdb4ef9fe597b20d871d42c1f2acb978)

(I added it in the first place because of the bad examples in the code base :cactus:)

Orzelius:
> > Please note if I need to update any documentation or add a note about the flag to the [getting started guide](https://docs.garden.io/getting-started/2-initialize-a-project)
> 
> Yes, that would be nice to have, see https://github.com/garden-io/garden/blob/master/docs/getting-started/2-initialize-a-project.md

all and any critique warmly welcomed [0a1236a](https://github.com/garden-io/garden/pull/2904/commits/0a1236a5fe7c0cd9add0f10dbc7503045527eaf2)

vvagaytsev (approved):

PR #2907, opened 2022-04-21
fix: ingress for microk8s and k8s 1.22

Orzelius:

* Fixes ingress not working for microk8s
* adds a warning if the networking api is at `networking.k8s.io/v1` and no ingressclass exists for the ingresses created by garden 

image

(the warning appears twice because two containers are created which both have ingress)

### DIFF
--- a/core/src/plugins/kubernetes/container/ingress.ts
+++ b/core/src/plugins/kubernetes/container/ingress.ts
@@ -22,9 +22,7 @@ import { LogEntry } from "../../../logger/log-entry"
 import chalk from "chalk"
 
 // Ingress API versions in descending order of preference
-// NOTE: We currently prefer the v1beta1 API version if available, since users may not have an IngressClass set up
-// correctly
-export const supportedIngressApiVersions = ["networking.k8s.io/v1beta1", "networking.k8s.io/v1", "extensions/v1beta1"]
+export const supportedIngressApiVersions = ["networking.k8s.io/v1", "networking.k8s.io/v1beta1", "extensions/v1beta1"]
 
 interface ServiceIngressWithCert extends ServiceIngress {
   spec: ContainerIngressSpec
--- a/core/src/plugins/kubernetes/container/ingress.ts
+++ b/core/src/plugins/kubernetes/container/ingress.ts
@@ -17,7 +17,7 @@ import { ConfigurationError, PluginError } from "../../../exceptions"
 import { ensureSecret } from "../secrets"
 import { getHostnamesFromPem } from "../../../util/tls"
 import { KubernetesResource } from "../types"
-import { ExtensionsV1beta1Ingress, V1Ingress, V1Secret } from "@kubernetes/client-node"
+import { ExtensionsV1beta1Ingress, V1Ingress, V1IngressClass, V1Secret } from "@kubernetes/client-node"
 import { LogEntry } from "../../../logger/log-entry"
 import chalk from "chalk"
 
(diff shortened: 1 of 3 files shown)


core/src/plugins/kubernetes/container/ingress.ts
```diff
@@ -22,9 +22,7 @@ import { LogEntry } from "../../../logger/log-entry"
 import chalk from "chalk"
 
 // Ingress API versions in descending order of preference
-// NOTE: We currently prefer the v1beta1 API version if available, since users may not have an IngressClass set up
-// correctly
-export const supportedIngressApiVersions = ["networking.k8s.io/v1beta1", "networking.k8s.io/v1", "extensions/v1beta1"]
+export const supportedIngressApiVersions = ["networking.k8s.io/v1", "networking.k8s.io/v1beta1", "extensions/v1beta1"]
```
Orzelius, line 25: This is the biggest change in this pr and I would definitely like to get it approved by people more experienced

core/src/plugins/kubernetes/container/ingress.ts
```diff
@@ -70,6 +70,24 @@ export async function createIngressResources(
...
+    const ingressclassWithCorrectName = ingressclasses.items.find(
+      (ic) => ic.metadata.name === provider.config.ingressClass
+    )
+    if (!ingressclassWithCorrectName) {
+      log.warn(
+        chalk.yellow(`ingressclass with name "${provider.config.ingressClass}" was not found, ingress might not work`)
+      )
+    }
+  }
+
```
Orzelius, line 90: Didn't really know where to put this code so it ended up here. Please say if it should be a separate function (something like `warnIfIngressClassDoesNotExist`?).

core/src/plugins/kubernetes/container/ingress.ts
vvagaytsev: I'm also not too experienced with k8s configs. @twelvemo could you review Albert's changes, please?

twelvemo: We spoke about this today and the change here is that we would now prefer `networking.k8s.io/v1` over the other API versions. So far we have been using `networking.k8s.io/v1beta1` whenever it was still possible (up until Kubernetes version 1.22). The newer `networking.k8s.io/v1`API version uses an `ingressClass` and not everyone had setup `ingressClass` yet. I believe as of today this should be more widely spread and as long as we make sure this change works with Garden it should be fine in my opinion. 
In any local kubernetes solution that uses Garden's `setupIngressController: nginx` this works, because we deploy an `ingressClass` resource with the name `nginx` or it is deployed by one of the add-ons. 
For kind clusters it depends on #2902  that updates the ingress controller for kind and adds an `ingressClass` to it.
For remote Kubernetes it would also work when using the Garden deployed ingress controller.
There is a chance that in a remote kubernetes cluster a custom ingress controller is installed that does not expose an `ingressClass`. In this case Garden would warn that an ingress was deployed that does not match any `ingressClass`.

core/src/plugins/kubernetes/container/ingress.ts
vvagaytsev: Looks good to me. Should we use "IngressClass" in the warning message for slightly better readability? :)

vvagaytsev (approved):
LGTM. Should we merge and release both this and #2902? cc @eysi09 @thsig @twelvemo

PR #290, opened 2018-09-21
refactor: rename `init env` cmd to `init` and remove login cmd

edvald:
This change folds the login handler into the prepareEnvironment
handler. The getEnvironmentStatus handler can now indicate if
prepareEnvironment will need user input, in which case the framework
may error out and guide the user to run `garden init`. This allows
providers to transparently initialize in cases where they don't need
user input, but retain the flexibility to ask for user input when
needed (e.g. for logins, asking for sudo permissions etc.).

PR #2927, opened 2022-05-04
fix(core): improve ingress warning

Orzelius:

Follow-up to #290
* make message more informative
* don't spam the error
* don't show when ingressClass has not been configured
* adds tests

Still shows up twice when the very first garden initialization is done as all the plugins are initialized twice for some reason on the very first init


Orzelius:
somehow this broke all e2e tests, will look into it tomorrow

vvagaytsev:
> somehow this broke all e2e tests, will look into it tomorrow

@Orzelius I saw the same issues in my reverse proxy branch.
So it looks like some env slowness in the CircleCI pipeline. Let's check tomorrow.

Orzelius:
Turns out the tests failing was just a momentary lapse of reason on circleci's part it seems 🤷

vvagaytsev (approved):

PR #2933, opened 2022-05-05
fix(cli): potential OOM error during module and error serialization

edvald:
Basically we were normalizing and serializing a recursive data structure, which could end up with a massive amount of data, which could kill the process in certain scenarios.

### DIFF
--- a/core/src/commands/scan.ts
+++ b/core/src/commands/scan.ts
@@ -24,7 +24,7 @@ export class ScanCommand extends Command {
     const graph = await garden.getConfigGraph({ log, emit: false })
     const modules = graph.getModules().map((m) => {
       return {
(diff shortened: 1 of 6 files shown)


vvagaytsev (approved):

core/src/commands/scan.ts
```diff
@@ -24,7 +24,7 @@ export class ScanCommand extends Command {
     const graph = await garden.getConfigGraph({ log, emit: false })
     const modules = graph.getModules().map((m) => {
       return {
-        ...omit(m, ["_config", "cacheContext", "serviceNames", "taskNames"]),
+        ...omit(m, ["_config", "cacheContext", "serviceNames", "taskNames", "buildDepdendencies"]),
```
carolinapc, line 27: Hi @edvald ! Is there a typo here? "buildDependencies"

PR #2935, opened 2022-05-05
improve(core): more efficient module resolution

edvald:
This especially improves resolution for large graphs, by improving concurrency
during the resolution process.


vvagaytsev (approved):

vvagaytsev:
@edvald  looks good! One comment about the commit message format. It looks like it has to start with "improvment", at least this is what our contributing guide says: https://github.com/garden-io/garden/blob/master/CONTRIBUTING.md

When I prepared recent releases the commits starting with "improve:" were not automatically included in the changelog, so I had to add those manually.

PR #2937, opened 2022-05-06
Transitive dependency handling

Orzelius:

See `test: add tests for build logic after modifications to modules and deps` commit for the cases that this pr fixes.

TLDR: fix transitive dependencies

If you have a dep tree like this, where A depends on B which depends on C:
```
A -> B -> C
```

If you change C and rebuild A you would expect the result to be different, but before this pr it wasn't as build dependencies were checked only one level down and only file changes were checked.

With this pr hash changes are the basis for updating.

Big part of the logic changes were lifted from @thsig branch https://github.com/garden-io/garden/tree/exec-build-deps-fix


core/test/unit/src/garden.ts
```diff
@@ -4468,8 +4584,8 @@ describe("Garden", () => {
         const moduleB = graph.getModule("module-b")
         const moduleC = graph.getModule("module-c")
         expect(moduleA.version.versionString).to.not.equal(moduleAVersionString)
-        expect(moduleB.version.versionString).to.equal(moduleBVersionString)
-        expect(moduleC.version.versionString).to.equal(moduleCVersionString)
+        expect(moduleB.version.versionString).to.not.equal(moduleBVersionString) // B depends on A so it changes as well
+        expect(moduleC.version.versionString).to.not.equal(moduleCVersionString) // C depends on B so it changes as well
```
edvald, line 4588: I don't think this is correct, since the version hashes are actually hard-coded.

docs/reference/commands.md
```diff
@@ -2261,12 +2261,8 @@ modules:
 
       # The version of each of the dependencies of the module.
       dependencyVersions:
+        # A Stack Graph node (i.e. module, service, task or test) version.
         <name>:
-          # The hash of all files belonging to the Garden module.
-          contentHash:
-
-          # List of file paths included in the version.
-          files:
```
edvald, line 2269: This looks wrong, what should the value schema look like? Also, this may be considered breaking.

vvagaytsev:
A note about commit message format. Some commit headers look a bit misleading. But that's not an issue if the PR is squashed before merging. This looks like a good case to do squashing because the PR fixes one product-level bug.

Also, we use commit headers to automatically generate the changelog for release notes. This means that each commit starting with `fix:` or `feat:`  will appear as a standalone bug fix or feature in the auto-generated changelog.

Of course, it is natural to use proper headers inside a PR branch and to keep the PR as a sequence of small enough and well-grained commits. It makes the CR much easier. But before merging it makes sense to revisit the scope of the PR and make sure that the in-branch features and fixes are "large enough" to be product-level fixes or features listed in the changelog. Otherwise, if the PR adds a single product-level bug fix or feature, it makes sense to do squashing.

core/test/unit/src/garden.ts
Orzelius: I don't get it, where are they hard coded?

These versions changes makes perfect sense to me. [Previous logic](https://github.com/garden-io/garden/pull/2937/files#diff-5f17dbb72e9d791ba8b214f07c86f9d435613e277ebf9c65b4399283efd73b32L178-L200) only  accounted for file changes resulting in B not changing as files were not changed. New logic uses dependency hashes instead, thus also updating B [that depends on A](https://github.com/garden-io/garden/blob/dependency-fix/core/test/data/test-projects/fixed-version-hashes-1/module-b/garden.yml#L7) and then updating C [that depends on B](https://github.com/garden-io/garden/blob/dependency-fix/core/test/data/test-projects/fixed-version-hashes-1/module-c/garden.yml#L8).

edvald: Those version strings being compared in the test are hard-coded, is what I was referring to. But I think I just read it wrong, you're quite right, looks good 👍

docs/reference/commands.md
Orzelius: The value schema is now just the [versionStringSchema](https://github.com/garden-io/garden/pull/2937/commits/1be24a1634bcb04e92c8fa062dfc2bc3c79aabd6#diff-1d9ef1ec0b89b1389227b0b5944d4735857d3fe2cb5b85884e0f78e6760fc643R560). I think the reason it's empty is due to the default being `{}` for the sake of generating this doc

core/test/unit/src/commands/build.ts
```diff
@@ -233,4 +244,156 @@ describe("BuildCommand", () => {
       "stage-build.module-d": {},
     })
   })
+
+  // adds a third level of dependants and tests rebuild logic after changes to modules
+  describe("tracking changes and rebuilding logic", () => {
```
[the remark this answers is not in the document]
Orzelius: makes sense, fixed in [da7b5c0](https://github.com/garden-io/garden/pull/2937/commits/da7b5c0f6ebb0b62a42410116d6155fda9e4a4c7)

docs/reference/commands.md
Orzelius: added a bit of a better desc in [9558a59](https://github.com/garden-io/garden/pull/2937/commits/9558a59452681c1d5f07219db57f9663d0116d4b)

Orzelius: > Also, this may be considered breaking

Yeah, if anybody is relying on the current schema it would break stuff for them

vvagaytsev (approved):

PR #2941, opened 2022-05-10
fix: deploy --force flag

vvagaytsev:
This PR fixes issue with `garden deploy --force` command when some deploy-level dependant services are not re-deployed in force mode.

The issue can be reproduced with [demo-project example](https://github.com/garden-io/garden/tree/master/examples/demo-project) by running the command `garden deploy --force`.

Here is the example output (with some extra logging):
```
✔ providers                 → Getting status... → Cached
   ℹ Run with --force-refresh to force a refresh of provider statuses.
frontend                  → new DeployTask(uid=b467dac0-d057-11ec-9efe-8352ecdfe161, force=true, serviceName=frontend)
backend                   → new DeployTask(uid=b4687700-d057-11ec-9efe-8352ecdfe161, force=true, serviceName=backend)
backend                   → new DeployTask(uid=b468ec34-d057-11ec-9efe-8352ecdfe161, force=false, serviceName=backend)
✔ backend                   → Getting build status for v-05b195be19... → Already built
✔ frontend                  → Getting build status for v-9f87fc75ef... → Already built
✔ backend                   → Deploying version v-2a8c9419c1... → Already deployed
   backend                   → DeployTask(uid=b468ec34-d057-11ec-9efe-8352ecdfe161, force=false, serviceName=backend).process()
   → Ingress: http://demo-project.local.app.garden/hello-backend
✔ frontend                  → Deploying version v-13c6cf9522... → Done (took 3.8 sec)
   frontend                  → DeployTask(uid=b467dac0-d057-11ec-9efe-8352ecdfe161, force=true, serviceName=frontend).process()
   ℹ frontend                  → Resources ready
   → Ingress: http://demo-project.local.app.garden/hello-frontend
   → Ingress: http://demo-project.local.app.garden/call-backend
```

The "backend" deploy-task is created twice:
1. First one (`new DeployTask(uid=b4687700-d057-11ec-9efe-8352ecdfe161, force=true, serviceName=backend)`) is created in [`initialTasks`](https://github.com/garden-io/garden/blob/latest-release/core/src/commands/deploy.ts#L208) in the method `DeployCommand.action()`.
2. Second one (`new DeployTask(uid=b468ec34-d057-11ec-9efe-8352ecdfe161, force=false, serviceName=backend)`) is created as a deploy-level dependency of a "frontend" deploy-task in the method [`DeployTask.resolveDependencies()`](https://github.com/garden-io/garden/blob/latest-release/core/src/tasks/deploy.ts#L104).

It looks like the second task replaces the original one, is it expected behavior @edvald @thsig?

If yes, then the fix is just to propagate the original task's `force` flag to the dependant deploy-tasks, as [we do for `TestTask`s](https://github.com/garden-io/garden/blob/latest-release/core/src/tasks/test.ts#L117).

Fixes #

Note, there is some extra debug logging in the PR, it will be removed before merging.

### DIFF
--- a/core/src/tasks/deploy.ts
+++ b/core/src/tasks/deploy.ts
@@ -105,7 +105,7 @@ export class DeployTask extends BaseTask {
         // statuses and results.
         return [statusTask, ...buildTasks, ...getServiceStatusDeps(this, deps), ...getTaskResultDeps(this, deps)]
       } else {
-        return [statusTask, ...buildTasks, ...getDeployDeps(this, deps, false), ...getTaskDeps(this, deps, false)]
+        return [statusTask, ...buildTasks, ...getDeployDeps(this, deps, this.force), ...getTaskDeps(this, deps, false)]
       }
     }
   }
--- a/core/src/tasks/deploy.ts
+++ b/core/src/tasks/deploy.ts
@@ -59,6 +59,10 @@ export class DeployTask extends BaseTask {
     hotReloadServiceNames,
   }: DeployTaskParams) {
     super({ garden, log, force, version: service.version })
+    log.info({
+      section: service.name,
+      msg: `new DeployTask(uid=${this.uid}, force=${force}, serviceName=${service.name})`,
+    })
     this.graph = graph
     this.service = service
     this.forceBuild = forceBuild
@@ -159,6 +163,11 @@ export class DeployTask extends BaseTask {
       status: "active",
       section: this.service.name,
       msg: `Deploying version ${version}...`,
+    })
+
+    log.info({
+      section: this.service.name,
+      msg: `DeployTask(uid=${this.uid}, force=${this.force}, serviceName=${this.service.name}).process()`,
     })
 
     if (!this.force && status.state === "ready" && (version === status.version || devModeSkipRedeploy)) {


vvagaytsev:
I'll try to prepare an integration test for it a bit later.

thsig:
Now that I think about this in detail, I think we may want to stick with the current behaviour.

Let's say I have an API service that has a runtime dependency on a database. If I want to force-deploy the API, I may not want to force-redeploy the database at the same time.

To do that, I can always run `garden deploy api,db --force`, but I'd want force-redeployments to be explicit (either a specified list of services or all services, when no service names are provided).

Thoughts? Was there a specific context where the current behaviour was problematic, or did this behaviour come across as unexpected/strange?

vvagaytsev:
@thsig yes, thanks for the example. When no services names are provided, i.e. when we execute `garden deploy --force`, we assume all services to be re-deployed. This is what the command description says: `"Force redeploy of service(s)."`

But now it does work in that way. If we execute `garden deploy --force` in `demo-project`, then only the `frontend` service will be redeployed, not the `backend`. This looks like a bug.

I'm just wondering if the suggested fix is accurate enough to fix it. It looks like it is not, because the command `garden deploy frontend --force` with the current fix causes the redeployment of the `backend` service too.

We need a better approach for handling force mode. Should we store the service names to be redeployed in an array, as we do for dev mode and hot reload (and for local mode too)?

vvagaytsev:
@thsig should `garden deploy <serviceNames> --force-build` trigger both force build and force deploy? I suppose it should.

Orzelius:
> @thsig should `garden deploy <serviceNames> --force-build` trigger both force build and force deploy? I suppose it should.

I don't know about that. It would mean that `test --force` would force build, force deploy and then force test as well?

edvald:
> @thsig should `garden deploy <serviceNames> --force-build` trigger both force build and force deploy? I suppose it should.

Hmm that doesn't sound right to me. I can take a closer look. I should note that all this is changing substantially in 0.13 so please do not make any complex changes here :)

thsig:
> @thsig should `garden deploy <serviceNames> --force-build` trigger both force build and force deploy? I suppose it should.

I don't think it should, no. That's the behaviour of `deploy --force --force-build`.

It's a little unusual to force-build and not force-deploy, but it's a valid combination of the CLI options as they're currently laid out.

> I don't know about that. It would mean that test --force would force build, force deploy and then force test as well?

Similarly, `garden test --force` should not force-build (there's a `--force-build` CLI flag for the `test` command which serves that purpose).

PR #2953, opened 2022-05-16
improvement(core): more efficient module resolution

edvald:
This especially improves resolution for large graphs, by improving concurrency
during the resolution process.


edvald:
@vvagaytsev updated the PR, seems to all be in order now but please do give it a quick spin. See #2961 for those plugin test failures, which are unrelated.

core/src/util/testing.ts
```diff
@@ -46,6 +46,35 @@ export function getLogMessages(log: LogEntry, filter?: (log: LogEntry) => boolea
...
+  serviceConfigs: [],
+  spec: {},
+  taskConfigs: [],
+  testConfigs: [],
+  type: "test",
+}
+
+export function moduleConfigWithDefaults(partial: PartialModuleConfig) {
+  return {
+    ...moduleConfigDefaults,
```
vvagaytsev, line 69: This and `...moduleConfigDefaults.build` in L72 creates a shallow copy. So there is still a way to modify the underlying constant `moduleConfigDefaults` via the `ModuleConfig` instance produced by this method.

But, we do not do any of those modifications, and the method is used in `TestGarden` only. So, maybe deep cloning does not make sense here.

core/src/task-graph.ts
```diff
@@ -91,23 +93,30 @@ export class TaskGraph extends EventEmitter2 {
...
+    this.lock = new AsyncLock()
   }
 
   async process(tasks: BaseTask[], opts?: ProcessTasksOpts): Promise<GraphResults> {
+    this.log.silly(`TaskGraph: Processing ${tasks.length} tasks: ${tasks.map((t) => t.getKey()).join(",")}`)
+
     let nodes: TaskNode[]
     try {
-      nodes = await this.nodesWithDependencies({ tasks, dependencyCache: {}, stack: [] })
+      nodes = await this.nodesWithDependencies({ tasks, nodeMap: {}, stack: [] })
```
vvagaytsev, line 104: Do we need a `try/catch` block here? The error is re-thrown immediately, so L102-L107 can be simplified to
```
const nodes = await this.nodesWithDependencies({ tasks, nodeMap: {}, stack: []
```

edvald: I only renamed the argument here, not sure why that try/catch was in place prior. I'm assuming @thsig was debugging something at some point. Fine to remove.

core/src/util/testing.ts
edvald: I think it's sensible to do a deep clone here, even if it is only used for testing. Never know when this might get picked up elsewhere or create gnarly test bugs :P

vvagaytsev (changes requested):
Looks good, I left a couple of comments. I also asked @Orzelius to take a look.

Orzelius (approved):
couldn't find anything to nit-pic :)

vvagaytsev (approved):

PR #2960, opened 2022-05-20
fix(k8s): avoid unnecessary namespace patch no-op on init

edvald:
Just a minor thing I happened to come across, some no-op k8s API calls being made on provider init.


edvald:
See #2961 for those plugin test failures.

vvagaytsev (approved):

Issue #2924, opened 2022-04-30
Garden crashes if default dashboard port is taken


eysi09:
### Current Behavior

If I run Garden in persistent mode in a given project, it starts the dashboard server at port 9777. 

When I run Garden in persistent mode (for a different project, probably not relevant) in another process it crashes because the dashboard port is taken:

```
✔ garden-enterprise         → Authorizing... → Done
   ℹ Visit project at https://eysi.dev.enterprise.garden.io/projects/1

Error: listen EADDRINUSE: address already in use 127.0.0.1:9777
    at Server.setupListenHandle [as _listen2] (net.js:1331:16)
    at listenInCluster (net.js:1379:12)
    at GetAddrInfoReqWrap.doListen (net.js:1516:7)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:73:8)
```

### Expected behavior

It should pick another random port.

### Reproducible example

See above.

### Workaround

Manually set the dashboard port via env var when starting the second process. E.g.: `GARDEN_SERVER_PORT=9778 garden dev`. 

### Suggested solution(s)

I'm wondering if this broke with #2891. Maybe start by seeing whether there's something fishy in that commit.

vvagaytsev:
There is a breaking change in [v.6.0.0](https://github.com/sindresorhus/get-port/releases/tag/v6.0.0), the package became [pure ESM](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c).

PR #2969, opened 2022-05-24
fix: correct port number when open multiple dashboards

### LINKED ISSUE #2188: When opening mutliple dashboards, the url in the logs contains the wrong port
### Current Behavior

When running multiple dashboards, the info log showing the dashboard url contains the url of the first opened dashboard.
Additionally (it seems) gets updated with the latest created dashboard port. 

See below:

Screenshot 2021-01-07 at 17 13 56

### Reproducible example

Just run the command `garden dashboard` in two separate terminals.

### Your environment

* **OS:** 

`garden version`

vvagaytsev:
This PR ensures that the displayed URL of the currently running dashboard process is correct.

Previously the method [`GP.getDashboardProcess`](https://github.com/garden-io/garden/blob/master/core/src/db/entities/garden-process.ts#L95) did not rely on the base URL (i.e. the hostname and the port) of the target dashboard process. There might be several dashboards running on the different ports. In that case, we need to apply extra filtering by the command's base URL (which includes the port).

Fixes #2188

There are no integration tests for [`cli.ts`](https://github.com/garden-io/garden/blob/master/core/src/cli/cli.ts) yet. An integration test for this case assumes starting 2 garden CLI processes and monitoring the logs of each process. The corresponding unit tests were fixed respectively.

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -7,7 +7,7 @@
  */
 
 import dotenv = require("dotenv")
-import { intersection, sortBy } from "lodash"
+import { find, intersection, sortBy } from "lodash"
 import { resolve, join } from "path"
 import chalk from "chalk"
 import { pathExists } from "fs-extra"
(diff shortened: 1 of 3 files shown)


core/src/cli/cli.ts
```diff
@@ -436,18 +436,37 @@ ${renderCommands(commands)}
...
-            if (dashboardProcess) {
-              url = `${dashboardProcess.serverHost}?key=${dashboardProcess.serverAuthKey}`
+            let dashboardProcess: GardenProcess | undefined
+            if (dashboardProcesses.length === 0) {
+              dashboardProcess = undefined
+            } else if (dashboardProcesses.length === 1) {
+              dashboardProcess = dashboardProcesses[0]
+            } else {
+              // If there are multiple dashboard processes running,
+              // pick up the process associated with the current command's server url and port.
```
edvald, line 455: The port of a separate process will always be different from that of the current process (unless you are running the dashboard command here, which should be a case of its own). So I don't think this bit makes sense. 

What's happening in the underlying issue, I _think_, is that the 2nd dashboard process attempts to use port 9999 but it's not available, so it goes back to the default. So in a sense that's the expected behavior, but we could maybe do with a warning to say the desired port isn't available.

vvagaytsev: I checked the behavior with the debugger. I started a few different dashboards for the same project (I used `demo-project`):
```
garden dashboard
garden dashboard --port 9999
garden dashboard --port 9900
```

All 3 dashboards started successfully, but the wrong link was reported in the 2nd and 3rd cases, all 3 had the same link pointing to the first dashboard with the default port `9700`.

Next, I changed the execution order:
```
garden dashboard --port 9999
garden dashboard
garden dashboard --port 9900
```
and all three got the link with the port `9999`.

That happened because the method [GP.getDashboardProcess](https://github.com/garden-io/garden/blob/master/core/src/db/entities/garden-process.ts#L95) returned only **a single** dashboard matching the criteria defined in its `scope` argument. But, there **can be multiple** dashboards matching those criteria.

The `scope`-criteria do not distinguish different processes running from the same working dir on the different ports. In that case, we need a tie-breaker to pick up the right dashboard by its base URL which contains the port number.

vvagaytsev: There is another issue https://github.com/garden-io/garden/issues/2924 which is about port conflict. That issue comes from the bug in the `get-port` package (we used version `5.1.1`). It looks like it was fixed in the version `6.0.0` but there is a breaking change related to ESM modules. See the comments in https://github.com/garden-io/garden/issues/2924

vvagaytsev:
I had a discussion with @edvald and we decided to close this as "Won't fix". Running a few dashboards concurrently is an edge case, and the current behavior is expected.

Issue #2883, opened 2022-04-04
Garden workflows fail silently


twelvemo:
### Current Behavior

Running Garden workflows with errors in them like: 

* a wrong yaml indentation
* an invalid trigger spec in which an environment is specified that does not exist
* an invalid trigger spec with a trigger event that does not exist

causes Garden to fail without any error messages. Running this workflow with log-level 5 does also not output any hints into what is potentially mis-configured.

![Screenshot 2022-04-04 at 11 00 44](https://user-images.githubusercontent.com/78752267/161510595-9f8b57c8-7482-4893-a06f-31f10484c7ef.png)

The error log mentioned is not created.

### Expected behavior

Some output or additional help into what is causing Garden to fail. If the cause is a mis-configured environment that should be in the output. In case there is a problem with the yaml it would be good to hint at that.

### Reproducible example

Run the following workflow with wrong yaml indentation:

```
kind: Workflow
name: test
description: Testing out Workflows
steps:
  - script: echo "Hello"
      description: Hello World
      name: hello-world
triggers:
  - events: [pull-request]
    environment: dev # <--- Use one of your own environments here
```

Or the following with a non-existing trigger event:

```
kind: Workflow
name: test
description: Testing out Workflows
steps:
  - description: Hello World
    name: hello-world
    script: echo "hello world"
triggers:
  - events: [ miaow ]
    environment: dev # <--- Use one of your own environments here
```

```
garden run workflow test
```

### Suggested solution(s)

Add some output of why a workflow might be failing.

### Your environment

* **OS:** 
* **How I'm running Kubernetes:**  

`garden version`

Issue #2898, opened 2022-04-08
Workflow steps producing empty logs with structured output (-o=json / -o=yaml)


RixTmobilender:
### Current Behavior
Running a workflow with the output flag set to either JSON or YAML produces an empty log key:
![image](https://user-images.githubusercontent.com/87778732/162537918-9d00bd73-fe0d-4909-bfff-6f169a32b6f3.png)

Perhaps related to https://github.com/garden-io/garden/issues/2883
### Expected behavior
The log key should be populated with the jsonstring output of my workflow step, the same way tasks do

### Reproducible example
```
kind: Workflow
name: reproducible-example
steps:
- script: |
    echo "[\"my_json_string\"]"
```
`garden run workflow reproducible-example -o=json`
### Your environment

* **OS:** Ubuntu 22.04 on WSL2
**Garden version**: `0.12.39`

Orzelius:
@RixTmobilender thank you for the issue, I wouldn't have noticed the bug if you hadn't pointed it out. Is fixed via https://github.com/garden-io/garden/pull/2972

PR #2972, opened 2022-05-25
fix: workflows fail silently and don't output script logs with --output option

Orzelius:
Fixes the issue introduced in https://github.com/garden-io/garden/pull/2754 where workflows will fail silently if they fail outside of the steps (e.g. invalid workflow config) issue: #2883

Also fixes scripts logs not being visible with the `--output` option, issue: #2898

closes: #2883 and #2898

With this logic:
1. If `--output` is not specified and error(s) occur in the workflow steps themselves they will be printed while running the task and generic aggregated error will be printed at the bottom.
image

2. If `--output` is not specified and error occurs *outside* the workflow steps it will be printed immediately
image

The `--output` option also works now and the `.garden/error.log` gets properly populated as well.

### DIFF
--- a/core/src/commands/run/workflow.ts
+++ b/core/src/commands/run/workflow.ts
@@ -184,7 +184,7 @@ export class RunWorkflowCommand extends Command<Args, {}> {
       }
       stepBodyLog.root.storeEntries = initSaveLogState
 
-      if (stepResult.errors) {
+      if (stepResult.errors && stepResult.errors.length > 0) {
         garden.events.emit("workflowStepError", getStepEndEvent(index, stepStartedAt))
         logErrors(outerLog, stepResult.errors, index, steps.length, step.description)
         stepErrors[index] = stepResult.errors
--- a/core/src/commands/run/workflow.ts
+++ b/core/src/commands/run/workflow.ts
@@ -152,6 +152,8 @@ export class RunWorkflowCommand extends Command<Args, {}> {
 
       const stepStartedAt = new Date()
 
+      const initSaveLogState = stepBodyLog.root.storeEntries
+      stepBodyLog.root.storeEntries = true
       try {
         if (step.command) {
           step.command = resolveTemplateStrings(step.command, stepTemplateContext).filter((arg) => !!arg)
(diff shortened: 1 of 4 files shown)


core/src/commands/run/workflow.ts
```diff
@@ -152,6 +152,8 @@ export class RunWorkflowCommand extends Command<Args, {}> {
 
       const stepStartedAt = new Date()
 
+      const initSaveLogState = stepBodyLog.root.storeEntries
+      stepBodyLog.root.storeEntries = true
```
Orzelius, line 156: This is hacky as hell but current logic just gets the logs from the logger store so this is the easiest option to get the logs with the `--output` specified without a bigger rewrite. If anyone has better ideas I'm all ears!

vvagaytsev: Hacky but compact solution :) I think it's fine by now, no better immediate ideas from me. It might make sense to put a fixme- or todo-comment and make a note that some less hacky solution could be suggested.
@thsig any alternative ideas on how to do this?

vvagaytsev: Hacky but compact solution :) I think it's fine by now, no better immediate ideas from me. It might make sense to put a fixme- or todo-comment and make a note that some less hacky solution could be suggested.
@thsig any alternative ideas on how to do this?

vvagaytsev (dismissed):

Orzelius:
@vvagaytsev you can take a look at the two commits I added and if you're ok with them it's fine to squash and merge from my side as well.

core/test/unit/src/commands/run/workflow.ts
```diff
@@ -507,7 +507,7 @@ describe("RunWorkflowCommand", () => {
     const { result, errors } = await cmd.action({ ...defaultParams, args: { workflow: "workflow-a" } })
 
     expect(result).to.exist
-    expect(errors).to.eql([])
+    expect(errors).to.eql(undefined)
```
vvagaytsev, line 510: Looks like we usually use `.to.be.undefined` instead of `.to.eql(undefined)` (which is actually deepEqual operation) or `.to.equal(undefined)`.
All three approaches are functionally equal for `undefined`, so it's fine to keep it as it is now.

vvagaytsev (approved):
LGTM. There is one integration test failure, but it's irrelevant here. It happens sporadically in different branches.

PR #2982, opened 2022-06-02
feat(cli): --full option for get modules command

thsig:

Added a `--full` option to the `get modules` command.

When used, the fully resolved module config is logged (with some internal/irrelevant fields omitted). Any Garden Cloud secrets that would appear in the logged output are hidden/sanitized.

This is useful for debugging/writing template strings for modules, and for reasoning about which files/variables factor into the  module's version calculation.

This PR replaces #2652, and reuses much of the rendering logic sketched out there.

### DIFF
--- a/core/src/commands/get/get-modules.ts
+++ b/core/src/commands/get/get-modules.ts
@@ -6,15 +6,19 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+import { emoji as nodeEmoji } from "node-emoji"
 import { Command, CommandParams } from "../base"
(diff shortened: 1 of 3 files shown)


thsig:
Note: We easily could make the `--full` output option the default (and rename the current default to something like `--table`), but I thought I'd make it opt-in for starters.

eysi09:
> This is useful for debugging/writing template strings for modules, and for reasoning about which files/variables factor into the module's version calculation.

Nice 💯

vvagaytsev (approved):

PR #2983, opened 2022-06-02
chore: update bundled Docker to 20.10.9

thsig:

Updated the bundled Docker CLI from 19.03.6 to 20.10.9.

This is prompted by some users running into trouble with local Docker builds because we were shipping an older version of Docker.

If the need arises, we could also add a config field to the `local-kubernetes` and `kubernetes` providers that allows using whatever Docker CLI binary is on the PATH instead of the Garden-bundled version (like we do with the `terraform` and `pulumi` version fields).

That said, this may be unnecessary in practice unless actual backwards-incompatibilities are encountered by our users.

### DIFF
--- a/core/src/plugins/container/container.ts
+++ b/core/src/plugins/container/container.ts
@@ -302,8 +302,8 @@ export const gardenPlugin = () =>
           {
             platform: "darwin",
             architecture: "amd64",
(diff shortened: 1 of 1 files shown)


vvagaytsev (approved):

vvagaytsev (changes requested):
There are some linter errors, please fix.

vvagaytsev (approved):

PR #2984, opened 2022-06-02
test: updated some old helm charts to the newer versions

vvagaytsev:
Some tests fail with errors:
```
Error: chart "postgresql" version "8.10.5" not found in https://charts.bitnami.com/bitnami repository
Error: chart "redis" version "10.5.7" not found in https://charts.bitnami.com/bitnami repository
```

The charts older than 6 months are not available in the [standard bitnami charts repo](https://charts.bitnami.com/bitnami) since June 1st.  The index file was too heavy and some older versions were removed from the index. Too heavy index file caused some CloudFront traffic limitations and errors like this:
```
failed to generate YAML for specified Helm chart: failed to pull chart: looks like "https://charts.bitnami.com/bitnami" is not a valid chart repository or cannot be reached: stream error: stream ID 1; INTERNAL_ERROR; received from peer
```
We observed the error above in our test failures. See https://github.com/bitnami/charts/issues/10539 and https://github.com/bitnami/charts/issues/8433 for more details.

There is still a way to use the old chart versions by explicitly specifying the repo https://raw.githubusercontent.com/bitnami/charts/eb5f9a9513d987b519f0ecd732e7031241c50328/bitnami
But that's not a recommended way, because it can cause test failures again (because of the heavy index file and the traffic issue described above).

This PR updates the `postgres` and `redis` helm charts the latest versions compatible with Kubernetes 1.12+:
- [postgres 10.16.1](https://staging.artifacthub.io/packages/helm/bitnami/postgresql/10.16.1)
- [redis 16.0.0](https://staging.artifacthub.io/packages/helm/bitnami/redis/16.0.0)

Fixes #

### DIFF
--- a/core/test/data/test-projects/helm/postgres/garden.yml
+++ b/core/test/data/test-projects/helm/postgres/garden.yml
@@ -4,7 +4,7 @@ type: helm
 name: postgres
 chart: postgresql
 repo: https://charts.bitnami.com/bitnami
-version: "8.10.5"
+version: "10.16.1"
 serviceResource:
   kind: StatefulSet
   name: postgres
--- a/core/test/integ/src/plugins/kubernetes/util.ts
+++ b/core/test/integ/src/plugins/kubernetes/util.ts
@@ -367,7 +367,7 @@ describe("util", () => {
         log,
         version: module.version.versionString,
       })
-      module.spec.serviceResource.name = `{{ template "postgresql.master.fullname" . }}`
+      module.spec.serviceResource.name = `{{ template "postgresql.primary.fullname" . }}`
       const result = await getServiceResource({
         ctx,
         log,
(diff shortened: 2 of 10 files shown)


twelvemo:
Just saw that this also makes the vote-helm example fail. It uses helm charts from bitnami for postgres and redis, which both don't exist any longer.

vvagaytsev:
Now there are other test failures like
```
Error: template: postgresql/templates/d03e5fadae183434.yaml:1:20: executing "postgresql/templates/d03e5fadae183434.yaml" at <{{template "postgresql.master.fullname" .}}>: template "postgresql.master.fullname" not defined
```

Postgres chart 10.0.0 had some [breaking changes](https://artifacthub.io/packages/helm/bitnami/postgresql/10.16.1#to-10-0-0), which should be addressed in this PR.

PR #2995, opened 2022-06-06
Core-cloud Interactivity

Orzelius:
Listen to events from cloud and create tasks based on them. Also dashboard command can also be used for it.

Will leave as draft until this has been properly tested and combined with the Cloud implementation and the command palette.

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -293,15 +293,15 @@ ${renderCommands(commands)}
     const footerLog = logger.placeholder()
 
     command.printHeader({ headerLog, args: parsedArgs, opts: parsedOpts })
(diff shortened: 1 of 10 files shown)


vvagaytsev:
Looks like this requires some extra work. Currently, `devMode` and `hotReload` are not supported. The corresponding arrays `devModeServiceNames` and `hotReloadServiceNames` are hard-coded to be `[]`.

Let's merge the local mode PR https://github.com/garden-io/garden/pull/2949 first, and then implement the full support of `devMode`, `hotReload`, and `localMode` for core-cloud interactivity.

I'll rebase https://github.com/garden-io/garden/pull/2949 and make the necessary initial changes in the interactivity part.

PR #2996, opened 2022-06-06
docs: add backend injection to terraform doc

thelongshanx:
**What this PR does / why we need it**: This PR explains how to use an exec module to inject environment variables into backend terraform files

### DIFF
--- a/docs/advanced/terraform.md
+++ b/docs/advanced/terraform.md
@@ -107,6 +107,48 @@ Here we imagine a Terraform stack that has a `my-database-uri` output, that we t
 
 Much like other modules, you can also reference Terraform definitions in other repositories using the `repositoryUrl` key. See the [Remote Sources](./using-remote-sources.md) guide for details.
 
+## Injecting Environment Variables Into Backend Manifests
+
+[Terraform does not interpolate named values in backend manifests](https://www.terraform.io/language/settings/backends/configuration). Below are two solutions (using an `exec` provider and using both an `exec` and a `teraform` module).
+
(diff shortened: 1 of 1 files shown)


worldofgeese (approved):
Ah, the [all-too-familiar](https://www.monterail.com/blog/chicken-or-egg-terraforms-remote-backend) [chicken-and-egg](https://charity.wtf/2016/03/30/terraform-vpc-and-why-you-want-a-tfstate-file-per-env/) quandary of creating a state bucket to hold managed state of your uncreated state bucket strikes again!

I would consider pivoting the initScript example to making a strong value argument for what garden's TF support gives you versus the Terraform's tfvars file.

I would then write out a disclaimer that if you're an AWS user using Garden to manage S3 buckets for state you can profit from your example.

I'd love to see this edge-case -- not at all relevant to your PR -- just solved inside garden so users never have to deal with the chicken-and-egg problem of AWS state buckets again.

Terragrunt deals with it its own way as does [pretf](https://pretf.readthedocs.io/en/latest/tutorial/aws-projects/), a Python wrapper, and [Terraformsh](https://github.com/pwillis-els/terraformsh/blob/3e82ff5c0a170a17cc0a8df2548421898db96066/terraformsh#L285) a Terraform bash wrapper.

I have no idea if any of our ace combat OSS coders want to take a look at that but I will @Orzelius and @vvagaytsev just in case.

@thelongshanx take all this with a big grain of salt and consider just telling me to take wing elsewhere! I otherwise approve 😻

PR #2999, opened 2022-06-08
fix a few ci problems

Orzelius:
FIx a bunch of small things that were messing up ci + add new k8s versions and remove old unsupported ones.

Please review this commit by commit (they have a lot of extra info in them)

Stuff that still sometimes fails (don't have any more time to spend on debugging, maybe will do so later)
* build step will still occasionally fail due to the concurrency issue in pkg. The delays seemed to help, but not completely fix (more info in the pkg commits)
* in vm e2e runs sometimes mutagen fails to sync stuff and crashes ([example](https://app.circleci.com/pipelines/github/garden-io/garden/12104/workflows/7e760b32-0793-46e7-8182-91b19118599c/jobs/196256))

### DIFF
--- a/cli/src/build-pkg.ts
+++ b/cli/src/build-pkg.ts
@@ -11,7 +11,7 @@ import { resolve, relative, join } from "path"
 import Bluebird from "bluebird"
 import { STATIC_DIR, GARDEN_CLI_ROOT, GARDEN_CORE_ROOT } from "@garden-io/core/build/src/constants"
 import { remove, mkdirp, copy, writeFile } from "fs-extra"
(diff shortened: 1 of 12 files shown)


cli/src/build-pkg.ts
```diff
@@ -240,6 +242,7 @@ async function pkgCommon({
 
   console.log(` - ${targetName} -> pkg`)
   await exec(pkgPath, [
+    "-d", // it seems to fix a concurrency big in pkg if this is added :(
```
vvagaytsev, line 245: "big" -> "bug"

PR #3003, opened 2022-06-13
feat: local mode k8s module

vvagaytsev:
Local mode for `kubernetes` and `helm` modules.

Fixes #


vvagaytsev:
@wojciechka the local mode for `kubernetes` modules has been implemented in this PR. We're planning to merge it next week. Feel free to test the latest build of the branch. Any feedback and comments are highly appreciated!

examples/local-mode-k8s/backend-local/main.go
```diff
@@ -0,0 +1,18 @@
...
+func handler(w http.ResponseWriter, r *http.Request) {
+	fmt.Println("Incoming request")
+	fmt.Fprint(w, "Hello from Local Go!")
+}
+
+func main() {
+	http.HandleFunc("/hello-backend", handler)
+	fmt.Println("Server running at port 8090...")
+
+	http.ListenAndServe(":8090", nil)
```
Orzelius, line 17: could you also catch the error here like in the backend-image/main.go

It's a pain to debug if it crashes because it will give zero output and just die. Happens often when the port is taken

examples/local-mode-k8s/backend/garden.yml
```diff
@@ -0,0 +1,80 @@
+kind: Module
+name: backend
+description: Backend service container
+type: kubernetes
+
+localMode:
+  localPort: 8090
+  command: [ "../backend-local/main" ]
```
Orzelius, line 8: could you explain this command?

docs/guides/running-service-in-local-mode.md
```diff
@@ -34,9 +34,13 @@ progress. So, there is a number of functional limitations in the first release:
   first `TCP` port from the list of ports or just the first one if no `TCP` ports defined. Thus, if the service needs to
   talk to some data sources like databases, message brokers, etc. then all these services are assumed to be running
   locally.
-* The _local mode_ is supported only by [container module type](./container-modules.md)
-  and [kubernetes provider](../reference/providers/kubernetes.md). The `kubernetes` and `helm` module types will be
-  supported later.
+* The _local mode_ is supported only by [`container`](./container-modules.md)
+  and [`kubernetes`](../reference/module-types/kubernetes.md) module types.
+  The [`helm`](../reference/module-types/helm.md) module type will be supported later.
```
thsig, line 39: "The helm module type will be supported later" -> "Support for the helm module type will be added soon."

docs/guides/running-service-in-local-mode.md
```diff
@@ -34,9 +34,13 @@ progress. So, there is a number of functional limitations in the first release:
...
+* Only one service can be executed in _local mode_ with [`kubernetes`](../reference/module-types/kubernetes.md)
```
thsig, line 40: "Only one service can be executed in local mode with kubernetes and helm module types" -> "Only one container can be run in local mode for each `kubernetes` service (the same will be the case for `helm` services when local mode is implemented there)."

docs/guides/running-service-in-local-mode.md
```diff
@@ -34,9 +34,13 @@ progress. So, there is a number of functional limitations in the first release:
   first `TCP` port from the list of ports or just the first one if no `TCP` ports defined. Thus, if the service needs to
   talk to some data sources like databases, message brokers, etc. then all these services are assumed to be running
   locally.
-* The _local mode_ is supported only by [container module type](./container-modules.md)
-  and [kubernetes provider](../reference/providers/kubernetes.md). The `kubernetes` and `helm` module types will be
-  supported later.
+* The _local mode_ is supported only by [`container`](./container-modules.md)
```
thsig, line 37: Replace all references to "the local mode" with "local mode" in this doc.

docs/guides/running-service-in-local-mode.md
```diff
@@ -116,7 +120,28 @@ services:
...
 
-An example can be found in the [demo-project's backend service](../../examples/demo-project/backend/garden.yml).
+An example can be found in the [`local-mode project`](../../examples/local-mode).
+
+### Configuring dev mode for `kubernetes` and `helm` modules
+
+```yaml
+kind: Module
+name: backend
+type: kubernetes # this example looks the same for helm modules (i.e. with `type: helm`)
```
thsig, line 130: We should probably delete this comment until we actually add support for local mode for `helm` services (and also remove the reference to `helm` modules from the section heading on line 125 above).

thsig (changes requested):
Great work! I've only got text-related comments on this one.

examples/local-mode-k8s/backend/garden.yml
vvagaytsev: It's explained in the documentation. But we can put a short comment here as well.

core/src/plugins/kubernetes/helm/config.ts
```diff
@@ -58,6 +61,7 @@ export interface HelmServiceSpec {
   chartPath: string
   dependencies: string[]
   devMode?: KubernetesDevModeSpec
+  localMode?: KubernetesLocalModeSpec
```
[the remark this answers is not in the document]
vvagaytsev: This will cause build failure in https://github.com/garden-io/garden/blob/3163961c370d89aec4efbc486ee475d8d6200c9f/core/src/plugins/kubernetes/kubernetes-module/handlers.ts#L385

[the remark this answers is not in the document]
thsig: Aha, gotcha—let's just leave it as is, then.

I'm sure we'll be adding Helm support soon anyway.

thsig:
Looks good to me—unless @Orzelius has any further comments, I think we're good to merge.

Orzelius (approved):
:rocket:

thsig (approved):

wojciechka:
> @wojciechka the local mode for `kubernetes` modules has been implemented in this PR. We're planning to merge it next week. Feel free to test the latest build of the branch. Any feedback and comments are highly appreciated!

Hi! I've tested the solution and it works as expected in my initial tests.

I'll work on trying to integrate it in our tooling and see if I can get it working end to end in our tooling. I couldn't easily do that with `container` module but should be doable now.

One question, though - is there a way to have local command write to stdout or where would the logs from the command be?

Seeing stdout/stderr in real time would be useful for our engineers.

PR #3005, opened 2022-06-14
fix: update nginx ingress controller for kind

Orzelius:
based on https://github.com/garden-io/garden/pull/2902, but is backwards compatible

this is currently blocking #2999

(ignore the broken ci, I have  #2999 for that)


twelvemo (approved):
Thanks so much for taking care of that!

PR #3007, opened 2022-06-14
fix(k8s): change kaniko default image to 1.8.1

TimBeyer:
Kaniko has a bug with not computing the correct cache keys when running multistage builds yielding images with outdated files inside. This version fixes the issue.


thsig:
Good catch!

My only comment is that we should probably use `fix(k8s)` instead of `chore` in the commit message, so that it shows up in the changelog.

thsig (approved):

PR #3011, opened 2022-06-16
[0.13] refactor: remove deprecated hot reload feature

edvald:
BREAKING CHANGE:

The legacy hot-reloading feature has been removed. Users should now use
_dev mode_, which is far more performant.


core/src/config/template-contexts/project.ts
```diff
@@ -182,7 +182,7 @@ class VcsContext extends ConfigContext {
   }
 }
 
-const commandHotExample = "${command.params contains 'hot-reload' && command.params.hot-reload contains 'my-service'}"
+const commandHotExample = "${command.params contains 'dev-mode' && command.params.dev-mode contains 'my-service'}"
```
Orzelius, line 185: maybe also rename this variable

Orzelius (approved):

core/src/config/template-contexts/project.ts
edvald: 👍  updated, and fixed the breaking tests.

Issue #2905, opened 2022-04-20
Incompatible with git 2.36.0


tjcutajar:
## Bug

upgrading to git 2.36.0 causes garden to error.

### Current Behavior

commands raise 
`Path /Users/myUser/.garden/bin/static/kubernetes/system is not in a git repository root. Garden must be run from within a git repo. Please run 'git init' if you're starting a new project and
repository, or move the project to an existing repository, and try again.`

### Your environment

* **OS:** MacOS 12.3.1

`garden version` 0.12.39

Orzelius:
what command are you running and are you running it within a git repository?

tjcutajar:
running either `deploy` or `delete` from within a git repo. Downgrading git to 2.35.3 resolves it.

tjcutajar:
@twelvemo maybe try deleting your .garden directory

alexkuretz:
We are getting this on Windows as well while trying to set up a new system. The user getting the error is using git 2.36.0 and my setup with git 2.35.1 does  not show the error. We are running within our git repository.

> Failed resolving provider kubernetes. Here is the output:
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> Path C:\Users\username\\.garden\bin\static\kubernetes\system is not in a git repository root. Garden must be run from within a git repo. Please run `git init` if you're starting a new project and repository, or move the project to an existing repository, and try again.
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> Failed resolving one or more providers:
> - kubernetes

alexkuretz:
Another coworker encountered this today and found that by going into the .garden\bin\static\kubernetes\system directory and doing a 'git status' he was able to set a permission that allowed garden to work. I believe it's related to the "Stricter repository ownership checks" change in git 2.36:
https://github.blog/2022-04-18-highlights-from-git-2-36/

edvald:
Thanks for the investigation @alexkuretz! I figure it's a simple fix, we'll get on it.

ITHedgeHog:
I got around this by running the command git config --add safe.directory C:\git\oemplatform which is the project root for me, this got me running.

vvagaytsev:
[...]

PR #3012, opened 2022-06-17
fix: git repository ownership issue

vvagaytsev:
This is an additional fix for #2905. The same issue happened with some sub-folder of the git repo in the Garden Cloud testing.

The suggested solution is to always consider any working directory in a git repository as a safe one.

Fixes #

### DIFF
--- a/core/src/vcs/git.ts
+++ b/core/src/vcs/git.ts
@@ -139,9 +139,7 @@ export class GitHandler extends VcsHandler {
       if (err.exitCode === 128 && err.stderr?.toLowerCase().includes("fatal: unsafe repository")) {
         log.warn(
           chalk.yellow(
-            "It looks like you're using Git 2.36.0 or newer " +
-              `and the Garden static directory "${path}" is owned by someone else. ` +
-              "It will be added to safe.directory list in the .gitconfig."
+            `It looks like you're using Git 2.36.0 or newer and the Garden static directory "${path}" is owned by someone else. It will be added to safe.directory list in the .gitconfig.`
           )
         )
         const gitConfigCompatiblePath = this.toGitConfigCompatiblePath(path, currentPlatformName)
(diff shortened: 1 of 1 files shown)


Orzelius (approved):

thsig (changes requested):
Good stuff!

While the logic itself looks fine, we're now performing two extra git CLI calls every time we call `vcs.getFiles` (which can happen once for each module when a command is first run).

How about reading the contents of the `safe.directory` block in the git config the first time this method is called, and initialising `this.gitSafeDirs` using the entries from there?

That should mean that after the path is added (e.g. the first time `ensureSafeDirGitRepo` is called on the path) we can almost always return semi-instantly from `ensureSafeGitDirRepo` without having to call git.

I'm mostly mentioning this because we recently became aware that the time/resource cost of all these git CLI calls really adds up for large projects with lots of modules.

eysi09 (approved):

core/src/vcs/git.ts
```diff
@@ -168,6 +197,8 @@ export class GitHandler extends VcsHandler {
     }
 
     await this.ensureSafeDirGitRepo(log, STATIC_DIR)
+    await this.ensureSafeDirGitRepo(log, path)
```
edvald, line 200: Sorry I'm late to the party here, but I should note that this will add up to hundreds of git calls for certain large projects. I wonder if this could be done more optimally, but maybe that could be included along with other optimizations.

vvagaytsev: Yes, we can do further optimizations later. Now we read all safe directories from the `.gitconfig` file, this will allow us to avoid extra `git status` calls for the directories which are explicitly marked as safe ones in the `.gitconfig`.

But `git status` will be executed for "normal" repos, i.e. the repos that do not have any ownership issues and should not be marked as safe ones.

We can simply cache the successful git status check results in memory, I did it in https://github.com/garden-io/garden/pull/3014

Of course, if there are any repo ownership changes done while a Garden CLI is running, then we might get the issue again. But restarting the Garden CLI will solve the issue.

vvagaytsev: One more idea on optimizing this. We can keep the git safe directories in a singleton object. In such case, all `GitHandler` instances will share the same in-memory cache with known safe directories.

PR #3014, opened 2022-06-21
improvement: optimized safety checks for git repos

vvagaytsev:
This PR minimizes number of `git status` calls in the `GitHandler.ensureSafeDirGitRepo(log: LogEntry, path: string)` method. Every successful `git status` result is cached in memory.

Fixes #

See https://github.com/garden-io/garden/pull/3012#discussion_r902252328 for more details.


edvald (approved):
Approving, but let's also take a deeper look soon :)

PR #3021, opened 2022-06-27
Interactivity improvements

Orzelius:
this addresses a few issues introduced in #2995

* improve logging (see pic below)
* clear cache for each cloud action (see commit for more detailed explanation)
* a few minor things

now looks like this:
image


vvagaytsev (approved):
Looks good! It might make sense to have something like "improvement: minor improvements in core/cloud interactivity" as a squashed commit message, and you can retain the original details as an additional commit description (which comes after a blank line).

PR #3024, opened 2022-06-30
chore(tests): used python 3 in test projects

vvagaytsev:
This PR replaces the ancient Python 2.7 with the latest stable 3.10 version in the test projects. It also makes the code-style more Pythonic.

Fixes #


Orzelius (approved):
Nice! I will just trust you that these still work as I didn't test myself

PR #3033, opened 2022-07-05
feat(k8s): local mode for helm modules

vvagaytsev:
Local mode for `helm` modules. This is based on the changes made in #2949 and #3003.

Fixes #

A commit-by-commit review should be an easy way to go.


docs/guides/running-service-in-local-mode.md
```diff
@@ -34,12 +34,11 @@ progress. So, there is a number of functional limitations in the first release:
...
   locally.
-* The _local mode_ is supported only by [`container`](./container-modules.md)
-  and [`kubernetes`](../reference/module-types/kubernetes.md) module types.
+* The _local mode_ is supported by [`container`](./container-modules.md),
+  [`kubernetes`](../reference/module-types/kubernetes.md), and [`helm`](../reference/module-types/helm.md) module types.
   Support for the [`helm`](../reference/module-types/helm.md) module type will be added soon.
-* Only one container can be run in local mode for each [`kubernetes`](../reference/module-types/kubernetes.md) service (
-  the same will be the case for [`helm`](../reference/module-types/helm.md) services when local mode is implemented
-  there).
+* Only one container can be run in local mode for each [`kubernetes`](../reference/module-types/kubernetes.md) or
```
Orzelius, line 40:
```suggestion
* Only one container can be ran in local mode for each [`kubernetes`](../reference/module-types/kubernetes.md) or
```
or
```suggestion
* Only one container can run in local mode for each [`kubernetes`](../reference/module-types/kubernetes.md) or
```

vvagaytsev: Why do we need "can be ran" here? It's a passive voice construction. So we need to use the past participle form, "run".

Orzelius: haha I don't know, it just sounded right in my head, but I think you're right so ignore this one

Orzelius:
Ah I didn't notice it was still WIP, sorry for jumping to the review

Orzelius (approved):
This looks good, but am I correct in that there's currently no integ/e2e test that actually makes a request to the fake proxy service and expects it to be handled locally? I think such test would be also good to have for all three module types to catch the scenario where all the right things are deployed but the proxy itself is not working for whatever reason. This could be a separate pr.

PR #3034, opened 2022-07-05
feat(core): opt for deleting services in dep order

### LINKED ISSUE #2782: [BUG]: Garden is allowing delete a dependency of a deployed service
### Current Behavior

When running `garden delete my-service-dependency` it is not taking in consideration that my-service is deployed and running and it is allowing to delete its dependency. 

Also when deleting all services by running `garden delete service` it is not deleting the services in order, which is important as some resources can't be deleted unless their child resources have been deleted.

### Expected behavior

I expect that garden throws an error when trying to delete a dependency of a service that was not set to delete together.
I expect that garden delete the services respecting the order of dependencies. 

### Suggested solution(s)

Unless I opt to delete all services or the services are being deleted along with its dependencies specific, this shouldn't be allowed and an error should be thrown. 
An optional flag could also be added to force delete the dependency in case we want that.
When deleting all services it could use the dependency graph to delete all the services that depend on a service before the dependent service is deleted.

### Your environment

* **OS:** linux distro

`garden version` 0.12.34-edge-7fdd7022

thsig:

Added a `--dependants-first` option to the `delete service` and `delete environment` commands.

When used, services will be deleted in reverse dependency order. That is, service dependants will be    deleted before their dependencies.

Also added a `--with-dependants` option to the `delete service` command, which deletes any running      dependants of the requested service/services.

Fixes #2782.

Note that the dependency logic here only uses service dependencies and doesn't traverse task dependencies to find transitive service dependencies.

This means that e.g. in the vote project, the `result` service won't be deleted before the `postgres` service, even though `result` transitively depends on `postgres` via the `db-init` task. Adding an explicit service dependency on `postgres` to the `result` service would change this.

If do we want to fully replicate the dependency behavior of `DeployTask` in reverse (which I think is probably unnecessary), we'd have to modify the dependency calculation logic there (which would be a bit fiddly, but not a big deal).

### DIFF
--- a/core/src/actions.ts
+++ b/core/src/actions.ts
@@ -850,6 +850,7 @@ export class ActionRouter implements TypeGuard {
     const servicesLog = log.info({ msg: chalk.white("Deleting services..."), status: "active" })
 
     const services = graph.getServices({ names })
+    const deleteServiceNames = services.map((s) => s.name)
 
     const deleteResults = await this.garden.processTasks(
       services.map((service) => {
@@ -858,7 +859,7 @@ export class ActionRouter implements TypeGuard {
           graph,
           service,
           log: servicesLog,
-          includeDependants: true,
+          deleteServiceNames,
         })
       })
     )
--- a/core/src/commands/delete.ts
+++ b/core/src/commands/delete.ts
@@ -9,14 +9,20 @@
 import { Command, CommandResult, CommandParams, CommandGroup } from "./base"
 import { NotFoundError } from "../exceptions"
 import dedent from "dedent"
-import { ServiceStatusMap, serviceStatusSchema } from "../types/service"
+import { ServiceStatus, ServiceStatusMap, serviceStatusSchema } from "../types/service"
 import { printHeader } from "../logger/util"
 import { DeleteSecretResult } from "../types/plugin/provider/deleteSecret"
 import { EnvironmentStatusMap } from "../types/plugin/provider/getEnvironmentStatus"
 import { DeleteServiceTask, deletedServiceStatuses } from "../tasks/delete-service"
 import { joi, joiIdentifierMap } from "../config/common"
 import { environmentStatusSchema } from "../config/status"
-import { StringParameter, StringsParameter } from "../cli/params"
+import { BooleanParameter, StringParameter, StringsParameter } from "../cli/params"
+import { deline } from "../util/string"
+import { Garden } from ".."
+import { ConfigGraph } from "../config-graph"
+import { LogEntry } from "../logger/log-entry"
+import { uniqByName } from "../util/util"
+import { flatten } from "lodash"
 
 export class DeleteCommand extends CommandGroup {
   name = "delete"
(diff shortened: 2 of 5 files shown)


core/src/tasks/delete-service.ts
```diff
@@ -19,21 +19,24 @@ export interface DeleteServiceTaskParams {
...
 export class DeleteServiceTask extends BaseTask {
   type: TaskType = "delete-service"
   concurrencyLimit = 10
   graph: ConfigGraph
   service: GardenService
-  includeDependants: boolean
+  deleteServiceNames: string[]
 
-  constructor({ garden, graph, log, service, includeDependants = false }: DeleteServiceTaskParams) {
+  constructor({ garden, graph, log, service, deleteServiceNames }: DeleteServiceTaskParams) {
```
Orzelius, line 35: why even have the `service` parameter here any more if all of them will be deleted anyway?

core/src/commands/delete.ts
```diff
@@ -167,8 +167,10 @@ export class DeleteServiceCommand extends Command {
       return { result: {} }
     }
 
+    const deleteServiceNames = services.map((s) => s.name)
+
     const deleteServiceTasks = services.map((service) => {
-      return new DeleteServiceTask({ garden, graph, log, service })
+      return new DeleteServiceTask({ garden, graph, log, service, deleteServiceNames })
```
Orzelius, line 173: This can also be taken out of the map I think, (if not I need an explanation on what's going on :small_airplane: )

thsig:
@edvald: The changes we talked about yesterday have been applied here, so I think we should be good to go.

vvagaytsev (approved):

vvagaytsev (changes requested):
Looks like there are some test failures: https://app.circleci.com/pipelines/github/garden-io/garden/12296/workflows/eac4a57b-75e9-40a5-acbd-4eeb83c13b56/jobs/201243

core/src/actions.ts
```diff
@@ -858,7 +859,7 @@ export class ActionRouter implements TypeGuard {
           graph,
           service,
           log: servicesLog,
-          includeDependants: true,
+          deleteServiceNames,
```
vvagaytsev, line 862: Should the `dependantsFirst` flag be propagated to the action router too?

core/test/unit/src/commands/delete.ts
```diff
@@ -143,6 +150,39 @@ describe("DeleteEnvironmentCommand", () => {
       "service-d": { forwardablePorts: [], state: "missing", detail: {}, outputs: {} },
     })
     expect(deletedServices.sort()).to.eql(["service-a", "service-b", "service-c", "service-d"])
+
+    // This means that the services weren't deleted in dependant order (i.e. they were all deleted at the same time).
+    expect(deleteOrder).to.eql(["service-a", "service-b", "service-c", "service-d"])
```
vvagaytsev, line 155: I think this condition is not stable enough. If the services are deleted simultaneously, the final deletion order cannot be predicted. We can use something like this here:
```
const expectedDeletedServices = shuffle(["service-a", "service-b", "service-c", "service-d"])
expect(new Set<string>(deleteOrder)).to.eql(new Set<string>(expectedDeletedServices))
``` 

But that would look redundant as it's equivalent to L152:
```
expect(deletedServices.sort()).to.eql(["service-a", "service-b", "service-c", "service-d"])
```

core/test/unit/src/commands/delete.ts
```diff
@@ -143,6 +150,39 @@ describe("DeleteEnvironmentCommand", () => {
...
+      expect(result!.serviceStatuses).to.eql({
+        "service-a": { forwardablePorts: [], state: "missing", detail: {}, outputs: {} },
+        "service-b": { forwardablePorts: [], state: "missing", detail: {}, outputs: {} },
+        "service-c": { forwardablePorts: [], state: "missing", detail: {}, outputs: {} },
+        "service-d": { forwardablePorts: [], state: "missing", detail: {}, outputs: {} },
+      })
+      expect(deletedServices.sort()).to.eql(["service-a", "service-b", "service-c", "service-d"])
+
+      // This means that the services were deleted in dependant order.
+      expect(deleteOrder).to.eql(["service-c", "service-d", "service-b", "service-a"])
```
vvagaytsev, line 184: Sometimes this fails with actual value `["service-b", "service-a", "service-c", "service-d"]`.
The dependency graph here looks like this (I omitted `service-` prefix):
```
c -> |
        | -> b -> a
d -> |
```

So, the deletion order would look like this: `((c, d), (b), (a))`. We mocked `deleteService` handler in the test to just write the service name to the list. but `deleteService` handler is called asynchronously, so the completion order might still be unstable.

thsig: I've fixed this now—I had originally intended to use the handwritten module configs for the `DeleteEnvironment` tests too, but was still using the modules from `projectRootB`.

To be extra sure, I ran these on a loop for ~30 repetitions, and never got a test failure. So I think we should be good to go now.

vvagaytsev (approved):

PR #3040, opened 2022-07-07
Improve dead links check

Orzelius:
please see individual commits for more info


Orzelius:
Aslo, this won't actually fix the currently failing check-docs ci stage as it's failing due to gitbook not working

scripts/check-docs.sh
```diff
@@ -9,20 +9,15 @@ git diff --quiet HEAD -- docs/ || (echo 'generated docs are not up-to-date! run
...
-}
-
-error_in_external_links=false
-check_external_links() {
+found_dead_links=false
+check_links() {
   for file in $@; do
     # markdown-link-check is configured to ignore relative links
-    ./node_modules/.bin/markdown-link-check --config markdown-link-check-config.json $file || error_in_external_links=true
+    ./node_modules/.bin/markdown-link-check --config markdown-link-check-config.json $file || found_dead_links=true
```
vvagaytsev, line 16: The issue https://github.com/tcort/markdown-link-check/issues/91 was resolved in `3.10.0` according to [this comment](https://github.com/tcort/markdown-link-check/issues/91#issuecomment-1073069387).

Should we upgrade `markdown-link-check` from `^3.9.3` to `^3.10.0`?

vvagaytsev: Ah, there was some [regression](https://github.com/tcort/markdown-link-check/issues/91#issuecomment-1102492323) in `3.10.0`, so we might be ok with `3.9.3`, or should consider using the latest available stable version.

Orzelius: I did manual testing and it did work for relative, anchor and normal links so I think we can run with this until we encounter the next issue

vvagaytsev (approved):

PR #3046, opened 2022-07-08
fix(cloud): handle "=" separator in named vars passed as command args

### LINKED ISSUE #2405: garden enterprise create secrets command cannot create secrets that include `=`
### Current Behavior

The `garden enterprise secrets create`  command cannot create any secrets that include an `=` , which is often the case in certificates or base64 encoded values. It cuts everything off after an `=` , using single quotes or quotes does not prevent that.

### Expected behavior

Secrets including `=` should be created with the full content including `=` and everything after it.
### Reproducible example

```
$ garden enterprise secrets create TEST="123=4"
```
With an initScript that prints the created secret:
```
...
providers:
  - name: exec
    initScript: echo ${secrets.TEST}
```
The output looks the following;

```
ℹ providers                 → Preparing environment...
   ℹ exec                      → Configuring...
   exec                      → Running init script
   123
```

Using single quotes around the secret creates the same output.

### Workaround

Using the `--from-file` flag and quoting the secret value results in the creation of the full secret including the `=` and everything after it. But this workaround only functions, if the whole secret value is in the same line. For multi-line secrets e.g. certificates it leads to this behaviour:

secret:
```
FOO="12
34=
5"
```

output from garden cli:

```
$ garden enterprise secrets create --from-file secrets
Create secrets 🔒

✔ garden-enterprise         → Authorizing... → Done
✔ garden-enterprise         → Initializing... → Ready
✖ secrets-command           → Creating secrets... → 2/2 → Error

Failed creating 1/2 secrets. See errors below:

→ Creating secret "34 "failed with error: Secret value is missing.

Successfully created 1 secret!

Command failed.
```
resulting secret `FOO` is "12".

### Suggested solution(s)

Escape `=` in quoted secret values on the command line.
### Your environment

* **OS:** MacOS

`garden version`
```
0.12.21
```

vvagaytsev:
This PR ensures proper handling of the `=` separators when named variables are passed as command-line arguments. The fix addresses `garden enterprise secrets create` and `garden enterprise users create` commands.

Fixes #2405

We do not have any tests for `commands.cloud` package yet, so no tests have been added.

### DIFF
--- a/core/src/commands/cloud/secrets/secrets-create.ts
+++ b/core/src/commands/cloud/secrets/secrets-create.ts
@@ -8,15 +8,15 @@
 
 import { CommandError, ConfigurationError, EnterpriseApiError } from "../../../exceptions"
 import { CreateSecretResponse } from "@garden-io/platform-api-types"
-import dotenv = require("dotenv")
 import { readFile } from "fs-extra"
 
 import { printHeader } from "../../../logger/util"
 import { Command, CommandParams, CommandResult } from "../../base"
 import { ApiCommandError, handleBulkOperationResult, makeSecretFromResponse, noApiMsg, SecretResult } from "../helpers"
 import { dedent, deline } from "../../../util/string"
-import { StringsParameter, PathParameter, IntegerParameter, StringParameter } from "../../../cli/params"
+import { IntegerParameter, PathParameter, StringParameter, StringsParameter } from "../../../cli/params"
 import { StringMap } from "../../../config/common"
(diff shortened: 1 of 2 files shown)


Orzelius (approved):
cool, didn't know dotenv exposed the parsing logic, but it's a perfect use case here. Maybe a separate issue should be created about adding the test cases?

vvagaytsev:
Thanks, @Orzelius! I created an internal task to add some tests.

PR #3049, opened 2022-07-09
refactor(k8s): remove in-cluster registry and cluster-docker build mode

edvald:
BREAKING CHANGE:

We no longer support the deprecated `cluster-docker` build mode. Please use
`cluster-buildkit` or `kaniko` for in-cluster building.

We no longer support deploying an in-cluster registry. Users must now configure
a `deploymentRegistry` in the `kubernetes` provider configuration when using
remote clusters. Please refer to the
[In-cluster building guide](https://docs.garden.io/guides/in-cluster-building)
for more information.


Orzelius (commented):
This is hard to review thoroughly, but from a quick look looks good. Currently the e2e tests are failing due to the remote kubernetes provider config now requiring a registry. Should we just provide our own or is there a different way you'd like to go about this?

edvald:
Good catch. We should indeed provide our own. And in fact, we should update all relevant examples such that the `kubernetes` provider has a registry configured. Want to have a swing at that @Orzelius, or leave it with me?

edvald:
Actually, I'll sort it out, can do it quickly

edvald:
@Orzelius added a commit. Please take another look when you can. The failing tests here seem unrelated btw.

Orzelius (approved):
gltm!

PR #3054, opened 2022-07-15
improvement: used async sleeps between retries in recoverable process

vvagaytsev:
I've noticed that sleep periods between retries do not work well when I use Garden binaries built by CI pipeline. Local builds work fine. I'm not sure if it's related to the sync sleeping implementation, but it makes sense to use async sleep like anywhere else in the project codebase.

Async sleep won't spawn a new OS process, it will use `Promise` instead.
Also, sleep is called from the [async function](https://github.com/garden-io/garden/blob/master/core/src/util/recoverable-process.ts#L423), so it makes sense to keep the sleep call async too.

This PR also fixes state management in `RecoverableProcess` class. Now both `"stopped"` and `failed"` are the valid final states. The former stands for forced process tree stop, the latter -- for the failure after all retries have been used.

If necessary, it is possible to introduce an intermediate `"stopped"` state later, and define `"terminated"` and `"failed"` states as the final ones. But now it's not the case, the recoverable process trees should last for the duration of the Garden session.

Fixes #

### DIFF
--- a/core/src/util/recoverable-process.ts
+++ b/core/src/util/recoverable-process.ts
@@ -17,7 +17,7 @@ export interface OsCommand {
   readonly cwd?: string
 }
 
-const renderOsCommand = (cmd: OsCommand): string => JSON.stringify(cmd)
+export const renderOsCommand = (cmd: OsCommand): string => JSON.stringify(cmd)
 
 export interface ProcessMessage {
   readonly pid: number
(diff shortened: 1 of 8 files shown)


vvagaytsev:
This makes things worse, rejecting it. We have to use sync sleep calls.

core/src/util/recoverable-process.ts
```diff
@@ -261,19 +262,19 @@ export class RecoverableProcess {
     node.descendants.forEach((descendant) => RecoverableProcess.recursiveAction(descendant, action))
   }
 
-  private stopNode(): void {
+  private stopNode(state: RecoverableProcessState): void {
```
Orzelius, line 265: what if I pass `"running"` here, should that be allowed?

Orzelius (approved):
Looks good, other than the one comment that's probably just me not understanding how the state management works

core/src/util/recoverable-process.ts
vvagaytsev: Ah, that should be refactored. We definitely don't want to accept any state here. That was a quick fix, I forgot to do it properly, good catch!

I'll push some changes soon.

vvagaytsev: Fixed this in https://github.com/garden-io/garden/pull/3054/commits/aa152c9d9113685d8e82e1514bc0d97119ca5a6c

vvagaytsev: I also added some unit tests and did some refactoring.

Orzelius (dismissed):

core/src/util/recoverable-process.ts
Orzelius: > I also added some unit tests and did some refactoring.

I think you forgot to commit that

vvagaytsev: True :) Got it rejected by the linter, and now everything is fixed and pushed.

Orzelius (approved):

PR #3056, opened 2022-07-20
refactor: renamed delete commands

vvagaytsev:
This PR only contains the user-facing parts for now. We can rename internals later. The internal includes the classes like `DeleteXxx[Command|Params|Result]`, some handlers, methods and functions in the action router, local variables, etc.

Changing the internal now can cause extra conflicts with https://github.com/garden-io/garden/tree/graph-v2.

Fixes #

### DIFF
--- a/core/src/commands/delete.ts
+++ b/core/src/commands/delete.ts
@@ -23,6 +23,7 @@ import { ConfigGraph } from "../config-graph"
 import { LogEntry } from "../logger/log-entry"
 import { uniqByName } from "../util/util"
 
+// TODO-G2 rename this to CleanupCommand, and do the same for all related classes, constants, variables and functions
 export class DeleteCommand extends CommandGroup {
   name = "cleanup"
   alias = "delete"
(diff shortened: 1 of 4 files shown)


core/src/commands/delete.ts
```diff
@@ -24,9 +24,8 @@ import { LogEntry } from "../logger/log-entry"
 import { uniqByName } from "../util/util"
 
 export class DeleteCommand extends CommandGroup {
-  name = "delete"
-  alias = "del"
```
edvald, line 28: I'd suggest keeping `delete` as an alias.

core/src/commands/delete.ts
```diff
@@ -99,35 +98,34 @@ interface DeleteEnvironmentResult {
 }
 
 export class DeleteEnvironmentCommand extends Command<{}, DeleteEnvironmentOpts> {
-  name = "environment"
-  alias = "env"
-  help = "Deletes a running environment."
+  name = "namespace"
+  alias = "ns"
```
edvald, line 102: Maybe throw a TODO-G2 comment in here. I had already made a plural `aliases` field in the `graph-v2` branch, and it'd be good to have `environment` and `env` in there as aliases as well.

Orzelius (approved):
Looks good! I guess the renaming of the files will be done afterwards as well

vvagaytsev:
@Orzelius thank you! Yes, we'll rename the files later.

PR #3069, opened 2022-07-22
improvement: support only single dotIgnoreFile

vvagaytsev:
This PR simplifies the `GitHandler` code to support a single ignore file instead of multiple ones. It's a **breaking** change.

Fixes #


core/src/vcs/vcs.ts
```diff
@@ -94,7 +94,7 @@ export abstract class VcsHandler {
   constructor(
     protected projectRoot: string,
     protected gardenDirPath: string,
-    protected ignoreFiles: string[],
+    protected ignoreFile: string,
```
vvagaytsev, line 97: Should we make this optional `protected ignoreFile?: string`?

core/src/vcs/git.ts
```diff
@@ -281,15 +279,9 @@ export class GitHandler extends VcsHandler {
     // List tracked but ignored files (we currently exclude those as well, so we need to query that specially)
     // TODO: change in 0.13 to no longer exclude these
```
vvagaytsev, line 280: Was this comment about executing `git` command for multiple files? If yes, then we can remove this TODO now.

core/test/unit/src/garden.ts
```diff
@@ -2491,11 +2491,15 @@ describe("Garden", () => {
     })
 
     it("should respect custom dotignore files", async () => {
-      const projectRoot = getDataDir("test-projects", "dotignore")
+      // In this project we have custom dotIgnoreFile: .customignore which overrides the default .gardenignore.
+      // Thus, all exclusions from .gardenignore will be skipped.
+      const projectRoot = getDataDir("test-projects", "dotignore-custom")
```
vvagaytsev, line 2496: Project `"dotignore-custom"` has been unused, and this case case has been equivalent to the one above (`should respect .gitignore and .gardenignore files`).

Now, this test uses the right project and has the corrected assertions.

core/src/vcs/git.ts
edvald: Actually this comment can be removed. We no longer plan on changing this behavior.

core/src/config/project.ts
```diff
@@ -307,18 +307,22 @@ export const projectDocsSchema = () =>
...
+
+        Note: prior to Garden 0.13.0, it was possible to specify _multiple_ ".ignore" files using the \`dotIgnoreFiles\` field in the project configuration.
 
         Note that these take precedence over the project \`module.include\` field, and module \`include\` fields, so any paths matched by the .ignore files will be ignored even if they are explicitly specified in those fields.
 
         See the [Configuration Files guide](${DOCS_BASE_URL}/using-garden/configuration-overview#including-excluding-files-and-directories) for details.
       `
         )
-        .example([".gardenignore", ".gitignore"]),
+        .example(".gardenignore"),
```
edvald, line 325: Maybe the example could be `.gitignore`?

edvald (changes requested):
Looks good! It does occur to me though that we should ease the migration here.

Maybe we could add something here: https://github.com/garden-io/garden/blob/9f639397050f1277c93502572ee9c5bcda9ccfb8/core/src/config/base.ts#L117

Basically if we find a Project kind, we do some similar pre-processing as we do with Module kinds. If `dotIgnoreFiles` is set with exactly one value, we print a yellow deprecation warning but map it to the new `dotIgnoreFile` field. If there's more than one value, we throw an error and tell people should specify only one file with the `dotIgnoreFile` field.

edvald:
Also, please make sure to add a `BREAKING CHANGE:` clause to the commit, so that we get it in the changelogs later.

core/src/config/project.ts
vvagaytsev: Yes, makes sense.

core/src/vcs/vcs.ts
vvagaytsev: Actually, it's better to keep it non-optional and supply a default value.

vvagaytsev:
@edvald I've implemented automated migration, see the latest commits:
* https://github.com/garden-io/garden/pull/3069/commits/3f26eebaa22ba1ce077c62d0c252c505a4fed227 - restored old config field `dotIgnoreFiles`
* https://github.com/garden-io/garden/pull/3069/commits/f9701fb2a7945041adca8d3d61609a9c9de4b07c - implemented automatic remapping + unit test
* https://github.com/garden-io/garden/pull/3069/commits/4aa80578e6e5d9123b957ca56cc22a8410239dd2 - excessive integration tests

What is a good place to print a warning in case if `dotIgnoreFiles` array contains a single value? Function [`prepareResource`](https://github.com/garden-io/garden/blob/9f639397050f1277c93502572ee9c5bcda9ccfb8/core/src/config/base.ts#L117) does not have a logger in its context.

edvald:
I think it's sensible to pass `log` to the `prepareResource` function, might as well just do that imo.

vvagaytsev:
@edvald now the deprecation warning is there:
* https://github.com/garden-io/garden/pull/3069/commits/4d13dc7fc3911173283435f2b2bb1f0eea8b7a91 - print the warninig
* https://github.com/garden-io/garden/pull/3069/commits/d0f643c2780b9c7f5958cab4cf208818522addc4 - deduplicate the warning

Orzelius (approved):
approve for merge permission (I didn't take a look myself, trusting Jon's review)

core/src/vcs/git.ts
```diff
@@ -279,17 +277,10 @@ export class GitHandler extends VcsHandler {
...
+      !this.ignoreFile
         ? []
-        : flatten(
-            await Promise.all(
-              this.ignoreFiles.map((f) =>
-                git("ls-files", "--ignored", ...lsFilesCommonArgs, "--exclude-per-directory", f)
-              )
-            )
-          )
+        : await git("ls-files", "--ignored", ...lsFilesCommonArgs, "--exclude-per-directory", this.ignoreFile)
```
vvagaytsev, line 283: @edvald I wanna double-check the old comment "TODO: change in 0.13 to no longer exclude these". Did it mean that we want Git to track the files listed in a .ignore-file?

If yes, then we need to remove this statement.

edvald: That TODO no longer applies. We should retain the current behavior and simply remove the TODO comment itself.

vvagaytsev:
@edvald this is ready for review. If we're fine with the in-memory warnings tracking implemented in https://github.com/garden-io/garden/pull/3069/commits/7a4ca0ff518c031881241407e44a9d99669e3a39, then we can merge it. Please re-review once you have time.

edvald (approved):

edvald:
Approved. Please make sure there's a `BREAKING CHANGE` clause in the squashed commit 🙏

PR #3071, opened 2022-07-25
fix: support --dependants-first option in `delete service` command

vvagaytsev:
Fixes the issue reported in https://github.com/garden-io/garden/issues/2782#issuecomment-1193226037.

Fixes #


thsig (approved):
My bad, not sure how this slipped through!

Orzelius (approved):

PR #3073, opened 2022-07-27
perf: optimized file watcher

vvagaytsev:
Need to avoid watcher issues by explicitly not detecting added modules. This allows us to reduce the scope of what the watcher is looking at by default, and more aggressively apply exclusions.

Need to change the default watcher scope to just config files + module/action roots, don’t watch the whole project root unless that’s also a module/action root.

Fixes #


core/src/garden.ts
```diff
@@ -396,19 +428,9 @@ export class Garden {
     bufferInterval?: number
   }) {
     const modules = graph.getModules()
-    const linkedPaths = (await getLinkedSources(this)).map((s) => s.path)
-    const paths = [this.projectRoot, ...linkedPaths]
+    const paths = await this.getWatchablePaths(modules)
+    const skipPaths = await this.getSkipPaths(skipModules)
```
edvald, line 432: I think we may be missing an important case here: What if one enabled module shares a path with a disabled module?

vvagaytsev: Ah, you're right! A possible scenario is that we have multi-module config and one of the modules is disabled.
Then we can watch the entire dir. Are there any other possible scenarios covering this case?

vvagaytsev: @edvald do we need to watch configs of disabled modules?

vvagaytsev: I added a test case in https://github.com/garden-io/garden/pull/3073/commits/1f6162a91cd3634be0a5e434c5f19cc69ec60bbc.

And I included only enabled modules to the watchlist in https://github.com/garden-io/garden/pull/3073/commits/dacfbe4eee9f70245b43e7ee03ac024791e8970a.

core/test/data/test-project-watch-flat-disabled/module.garden.yml
```diff
@@ -0,0 +1,33 @@
+kind: Module
```
edvald, line 1: A general note: We tend to prefer putting module configs inline in tests instead of making more and more test projects in the data directory. Look for tests that use `garden.setModuleConfigs(...)` to see examples of that.

core/src/garden.ts
edvald: The changes cover explicitly disabled modules nicely now, I _think_, but a similar case remains where you have two enabled modules with the same root path and one is in dev mode, the other not. I suspect the flow needs to be a little different to accommodate all these scenarios.

core/test/data/test-project-watch-flat-disabled/module.garden.yml
vvagaytsev: Ok, thanks! Is there an easy way to dump `garden.yml` config files to disk when project and module configs are created programmatically? Maybe we have some helper method.

vvagaytsev: I retained only one new test project in https://github.com/garden-io/garden/pull/3073/commits/96192166588cccdae46c6f8565723c31e2e01874 because I need to have config files on the disk.

core/src/garden.ts
vvagaytsev: Looks like the dev/local-mode modules case is already covered. Modules with enabled `devMode` (and `localMode`) are passed over the `skipModules` argument, it's getting resolved later, in the [`Watcher.start(...)`](https://github.com/garden-io/garden/blob/master/core/src/watch.ts#L108). 

We can handle `devMode`/`localMode` modules here, instead of unwatching them later. In the `Watcher` class we can process only `Garden.moduleExcludePatterns`, wdyt?

vvagaytsev:
This has been irrelevant since #3702 was merged.

PR #3078, opened 2022-07-28
chore: require environments field in project configuration

Orzelius:
Now environment field is required in project configs.
This also eliminates the need for an internal default env (used to be "local"), now if default env is not set just the first one will be selected (as before) and error is thrown if envs is undefined or an empty array.

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -38,11 +38,11 @@ import {
 } from "./helpers"
 import { Parameters, globalOptions, OUTPUT_RENDERERS, GlobalOptions, ParameterValues } from "./params"
 import {
-  defaultEnvironments,
   ProjectConfig,
   defaultNamespace,
   parseEnvironment,
   ProjectResource,
+  EnvironmentConfig,
 } from "../config/project"
 import { ERROR_LOG_FILENAME, DEFAULT_API_VERSION, DEFAULT_GARDEN_DIR_NAME, LOGS_DIR_NAME } from "../constants"
 import { generateBasicDebugInfoReport } from "../commands/get/get-debug-info"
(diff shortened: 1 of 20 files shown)


core/src/cli/cli.ts
```diff
@@ -66,9 +66,9 @@ import { renderDivider } from "../logger/util"
 import { emoji as nodeEmoji } from "node-emoji"
 
 export async function makeDummyGarden(root: string, gardenOpts: GardenOpts) {
-  const environments = gardenOpts.environmentName
+  const environments: EnvironmentConfig[] = gardenOpts.environmentName
     ? [{ name: parseEnvironment(gardenOpts.environmentName).environment, defaultNamespace, variables: {} }]
-    : defaultEnvironments
+    : [{ defaultNamespace: "default", name: "dummy-env", variables: {} }]
```
edvald, line 71: I'm not quite sure where but I sort of suspect the `"dummy-env"` string might bleed through somewhere. Maybe just `"default"` instead?

core/src/config/project.ts
```diff
@@ -442,7 +423,8 @@ export function resolveProjectConfig({
       ...globalConfig,
       name,
       defaultEnvironment,
-      environments: [],
+      // environments are validated later
+      environments: [{ defaultNamespace: null, name: "fake-env-only-here-for-inital-load", variables: {} }],
```
Orzelius, line 427: I'm not too sure about this line here, if anyone has better ideas I'm all ears

edvald: Hmm. Does this actually need to be changed?

Orzelius: It is, if I leave it to an empty array it won't pass the schema check and if I pass the real ends the vars won't be resolved and I'll get regex match errors

core/src/cli/cli.ts
Orzelius: makes sense, fixed in [e0d27aa](https://github.com/garden-io/garden/pull/3078/commits/e0d27aa78210d5e9b4ac17f0caf10333e815be46)

edvald (approved):

edvald:
Approved 👍  When merging, please tag the squashed commit as `refactor` and include a `BREAKING CHANGE:` clause so that we catch it in changelogs.

PR #3080, opened 2022-07-29
refactor: removed deprecated options from garden commands

vvagaytsev:
This PR removes deprecated opts from garden CLI commands.

Fixes #

### DIFF
--- a/core/test/unit/src/commands/logs.ts
+++ b/core/test/unit/src/commands/logs.ts
@@ -260,17 +260,6 @@ describe("LogsCommand", () => {
 
       expect(out[0]).to.eql(msgColor("Yes, this is log"))
     })
-    it("should optionally show the container name", async () => {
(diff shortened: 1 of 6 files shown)


core/test/unit/src/commands/logs.ts
```diff
@@ -260,17 +260,6 @@ describe("LogsCommand", () => {
 
       expect(out[0]).to.eql(msgColor("Yes, this is log"))
     })
-    it("should optionally show the container name", async () => {
```
vvagaytsev, line 263: This can be removed as there is the dedicated test `"should optionally print tags with --show-tags"` in https://github.com/garden-io/garden/blob/5192f66162d5071efe7aabffb8676cb49d0a056c/core/test/unit/src/commands/logs.ts#L496

edvald (approved):

PR #3087, opened 2022-08-02
feat: `local.arch` template variable

### LINKED ISSUE #3086: [FEATURE]: local.arch template variable
## Feature Request
provide access to a `local.arch` template variable similar to [platform](https://docs.garden.io/reference/template-strings/modules#usd-local.platform)
### Background / Motivation
for local development environments there are some images that are specifically needed (i.e developing on M1's vs intels).
providing access to this would allow templated garden modules to swap out container ids based on  arch during development

ideally it should beable to be used the same way as `local.platform` or any other provided local variable simple example

currently without this feature we're having developers just manage this external to garden with bash scripts and if statements to setup operators with the correct images before running our garden workflow

similar to `local.platform`

vvagaytsev:
This PR adds `local.arch` template variable.

Fixes #3086

### DIFF
--- a/core/src/config/template-contexts/project.ts
+++ b/core/src/config/template-contexts/project.ts
@@ -51,7 +51,7 @@ class LocalContext extends ConfigContext {
         "A string indicating the platform that the framework is running on " +
           "(see https://nodejs.org/api/process.html#process_process_platform)"
       )
-      .example("posix")
+      .example("linux")
   )
   public platform: string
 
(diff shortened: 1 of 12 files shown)


vvagaytsev:
@edvald is it ok to merge it to the main branch, or should we include this in `0.13`?

Orzelius (approved):

vvagaytsev:
It looks like it's ok to merge it to the main branch. I've just checked the graph-v2, there were no changes done in https://github.com/garden-io/garden/blob/graph-v2/core/src/config/template-contexts/project.ts

PR #3088, opened 2022-08-02
improvement: removed error-prone local ip inference

vvagaytsev:
This PR removes the usage of `LocalAddress` entity. A default local address will be used if no specific address is given in the Garden env settings.

Fixes #


edvald:
One quick note here before proceeding: It would still be useful to remember the _port_ used as a fallback when the desired port is unavailable. I'm not sure we want to get into it in the same PR, but it's something to keep in mind. 

We could also, instead of remembering the port, just use a predictable progression of port numbers upon conflict so that we deterministically try alternative ports instead of randomly.

vvagaytsev:
@edvald sounds good! Let's create a separate task in the `0.13` shortlist for that. Having a progression of port numbers might be a better idea because we won't have to store any extra state in the DB.

vvagaytsev:
I removed the DB migration to ensure backward compatibility with 0.12.x.

edvald (approved):

PR #3095, opened 2022-08-05
chore: added a new cloud event with context of a command

mkhq:

Adds a new cloud event including more context related to the running command.

Fixes #

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -478,9 +478,8 @@ ${renderCommands(commands)}
           }
           this.bufferedEventStream.connect(connectParams)
           if (streamEvents) {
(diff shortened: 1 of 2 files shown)


thsig (dismissed):
Looks good to me! Would have been neat to simply add these fields to the `commandInfo` event, but this approach is probably more pragmatic with respect to API compatibility as things stand.

mkhq:
@thsig Thanks for the comment. I've tried another approach that extends `CommandInfo` creating a new Payload type to be passed in the event. Think this is a bit cleaner and also avoids an extra event send.

thsig (approved):

thsig (approved):

PR #3096, opened 2022-08-05
improvement(pulumi): improve preview output

thsig:

We now only show the output of `pulumi preview` when the generated plan is not a no-op. This makes the output much easier to parse visually during a manual review process.

We now also copy plans that aren't no-ops to a subdirectory of the preview directory for easier review and deployment, and write a total summary of the generated plans (including preview links and operation counts by affected service, and total counts).

The latter is designed to be useful for further automation, e.g. in a CI/CD context.


plugins/pulumi/cli.ts
```diff
@@ -160,4 +208,5 @@ export const pulumiCliSPecs: { [version: string]: PluginToolSpec } = {
 export const supportedVersions = Object.keys(pulumiCliSPecs)
 
 // Default to latest pulumi version
-export const defaultPulumiVersion = "3.25.1"
+export const defaultPulumiVersion = "3.40.0"
+// export const defaultPulumiVersion = "3.25.1"
```
vvagaytsev, line 212: Should this line be removed?

vvagaytsev (approved):
LGTM 🚀

plugins/pulumi/cli.ts
thsig: Yes it should, good catch!

PR #3097, opened 2022-08-05
chore: check for requirements being installed

Orzelius:
checks if git and rsync are installed with valid versions
and gives readable error otherwise.

check results are cached and are not checked again once passed

I opted for the global config instead of the sqlite because connecting to the db and starting it took too long.

### DIFF
--- a/core/src/build-staging/rsync.ts
+++ b/core/src/build-staging/rsync.ts
@@ -27,45 +27,51 @@ const versionDetectFailure = new RuntimeError(
   {}
 )
 
+/**
+ * throws if no rsync is installed or version is too old
+ */
+export async function validateRsyncInstall() {
+  let version: string | undefined = undefined
+
+  try {
+    const versionOutput = (await exec("rsync", ["--version"])).stdout
+    version = versionOutput.split("\n")[0].match(versionRegex)?.[1]
+  } catch (error) {
+    throw new RuntimeError(
+      deline`
+      Could not find rsync binary.
+      Please make sure rsync (version ${minRsyncVersion} or later) is installed and on your PATH.
+      `,
+      { error }
+    )
+  }
+
+  if (!version) {
+    throw versionDetectFailure
+  }
+
+  let versionGte = true
+
+  try {
+    versionGte = semver.gte(version, minRsyncVersion)
+  } catch (_) {
+    throw versionDetectFailure
+  }
+
+  if (!versionGte) {
+    throw new RuntimeError(
+      deline`
+      Found rsync binary but the version is too old (${version}).
+      Please install version ${minRsyncVersion} or later.
+      `,
+      { version }
+    )
+  }
+}
+
 export class BuildDirRsync extends BuildStaging {
   static async factory(projectRoot: string, gardenDirPath: string) {
-    // Make sure rsync is installed and is recent enough
-    let version: string | undefined = undefined
-
-    try {
-      const versionOutput = (await exec("rsync", ["--version"])).stdout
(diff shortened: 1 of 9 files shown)


core/src/cli/helpers.ts
```diff
@@ -69,6 +71,14 @@ export async function checkForStaticDir() {
   }
 }
 
+/**
+ * checks if requirements to run garden are installed, throws if they are not
+ */
+export async function checkRequirements() {
+  await validateRsyncInstall()
```
vvagaytsev, line 78: If we add more checks in the future, we can use `Promise.allSettled()` here. It executes all checks asynchronously and collects the results. If multiple requirements are not satisfied, we'll be able to show all errors at once.

core/src/build-staging/rsync.ts
```diff
@@ -27,45 +27,51 @@ const versionDetectFailure = new RuntimeError(
...
-        Could not find rsync binary.
-        Please make sure rsync (version ${minRsyncVersion} or later) is installed and on your PATH.
-        `,
-        { error }
-      )
-    }
+/**
+ * throws if no rsync is installed or version is too old
+ */
+export async function validateRsyncInstall() {
```
vvagaytsev, line 33: This and `validateGitInstall` have the same structure of the function body. If we add more checkers in the future, it might make sense to extract a template function that accepts only necessary tool-specific details.

vvagaytsev (approved):
LGTM! Left some comments, but those are not mandatory, and can be addressed in a separate PR later.

edvald:
Haven't had a chance to comment, but have one or two thoughts. Will post later today.

core/src/cli/cli.ts
```diff
@@ -764,3 +768,32 @@ ${renderCommands(commands)}
...
+export async function validateRuntimeRequirementsCached(
+  log: Logger,
+  globalConfig: GlobalConfigStore,
+  requirementCheckFunction: () => Promise<void>
+) {
+  let requirementsCheck: RequirementsCheck | undefined
+  try {
+    requirementsCheck = (await globalConfig.get(["requirementsCheck"])) as RequirementsCheck
+  } catch {} // throws if requirementsCheck not yet populated
+  if (!requirementsCheck || !requirementsCheck.passed) {
```
edvald, line 781: What if we add or modify a requirement at some point? I figure we need to be able to invalidate based on what's being checked for, even if it's just an incrementing constant we update.

core/src/cli/helpers.ts
edvald: Another potential issue is that we don't _always_ require rsync. If you set `GARDEN_K8S_BUILD_SYNC_MODE=mutagen` (which btw we should make default in 0.13) and are not running Windows, rsync is not needed.

core/src/cli/cli.ts
Orzelius: That's why I added the `lastRunGardenVersion` [field](https://github.com/garden-io/garden/pull/3097/files#diff-a5f5e0054cf73a57a1857a593d39daadf159a90d1cbf3d748771928e3719b3a5R281). In the future we can check by that and rerun the validation

core/src/build-staging/rsync.ts
Orzelius: good point, extracted in [f1002ed](https://github.com/garden-io/garden/pull/3097/commits/f1002ed6c4a5c16db4a699dd880c42ddb47de5d6) which also allowed to test the logic so that's what I did in [e336321](https://github.com/garden-io/garden/pull/3097/commits/e3363217f9967b9920af4d0865761e5269133d23)

edvald (approved):

PR #3102, opened 2022-08-10
WIP: refactor(k8s): use faux SSH command to switch away from Mutagen fork

### LINKED ISSUE #5420: Move away from Mutagen fork
Right now we use a Mutagen fork to enable syncing to Kubernetes Pods.

This prevents us from updating to the latest version of Mutagen, which means that users might be exposed to bugs that are already fixed in the latest version of Mutagen.

See also https://github.com/garden-io/garden/pull/3102

xenoscopic:

This PR is a draft proposal for switching Garden back to the official Mutagen releases and dispensing with the Garden fork of Mutagen.  The last remaining blocker for this transition is Mutagen's lack of support for custom transports and Garden's need to use a `kubectl exec`-based transport.  To work around this limitation, this commit introduces a faux `ssh` executable (written in Go) that can be provided to the Mutagen daemon to serve as a transport.  This faux `ssh` command smuggles `kubectl exec` parameters through the SSH hostname argument (as Base64-encoded JSON) and then invokes `kubectl exec`, forwarding signals and standard input closure to ensure proper termination signaling from the Mutagen daemon.

This PR is fairly close to complete, but there are a few remaining questions:

1. Is this approach satisfactory overall?
2. Should the faux `ssh` command be rewritten in TypeScript?
3. How should the faux `ssh` command be source-controlled and distributed?
4. Are there any transition considerations for active `garden dev` commands?

Regarding (1), I think this is the best approach because it gets Garden back on the official Mutagen releases with the minimal amount of effort and the implementation strategy is very similar to what custom transports will eventually look like in Mutagen, meaning that the faux SSH command will probably be convertible to a `mutagen-transport-garden` (name TBD) command later on.

Regarding (2), that's a policy decision I'd leave to Garden.  The critical aspect is just the proper signal and standard input closure forwarding.

Regarding (3), that's probably a good point for discussion.  I would imagine it would be distributed/downloaded in a similar manner to the `mutagen` executable, but I don't know the best location to perform its build or serve its binaries.  There's also the consideration of its versioning w.r.t. Garden, especially if the format of the JSON changes, but that's likely to see very little churn once implemented, so manually managing its versioning as some external project might be suitable.  Once this is decided and implemented, the `MUTAGEN_SSH_PATH` environment variable will need to be set for `mutagen daemon run`; a TODO has been indicated for this in the code.

Regarding (4), the simple answer is that the project would probably just need to be restarted entirely.  If the same `MUTAGEN_DATA_DIRECTORY` is used, then the standard Mutagen daemon will just reject and ignore the prior `exec`-based sessions, but testing of this would be good before shipping.

CC @thsig @vvagaytsev

Fixes #5420

### DIFF
--- a/core/src/plugins/kubernetes/mutagen.ts
+++ b/core/src/plugins/kubernetes/mutagen.ts
@@ -162,6 +162,7 @@ export async function ensureMutagenDaemon(ctx: PluginContext, log: LogEntry) {
       env: {
         MUTAGEN_DATA_DIRECTORY: dataDir,
         MUTAGEN_LOG_LEVEL: "debug",
+        // TODO: Add MUTAGEN_SSH_PATH with Garden-provided faux SSH
       },
       maxRestarts,
       sleep: 3000,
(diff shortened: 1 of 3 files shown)


vvagaytsev:
Thanks for opening this @xenoscopic! We'll take a look a bit later, now we're focused on the preparation of the `0.13` release branch.

vvagaytsev:
Superseded by #5551.

PR #3103, opened 2022-08-10
refactor: remove deprecated providers config

vvagaytsev:
BREAKING CHANGE

The old field `environments[].providers[]` is no longer supported in the project configuration.

Fixes #


Orzelius (changes requested):
There are still a few examples and test projects left that use the old api. Search for `^  providers` and you'll find 'em

vvagaytsev:
> There are still a few examples and test projects left that use the old api. Search for `^ providers` and you'll find 'em

Yes, I know. There is only one project left -- it is https://github.com/garden-io/garden/blob/master/core/test/data/test-projects/v10-configs/garden.yml. It's used in https://github.com/garden-io/garden/blob/master/core/test/unit/src/commands/migrate.ts to test the migrations from the older Garden versions. So, I left it without changes. Migration tests do not cause any config schema validation failures.

Orzelius:
There are also some docs and examples that use that

vvagaytsev:
@Orzelius the examples and guides have been fixed in https://github.com/garden-io/garden/pull/3103/commits/3258356e5212ce2c07c508a9f8f4bc034bb0e4cb.

Orzelius (approved):

PR #3105, opened 2022-08-15
feat: configurable maven phases in jib modules

### LINKED ISSUE #3026: [FEATURE]: Option for exposing package phase of maven with jib-module
## Feature Request
Garden should expose an option to use any maven phases [clean compile/clean package] in jib-module. 

### Background / Motivation
We have a use case to package the fat jar into jib which get build using `mvn clean package jib:build` but there is a limitation with garden that it embeds the maven command `mvn compile jib:build` for jib-module.

Developer should have an option while configuring jib-module, to use any phase [clean compile/clean package]. Garden will use the defined phase to build the jib image. 

### Why do they want to do this?

Without this option, we need to defined the separate exec module having custom mvn command 
`mvn clean package jib:build -Djib.to.image=${var.dockerRepo}/${parent.name}:$GARDEN_MODULE_VERSION`

#### Priority: Probably pretty low, though I would greatly appreciate it :)

vvagaytsev:
This PR allows configurable Maven phases in `jib-module`.

Fixes #3026

The changes are backward compatible. This PR adds an optional configuration field called `mavenPhases`. The Maven `compile` phase is used as a default value.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202534139264934
  - https://app.asana.com/0/0/1202799942680967


plugins/jib/index.ts
```diff
@@ -66,6 +66,14 @@ const jibModuleSchema = () =>
         .allow("docker", "oci")
         .default("docker")
         .description("Specify the image format in the resulting tar file. Only used if `tarOnly: true`."),
+      mavenPhases: joi
+        .object()
+        .optional()
+        .keys({
+          start: joi.string().optional().description("Start Maven phase."),
```
thsig, line 73: I'm not an expert on maven, but why don't we simply allow an array of phases (strings) here, with the default being `["compile"]`?

I have a feeling that a user will end up requesting support for 3+ phases at some point.

edvald: I agree, best to leave it open imo.

vvagaytsev: Right, setting start and end goals is the usual use case, but it's still possible to add more custom phases. I'll change this.

vvagaytsev: Fixed in https://github.com/garden-io/garden/pull/3105/commits/f5ad5cc435f3652fa2ab976a4c87fb7bb7584ed4

examples/jib-container/spring-boot/spring-boot.garden.yml
```diff
@@ -8,6 +8,9 @@ build:
   # Note: You could also specify a deploymentRegistry or set the `image` field if you prefer to push to a registry,
   #       but that'll obviously be much slower.
   dockerBuild: ${environment.name == "local"}
+  mavenPhases:
+    start: clean
+    end: package
```
edvald, line 13: This needs to be updated

plugins/jib/index.ts
```diff
@@ -66,6 +66,11 @@ const jibModuleSchema = () =>
         .allow("docker", "oci")
         .default("docker")
         .description("Specify the image format in the resulting tar file. Only used if `tarOnly: true`."),
+      mavenPhases: joi
+        .array()
+        .items(joi.string())
+        .optional()
+        .description("Defines the Maven phases to be executed during the Garden build step."),
```
edvald, line 73: Shouldn't this have a default value of `["compile"]` for clarity?

examples/jib-container/spring-boot/spring-boot.garden.yml
vvagaytsev: Fixed in https://github.com/garden-io/garden/pull/3105/commits/89f6537dad940a396744c8f1e80cb2c7be8d22c2.

plugins/jib/index.ts
vvagaytsev: Changed in https://github.com/garden-io/garden/pull/3105/commits/ee90a6e0f1460f2b8abb1381107488c8bf1e6f59.

plugins/jib/maven.ts
```diff
@@ -104,3 +104,7 @@ export async function mvn({
     return res
   })
 }
+
+export function resolveMavenPhases(mavenPhases?: string[]): string[] {
+  return !mavenPhases || mavenPhases.length === 0 ? ["compile"] : mavenPhases
```
edvald, line 109: Is this necessary, after the default has been set on the schema? Also, I think it should be allowed to put in an empty array.

vvagaytsev: Right, this can be simplified. And an empty array is still a valid case if we don't want to rerun a Maven build. Changed this in https://github.com/garden-io/garden/pull/3105/commits/d7f423f5976cd3217d99a37c11b67489b5f8d7f6.

vvagaytsev:
@thsig @edvald I've applied some fixes based on the latest CR. Can you review this again, please?

PR #3108, opened 2022-08-16
improvement: garden namespace truncation

Orzelius:
Truncate garden namespaces when needed. If truncated a hash is added.

image

### DIFF
--- a/core/src/config/common.ts
+++ b/core/src/config/common.ts
@@ -281,19 +281,32 @@ joi = joi.extend({
   messages: {
     base: "{{#label}} must be a valid environment name or <namespace>.<environment>",
     multipleDelimiters: "{{#label}} may only contain a single delimiter",
+    invalidNamespace:
+      "invalid namespace specified, {{#label}} must be a valid environment name or <namespace>.<environment>",
   },
   validate(value: string, { error }) {
-    const baseSchema = joi.string().hostname()
-    const result = baseSchema.validate(value)
-
-    if (result.error) {
-      return { value, errors: error("base") }
-    }
-
     const split = value.split(".")
 
     if (split.length > 2) {
       return { value, errors: error("multipleDelimiters") }
+    }
+
+    if (split.length === 2) {
+      const resultNs = joiIdentifierLenLimit255().validate(split[0])
+      if (resultNs.error) {
+        return { value, errors: error("invalidNamespace") }
+      }
+      const resultEnv = environmentNameSchema().validate(split[1])
+      if (resultEnv.error) {
+        return { value, errors: error("base") }
+      }
+    }
+
+    if (split.length === 1) {
+      const resultEnv = environmentNameSchema().validate(split[0])
+      if (resultEnv.error) {
+        return { value, errors: error("base") }
+      }
     }
 
     return { value }
--- a/core/src/config/common.ts
+++ b/core/src/config/common.ts
@@ -281,8 +281,7 @@ joi = joi.extend({
   messages: {
     base: "{{#label}} must be a valid environment name or <namespace>.<environment>",
     multipleDelimiters: "{{#label}} may only contain a single delimiter",
-    invalidNamespace:
-      "invalid namespace specified, {{#label}} must be a valid environment name or <namespace>.<environment>",
+    invalidNamespace: "invalid namespace specified, must be a valid environment name or <namespace>.<environment>",
   },
   validate(value: string, { error }) {
     const split = value.split(".")
--- a/core/src/config/common.ts
+++ b/core/src/config/common.ts
@@ -607,3 +607,6 @@ export function allowUnknown<T extends Joi.Schema>(schema: T) {
 
   return schema
 }
+
+export const environmentNameSchema = () =>
+  joiUserIdentifier().required().description("The name of the environment.").example("dev")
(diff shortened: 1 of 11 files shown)


core/src/config/project.ts
```diff
@@ -576,7 +578,10 @@ export const pickEnvironment = profileAsync(async function _pickEnvironment({
     projectRoot: projectConfig.path,
   })
 
-  namespace = getNamespace(environmentConfig, namespace)
+  const namespace = truncateGardenNamespaceIfNeeded(
```
Orzelius, line 581: with current implementation if a user specifies a namespace via the cli that exceeds the limits it will be truncated. I'm not 100% sure if that's the best way to go about it as it might be confusing and unexpected

Orzelius, line 581: Currently there is no warning shown when a truncation takes place as the logger is not yet initialized at this time. If anybody has ideas on how to get a warning out I'm all ears

edvald: The logger should be initialized by this time, just hasn't been passed to the function. So that should be easy to add in.

core/src/config/common.ts
```diff
@@ -447,6 +448,10 @@ export const joiIdentifierDescription =
   "valid RFC1035/RFC1123 (DNS) label (may contain lowercase letters, numbers and dashes, must start with a letter, " +
   "and cannot end with a dash) and must not be longer than 63 characters."
 
+export const joiIdentifierDescriptionLen255 =
+  "string that may contain lowercase letters, numbers and dashes, must start with a letter, " +
+  "and cannot end with a dash) and must not be longer than 255 characters."
```
vvagaytsev, line 453: There is no opening parenthesis matching the "dash)".

edvald:
There was an actual good reason to keep namespace names at 63 characters, namely that it's the maximum length of a DNS label. Besides, it's going to be annoying in the web UI to have to support up to 255 characters for an environment/namespace name. So unfortunately I think we probably should stick to that limit.

core/src/config/project.ts
```diff
@@ -613,30 +618,50 @@ export const pickEnvironment = profileAsync(async function _pickEnvironment({
...
+  const lenProjName = projectName.length
+  const maxK8sNsLen = 63
+
+  // don't truncate if <namespace>-<projectName> is under max
+  // as that is often used for remote k8s namespaces
+  if (lenNs + lenProjName + 1 <= maxK8sNsLen) {
+    return namespace
+  }
+
+  const sliceLen = maxK8sNsLen - 1 - lenProjName - 6 // last five chars will be a hash and a '-'
```
vvagaytsev, line 635: Nit-pick :) It might be more clear to write something like "reserve last 6 characters for 5-char hash and a dash-symbol ('-')". Otherwise, it might look like 5 chars are reserved for hash and '-'. meaning that hash takes only 4 chars :)

core/src/config/project.ts
```diff
@@ -613,30 +618,50 @@ export const pickEnvironment = profileAsync(async function _pickEnvironment({
...
+  const nonHashedPart = namespace.slice(0, sliceLen)
+  const fiveCharHas = hasha(namespace).slice(-5)
```
vvagaytsev, line 637: This could be renamed to `fiveCharHash`.

Orzelius:
> There was an actual good reason to keep namespace names at 63 characters, namely that it's the maximum length of a DNS label. Besides, it's going to be annoying in the web UI to have to support up to 255 characters for an environment/namespace name. So unfortunately I think we probably should stick to that limit.

the 255 limit is just so the validation doesn't fail if more than 63 is inputted (most likely via templating), internally it will always be truncated to less than 63. If we set the limit to 63 then there is no truncation to be implemented in the first place.

You are right in that I didn't think about how this would go with cloud. I will pair with somebody from there to make sure this is compatible.

edvald:
Aha, I see! That makes sense. But yes, give them a quick ping as well.

core/src/config/project.ts
Orzelius: okay with [341a3d1](https://github.com/garden-io/garden/pull/3108/commits/341a3d1f95837bc355dbe085d88c221ed8e66403) it is truncated

core/src/config/project.ts
Orzelius: yep :) [ed4d3d1](https://github.com/garden-io/garden/pull/3108/commits/ed4d3d1ab5dffffe2678cf08f902c8b396b1b7bd)

core/src/config/project.ts
Orzelius: nice catch [829786c](https://github.com/garden-io/garden/pull/3108/commits/829786cb2ec05f48c10c3558d943b161636e7db3)

Orzelius: nice catch [829786c](https://github.com/garden-io/garden/pull/3108/commits/829786cb2ec05f48c10c3558d943b161636e7db3)

Orzelius: good point. I also elaborated on the `- 1` as I myself got confused for a bit about it being there [1344cea](https://github.com/garden-io/garden/pull/3108/commits/1344cea1d59b59cb2263353e3aa55b037a8ecdb9)

core/src/config/common.ts
Orzelius: good catch [829786c](https://github.com/garden-io/garden/pull/3108/commits/829786cb2ec05f48c10c3558d943b161636e7db3)

PR #3116, opened 2022-08-22
docs: add permissions information

thelongshanx:
Adds information about permissions required for remote clusters

Fixes #

### DIFF
--- a/docs/guides/remote-kubernetes.md
+++ b/docs/guides/remote-kubernetes.md
@@ -92,7 +92,7 @@ images when deploying. This should generally be a _private_ container registry, 
 public registry.
 
 Similarly to the below TLS configuration, you may also need to set up auth for the registry using K8s Secrets, in this
-case via the `kubectl create secret docker-registry` helper. You can read more about using and setting up private 
+case via the `kubectl create secret docker-registry` helper. You can read more about using and setting up private
 registries [here](https://kubernetes.io/docs/concepts/containers/images/#using-a-private-registry).
 
 _Note that you do not need to configure the authentication and imagePullSecrets when using GKE along with GCR,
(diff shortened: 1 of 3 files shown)


docs/guides/remote-kubernetes.md
```diff
@@ -133,8 +133,35 @@ By default, Garden will not install an ingress controller for remote environment
...
+        hostname: ${local.env.PR_NUMBER}-my-project.preview.mydomain.com
+providers:
+  - name: kubernetes
+    environments: [dev, preview]
+    namespace: ${environment.namespace}
+    buildMode: cluster-buildkit
+    defaultHostname: ${var.hostname}
+```
+
+If you would like to manage TLS for development environments, we recommend using an application load balancer and your cloud provider's certificate management service. You can find the documentation for (AWS here)[https://aws.amazon.com/premiumsupport/knowledge-center/associate-acm-certificate-alb-nlb/] and for (GCP here)[https://cloud.google.com/load-balancing/docs/ssl-certificates/google-managed-certs].
```
twelvemo, line 164:
```suggestion
If you would like to manage TLS for development environments, we recommend using your cloud provider's certificate management service in combination with a load balancer. You can find the documentation for (AWS here)[https://aws.amazon.com/premiumsupport/knowledge-center/associate-acm-certificate-alb-nlb/] and for (GCP here)[https://cloud.google.com/load-balancing/docs/ssl-certificates/google-managed-certs].
```
This would also work with other types of load balancers like network loadbalancers, so removing the application here.
I have been thinking once more about this section and I think it requires its own page with detailed instructions for the different possibilities and cloud providers. The sentence above is actually not enough to really help someone new to Kubernetes implement load balancing with TLS. 
I think we can merge this for now but I would really like to create a proper guide around it that covers different cloud providers and options in more detail because something that works for EKS in AWS might not work for GKE Autopilot clusters in GCP. cc @worldofgeese

twelvemo (changes requested):
Thanks for creating the PR again! Re-reading it had a couple of further comments.

docs/guides/remote-kubernetes.md
thelongshanx: Thank you! I've made an Asana ticket for this 👍🏻

twelvemo (approved):
👍

PR #3120, opened 2022-08-23
Graph v2 build tests albert

Orzelius:
fix build issues in a couple of unit test modules.

### DIFF
--- a/core/test/unit/src/config/template-contexts/module.ts
+++ b/core/test/unit/src/config/template-contexts/module.ts
@@ -13,12 +13,13 @@ import { keyBy } from "lodash"
 import { ConfigContext } from "../../../../../src/config/template-contexts/base"
 import { expectError, makeTestGardenA, TestGarden } from "../../../../helpers"
 import { prepareRuntimeContext } from "../../../../../src/runtime-context"
-import { GardenService } from "../../../../../src/types/service"
 import { resolveTemplateString } from "../../../../../src/template-string/template-string"
 import { ModuleConfigContext } from "../../../../../src/config/template-contexts/module"
 import { WorkflowConfigContext, WorkflowStepConfigContext } from "../../../../../src/config/template-contexts/workflow"
 import { GardenModule } from "../../../../../src/types/module"
 import { ConfigGraph } from "../../../../../src/graph/config-graph"
+import { BuildAction } from "../../../../../src/actions/build"
+import { GraphResults } from "../../../../../src/graph/results"
 
 type TestValue = string | ConfigContext | TestValues | TestValueFunction
 type TestValueFunction = () => TestValue | Promise<TestValue>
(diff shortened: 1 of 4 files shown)


core/test/unit/src/enterprise/event-handlers.ts
```diff
@@ -67,7 +67,6 @@ describe("cloudEventHandlers", () => {
       })
       expect(deployTask["devModeDeployNames"]).to.eql([])
       expect(deployTask["localModeDeployNames"]).to.eql([])
-      expect(deployTask.service.name).to.eql("service-a")
```
vvagaytsev, line 70: I think it's better to preserve this condition. Otherwise, we'll weaken the test semantics. It can be commented out with a todo-label, or rewritten like `expect(deployTask["service"].name).to.eql("service-a")`

Now the method `cloudEventHandlers.deployRequested(..,)` returns an array of `DeployTask` instead of a single task. So, in future, this will likely be fixed with `expect(deployTask[0].action.name).to.eql("service-a")`.

vvagaytsev (changes requested):
Left a comment, please check it.

core/test/unit/src/enterprise/event-handlers.ts
Orzelius: Good catch, fixed in [200a3d4](https://github.com/garden-io/garden/pull/3120/commits/200a3d4496da7dd5d539a8e5b77ef14ee61491d7). I'm actually not sure why I had removed this line originally, probably by accident or it was just lost in the work progress.

core/test/unit/src/config/template-contexts/module.ts
```diff
@@ -136,56 +141,36 @@ describe("ModuleConfigContext", () => {
...
     let withRuntime: ModuleConfigContext
-    let serviceA: GardenService
+    let buildA: BuildAction
 
     before(async () => {
       const modules = graph.getModules()
-      serviceA = graph.getService("service-a")
-      const serviceB = graph.getService("service-b")
-      const taskB = graph.getTask("task-b")
+      buildA = graph.getBuild("build-a")
```
vvagaytsev, line 148: I think we should preserve the dependencies between the actions here. I did something similar in https://github.com/garden-io/garden/pull/3121/files#diff-7dd75283963156981c657ddf2fcaed7827d6a421a11e33f46619c71dfe295e30R76 -- the missing part there is the `graphResult.setResult()` invocation. That will be fixed in the next iteration.

Let's comment out the old code to keep the past context visible. Afaiui, `getService(...)` should be mapped to `getDeploy(...)`.

Orzelius: I see, kinda like this c3b401a72a01e9ee1382aa91ec792f16313f117a then?

vvagaytsev: Yep, I think so :)

vvagaytsev (approved):

Orzelius:
rebased as there's some useful info in the commits and we can squash the collector branch later if we so wish

PR #3145, opened 2022-08-25
improvement: allow mutagen code sync outside subpath

Orzelius:
allow mutagen code sync outside subpath and add dev-mode to the [build dependencies example](https://github.com/garden-io/garden/tree/sync-build-deps/examples/build-dependencies)

### DIFF
--- a/core/src/plugins/container/config.ts
+++ b/core/src/plugins/container/config.ts
@@ -253,15 +253,14 @@ export const syncDefaultGroupSchema = () =>
 const devModeSyncSchema = () =>
   hotReloadSyncSchema().keys({
     exclude: syncExcludeSchema(),
-    source: joi // same as hotReloadSyncSchema but no subPathOnly
+    source: joi // same as hotReloadSyncSchema but no subPathOnly and relativeOnly
       .posixPath()
-      .relativeOnly()
       .allowGlobs()
       .default(".")
       .description(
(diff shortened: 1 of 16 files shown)


core/src/plugins/container/config.ts
```diff
@@ -253,6 +253,17 @@ export const syncDefaultGroupSchema = () =>
...
   hotReloadSyncSchema().keys({
     exclude: syncExcludeSchema(),
+    source: joi // same as hotReloadSyncSchema but no subPathOnly
+      .posixPath()
+      .relativeOnly()
+      .allowGlobs()
+      .default(".")
+      .description(
+        deline`
+      POSIX-style path of the directory to sync to the target, relative to the module's top-level directory.
```
vvagaytsev, line 263: Does it still have to be a relative path?

edvald: I think we should allow a full path actually, because absolute paths may be templated in.

Orzelius: makes sense, allowed in [0aa7cc0](https://github.com/garden-io/garden/pull/3145/commits/0aa7cc05c54854322e396d95825b5784aaf2662f)

vvagaytsev (approved):

eysi09 (approved):
LGTM.

PR #3147, opened 2022-08-25
test: partially fixed compilation errors in actions.ts

vvagaytsev:
Continuation of https://github.com/garden-io/garden/pull/3143
Still a partial fix.

Fixes #

Another iteration will happen as a new PR.

### DIFF
--- a/core/test/unit/src/actions.ts
+++ b/core/test/unit/src/actions.ts
@@ -37,6 +37,12 @@ import { testFromModule, testFromConfig } from "../../../src/types/test"
 import { ConfigGraph } from "../../../src/graph/config-graph"
 import { ResolvedRuntimeAction } from "../../../src/actions/base"
 import { ResolvedBuildAction } from "../../../src/actions/build"
+import {
+  execBuildActionSchema,
+  execDeployActionSchema,
+  execRunActionSchema,
+  execTestActionSchema,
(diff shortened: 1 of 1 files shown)


core/test/unit/src/actions.ts
```diff
@@ -70,8 +70,8 @@ describe("ActionRouter", () => {
     actions = await garden.getActionRouter()
     graph = await garden.getConfigGraph({ log: garden.log, emit: false })
     module = graph.getModule("module-a")
-    service = graph.getService("service-a")
-    task = graph.getTask("task-a")
+    resolvedDeployAction = graph.getDeploy("service-a").resolve() // todo: pass parameters
```
vvagaytsev, line 73: @edvald what is the right way to get a `ResolvedRuntimeTask` instance here? Should we run `garden.processTasks(...)` first?

cc @Orzelius @thsig

Orzelius: I usually do something like this
```ts
const rawAction = graph.getDeploy("some-deploy")
const action = await garden.resolveAction({
  action: rawAction,
  log: garden.log,
  graph: await garden.getConfigGraph({ log: garden.log, emit: false }),
})
```

vvagaytsev: Thanks!

PR #3155, opened 2022-08-29
test: fixed compilation errors in helm

vvagaytsev:

Fixes #

### DIFF
--- a/core/test/integ/src/plugins/kubernetes/helm/common.ts
+++ b/core/test/integ/src/plugins/kubernetes/helm/common.ts
@@ -11,7 +11,6 @@ import { resolve } from "path"
 import { expect } from "chai"
 import { first, uniq } from "lodash"
 import {
-  containsSource,
   getBaseModule,
   getChartPath,
   getChartResources,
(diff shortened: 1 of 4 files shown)


core/test/integ/src/plugins/kubernetes/helm/common.ts
```diff
@@ -103,29 +104,16 @@ describe("Helm common functions", () => {
...
-      expect(await containsSource(module)).to.be.false
-    })
-  })
-
   describe("renderTemplates", () => {
     it("should render and return the manifests for a local template", async () => {
-      const module = graph.getModule("api")
-      const imageModule = graph.getModule("api-image")
+      const deployAction = graph.getDeploy("api")
+      const buildImageAction = graph.getBuild("api-image")
```
Orzelius, line 110: just wondering why is this one a build?

core/test/integ/src/plugins/kubernetes/helm/common.ts
```diff
@@ -611,10 +594,10 @@ ${expectedIngressOutput}
     })
 
     it("localMode should always take precedence over devMode when add a --set flag", async () => {
-      const module = graph.getModule("api")
-      module.spec.valueFiles = []
-      const gardenValuesPath = getGardenValuesPath(module.buildPath)
-      expect(await getValueArgs(module, true, true)).to.eql([
+      const action = await garden.resolveAction({ action: graph.getDeploy("api"), log, graph })
+      action.getSpec().valueFiles = []
```
Orzelius, line 598: I don't think this line does anything. If you want to modify the spec you need to go into the `._config.spec` itself. You can get a typed version with the trick I messaged you in slack

edit: also happens in a couple places other than here as well

Orzelius (approved):
you can fix the things I mentioned in the comment if you want to I guess

core/test/integ/src/plugins/kubernetes/helm/common.ts
vvagaytsev: The module `api-image` is used for docker build only. there are no services defined in it. So, the build option looked like the best match.

core/test/integ/src/plugins/kubernetes/helm/common.ts
vvagaytsev: Yes, it does. The method `getSpec()` is a shortcut for `._config.spec`, and it returns a reference to a mutable object. So we can easily modify the state of the object over the reference.

vvagaytsev: Added some explicit typing in https://github.com/garden-io/garden/pull/3155/commits/f9dd825477a1e39c36ca15f665645a86006aa2d2

PR #3156, opened 2022-08-29
SGv2: output.ts, hadolint and kubernetes/util build issues

Orzelius:
fix output.ts, hadolint and kubernetes/util build issues

### DIFF
--- a/core/test/integ/src/plugins/kubernetes/util.ts
+++ b/core/test/integ/src/plugins/kubernetes/util.ts
@@ -11,6 +11,7 @@ import { flatten, find, first } from "lodash"
 import stripAnsi from "strip-ansi"
 import { TestGarden, expectError } from "../../../../helpers"
 import { ConfigGraph } from "../../../../../src/graph/config-graph"
+import { actionFromConfig } from "../../../../../src/graph/actions"
 import { Provider } from "../../../../../src/config/provider"
 import { DeployTask } from "../../../../../src/tasks/deploy"
 import { KubeApi } from "../../../../../src/plugins/kubernetes/api"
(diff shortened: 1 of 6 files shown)


core/test/integ/src/plugins/kubernetes/util.ts
```diff
@@ -356,49 +411,53 @@ describe("util", () => {
...
           localModeDeployNames: [],
+          fromWatch: false,
         })
 
         await helmGarden.processTasks({ tasks: [deployTask], throwOnError: true })
       })
 
       it("returns running Pod if one is found matching podSelector", async () => {
-        const module = helmGraph.getModule("api")
+        const rawAction = helmGraph.getActions({ refs: [{ kind: "Deploy", name: "api" }] })[0] as DeployAction
```
vvagaytsev, line 459: It looks like this (and the other similar places) can be simplified to `const rawAction = helmGraph.getDeploy("api")`.

core/test/integ/src/plugins/kubernetes/util.ts
```diff
@@ -85,23 +117,31 @@ describe("util", () => {
         const graph = await garden.getConfigGraph({ log: garden.log, emit: false })
         const provider = (await garden.resolveProvider(garden.log, "local-kubernetes")) as Provider<KubernetesConfig>
 
-        const service = graph.getService("simple-service")
+        const rawAction = graph.getDeploy("simple-service")
+        const action = await garden.resolveAction({
+          action: rawAction,
+          log: garden.log,
+          graph: await garden.getConfigGraph({ log: garden.log, emit: false }),
```
vvagaytsev, line 124: Can we pass the existing `graph` instance to `garden.resolveAction(...)` here? I'm not sure if we need to re-build the config graph here.

vvagaytsev (approved):
LGTM. Left a couple of comments.

vvagaytsev (approved):

PR #3160, opened 2022-08-30
test: fix integ plugins/kubernetes/run.ts build issues

Orzelius:
I think this is the last one, right?


vvagaytsev (approved):

vvagaytsev:
> I think this is the last one, right?

I hope so, but there might be a few leftovers to be fixed. Let's merge all open PRs and rerun the compilation.

PR #3162, opened 2022-08-30
Graph v2 build tests albert

Orzelius:
a few fixes more

### DIFF
--- a/core/test/unit/src/plugins/container/container.ts
+++ b/core/test/unit/src/plugins/container/container.ts
@@ -26,9 +26,7 @@ import {
 } from "../../../../../src/plugins/container/helpers"
 import { getDockerBuildFlags } from "../../../../../src/plugins/container/build"
 import { actionFromConfig } from "../../../../../src/graph/actions"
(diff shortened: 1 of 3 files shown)


core/test/unit/src/plugins/container/container.ts
```diff
@@ -99,7 +97,7 @@ describe("plugins.container", () => {
       garden,
       log,
       config: parsed.moduleConfig,
```
vvagaytsev, line 99: Should this also be `config: parsed.moduleconfig.buildConfig`? If yes, then we'll have 2 identical methods here and in `helpers.ts`.

vvagaytsev (approved):
LGTM, please check one comment that I left.

core/test/unit/src/plugins/container/container.ts
Orzelius: probably yes, but because this `configure` here has no types and can't say for sure. I think we'll just have to come back to this once we can run the tests

PR #3163, opened 2022-08-30
fix(core): fixed concurrency issue with plugin tool downloads

edvald:
Fixes issues like

```
ENOTEMPTY: directory not empty, rename 
'/github/home/.garden/tools/kubectl/49eb930aa565a80f.1480f616' -> 
'/github/home/.garden/tools/kubectl/49eb930aa565a80f'
```

when fetching plugin tools automatically.


stefreak (approved):
good catch 👍

PR #3175, opened 2022-09-02
 improvement: upgraded gradle version to 7.5.1

vvagaytsev:

Fixes #

### DIFF
--- a/plugins/jib/gradle.ts
+++ b/plugins/jib/gradle.ts
@@ -15,12 +15,14 @@ import { resolve } from "path"
 import { pathExists } from "fs-extra"
 import { Writable } from "stream"
 
+const gradleVersion = "7.1.1"
+
 const spec = {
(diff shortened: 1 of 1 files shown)


plugins/jib/gradle.ts
```diff
@@ -15,12 +15,14 @@ import { resolve } from "path"
 import { pathExists } from "fs-extra"
 import { Writable } from "stream"
 
+const gradleVersion = "7.1.1"
```
Walther, line 18: typofix: need to update the version const to match the sha256sum

vvagaytsev: Good catch, thanks! Fixed it in https://github.com/garden-io/garden/pull/3175/commits/00258199f8d552eb20a1d91857265667f5193fdc

Walther (approved):

PR #3181, opened 2022-09-05
test fixes sgv2

Orzelius:
please see individual commits for more details

### DIFF
--- a/core/src/config/base.ts
+++ b/core/src/config/base.ts
@@ -69,7 +69,7 @@ export async function loadConfigResources(
   configPath: string,
   allowInvalid = false
 ): Promise<GardenResource[]> {
-  let fileData: Buffer = await readConfigFile(configPath, projectRoot)
+  const fileData = await readConfigFile(configPath, projectRoot)
 
   const resources = await validateRawConfig(fileData.toString(), configPath, projectRoot, allowInvalid)
 
--- a/core/src/config/base.ts
+++ b/core/src/config/base.ts
@@ -71,17 +71,22 @@ export async function loadConfigResources(
 ): Promise<GardenResource[]> {
   const fileData = await readConfigFile(configPath, projectRoot)
 
-  const resources = await validateRawConfig(fileData.toString(), configPath, projectRoot, allowInvalid)
+  const resources = await validateRawConfig({ rawConfig: fileData.toString(), configPath, projectRoot, allowInvalid })
 
   return resources
 }
 
-export async function validateRawConfig(
-  rawConfig: string,
-  configPath: string,
-  projectRoot: string,
-  allowInvalid = false
-) {
+export async function validateRawConfig({
+  rawConfig,
+  configPath,
+  projectRoot,
+  allowInvalid = false,
+}: {
+  rawConfig: string
+  configPath: string
+  projectRoot: string
+  allowInvalid?: boolean
+}) {
   let rawSpecs = await loadAndValidateYaml(rawConfig, configPath)
 
   // Ignore empty resources
--- a/core/src/config/base.ts
+++ b/core/src/config/base.ts
@@ -69,15 +69,20 @@ export async function loadConfigResources(
   configPath: string,
   allowInvalid = false
 ): Promise<GardenResource[]> {
-  let fileData: Buffer
+  let fileData: Buffer = await readConfigFile(configPath, projectRoot)
 
-  try {
-    fileData = await readFile(configPath)
-  } catch (err) {
-    throw new FilesystemError(`Could not find configuration file at ${configPath}`, { projectRoot, configPath })
-  }
+  const resources = await validateRawConfig(fileData.toString(), configPath, projectRoot, allowInvalid)
 
-  let rawSpecs = await loadAndValidateYaml(fileData.toString(), configPath)
+  return resources
+}
+
+export async function validateRawConfig(
+  rawConfig: string,
+  configPath: string,
+  projectRoot: string,
+  allowInvalid = false
+) {
+  let rawSpecs = await loadAndValidateYaml(rawConfig, configPath)
 
   // Ignore empty resources
   rawSpecs = rawSpecs.filter(Boolean)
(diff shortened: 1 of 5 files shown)


core/test/helpers.ts
```diff
@@ -706,3 +719,16 @@ export async function assertAsyncError(action: () => Promise<any>, ...expectedEr
     expectedErrorMessages.forEach((msg) => expect(errorString).to.contain(msg))
   }
 }
+
+// https://stackoverflow.com/a/66836940
+// useful for typesafe stubbing
+export function getPropertyName<T>(
```
edvald, line 725: Interesting. Can you elaborate a bit on why/how this is useful? Any chance a wrapper around `td.replace` might be easier to grok?

core/src/config/base.ts
```diff
@@ -69,26 +69,41 @@ export async function loadConfigResources(
   configPath: string,
   allowInvalid = false
 ): Promise<GardenResource[]> {
-  let fileData: Buffer
+  let fileData: Buffer = await readConfigFile(configPath, projectRoot)
```
edvald, line 72:
```suggestion
  const fileData = await readConfigFile(configPath, projectRoot)
```

core/src/config/base.ts
```diff
@@ -69,26 +69,41 @@ export async function loadConfigResources(
...
 
-  let rawSpecs = await loadAndValidateYaml(fileData.toString(), configPath)
+  return resources
+}
+
+export async function validateRawConfig(
+  rawConfig: string,
+  configPath: string,
+  projectRoot: string,
+  allowInvalid = false
```
edvald, line 83: Suggest converting these params to destructured object. I use that refactoring shortcut in VSCode all the time, and imo it's generally advisable when there are more than 2 params, especially when they share a type that can get confused without the compiler noticing.

core/src/config/base.ts
```diff
@@ -69,26 +69,41 @@ export async function loadConfigResources(
...
   )
-
   return resources
 }
 
+export async function readConfigFile(configPath: string, projectRoot: string) {
+  let fileData: Buffer
+
+  try {
+    fileData = await readFile(configPath)
```
vvagaytsev, line 100: This can be a bit simplified to `return await readFile(configPath)`. It will make the local variable `fileData` unnecessary.

core/test/unit/src/config/base.ts
```diff
@@ -159,9 +161,16 @@ describe("loadConfigResources", () => {
...
   it("throws if basic fields contain template strings", async () => {
     for (const field of noTemplateFields) {
+      const basicProjecConfig = { ...defaultProjectConfig }
+      basicProjecConfig[field] = '${camelCase("No templating should be allowed here")}'
+      const configRaw = safeDumpYaml(basicProjecConfig)
+      await expectError(
+        async () => validateRawConfig(configRaw, "fake/path", "fake/projec/root"),
+        (err) => {
+          expectFuzzyMatch(err, "does not allow templating")
+        }
```
vvagaytsev, line 171: Now this callback can be replaced with `{ contains: "does not allow templating" }`.

core/test/unit/src/config/base.ts
```diff
@@ -159,9 +161,16 @@ describe("loadConfigResources", () => {
 
   it("throws if basic fields contain template strings", async () => {
     for (const field of noTemplateFields) {
+      const basicProjecConfig = { ...defaultProjectConfig }
```
vvagaytsev, line 164: This creates a shallow copy of the object. This is fine as long as we do the full field reassignment by the object key, like `obj["someKey"] = someValue`. But, if we need to modify any "composite" field (i.e. with a value of type array or object), then the shallow copy won't prevent the pollution of the original template object `defaultProjectConfig`.

Maybe composite field modification is not a frequent use case in our tests, but it might make sense to prevent that at all. Otherwise, we might face some weird errors in the tests at some point in the future.

We can do something like this:
```
export const defaultProjectConfigTemplate: ProjectConfig = {...}
export const defaultProjectConfig = (): ProjectConfig => cloneDeep(defaultProjectConfigTemplate)
```

The same pattern can be applied to all project/module/action configs in the tests.

core/test/helpers.ts
Orzelius: Can be used like [here](https://github.com/garden-io/garden/pull/3181/commits/e76a91251aec5fedbd258863343b0a84650f31aa#diff-662cac65dad0f8b07d88326c30ed2f73a46411c23f05bc66d80d745f8787635fL78-L125) to type stubs so when changing method names tests don't break. But a wrapper does seem like a simpler approach and I'll look into implementing it

core/src/config/base.ts
Orzelius: That's a nice helper, didn't know it existed. Thanks for pointing it out [8325284](https://github.com/garden-io/garden/pull/3181/commits/8325284486adb6e866326ca76d1b40a14b3b1f98)

core/src/config/base.ts
Orzelius: [03a6325](https://github.com/garden-io/garden/pull/3181/commits/03a6325f24ad029d8cb1eb075da7824a7ea4b4a6) It's soo clean now :blue_book:

core/test/unit/src/config/base.ts
Orzelius: :+1: [f82ee17](https://github.com/garden-io/garden/pull/3181/commits/f82ee17a41ce889606b9f623d7677a9a932a09dd)

Orzelius:
Both of you, thanks for the thorough code review!

core/test/helpers.ts
Orzelius: Couldn't come up with a good wrapper, instead added a primitive type which should make things easier to understand [c6ea83c](https://github.com/garden-io/garden/pull/3181/commits/c6ea83c3e837e9fdce4c457a37d6fb1992ce10cc)

PR #3183, opened 2022-09-05
graph-v2: fix router typing

edvald:


core/src/config/base.ts
```diff
@@ -194,6 +196,22 @@ export function prepareProjectResource(spec: any): ProjectResource {
   )
 }
 
+export function prepareActionResource(spec: any, configPath: string, relPath: string) {
+  delete spec.path
+
+  if (spec.internal) {
+    throw new ConfigurationError(`yFound invalid ke "internal" in config at ${relPath}`, {
```
Orzelius, line 203: A typo I presume

edvald: Indeed

core/src/actions/helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+    return new ResolvedRunAction({ ...action["params"], ...params })
+  } else if (isTestAction(action)) {
+    return new ResolvedTestAction({ ...action["params"], ...params })
+  } else {
+    // This should never happen
+    throw new InternalError(`Unexpected action kind`, {})
+  }
+}
+
+export function executeAction<T extends ResolvedAction>(
```
thsig, line 31: Not sure it's worth mentioning, but if this function ends up being called very frequently, it would be more efficient to use a lookup table rather than running through these conditionals every time.

core/src/graph/config-graph.ts
```diff
@@ -427,6 +427,7 @@ export abstract class BaseConfigGraph<
   }
 
   protected addActionInternal(action: Action) {
+    this.actions[action.kind][action.name] = <any>action
```
vvagaytsev, line 430: The existing conditional type can be applied instead of casting to`<any>`:
```suggestion
  protected addActionInternal<K extends ActionKind>(action: Action) {
    this.actions[action.kind][action.name] = <PickTypeByKind<K, B, D, R, T>>action
```

core/src/actions/base.ts
```diff
@@ -525,23 +523,14 @@ export abstract class RuntimeAction<
...
-  resolve(params: ResolveActionParams<C>): Resolved<this> {
-    // TODO-G2: validate static outputs here
-    const constructor: ResolvedActionConstructor<this> = Object.getPrototypeOf(this).constructor
-    return constructor({ ...this.params, ...params })
-  }
 }
 
 export interface ResolvedActionConstructor<A extends BaseAction> {
-  (params: ResolvedActionWrapperParams<A["_config"]>): Resolved<A>
+  new (params: ResolvedActionWrapperParams<A["_config"]>): Resolved<A>
```
vvagaytsev, line 529: It looks like this interface and `ExecutedActionConstructor` in L532 are no longer used. Should we remove both?

core/src/tasks/build.ts
```diff
@@ -75,7 +76,9 @@ export class BuildTask extends ExecuteActionTask<BuildAction, BuildStatus> {
         append: true,
       })
 
-      return { ...result, executedAction: action.execute({ status: result }) }
+      console.log(result)
```
vvagaytsev, line 79: `console.log(...)` will be rejected by the linter. Let's remove this or use the logger instead.

core/src/actions/base.ts
edvald: Yeah I can remove both. We had these `resolve()` methods that I couldn't quite get working as I'd liked.

core/src/actions/helpers.ts
edvald: I'm not sure that's the case, but either way the reason for going this route is to use the type guards. I had issues with type checking otherwise.

core/src/tasks/build.ts
edvald: I just forgot it in there, will fix.

vvagaytsev (approved):

Issue #3166, opened 2022-08-31
[FEATURE]: Use native tools on ARM64 mac


stefreak:
## Feature Request

Garden should download and use ARM64 native tools on M1 mac.

### Background / Motivation

I am hitting a terraform problem in conjunction with rosetta, where terraform apply fails randomly on M1 macs. [This can be fixed by installing ARM64 native terraform](https://github.com/hashicorp/terraform-provider-aws/issues/22582#issuecomment-1015302978). But with garden, I do not have control to do that.

Because of that I get errors like this, making terraform completely unusable:
```
Error: Unrecognized remote plugin message:

This usually means that the plugin is either invalid or simply
needs to be recompiled to support the latest protocol.


Command "/Users/steffen/.garden/tools/terraform-0-14-7/5bca2f75c0636f7a/terraform apply -var-file /Users/steffen/work/github/garden-platform/cloud/customers/terraform/garden.tfvars.json" failed with code 1:
```

There might be similar rosetta incompatibilities with other tools.

### What should the user be able to do?

When on M1 mac, garden should simply use the appropriate ARM64 tools where available.

### Why do they want to do this? What problem does it solve?

For example, terraform does not work with rosetta, but it works fine natively. There might be other tools with issues like that.

### Suggested Implementation(s)

I am not sure. This might be related to https://github.com/garden-io/garden/issues/2333

### How important is this feature for you/your team?

🌵 Not having this feature makes using Garden painful

... it's very, very painful. I wrote a bash script that runs garden in a docker container to work around this issue, and it's very slow, and it was not easy to make it work in the first place.

vvagaytsev:
Thanks for raising this @stefreak! Looks related to https://github.com/garden-io/garden/issues/2333 We can reconsider both issues after the Garden `0.13` release. I also use an Apple machine with an M1 chip :)

PR #3185, opened 2022-09-06
feat: use m1 native tools where available

stefreak:

Use M1 native tools where upstream provides official builds.

The idea of this PR is to already download and call native M1 tools where available.

We should actually change our build processes to provide native ARM Garden builds for Linux and MacOS, but this change is simpler and already relieves some pain for tools with ARM incompatibilities (like with terraform, see #3166).

Also in case of Java, users can benefit from faster builds with native execution.

Implements #3166

I gave my best to make sure that all the hashes and instances of the string `arm64`/`aarch64` are correct. I smoke tested using this command:

```
garden tools 2>&1 | grep -E '\[(binary|library)\]' | awk -F' ' '{print $1}' | xargs -I'{}' -n 1 -P 20 garden tools '{}' -- version
```

all the downloads succeeded on my M1 mac.

If you want, I can add unit tests for this stuff, but it would take some time

### DIFF
--- a/core/src/plugins/kubernetes/helm/helm-cli.ts
+++ b/core/src/plugins/kubernetes/helm/helm-cli.ts
@@ -32,7 +32,7 @@ export const helm3Spec: PluginToolSpec = {
     },
     {
       platform: "darwin",
-      architecture: "amd64",
+      architecture: "arm64",
       url: "https://get.helm.sh/helm-v3.7.2-darwin-arm64.tar.gz",
       sha256: "260d4b8bffcebc6562ea344dfe88efe252cf9511dd6da3cccebf783773d42aec",
       extract: {
--- a/core/src/plugins/kubernetes/helm/helm-cli.ts
+++ b/core/src/plugins/kubernetes/helm/helm-cli.ts
@@ -31,6 +31,16 @@ export const helm3Spec: PluginToolSpec = {
       },
     },
     {
+      platform: "darwin",
+      architecture: "amd64",
+      url: "https://get.helm.sh/helm-v3.7.2-darwin-arm64.tar.gz",
+      sha256: "260d4b8bffcebc6562ea344dfe88efe252cf9511dd6da3cccebf783773d42aec",
+      extract: {
+        format: "tar",
+        targetPath: "darwin-arm64/helm",
+      },
+    },
+    {
       platform: "linux",
       architecture: "amd64",
       url: "https://get.helm.sh/helm-v3.7.2-linux-amd64.tar.gz",
--- a/core/src/util/ext-tools.ts
+++ b/core/src/util/ext-tools.ts
@@ -250,8 +250,6 @@ export class PluginTool extends CliWrapper {
     const architecture = getArchitecture()
     const isEmulated = isRosetta()
 
-    console.log("===================", { _platform, architecture, isEmulated })
-
     let buildSpec: ToolBuildSpec
     // first look for native arch, if not found, then try (potentially emulated) arch
     if (isEmulated) {
--- a/core/src/util/ext-tools.ts
+++ b/core/src/util/ext-tools.ts
@@ -6,7 +6,6 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import { platform } from "os"
 import split2 from "split2"
 import { pathExists, createWriteStream, ensureDir, chmod, remove, move, createReadStream } from "fs-extra"
 import { ConfigurationError, ParameterError, GardenBaseError, RuntimeError } from "../exceptions"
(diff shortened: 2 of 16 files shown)


core/src/plugins/kubernetes/helm/helm-cli.ts
```diff
@@ -30,6 +30,16 @@ export const helm3Spec: PluginToolSpec = {
         targetPath: "darwin-amd64/helm",
       },
     },
+    {
+      platform: "darwin",
+      architecture: "amd64",
```
stefreak, line 35:
```suggestion
      architecture: "arm64",
```

plugins/jib/openjdk.ts
```diff
@@ -103,21 +108,37 @@ const jdk17Version: JdkVersion = {
...
+        targetPath: posix.join(jdkVersion.versionName, "Contents", "Home"),
+      },
+    },
+  ]
+
+  if (jdkVersion.mac_arm64) {
+    macBuilds.push({
+      platform: "darwin",
+      architecture: "arm64",
+      url: jdkVersion.baseUrl + jdkVersion.mac_amd64.filename,
```
stefreak, line 128:
```suggestion
      url: jdkVersion.baseUrl + jdkVersion.mac_arm64.filename,
```

core/src/util/util.ts
```diff
@@ -703,6 +704,21 @@ export function getArchitecture() {
   return archMap[arch] || arch
 }
 
+export function isRosetta() {
```
edvald, line 707: Could we memoize this?

stefreak: done, good point

core/src/util/ext-tools.ts
```diff
@@ -229,17 +245,28 @@ export class PluginTool extends CliWrapper {
...
+    const _platform = getPlatform()
+    const architecture = getArchitecture()
+    const isEmulated = isRosetta()
+
+    let buildSpec: ToolBuildSpec
+    // first look for native arch, if not found, then try (potentially emulated) arch
+    if (isEmulated) {
+      buildSpec = findBuildSpec(spec, _platform, "arm64") || findBuildSpec(spec, _platform, architecture)
+    } else {
+      buildSpec = findBuildSpec(spec, _platform, architecture)
```
stefreak, line 258: When garden will go fully native on arm64, this code will fail as some tools are not available in arm64.

We should already prepare that and prefer arm64, but fall back to rosetta.

do you agree @Orzelius @edvald?

stefreak: Implemented that in [945e300](https://github.com/garden-io/garden/pull/3185/commits/945e3005c3571fbfc0108e755074a978e572897b)

core/src/util/ext-tools.ts
```diff
@@ -202,9 +201,8 @@ export class CliWrapper {
   }
 }
 
-interface PluginToolOpts {
-  platform?: string
-  architecture?: string
+const findBuildSpec = (spec: PluginToolSpec, plat: string, arch: string) => {
+  return spec.builds.find((build) => build.platform === plat && build.architecture === arch)!
```
Orzelius, line 205: I don't think using `!` here is a good idea, because due to our error we might have a case where nothing is found. Instead maybe just throw a helpful error if the result is undefined.

Orzelius:
You can ignore the flaky e2e tests, there's something wrong with the inc-cluster registry and we've had no time to debug it. Lint unhappiness should be fixed though.

core/src/util/ext-tools.ts
stefreak: it's actually being checked in line 244. that check works, I tested it - there is a nice user friendly error message.

stefreak: I can eliminate the `!` here, but then I can't assign directly to `this.buildspec` in line  239 and I'd need to introduce another variable. Would you prefer that?

stefreak: I moved it to lines 239 and 241, then it might be more clear. do you agree?

core/src/util/ext-tools.ts
```diff
@@ -236,9 +236,9 @@ export class PluginTool extends CliWrapper {
 
     if (darwinARM) {
       // first look for native arch, if not found, then try (potentially emulated) arch
-      this.buildSpec = findBuildSpec(spec, platform, "arm64") || findBuildSpec(spec, platform, "amd64")
+      this.buildSpec = findBuildSpec(spec, platform, "arm64")! || findBuildSpec(spec, platform, "amd64")!
```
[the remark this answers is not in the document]
stefreak: Thanks, @Orzelius, now the code looks much better, you're right. Do you like it now?

Orzelius (approved):
happy

PR #3191, opened 2022-09-06
Fixed error assertions in `config-graph.ts` unit tests

vvagaytsev:

Fixes #


core/test/unit/src/config-graph.ts
```diff
@@ -34,7 +41,8 @@ describe("ConfigGraph", () => {
     await expectError(
       () => garden.getConfigGraph({ log: garden.log, emit: false }),
       (err) =>
-        expect(err.message).to.equal(
+        expectErrorMessageContains(
```
edvald, line 44: I have a suggestion: We can allow a string as the second argument to `expectError` which would basically do exactly this. Same result, just a little cleaner :)

vvagaytsev: @edvald `expectError` already accepts an optional 2nd argument of type `string | function`. I created a new helper function that delegates to `expectError` in https://github.com/garden-io/garden/pull/3191/commits/6d5a529730bbde82c9a2adb060eeea16a706a2e6.

core/src/graph/config-graph.ts
```diff
@@ -521,7 +521,7 @@ export class ConfigGraphNode {
     return {
       name: this.name,
       kind: this.type,
-      key: this.name,
+      key: nodeKey(this.type, this.name),
```
vvagaytsev, line 524: @edvald should we include the `ActionKind` into the node's key?

vvagaytsev:
@edvald now the only failing test in `config-graph.ts` is https://github.com/garden-io/garden/blob/graph-v2-fix-config-graph-tests/core/test/unit/src/config-graph.ts#L469

Is this a bug or a feature? I assume it's a bug, so we still need to include disabled modules in the build dependencies, right?

core/src/graph/config-graph.ts
edvald, line 524: Hmm, this doesn't smell right... I think `this.type` should be renamed to `this.kind`. And `key` should really just be called `name`, otherwise we're repeating the _kind_ in the key.

vvagaytsev: This discussion is continued in https://github.com/garden-io/garden/pull/3191#discussion_r964759175.

vvagaytsev: Fixed this in:
* https://github.com/garden-io/garden/pull/3191/commits/c45584a6b623b2f5e28a95042e15203341707d42
* https://github.com/garden-io/garden/pull/3191/commits/ebcdd0d5c453f624f64dcc7bd6248f12b7a564b9
* https://github.com/garden-io/garden/pull/3191/commits/fe58c6495b3373313b4811071da655a33eca53fe

core/test/unit/src/config-graph.ts
vvagaytsev: Refactored in https://github.com/garden-io/garden/pull/3191/commits/af2faf148f78b938b9401b9817447fbb9a194118

PR #3192, opened 2022-09-07
fix: empty action group when converting basic exec module

edvald:
cc @vvagaytsev


vvagaytsev (approved):
Thank you! Looks good 👍

PR #3195, opened 2022-09-07
WIP: try to fix getPublicImageId helper tests

Orzelius:
WIP

Attempt to fix tests for the docker getPublicImageId helper. Instead of a module, create an action dynamically. Current issue: `garden.resolveAction` with the dynamic action passed returns undefined.

### DIFF
--- a/core/src/config/base.ts
+++ b/core/src/config/base.ts
@@ -69,15 +69,20 @@ export async function loadConfigResources(
   configPath: string,
   allowInvalid = false
 ): Promise<GardenResource[]> {
-  let fileData: Buffer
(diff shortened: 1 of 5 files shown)


edvald:
There's an error here (introduced by myself, I should note): https://github.com/garden-io/garden/blob/46a8f05bd170b21fa662611567cfa3cf1cdd1840/core/src/graph/actions.ts#L207

The `GraphResults` class should not be indexed like that. `results.results.getResult(...)` should be used instead.

I wonder if it's possible to make the type-checking more strict to catch this... I've looked for this but there might still be more places where this error lingers.

Orzelius:
closing in favor of #3181

PR #3206, opened 2022-09-09
feat: add option to run git pulls in parallel

ryanhair:

This adds the ability for git pulls to happen in parallel. For projects with large amounts of repositories, it can start taking a _long_ time to update all repos.

### DIFF
--- a/core/src/commands/update-remote/all.ts
+++ b/core/src/commands/update-remote/all.ts
@@ -14,15 +14,27 @@ import { updateRemoteModules } from "./modules"
 import { SourceConfig, projectSourceSchema, moduleSourceSchema } from "../../config/project"
 import { printHeader } from "../../logger/util"
 import { joi, joiArray } from "../../config/common"
+import { BooleanParameter } from "../../cli/params"
 
 export interface UpdateRemoteAllResult {
   projectSources: SourceConfig[]
   moduleSources: SourceConfig[]
 }
 
-export class UpdateRemoteAllCommand extends Command {
+const updateRemoteAllOptions = {
+  parallel: new BooleanParameter({
+    help: "Allow git updates to happen in parallel",
+    defaultValue: false,
+  }),
+}
+
+type Opts = typeof updateRemoteAllOptions
+
+export class UpdateRemoteAllCommand extends Command<{}, Opts> {
   name = "all"
   help = "Update all remote sources and modules."
+
+  options = updateRemoteAllOptions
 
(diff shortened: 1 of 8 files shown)


edvald:
Thanks so much @ryanhair 👍! 

I'll leave my team to review in more detail but one thing I might flag is that we should consider what happens when a username/password prompt comes up in one or more of the updates. That scenario was the reason this wasn't done to begin with.

I think it might be best to do something like this: https://stackoverflow.com/questions/64319349/force-git-to-fail-when-password-required-instead-of-prompting-for-password
And then document this implication in the help string for the `--parallel` flag.

ryanhair:
@edvald I made some updates, totally agree that we should force Git to fail with a warning in the help text

Orzelius (approved):
@ryanhair Thank you so much for this. It's a very nice implementation and makes sense if there's a lot to fetch. The only potential problem I see is git providers looking at this as a ddos attack if 10 requests are made in the same time, but we can deal with that issue if we ever run into it.

Orzelius:
also on another note, how was the developer experience? Any pain points getting up and running? We know the contribution guide is a bit outdated but is there something else we could improve about the process?

ryanhair:
> @ryanhair Thank you so much for this. It's a very nice implementation and makes sense if there's a lot to fetch. The only potential problem I see is git providers looking at this as a ddos attack if 10 requests are made in the same time, but we can deal with that issue if we ever run into it.

Yeah, I thought about batching the promises, but I wasn't sure we wanted to add that yet. You're right, it'll depend on the provider, for ours, I nuked my `.garden` cache and ran it to see what would happen with our 96 (yeah, ridiculous amount) repos that get cloned, and it succeeded, but that's just our case...

ryanhair:
> also on another note, how was the developer experience? Any pain points getting up and running? We know the contribution guide is a bit outdated but is there something else we could improve about the process?

To be honest, I don't think I figured out the right flow. I tried `yarn dev`, but had problems (which I honestly can't recall off the top of my head). Instead, I ended up running from the `core` folder a script that would move up to the parent folder, run a build, then move back into core and run whatever it was I wanted. Not elegant, not really fast, but it worked for what I needed for this

ryanhair:
One last question @Orzelius, what do I need to do to get this merged?

Orzelius:
> One last question @Orzelius, what do I need to do to get this merged?

Nothing :clinking_glasses:
Thank you very much!

PR #3218, opened 2022-09-13
fix(core): fix Rosetta detection logic for OSX

thsig:

Fixes an issue where Garden would crash on Intel macs. This is a patch for https://github.com/garden-io/garden/pull/3185.

Closes https://github.com/garden-io/garden/issues/3166

### DIFF
--- a/core/src/util/util.ts
+++ b/core/src/util/util.ts
@@ -717,10 +717,12 @@ export const isDarwinARM = memoize(() => {
     // see also https://developer.apple.com/documentation/apple-silicon/about-the-rosetta-translation-environment
     // We use execSync here, because this function is called in a constructor
     // otherwise we'd make the function async and call `spawn`
-    const stdout = execSync("sysctl -n sysctl.proc_translated", { encoding: "utf-8" })
-    if (stdout === "1\n") {
-      return true
+    try {
+      execSync("sysctl -n -q sysctl.proc_translated", { encoding: "utf-8" })
+    } catch (err) {
+      return false
     }
+    return true
   }
 
   return false


core/src/util/util.ts
```diff
@@ -717,10 +717,12 @@ export const isDarwinARM = memoize(() => {
     // see also https://developer.apple.com/documentation/apple-silicon/about-the-rosetta-translation-environment
     // We use execSync here, because this function is called in a constructor
     // otherwise we'd make the function async and call `spawn`
-    const stdout = execSync("sysctl -n sysctl.proc_translated", { encoding: "utf-8" })
-    if (stdout === "1\n") {
-      return true
+    try {
+      execSync("sysctl -n -q sysctl.proc_translated", { encoding: "utf-8" })
```
thsig, line 721: I added the `-q` (quiet) option here, since without it I always got a log line saying `sysctl: unknown oid 'sysctl.proc_translated'`.

This way, we just rely on the exit code.

stefreak: Oh, should have tested this one! Sorry for that.

stefreak (approved):
👍 LGTM, sorry for introducing this regression

core/src/util/util.ts
vvagaytsev: Just to confirm. we don't have to check the output returned by the `sysctl -n -q sysctl.proc_translated` command, right?

vvagaytsev: I think it should be ok. Anyway, we tested it on both ARM and Intel machines.

vvagaytsev (approved):

core/src/util/util.ts
stefreak: @vvagaytsev yes, apparently the command will fail on non-ARM macs. I assumed in the original change, that the sysctl will work on non-ARM macs as well, always returning 0.

PR #3222, opened 2022-09-13
feat: allow module overlap if one or both of them are disabled

stefreak:
This fixes module validation to allow for modules to overlap, if
one of them is disabled. Example use case: one environment deploys
to kubernetes, the other environment deploys to a serverless env.

This also changes the error message: The goal was to make clear that
there are two ways to get rid of the error if the overlap is intentional.

Fixes #

- [x] TODO: I failed running the tests locally, https://github.com/garden-io/garden/blob/master/CONTRIBUTING.md#tests seems to be outdated: `error Command "test" not found. Did you mean "jest"?` – how can I run the fs.ts tests locally?

### DIFF
--- a/core/src/garden.ts
+++ b/core/src/garden.ts
@@ -1132,7 +1132,7 @@ export class Garden {
       })
       .join("\n\n")
     const message = chalk.red(dedent`
-      There are multiple enabled modules that share the same garden.yml file or are nested within another:
+      Found multiple enabled modules that share the same garden.yml file or are nested within another:
 
       ${overlapList}
 
--- a/core/src/garden.ts
+++ b/core/src/garden.ts
@@ -1132,11 +1132,18 @@ export class Garden {
       })
       .join("\n\n")
     const message = chalk.red(dedent`
-      Missing ${chalk.bold("include")} and/or ${chalk.bold("exclude")} directives on modules with overlapping paths.
-      Setting includes/excludes is required when modules have the same path (i.e. are in the same garden.yml file),
-      or when one module is nested within another.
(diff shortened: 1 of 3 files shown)


core/test/unit/src/util/fs.ts
```diff
@@ -155,6 +167,19 @@ describe("detectModuleOverlap", () => {
       expect(detectModuleOverlap({ projectRoot, gardenDirPath, moduleConfigs: [moduleA, moduleB] })).to.be.empty
     })
 
+    it("should ignore modules that set excludes", () => {
```
stefreak, line 170:
```suggestion
    it("should ignore modules that are disabled", () => {
```

core/src/garden.ts
```diff
@@ -1132,11 +1132,18 @@ export class Garden {
       })
       .join("\n\n")
     const message = chalk.red(dedent`
-      Missing ${chalk.bold("include")} and/or ${chalk.bold("exclude")} directives on modules with overlapping paths.
-      Setting includes/excludes is required when modules have the same path (i.e. are in the same garden.yml file),
-      or when one module is nested within another.
+      There are multiple enabled modules that share the same garden.yml file or are nested within another:
```
stefreak, line 1135:
```suggestion
      Found multiple enabled modules that share the same garden.yml file or are nested within another:
```

vvagaytsev:
@stefreak to run the tests you need to locate the `core` directory first:
```
cd core
yarn test
```

I tried running `yarn test` from the repo root and got the same error. But running it from `core` went well.

vvagaytsev (approved):
LGTM, this could be squashed and merged.

PR #3223, opened 2022-09-13
fix: make copy target actually default to source

stefreak:
The copy target did not actually default to source, as
stated in the documentation. This commit changes that. Now instead of writing

```
 build:
   dependencies:
     - name: other
       copy:
         - source: "src"
           target: "src"
```
 
you can actually write

```
 build:
   dependencies:
     - name: other
       copy:
         - source: "src"
```

without a confusing error message:

```
Error validating Module 'module' (module): key .build.dependencies[0][copy][0][target] is not allowed to be empty
```

Fixes #

- [ ] TODO: add a unit test. Where is a good location for this? Do we have tests for the validation?

### DIFF
--- a/core/src/config/module.ts
+++ b/core/src/config/module.ts
@@ -43,9 +43,12 @@ const copySchema = () =>
       .subPathOnly()
       .required()
       .description("POSIX-style path or filename of the directory or file(s) to copy to the target."),
-    target: joi.posixPath().subPathOnly().default("").description(dedent`
+    target: joi
+      .posixPath()
+      .subPathOnly()
+      .default((parent) => parent.source).description(dedent`
         POSIX-style path or filename to copy the directory or file(s), relative to the build directory.
-        Defaults to to same as source path.
+        Defaults to the same as source path.
       `),
   })
 
--- a/docs/reference/commands.md
+++ b/docs/reference/commands.md
@@ -1254,7 +1254,7 @@ providers:
                   source:
 
                   # POSIX-style path or filename to copy the directory or file(s), relative to the build directory.
-                  # Defaults to to same as source path.
+                  # Defaults to the same as source path.
                   target:
 
           # Maximum time in seconds to wait for build to finish.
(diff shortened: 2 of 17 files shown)


core/src/config/module.ts
```diff
@@ -43,7 +43,10 @@ const copySchema = () =>
       .subPathOnly()
       .required()
       .description("POSIX-style path or filename of the directory or file(s) to copy to the target."),
-    target: joi.posixPath().subPathOnly().default("").description(dedent`
+    target: joi
+      .posixPath()
+      .subPathOnly()
+      .default((parent) => parent.source).description(dedent`
         POSIX-style path or filename to copy the directory or file(s), relative to the build directory.
         Defaults to to same as source path.
```
vvagaytsev, line 51: Double "to"? This likely should be "Defaults to the same as the source path."

stefreak:
```suggestion
        Defaults to the same as source path.
```

vvagaytsev:
@stefreak please rerun the `yarn build` command to refresh the auto-generated documentation, and commit those changes as well.

I also left [one comment](https://github.com/garden-io/garden/pull/3223/files#r970419644) that is not relevant to the changes made in this PR. Looks like we have had a typo in the docs, let's fix it here as well.

core/src/config/module.ts
stefreak: Actually this needs fixing in a couple of places: https://github.com/garden-io/garden/search?q=%22to+to+same%22

stefreak: Good catch @vvagaytsev 👍

vvagaytsev: @stefreak no need to fix those `*.md` files, those are automatically generated. All you need is to rerun `yarn build` command. It will re-generate those docs.

vvagaytsev: So, every time we deal with any changes in the parameters schema description, we should remember to regenerate the docs.

vvagaytsev:
LGTM, let's use "Squash and merge" for it. The PR's header would be a good commit message :)

core/src/config/module.ts
stefreak: thank you vova

PR #3232, opened 2022-09-19
fix: ensure semver format in git/rsync version regex

vvagaytsev:
This PR fixes the regex patterns for `git` and `rsync` version checks. The previous pattern was not accurate enough. Fow Windows output it grabbed the whole version like `2.37.0.windows.1` instead of semver-only sub-part like `2.37.0`.

Now it's fixed. The relevant error message was not informative enough. The error handling will be fixed in a separate PR.

This PR also ensures that `git` and `rsync` are installed and up-to-date in `test-macos` and `test-windows` steps of the CircleCI pipeline.

Fixes #

### DIFF
--- a/core/src/build-staging/rsync.ts
+++ b/core/src/build-staging/rsync.ts
@@ -14,7 +14,7 @@ import { BuildStaging, SyncParams } from "./build-staging"
 import { validateInstall } from "../util/validateInstall"
 
 const minRsyncVersion = "3.1.0"
-const versionRegex = /rsync  version [v]*([\d\.]+)  /
+const versionRegex = /rsync\s+version\s+v?(\d+.\d+.\d+)/
 
 /**
  * throws if no rsync is installed or version is too old
--- a/core/src/vcs/vcs.ts
+++ b/core/src/vcs/vcs.ts
@@ -38,7 +38,7 @@ export const NEW_MODULE_VERSION = "0000000000"
 const fileCountWarningThreshold = 10000
 
 const minGitVersion = "2.14.0"
-export const gitVersionRegex = /git version [v]*([\d\.]+)/
+export const gitVersionRegex = /git\s+version\s+v?(\d+.\d+.\d+)/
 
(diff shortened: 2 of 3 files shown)


core/src/build-staging/rsync.ts
```diff
@@ -14,7 +14,7 @@ import { BuildStaging, SyncParams } from "./build-staging"
 import { validateInstall } from "../util/validateInstall"
 
 const minRsyncVersion = "3.1.0"
-const versionRegex = /rsync  version [v]*([\d\.]+)  /
```
stefreak, line 17: can `v` really repeat multiple times? looks on first glance to me that it should be `v?` (you can test/ see on https://regex101.com/)

core/src/build-staging/rsync.ts
```diff
@@ -14,7 +14,7 @@ import { BuildStaging, SyncParams } from "./build-staging"
...
+const versionRegex = /rsync\s+version\s+[v]*(\d+.\d+.\d+)/
```
stefreak, line 17:
```suggestion
const versionRegex = /rsync\s+version\s+v?(\d+.\d+.\d+)/
```

core/src/vcs/vcs.ts
```diff
@@ -38,7 +38,7 @@ export const NEW_MODULE_VERSION = "0000000000"
 const fileCountWarningThreshold = 10000
 
 const minGitVersion = "2.14.0"
-export const gitVersionRegex = /git version [v]*([\d\.]+)/
+export const gitVersionRegex = /git\s+version\s+[v]*(\d+.\d+.\d+)/
```
stefreak, line 41:
```suggestion
export const gitVersionRegex = /git\s+version\s+v?(\d+.\d+.\d+)/
```

stefreak (approved):
LGTM

I'm still in huge favor of somehow getting rid of the requirement of an updated rsync, but until then: 👍

PR #3236, opened 2022-09-20
chore: remove get-vaccine

stefreak:
the command does not work as the impfstoff.link API does not exist anymore.

Fixes #


edvald:
Can we wrap this one up? If it hadn't been resolved prior, the reason `yarn generate-docs` might still have been generating outdated docs would be that you hadn't built the updated code first.

stefreak:
Thank you @edvald! The PR is now rebased and docs are regenerated, ready for review.

PR #3239, opened 2022-09-20
feat: simple mode=max support with a list of not supported registries

stefreak:
- [ ] TODO: This code is a WIP and not tested yet

Fixes #


core/src/plugins/kubernetes/config.ts
```diff
@@ -465,6 +466,30 @@ export const kubernetesConfigBase = () =>
     clusterBuildkit: joi
       .object()
       .keys({
+        overrideMultiStageCacheSupport: joi
```
edvald, line 469: I wonder if we should change this to something like `overrideCacheMode` and make it a string value (min or max)? Perhaps one might want to force the other way, and it's conceivable that more modes get added at some point.

stefreak: yes, that's a good idea.

core/src/plugins/kubernetes/config.ts
```diff
@@ -465,6 +513,35 @@ export const kubernetesConfigBase = () =>
     clusterBuildkit: joi
       .object()
       .keys({
+        cache: joi
+          .array()
+          .items(buildkitCacheConfigurationSchema())
+          // TODO: fix default value, to include defaults automatically
+          .default([{ type: "registry", mode: "auto", tag: "_buildcache", export: true }])
```
stefreak, line 520: @vvagaytsev this is a weird one, can you help us with this one?

We don't want to repeat the defaults in the `buildkitCacheConfigurationSchema`, but when we just use `.default([{type: "registry"}])` it will not deep-default the `mode`, `tag`, etc.

core/src/plugins/kubernetes/config.ts
stefreak: @edvald @TimBeyer Tim and me paired on this today, and came up with a different config format, see the PR description for a couple of examples: https://github.com/garden-io/garden/pull/3239

What do you think of this?

We'll add some unit tests tomorrow and document it better, but it already works as of now in this PR

core/src/plugins/kubernetes/config.ts
```diff
@@ -447,6 +455,46 @@ const tlsCertificateSchema = () =>
...
+      `
+      ),
+    tag: joi
+      .string()
+      .default("_buildcache")
+      .description(
+        dedent`
+        This is the tag name for the registry build cache. Default is \`_buildcache\`
+
+        **NOTE**: tag can only be used together with the \`registry\` cache type
```
stefreak, line 485: maybe we should use `joi.when`, so the tag option only exists when `type: registry`

docs/reference/providers/local-kubernetes.md
```diff
@@ -440,9 +529,134 @@ For more details on all the different options and what makes sense to use for yo
...
+```
+
+We also support more advanced cache configurations, like the following:
+
+```
+clusterBuildkit:
+  cache:
+    - type: registry
+      mode: max
+      tag: _buildcache-\${camelCase(git.branch)}
```
stefreak, line 584: unfortunately I think the backslash here is a in our docs generator

in a javascript backtick string that we use, I must escape the dollar sign, otherwise it will try to call `camelCase`. The resulting string should not contain the backslash.

Or am I missing something?

stefreak: It's not the first render issue in our docs, so maybe let's open an upstream bug report and ignore this for now?

stefreak: Made a separate bug report for this: https://github.com/garden-io/garden/issues/3245

stefreak: applied the workaround in the bug description.

docs/reference/providers/kubernetes.md
```diff
@@ -487,10 +602,211 @@ For more details on all the different options and what makes sense to use for yo
...
+| -------- | -------- |
+| `string` | Yes      |
+
+### `providers[].clusterBuildkit.cache[].registry`
+
+[providers](#providers) > [clusterBuildkit](#providersclusterbuildkit) > [cache](#providersclusterbuildkitcache) > registry
+
+The registry from which the cache should be imported from, or which it should be exported to.
+
+Important: You must make sure `imagePullSecrets` includes authentication with the specified cache registry, that has the appropriate write privileges (usually full write access to the configured `deploymentRegistry.namespace`).
```
stefreak, line 698:
```suggestion
Important: You must make sure `imagePullSecrets` includes authentication with the specified cache registry, that has the appropriate write privileges (usually full write access to the configured `registry.namespace`).
```

core/src/plugins/kubernetes/config.ts
stefreak: so far we only have registry. can be done later 🤷

stefreak:
The test failure seems to be unrelated to this change.

core/src/plugins/kubernetes/config.ts
```diff
@@ -465,6 +536,77 @@ export const kubernetesConfigBase = () =>
...
       .object()
       .keys({
+        cache: joi
+          .array()
+          .items(buildkitCacheConfigurationSchema())
+          // TODO: fix default value, to include defaults automatically
+          .default([{ type: "registry", mode: "auto", tag: "_buildcache", export: true }])
+          .description(
+            dedent`
+            Use the \`cache\` configuration to customize the default cluster-buildkit cache behaviour.
```
edvald, line 546: I think it would be good to add a section to the in-cluster building guide on this.

core/src/plugins/kubernetes/config.ts
```diff
@@ -447,6 +456,68 @@ const tlsCertificateSchema = () =>
...
+      .default(true)
+      .description(
+        dedent`
+        If this is false, only pass the \`--import-cache\` option to buildkit, and not the \`--export-cache\` option. Defaults to true.
+      `
+      ),
+  })
+
+// work around docs generator bug #3245
+const camelCaseGitBranch = "${camelCase(git.branch)}"
```
edvald, line 519: Maybe unimportant, but would kebab case make more sense here?

core/src/plugins/kubernetes/container/build/buildkit.ts
```diff
@@ -254,6 +242,100 @@ export function getBuildkitFlags(module: ContainerModule) {
...
+}
+
+function getCacheImageName(moduleOutputs: ContainerModuleOutputs, cacheConfig: ClusterBuildkitCacheConfig): string {
+  if (cacheConfig.registry === undefined) {
+    return moduleOutputs["deployment-image-name"]
+  }
+
+  const { hostname, port, namespace } = cacheConfig.registry
+  const portPart = port ? `:${port}` : ""
+  return `${hostname}${portPart}/${namespace}/${moduleOutputs["local-image-name"]}`
```
edvald, line 314: You may want to use the `unparseImageId` helper here: https://github.com/garden-io/garden/blob/6bed7f664cd2f337a0e546852defcff224bb073d/core/src/plugins/container/helpers.ts#L227

core/src/plugins/kubernetes/container/build/buildkit.ts
```diff
@@ -254,6 +242,100 @@ export function getBuildkitFlags(module: ContainerModule) {
...
+    let registryExtraSpec = ""
+    if (cache.registry === undefined) {
+      registryExtraSpec = deploymentRegistryExtraSpec
+    } else if (cache.registry?.insecure === true) {
+      registryExtraSpec = ",registry.insecure=true"
+    }
+
+    // subtle: it is important that --import-cache arguments are in the same order as the cacheConfigs
+    // buildkit will go through them one by one, and use the first that has any cache hit for all following
+    // layers, so it will actually never use multiple caches at once
```
edvald, line 285: Would it make sense to explain this in the docs as well?

core/src/plugins/kubernetes/container/build/buildkit.ts
```diff
@@ -254,6 +242,100 @@ export function getBuildkitFlags(module: ContainerModule) {
...
+
+  // Detect AWS ECR
+  if (deploymentImageName.includes(".dkr.ecr.")) {
+    return "inline"
+  }
+
+  // Detect gcr.io
+  if (deploymentImageName.includes("gcr.io")) {
+    return "inline"
+  }
```
edvald, line 333: It may be unnecessary, not sure, but we could use `parseImageId` to narrow this detection to the hostname part specifically.

core/src/plugins/kubernetes/container/build/buildkit.ts
```diff
@@ -254,6 +242,100 @@ export function getBuildkitFlags(module: ContainerModule) {
...
+
+  // Default to true for all others
```
edvald, line 335:
```suggestion
  // Default to max for all others
```

edvald (changes requested):
Had some comments and suggestions, but overall this is great and a very welcome improvement 👏 

I'd love to see more documentation in the in-cluster building guide around this as well, as opposed to just documenting in the reference. A suggested good practice in the guide would make a lot of sense imo.

core/src/plugins/kubernetes/config.ts
```diff
@@ -465,6 +536,77 @@ export const kubernetesConfigBase = () =>
...
+          .array()
+          .items(buildkitCacheConfigurationSchema())
+          // TODO: fix default value, to include defaults automatically
+          .default([{ type: "registry", mode: "auto", tag: "_buildcache", export: true }])
+          .description(
+            dedent`
+            Use the \`cache\` configuration to customize the default cluster-buildkit cache behaviour.
+
+            The default value is:
+            \`\`\`
```
thsig, line 549: Add `yaml` here (for syntax highlighting in the rendered docs).
```suggestion
            \`\`\`yaml
```

core/src/plugins/kubernetes/config.ts
```diff
@@ -465,6 +536,77 @@ export const kubernetesConfigBase = () =>
...
+
+            | Registry Name                   | Detection string | Assumed \`mode=max\` support |
+            |---------------------------------|------------------|------------------------------|
+            | AWS Elastic Container Registry  | \`.dkr.ecr.\`    | No                           |
+            | Google Cloud Container Registry | \`gcr.io\`       | No                           |
+            | Any other registry              | -                | Yes                          |
+
+            In case you need to override the defaults for your registry, you can do it like so:
+
+            \`\`\`
```
thsig, line 576:
```suggestion
            \`\`\`yaml
```

core/src/plugins/kubernetes/config.ts
```diff
@@ -465,6 +536,77 @@ export const kubernetesConfigBase = () =>
...
+            clusterBuildkit:
+              cache:
+                - type: registry
+                  mode: inline
+            \`\`\`
+
+            We also support more advanced cache configurations, like the following:
+
+            \`\`\`
```
thsig, line 585:
```suggestion
            \`\`\`yaml
```

core/src/plugins/kubernetes/config.ts
```diff
@@ -465,6 +536,77 @@ export const kubernetesConfigBase = () =>
...
+
+            Using this cache configuration, we every build will first look for a cache specific to your feature branch.
+            If it does not exist yet, it will import caches from the main branch builds (\`_buildcache-main\`).
+            When the build is finished, it will only export caches to your feature branch, and avoid polluting the \`main\` branch caches.
+            A configuration like that may improve your cache hit rate and thus save time.
+
+            If you need to disable caches completely you can achieve that with the following configuration:
+
+            \`\`\`
```
thsig, line 604:
```suggestion
            \`\`\`yaml
```

core/src/plugins/kubernetes/config.ts
```diff
@@ -447,6 +456,68 @@ const tlsCertificateSchema = () =>
...
+          Use the Docker registry configured at \`deploymentRegistry\` to retrieve and store buildkit cache information.
+
+          See also the [buildkit registry cache documentation](https://github.com/moby/buildkit#registry-push-image-and-cache-separately)
+        `
+      ),
+    registry: containerRegistryConfigSchema().description(
+      dedent`
+      The registry from which the cache should be imported from, or which it should be exported to.
+
+      Important: You must make sure \`imagePullSecrets\` includes authentication with the specified cache registry, that has the appropriate write privileges (usually full write access to the configured \`deploymentRegistry.namespace\`).
```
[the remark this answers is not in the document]
stefreak: Absolutely, will do that

core/src/plugins/kubernetes/config.ts
```diff
@@ -447,6 +456,68 @@ const tlsCertificateSchema = () =>
       .example("cert-manager"),
   })
 
+const buildkitCacheConfigurationSchema = () =>
+  joi.object().keys({
+    type: joi
+      .string()
+      .allow("registry")
```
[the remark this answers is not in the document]
stefreak: Makes sense 👍 
There might be future additions like `s3`

core/src/plugins/kubernetes/config.ts
```diff
@@ -465,6 +536,77 @@ export const kubernetesConfigBase = () =>
...
+                - type: registry
+                  mode: max
+                  tag: _buildcache-${camelCaseGitBranch}
+                - type: registry
+                  mode: max
+                  tag: _buildcache-main
+                  export: false
+            \`\`\`
+
+            Using this cache configuration, we every build will first look for a cache specific to your feature branch.
```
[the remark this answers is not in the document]
stefreak:
```suggestion
            Using this cache configuration, every build will first look for a cache specific to your feature branch.
```

core/test/integ/src/plugins/kubernetes/container/build/buildkit.ts
```diff
@@ -82,7 +82,7 @@ grouped("cluster-buildkit").describe("ensureBuildkit", () => {
 
       const nodeSelector = { "kubernetes.io/os": "linux" }
 
-      provider.config.clusterBuildkit = { nodeSelector }
+      provider.config.clusterBuildkit = { nodeSelector, cache: [] }
```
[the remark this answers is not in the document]
stefreak: 👍

core/src/plugins/kubernetes/config.ts
stefreak: Yes you're right. Is there also a function that limits the length of the string? Branch names can be longer than allowed tag names 🤔

core/test/unit/src/plugins/kubernetes/container/build/buildkit.ts
```diff
@@ -29,3 +34,162 @@ describe("getBuildkitFlags", () => {
...
+    "local-image-id": "name:v-xxxxxx",
+    "local-image-name": "name",
+    "deployment-image-id": `${registry}/namespace/name:v-xxxxxx`,
+    "deployment-image-name": `${registry}/namespace/name`,
+  })
+
+  it("returns mode=max cache flags with default config by default", async () => {
+    const flags = getBuildkitImageFlags(defaultConfig, moduleOutputs({ registry: "hub.docker.com" }), false)
+
+    expect(flags).to.eql(modeFlags({ mode: "max", registry: "hub.docker.com" }))
```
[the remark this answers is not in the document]
stefreak: No I was actually also thinking about that. That is a very good point. I did it mostly to make it a little bit more obvious, what exactly I am expecting in this test case, as the important strings drown in all of the noise.

I think I have an idea how I can achieve that without the magic 👍

docs/reference/commands.md
```diff
@@ -302,7 +302,7 @@ Examples:
 
 #### Usage
 
-    garden config analytics-enabled [enable]
+    garden config analytics-enabled [enable]
```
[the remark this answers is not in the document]
stefreak: Unsure, I will rebase with master later and see if this still happens then

core/src/plugins/kubernetes/config.ts
```diff
@@ -465,6 +538,77 @@ export const kubernetesConfigBase = () =>
     clusterBuildkit: joi
       .object()
       .keys({
+        cache: joi
+          .array()
+          .items(buildkitCacheConfigurationSchema())
+          // TODO: fix default value, to include defaults automatically
```
TimBeyer, line 544: I feel like we can just ignore that and keep hardcoding the default. That way it's also very explicit what the default is.

stefreak:
```suggestion
```

core/src/plugins/kubernetes/config.ts
stefreak: decided that an explicit default is ok

docs/README.md
```diff
@@ -101,11 +101,6 @@
 * [Project Configuration](./reference/project-config.md)
 * [Module Template Configuration](./reference/module-template-config.md)
 * [Workflow Configuration](./reference/workflow-config.md)
-* [Action Types](./reference/action-types/Build)
-  * [Build](./reference/action-types/Build/container.md)
-    * [`container` Build](./reference/action-types/Build/container.md)
-    * [`exec` Build](./reference/action-types/Build/exec.md)
-    * [`jib-container` Build](./reference/action-types/Build/jib-container.md)
```
stefreak, line 108: I have no idea why the docs generator removed this. This doesnt look right

core/test/integ/src/plugins/kubernetes/container/build/buildkit.ts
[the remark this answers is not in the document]
stefreak: I think all the integration tests that push to registries, e.g. the gcr.io tests and the in-cluster registry tests are already testing this. Nonetheless I added the defaultCache to the `buildkit.ts` tests. They will only spin up a buildkit and not test actually building something, so I did not add any cache specific tests to `buildkit.ts`

core/src/plugins/kubernetes/container/build/buildkit.ts
stefreak: I tried this, and it makes things very very complicated ... and it doesn't support the port (the in-cluster-registry test fails)

diff:

```
diff --git a/core/src/plugins/container/helpers.ts b/core/src/plugins/container/helpers.ts
index 190f66a2..2497e9f3 100644
--- a/core/src/plugins/container/helpers.ts
+++ b/core/src/plugins/container/helpers.ts
@@ -45,7 +45,7 @@ export const minDockerVersion: DockerVersion = {
   server: "17.07.0",
 }
 
-interface ParsedImageId {
+export interface ParsedImageId {
   host?: string
   namespace?: string
   repository: string
diff --git a/core/src/plugins/kubernetes/container/build/buildkit.ts b/core/src/plugins/kubernetes/container/build/buildkit.ts
index 05a109ed..8863f701 100644
--- a/core/src/plugins/kubernetes/container/build/buildkit.ts
+++ b/core/src/plugins/kubernetes/container/build/buildkit.ts
@@ -37,7 +37,8 @@ import { getRunningDeploymentPod, millicpuToString, megabytesToString, usingInCl
 import { PodRunner } from "../../run"
 import { prepareSecrets } from "../../secrets"
 import { ContainerModuleOutputs } from "../../../container/container"
-import { containerHelpers } from "../../../container/helpers"
+import { containerHelpers, ParsedImageId } from "../../../container/helpers"
+import { PluginError } from "../../../../exceptions"
 
 export const buildkitImageName = "gardendev/buildkit:v0.9.3-1"
 export const buildkitDeploymentName = "garden-buildkit"
@@ -260,7 +261,7 @@ export function getBuildkitImageFlags(
 
     for (const cache of inlineCaches) {
       const cacheImageName = getCacheImageName(moduleOutputs, cache)
-      imageNames.push(`${cacheImageName}:${cache.tag}`)
+      imageNames.push(containerHelpers.unparseImageId(cacheImageName))
     }
   }
 
@@ -281,10 +282,12 @@ export function getBuildkitImageFlags(
       registryExtraSpec = ",registry.insecure=true"
     }
 
+    const parsedCacheImageName = containerHelpers.unparseImageId(cacheImageName)
+
     // subtle: it is important that --import-cache arguments are in the same order as the cacheConfigs
     // buildkit will go through them one by one, and use the first that has any cache hit for all following
     // layers, so it will actually never use multiple caches at once
-    args.push("--import-cache", `type=registry,ref=${cacheImageName}:${cache.tag}${registryExtraSpec}`)
+    args.push("--import-cache", `type=registry,ref=${parsedCacheImageName}${registryExtraSpec}`)
 
     if (cache.export === false) {
       continue
@@ -296,46 +299,56 @@ export function getBuildkitImageFlags(
       continue
     }
 
-    args.push(
-      "--export-cache",
-      `type=registry,ref=${cacheImageName}:${cache.tag},mode=${cacheMode}${registryExtraSpec}`
-    )
+    args.push("--export-cache", `type=registry,ref=${parsedCacheImageName},mode=${cacheMode}${registryExtraSpec}`)
   }
 
   return args
 }
 
-function getCacheImageName(moduleOutputs: ContainerModuleOutputs, cacheConfig: ClusterBuildkitCacheConfig): string {
+function getCacheImageName(
+  moduleOutputs: ContainerModuleOutputs,
+  cacheConfig: ClusterBuildkitCacheConfig
+): ParsedImageId {
   if (cacheConfig.registry === undefined) {
-    return moduleOutputs["deployment-image-name"]
+    return {
+      ...containerHelpers.parseImageId(moduleOutputs["deployment-image-name"]),
+      tag: cacheConfig.tag,
+    }
   }
 
   const { hostname, port, namespace } = cacheConfig.registry
   const portPart = port ? `:${port}` : ""
-  return containerHelpers.unparseImageId({
-    repository: moduleOutputs["local-image-name"],
+  return {
+    ...containerHelpers.parseImageId(moduleOutputs["local-image-name"]),
     host: `${hostname}${portPart}`,
     namespace,
-  })
+    tag: cacheConfig.tag,
+  }
 }
 
 export const getSupportedCacheMode = (
   cache: ClusterBuildkitCacheConfig,
-  deploymentImageName: string
+  deploymentImageName: ParsedImageId
 ): ClusterBuildkitCacheConfig["mode"] => {
   if (cache.mode !== "auto") {
     return cache.mode
   }
 
-  const { host } = containerHelpers.parseImageId(deploymentImageName)
+  const host = deploymentImageName.host
+  if (host === undefined) {
+    throw new PluginError("getSupportedCacheMode: unexpectedly missing host in deploymentImageName", {
+      deploymentImageName,
+      cacheConfigItem: cache,
+    })
+  }
 
   // Detect AWS ECR
-  if (host!.includes(".dkr.ecr.")) {
+  if (host.includes(".dkr.ecr.")) {
     return "inline"
   }
 
   // Detect gcr.io
-  if (host!.includes("gcr.io")) {
+  if (host.includes("gcr.io")) {
     return "inline"
   }

```

core/src/plugins/kubernetes/container/build/buildkit.ts
stefreak: Same with https://github.com/garden-io/garden/pull/3239/files#r978832787

the test for in-cluster registry will fail.... all the code looks very complex, I thought about adding port support but I'd be very afraid of breaking something else with that

```
diff --git a/core/src/plugins/kubernetes/container/build/buildkit.ts b/core/src/plugins/kubernetes/container/build/buildkit.ts
index 980b9378..37e418a9 100644
--- a/core/src/plugins/kubernetes/container/build/buildkit.ts
+++ b/core/src/plugins/kubernetes/container/build/buildkit.ts
@@ -37,6 +37,7 @@ import { getRunningDeploymentPod, millicpuToString, megabytesToString, usingInCl
 import { PodRunner } from "../../run"
 import { prepareSecrets } from "../../secrets"
 import { ContainerModuleOutputs } from "../../../container/container"
+import { containerHelpers } from "../../../container/helpers"

 export const buildkitImageName = "gardendev/buildkit:v0.9.3-1"
 export const buildkitDeploymentName = "garden-buildkit"
@@ -322,13 +323,15 @@ export const getSupportedCacheMode = (
     return cache.mode
   }

+  const host = containerHelpers.parseImageId(deploymentImageName).host!
+
   // Detect AWS ECR
-  if (deploymentImageName.includes(".dkr.ecr.")) {
+  if (host.includes(".dkr.ecr.")) {
     return "inline"
   }

   // Detect gcr.io
-  if (deploymentImageName.includes("gcr.io")) {
+  if (host.includes("gcr.io")) {
     return "inline"
   }
```

core/src/plugins/kubernetes/config.ts
stefreak: Done!

stefreak:
@edvald thank you for the review! I tried to address all the comments, and when I didn't I tried to explain why.

See you after vacation 👍 🍹

docs/README.md
edvald: This wasn't supposed to be in the docs (it seems to have slipped in from 0.13), so this is all good.

thsig (approved):

edvald (approved):

PR #3242, opened 2022-09-21
Graph v2 partial fixes in `VcsHandler` unit tests

vvagaytsev:
TODOs:
* [x] add new unit tests for the functions declared as exported in this PR
* [x] fix the runtime error in the failing test
* ~fix the assertion in the failing unit test~

Fixes #

### DIFF
--- a/core/test/unit/src/vcs/vcs.ts
+++ b/core/test/unit/src/vcs/vcs.ts
@@ -355,6 +355,7 @@ describe("getModuleVersionString", () => {
     const garden = await makeTestGarden(projectRoot, { noCache: true })
     const module = await garden.resolveModule("module-a")
 
+    // TODO-G2: the assertion below still fails, check if there is something changed in the hash calculation
     const fixedVersionString = "v-2db99eee24"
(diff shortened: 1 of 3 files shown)


core/test/unit/src/vcs/vcs.ts
```diff
@@ -347,6 +355,7 @@ describe("getModuleVersionString", () => {
     const garden = await makeTestGarden(projectRoot, { noCache: true })
     const module = await garden.resolveModule("module-a")
 
+    // TODO-G2: the assertion below still fails, check if there is something changed in the hash calculation
```
vvagaytsev, line 358: @edvald @Orzelius has anything changed in the version hash calculation in SGv2? I'm getting `v-fd01b4fd6b` now instead of the old hard-coded value `v-2db99eee24`.

I'm wondering if it's a bug or an expected change.

edvald: Yeah the version hashes will probably change in all cases after the change.

PR #3244, opened 2022-09-22
Various test and schema fixes SGV2

Orzelius:
I'm still working on the actions.ts tests

Currently there seems to be a bug in the resolveModules functions resulting in it returning no actions. I will look into that next.

Please see individual commits for more details.

Some of the code was written in pair with @thsig


core/src/plugin/handlers/provider/augmentGraph.ts
```diff
@@ -6,22 +6,22 @@
...
 
 export interface AugmentGraphParams extends PluginActionParamsBase {
   actions: BaseAction[]
   providers: ProviderMap
 }
 
 interface AddDependency {
-  by: string
-  on: string
+  by: { kind: ActionKind; name: string }
```
vvagaytsev, line 23: Maybe it's an overuse of typing, but `type Dependency = { kind: ActionKind; name: string }` can be extracted here. So the `AddDependency` would look like:
```
interface AddDependency {
  by: Dependency
  on: Dependency

**Review by vvagaytsev — APPROVED** [review]
LGTM. But it looks like the last commit contains some unfinished changes. Please check the items below, otherwise looks good.
* In https://github.com/garden-io/garden/pull/3244/commits/04747027b1c2754b9296fd32ef790e96586f3431#diff-af4130d3ce6bfd63ecc6bb23a507a76e14823eb4611425b1abf16e93f15cc4b6R86-R88 the statement `graph.getBuild("build.module-a")` looks a bit odd, the method expects an action name instead of an action key.
* https://github.com/garden-io/garden/pull/3244/commits/04747027b1c2754b9296fd32ef790e96586f3431#diff-af4130d3ce6bfd63ecc6bb23a507a76e14823eb4611425b1abf16e93f15cc4b6R131 looks like leftover from the debugging
* In https://github.com/garden-io/garden/pull/3244/commits/04747027b1c2754b9296fd32ef790e96586f3431#diff-af4130d3ce6bfd63ecc6bb23a507a76e14823eb4611425b1abf16e93f15cc4b6R174 should the field `addDependencies` have type `AddDependency[]`?

File: core/src/plugin/handlers/provider/augmentGraph.ts (hunk 1 above)
Comment (Orzelius) [reply core/src/plugin/handlers/provider/augmentGraph.ts]: good point, fixed in [9cc599b](https://github.com/garden-io/garden/pull/3244/commits/9cc599b65f4806aa03466486f582d407067376f5)

**Comment by Orzelius** [comment]
> But it looks like the last commit contains some unfinished changes

Yeah I'm sill working on fixing the actual tests I just wanted to get the core changes out so they may help others

**Comment by vvagaytsev** [comment]
Great, it's ready to go then :) 🚀

File: core/src/config/common.ts [hunk 2]
```diff
@@ -146,7 +146,7 @@ export interface PosixPathSchema extends Joi.StringSchema {
 
 interface ActionReferenceSchema extends Joi.AnySchema {
   kind(kind: ActionKind): this
-  actionType(type: string): this
+  name(type: string): this
```
edvald, line 149: We're mixing up name and type here. Since we won't be able to validate the type of the action from the reference itself, I suggest commenting this out for now and removing usage of the method.

core/test/unit/src/actions.ts
```diff
@@ -83,7 +83,9 @@ describe("ActionRouter", () => {
     actionRouter = await garden.getActionRouter()
     graph = await garden.getConfigGraph({ log: garden.log, emit: false })
     module = graph.getModule("module-a")
-    const buildAction = graph.getBuild("module-a")
+    // TODO
+    const actions = graph.getActions()
+    const buildAction = graph.getBuild("build.module-a")
```
edvald, line 88: Should be just `graph.getBuild("module-a")`

core/test/unit/src/actions.ts
```diff
@@ -126,7 +128,7 @@ describe("ActionRouter", () => {
   // Note: The test plugins below implicitly validate input params for each of the tests
   describe("environment actions", () => {
     describe("configureProvider", () => {
-      it("should configure the provider", async () => {
+      it.only("should configure the provider", async () => {
```
edvald, line 131: Left the `.only` here

vvagaytsev: I fixed that in `graph-v2` after the merge.

PR #3250, opened 2022-09-25
Couple of fixes for graph v2

edvald:
cc @vvagaytsev @Orzelius

### DIFF
--- a/core/test/unit/src/actions.ts
+++ b/core/test/unit/src/actions.ts
@@ -2100,7 +2100,7 @@ const testPlugin = createGardenPlugin({
 
           const buildAction: BuildActionConfig = {
             kind: "Build",
-            type: "exec",
+            type: "test",
             name: module.name,
 
             ...params.baseFields,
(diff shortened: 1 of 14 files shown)


core/test/unit/src/actions.ts
```diff
@@ -2086,9 +2086,98 @@ const testPlugin = createGardenPlugin({
...
+            convertBuildDependency,
+            convertRuntimeDependencies,
+          } = params
 
-          return {}
+          const actions: (BuildActionConfig | BaseRuntimeActionConfig)[] = []
+
+          const buildAction: BuildActionConfig = {
+            kind: "Build",
+            type: "exec",
```
Orzelius, line 2103: why are some actions here converted to an exec type and others test?

Orzelius (approved):

core/test/unit/src/actions.ts
edvald: Because copy paste :P This should indeed be a test type.

PR #3252, opened 2022-09-26
Linux Dev Env Setup and asdf-based dev setup script

highb:
- chore: add install-asdf-dependencies.sh script
- docs: update linux development documentation

Updates the contribution docs to note that the provided scripts work when using Linux Brew, and adds a second script for dev setup that uses the tool [asdf](https://asdf-vm.com/) to manage several of the tool versions via the `.tool-versions` config file, which I would propose is a much easier method of dealing with tool versions during development and makes it much easier to swap out versions of the tools.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,0 +1,5 @@
+jq 1.6
+nodejs 14.20.1
+kubectl 1.24.6
+helm 3.10.0
+stern 1.21.0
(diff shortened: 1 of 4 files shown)


Orzelius (commented):
Just one typo, but other than that this is very cool. I installed `asdf` on my system and tried out the script and it worked like a charm.

highb:
Ah, thanks for catching that!

vvagaytsev (approved):

PR #3253, opened 2022-09-26
chore: fix cli build-pkg

highb:
`pkg` creates the output directory automatically and will fail
if the directory already exists.

Fixes CI, allows actually building artifacts.

Fixes #


highb:
Hm, well it builds locally...

PR #3257, opened 2022-09-26
improvement(local-mode): multiple port-forwards

vvagaytsev:

Fixes #


core/src/util/recoverable-process.ts
```diff
@@ -259,6 +261,10 @@ export class RecoverableProcess {
     return this.lastKnownPid
   }
 
+  public getProcessDescription(): string {
+    return !!this.command.description ? this.command.description : renderOsCommand(this.command)
```
Orzelius, line 265: Maybe would be easier to read like this
```suggestion
    return this.command.description || renderOsCommand(this.command)
```

core/src/plugins/kubernetes/local-mode.ts
```diff
@@ -474,16 +474,16 @@ function getLocalAppCommand({ spec: localModeSpec, gardenService }: StartLocalMo
...
-  const localServiceCmd = getLocalAppCommand(configParams)
+  const localAppCmd = getLocalAppCommand(configParams)
   const { ctx, gardenService, log } = configParams
 
   // This covers Win \r\n, Linux \n, and MacOS \r line separators.
   const eolRegex = /\r?\n?$/
   const stripEol = (message: string) => message.replace(eolRegex, "")
 
-  return !!localServiceCmd
+  return !!localAppCmd
```
Orzelius, line 482: not directly related to the pr code, but maybe put this return in a variable, the conditional statement here is like 20+ lines long

Orzelius (commented):
Added a few comments. Also @Walther do you want to take a look at this? You can see the [example project](https://github.com/garden-io/garden/tree/main/examples/local-mode) to get familiar with the concept first.

core/src/util/recoverable-process.ts
vvagaytsev: Nice!

core/src/plugins/kubernetes/local-mode.ts
vvagaytsev: Good point! Fixed it in https://github.com/garden-io/garden/pull/3257/commits/56c4776022740681a179b7b8be03641f647e193c

Walther (approved):
🚢

PR #3260, opened 2022-09-28
SHV2: Fix some action tests, some schemas and some logic

Orzelius:
please see individual commits for more details :)

Please note that some changes include logic fixes in addition to tests

### DIFF
--- a/core/src/actions/base.ts
+++ b/core/src/actions/base.ts
@@ -244,7 +244,7 @@ export const baseRuntimeActionConfig = () =>
 
 export const actionStatusSchema = (detailSchema?: Joi.ObjectSchema) =>
   joi.object().keys({
-    status: joi
+    state: joi
       .string()
       .allow(...actionStateTypes)
       .only()
--- a/core/src/events.ts
+++ b/core/src/events.ts
@@ -19,6 +19,7 @@ import { BuildState } from "./plugin/handlers/build/get-status"
 import { ActionReference } from "./config/common"
 import { GraphResult } from "./graph/results"
 import { sanitizeObject } from "./logger/logger"
+import { ModuleVersion } from "./vcs/vcs"
 
 export type GardenEventListener<T extends EventName> = (payload: Events[T]) => void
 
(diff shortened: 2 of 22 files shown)


core/src/events.ts
```diff
@@ -223,7 +224,7 @@ export interface Events extends LoggerEvents {
 
     // DEPRECATED: remove in 0.14
     moduleName: string
-    moduleVersion: string
+    moduleVersion: ModuleVersion
```
vvagaytsev, line 227: Do we need to include the whole `ModuleVersion` object in the events? That can be quite heavy. How about storing only `moduleVersionString` here?

core/test/unit/src/actions.ts
```diff
@@ -381,7 +395,7 @@ describe("ActionRouter", () => {
         //   {
         //     name: "test",
         //     dependencies: [],
-        //     disabled: false,
+        //     disabled: false,module actions
```
vvagaytsev, line 398: This looks like some leftover or unfinished comment.

vvagaytsev:
Looks good! 👍  I left a couple of comments, please check.

core/src/actions/base.ts
```diff
@@ -244,7 +244,7 @@ export const baseRuntimeActionConfig = () =>
 
 export const actionStatusSchema = (detailSchema?: Joi.ObjectSchema) =>
   joi.object().keys({
-    status: joi
+    state: joi
```
vvagaytsev, line 247: I think this change is robust, so we can re-generate the docs in the same PR.

Orzelius: Done in [f92212d](https://github.com/garden-io/garden/pull/3260/commits/f92212dab3ab7385782f726f80f86ad7f53b6e9f)

core/src/events.ts
Orzelius: makes sense, can always revisit if need be [c6d0d60](https://github.com/garden-io/garden/pull/3260/commits/c6d0d60a8ec97d35bc5c8d47603a60e79dbc3bad)

core/test/unit/src/actions.ts
Orzelius: oops [706705b](https://github.com/garden-io/garden/pull/3260/commits/706705bfb7a8a0e7ce22863eee94461854364abd)

vvagaytsev (approved):

PR #3262, opened 2022-09-28
docs: regenerated docs

vvagaytsev:

Fixes #


Orzelius (approved):
probably no point in rendering them as they will change so much during development, but why not

vvagaytsev:
> probably no point in rendering them as they will change so much during development, but why not

They will change, that's right. But now I'm getting tons of unwanted docs changes after rebuilding the project when I work in my branches :)

PR #3264, opened 2022-09-29
Fix more action things

Orzelius:
Fix some logic and some tests, mostly related to actions and module resolution

Please see individual commits for more detail.

### DIFF
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -276,7 +276,7 @@ export async function executeAction<T extends Action>({
 
   const results = await garden.processTasks({ tasks: [task], log, throwOnError: true })
 
-  return <Executed<T>>(<unknown>results.results.getResult(task)!.result!.outputs.executedAction)
+  return <Executed<T>>(<unknown>results.results.getResult(task)!.result!.executedAction)
 }
 
 const getActionConfigContextKeys = memoize(() => {
--- a/core/test/unit/src/actions.ts
+++ b/core/test/unit/src/actions.ts
@@ -386,25 +386,19 @@ describe("ActionRouter", () => {
     })
 
     describe("test.run", () => {
+      const actionConfig: TestActionConfig = {
+        name: "test",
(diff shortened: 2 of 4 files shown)


core/src/graph/actions.ts
```diff
@@ -276,7 +276,7 @@ export async function executeAction<T extends Action>({
 
   const results = await garden.processTasks({ tasks: [task], log, throwOnError: true })
 
-  return <Executed<T>>(<unknown>results.results.getResult(task)!.result!.outputs.executedAction)
+  return <Executed<T>>(<unknown>results.results.getResult(task)!.result!.executedAction)
```
vvagaytsev, line 279: I'm not that deep into actions router yet, but should the same fi be applied to `resolveAction(s)` functions in https://github.com/garden-io/garden/blob/graph-v2/core/src/graph/actions.ts#L209 and https://github.com/garden-io/garden/blob/graph-v2/core/src/graph/actions.ts#L248?

core/test/unit/src/actions.ts
```diff
@@ -532,45 +489,35 @@ describe("ActionRouter", () => {
...
-        })
+
+        expect(result.outputs).to.eql({})
+        expect(result.detail?.log).to.eql("bla bla")
+        expect(result.state).to.eql("ready")
       })
     })
 
     it("should emit a testStatus event", async () => {
+      const all = graph.getActions()
```
vvagaytsev, line 507: Unused variable declaration

vvagaytsev (changes requested):
Looks good, but I left a couple of comments. Please check.

core/test/unit/src/actions.ts
Orzelius: good catch, had this left from debugging. [addea3b](https://github.com/garden-io/garden/pull/3264/commits/addea3bace2ad5bd63a7194112487af8725dca9a)

vvagaytsev (approved):

PR #3277, opened 2022-09-30
fix(cli): fixed alias of `--skip-dependencies` flag in garden commands

### LINKED ISSUE #3275: [BUG] --no-deps option of deploy command not working
### Current Behavior

The `deploy` command does not recognize the `--no-deps` option:
```
$ garden deploy --no-deps my-service
Unrecognized option flag --deps
```

However, using the 'long' version of the same option works:
```
$ garden deploy --skip-dependencies my-service
```

### Expected behavior

Either `--no-deps` should work like `--skip-dependencies` or `--no-deps` should be removed from the `garden deploy --help` output.

`garden version`
0.12.44

vvagaytsev:
It looks like the alias name cannot contain any dash characters (`-`).
So, the `--no-deps` alias for `--skip-dependencies` hasn't been working properly since the `0.12.34` - the release when the `--skip-dependencies` flag was introduced.

This PR fixes the issue by replacing `--no-deps` alias with `--nodeps`.

Fixes #3275

Formally, it's a breaking change. But, in fact, it is not, because the old flag's alias hasn't been working properly for a long time.

### DIFF
--- a/core/src/commands/deploy.ts
+++ b/core/src/commands/deploy.ts
@@ -87,7 +87,7 @@ export const deployOpts = {
     This can be useful e.g. when your stack has already been deployed, and you want to deploy a subset of services in
     dev mode without redeploying any service dependencies that may have changed since you last deployed.
     `,
-    alias: "no-deps",
+    alias: "nodeps",
   }),
   "forward": new BooleanParameter({
     help: deline`Create port forwards and leave process running without watching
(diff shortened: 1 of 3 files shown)


thsig (approved):

vvagaytsev:
Hmm... I've just noticed that in the generated reference docs we still have invalid single-dashed alias like `-nodeps`. In the `help` command, everything is ok, it was fixed in https://github.com/garden-io/garden/commit/51ab495af336a215e5ad4f059c5e8da7ae973b19

This looks like a general issue for all multi-char aliases, I'll fix that in a separate PR.

PR #3281, opened 2022-10-02
improvement(cli): improve header/footer text when connected to Cloud

eysi09:
Update header/footer text

* Remove Cloud link from header, it was too noisy
* Always show footer link, also for non-persistent commands
* Update text
* Update link to session

Screenshot 2022-10-02 at 15 15 29

Fixes #


eysi09:
@hph, assigned you as well to double check that the URL makes sense.

thsig (approved):

PR #3293, opened 2022-10-10
improvement(container): configurable deployment strategy 

### LINKED ISSUE #3237: [FEATURE]:  deploymentStrategy. should be available in container module
### Background / Motivation

We have a one single-replica container that we deploy via the garden container module that has attached PVCs.  We would like this  deployed with the `recreate` strategy to minimize reattachment delays when we redeploy it. While most other deployments in our project use helm, we still have a few that we would like to simply use rolling update to minimize downtime.  

Garden allows deploymentStrategy to be specified at the project level, but not at the module level, where it arguably makes more sense. Thus ALL of our garden container modules end up having to use the `recreate` strategy

Specify a deployment strategy under the container/service section. For example:

```
services:
  - name: postgres
    replicas: 1
    deploymentStrategy: recreate
```

Keeps us from using recreate as the default deployment strategy.

See above

We can work around this by using helm or k8s manifests for deployment,  but we *do* like using container modules for simple deploys.

vvagaytsev:
This PR makes the deployment strategy configurable for `container` module type.

Fixes #3237

Please check my comments in the change set.

### DIFF
--- a/core/src/plugins/container/config.ts
+++ b/core/src/plugins/container/config.ts
@@ -111,6 +111,10 @@ interface Annotations {
   [name: string]: string
 }
 
+const deploymentStrategies = ["RollingUpdate", "Recreate"] as const
+export type DeploymentStrategy = typeof deploymentStrategies[number]
+export const defaultDeploymentStrategy: DeploymentStrategy = "RollingUpdate"
+
 export interface ContainerServiceSpec extends CommonServiceSpec {
   annotations: Annotations
   command?: string[]
@@ -134,6 +138,7 @@ export interface ContainerServiceSpec extends CommonServiceSpec {
   tty?: boolean
   addCapabilities?: string[]
   dropCapabilities?: string[]
+  deploymentStrategy: DeploymentStrategy
 }
 
 export const commandExample = ["/bin/sh", "-c"]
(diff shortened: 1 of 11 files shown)


core/src/plugins/container/config.ts
```diff
@@ -742,6 +747,11 @@ const containerServiceSchema = () =>
       ),
     addCapabilities: containerAddCapabilitiesSchema("service"),
     dropCapabilities: containerDropCapabilitiesSchema("service"),
+    deploymentStrategy: joi
```
vvagaytsev, line 750: The compatibility is ensured by the default value here.
The default value is `"RollingUpdate"`.
Validation ensures informative error messages.

core/src/plugins/container/config.ts
```diff
@@ -134,6 +138,7 @@ export interface ContainerServiceSpec extends CommonServiceSpec {
   tty?: boolean
   addCapabilities?: string[]
   dropCapabilities?: string[]
+  deploymentStrategy: DeploymentStrategy
```
vvagaytsev, line 141: This field is not optional, but the changes are backward compatible. See the [comment below](https://github.com/garden-io/garden/pull/3293/commits/58e151bc2332ca3fefa5eeb620c61094d2dca6fb#r991278796).

vvagaytsev:
Converted to draft, there are some test failures.

PR #3291, opened 2022-10-07
SGV2 - Fix more actions tests 

Orzelius:
nearly half way there

PR #3296, opened 2022-10-11
SGv2: Fix actions tests and do some cleaning

Orzelius:
* move main action tests to separate files
* create helpers to get test plugin and test data
* fix a couple of tests for good measure

This comment by @vvagaytsev  is also addressed via the logic separation here https://github.com/garden-io/garden/pull/3291#discussion_r991423995

The only test failing under the new /router dir is the on with the graph solver error we paired on a while back, everything else is green.

Next steps:
* extracts module and environment actions into separate files
* basically rewrite all router and handler tests from scratch. The current ones assume logic that is too different or no longer exists


core/test/unit/src/router/build.ts
```diff
@@ -0,0 +1,127 @@
...
+    const data = await getRouterTestData()
+    garden = data.garden
+    graph = data.graph
+    log = data.log
+    actionRouter = data.actionRouter
+    returnWrongOutputsCfgKey = data.returnWrongOutputsCfgKey
+    resolvedBuildAction = data.resolvedBuildAction
+    module = data.module
+    dateUsedForCompleted = data.dateUsedForCompleted
+  })
```
Orzelius, line 38: If anyone knows how to do this declaration without duplicates like this I'm all ears

vvagaytsev (approved):
LGTM 🚀

PR #3302, opened 2022-10-12
feat(exec): live log streaming for exec modules

thsig:

We now stream build, deploy, test, task and service logs for `exec` modules.

This brings the Cloud logging experience for `exec` modules to full parity with our k8s-based module types.

I've verified that build, test, deploy, task and service log events are emitted, and that they show up in the logging UI in Cloud.

Would be nice to have proper automated testing for the end-to-end flow here, but since this log event emission logic is rarely changed after its initial implementation, it may not be worth the effort right now.


eysi09 (approved):
LGTM!

PR #3309, opened 2022-10-18
improvement(logs): run and test commands error handling

### LINKED ISSUE #3297: [FEATURE]: Different exit codes for different failure modes in Garden commands
### Background / Motivation
We're using Garden for CI and inner loop development, so we utilize the `garden test` and the `garden deploy` commands pretty heavily. There are a few reasons that these commands might fail:
- The tests we wanted to run failed
- The pods took too long to deploy
- The Garden namespace couldn't be created
- Helm chart we use couldn't be pulled
- Some other Kubernetes errors that are not related to our tests - image pull error, context deadline exceeded etc

We would like to be able to automatically retry some of these failures in CI. For example, if it's something unrelated to the test run - the helm chart couldn't be pulled or if the Garden namespace couldn't be created, we'd like to retry twice because we noticed that those are usually flakes.

An easy way to achieve this would be to set retry rules based on the exit code of the Garden commands. Right now the commands would produce an exit code `1` no matter what the failure is, so all those failure modes get lumped together. It would be a huge help for us if say K8s related errors exit at `103`, timeouts exit at `104` and user errors (say tests failed) exit at `1` or something like that.

The user should be able to know the type of error based on the exit code of the Garden command

We'd like to set different auto retry rules for different failure modes. If the Garden command produced different exit codes, auto retry would be very easy to set up for us. We don't want to just retry everything 3 times because the tests take like 20min to run and we don't want flaky tests to make it into the main branch, but if it's a k8s flake then we'd like to retry away and not have our engineers see those errors if possible.

When emitting an error from the Garden command, check what it is (whether it was the command it ran that failed or some k8s failure) and produce a different exit code accordingly. To start we'd be very happy if the commands just produced 2 different errors - command failure vs everything else. More granularity would help but we could get started with just 2.

vvagaytsev:
This PR unifies error handling for `garden run` and `garden test` commands when running those with and without artifacts.

Partially resolves #3297

### DIFF
--- a/core/src/plugins/kubernetes/run.ts
+++ b/core/src/plugins/kubernetes/run.ts
@@ -810,14 +810,6 @@ export class PodRunner extends PodRunnerParams {
     let mainContainerLogs = ""
     const mainContainerName = this.getMainContainerName()
 
-    const getDebugLogs = async () => {
-      try {
-        return this.getMainContainerLogs()
-      } catch (err) {
-        return ""
-      }
-    }
-
     const stream = new Stream<RunLogEntry>()
     void stream.forEach((entry) => {
       const { msg, timestamp } = entry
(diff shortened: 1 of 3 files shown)


Orzelius (approved):
Looks good. I very much thank you for the pristine commits

PR #3310, opened 2022-10-19
fix(cloud): fix duplicate event emission

thsig:

**fix(cloud): fix duplicate event emission**

Before this fix, we were sending each event twice to websocket subscribers in live Cloud sessions.

This is because the `ignoreHost` logic in `DashboardEventStream` expected a `serverHost` without an auth token appended, which was broken when the dashboard auth tokens were introduced a while back. The fix was to simply compare the base server URLs.

**chore(cloud): don't emit client requests as events**

Previously, we were emitting build, test, deploy and task requests on the event bus and handling    them in `processModules`.

This was unnecessarily complicated, and resulted in requests from Cloud being emitted back to it as events (which wasn't the desired behavior for our event logging).

Fixes duplicate event logging in the Cloud UI.

Both the dashboard and Cloud's logging interface (also when used with interactivity) seem to be working as desired after these changes.


core/src/server/client-router.ts
```diff
@@ -0,0 +1,261 @@
...
+  constructor(garden: Garden, log: LogEntry) {
+    this.garden = garden
+    this.log = log
+  }
+
+  async dispatch<T extends ClientRequestType>(requestType: T, requestParams: ClientRequests[keyof ClientRequests]) {
+    // TODO: Once Cloud request types have been renamed from "buildRequested" etc. to "build", remove the cases
+    // handling the suffixed types from this switch statement.
+    switch (requestType) {
+      case "build":
```
thsig, line 33: We should get rid of `buildRequested` etc. here eventually—that will require changes on the Cloud side too, though.

So I'm keeping both names (`build` + `buildRequested` etc.) here for compatibility until we've implemented & rolled this out on the Cloud side.

core/src/tasks/test.ts
```diff
@@ -204,6 +205,9 @@ export class TestTask extends BaseTask {
     return result
   }
 
+  // We memoize this method for performance and to avoid emitting two `testStatus` events when the task is added
+  // and processed (since this method is called in `resolveDependencies` and `process`).
+  @pMemoizeDecorator()
```
thsig, line 210: This fixes an issue in Cloud where "test X has already passed" would appear twice when running tests via the command palette (since `getTestResult` was being called in `resolveDependencies` and in `process`, resulting in two test status events being emitted).

This only happens when using the command palette, since we call `garden.clearCaches()` in the Cloud request handlers, which clears the task graph cache (that cache prevents the event from getting emitted twice).

Note that we're doing away with a lot of these extra caching layers in 0.13 anyway.

core/src/server/client-router.ts
```diff
@@ -0,0 +1,261 @@
...
+      case "deployRequested":
+        return await this.deploy(requestParams as ClientRequests["deploy"])
+      case "test":
+      case "testRequested":
+        return await this.test(requestParams as ClientRequests["test"])
+      case "task":
+      case "taskRequested":
+        return await this.run(requestParams as ClientRequests["task"])
+      default:
+        // This will result in a type error if one or more event types is forgotten above.
```
eysi09, line 46: Nice.

eysi09 (changes requested):
Awesome! Thanks so much for looking into this.

core/src/events.ts
```diff
@@ -282,56 +282,13 @@ export interface Events extends LoggerEvents {
   }
 
   // Cloud UI events
```
vvagaytsev, line 284:
```suggestion
```

vvagaytsev: This looks like a dangling comment that can be removed.

core/src/server/server.ts
```diff
@@ -352,17 +356,15 @@ export class GardenServer {
         isAlive = true
       })
 
-      // Pipe everything from the event bus to the socket, as well as from the /events endpoint
+      // Pipe everything from the event bus to the socket.
```
vvagaytsev, line 359: How do we define listeners for `this.incomingEvents` now? I can see that we emit some events to that bus, but it looks like we do not have any listeners defined.

vvagaytsev: The `ClientRouter` is a functional replacement for the `incomingEvents` event bus, isn't it? The `ClientRouter` is called from `addWebsocketEndpoint()` method which, in turn, is called from `/events` endpoint handler.

It looks like `this.incomingEvents` bus can be safely removed.

core/src/server/server.ts
```diff
@@ -147,6 +150,7 @@ export class GardenServer {
 
     this.garden = garden
     this.garden.log = this.debugLog
+    this.clientRouter = new ClientRouter(this.garden, this.log)
```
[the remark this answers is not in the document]
vvagaytsev: Sounds reasonable. This looks a bit like [feature envy](https://refactoring.guru/smells/feature-envy). Otoh, we have the Cloud requests handling logic placed in a separate component. In some sense that fits the SRP over low-granular composition.

We can leave it as it is now, but attempt to do a bigger refactoring later. I'd like to submit a separate PR for that with some ideas.

There are more places to be refactored in the `GardenServer` class itself. Some methods seem to be too long and hard to read. We can decompose it by splitting the effective request-handling logic and the app/endpoint creation machinery.

vvagaytsev (dismissed):
Looks good 👍 I left a few comments, please check :) Let's try to get it merged by Monday. it will be included in `0.12.46` then.

eysi09 (dismissed):

thsig:
We noticed a regression in one place, marking this as a draft until I've had time to look into it.

core/src/server/server.ts
eysi09: I think we actually still need the incoming events bus to handle events from other Garden processes.

eysi09 (approved):

PR #3316, opened 2022-10-21
Fix the last of the action tests

Orzelius:
The final pr to get all the action tests fixed. A few of the ones that fail I pointed out below and I'd appreciate help with them.


core/test/unit/src/router/base.ts
```diff
@@ -350,4 +340,93 @@ describe("BaseActionRouter", () => {
...
+
+    it("should call the handler with a base argument if the handler is overriding another", async () => {
+      const { router } = await createTestRouter(testPlugins, garden)
+
+      const result = await router.callHandler({
+        handlerType: "getStatus",
+        params: { graph, log, action: resolvedBuildAction },
+      })
+
+      expect(result.outputs.base).to.not.equal(undefined)
```
Orzelius, line 401: This fails. I don't know why. The base action type is properly defined in [the test-plugin-a](https://github.com/garden-io/garden/blob/f955c4e552a5a8ceaa5cc784ea124f8e558f8ce1/core/test/unit/src/router/_helpers.ts#L399).

Orzelius:
Also, [this test fails](https://github.com/garden-io/garden/blob/f955c4e552a5a8ceaa5cc784ea124f8e558f8ce1/core/test/unit/src/router/deploy.ts#L186). It seems like a bug deep within the solver and I was scared to dive in too deep.

Error:
> Aborted process test Deploy service-a(from module service-a) because dependency (process test Build module-a(from module module-a)) failed due to an error: GraphResults object does not have task resolve-action.module-a.
This is a bug. Please report it!.

Orzelius:
The failing tests can be fixed later. I will merge this so we can get `graph-v2` into `0.13` asap

vvagaytsev (approved):

PR #3319, opened 2022-10-21
Graph v2 use exec handlers for test plugin + some Build command test fixes

vvagaytsev:

Fixes #

### DIFF
--- a/core/test/helpers.ts
+++ b/core/test/helpers.ts
@@ -8,7 +8,7 @@
 
 import td from "testdouble"
 import { join, relative, resolve } from "path"
-import { cloneDeep, extend, intersection, mapValues, merge, pick } from "lodash"
+import { cloneDeep, extend, intersection, mapValues, merge, omit, pick } from "lodash"
 import { copy, ensureDir, mkdirp, pathExists, remove, truncate } from "fs-extra"
 
 import {
@@ -486,7 +486,7 @@ export function stubRouterAction<K extends ActionKind, H extends keyof WrappedAc
 }
 
 export function taskResultOutputs(results: ProcessCommandResult) {
-  return mapValues(results.graphResults, (r) => r && r.result)
+  return mapValues(results.graphResults, (r) => r?.result && omit(r.result, "executedAction"))
 }
 
 export const cleanProject = async (gardenDirPath: string) => {
--- a/core/test/helpers.ts
+++ b/core/test/helpers.ts
@@ -12,12 +12,11 @@ import { cloneDeep, extend, intersection, mapValues, merge, pick } from "lodash"
 import { copy, ensureDir, mkdirp, pathExists, remove, truncate } from "fs-extra"
 
 import {
-  containerDeploySchema,
   containerModuleSpecSchema,
   containerModuleTestSchema,
   containerTaskSchema,
 } from "../src/plugins/container/moduleConfig"
-import { buildExecAction, prepareExecBuildAction } from "../src/plugins/exec/exec"
+import { buildExecAction, convertExecModule } from "../src/plugins/exec/exec"
 import { joi, joiArray } from "../src/config/common"
 import { createGardenPlugin, GardenPluginSpec, ProviderHandlers, RegisterPluginParam } from "../src/plugin/plugin"
 import { Garden, GardenOpts } from "../src/garden"
@@ -42,10 +41,10 @@ import { ConfigurationError } from "../src/exceptions"
 import Bluebird = require("bluebird")
 import execa = require("execa")
 import timekeeper = require("timekeeper")
-import { execBuildSpecSchema } from "../src/plugins/exec/moduleConfig"
+import { execBuildSpecSchema, ExecModule } from "../src/plugins/exec/moduleConfig"
 import {
-  ExecActionConfig,
   execBuildActionSchema,
+  execDeployActionSchema,
   ExecRun,
   execRunActionSchema,
   ExecTest,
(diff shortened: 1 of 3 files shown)


core/test/helpers.ts
```diff
@@ -282,24 +277,26 @@ export const testPlugin = () =>
...
-            return {
-              group: {
-                // This is an annoying TypeScript limitation :P
-                kind: <"Group">"Group",
-                name: module.name,
-                path: module.path,
-                actions,
-              },
-            }
+            // const execBuildAction = prepareExecBuildAction(params)
```
thsig, line 285: Delete commented-out stuff here.

core/test/unit/src/commands/build.ts
```diff
@@ -18,14 +18,21 @@ import {
...
 import { LogEntry } from "../../../../src/logger/log-entry"
 import { writeFile } from "fs-extra"
 import { join } from "path"
+import { ProcessCommandResult } from "../../../../src/commands/base"
 
 // TODO-G2: rename test cases to match the new graph model semantics
 describe("BuildCommand", () => {
+  function getBuildModuleVersion(result: ProcessCommandResult, moduleName: string) {
+    const buildActionResults = result!.graphResults
+    const buildModuleResult = buildActionResults[`build.${moduleName}`]
```
thsig, line 32: Maybe there's a helper somewhere that we can use instead of `build.${moduleName}`? No big deal if not, imho.

core/test/unit/src/commands/build.ts
```diff
@@ -43,43 +50,39 @@ describe("BuildCommand", () => {
...
-    expect(buildActionResults).to.eql({
-      "module-a": {
-        fresh: true,
-        buildLog: "A",
-        aborted: false,
-        durationMsec: 0,
-        error: undefined,
-        success: true,
-        version: modules["module-a"].version.versionString,
+    // TODO-G2: need to check if results.graphResults["build.module-a"]?.outputs still makes any sense.
```
thsig, line 53: It does still make sense, but only some plugins make use of it (so I think we can delete this comment).

The `exec` plugin's build action's output type is `{}` (and that's the type of this test case's build action). See:

https://github.com/garden-io/garden/blob/graph-v2/core/src/plugins/exec/config.ts#L30

In contrast, the output type of `container` builds has a bunch of fields:

https://github.com/garden-io/garden/blob/graph-v2/core/src/plugins/container/config.ts#L871-L883

These outputs are available in template strings via `ActionResultContext`, similarly to how module outputs were used in 0.12 for stuff like this:

https://github.com/garden-io/garden/blob/main/examples/vote-helm/vote/garden.yml#L17

thsig (changes requested):
Looks good! Had a few suggestions there.

core/test/helpers.ts
```diff
@@ -282,24 +277,26 @@ export const testPlugin = () =>
         needsBuild: true,
         handlers: {
           convert: async (params: ConvertModuleParams) => {
-            const module = params.module
+            const module: ExecModule = params.module
+            return convertExecModule({ ...params, module })
+
             // We want the build action from the exec conversion,
             // and all the other actions from the container conversion.
```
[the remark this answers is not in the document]
vvagaytsev: I updated the comments in the rewritten commit https://github.com/garden-io/garden/pull/3319/commits/bf00343e01547f04b5e9e382e40e2cd7205a1116

core/test/helpers.ts
vvagaytsev: Done in https://github.com/garden-io/garden/pull/3319/commits/bf00343e01547f04b5e9e382e40e2cd7205a1116

core/test/unit/src/commands/build.ts
vvagaytsev: Good point, done in https://github.com/garden-io/garden/pull/3319/commits/13baeee7b22d6b1bae7c9c68a7f87494f96c7759

core/test/unit/src/commands/build.ts
```diff
@@ -43,43 +50,39 @@ describe("BuildCommand", () => {
...
     })
+
+    const graph = await garden.getConfigGraph({ log, emit: false })
+    const modules = keyBy(graph.getModules(), "name")
+
+    const buildModuleAVersion = getBuildModuleVersion(result!, "module-a")
+    const buildModuleBVersion = getBuildModuleVersion(result!, "module-b")
+    const buildModuleCVersion = getBuildModuleVersion(result!, "module-c")
+
+    // TODO-G2: fix versions
```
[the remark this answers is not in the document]
vvagaytsev: I've updated this comment in https://github.com/garden-io/garden/pull/3319/commits/2bebf83911b9541e0c1c501b6d98bd73661860b4. The version assertions still fail, I'm wondering if it's a bug or some new semantics. I need to dig into that, will do it in the next iteration.

core/test/unit/src/commands/build.ts
vvagaytsev: Good to know, thank you! I updated the TODO-comments in https://github.com/garden-io/garden/pull/3319/commits/2bebf83911b9541e0c1c501b6d98bd73661860b4

vvagaytsev: And how about outputs from `result`? See `results.graphResults["build.module-a"]?.result?.outputs` - could it be different from `results.graphResults["build.module-a"]?.outputs`?

thsig: The `GraphResult.outputs` field is always initialized from `GraphResult.result.outputs`. It has the same type:

https://github.com/garden-io/garden/blob/graph-v2/core/src/graph/results.ts#L30

And (as far as I can tell) is only initialized here:

https://github.com/garden-io/garden/blob/graph-v2/core/src/graph/nodes.ts#L133

I think the reason this field is here is simply to avoid the awkwardness of `result.result.outputs` when getting at the outputs from a `GraphResult` object.

Maybe @edvald can confirm that was the intent behind copying/duplicating the data here?

thsig (approved):

PR #3323, opened 2022-10-26
SGV2 - Fix publish tests

Orzelius:
Fixes all but one tests for the publish command. The one remaining tests fails due to I bug in the solver I think?

Currently no builds are being triggered by the publish task. I'm not sure if that's a deliberate change or not.

Also removed build metadata logic.


Orzelius:
closing in favor of #3328 because graph-v2 was rebased

PR #3345, opened 2022-11-11
SGV2: fixed single test for build command

vvagaytsev:
Fixed assertions for module versions.

Fixes #

### DIFF
--- a/core/test/unit/src/commands/build.ts
+++ b/core/test/unit/src/commands/build.ts
@@ -18,7 +18,6 @@ import {
   withDefaultGlobalOpts,
 } from "../../../helpers"
 import { taskResultOutputs } from "../../../helpers"
-import { keyBy, omit } from "lodash"
 import { ModuleConfig } from "../../../../src/config/module"
 import { LogEntry } from "../../../../src/logger/log-entry"
 import { writeFile } from "fs-extra"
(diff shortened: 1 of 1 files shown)


Orzelius (commented):
why not just get the builds from the graph and then use the `moduleVersion` method? This seems kinda hacky for no reason

vvagaytsev:
> why not just get the builds from the graph and then use the `moduleVersion` method? This seems kinda hacky for no reason

Good point, thanks! Simplified the code in https://github.com/garden-io/garden/pull/3345/commits/f5247f830bbcd24c110f321ad8501e82cceaa507

Orzelius (approved):
Not quite what I meant but this works as well :smiley:

vvagaytsev:
> Not quite what I meant but this works as well 😃

And what did you mean? In-place comments in the code could be more helpful to understand the intent :) If there is a better way to do that, let's just do it!

Orzelius:
> And what did you mean? In-place comments in the code could be more helpful to understand the intent :) If there is a better way to do that, let's just do it!

It's my bad, I took a closer look and now realize this is the way to go. At first I thought why not just get the actions and modules from the graph and compare the versions, but I now realize that would not test the task output results.

PR #3346, opened 2022-11-13
chore: re-enable edge releases

thsig:

We disabled edge releases on merges to `main` back in September, since we were concerned about notification spam. This seems to have been unnecessary, so here we re-enable edge releases.


vvagaytsev (approved):

Orzelius (approved):

stefreak (approved):
👍

PR #3352, opened 2022-11-16
feat: allow to set no resource limits

### LINKED ISSUE #3337: [FEATURE]: Allow to not have have CPU/Memory Resource Limits
### Background / Motivation

We use garden to deploy a number of micro-services into a dev environment dedicated to a developer.  We have control over the resourced dedicated to the environment at large.  Garden will assign default CPU/Memory limits and will not allow to explicitly specify 0/null limits.  However, there is no need for CPU limits at all in that environment and we would also like to remove memory limits to enable to deploy more services without having to manually manage the limits.  For us. it is ok to oversubscribe on memory and only ever apply limits for missbehaving services.

Be able to pass in null or 0 into [CPU limits](https://docs.garden.io/reference/module-types/container#services-.cpu.max) and [memory limits](https://docs.garden.io/reference/module-types/container#services-.memory.max)

We don't want to pay double for larger instances in AWS or invest engineering time to tune k8s resource limits for a development environment.

Either have a project/container module level setting to disable default limits or allow to pass in 0/null values into existing configurations referenced above.

Orzelius:
closes #3337

If the max value of memory or cpu under the container service is set to
null, no resource limit will be set. If they are NOT set to anything at
all (undefined), a default value is used like before.

### DIFF
--- a/core/src/plugins/container/config.ts
+++ b/core/src/plugins/container/config.ts
@@ -539,7 +539,8 @@ export const containerCpuSchema = (targetType: string) =>
         The minimum amount of CPU the ${targetType} needs to be available for it to be deployed, in millicpus
         (i.e. 1000 = 1 CPU)
       `),
-    max: joi.number().default(defaultContainerResources.cpu.max).min(10).allow(null).description(deline`
+    max: joi.number().default(defaultContainerResources.cpu.max).min(defaultContainerResources.cpu.min).allow(null)
+      .description(deline`
         The maximum amount of CPU the ${targetType} can use, in millicpus (i.e. 1000 = 1 CPU).
         If set to null will result in no limit being set.
       `),
--- a/core/src/plugins/container/config.ts
+++ b/core/src/plugins/container/config.ts
@@ -99,11 +99,11 @@ export interface ServiceLimitSpec {
 export interface ContainerResourcesSpec {
   cpu: {
     min: number
-    max: number
+    max: number | null
   }
   memory: {
     min: number
-    max: number
+    max: number | null
   }
 }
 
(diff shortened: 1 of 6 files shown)


core/test/unit/src/plugins/kubernetes/container/util.ts
```diff
@@ -0,0 +1,61 @@
...
+        cpu: "1m",
+        memory: "1Mi",
+      },
+      limits: {
+        cpu: "1m",
+      },
+    })
+  })
+
+  it("shoudl prioritize deprecated limits param", () => {
```
vvagaytsev, line 47: Typo: shoudl -> should

vvagaytsev (changes requested):
Looks good! I've left a couple of minor comments, please check.

core/src/plugins/container/config.ts
```diff
@@ -536,14 +536,13 @@ const limitsSchema = () =>
...
-        `),
-    max: joi
-      .number()
-      .default(defaultContainerResources.cpu.max)
-      .min(10)
-      .description(`The maximum amount of CPU the ${targetType} can use, in millicpus (i.e. 1000 = 1 CPU)`),
+        The minimum amount of CPU the ${targetType} needs to be available for it to be deployed, in millicpus
+        (i.e. 1000 = 1 CPU)
+      `),
+    max: joi.number().default(defaultContainerResources.cpu.max).min(10).allow(null).description(deline`
```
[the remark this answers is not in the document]
Orzelius: nice catch [cdd7b17](https://github.com/garden-io/garden/pull/3352/commits/cdd7b17a0f2069142583bc6507d59afe38345e7e)

core/test/unit/src/plugins/kubernetes/container/util.ts
Orzelius: yeah [616f99b](https://github.com/garden-io/garden/pull/3352/commits/616f99b77bc73d9dd682f517492da69ae9b3477c)

vvagaytsev (approved):

PR #3354, opened 2022-11-16
fix: fixed autoResolveIncludes for multi-stage dockerfiles

ivanbabenko:

Fixes bug in auto-resolve logic for includes running on multi-stage Dockerfiles that has multiple arguments for COPY command.

`autoResolveIncludes` logic was relying on having just one argument in ADD or COPY commands, and therefore filtering was done based only on the first argument (args[0]).

However, there are cases when both `--chown` and `--from` arguments can be used for COPY command (in multi-stage builds), and in such case Garden would handle includes successfully only if `--from` was the first argument. For cases when it's not, e.g. `COPY --chown=bla --from=bla file-b file-c`, existing logic would try to use `--from=bla` as an include parameter and fail with a cryptic message:
```
Error validating configure container output from provider kubernetes: 
key .moduleConfig.include[2] must be a relative sub-path (may not contain '..' segments or be an absolute path)
```

### DIFF
--- a/core/src/plugins/container/helpers.ts
+++ b/core/src/plugins/container/helpers.ts
@@ -372,12 +372,12 @@ const helpers = {
       flatten(
         commands.map((cmd) => {
           const args = cmd.args as string[]
-          if (args[0].startsWith("--chown")) {
-            // Ignore --chown args
-            return args.slice(1, -1)
-          } else if (args[0].startsWith("--from")) {
+          if (args.find((arg) => arg.startsWith("--from"))) {
             // Skip statements copying from another build stage
             return []
+          } else if (args[0].startsWith("--chown")) {
+            // Ignore --chown args
+            return args.slice(1, -1)
           } else {
             return args.slice(0, -1)
           }
--- a/core/test/unit/src/plugins/container/helpers.ts
+++ b/core/test/unit/src/plugins/container/helpers.ts
@@ -543,6 +543,19 @@ describe("containerHelpers", () => {
       expect(await helpers.autoResolveIncludes(config, log)).to.eql(["file-a", "Dockerfile"])
     })
 
+    it("should handle COPY statements with multiple arguments in any order", async () => {
+      await writeFile(
+        dockerfilePath,
+        dedent`
+        FROM foo
+        ADD --chown=bla file-a /
+        COPY --chown=bla --from=bla /file-b file-c /
+        COPY --from=bla --chown=bla  /file-c file-d /
+        `
+      )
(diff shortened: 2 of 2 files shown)


vvagaytsev:
Huge thanks for fixing this @ivanbabenko! We'll review it soon!

ivanbabenko:
Thanks for a quick response!
One of integration tests jobs is failing, but I cannot see how this can be related to my changes. Please take a look when you can.

stefreak:
@ivanbabenko thanks for contributing, this looks like a bug fix many users will benefit from. I restarted the integration test, it might have flaked.

Also looking into the code; I'm not very familiar with the docker file parsing code yet so I'll need a couple minutes.

core/src/plugins/container/helpers.ts
```diff
@@ -372,12 +372,12 @@ const helpers = {
...
         commands.map((cmd) => {
           const args = cmd.args as string[]
-          if (args[0].startsWith("--chown")) {
-            // Ignore --chown args
-            return args.slice(1, -1)
-          } else if (args[0].startsWith("--from")) {
+          if (args.find((arg) => arg.startsWith("--from"))) {
             // Skip statements copying from another build stage
             return []
+          } else if (args[0].startsWith("--chown")) {
```
stefreak, line 378: is there a particular reason why we are looking for `--chown` only at the beginning? I wonder wether we should use `args.find` for both cases

stefreak: I see, we assume that there are no other flags than `--from` and `--chown` in this code.

I think I'll approve this PR for now, we can improve this code in another PR.

ivanbabenko: Hi @stefreak, thanks for the review!
There are of course other arguments that can be used with ADD and COPY commands (e.g. `--link` or `--checksum`, see [Docker documentation](https://docs.docker.com/engine/reference/builder/#copy)) and ideally we should reason about all of them, but they are not that common though will increase logic a lot. My intention was at least to fix it for the current scope of supported arguments.

stefreak (approved):
LGTM, I will follow up on this with another PR later today.

core/src/plugins/container/helpers.ts
stefreak: @ivanbabenko what do you think of my take on generalizing this in https://github.com/garden-io/garden/pull/3359 ?

Do you think there are some flags that we missed, and need to be taken care of?

Thanks for your time on this 👍

ivanbabenko: @stefreak nice, much better solution! Only `--from` requires explicit ignore currently, and I guess now we are safe even from new arguments introduced by Docker and can handle different combinations 👍

PR #3357, opened 2022-11-17
Separate resource and toleration configuration for garden-util pod

### LINKED ISSUE #3197: [FEATURE]: Separate configuration for garden-util pods
### Background / Motivation

Currently, `garden-util` and `kaniko-build` pods share the same tolerations, which means they cannot be scheduled to run on different nodes.

Additionally, there is no way to configure the resource request/limit for the `garden-util` pods and they will use whatever the default values are for the Kubernetes cluster.

See [this forum thread](https://community.garden.io/t/best-way-to-use-garden-util-in-dynamic-ci-environments/191/4?u=masihaz) for additional discussion.

The user should be able to specify separate tolerations and resource requests/limits for `garden-util` pods in the Kubernetes provider config. The resource request/limits can also have a reasonable default (some testing under load might be worthwhile for the ideal amount).

Some teams use a dynamically-scaling nodepool with their cloud provider, and may have additional resources allocated to their build nodes (making them cost more). It would be preferable to be able to limit the pods running on these nodes to the ephemeral build pods and run the `garden-util` pods elsewhere, and also to configure their resource requests/limits. This will allow users to more effectively manage the infrastructure footprint of their builds and minimize costs.

An additional section in the Kubernetes provider config under the `kaniko` settings for the `garden-util` pod resources and tolerations.

Example:
```yaml
providers:
  resources:
    builder:
      limits:
        cpu:
        memory:
        ephemeralStorage:
      requests:
        cpu:
        memory:
        ephemeralStorage:
    util:
      limits:
        cpu:
        memory:
        ephemeralStorage:
      requests:
        cpu:
        memory:
        ephemeralStorage:
  kaniko:
    tolerations:
      builder:
        - effect:
          key:
          operator:
          value:
      util:
        - effect:
          key:
          operator:
          value:
```

Orzelius:
closes #3197

Now custom resource limits and requests and tolerations can be set for the `garden-util` pod. See schema changes in the file diff.

### DIFF
--- a/core/src/plugins/kubernetes/config.ts
+++ b/core/src/plugins/kubernetes/config.ts
@@ -158,17 +158,15 @@ export interface CertManagerConfig {
   acmeServer?: LetsEncryptServerType
 }
 
+export interface KuberetesResourceConfig {
+  cpu: number
+  memory: number
+  ephemeralStorage?: number
+}
+
 export interface KubernetesResourceSpec {
-  limits: {
-    cpu: number
-    memory: number
-    ephemeralStorage?: number
-  }
-  requests: {
-    cpu: number
-    memory: number
-    ephemeralStorage?: number
-  }
+  limits: KuberetesResourceConfig
+  requests: KuberetesResourceConfig
 }
 
 interface KubernetesResources {
--- a/core/src/plugins/kubernetes/config.ts
+++ b/core/src/plugins/kubernetes/config.ts
@@ -175,6 +175,7 @@ interface KubernetesResources {
   builder: KubernetesResourceSpec
   registry: KubernetesResourceSpec
   sync: KubernetesResourceSpec
+  util: KubernetesResourceSpec
 }
 
 interface KubernetesStorageSpec {
@@ -290,6 +291,16 @@ export const defaultResources: KubernetesResources = {
     requests: {
       cpu: 100,
       memory: 90,
+    },
+  },
+  util: {
+    limits: {
+      cpu: 256,
+      memory: 512,
+    },
+    requests: {
+      cpu: 256,
+      memory: 512,
     },
   },
 }
@@ -762,6 +773,12 @@ export const kubernetesConfigBase = () =>
 
             This is shared across all users and builds, so it should be resourced accordingly, factoring
             in how many concurrent builds you expect and how large your images tend to be.
+          `),
+        util: resourceSchema(defaultResources.util, false).description(dedent`
+            Resource requests and limits for the util pod for in-cluster builders.
+            This pod is used to get, start, stop and inquire the status of the builds.
+
+            This pod is created in each garden namespace.
           `),
         sync: resourceSchema(defaultResources.sync, true)
           .description(
--- a/core/src/plugins/kubernetes/config.ts
+++ b/core/src/plugins/kubernetes/config.ts
@@ -229,6 +229,9 @@ export interface KubernetesConfig extends BaseProviderConfig {
     namespace?: string | null
     nodeSelector?: StringMap
     tolerations?: V1Toleration[]
+    util?: {
+      tolerations?: V1Toleration[]
+    }
   }
   context: string
   defaultHostname?: string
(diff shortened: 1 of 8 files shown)


core/src/plugins/kubernetes/config.ts
```diff
@@ -763,6 +783,12 @@ export const kubernetesConfigBase = () =>
             This is shared across all users and builds, so it should be resourced accordingly, factoring
             in how many concurrent builds you expect and how large your images tend to be.
           `),
+        util: resourceSchema(defaultResources.util, false).description(dedent`
```
stefreak, line 786: joi allows to pass a function to the `default` method with a `parent` argument like this:

```
   .default((parent) => parent.kaniko)
```

Here it seems to be complicated as we use the `resourceSchema` function, but in general this can be useful to avoid if statements later in the code.
```

Comment (Orzelius) [reply core/src/plugins/kubernetes/config.ts:786]: I think I prefer to keep my schema stupid and account for logic behavior in code. In general that seems to be the pattern in garden.

File: core/src/plugins/kubernetes/container/build/common.ts [hunk 2]
```diff
@@ -506,10 +506,19 @@ export function getUtilContainer(authSecretName: string): V1Container {
       },
     },
     resources: {
-      // This should be ample
       limits: {
-        cpu: "256m",
-        memory: "512Mi",
+        cpu: millicpuToString(provider.config.resources.util.limits.cpu),
```
stefreak, line 510: maybe transforming the limits / request config to pod spec could be done in a helper function, as it's repeated

core/src/plugins/kubernetes/config.ts
```diff
@@ -292,6 +296,16 @@ export const defaultResources: KubernetesResources = {
       memory: 90,
     },
   },
+  util: {
+    limits: {
+      cpu: 256,
+      memory: 512,
+    },
+    requests: {
+      cpu: 256,
```
stefreak, line 305: in the old util pod spec, we only specified limits bot no requests. 

Specifying requests leads to scheduling problems for our customers, the scheduler will reserve this amount of cpu and memory for the util pod (that doesn't need much) even when there is ample space available for other services.

stefreak (changes requested):
Everything looks good to me, except for introducing requests for the util pod; I think limits are enough, so we don't hog customer's resources

Great stuff! 👍 🥳

core/src/plugins/kubernetes/config.ts
Orzelius: From the kuberetes doc: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/

> If you specify a limit for a resource, but do not specify any request, and no admission-time mechanism has applied a default request for that resource, then Kubernetes copies the limit you specified and uses it as the requested value for the resource.

so it seems people should just set the request to a low value if they want to save on resources

srihas-g: I agree with @Orzelius if we don't specify `requests` we can't ensure if util pod can even be scheduled or not. Also specifying `requests` will help our service avoid being throttled on CPU since minimum requested CPU is always made sure to be available for our service, for e.g.

stefreak: Ah, I didn't know that, thanks @Orzelius

stefreak (approved):
Thanks for clarifying that when you don't specify a request, kubernetes will use the limits as requests. I learned something today :)

looks good to me

core/test/unit/src/plugins/kubernetes/container/build/common.ts
```diff
@@ -31,4 +36,134 @@ describe("common build", () => {
...
+          spec: {
+            replicas: 1,
+            selector: { matchLabels: { app: "garden-util" } },
+            template: {
+              metadata: { labels: { app: "garden-util" } },
+              spec: {
+                containers: [
+                  {
+                    name: "util",
+                    image: "gardendev/k8s-util:0.5.4",
```
vvagaytsev, line 69: This value can be replaced with the named constant `k8sUtilImageName` from `core/src/plugins/kubernetes/constants.ts`. Otherwise, the test will get broken on every k8s util image version update.

core/src/plugins/kubernetes/container/build/common.ts
Orzelius: Good idea [787cd1a](https://github.com/garden-io/garden/pull/3357/commits/787cd1a8b36f3872f1e10f736afb20531ed9f1c9)

Orzelius: I just read your comment again and realized you probably meant this [88bbd59](https://github.com/garden-io/garden/pull/3357/commits/88bbd590b21b40dcbf1988b0535e6d58c9939bcd), but some extra cleanup never hurts

vvagaytsev (approved):

stefreak (approved):
🚢 👍

srihas-g:
Would we also want to enable `[ephemeral-storage](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#configurations-for-local-ephemeral-storage)` as a variable so we satisfy all existing fields for resource requests/limits? 
I don't really know that we need it for this particular case or not though.

core/src/plugins/kubernetes/container/build/common.ts
stefreak: looks much better now 👍

Orzelius:
Approve plz, I just squashed the fixup commit

vvagaytsev (approved):

PR #3359, opened 2022-11-17
fix(autoResolve): handle unknown Dockerfile flags

stefreak:
This fixes the `autoResolveIncludes` function so it ignores
all unknown ADD/COPY flags. That way users do not receive
strange looking error messages like

```
Error validating configure container output from provider kubernetes:
key .moduleConfig.include[2] must be a relative sub-path (may not contain '..' segments or be an absolute path)
```

This improves further upon the bug fix in #3354

Thanks again to @ivanbabenko to bring this up.

Fixes #


vvagaytsev (approved):
Nice fix, thanks @stefreak! 👍

PR #3365, opened 2022-11-22
feat(k8s): custom annotations for builder pods

### LINKED ISSUE #2628: [FEATURE]: allow setting annotations on garden managed pods
## Feature Request

Allow setting annotations on garden managed pods like build pods (kaniko or cluster-buildkit) and pods used for running tasks in helm, kubernetes and container modules.

### Background / Motivation

When using cluster-autoscaler, all pods without the annotation `cluster-autoscaler.kubernetes.io/safe-to-evict: false` can be evicted in a scale-down and scheduled on a new node. 
For Garden build pods like kaniko or cluster buildkit, as well as for tasks run through container, kubernetes or helm modules it is necessary to set this annotation, as eviction and re-deploying of these pods results in longer waits or even an undefined state (e.g. task is evicted mid-run).

Specify annotations for build containers in the provider configuration and for tasks in the task definition in a module.

Tasks and build pods are not getting evicted by scale-down operations from cluster-autoscaler, as well as other possible use-cases for annotations on pods.

Specify annotations for build containers in the respective provider configuration e.g.:

```
providers:
  - name: kubernetes
    # Use the kaniko build mode
    buildMode: kaniko
    kaniko:
      namespace: null
      annotations:
          cluster-autoscaler.kubernetes.io/safe-to-evict: false
```

And for tasks:

```
tasks:
  - name: echo-task
    command: [/bin/sh, -c]
    args: ["echo hello"]
    annotations: 
        cluster-autoscaler.kubernetes.io/safe-to-evict: false
```

stefreak:
Allow specifying custom annotations for garden-
managed pods when using in-cluster building with
buildkit or kaniko.

This is useful when users want to alter how certain
components, like the cluster-autoscaler treat these
pods.

Fixes #2628

### DIFF
--- a/core/src/plugins/kubernetes/config.ts
+++ b/core/src/plugins/kubernetes/config.ts
@@ -707,23 +707,23 @@ export const kubernetesConfigBase = () =>
           dedent`
             Exposes the \`nodeSelector\` field on the PodSpec of the Kaniko pods. This allows you to constrain the Kaniko pods to only run on particular nodes.
 
-            [See here](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/) for the official Kubernetes guide to assigning Pods to nodes.
+            [See here](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/) for the official Kubernetes guide to assigning pods to nodes.
           `
         ),
         tolerations: joiSparseArray(tolerationSchema()).description(
-          deline`Specify tolerations to apply to each Kaniko builder Pod. Useful to control which nodes in a cluster can run builds.
+          deline`Specify tolerations to apply to each Kaniko builder pod. Useful to control which nodes in a cluster can run builds.
           Same tolerations will be used for the util pod unless they are specifically set under \`util.tolerations\``
         ),
         annotations: annotationsSchema().description(
-          deline`Specify annotations to apply to each Kaniko builder Pod. Annotations may have an effect on the behaviour of certain components, for example autoscalers.
-          Same anotations will be used for the util pod unless they are specifically set under \`util.annotations\``
+          deline`Specify annotations to apply to each Kaniko builder pod. Annotations may have an effect on the behaviour of certain components, for example autoscalers.
+          Same anotations will be used for each util pod unless they are specifically set under \`util.annotations\``
         ),
         util: joi.object().keys({
           tolerations: joiSparseArray(tolerationSchema()).description(
-            "Specify tolerations to apply to the garden-util Pod."
+            "Specify tolerations to apply to each garden-util pod."
           ),
           annotations: annotationsSchema().description(
-            "Specify annotations to apply to the garden-util Pod and Deployment."
+            "Specify annotations to apply to each garden-util pod and deployments."
           ),
         }),
       })
--- a/core/src/plugins/kubernetes/config.ts
+++ b/core/src/plugins/kubernetes/config.ts
@@ -214,6 +214,7 @@ export interface KubernetesConfig extends BaseProviderConfig {
     rootless?: boolean
     nodeSelector?: StringMap
     tolerations?: V1Toleration[]
+    annotations?: StringMap
   }
   clusterDocker?: {
     enableBuildKit?: boolean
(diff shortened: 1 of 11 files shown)


stefreak:
@Orzelius I have no clue why the circleci tests did not run on this PR 🤔

core/test/unit/src/plugins/kubernetes/container/build/buildkit.ts
```diff
@@ -6,67 +6,340 @@
...
+    let provider = _provider as KubernetesProvider
+    beforeEach(() => {
+      provider = _provider as KubernetesProvider
+    })
 
-    module.spec.build.targetImage = "foo"
+    it("should return a Kubernetes Deployment manifest for buildkit in-cluster-builder", () => {
+      const result = getBuildkitDeployment(provider, "authSecretName", [{ name: "imagePullSecretName" }])
+      expect(result.kind).eql("Deployment")
+      expect(result.metadata).eql({
```
stefreak, line 40: I split this into separate `expect()`s because otherwise the diffs would be useless because of https://stackoverflow.com/questions/70962366/with-mocha-chai-how-can-i-make-it-show-lines-skipped

core/test/unit/src/plugins/kubernetes/container/build/kaniko.ts
```diff
@@ -60,6 +67,145 @@ describe("kaniko build", () => {
...
+          ],
+          imagePullSecrets: [],
+          initContainers: [
+            {
+              command: [
+                "/bin/sh",
+                "-c",
+                'echo "Copying from sourceURL to /.garden/context"\nmkdir -p /.garden/context\nn=0\nuntil [ "$n" -ge 30 ]\ndo\n  rsync arg1 arg2 && break\n  n=$((n+1))\n  sleep 1\ndone\necho "Done!"',
+              ],
+              image: "gardendev/k8s-util:0.5.4",
```
stefreak, line 141: need to reference the variable here

core/src/plugins/kubernetes/container/build/kaniko.ts
```diff
@@ -395,6 +358,81 @@ async function runKaniko({
     spec,
   }
 
+  return pod
+}
+
+interface RunKanikoParams {
```
Orzelius, line 364: I think if the interface is not used more than once it doesn't deserve to be declared with a name and should be inlined instead like this
```ts
someFunc({
  garden,
  log,
  ...
}: {
  garden: Garden
  log: LogEntry
  ...
}) {...}
```

core/test/unit/src/plugins/kubernetes/container/build/kaniko.ts
stefreak: fixed that :)

Orzelius (commented):
Left a few comments. Good job on the tests!

core/src/plugins/kubernetes/container/build/kaniko.ts
Orzelius: Same thing here https://github.com/garden-io/garden/pull/3365/files#diff-6c3fdefa4bb9119db673f06fee677b0528c922b94e4fb1eb1562f8645283ca13R214

core/src/plugins/kubernetes/config.ts
```diff
@@ -708,9 +714,16 @@ export const kubernetesConfigBase = () =>
...
           Same tolerations will be used for the util pod unless they are specifically set under \`util.tolerations\``
         ),
+        annotations: annotationsSchema().description(
+          deline`Specify annotations to apply to each Kaniko builder Pod. Annotations may have an effect on the behaviour of certain components, for example autoscalers.
+          Same anotations will be used for the util pod unless they are specifically set under \`util.annotations\``
+        ),
         util: joi.object().keys({
           tolerations: joiSparseArray(tolerationSchema()).description(
-            "Specify tolerations to apply to each garden-util Pod."
+            "Specify tolerations to apply to the garden-util Pod."
```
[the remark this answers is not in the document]
stefreak: Hm, per namespace there is only one; if I run garden deploy it will only change it in one namespace, right? That's why I preferred the signular. 

But I am ok with changing it to the plural, changed it in the latest commit 👍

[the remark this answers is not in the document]
Orzelius: > if I run garden deploy it will only change it in one namespace, right?

yes, but if you change the project config it will change for everyone resulting in changes in all namespaces

Orzelius (dismissed):

Orzelius (approved):

PR #3379, opened 2022-11-24
chore(update): update kubelogin version to v0.0.24

srihas-g:
Helps with bumping the version of `kubelogin` binary used in Azure based images

Fixes #

happy to pair and test locally.


vvagaytsev (commented):
Thanks @srihas-g! Looks good! Are there any possible compatibility issues? SemVer number assume there are not, but let's double check :)

Orzelius (approved):
Yeah I think this approach is better after all as the version won't change from under our feet

PR #3384, opened 2022-11-25
feat(k8s): add service account and irsa support for in-cluster-builder

### LINKED ISSUE #2931: [FEATURE]: Allow custom annotations for image builders so that IRSA can be used
### Background / Motivation

According to the [AWS EKS Best Practices Guide](https://aws.github.io/aws-eks-best-practices/security/docs/iam/#restrict-access-to-the-instance-profile-assigned-to-the-worker-node), when using IRSA you should set the instance metadata service to IMDSv2 and the hop count to 1. This means that pods will no longer be able to talk to the regular instance metadata endpoint and inherit the role of the node.

Under normal conditions, kaniko or cluster-buildkit will just use the default Service Account, and not have any specific IAM role that grants it ECR push permissions, so it will rely on the instance role and ECR push permissions associated with it. If we want to enforce the above security settings, any ECR pushes will fail. Since there's no opportunity to add annotations to these pods, we can't tell IRSA which IAM role the builder should assume, and hence can't give it the right permissions. This will break effectively all modules using in-cluster built images.

Add a configuration in the Garden project configuration associated with the chosen builder service, that adds the K8s annotation to the builder's service account. This probably also means that a build service account needs to be created by Garden.

See the description above.

This could just be added to the existing configuration map that you pass for the builder in the project configuration. I'm not sure how things like for GKE or AKS works with respect to pod roles, so I guess it would be good to have a solution that works equally well for those platforms (although I'm not using them).

It's a non-functional requirement as far as the day to day developer experience goes, but not having this means that any Garden pod can assume the instance IAM role, and use any permissions a given EKS node can use. This is not great for security.

stefreak:
This PR introduces a new service account, and makes the annotations
configurable in kaniko and buildkit in-cluster-builders.

This change enables the use of IRSA for in-cluter-building which makes
in-cluster building more secure.

Fixes #2931

### DIFF
--- a/core/src/plugins/kubernetes/container/build/buildkit.ts
+++ b/core/src/plugins/kubernetes/container/build/buildkit.ts
@@ -223,8 +223,8 @@ export async function ensureBuildkit({
 
     // if the service account changed, all pods part of the deployment must be restarted
     // so that they receive new credentials (e.g. for IRSA)
-    if (status.remoteResources.length && serviceAccountChanged) {
-      await cycleDeployment({ ctx, provider, deployment: manifest, api, namespace, deployLog })
+    if (status.remoteResources.length > 0 && serviceAccountChanged) {
+      await cycleDeployment({ ctx, provider, deployment, api, namespace, deployLog })
     }
 
     if (status.state === "ready") {
--- a/core/src/plugins/kubernetes/container/build/buildkit.ts
+++ b/core/src/plugins/kubernetes/container/build/buildkit.ts
@@ -224,7 +224,7 @@ export async function ensureBuildkit({
     // if the service account changed, all pods part of the deployment must be restarted
     // so that they receive new credentials (e.g. for IRSA)
     if (status.remoteResources.length > 0 && serviceAccountChanged) {
-      await cycleDeployment({ ctx, provider, deployment, api, namespace, deployLog })
+      await cycleDeployment({ ctx, provider, deployment: manifest, api, namespace, deployLog })
     }
 
     if (status.state === "ready") {
--- a/core/src/plugins/kubernetes/container/build/buildkit.ts
+++ b/core/src/plugins/kubernetes/container/build/buildkit.ts
@@ -30,6 +30,9 @@ import {
   getUtilContainer,
   ensureBuilderSecret,
   builderToleration,
+  inClusterBuilderServiceAccount,
+  ensureServiceAccount,
+  cycleDeployment,
 } from "./common.js"
 import { getNamespaceStatus } from "../../namespace.js"
 import { sleep } from "../../../../util/util.js"
@@ -187,6 +190,14 @@ export async function ensureBuildkit({
   api: KubeApi
   namespace: string
 }) {
+  const serviceAccountChanged = await ensureServiceAccount({
+    ctx,
+    log,
+    api,
+    namespace,
+    annotations: provider.config.clusterBuildkit?.serviceAccountAnnotations,
+  })
+
   return deployLock.acquire(namespace, async () => {
     const deployLog = log.createLog()
 
@@ -210,12 +221,18 @@ export async function ensureBuildkit({
       log: deployLog,
     })
 
+    // if the service account changed, all pods part of the deployment must be restarted
+    // so that they receive new credentials (e.g. for IRSA)
+    if (status.remoteResources.length && serviceAccountChanged) {
+      await cycleDeployment({ ctx, provider, deployment: manifest, api, namespace, deployLog })
+    }
+
     if (status.state === "ready") {
       // Need to wait a little to ensure the secret is updated in the deployment
       if (secretUpdated) {
         await sleep(1000)
       }
-      return { authSecret, updated: false }
+      return { authSecret, updated: serviceAccountChanged }
     }
 
     // Deploy the buildkit daemon
@@ -397,6 +414,7 @@ export function getBuildkitDeployment(
           annotations: provider.config.clusterBuildkit?.annotations,
         },
         spec: {
+          serviceAccountName: inClusterBuilderServiceAccount,
           containers: [
             {
               name: buildkitContainerName,
--- a/core/src/plugins/kubernetes/container/build/common.ts
+++ b/core/src/plugins/kubernetes/container/build/common.ts
@@ -365,7 +365,7 @@ export async function ensureUtilDeployment({
 
     // if the service account changed, all pods part of the deployment must be restarted
     // so that they receive new credentials (e.g. for IRSA)
-    if (status.remoteResources.length && serviceAccountChanged) {
+    if (status.remoteResources.length > 0 && serviceAccountChanged) {
       await cycleDeployment({ ctx, provider, deployment, api, namespace, deployLog })
     }
 
--- a/core/src/plugins/kubernetes/container/build/common.ts
+++ b/core/src/plugins/kubernetes/container/build/common.ts
@@ -192,7 +192,7 @@ export async function skopeoBuildStatus({
   const outputs = k8sGetContainerBuildActionOutputs({ action, provider })
 
   const remoteId = outputs.deploymentImageId
-  const skopeoCommand = ["skopeo", "--command-timeout=30s", "inspect", "--raw", "--authfile", "/.docker/config.json"]
+  const skopeoCommand = ["skopeo", "--command-timeout=30s", "inspect", "--raw", "--authfile", "~/.docker/config.json"]
 
   if (deploymentRegistry?.insecure === true) {
     skopeoCommand.push("--tls-verify=false")
@@ -280,6 +280,11 @@ export async function ensureServiceAccount({
 }): Promise<boolean> {
   return deployLock.acquire(namespace, async () => {
     const serviceAccount = getBuilderServiceAccountSpec(annotations)
+
+    // the manifest does not contain the namespace, so we need to set it here
+    // otherwise the comparison with deployed resources returns success if the
+    // serviceAccount exists in any namespace
+    serviceAccount.metadata.namespace = namespace
 
     const status = await compareDeployedResources({
       ctx: ctx as KubernetesPluginContext,
--- a/core/src/plugins/kubernetes/container/build/common.ts
+++ b/core/src/plugins/kubernetes/container/build/common.ts
@@ -290,7 +290,7 @@ export async function ensureServiceAccount({
     })
 
     // NOTE: This is here to make sure that we remove annotations in case they are removed in the garden config.
-    // `compareDeployedResources` as of today only checks wether the manifest is a subset of the deployed manifest.
+    // `compareDeployedResources` as of today only checks whether the manifest is a subset of the deployed manifest.
     // The manifest is still a subset of the deployed manifest, if an annotation has been removed. But we want the
     // annotation to be actually removed.
     // NOTE(steffen): I tried to change the behaviour of `compareDeployedResources` to return "outdated" when the
--- a/core/src/plugins/kubernetes/container/build/common.ts
+++ b/core/src/plugins/kubernetes/container/build/common.ts
@@ -354,7 +354,7 @@ export async function ensureUtilDeployment({
     const imagePullSecrets = await prepareSecrets({ api, namespace, secrets: provider.config.imagePullSecrets, log })
 
     // Check status of the util deployment
-    const { deployment, service } = getUtilManifests(provider, authSecret.metadata.name, imagePullSecrets, namespace)
+    const { deployment, service } = getUtilManifests(provider, authSecret.metadata.name, imagePullSecrets)
     const status = await compareDeployedResources({
       ctx: ctx as KubernetesPluginContext,
       api,
@@ -576,8 +576,7 @@ export function getUtilContainer(authSecretName: string, provider: KubernetesPro
 export function getUtilManifests(
   provider: KubernetesProvider,
   authSecretName: string,
-  imagePullSecrets: { name: string }[],
-  namespace?: string
+  imagePullSecrets: { name: string }[]
 ) {
   const kanikoTolerations = [
     ...(provider.config.kaniko?.util?.tolerations || provider.config.kaniko?.tolerations || []),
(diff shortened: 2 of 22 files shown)


core/src/plugins/kubernetes/container/build/buildkit.ts
```diff
@@ -193,30 +204,43 @@ export async function ensureBuildkit({
...
+      ctx as KubernetesPluginContext,
+      api,
+      namespace,
+      [deployment],
+      deployLog
+    )
+
+    // if the service account changed, all pods part of the deployment must be restarted
+    // so that they receive new credentials (e.g. for IRSA)
+    if (status.remoteResources.length && serviceAccountChanged) {
```
vvagaytsev, line 219: Minor nit-pick :) Here and in [L413](https://github.com/garden-io/garden/pull/3384/files#diff-e7c65544747fe8f12648a5aa212bca83e7ebb6636cefb76985963f26015b035dR413) we use `.length` as a `boolean` condition, and in [L344](https://github.com/garden-io/garden/pull/3384/files#diff-e7c65544747fe8f12648a5aa212bca83e7ebb6636cefb76985963f26015b035dR344) we use explicit `.length > 0`. I think those checks are equivalent, because `0` is falsey in JS/TS and the `length` property cannot be negative.

I'm just wondering what is the most typescriptic way of writing such expressions. Is there any convention or recommended style?

docs/guides/in-cluster-building.md
```diff
@@ -293,55 +293,83 @@ Note: If you're using the [`kaniko`](#kaniko) or [`cluster-docker`](#cluster-doc
 
 ### Using in-cluster building with ECR
 
-For AWS ECR (Elastic Container Registry), you need to enable the ECR credential helper once for the repository by adding an `imagePullSecret` for you ECR repository.
+### Using in-cluster building with ECR
+
+For using Garden with AWS ECR (Elastic Container Registry) we need to
+
+1. Create an IAM policy that allows your Kubernetes worker nodes to pull images from the ECR registry
```
twelvemo, line 300:
```suggestion
1. Create an IAM policy that allows your Kubernetes worker nodes to pull images from the ECR registry. Depending on how you deployed your EKS cluster this might not be necessary, as this policy is attached to the worker nodes by default in some scenarios. For more info please check [here](https://docs.aws.amazon.com/AmazonECR/latest/userguide/ECR_on_EKS.html).
```

vvagaytsev (changes requested):
Thanks, great job! 💪  I've left a few comments and questions, please check.

docs/guides/in-cluster-building.md
```diff
@@ -293,55 +293,83 @@ Note: If you're using the [`kaniko`](#kaniko) or [`cluster-docker`](#cluster-doc
...
+2. Create a [web identity IAM role](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-idp_oidc.html), allowing the `garden-in-cluster-builder` service accounts managed by Garden to push images to the ECR registry
```
twelvemo, line 301: I think we should keep the current documentation on how to add a policy to ECR that allows the worker nodes role push access to ECR. IRSA is another, more secure way of handling this, but the original option is still feasible especially for AWS accounts only used for development. What do you think?

core/src/plugins/kubernetes/container/build/common.ts
```diff
@@ -306,6 +308,59 @@ export async function getUtilDaemonPodRunner({
...
+    )
+
+    // NOTE: This is here to make sure that we remove annotations in case they are removed in the garden config.
+    // `compareDeployedResources` as of today only checks wether the manifest is a subset of the deployed manifest.
+    // The manifest is still a subset of the deployed manifest, if an annotation has been removed. But we want the
+    // annotation to be actually removed.
+    // NOTE(steffen): I tried to change the behaviour of `compareDeployedResources` to return "outdated" when the
+    // annotations have changed. But a lot of code actually depends on the behaviour of it with missing annotations.
+    const annotationsNeedUpdate =
+      status.remoteResources.length > 0 && !isEqualAnnotations(serviceAccount, status.remoteResources[0])
```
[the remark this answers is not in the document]
stefreak: No, it's 0 or 1 (depending on wether this is a fresh deployment or updating an existing one).
We only want to cycle the deployment if it's existing, to get the fresh credentials via IRSA.

docs/guides/in-cluster-building.md
```diff
@@ -293,55 +293,83 @@ Note: If you're using the [`kaniko`](#kaniko) or [`cluster-docker`](#cluster-doc
...
+3. Create a Kubernetes secret of type `kubernetes.io/dockerconfigjson` and reference it in the Garden configuration
 
-First create a `config.json` somewhere with the following contents (`<aws_account_id>` and `<region>` are placeholders that you need to replace for your repo):
+#### IAM policies and roles
+
+To allow access to the Docker images in your ECR repositories from the Kubernetes worker nodes and from the Garden in-cluster builder Pods, you need to configure an IAM policy and a [web identity IAM role](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-idp_oidc.html). To learn more about ECR IAM policies, please refer to [the AWS documentation](https://docs.aws.amazon.com/AmazonECR/latest/userguide/security_iam_id-based-policy-examples.html).
+
+Create this IAM policy to allow the Kubernetes nodes to pull images from your ECR repositories:
```
twelvemo, line 308:
```suggestion
If not already present, create this IAM policy to allow the Kubernetes nodes to pull images from your ECR repositories:
```

core/src/plugins/kubernetes/container/build/common.ts
```diff
@@ -306,6 +308,59 @@ export async function getUtilDaemonPodRunner({
...
+    const status = await compareDeployedResources(
+      ctx as KubernetesPluginContext,
+      api,
+      namespace,
+      [serviceAccount],
+      deployLog
+    )
+
+    // NOTE: This is here to make sure that we remove annotations in case they are removed in the garden config.
+    // `compareDeployedResources` as of today only checks wether the manifest is a subset of the deployed manifest.
```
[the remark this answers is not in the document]
stefreak:
```suggestion
    // `compareDeployedResources` as of today only checks whether the manifest is a subset of the deployed manifest.
```

docs/guides/in-cluster-building.md
```diff
@@ -350,13 +378,76 @@ To grant your service account the right permission to push to ECR, add this poli
...
+name: my-project
+...
+providers:
+  - name: kubernetes
+    ...
+    # If you use the kaniko build mode
+    buildMode: kaniko
+    kaniko:
+      serviceAccountAnnotations:
+        eks.amazonaws.com/role-arn: arn:aws:iam::<account-id>:role/<web-identity-role-name>
```
twelvemo, line 440: So nice 🎉

twelvemo (changes requested):
Reviewed the docs! Thanks for implementing this, this is super useful!
I have two suggestions concerning the structure of the docs as they contain significantly more content now than before:
1. Split into documentation about pulling images and pushing images.
2. For pushing images, keep the existing docs on how to allow all nodes to push to ECR and display using IRSA as an improved and recommended alternative.

docs/guides/in-cluster-building.md
stefreak: OK, I'll create a new guide for IRSA then.

I decided against that initially as almost always the ECR registry will be used for production deployments.

But at the same time I acknowledge that there is a trade off between ease of setup and security here.

twelvemo: But since we are suggesting IAM resource based policies in ECR we are only allowing push access to specific repositories. And the production ECR registry could also be in another account.

core/src/plugins/kubernetes/container/build/common.ts
```diff
@@ -306,6 +308,59 @@ export async function getUtilDaemonPodRunner({
...
+
+    return false
+  })
+}
+
+function isEqualAnnotations(r1: KubernetesResource, r2: KubernetesResource): boolean {
+  // normalize annotations before comparison
+  const a1 = r1.metadata.annotations !== undefined ? r1.metadata.annotations : {}
+  const a2 = r2.metadata.annotations !== undefined ? r2.metadata.annotations : {}
+  return JSON.stringify(a1) === JSON.stringify(a2)
```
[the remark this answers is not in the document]
stefreak: good point, using lowdash's `isEqual` now

docs/guides/in-cluster-building.md
stefreak: changed this

docs/guides/in-cluster-building.md
stefreak: Added a link to these docs. thanks for the good feedback!

core/src/plugins/kubernetes/container/build/common.ts
```diff
@@ -345,12 +408,18 @@ export async function ensureUtilDeployment({
       deployLog
     )
 
+    // if the service account changed, all pods part of the deployment must be restarted
```
srihas-g, line 411: what does `changed` mean here, name of the service account or the metadata associated to it?



Aren't credentials mounted via serviceAccount tokens in pods which are projected volumes? which means changing annotations doesn't need to restart pods to apply changes I believe.

I think Secret token was removed from serviceAccount object to make it dynamic since tokens are static and don't have a TTL.

source: https://kubernetes.io/docs/concepts/storage/projected-volumes/#serviceaccounttoken

stefreak: @srihas-g yes that's right. But this is specific to make IRSA work; if you change or add the IRSA annotation on the service account the pods need to be re-created to receive updated `AWS_*` environment variables.

jamesloosli:
This PR solves an issue we're having as well.

Any chance this will make the next release?

stefreak:
@jamesloosli I absolutely want to finish this but we won't make it before the 0.13 release (will happen this month).

stefreak:
Changed the base back to main, as this would need to land in main and not 0.12

highb:
@stefreak @twelvemo Any ideas of when this PR will land and ship? This would be a massive improvement for our security posture during in-cluster builds.

stefreak:
@highb right now we are focused on bug fixes and stability improvements– but this is definitely something I want to look back into in the coming months.
Sorry for the delay here.

twelvemo:
Picking this up again now and also making sure this works with GCP workload identity for the in-cluster builders as well.

core/test/integ/src/plugins/kubernetes/container/build/build.ts
```diff
@@ -566,3 +579,168 @@ describe.skip("Kubernetes Container Build Extension", () => {
...
+
+      const status = await compareDeployedResources({
+        ctx: ctx as KubernetesPluginContext,
+        api,
+        namespace: projectNamespace,
+        manifests: [originalServiceAccount],
+        log,
+      })
+      // Both annotations should be present
+      expect(isEqualAnnotations(originalServiceAccount, status.remoteResources[0]))
```
stefreak, line 645: I'm not sure if expect will actually fail if you do not explicitly assert for true 
```suggestion
      expect(isEqualAnnotations(originalServiceAccount, status.remoteResources[0])).to.be.true
```

images/skopeo/garden.yml
```diff
@@ -2,6 +2,6 @@ kind: Module
 type: container
 name: skopeo
 description: Used by the kubernetes provider for interacting with container registries within a cluster
-image: gardendev/skopeo:1.41.0-3
+image: gardendev/skopeo:1.41.0-4
```
stefreak, line 5: We do not seem to reference this image from Garden, but this can be cleaned up in a separate PR

stefreak (commented):
Looks good to me 👍

core/test/integ/src/plugins/kubernetes/container/build/build.ts
stefreak: We should install https://www.npmjs.com/package/eslint-plugin-chai-expect

twelvemo: Will create a new PR because some existing tests fail the new linter rules.

core/test/integ/src/plugins/kubernetes/container/build/build.ts
```diff
@@ -566,3 +579,168 @@ describe.skip("Kubernetes Container Build Extension", () => {
...
+
+      const status = await compareDeployedResources({
+        ctx: ctx as KubernetesPluginContext,
+        api,
+        namespace: projectNamespace,
+        manifests: [originalServiceAccount],
+        log,
+      })
+      // Both annotations should be present
+      expect(isEqualAnnotations(originalServiceAccount, status.remoteResources[0])).to.be.true
```
stefreak, line 645: I'm ok with this, but if the test fails this will give a better error message with the actual difference:
```suggestion
      expect(isEqualAnnotations(originalServiceAccount)).to.equal(status.remoteResources[0])
```

TimBeyer: This doesn't really work like that since `isEqualAnnotations` is a function that takes two parameters, and it checks the following
```
  const a1 = r1.metadata.annotations !== undefined ? r1.metadata.annotations : {}
  const a2 = r2.metadata.annotations !== undefined ? r2.metadata.annotations : {}
  return isEqual(a1, a2)
```

So I think what would be needed is 
```
expect(originalServiceAccount.metadata.annotations).to.eql(status.removeResources[0].metadata.annotations)
```

`to.equal` is strict equality btw so not a good match here.

TimBeyer: I see that change was already fixed so I'll go ahead and resolve that now :)

vvagaytsev (approved):

core/test/integ/src/plugins/kubernetes/container/build/build.ts
stefreak: Ah yes, my snippet was flawed but anna correctly used deep.equal on just the annotations; That should work fine 👍

PR #3386, opened 2022-11-25
fix(k8s): compareDeployedResources & manifest hash annotations

stefreak:

* **fix(k8s): compareDeployedResources should not ignore manifest hash**

compareDeployedResources actually just skipped isSubset check in case
the manifest hash was present.

Now it actually returns "outdated" when manifest hash differs.

This is necessary if the manifest removes annotations that are
present in the deployed manifest.

E.g.:

1. upsert annotations: {"some-annotation": "present"}
1. upsert annotations: {}

before this change, compareDeployedResources just returned "ready"
even though the deployedManifest was actually outdated (the annotation
needed to be removed)

This commit fixes that

* **fix(k8s): add manifest-hash annotations also for upsert operations**

previously we added garden manifest-hash only for kubectl apply ops.

This commit makes sure to add the manifest-hash also for upserts.

This enables for more accurate (and quicker) comparison results in
compareDeployedResources for the affected resources.

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/api.ts
+++ b/core/src/plugins/kubernetes/api.ts
@@ -54,14 +54,14 @@ import {
 } from "./types"
 import { LogEntry } from "../../logger/log-entry"
 import { kubectl } from "./kubectl"
-import { deline, urlJoin } from "../../util/string"
+import { deline, gardenAnnotationKey, urlJoin } from "../../util/string"
 import { KubernetesProvider } from "./config"
 import { StringMap } from "../../config/common"
 import { PluginContext } from "../../plugin-context"
 import { Writable, Readable, PassThrough } from "stream"
 import { WebSocketHandler } from "@kubernetes/client-node/dist/web-socket-handler"
 import { getExecExitCode } from "./status/pod"
-import { labelSelectorToString } from "./util"
+import { hashManifest, labelSelectorToString } from "./util"
 
 interface ApiGroupMap {
   [groupVersion: string]: V1APIGroup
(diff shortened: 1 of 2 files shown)


core/src/plugins/kubernetes/status/status.ts
```diff
@@ -381,15 +381,28 @@ export async function compareDeployedResources(
     if (deployedResource.metadata && deployedResource.metadata.annotations) {
       const lastAppliedHashed = deployedResource.metadata.annotations[gardenAnnotationKey("manifest-hash")]
 
-      // The new manifest matches the last applied manifest
-      if (lastAppliedHashed && (await hashManifest(manifest)) === lastAppliedHashed) {
-        continue
+      if (lastAppliedHashed) {
+        const calculated = await hashManifest(manifest)
+        if (calculated !== lastAppliedHashed) {
```
stefreak, line 386: Probably we should keep the continue if it's equal – that way we are saving some cpu cycles in that case.

stefreak:
```suggestion
        if (calculated === lastAppliedHashed) {
        continue
        } else {
```

core/src/plugins/kubernetes/api.ts
```diff
@@ -597,6 +597,17 @@ export class KubeApi {
     obj: O
     log: LogEntry
   }) {
+    // Hash the raw input and add as an annotation on each manifest (this is helpful beyond kubectl's own annotation,
```
stefreak, line 600: this code block is copied from `apply()` and should be extracted to a helper method

vvagaytsev:
There are some test failures. I'll convert this to a draft, and will try to take a look later this week.

Issue #3254, opened 2022-09-26
[FEATURE]: A better, more informative, append-only logger


eysi09:
# Feature Request
## Background / Motivation
The Garden Core logger currently has a few issues:

* It’s a little buggy:
  * The spinner goes out of whack if there's a lot of output.
* It could be a lot more informative:
  * The default log level shows too little information in terms of what’s happening in the user’s system. You might know that Garden is building a given module but can’t see what’s happening in the build. If it e.g. takes longer than expected, you’ll need to re-run the command with a higher log level to see what actually happened (assuming you’re not using Cloud).
  * When using the basic logger or on higher log levels you don’t see the context of the log lines. You might see something like: `#29 [built 1/1] RUN yarn run build` but can’t tell what build it belongs to.
  * Sections are confusing. Sometimes they refer to internal things and sometimes to stack graph nodes with out making it clear what the type of the node is.
  * When you install a Helm chart, Helm will print documentation about the chart from the maintainer. E.g. deprecation warnings. Garden doesn't show this information.
* Error messages are confusing:
  * We sometimes get complaints about error messages being hard to parse, in particular from developers.

## Suggested Implementation(s)
Below are some _suggestions_ for improvements. We should make sure to experiment with this and see what makes sense and what doesn’t. In particular for larger projects and focusing on the non-happy path.

* **Make the logger append-only**
  * We should completely remove the `setState` functionality and stop editing lines in place. Rather, each new log line should be, well, a new line.
  * We can have a single status line at the bottom with a spinner when in watch mode that makes it clear when Garden is responding to events.
* **Informative sections**
  * Sections should map directly to stack graph nodes, showing both the type and name of the node. E.g. `build.api`. We can reserve a keyword for internal sections, e.g.: `[internal] graph`
  * Sections should display the name of any external tool Garden is shelling out to along with the command name when applicable. E.g. `build.api ⇒ docker build ⇒ …` or  `deploy.api => helm install => ...`
  * All log lines should have sections. So instead of seeing `#29 [built 1/1] RUN yarn run build` you see `build.api ⇒ docker build ⇒ #29 [built 1/1] RUN yarn run build`.
* **Hints**
  * We could maintain a map of hints that we apply if applicable. Attaching info about external tools or providers to the error context could be helpful here. In pseudo code, something like:

```tsx
if (error.detail?.tool === "helm" && error.message.match(/timedout waiting for contition/) {
  log(hints.HELM_TIME_OUT_ERROR)
}
```

[...]

PR #3388, opened 2022-11-29
improvement(k8s): better error handling and logging in `PodRunner`

### LINKED ISSUE #3297: [FEATURE]: Different exit codes for different failure modes in Garden commands

### LINKED ISSUE #3330: [FEATURE]: Improve error messages when Kubernetes pod fails
### Background / Motivation

As reported by some of our users, whenever a pod running a task or a test fail because of a Kubernetes error, Garden will only report the error code, which can be confusing or disorienting.

For example, in the case of an OOM error that instructs Kubernetes to evict the pod, Garden will only print: "received 129 error from pod".

In case of an error that happens on the Kubernetes side (or any provider really), the user should receive actionable, useful information describing the cause of the issue and a potential solution.

Getting suggestions from Garden itself on how to solve an issue instead of receiving a generic error message, makes debugging and resolving the issue easier.

We could create a mapping between K8s errors and suggest solutions. E.g. in the case of 129, we could suggest how to configure the pods with increased resources. 
Additionally, we could print out the PodStatus and the lifecycles messages.

On a more general note, we could have a generic `PluginErrorHandler` that could be extended based on the error codes received when executing operations.

vvagaytsev:
* Refactored error handling and log retrieval for `garden run` and `garden test` commands.
* Aligned error code handling for running tests with and without artifacts.
* Extracted a number of helper functions to avoid too long function bodies

Fixes #3330.

This should partially resolve the issues reported in #3330. The error messages now are more structured and should be more informative. However, mappings between K8s errors and suggested solutions haven't been implemented yet. It's expected to be done in #3254.

Partially resolves #3297.

This PR also fixes flaky-tests like [this](https://app.circleci.com/pipelines/github/garden-io/garden/13888/workflows/f3ae8d4e-48f3-4bc3-a604-7016d52d4012/jobs/232512) (https://github.com/garden-io/garden/commit/9de010466794b220a664c8260751403998797bbf).

Note, that there is still an issue with duplicate logs printed on failures of the `garden run` command. Verified with `garden run task` and `garden run test` sub-commands.

There are some open discussions to clarify the ideas behind the changes made.

Manual tests can be found in [this branch](https://github.com/garden-io/garden/tree/refactor/run-and-test-logs-manual-test).

### DIFF
--- a/core/src/plugins/kubernetes/run.ts
+++ b/core/src/plugins/kubernetes/run.ts
@@ -432,6 +432,48 @@ async function runWithoutArtifacts({
   return result
 }
 
+/**
+ * Wraps a given {@code cmd} into a script to redirect its stdout and stderr to the same tmp file.
+ * See https://stackoverflow.com/a/20564208
+ * @param cmd the command to wrap
+ */
+function commandExecutionScript(cmd: string[]) {
+  return `
+exec 1<&-
+exec 2<&-
+exec 1<>/tmp/output
+exec 2>&1
+
+${cmd.join(" ")}
+`
+}
+
+/**
+ * For given {@code artifacts} prepares a script which will:
+ *   1. Create temp directory in the container
+ *   2. Create directories for each target, as necessary
+ *   3. Recursively (and silently) copy all specified artifact files/directories into the temp directory
+ *   4. Tarball the directory and pipe to stdout
+ * @param artifacts the artifacts to be processed
+ */
+function artifactsTarScript(artifacts: ArtifactSpec[]) {
+  // TODO: only interpret target as directory if it ends with a slash (breaking change, so slated for 0.13)
+  const directoriesToCreate = artifacts.map((a) => a.target).filter((target) => !!target && target !== ".")
+  const tmpPath = "/tmp/.garden-artifacts-" + randomString(8)
+
+  // TODO: escape the user paths somehow?
+  return `
+rm -rf ${tmpPath} >/dev/null || true
+mkdir -p ${tmpPath}
+cd ${tmpPath}
+touch .garden-placeholder
+${directoriesToCreate.map((target) => `mkdir -p ${target}`).join("\n")}
+${artifacts.map(({ source, target }) => `cp -r ${source} ${target || "."} >/dev/null || true`).join("\n")}
+tar -c -z -f - . | cat
+rm -rf ${tmpPath} >/dev/null || true
+`
+}
+
 async function runWithArtifacts({
   ctx,
   api,
@@ -553,15 +595,7 @@ async function runWithArtifacts({
     const cmd = [...command!, ...(args || [])].map((s) => JSON.stringify(s))
 
     try {
-      // See https://stackoverflow.com/a/20564208
-      const commandScript = `
-exec 1<&-
-exec 2<&-
-exec 1<>/tmp/output
-exec 2>&1
-
-${cmd.join(" ")}
-`
+      const commandScript = commandExecutionScript(cmd)
 
       const res = await runner.exec({
         // Pipe the output from the command to the /tmp/output pipe, including stderr. Some shell voodoo happening
@@ -595,26 +629,7 @@ ${cmd.join(" ")}
       })
     }
 
-    // TODO: only interpret target as directory if it ends with a slash (breaking change, so slated for 0.13)
-    const directoriesToCreate = artifacts.map((a) => a.target).filter((target) => !!target && target !== ".")
-    const tmpPath = "/tmp/.garden-artifacts-" + randomString(8)
-
-    // This script will
-    // 1. Create temp directory in the container
-    // 2. Create directories for each target, as necessary
-    // 3. Recursively (and silently) copy all specified artifact files/directories into the temp directory
-    // 4. Tarball the directory and pipe to stdout
-    // TODO: escape the user paths somehow?
-    const tarScript = `
-rm -rf ${tmpPath} >/dev/null || true
-mkdir -p ${tmpPath}
-cd ${tmpPath}
-touch .garden-placeholder
-${directoriesToCreate.map((target) => `mkdir -p ${target}`).join("\n")}
-${artifacts.map(({ source, target }) => `cp -r ${source} ${target || "."} >/dev/null || true`).join("\n")}
-tar -c -z -f - . | cat
-rm -rf ${tmpPath} >/dev/null || true
-    `
+    const tarScript = artifactsTarScript(artifacts)
 
     // Copy the artifacts
     const tmpDir = await tmp.dir({ unsafeCleanup: true })
--- a/core/src/plugins/kubernetes/run.ts
+++ b/core/src/plugins/kubernetes/run.ts
@@ -734,7 +734,7 @@ interface RunAndWaitResult {
 }
 
 export interface PodErrorDetails {
-  logs: string
+  logs?: string
   // optional details
   exitCode?: number
   containerStatus?: V1ContainerStatus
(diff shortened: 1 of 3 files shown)


core/src/plugins/kubernetes/run.ts
```diff
@@ -1080,4 +1030,80 @@ export class PodRunner extends PodRunnerParams {
...
+          }
+
+          if (!!logs) {
+            errorDesc += `Here are the logs until the error occurred:\n\n${logs}`
+          }
+
+          return errorDesc
+        // Unknown/unexpected error, rethrow it
+        default:
+          throw err
```
vvagaytsev, line 1093: Most likely, this should never happen. I guess it might be better to print some error message instead of throwing an error here.

core/src/plugins/kubernetes/run.ts
```diff
@@ -1031,16 +981,16 @@ export class PodRunner extends PodRunnerParams {
...
+  async getMainContainerLogs(): Promise<string> {
+    const mainContainerName = this.getMainContainerName()
     try {
-      return (await this.getMainContainerLogs()).trim()
+      const allLogs = await this.getLogs()
+      const containerLogs = allLogs.find((l) => l.containerName === mainContainerName)?.log?.trim()
+      return containerLogs || ""
     } catch (err) {
-      return ""
+      return `[Could not retrieve logs for container '${mainContainerName}': ${
```
vvagaytsev, line 991: I had an idea to introduce a flag like `silentOnFailure` here. Setting it to `false` in the `handlePodError(...)` method and `true` in the other places could help to do a retry for container logs retrieval.

But container logs retrieval might not be the most frequently failing operation, so I decided to keep the code simple. Also, getters with conditional parameters can be confusing :)

vvagaytsev:
Converting to draft, there are some test failures.

core/src/plugins/kubernetes/run.ts
vvagaytsev: This comment is no longer relevant. Alternative log retrieval has been removed from `handlePodError(...)`. All logs are now collected in the places where errors occur.

core/src/plugins/kubernetes/run.ts
vvagaytsev: This is no longer valid too. Some unexpected errors can occur, see the example in https://app.circleci.com/pipelines/github/garden-io/garden/13873/workflows/f3b63869-0a9c-4f2e-a27e-3b4f208ca2ad/jobs/232172

This method has been refactored in https://github.com/garden-io/garden/pull/3388/commits/82985ac646212f649d8ec6dd7bf0db3d216dbc55. Type-safe error rendering has been implemented.

core/src/plugins/kubernetes/run.ts
```diff
@@ -1034,11 +1038,7 @@ export class PodRunner extends PodRunnerParams {
     // We never want to restart containers in these ephemeral pods
     pod.spec.restartPolicy = "Never"
 
-    try {
-      await this.api.createPod(this.namespace, pod)
-    } catch (error) {
-      throw new PodRunnerError(`Failed to create Pod ${this.podName}: ${error.message}`, { error })
-    }
+    await this.api.createPod(this.namespace, pod)
```
vvagaytsev, line 1055: This also prevents self-repeating in the error messages like [here](https://app.circleci.com/pipelines/github/garden-io/garden/13882/workflows/c13c6a22-af12-4d5e-80ad-5387a4e99ee5/jobs/232341).

core/src/plugins/kubernetes/run.ts
```diff
@@ -1054,25 +1054,42 @@ export class PodRunner extends PodRunnerParams {
...
+    // Some types and predicates to identify known errors
+    const knownErrorTypes = ["out-of-memory", "timeout", "pod-runner"] as const
+    type KnownErrorType = typeof knownErrorTypes[number]
+    // A known error is always an instance of a subclass of GardenBaseError
+    type KnownError = {
+      message: string
+      type: KnownErrorType
+      detail: PodErrorDetails
+    }
+    const isKnownError = (error: any): error is KnownError => {
```
Orzelius, line 1080: haha what is that return type, I've never seen something like that before. Is there a reason it couldn't just be `boolean`?

core/src/plugins/kubernetes/run.ts
```diff
@@ -1120,6 +1120,8 @@ export class PodRunner extends PodRunnerParams {
           }
 
           return errorDesc
+        case "kubernetes":
+          return `Unable to start command execution. Kubernetes cluster returned error:\n${error.message}\n\nPlease check the cluster health.`
```
Orzelius, line 1138: I don't think cluster itself returns errors. Perhaps this should be changed to something like

```suggestion
          return `Unable to start command execution. Failed to initiate test-runner pod with error:\n${error.message}\n\nPlease check the cluster health.`
```

Orzelius (commented):
You're doing gods work! Left a few comments, feel free to address (or not) and re-ping for a second round of review.

core/src/plugins/kubernetes/run.ts
vvagaytsev: That's a special type :) See [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) for the details. It ensures that `error` is of type `KnownError` if the predicate returns true. Som later in the code, TS understands it as `KnownError` instead of `any`.

core/src/plugins/kubernetes/run.ts
```diff
@@ -946,16 +946,19 @@ export class PodRunner extends PodRunnerParams {
...
-
-
-        throw new PodRunnerError(`Failed to start Pod ${podName}.`, errorDetails)
       }
 
       // reason "Completed" means main container is done, but sidecars or other containers possibly still alive
       if (state === "stopped" || exitReason === "Completed") {
+        if (!!exitCode) {
+          throw new PodRunnerError(`Command exited with code ${exitCode}.`, errorDetails)
+        }
```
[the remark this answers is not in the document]
vvagaytsev: It identifies a non-zero exit code and throws an error.

core/src/plugins/kubernetes/run.ts
Orzelius: that's very cool, learn something new every day

core/src/plugins/kubernetes/run.ts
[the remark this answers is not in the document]
vvagaytsev: So, a missing error code or `0` are interpreted as success.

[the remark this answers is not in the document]
Orzelius: Perhaps the statement should be `exitcode !== 0` then? I find that `!!` requires a bit of mental gymnastics to figure out with this `number | undefined` type

core/src/plugins/kubernetes/run.ts
```diff
@@ -697,18 +693,19 @@ function handlePodError({
...
 }: {
   err: any
   containerLogs: string
   command: string[]
   startedAt: Date
-  exitCode: number
   version: string
   moduleName
 }) {
+  const errorDetail = <PodErrorDetails>err.detail
```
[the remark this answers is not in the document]
vvagaytsev: This cast was removed in https://github.com/garden-io/garden/commit/9de010466794b220a664c8260751403998797bbf

The basic idea in that commit was to identify the "supported" (i.e. expected and handleable) errors. I did not want to make any bigger changes there. So I preserved the existing way of identifying errors by their `type` field. All custom errors derived from `GardenBaseError` class have that field. Any other errors can be interpreted as unexpected ones and re-thrown.

We use tagged classes in the error class hierarchy where the `type` field stands for the tag. We could try using [class hierarchy instead of tagged classes](https://dev.to/kylec32/effective-java-prefer-class-hierarchies-to-tagged-classes-59l9) by adding abstract method `GardenBaseError.render()` to hide each error's rendering in its class definition, but it would be a bigger change.

Another thing is that we do not have any type restrictions for `detail` field in `GardenBaseError`, it's `detail: any`. Declaring it as generic might not make much sense. So I introduced `type PodErrorDetails` -- an "expected type of error detail" that can contain valuable error details.

core/src/plugins/kubernetes/run.ts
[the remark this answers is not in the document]
vvagaytsev: Well, `undefined` is also a legal value here, but it can be rewritten in more understandable way, I agree 👍

core/src/plugins/kubernetes/run.ts
vvagaytsev: Yes, this also helps in the `switch` case below in the code. If we remove any of the `case` statements, then the `_exhaustiveCheck: never` variable declaration will fail with a TS compilation error.

core/src/plugins/kubernetes/run.ts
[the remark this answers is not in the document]
vvagaytsev: Made it more explicit in https://github.com/garden-io/garden/pull/3388/commits/fbc4ea8e81b5a7dadc33988c13f6670acace8445

core/src/plugins/kubernetes/run.ts
vvagaytsev: Well, the underlying library function `KubeApi.createNamespacedPod(...)` can throw errors, and we always wrap 'em as `KubernetesError`. Effectively, those are always pod-creation errors.

So we can change the error message as you suggested, but remove "test-" prefix. It can be a task execution.

vvagaytsev: Changed this in https://github.com/garden-io/garden/pull/3388/commits/0bcef414c979591eabc19668b2ad26fc4904c04f

PR #3374, opened 2022-11-23
update k8s client library and fix teleport issue

Orzelius:
commits have more details

PR #3389, opened 2022-11-29
chore: remove dns-lookup-cache

Walther:

Fixes the regression in local Docker Desktop Kubernetes introduced in https://github.com/garden-io/garden/pull/3374 due to some adverse interaction with the DNS caching library we used.

Replaces the need to revert the changes with https://github.com/garden-io/garden/pull/3387

The `dns-lookup-cache` library has its latest commit at nearly 3 years ago, so we should remove it anyway. If we absolutely need a DNS caching solution, we should figure out an alternative.

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -59,7 +59,6 @@
     "deline": "^1.0.4",
     "dependency-graph": "^0.9.0",
     "directory-tree": "^2.2.4",
-    "dns-lookup-cache": "^1.0.4",
     "docker-file-parser": "^1.0.4",
     "dockerode": "^3.2.0",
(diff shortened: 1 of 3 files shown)


Orzelius:
Wow good job on catching this, I was scratching my head pretty bad with this

vvagaytsev:
Great job, @Walther! Thank you! 🚀

vvagaytsev (approved):

edvald:
FYI, the original reason for including that library was an issue for users running Rancher, bumping into dropped DNS requests and therefore connection issues with clusters. We should be aware of potential implications but I still agree that we should likely drop this library.

PR #3391, opened 2022-11-29
feat(k8s): custom nodeSelector for util pods

Walther:

Enables customizing `nodeSelector` requirements for the `garden-util` pods when using `kaniko` in kubernetes.

Fixes #

Taking inspiration from https://github.com/garden-io/garden/pull/3365.

This is currently implemented on top of existing code. This means that we configure `nodeSelector` for the `util` manifests in the `common.ts` file, even if we only expose the configuration options under `kaniko.util` specifically. If we want to, we can later refactor this, for example move the configuration logic to `kaniko.ts` specifically, or keep the configuration reading logic in `common.ts` and move the `util` configurations upwards in the configmap, or something else.

I've also verified the PR locally by editing the `examples/demo-project/garden.yml` file
```
providers:
  - name: local-kubernetes
    environments: [local]
    buildMode: kaniko
    kaniko:
      util:
        nodeSelector:
          "kubernetes.io/os": "linux"
```

- using `"kubernetes.io/os": "linux"` to verify a successful local deployment
- using `"kubernetes.io/os": "macos"` to verify an unsuccessful deployment due to no matching nodes being available.

---

Please let me know if I missed anything!

### DIFF
--- a/core/src/plugins/kubernetes/config.ts
+++ b/core/src/plugins/kubernetes/config.ts
@@ -706,18 +706,18 @@ export const kubernetesConfigBase = () =>
           ),
         nodeSelector: joiStringMap(joi.string()).description(
           dedent`
-            Exposes the \`nodeSelector\` field on the PodSpec of the Kaniko pods. This allows you to constrain the Kaniko pods to only run on particular nodes.
+            Exposes the \`nodeSelector\` field on the PodSpec of the Kaniko pods. This allows you to constrain the Kaniko pods to only run on particular nodes. The same nodeSelector will be used for each util pod unless they are specifically set under \`util.nodeSelector\`.
 
             [See here](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/) for the official Kubernetes guide to assigning pods to nodes.
           `
         ),
         tolerations: joiSparseArray(tolerationSchema()).description(
           deline`Specify tolerations to apply to each Kaniko builder pod. Useful to control which nodes in a cluster can run builds.
-          Same tolerations will be used for the util pod unless they are specifically set under \`util.tolerations\``
+          The same tolerations will be used for each util pod unless they are specifically set under \`util.tolerations\``
         ),
         annotations: annotationsSchema().description(
           deline`Specify annotations to apply to each Kaniko builder pod. Annotations may have an effect on the behaviour of certain components, for example autoscalers.
-          Same annotations will be used for each util pod unless they are specifically set under \`util.annotations\``
+          The same annotations will be used for each util pod unless they are specifically set under \`util.annotations\``
         ),
         util: joi.object().keys({
           tolerations: joiSparseArray(tolerationSchema()).description(
--- a/core/src/plugins/kubernetes/config.ts
+++ b/core/src/plugins/kubernetes/config.ts
@@ -232,6 +232,7 @@ export interface KubernetesConfig extends BaseProviderConfig {
     util?: {
       tolerations?: V1Toleration[]
       annotations?: StringMap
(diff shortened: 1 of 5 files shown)


core/src/plugins/kubernetes/container/build/common.ts
```diff
@@ -583,8 +583,9 @@ export function getUtilManifests(
   }
 
   // Set the configured nodeSelector, if any
-  if (!isEmpty(provider.config.kaniko?.nodeSelector)) {
-    deployment.spec!.template.spec!.nodeSelector = provider.config.kaniko?.nodeSelector
+  const nodeSelector = provider.config.kaniko?.util?.nodeSelector || provider.config.kaniko?.nodeSelector
+  if (!isEmpty(nodeSelector)) {
+    deployment.spec!.template.spec!.nodeSelector = nodeSelector
   }
```
Walther, line 589: Do you think it is okay to keep this here as it has been so far, or should we refactor this to `kaniko.ts` specifically?

core/test/unit/src/plugins/kubernetes/container/build/common.ts
```diff
@@ -197,5 +198,27 @@ describe("common build", () => {
...
+      provider.config.kaniko = {
+        util: {
+          nodeSelector: { "kubernetes.io/os": "linux" },
+        },
+      }
+      const result = getUtilManifests(provider, "test", [])
+
+      const podNodeSelector = result.deployment.spec.template.spec?.nodeSelector
+      expect(podNodeSelector).to.eql(provider.config.kaniko.util?.nodeSelector)
+    })
```
Walther, line 222: related to above; if we want to refactor the configuration from `common.ts` to `kaniko.ts`, we could also move these tests respectively

docs/reference/providers/kubernetes.md
```diff
@@ -266,7 +266,7 @@ providers:
           value:
 
       # Specify annotations to apply to each Kaniko builder pod. Annotations may have an effect on the behaviour of
-      # certain components, for example autoscalers. Same anotations will be used for each util pod unless they are
+      # certain components, for example autoscalers. Same annotations will be used for each util pod unless they are
```
stefreak, line 269: 👍

core/src/plugins/kubernetes/config.ts
```diff
@@ -725,6 +726,9 @@ export const kubernetesConfigBase = () =>
           annotations: annotationsSchema().description(
             "Specify annotations to apply to each garden-util pod and deployments."
           ),
+          nodeSelector: joiStringMap(joi.string()).description(
```
stefreak, line 729: maybe let's add a hint to the other `nodeSelector` config flag, that `kaniko.nodeSelector` will be applied to the util container unless the `kaniko.util.nodeSelector` is specified (Similar to the description of `tolerations` and `annotations`).

stefreak (changes requested):
Thanks for the good work, only had some minor comments 👍

Orzelius (commented):
I have nothing to add to @stefreak review. I'm really impressed with how fast you got this done

core/src/plugins/kubernetes/config.ts
Walther: Added! Unified some of the wording while at it.

Orzelius (approved):

stefreak (approved):
🥳

core/src/plugins/kubernetes/container/build/common.ts
Walther: Based on slack conversations, let's do the refactor later so we can get this feature into the next release window quickly.

PR #3394, opened 2022-11-30
fix(framework): idempotent config store deletes

eysi09:

Previously the ConfigStore class would throw if attempting to delete an entry with a key that doesn't exist.

So e.g. running delete twice on the same key would fail the second time.

I'm 99.9% sure this that is just a mistake that was made years ago and hasn't surfaced before.

This fix is needed to handle a test error on a different branch.

Fixes #

### DIFF
--- a/core/test/unit/src/config-store.ts
+++ b/core/test/unit/src/config-store.ts
@@ -141,15 +141,9 @@ describe("ConfigStore", () => {
       throw new Error("Expected error, got " + res)
     })
 
-    it("should throw if key is not found", async () => {
-      let res
-      try {
-        res = await config.delete(["key"])
-      } catch (err) {
-        expect(err.type).to.equal("local-config")
-        return
-      }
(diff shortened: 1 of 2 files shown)


core/test/unit/src/config-store.ts
```diff
@@ -140,16 +140,5 @@ describe("ConfigStore", () => {
       }
       throw new Error("Expected error, got " + res)
     })
-
-    it("should throw if key is not found", async () => {
```
stefreak, line 144: if the desired behaviour is no error on delete if the key isn't there, let's have a test for that

vvagaytsev: Agree, a new test can ensure that nothing is thrown on repeated deletion.

eysi09: Good point. I re-introduced the test.

thsig (approved):

PR #3395, opened 2022-12-01
fix: send a periodic ping over the websocket for exec

### LINKED ISSUE #3324: Garden not detecting finished buildkit job
### Current Behavior

Garden does not detect that a buildkit job is finished, and basically hangs. If the local CLI command is stopped and re-run, then it's correctly registered as already built, and the steps continue.

### Expected behavior

Correctly detect a finished build

### Reproducible example

(Note that I haven't yet tried to replicate this)
The dockerfile in particular is of the following format
```dockerfile
# The below get passed from other garden-build modules as buildargs
ARG BASE_IMAGE_VERSION
ARG BASE_IMAGE_NAME
FROM ${BASE_IMAGE_NAME}:${BASE_IMAGE_VERSION} as builder
WORKDIR /app
COPY . .
RUN ./command_that_builds.sh && \
  mkdir compiled-binaries && \
  ./copy_binaries.sh compiled-binaries && \
  ./clean_compilation.sh

# The below is only used as a binary container. I'd use scratch, but kaniko did not like that
FROM alpine
COPY --from=builder /app/compiled-binaries /
COPY --from=builder /app/resources /
```

### Your environment

* **OS:** Linux
* **How I'm running Kubernetes:** AKS

`garden version` 0.12.46

TimBeyer:

Currently it can happen that the websocket connection is timed out somewhere in the network, and no error event is triggered. Garden believes something is still executing, and never receives the information about completion or failure. This sends a periodic ping so that hopefully the connection stays alive.

As an extra step we could implement the full timeout logic in https://github.com/websockets/ws#how-to-detect-and-close-broken-connections

Fixes #3324

This can be reproduced by starting a build with a Dockerfile containing `RUN sleep 200`, and then running `garden build backend -f -l5`

When the following log output is printed to the screen:
```
  → #11 [stage1 2/3] RUN sleep 200
```

One would need to disconnect from the network.

Without this patch, garden would just hang forever.

With this patch, the user receives this error message:

```
✖ backend                   → Building image backend:v-6615eb48cc...

Failed building backend. Here is the output:
Lost connection to the Kubernetes WebSocket API (Timed out)
Error: Lost connection to the Kubernetes WebSocket API (Timed out)
    at Timeout.<anonymous> (/Users/steffen/work/github/garden/core/src/plugins/kubernetes/api.ts:863:24)
    at listOnTimeout (internal/timers.js:557:17)
    at processTimers (internal/timers.js:500:7)
✖ tasks                     → Failed task build.backend.f2def460-7263-11ed-b60c-2ffde9c2bdad
ℹ Remaining tasks 0
ℹ Remaining tasks 0
Done flushing all events and log entries.
1 build action(s) failed!
Error: 1 build action(s) failed!
    at handleProcessResults (/Users/steffen/work/github/garden/core/src/commands/base.ts:532:19)
    at BuildCommand.action (/Users/steffen/work/github/garden/core/src/commands/build.ts:148:32)
    at GardenCli.runCommand (/Users/steffen/work/github/garden/core/src/cli/cli.ts:512:20)
    at GardenCli.run (/Users/steffen/work/github/garden/core/src/cli/cli.ts:671:26)
    at Object.runCli (/Users/steffen/work/github/garden/cli/src/cli.ts:41:14)

Error Details:

results:
  build.backend:
    type: build
    description: building backend
    key: build.backend
    name: backend
    error: {}
    startedAt: '2022-12-02T17:08:32.051Z'
    completedAt: '2022-12-02T17:09:30.926Z'
    batchId: 9d6e7d56-c453-4b86-98d5-563e410a327c
    version: v-6615eb48cc


Done flushing all events and log entries.
```


agates4:
I've ran our long build command with this version of Garden and I've confirmed it results in accurately reporting the results back to our shell! 
LGTM, as far as fixing the issue in https://github.com/garden-io/garden/issues/3324

vvagaytsev (approved):

core/src/plugins/kubernetes/api.ts
```diff
@@ -854,6 +843,50 @@ export class KubeApi {
...
+  }, WEBSOCKET_KEEPALIVE_INTERVAL)
+
+  let pingTimeout: NodeJS.Timeout | undefined
+
+  function heartbeat() {
+    if (pingTimeout) {
+      clearTimeout(pingTimeout)
+    }
+    pingTimeout = setTimeout(() => {
+      ws.emit("error", new Error("Lost connection to the Kubernetes WebSocket API (Timed out)"))
```
vvagaytsev, line 861: Should we include the timeout value in the error message?

core/src/plugins/kubernetes/api.ts
```diff
@@ -769,69 +768,59 @@ export class KubeApi {
...
+          )
+        )
+
+        ws.on("error", (err) => {
+          done = true
+          reject(err)
         })
-        .catch(reject)
+      } catch (err) {
+        reject(err)
```
vvagaytsev, line 819: Should `done` get any value here?

vvagaytsev (changes requested):
Looks good! I left a couple of comments/questions, please take a look.

core/src/plugins/kubernetes/api.ts
TimBeyer: It did not previously and I think that's fine.
`done` is only relevant once the `execHandler` is running and different callbacks might get called that only need to resolve once.
In this `try/catch` block, the only way we throw is if the initialization of the `exec` was interrupted and once that is the case, none of the other completion handlers will be triggered anymore.

core/src/plugins/kubernetes/api.ts
vvagaytsev: Yes, let's add it.

TimBeyer: Done

vvagaytsev (approved):

agates4:
Thanks for the really quick turnaround on this fix. Is this something that would be released in `0.12.48`? This currently **blocks** our ability to build our production images

PR #3400, opened 2022-12-02
improvement(pulumi): upgrade pulumi version

ishantagarwal-x213235:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

PR #3401, opened 2022-12-02
docs(pulumi): re-generated reference docs

vvagaytsev:
Patches #3400

Fixes #

### DIFF
--- a/docs/reference/providers/pulumi.md
+++ b/docs/reference/providers/pulumi.md
@@ -32,7 +32,7 @@ providers:
     environments:
 
     # The version of pulumi to use. Set to `null` to use whichever version of `pulumi` is on your PATH.
-    version: 3.40.0
+    version: 3.48.0
 
     # Overrides the default plan directory path used when deploying with the `deployFromPreview` option for pulumi
     # modules.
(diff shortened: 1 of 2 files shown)


stefreak (changes requested):
I think we should also revert `_includeInGardenImage: false` right?
Do we want to continue to include pulumi in our workflow runner docker images?

vvagaytsev:
> I think we should also revert `_includeInGardenImage: false` right? Do we want to continue to include pulumi in our workflow runner docker images?

I'm not 100% sure. It looks like we have had `_includeInGardenImage: false` for every default Pulumi version, buy mabe we don't need to include it in workflow runner images.

@thsig wdyt?

thsig:
I don't think we need to include it.

We can always change our mind later and start including the latest version.

stefreak (approved):
ok then LGTM 🚢 👍

stefreak:
> In #3400 we got `_includeInGardenImage: false` for Pulumi `3.40.0` because it stopped being the latest version. And we got `_includeInGardenImage: true` for Pulumi `3.48.0` because it became the latest one.
> 
> So, it should not make any harm. We can keep it as it is, and exclude it in a separate PR.

Ah right, got it! My fault I didn't look at the diff closely enough. Thanks for clarifying!

PR #3402, opened 2022-12-02
fix: run action result processing

Orzelius:
The statement was the wrong way around

Big thanks to @vvagaytsev for helping to find this!


vvagaytsev (approved):
🚀

PR #3413, opened 2022-12-06
improvement: decoupled the cloud api login from project configuration and verification

mkhq:

The cloud API required both a `domain` and a `projectId` to be initialised. This PR removes the requirement of a `projectId` which makes it possible to implement new cloud commands that does not need a `projectId`. An example would be a command to create garden cloud projects.

A new environment variable `GARDEN_CLOUD_DOMAIN` is introduced that makes it possible to override a cloud domain configured in the project for easier configuration/testing.

Fixes #


vvagaytsev (approved):
Looks good, thank you! 🚀

PR #3423, opened 2022-12-07
fix: support local builds in microk8s in multipass

TimBeyer:
When microk8s runs within a multipass VM, we need to explicitly transfer any files into the VM before having access to them. This broke builds with local docker on macOS.

Since there is no simple way of determining whether microk8s runs within multipass, we instead ask multipass to list all its VMs and see if the default microk8s machine name is included. In that case we assume that we are running in multipass and transfer the file over.

Another option would be to mount the temporary directory in and potentially reduce the file transfer overhead

```ts
if (isInMultipassVM) {
  await exec("multipass", ["mount", parsedTempFilePath.dir, `${MULTIPASS_VM_NAME}:${parsedTempFilePath.dir}`])
}

await exec("microk8s", ["ctr", "image", "import", sourceFilePath])

// Clean up the file within the VM by deleting it explicitly
if (isInMultipassVM) {
  await exec("multipass", ["umount", `${MULTIPASS_VM_NAME}:${parsedTempFilePath.dir}`])
}
```

This does work well, but it is potentially not concurrency safe in case multiple images build at the same time since they could unmount each others' directories when the same temporary directory is chosen.
It will also mount all the other files within that directory.

### DIFF
--- a/core/src/plugins/kubernetes/local/microk8s.ts
+++ b/core/src/plugins/kubernetes/local/microk8s.ts
@@ -17,6 +17,7 @@ import chalk from "chalk"
 import { naturalList, deline } from "../../../util/string"
 import { ExecaReturnValue } from "execa"
 import { PluginContext } from "../../../plugin-context"
+import { parse as parsePath } from "path"
 
 export async function configureMicrok8sAddons(log: LogEntry, addons: string[]) {
   let statusCommandResult: ExecaReturnValue | undefined = undefined
@@ -66,6 +67,28 @@ export async function getMicrok8sImageStatus(imageId: string): Promise<BuildStat
   return { ready: res.stdout.split("\n").includes(clusterId) }
 }
 
+const MULTIPASS_VM_NAME = "microk8s-vm"
+
+type MultipassListOutput = {
+  list: {
+    ipv4: string[]
+    name: string
+    release: string
+    state: string
+  }[]
+}
+
+async function isMicrok8sRunningInMultipassVM(): Promise<boolean> {
+  try {
+    const res = await exec("multipass", ["list", "--format", "json"])
+
+    const data = JSON.parse(res.stdout) as MultipassListOutput
+    return data.list.some((vm) => vm.name === MULTIPASS_VM_NAME)
+  } catch (_err) {
+    return false
+  }
+}
+
 export async function loadImageToMicrok8s({
   module,
   imageId,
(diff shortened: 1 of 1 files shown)


Orzelius:
> it is potentially not concurrency safe in case multiple images build at the same time since they could unmount each others' directories when the same temporary directory is chosen.

I don't think we can look over this limitation. Building most if not all images at once is a super common workflow

TimBeyer:
@Orzelius that was just to explain why I didn't go for the mounting approach. The actual implementation does not have that issue.

core/src/plugins/kubernetes/local/microk8s.ts
```diff
@@ -66,6 +67,27 @@ export async function getMicrok8sImageStatus(imageId: string): Promise<BuildStat
...
+
+type MultipassListOutput = {
+  list: {
+    ipv4: string[]
+    name: string
+    release: string
+    state: string
+  }[]
+}
+export async function microk8sIsRunningInMultipassVM(): Promise<boolean> {
```
Orzelius, line 80: a small nitpick but I think it should be `isMicrok8sRunningInMultipassVM` or even something like `getMicrok8sMulipassVMStatus` that returns `{ isMultipass: boolean }`. In cases like this I [miss go's named returns tbh](https://go.dev/tour/basics/7)

Orzelius: Also, I don't think there's a reason to export this

Orzelius (commented):
:facepalm: sorry it seems I powerread through the important bit in the description.

I have a few nitpicks but overall looks good to me

Orzelius (approved):

core/src/plugins/kubernetes/local/microk8s.ts
```diff
@@ -86,7 +109,32 @@ export async function loadImageToMicrok8s({
...
+      // And then later on remove it again manually
+
+      // We only grab the base name of the temp file
+      // since else we would need to create the entire path of the temp file first
+      // Once microk8s releases with multipass v1.11.0,
+      // we can use the `-p` flag and simplify this code again
+      const filePath = isInMultipassVM ? `/tmp/${parsedTempFilePath.base}` : sourceFilePath
+
+      // Transfer the file from the source path into the new destination path within the VM
+      if (isInMultipassVM) {
```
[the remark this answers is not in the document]
TimBeyer: If this throws then we wouldn't have anything to clean up.
The `microk8s` command could fail, while the `multipass transfer` succeeded.
In that case we probably would want to clean up the file, though.

However, if we were to do an `await` in the `finally` block, in cases or an actual caught exception before, we'd most likely end up with an `UnhandledPromiseRejectionWarning` being logged.

Also the `finally` could also fail and throw, so it would also need to be caught so that we can throw the correct `RuntimeError`.

Since all of this data is anyway put within `/tmp`, it'll eventually get cleaned up even if we don't perform the manual cleanup now.

I think that is an alright tradeoff versus having a much more complicated flow where we could have a potentially throwing `await` within a `finally` block.

vvagaytsev (approved):
LGTM! 🚀

PR #3425, opened 2022-12-07
chore: add runtime.tasks.<task-name>.outputs.log schema to helm and kubernetes modules

### LINKED ISSUE #3378: [FEATURE]: Add support for ${runtime.tasks.<task-name>.outputs.log} to Helm modules.
## Feature Request

To keep consistent with other module types I'd love if if this runtime output as available.

### Background / Motivation

Other module types already support this and it seems an oversigth that the Helm module does not.

Work in a similar manner as the k8s module does for the same log output.

It makes the Helm module have feature parity with the k8s module.

I assume it's fairly easy to do what is done for the k8s module type.

We are a Helm chart based organization, and being able to capture task log output would be very helpful.

Walther:

Adds the `joi` schema for `runtime.tasks.<task-name>.outputs.log` for both `helm` and `kubernetes` module types.
The necessary code for passing the logs already exists for both [helm](https://github.com/garden-io/garden/blob/7ba352815890ebcf727ae5731dd06bd3599aa866/core/src/plugins/kubernetes/helm/run.ts#L185) as well as [kubernetes](https://github.com/garden-io/garden/blob/7ba352815890ebcf727ae5731dd06bd3599aa866/core/src/plugins/kubernetes/kubernetes-module/run.ts#L73), but they were not documented correctly.

Fixes #3378 

### DIFF
--- a/core/src/plugins/container/container.ts
+++ b/core/src/plugins/container/container.ts
@@ -24,6 +24,7 @@ import { dedent } from "../../util/string"
 import { Provider, GenericProviderConfig, providerConfigBaseSchema } from "../../config/provider"
 import { isSubdir } from "../../util/util"
 import { GetModuleOutputsParams } from "../../types/plugin/module/getModuleOutputs"
+import { taskOutputsSchema } from "../kubernetes/task-results"
 
 export interface ContainerProviderConfig extends GenericProviderConfig {}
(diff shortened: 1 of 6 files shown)


vvagaytsev (approved):
Great job! 🚀

PR #3426, opened 2022-12-07
fix(server): send better close codes on ws connection close

eysi09:

These better websocket close / ready messages are helpful for Cloud (and future clients) when handling websocket connections.

Fixes #


vvagaytsev (approved):
👍

PR #3430, opened 2022-12-08
fix(k8s): fixed error detail format in `PodRunner.exec()`

vvagaytsev:
This patches https://github.com/garden-io/garden/pull/3388. If the method `PodRunner.exec()` throws an error, then the error's `detail` field must contain the `result` field to ensure the proper error handling in [`skopeoBuildStatus`](https://github.com/garden-io/garden/blob/main/core/src/plugins/kubernetes/container/build/common.ts#L198) function.

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/api.ts
+++ b/core/src/plugins/kubernetes/api.ts
@@ -170,6 +170,14 @@ type WrappedApi<T> = {
   T[P]
 }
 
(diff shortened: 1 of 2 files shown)


10ko (approved):
Approving since it's a breaking regression, but it would be nice to have integration tests for this use-case, as @stefreak pointed out.

PR #3432, opened 2022-12-12
chore: bump platform api types to 1.271.0

mkhq:

This updates the cloud platform API types to the latest version.

Fixes #

### DIFF
--- a/core/src/analytics/analytics.ts
+++ b/core/src/analytics/analytics.ts
@@ -21,7 +21,7 @@ import dedent from "dedent"
 import { getGitHubUrl } from "../docs/common"
 import { Profile } from "../util/profiling"
 import { ModuleConfig } from "../config/module"
-import { UserResponse } from "@garden-io/platform-api-types"
+import { UserResult } from "@garden-io/platform-api-types"
 
 const API_KEY = process.env.ANALYTICS_DEV ? SEGMENT_DEV_API_KEY : SEGMENT_PROD_API_KEY
 const CI_USER = "ci-user"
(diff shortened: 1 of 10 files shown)


core/src/analytics/analytics.ts
```diff
@@ -21,7 +21,7 @@ import dedent from "dedent"
 import { getGitHubUrl } from "../docs/common"
 import { Profile } from "../util/profiling"
 import { ModuleConfig } from "../config/module"
-import { UserResponse } from "@garden-io/platform-api-types"
+import { UserResult } from "@garden-io/platform-api-types"
```
eysi09, line 24: Instead of using `UserResult` here, we should import `UserResponse` and reference the results as `UserResponse["data"]`. 

Otherwise we're depending on an internal variable name that may change.

core/src/commands/cloud/helpers.ts
```diff
@@ -6,7 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import { SecretResponse, UserResponse } from "@garden-io/platform-api-types"
+import { SecretResult as SecretResultApi, UserResult as UserResultApi } from "@garden-io/platform-api-types"
```
eysi09, line 9: Let's also use the Response as opposed to the Result here.

core/src/commands/cloud/users/users-create.ts
```diff
@@ -7,7 +7,11 @@
  */
 
 import { CommandError, ConfigurationError } from "../../../exceptions"
-import { CreateUserBulkRequest, CreateUserBulkResponse, UserResponse } from "@garden-io/platform-api-types"
+import {
+  CreateUserBulkRequest,
+  CreateUserBulkResponse,
+  UserResult as UserResultApi,
```
eysi09, line 13: And here :)

eysi09 (changes requested):
A handful of small comments. Otherwise looks good!

eysi09 (approved):

PR #3435, opened 2022-12-12
Merge main into 0.13

Orzelius:
the commit [5f2a54a](https://github.com/garden-io/garden/pull/3435/commits/5f2a54afe5ade93553bd19ea17f2b02c76b47274) merges main to 0.13 and the next commit fixes the merge conflicts. A few commits that follow fix miscellaneous issues caused by the merge. I'm not sure why the commits from the main branch appear in the pr, in my git log it looks normal so I presume it's github being funny
image
But thanks to @Walther for helping with this!


core/package.json
```diff
@@ -26,7 +26,7 @@
   "dependencies": {
     "@codenamize/codenamize": "^1.1.1",
     "@hapi/joi": "git+https://github.com/garden-io/joi.git#master",
-    "@kubernetes/client-node": "^0.14.3",
+    "@kubernetes/client-node": "git+https://github.com/garden-io/javascript.git#master",
```
edvald, line 29: Did we actually end up needing this @Orzelius?

core/src/commands/deploy.ts
```diff
@@ -103,12 +103,13 @@ export class DeployCommand extends Command<Args, Opts> {
...
-        garden deploy --local              # deploys everything, enabling local mode on all applicable actions
-        garden deploy --env stage          # deploy to an environment called stage
-        garden deploy --skip service-b     # deploy everything except service-b
+        garden deploy --dev=my-service     # deploys all services, with dev mode enabled for my-service
+        garden deploy --dev                # deploys all compatible services with dev mode enabled
+        garden deploy --local=my-service   # deploys all services, with local mode enabled for my-service
+        garden deploy --local              # deploys all compatible services with local mode enabled
+        garden deploy --env stage          # deploy your services to an environment called stage
+        garden deploy --skip service-b     # deploy all services except service-b
+        garden deploy --forward            # deploy all services and start port forwards without watching for changes
```
edvald, line 112: This shouldn't have been overwritten fully. We deliberately are trying to remove the "service" terminology.

.circleci/config.yml
```diff
@@ -1025,17 +1026,16 @@ workflows:
 
       ### MAIN ONLY ###
 
-      # TODO: Re-enable this if we can figure out a way to not spam users who have starred the repo.
```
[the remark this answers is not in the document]
Orzelius: The only thing we figured out is that no direct notification is sent about edge releases and everybody else are doing them just the same way we do em so we turned them back on a while back

core/src/events.ts
```diff
@@ -305,58 +304,13 @@ export interface Events extends LoggerEvents {
...
-    build: boolean
-  }
-  setDeployOnWatch: {
-    serviceName: string
-    deploy: boolean
-  }
-  setTestOnWatch: {
-    moduleName: string
-    test: boolean
-  }
```
edvald, line 346: Was this all deliberately dropped?

core/src/plugins/kubernetes/api.ts
```diff
@@ -774,75 +776,66 @@ export class KubeApi {
       }
     }
 
-    const execHandler = new Exec(this.config, new WebSocketHandler(this.config))
+    const execHandler = new Exec(this.config)
 
-    return new Promise((resolve, reject) => {
+    return new Promise(async (resolve, reject) => {
```
edvald, line 781: Unrelated to the PR, but I believe this is not a good idea: https://stackoverflow.com/questions/43036229/is-it-an-anti-pattern-to-use-async-await-inside-of-a-new-promise-constructor

Not sure whom to ask for a fix but we should get on that :)

.circleci/config.yml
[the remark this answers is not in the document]
Orzelius: I don't see the draft releases in my github feed. I only see a full release if a person I'm following releases something or sometimes a I see a non draft release of a repo I've starred under the new "For you" section.

core/src/events.ts
Orzelius: Yes, this section needs to be rewritten. We discussed it a bit back in the dm-s on slack with you.

core/src/plugins/kubernetes/api.ts
Orzelius: Looks like it was added here https://github.com/garden-io/garden/pull/3395

I'll see if Tim feels like fixing it once he's back as it's not a super burning issue

core/src/commands/deploy.ts
Orzelius: This will probably conflict with your upcoming pr but I fixed the terminology here [be90709](https://github.com/garden-io/garden/pull/3435/commits/be90709e01627f9da6d8bbbd89784fbf6bcac04a)

edvald (approved):

PR #3443, opened 2022-12-14
chore(docs): update rbac guide for shared cluster

srihas-g:
RBAC config guide for scoping permissions when there are no dedicated Dev k8s clusters.

Fixes #

### DIFF
--- a/docs/advanced/rbac-config.md
+++ b/docs/advanced/rbac-config.md
@@ -3,173 +3,138 @@ order: 4
 title: Minimal RBAC Configuration for Development Clusters
 ---
 
-The following describes the minimal RBAC roles and permissions required for day-to-day use by developers for Garden when using the `kubernetes` plugin. These should be created along with the kubeconfig/kubecontext for the user in their namespace, replacing the `<username>`, `<service-accounts-namespace>` and `<project-namespace>` values as appropriate.
+If there is no dedicated development k8s cluster, the following configurations can be used to limit the permissions, of developers on a shared cluster using RBAC (Role-Based Access Control), to specific dev namespaces only. This can help to ensure that developers have the appropriate access to resources on the shared cluster.
 
-```
----
-# The user service account
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: user-<username>
-  namespace: <service-accounts-namespace>
+> We have verified to ensure that the following configuration is effective for an Azure IAM group that includes all developers. Similar approach can be followed for any other k8s group mapping. (Contributions, to this doc, are welcome)
 
----
+In order to achieve scoped permissions for desired namespaces only, we are using benefits of [hierarchical namespaces](https://kubernetes.io/blog/2020/08/14/introducing-hierarchical-namespaces/)
 
-# Project namespaces
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: <project-namespace>
-  # Some required annotations
-  annotations:
-    garden.io/version: "0.11.3"
+1. Install hierarchical namespace controller in the k8s cluster.
+    ```
+    kubectl apply -f https://github.com/kubernetes-sigs/hierarchical-namespaces/releases/download/v1.0.0/default.yaml
+    ```
+2.  Create an Azure IAM group for the developers who are allowed to deploy to dev namespaces only. Add all the Users to the group and note the group ID.
 
----
+3. Create a root namespace (`webdev-root` for example) which will have a role that can be inherited by sub-namespaces (all dev namespaces).
 
-# Allow reading namespaces and persistent volumes, which are cluster-scoped
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: user-<username>
-rules:
-- apiGroups: [""]
-  resources: ["namespaces", "persistentvolumes"]
-  verbs: ["get", "list"]
----
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
-metadata:
-  name: user-<username>
-  namespace: <service-accounts-namespace>
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: user-<username>
-subjects:
-- namespace: <service-accounts-namespace>
-  kind: ServiceAccount
-  name: user-<username>
+4. Create a clusterRole and ClusterRoleBinding.
+    ```
+    apiVersion: rbac.authorization.k8s.io/v1
+    kind: ClusterRole
+    metadata:
+      name: namespacesAndPVCsCreateAndList
+    rules:
+    - apiGroups:
+      - ""
+      resources:
+      - namespaces
+      - persistenvolumes
+      verbs:
+      - create
+      - get
+      - list
+      - watch
+    ---
 
----
+    apiVersion: rbac.authorization.k8s.io/v1
+    kind: ClusterRoleBinding
+    metadata:
+      name: webdevsClusterRoleBinding
+    roleRef:
+      apiGroup: rbac.authorization.k8s.io
+      kind: ClusterRole
+      name: namespacesAndPVCsCreateAndList
+    subjects:
+    - apiGroup: rbac.authorization.k8s.io
+      kind: Group
+      name: <Azure IAM group ID>
+    ```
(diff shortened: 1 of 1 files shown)


docs/advanced/rbac-config.md
```diff
@@ -3,173 +3,138 @@ order: 4
...
+
+In order to achieve scoped permissions for desired namespaces only, we are using benefits of [hierarchical namespaces](https://kubernetes.io/blog/2020/08/14/introducing-hierarchical-namespaces/)
+
+1. Install hierarchical namespace controller in the k8s cluster.
+    ```
+    kubectl apply -f https://github.com/kubernetes-sigs/hierarchical-namespaces/releases/download/v1.0.0/default.yaml
+    ```
+2.  Create an Azure IAM group for the developers who are allowed to deploy to dev namespaces only. Add all the Users to the group and note the group ID.
+
+3. Create a root namespace (`webdev-root` for example) which will have a role that can be inherited by sub-namespaces (all dev namespaces).
```
stefreak, line 18: maybe we could call it just `dev-root`? Then its also fine for non-webdev projects

stefreak: another option could be `garden-dev-root`

docs/advanced/rbac-config.md
```diff
@@ -3,173 +3,138 @@ order: 4
...
+
+> We have verified to ensure that the following configuration is effective for an Azure IAM group that includes all developers. Similar approach can be followed for any other k8s group mapping. (Contributions, to this doc, are welcome)
+
+In order to achieve scoped permissions for desired namespaces only, we are using benefits of [hierarchical namespaces](https://kubernetes.io/blog/2020/08/14/introducing-hierarchical-namespaces/)
+
+1. Install hierarchical namespace controller in the k8s cluster.
+    ```
+    kubectl apply -f https://github.com/kubernetes-sigs/hierarchical-namespaces/releases/download/v1.0.0/default.yaml
+    ```
+2.  Create an Azure IAM group for the developers who are allowed to deploy to dev namespaces only. Add all the Users to the group and note the group ID.
```
twelvemo, line 16:
```suggestion
2. Next we will need to take care to map your Cloud Provider's IAM groups or users to Kubernetes RBAC. This differs slightly between different cloud providers. The following should give you a brief overview, but might not cover all use cases. Please consider your Cloud Provider's documentation for further info.  
For Azure: Create an IAM group for the developers who are allowed to deploy to dev namespaces only. Add all the Users to the group and note the group ID. This group ID can be directly used in `roleBindings` and `clusterRoleBindings`. For more info see [here](https://learn.microsoft.com/en-us/azure/aks/azure-ad-rbac).
For AWS: Add your users to the [aws-auth configmap](https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html) and add them to an `apiGroup` that you want to use, we will call it `developers` here. This will be your group ID going forward.
For GCP: You can [create a Google group or add IAM users](https://cloud.google.com/kubernetes-engine/docs/how-to/role-based-access-control#rolebinding) directly to `roleBindings` and `clusterRoleBindings` We recommend creating a group, as adding single users to several `roleBindings` and `clusterRoleBindings` can become rather tedious.   
```

docs/advanced/rbac-config.md
```diff
@@ -3,173 +3,138 @@ order: 4
...
+    metadata:
+      name: webdevsClusterRoleBinding
+    roleRef:
+      apiGroup: rbac.authorization.k8s.io
+      kind: ClusterRole
+      name: namespacesAndPVCsCreateAndList
+    subjects:
+    - apiGroup: rbac.authorization.k8s.io
+      kind: Group
+      name: <Azure IAM group ID>
```
twelvemo, line 50:
```suggestion
      name: <group created in step 2>
```

docs/advanced/rbac-config.md
```diff
@@ -3,173 +3,138 @@ order: 4
...
+      name: webdevsRoleBinding
+      namespace: webdev-root
+    roleRef:
+      apiGroup: rbac.authorization.k8s.io
+      kind: Role
+      name: webdevNamespaceFullAccess
+    subjects:
+    - apiGroup: rbac.authorization.k8s.io
+      kind: Group
+      name: <Azure IAM group ID>
```
twelvemo, line 77:
```suggestion
      name: <group created in step 2>
```

docs/advanced/rbac-config.md
```diff
@@ -3,173 +3,138 @@ order: 4
...
+      name: gardenSystemFullAccessRoleBinding
+      namespace: garden-system
+    roleRef:
+      apiGroup: rbac.authorization.k8s.io
+      kind: Role
+      name: gardenSystemFullAccess
+    subjects:
+    - apiGroup: rbac.authorization.k8s.io
+      kind: Group
+      name: <Azure IAM group ID>
```
twelvemo, line 104:
```suggestion
      name: <group created in step 2>
```

docs/advanced/rbac-config.md
```diff
@@ -3,173 +3,138 @@ order: 4
...
+1. Install hierarchical namespace controller in the k8s cluster.
+    ```
+    kubectl apply -f https://github.com/kubernetes-sigs/hierarchical-namespaces/releases/download/v1.0.0/default.yaml
+    ```
+2.  Create an Azure IAM group for the developers who are allowed to deploy to dev namespaces only. Add all the Users to the group and note the group ID.
+
+3. Create a root namespace (`webdev-root` for example) which will have a role that can be inherited by sub-namespaces (all dev namespaces).
+
+4. Create a clusterRole and ClusterRoleBinding.
+    ```
```
twelvemo, line 21:
```suggestion
    ```
    # Allow reading namespaces and persistent volumes, which are cluster-scoped
```

docs/advanced/rbac-config.md
```diff
@@ -3,173 +3,138 @@ order: 4
...
+      kind: Role
+      name: gardenSystemFullAccess
+    subjects:
+    - apiGroup: rbac.authorization.k8s.io
+      kind: Group
+      name: <Azure IAM group ID>
+    ```
+
+7. Update project.garden.yaml in order to make sure that sub-namespaces have an annotation (`hnc.x-k8s.io/subnamespace-of: webdev-root` for example) of root namespace where permissions are inherited from.
+    ```
```
twelvemo, line 108: I think providing an example garden project is a great idea, but let's remove the unneeded parts like the local kubernetes env and providers and everything else that is not strictly required here.

twelvemo (changes requested):
Thanks for writing this up!
I would love this Guide to cover AWS, Azure and GCP and I think it is doable. I made some suggestions around that. What do you think?

docs/advanced/rbac-config.md
```diff
@@ -3,173 +3,138 @@ order: 4
...
-  name: user-<username>-kaniko
-  apiGroup: ""
-subjects:
-- namespace: <service-accounts-namespace>
-  kind: ServiceAccount
-  name: user-<username>
-```
+If there is no dedicated development k8s cluster, the following configurations can be used to limit the permissions, of developers on a shared cluster using RBAC (Role-Based Access Control), to specific dev namespaces only. This can help to ensure that developers have the appropriate access to resources on the shared cluster.
+
+> We have verified to ensure that the following configuration is effective for an Azure IAM group that includes all developers. Similar approach can be followed for any other k8s group mapping. (Contributions, to this doc, are welcome)
```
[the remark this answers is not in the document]
srihas-g: Before I go ahead and make changes, I feel we are repeating ourselves especially with this doc. 
In my opinion an optimal solution, to make this doc to be valid for all three major* cloud providers, is not keeping it in git book but rather having it as a README in three different example projects (per cloud provider) and having cross reference to that. This way we can also serve with updated config or so. It also makes quite easier to understand when we look at an example than going through the doc. Wdyt?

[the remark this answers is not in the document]
twelvemo: Sounds like a good idea too. My thinking was that we get the differences covered [here](https://github.com/garden-io/garden/pull/3443#discussion_r1049522425) and the rest can more or less stay as is. But the example projects are also a good idea, i will leave that to you :-)

eysi09:
This is pretty old. 

Can we resolve the conflicts and get this merged? (cc @twelvemo @srihas-g)

PR #3446, opened 2022-12-14
improvement(jib): configurable gradle binary

vvagaytsev:
This PR adds an option to make the `gradle` binary path configurable.
A similar change was done for `maven` in #3173.

There is a lot of code duplication in [`maven.ts`](https://github.com/garden-io/garden/blob/improvement/configurable-gradle-binary/plugins/jib/maven.ts) and [`gradle.ts`](https://github.com/garden-io/garden/blob/improvement/configurable-gradle-binary/plugins/jib/gradle.ts). Those parts will be refactored in a separate PR.

### DIFF
--- a/docs/reference/module-types/jib-container.md
+++ b/docs/reference/module-types/jib-container.md
@@ -101,6 +101,8 @@ build:
 
   # Defines the location of the custom executable Maven binary.
   #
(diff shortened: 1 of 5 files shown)


stefreak (approved):
good work 👍

PR #3462, opened 2022-12-20
improvement(cloud): automate remote project creation

mkhq:

This is an experimental integration with Garden Cloud that automates cloud project creation for a signed-in user.

Fixes #

### DIFF
--- a/core/src/cloud/api.ts
+++ b/core/src/cloud/api.ts
@@ -30,6 +30,8 @@ import { ProjectResource } from "../config/project"
 const gardenClientName = "garden-core"
 const gardenClientVersion = getPackageVersion()
 
+export class EnterpriseApiDuplicateProjectsError extends EnterpriseApiError {}
+
 // If a GARDEN_AUTH_TOKEN is present and Garden is NOT running from a workflow runner pod,
 // switch to ci-token authentication method.
 export const authTokenHeader =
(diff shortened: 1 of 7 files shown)


core/src/cloud/api.ts
```diff
@@ -323,15 +361,95 @@ export class CloudApi {
...
+
+      response = await this.post<CreateProjectsForRepoResponse>(`/projects/`, {
+        body: createRequest,
+      })
+    } catch (err) {
+      this.log.debug(`Create project request failed with error, ${err}`)
+      throw err
+    }
+
+    if (response.status === "error" || (response.status === "success" && response.data.length !== 1)) {
```
eysi09, line 446: Only check: `if (response.data.length !== 1` and log an informative and actionable error.

core/src/commands/cloud/secrets/secrets-list.ts
```diff
@@ -65,6 +65,10 @@ export class SecretsListCommand extends Command<{}, Opts> {
 
     const project = await api.getProject()
 
+    if (!project) {
+      throw new EnterpriseApiError(`Project ${garden.projectName} is not a Garden Cloud project`, {})
```
eysi09, line 69: Replace "Garden Cloud" with `${getCloudDistributionName(domain}`. Also in other places.

core/src/commands/cloud/users/users-list.ts
```diff
@@ -56,6 +56,11 @@ export class UsersListCommand extends Command<{}, Opts> {
     }
 
     const project = await api.getProject()
+
+    if (!project) {
+      throw new EnterpriseApiError(`Project ${garden.projectName} is not a Garden Cloud project`, {})
```
eysi09, line 61: Also update string.

core/src/cloud/api.ts
eysi09: Where we catch the error in the Garden class we need to check if it's a "duplicate project error", and if so, fail hard with, again, an actionable message for the user. 

This is what happened, this is what you need to do.

eysi09 (changes requested):

core/src/cloud/api.ts
mkhq: Opted for a warning message instead and continuing without a project configured since that's how other errors are handled when we cannot use a project.

eysi09 (approved):

PR #3481, opened 2023-01-02
fix: re-initialize providers changing environments

stefreak:
When users change environments, we need to re-initialize providers.

Let's consider the following example:

User has a provider config like so:
```
providers:
  - name: exec
    initScript: echo ${environment.name} > someFile
```

If user runs a garden command with --env foo, and then with --env bar,
and then again with --env foo we would expect `someFile` to contain `foo`. But
the file actually contains bar.

Reason for this is that garden skipped executing the initScript in the foo env,
because it has already been executed. But the cache logic is not aware that the
last command that ran was in the bar env.

This commit fixes that problem.

Fixes #

### DIFF
--- a/core/test/integ/src/plugins/exec/resolve.ts
+++ b/core/test/integ/src/plugins/exec/resolve.ts
@@ -14,49 +14,48 @@ import { getDataDir, makeTestGarden, TestGarden } from "../../../../helpers"
 describe("exec provider initialization cache behaviour", () => {
   let gardenOne: TestGarden
   let tmpDir: string
+  let fileLocation: string
 
   beforeEach(async () => {
     gardenOne = await makeTestGarden(getDataDir("exec-provider-cache"), { environmentName: "one" })
-    tmpDir = await gardenOne.getRepoRoot()
+
+    tmpDir = join(await gardenOne.getRepoRoot(), "project")
+    fileLocation = join(tmpDir, "theFile")
+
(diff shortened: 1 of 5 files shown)


core/test/integ/src/plugins/exec/resolve.ts
```diff
@@ -0,0 +1,54 @@
...
+  let gardenOne: TestGarden
+  let tmpDir: string
+
+  beforeEach(async () => {
+    gardenOne = await makeTestGarden(getDataDir("exec-provider-cache"), { environmentName: "one" })
+    tmpDir = await gardenOne.getRepoRoot()
+  })
+
+  it("writes the environment name to theFile as configured in the initScript", async () => {
+    await gardenOne.resolveProvider(gardenOne.log, "exec")
```
Walther, line 16: nit: we can probably move this to `beforeEach` as well, as every test starts with verifying the initial `one` on the log

stefreak:
I wonder if the new test will just run on circleci, or if I have to change the circleci yaml to include the new `exec` integration test somehow?

Orzelius (approved):
Looks good! The integ test is very easy to understand.

edvald:
The integ test should just run, along with the rest of the integ test suite. You should be able to confirm by looking at the vm test logs.

edvald (approved):

PR #3482, opened 2023-01-03
Fix solver issues and deploy command tests

edvald:
See individual commits for detail. This includes some related fixes improvements used to identify the root issue.


core/test/unit/src/tasks/base.ts
```diff
@@ -0,0 +1,134 @@
...
+          force: true,
+          forceBuild: false,
+          skipRuntimeDependencies: true, // <-----
+          devModeDeployNames: [],
+          localModeDeployNames: [],
+          fromWatch: false,
+        })
+
+        const deps = task.resolveProcessDependencies({ status: null })
+        expect(deps.find((dep) => dep.type === "deploy" || dep.type === "task")).to.be.undefined
```
thsig, line 130: Shouldn't this be `dep.type === "run"`?

core/test/unit/src/tasks/base.ts
```diff
@@ -0,0 +1,134 @@
...
+        fromWatch: false,
+      })
+
+      const deps = task.resolveProcessDependencies({ status: { state: "ready", outputs: {} } })
+
+      expect(deps.map((d) => d.getBaseKey())).to.eql(["resolve-action.test.module-a-integ"])
+    })
+
+    context("when skipRuntimeDependencies = true", () => {
+      it("doesn't return deploy or task dependencies", async () => {
```
thsig, line 113: "doesn't return deploy or task dependencies" -> "doesn't return Deploy or Run dependencies"

core/test/unit/src/commands/deploy.ts
```diff
@@ -72,14 +59,15 @@ const testProvider = () => {
...
-          schema: execDeployActionSchema(),
+          schema: testDeploySchema(),
           handlers: {
             deploy: async (params) => {
               const newStatus: ActionStatus = { state: "ready", detail: { state: "ready", detail: {} }, outputs: {} }
               testStatuses[params.action.name] = newStatus
               return newStatus
             },
             getStatus: async (params) => {
+              console.log("GET STATUS", params.action.name)
```
thsig, line 70: Remove `console.log`?

core/test/unit/src/commands/deploy.ts
```diff
@@ -88,35 +76,47 @@ const testProvider = () => {
...
-  const plugins = [testProvider()]
   const projectRootB = join(dataDir, "test-project-b")
   const projectRootA = join(dataDir, "test-project-a")
 
   // TODO: Verify that services don't get redeployed when same version is already deployed.
   // TODO: Test with --watch flag
 
-  it("should build and deploy all modules in a project", async () => {
-    const garden = await makeTestGarden(projectRootB, { plugins })
+  it("should build and deploy all modules in a project, and run any task dependencies", async () => {
```
thsig, line 118: "run any task dependencies" -> "execute any Run dependencies"

core/test/unit/src/commands/deploy.ts
```diff
@@ -247,13 +174,13 @@ describe("DeployCommand", () => {
...
-    const deployServiceBUid = getDeployUid("service-b")
-    const deployServiceCUid = getDeployUid("service-c")
-    const deployServiceDUid = getDeployUid("service-d")
+    const deployServiceAUid = getActionUid("service-a")
+    const deployServiceBUid = getActionUid("service-b")
+    const deployServiceDUid = getActionUid("service-d")
 
-    const runTaskAUid = getRunTaskUid("task-a")
-    const runTaskCUid = getRunTaskUid("task-c")
+    // Note: tasks A and C should not run or be queried for status because service-a is ready beforehand
```
thsig, line 181: "tasks A and C" -> "Runs A and C"

... we still haven't renamed the Runs from `task-a` etc. in our example projects, so it's easy to slip back into the old, familiar nomenclature (calling them tasks instead of Runs).

core/test/unit/src/graph/solver.ts
```diff
@@ -203,6 +203,26 @@ describe("GraphSolver", () => {
     expect(resultB?.outputs.callbackResult).to.equal(taskA.getId())
   })
 
+  it("processes a complex graph correctly", async () => {
+    throw "TODO"
```
thsig, line 207: Should these be commented out?

Or was the intent to introduce new, failing test cases as a reminder of work to be done in subsequent PRs?

thsig (commented):
Good stuff! Had a few comments there.

core/test/unit/src/commands/deploy.ts
edvald: Yeah that's the thing, I haven't been chasing all that down but it's good to make those edits as we come across them.

core/test/unit/src/tasks/base.ts
edvald: That's true, good catch.

core/test/unit/src/graph/solver.ts
edvald: No I opted to leave these in. Basically this is all implicitly tested elsewhere but I wanted to make more explicit and narrow tests and these should be done before we sign off on a release imo.

core/src/actions/base.ts
```diff
@@ -337,10 +337,10 @@ export abstract class BaseAction<C extends BaseActionConfig = BaseActionConfig,
    * Verbose string description of the action. Useful for logging and error messages.
    */
   longDescription(): string {
-    let d = `${this.type} ${this.kind} ${chalk.bold.white(this.name)}`
+    let d = `${chalk.white(this.kind)} type=${chalk.bold.white(this.type)} name=${chalk.bold.white(this.name)}`
 
     if (this._moduleName) {
-      d += `(from module ${chalk.bold.white(this.name)})`
+      d += ` (from module ${chalk.bold.white(this.name)})`
```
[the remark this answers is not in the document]
edvald: 'tis

core/test/unit/src/tasks/base.ts
edvald: Actually noticing that this should also include Test dependencies, will fix that as well.

core/src/router/build.ts
```diff
@@ -26,7 +26,8 @@ export const buildRouter = (baseParams: BaseRouterParams) =>
         defaultHandler: async () => ({ state: <ActionState>"unknown", detail: {}, outputs: {} }),
       })
 
-      await router.validateActionOutputs(action, status.outputs)
+      // TODO-G2: only validate if state is ready?
```
thsig, line 29: I think it could make sense to only validate if the status indicates success—it's easy to imagine build handlers that only return valid values for certain output fields if the build succeeded (e.g. an artifact ID for the build in some third-party system, which doesn't get generated if the build failed).

thsig (approved):
Good stuff! Had one comment/suggestion there, but feel free to merge.

core/src/router/build.ts
edvald: Yeah, there's some nuance to consider but this may well be correct. I'll follow up on it separately though.

PR #3483, opened 2023-01-03
2023 01 03 tiny fixes

Walther:

Small improvements to kubernetes logging. This should make reading CI logs and other `-l5` level logs much easier, and significantly reduce the amount of requests we are making to the kubernetes api for the logs during service startup.

N/A

WIP do not merge yet

### DIFF
--- a/core/src/plugins/kubernetes/logs.ts
+++ b/core/src/plugins/kubernetes/logs.ts
@@ -231,6 +231,7 @@ export class K8sLogFollower<T> {
     await this.createConnections(opts)
 
     this.intervalId = setInterval(async () => {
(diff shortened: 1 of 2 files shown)


core/src/plugins/kubernetes/logs.ts
```diff
@@ -231,6 +231,7 @@ export class K8sLogFollower<T> {
     await this.createConnections(opts)
 
     this.intervalId = setInterval(async () => {
+      this.log.silly(`calling createConnections again with retryIntervalMs: ${this.retryIntervalMs}`)
```
Walther, line 234: should be removed; this was how i found out the retry interval was 10ms

Walther:
Closing, we should revisit this later.

PR #3485, opened 2023-01-03
chore(ci): optimize dist build and (hopefully) fix flaky pkg-fetch problem

edvald:
Mostly attempting to fix the flaky dist build step in CI, but spotted an opportunity to do some caching along the way.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -231,11 +231,14 @@ commands:
       - npm_install
       - *attach-workspace
       - run:
-          environment:
-            # pkg doesn't use a normal cache directory by default
-            # see https://github.com/vercel/pkg/issues/1225
-            PKG_CACHE_PATH: tmp/pkg/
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -231,18 +231,29 @@ commands:
...
+            - pkg-v1-{{ checksum ".pkgversion" }}
+      - run:
+          name: Run dist script
           command: yarn run dist --version "<<parameters.version>>"
       - persist_to_workspace:
           root: ./
           paths:
             - dist/
             # Needed for the alpine docker build
             - tmp/pkg/
```
Walther, line 248: does this need to be changed as well, or is this a separate thing?
```suggestion
            - tmp/pkg-fetch/
```

Walther (approved):
thanks for the optimization! let's hope this makes `build-dist` more reliable 🤞

PR #3486, opened 2023-01-03
improvement(cli): trim binary size by approx 1/3

edvald:
This is a cheap improvement enabled by a compression feature in pkg.

pkg docs suggest this would generally improve startup time but in my manual testing I could find no meaningful changes on that front in either direction, but that's all good.


Walther (approved):
nice!

thsig (approved):

PR #3492, opened 2023-01-04
ci(edge-release): update existing instead of re-create

stefreak:

To avoid spamming our fellow coders GitHub timelines, update the
existing edge release with new binaries and a new SHA target instead
of creating a new edge release every time.

Fixes #


Walther (approved):
looks good to me 👍
and as discussed, noting that the docker image tag needs to be manually pushed before merging

images/circleci-release/garden.yml
```diff
@@ -0,0 +1,6 @@
+kind: Module
+type: container
+name: circleci-release
```
edvald, line 3: Might I suggest just using the `circleci-runner` for this? Seems like unnecessary overhead to add a while new image. Also curious why we can't use an upstream image to begin with?

stefreak: Why a custom image is necessary: If we update the existing release, we need another utility called `gh` that is not included together with `ghr` in any upstream container I looked at.

I can change this to circleci-runner if you prefer

edvald: Aha I see, makes sense. But yeah, I think just the one CircleCI image makes most sense.

stefreak: OK, I'll change that

stefreak: @edvald OK, changed it now. But to be able to upload the image I need permission on dockerhub

stefreak: CI ran with the new image now, feel free to approve and then I'd just merge this and we could look at whether this actually works

edvald:
Can't say if it's related, but the `test-windows` check seems to fail reliably. Could you take a look at the @stefreak? We'll want to fix it whether it's related to these changes or not.

stefreak:
@edvald this test also fails on main: https://app.circleci.com/pipelines/github/garden-io/garden/14309/workflows/e38043f4-eb14-4713-aa31-06edbc541e70/jobs/242005

Walther:
> Can't say if it's related, but the `test-windows` check seems to fail reliably. Could you take a look at the @stefreak? We'll want to fix it whether it's related to these changes or not.

gcloud auth plugin component missing, PR here https://github.com/garden-io/garden/pull/3519

edvald:
Yep all good! PR looks solid, just wanted to check for awareness on that.

Sent via Superhuman iOS ( ***@***.*** )

vvagaytsev (approved):

Issue #3468, opened 2022-12-30
Pulumi outputs don't display to the user on `garden deploy`


worldofgeese:
## Bug

I have defined a Cloud Run application in Pulumi over [here](https://github.com/worldofgeese/flask-cockroachdb-devcontainer/blob/gardenify/pulumi/__main__.py). `pulumi.export("url", default.statuses[0].url)` exports the URL to access the container service but it's never displayed to the user.

### Current Behavior

 `garden deploy` runs but the output values are never displayed:

![2022-12-30_14h04_26](https://user-images.githubusercontent.com/59834693/210073656-81e7d163-ab02-466d-b0fe-de0940bcd2e3.gif)

### Expected behavior

`garden deploy` should, by default, output all Pulumi outputs.

### Reproducible example

A full-fat example is available at https://github.com/worldofgeese/flask-cockroachdb-devcontainer/tree/gardenify. It may require just a few modifications around `deploymentRegistry` to work.

### Workaround

Browsing to the Cloud Run console from a web browser to fetch the URL. I assume a `gcloud` invocation would also work!

### Your environment

* **OS:** Windows 11
* **How I'm running Kubernetes:** GKE

`garden version`
0.12.48

stefreak:
@worldofgeese pulumi output should be logged now with `garden deploy -l3` (which might become the new default in #3254) – can you verify?

PR #3493, opened 2023-01-05
refactor(pulumi): remove dead code

Walther:
tiny cleanup, the `serviceStatus` block below handles the same case in the same way.

found this while reading through the code when debugging https://github.com/garden-io/garden/issues/3468 and thought might as well remove it, even if it doesn't really matter. we can also let it be, if we prefer.


Orzelius (approved):
Your ability to spot such issues amazes me

thsig:
Good catch! Something must have moved around as I refactored/changed the logic around this back in the day.

PR #3498, opened 2023-01-09
Allow to specify commit for remote sources

### LINKED ISSUE #3172: [FEATURE]: use git SHA in remote source declaration
### Background / Motivation

We are running a poly repo setup.
The `project.garden.yml` lives in an umbrella repository which declares all other repos as remote sources.
In our CI pipeline of the individual repos we trigger a child pipeline in the umbrella repo where `garden test && garden deploy` is run.
When triggering this child pipeline we pass a specific context as payload to the umbrella pipeline:
- which repo has triggered this pipeline run
- from which git SHA was this pipeline triggered

The idea is to integrate the specific git commit which triggered the pipeline with the other repos main development branches.

We'd like to use garden templating to adjust the remote source definition of the project.
In the remote source declaration we would like to be able to point to a git SHA.
Currently, only branches are allowed.

This way we could use gardens templating capabilities to use a specific git SHA instead of the main development branch in our CI pipeline.

Our current workaround is in our pipeline to clone the specific repository manually, checkout the required git SHA and use `garden link`.

🌹 It’s a nice to have, but nice things are nice

Orzelius:
Will eventually fix #3172

I'm sorry :(

### DIFF
--- a/core/test/unit/src/vcs/git.ts
+++ b/core/test/unit/src/vcs/git.ts
@@ -34,13 +34,19 @@ async function commit(msg: string, repoPath: string) {
   await createFile(filePath)
   await execa("git", ["add", filePath], { cwd: repoPath })
   await execa("git", ["commit", "-m", msg], { cwd: repoPath })
-  return uniqueFilename
+  const commitSHA = (await execa("git", ["rev-parse", "HEAD"], { cwd: repoPath })).stdout
+  return { uniqueFilename, commitSHA }
+}
+
+async function createGitTag(tag: string, message: string, repoPath: string) {
+  await execa("git", ["tag", "-a", tag, "-m", message], { cwd: repoPath })
 }
 
 async function makeTempGitRepo() {
   const tmpDir = await tmp.dir({ unsafeCleanup: true })
   const tmpPath = await realpath(tmpDir.path)
   await execa("git", ["init", "--initial-branch=main"], { cwd: tmpPath })
+  // await execa("git", ["config", "--global", "protocol.file.allow", "always"], { cwd: tmpPath })
 
   return tmpDir
 }
@@ -65,7 +71,12 @@ describe("GitHandler", () => {
     log = garden.log
     tmpDir = await makeTempGitRepo()
     tmpPath = await realpath(tmpDir.path)
-    handler = new GitHandler(tmpPath, join(tmpPath, ".garden"), [defaultIgnoreFilename], garden.cache)
+    handler = new GitHandler(
+      tmpPath,
+      join(tmpPath, ".garden"),
+      [defaultIgnoreFilename, "-c", "protocol.file.allow=always"],
+      garden.cache
+    )
     git = (<any>handler).gitCli(log, tmpPath)
   })
 
@@ -486,7 +497,7 @@ describe("GitHandler", () => {
       beforeEach(async () => {
         submodule = await makeTempGitRepo()
         submodulePath = await realpath(submodule.path)
-        initFile = await commit("init", submodulePath)
+        initFile = (await commit("init", submodulePath)).uniqueFilename
 
         await execa("git", ["submodule", "add", "--force", "--", submodulePath, "sub"], { cwd: tmpPath })
         await execa("git", ["commit", "-m", "add submodule"], { cwd: tmpPath })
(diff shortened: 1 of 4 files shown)


core/test/unit/src/vcs/git.ts
```diff
@@ -65,7 +71,12 @@ describe("GitHandler", () => {
     log = garden.log
     tmpDir = await makeTempGitRepo()
     tmpPath = await realpath(tmpDir.path)
-    handler = new GitHandler(tmpPath, join(tmpPath, ".garden"), [defaultIgnoreFilename], garden.cache)
+    handler = new GitHandler(
+      tmpPath,
+      join(tmpPath, ".garden"),
+      [defaultIgnoreFilename, "-c", "protocol.file.allow=always"],
```
Orzelius, line 77: The file protocol is disabled by default on newer git versions so our tests started failing

core/test/unit/src/vcs/git.ts
```diff
@@ -34,13 +34,19 @@ async function commit(msg: string, repoPath: string) {
...
+
+async function createGitTag(tag: string, message: string, repoPath: string) {
+  await execa("git", ["tag", "-a", tag, "-m", message], { cwd: repoPath })
 }
 
 async function makeTempGitRepo() {
   const tmpDir = await tmp.dir({ unsafeCleanup: true })
   const tmpPath = await realpath(tmpDir.path)
   await execa("git", ["init", "--initial-branch=main"], { cwd: tmpPath })
+  // await execa("git", ["config", "--global", "protocol.file.allow", "always"], { cwd: tmpPath })
```
eysi09, line 49: Remove comment

Walther:
Sidenote, some example prior art in package managers that allow specifying git revision for remote sources:
- `npm` https://github.com/npm/git/blob/42826b692719de97d98eb3624cfd5094ddc5a48a/lib/clone.js#L90-L105
- `cargo` https://github.com/rust-lang/cargo/blob/abc6abe34d28249a556ac17593a4951519b59f30/src/cargo/sources/git/utils.rs

Walther (dismissed):

Walther (approved):

core/test/unit/src/vcs/git.ts
vvagaytsev: Sorry for the late notification. It looked fine at a glance, but I've got some findings after having a closer look.

**Findings.**
This might not fix the issue completely. The array `[defaultIgnoreFilename, "-c", "protocol.file.allow=always"]` will be stored in the `ignoreFiles[]: string` field of the `VcsHandler` class. Here is what I have found so far:
* I've checked its usage and it doesn't seem to participate in any Git CLI calls.
* The Git config options should not be stored as ignored filenames.
* This change also brings some semantical conflicts with the `0.13` changes that were done in #3069

**Ideas on failure cause.**
I tried to run the tests locally with Git 2.39.0 on macOS and got the same errors about the file transport. It looks like the tests are failing because we still use the `execa("git", "submodule", ...)` commands to build the test Git repos.

**Ideas on fixes.**
I can see some quick fixes for this:
* We can just change all `execa("git", "submodule", ...)` commands to be `execa("git", "-c", "protocol.file.allow=always", "submodule", ...)` in all places where we configure the test data/repos. (We can also extract a helper function for that).
* We can ensure that the file protocol is allowed in all  `GitHandler.gitCli` calls (something like [this](https://github.com/garden-io/garden/commit/3eb32cbc3342d9c6eb804e0ce5a744dfb4f2195e)) and use [`GitHandler.getCli`](https://github.com/garden-io/garden/blob/9b3c0a5cb22bdb7574644c174d5c36c0d4940473/core/test/unit/src/vcs/git.ts#L80) to in all submodule-related actions while test data preparation.

It might be not the safest way to go to silently allow `file.protocol` by default. So, the first option might be a better one.

I'm wondering why CI tests passed w/o errors. Unfortunately, I can't see what Git version was used there.

Orzelius: > The Git config options should not be stored as ignored filenames.

ah, I didn't notice that. Good catch

> why CI tests passed w/o errors

I think the git version isn't just new enough yet

> I tried to run the tests locally with Git 2.39.0 on macOS and got the same errors about the file transport.

hmm that's odd, they run just fine on my machine with the same git version

vvagaytsev: > hmm that's odd, they run just fine on my machine with the same git version

Did you have `protocol.file.allow=always` in your global `.gitconfig`?

hnicke:
After #3873 has been implemented I was able to test #3172.
It seems to work well with long git SHAs.
Unfortunately, short SHAs don't work:
```
Downloading remote project failed with error: 
Error: Command "git -c protocol.file.allow=always clone --recursive --depth=1 --shallow-submodules --branch=442cb8443f4c git@bitbucket.org:xxxx/xxxxx.git ." failed with code 128:
Cloning into '.'...
Warning: Permanently added the ECDSA host key for IP address 'xxxxxxxxx' to the list of known hosts.
warning: Could not find remote branch 442cb8443f4c to clone.
fatal: Remote branch 442cb8443f4c not found in upstream origin
fatal: the remote end hung up unexpectedly
```

Would it be possible to support short SHAs as well?
I am aware that it's possible to derive the long SHA from the short SHA by running `git rev-parse $SHORT_SHA` - however this requires the repo to be checked out manually.

PR #3505, opened 2023-01-12
Run and Test command rework

edvald:
This makes several changes to commands and fixes related issues uncovered in the process. The idea is for `garden run ...` to no longer be a category of commands, and instead we get `garden run` as a command of its own, analogous to the build, deploy and test commands.

Individual commits reveal more detail but here's a summary of changes:

* The `run service/deploy` command has been removed.
* The `run module/build` command has been removed.
* The `run workflow` command has been renamed to `run-workflow` (note the dash).
* The `run test` command has been removed and its specific functionality moved to the `test` command (mainly the `--interactive` flag).
* The `run task` command is now simply `run`.
* Fixed several issues related to the `test` command. It now works properly for e.g. the `demo-project` example.

There's more work to be done around unit tests and updating the Run command, but it'd be good to get this checkpoint merged.


core/src/commands/get/get-test-result.ts
```diff
@@ -12,10 +12,24 @@ import { printHeader } from "../../logger/util"
...
+import { ParameterValues, StringParameter } from "../../cli/params"
+import { ParameterError } from "../../exceptions"
+import { ConfigGraph } from "../../graph/config-graph"
+import { GardenModule, moduleTestNameToActionName } from "../../types/module"
+import { findByName, getNames } from "../../util/util"
+
+const getTestResultArgs = {
+  name: new StringParameter({
+    help:
+      "The test to run. If using modules, specify the module name here and the test name from the module in the second argument",
```
thsig, line 25:
```suggestion
      "The name of the test. If this test belongs to a module, specify the module name here instead, and specify the test name from the module in the second argument.",
```
Suggesting this change since here we're fetching the result, not running the test.

core/src/commands/get/get-test-result.ts
```diff
@@ -12,10 +12,24 @@ import { printHeader } from "../../logger/util"
...
+    required: true,
+  }),
+  moduleTestName: new StringParameter({
+    help: "The name of the test to run in a module.",
```
thsig, line 29:
```suggestion
    help: "When the test belongs to a module, specify its name here (i.e. as the second argument).",
```

core/src/commands/run.ts
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  async action({ garden, log, footerLog, args, opts }: CommandParams<Args, Opts>) {
+    this.garden = garden
+
+    if (this.server) {
+      this.server.setGarden(garden)
+    }
+
+    // if (opts.interactive && opts.watch) {
```
thsig, line 132: Remove commented-out code.

core/src/config/task.ts
```diff
@@ -25,7 +25,7 @@ export const cacheResultSchema = () =>
     .default(true)
     .description(
       dedent`
-    Set to false if you don't want the task's result to be cached. Use this if the task needs to be run any time your project (or one or more of the task's dependants) is deployed. Otherwise the task is only re-run when its version changes (i.e. the module or one of its dependencies is modified), or when you run \`garden run task\`.
+    Set to false if you don't want the task's result to be cached. Use this if the task needs to be run any time your project (or one or more of the task's dependants) is deployed. Otherwise the task is only re-run when its version changes (i.e. the module or one of its dependencies is modified), or when you run \`garden run\`.
```
thsig, line 28: Out of scope for this PR, but since the line appears in the diff I thought I'd ask the question:

Should we start referring to tasks as Runs and services as Deploys in the docstrings for the old module-related schemas that use the old nomenclature?

I think it may be clearer for the user to just always say Deploys and Runs, and to speak of "tasks" and "services" as deprecated aliases for Deploys / Runs.

core/src/plugin/base.ts
```diff
@@ -141,18 +124,18 @@ export interface RunResult {
...
-      exitCode: joi.number().integer().description("The exit code of the run (if applicable)."),
-      startedAt: joi.date().required().description("When the module run was started."),
-      completedAt: joi.date().required().description("When the module run was completed."),
-      log: joi.string().allow("").default("").description("The output log from the run."),
-      namespaceStatus: namespaceStatusSchema().optional(),
-    })
+export const runResultSchema = createSchema({
+  name: "run-result",
+  keys: {
+    success: joi.boolean().required().description("Whether the module was successfully run."),
```
thsig, line 130:
```suggestion
    success: joi.boolean().required().description("Whether the Run was successful."),
```
Also: We should probably override the description strings in `testResultSchema` to refer to Tests instead of Runs, right?

core/src/plugin/base.ts
```diff
@@ -141,18 +124,18 @@ export interface RunResult {
...
+    exitCode: joi.number().integer().description("The exit code of the run (if applicable)."),
+    startedAt: joi.date().required().description("When the module run was started."),
```
thsig, line 132:
```suggestion
    startedAt: joi.date().required().description("When the Run was started."),
```

core/src/plugin/base.ts
```diff
@@ -141,18 +124,18 @@ export interface RunResult {
...
+    completedAt: joi.date().required().description("When the module run was completed."),
```
thsig, line 133:
```suggestion
    completedAt: joi.date().required().description("When the Run was completed."),
```

core/src/plugin/base.ts
```diff
@@ -141,18 +124,18 @@ export interface RunResult {
...
+    log: joi.string().allow("").default("").description("The output log from the run."),
```
thsig, line 134:
```suggestion
    log: joi.string().allow("").default("").description("The output log from the Run."),
```

core/src/commands/run.ts
```diff
@@ -0,0 +1,292 @@
...
+          fromWatch: false,
+          action,
+          devModeDeployNames: [],
+          localModeDeployNames: [],
+          skipRuntimeDependencies,
+          // interactive: opts.interactive,
+        })
+    )
+
+    // if (opts.interactive && initialTasks.length !== 1) {
```
thsig, line 253: Remove commented-out code.

thsig (changes requested):
Great stuff! Feels good to see the command list tidied up.

Had a few minor comments there (mostly text-related).

core/src/commands/run.ts
```diff
@@ -0,0 +1,292 @@
...
+}
+
+const runOpts = {
+  "force": new BooleanParameter({
+    help: "Run even if the action is disabled for the environment, and/or a successful result is found in cache.",
+  }),
+  "force-build": new BooleanParameter({
+    help: "Force re-build of Build dependencies before running.",
+  }),
+  // "interactive": new BooleanParameter({
```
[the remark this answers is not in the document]
edvald: I'd actually like to keep this. It related to a TODO comment above.

core/src/commands/run.ts
edvald: Same as last comment, keeping it around.

core/src/config/task.ts
edvald: I agree with that. I haven't been chasing everything as I go, more opportunistically. I figure it's best if we starting tracking this down more deliberately once most changes are settled down.

core/src/plugin/base.ts
edvald: This schema doesn't actually refer specifically or only to the `Run` action type. It is also used for Test actions.

core/src/plugin/base.ts
edvald: This is fine as-is (see prior comment).

thsig (approved):

core/src/commands/test.ts
```diff
@@ -161,24 +168,28 @@ export class TestCommand extends Command<Args, Opts> {
...
 
-    const filterNames = opts.name || []
     const force = opts.force
     const skipRuntimeDependencies = opts["skip-dependencies"]
     const skipped = opts.skip || []
 
     const actions = getTestActions({ graph, modules, filterNames })
 
+    console.log(filterNames)
+
```
Orzelius, line 179: should probably delete this

Issue #2551, opened 2021-08-18
Deprecate cert-manager integration


10ko:
### Current Behavior

The current `cert-manager` version supported in the Garden integration is way too old (`v0.11.0`) and needs an update.

### Expected behavior

As a user I want to be able to generate certificates using the latest version of `cert-manager`
### Reproducible example

N/A

### Workaround

You could potentially install `cert-manager` manually and use it to manage tlsCertificates but we haven't tested with the latest `cert-manager` version.

### Your environment

* **OS:** 
* **How I'm running Kubernetes:**  

`garden version`

ohookins:
Maybe a first step on the path to deprecation would be to fix this:

```
Failed resolving provider kubernetes. Here is the output:
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Error validating provider configuration: key .certManager.email is required
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Failed resolving one or more providers:
- kubernetes
```

This happens when you set `certManager.install: false`

ohookins:
Probably not worth solving if certmanager is going to be removed anyway. Stalebot can close it.

Walther:
Deprecated in #3517 
Fully deleted in #4085
Closing this issue as completed.

PR #3517, opened 2023-01-16
chore: mark cert-manager integration as deprecated

Walther:

Marking `cert-manager` as deprecated and warning on the use of it.

Related to https://github.com/garden-io/garden/issues/2551 but not closing it.
Let's close it during 0.13 when it will be actually removed.

This has been asked about in the community Discord as well, so we should definitely get these docs up to date.

### DIFF
--- a/core/src/plugins/kubernetes/integrations/cert-manager.ts
+++ b/core/src/plugins/kubernetes/integrations/cert-manager.ts
@@ -176,6 +176,11 @@ export async function checkCertManagerStatus({
   provider: KubernetesProvider
   namespace?: string
 }): Promise<ServiceState> {
+  log.warn({
+    symbol: "warning",
+    msg: "The cert-manager integration is deprecated and will be removed in a future release",
+  })
+
   const api = await KubeApi.factory(log, ctx, provider)
   const systemPods = await api.core.listNamespacedPod(namespace)
   const certManagerPods: KubernetesServerResource<V1Pod>[] = []
--- a/core/src/plugins/kubernetes/integrations/cert-manager.ts
+++ b/core/src/plugins/kubernetes/integrations/cert-manager.ts
@@ -26,6 +26,7 @@ import { EnvironmentStatus } from "../../../types/plugin/provider/getEnvironment
 import { PrimitiveMap } from "../../../config/common"
 import chalk from "chalk"
 import { defaultIngressClass } from "../constants"
+import { emitWarning } from "../../../warnings"
 
 /**
  * Given an array of certificate names, check if they are all existing and Ready.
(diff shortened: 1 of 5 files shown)


vvagaytsev:
Looks good! 👍  One general question - should we state that it will be removed exactly in the `0.13` release instead of just a future release?

core/src/plugins/kubernetes/integrations/cert-manager.ts
```diff
@@ -176,6 +176,11 @@ export async function checkCertManagerStatus({
   provider: KubernetesProvider
   namespace?: string
 }): Promise<ServiceState> {
+  log.warn({
+    symbol: "warning",
+    msg: "The cert-manager integration is deprecated and will be removed in a future release",
```
vvagaytsev, line 181: In some (or many?) places we use `chalk.yellow` to emphasize deprecation warnings.

Also, if the warning is printed too often, it might be ok to use [`emitWarning`](https://github.com/garden-io/garden/blob/main/core/src/warnings.ts#L11) function. But it might be slow as it seems to make some DB calls.

In-memory caching of warnings was implemented in #3069, see [`warnings.ts`](https://github.com/garden-io/garden/pull/3069/files#diff-504b46113a07411817847034fbd1ce40f2773a1916587da38aca7c68f6eca047) file. But it's available only in the `0.13` branch now. If this warning is not printed too often, let's just avoid any extra complexity and caching.

vvagaytsev: Yeah, makes sense. I remember the `emitWarning` didn't work for my case for some reason, so I came up with having a lightweight in-memory caching :)

docs/reference/providers/kubernetes.md
```diff
@@ -2541,6 +2515,10 @@ The namespace where the secret is stored. If necessary, the secret may be copied
 
 [providers](#providers) > [tlsCertificates](#providerstlscertificates) > managedBy
 
+{% hint style="warning" %}
+**Deprecated**: This field will be removed in a future release.
```
vvagaytsev, line 2519: The generated docs seem to have old values. Please re-generate them as well :)

Walther: The docs have been regenerated - this specific message is probably coming from https://github.com/garden-io/garden/blob/78ddb509e932d3041234c5af81062f6954158353/core/src/docs/config.ts#L94

vvagaytsev: Good catch! We can fix it in a separate PR.

vvagaytsev (approved):
Great job, thank you! 🚀

Issue #3506, opened 2023-01-12
SGv2: Some trivial fixes in error message assertions


vvagaytsev:
Some error message assertions should be fixed:

* [x] base.ts
* [x] common.ts
* [x] module-template.ts
* [x] project.ts

Let's use fuzzy-matching comparison like in  https://github.com/garden-io/garden/blob/0.13/core/test/unit/src/config/base.ts#L170 instead of `expect(stripAnsi(err.message)).to.equal("...")`

Walther:
Fixed in https://github.com/garden-io/garden/pull/3527 (not sure why this wasn't auto-closed)

PR #3527, opened 2023-01-17
test: unit test assertion fixes on config directory

Walther:

Fixes https://github.com/garden-io/garden/issues/3506

### DIFF
--- a/core/test/unit/src/config/base.ts
+++ b/core/test/unit/src/config/base.ts
@@ -120,7 +120,7 @@ describe("loadConfigResources", () => {
     await expectError(
       async () => await loadConfigResources(projectPath, resolve(projectPath, "missing-kind", "garden.yml")),
       (err) => {
-        expect(err.message).to.equal("Missing `kind` field in config at missing-kind/garden.yml")
+        expect(err.message).to.contain("Missing `kind` field in config at missing-kind/garden.yml")
       }
     )
   })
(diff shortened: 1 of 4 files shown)


core/test/unit/src/config/module-template.ts
```diff
@@ -499,10 +499,15 @@ describe("module templates", () => {
 
       await expectError(
         () => resolveTemplatedModule(garden, config, _templates),
-        (err) =>
-          expect(stripAnsi(err.message)).to.equal(
-            'ModuleTemplate test returned an invalid module (named module-${modules.foo.version}-test) for templated module test: Error validating module (modules.garden.yml): key .name with value "module-${modules.foo.version}-test" fails to match the required pattern: /^(?!garden)(?=.{1,63}$)[a-z][a-z0-9]*(-[a-z0-9]+)*$/. Note that if a template string is used in the name of a module in a template, then the template string must be fully resolvable at the time of module scanning. This means that e.g. references to other modules or runtime outputs cannot be used.'
+        (err) => {
+          const msg = stripAnsi(err.message)
```
Walther, line 503: Splitting this because we print the helpful `Available keys:` listings in all the messages - in this one it just happens to be in the middle of the two messages. In the other tests, that listing happens after the expected message, so replacing `equal` with `contain` is enough for the fix and avoids repetition

Orzelius (approved):
⚡

core/test/unit/src/config/base.ts
```diff
@@ -140,7 +140,7 @@ describe("loadConfigResources", () => {
     await expectError(
       async () => await loadConfigResources(projectPath, resolve(projectPath, "missing-type", "garden.yml")),
       (err) => {
-        expect(stripAnsi(err.message)).to.equal(
+        expect(stripAnsi(err.message)).to.contain(
```
vvagaytsev, line 143: Sorry for the late reply. We have a special helper function for fuzzy-match string comparison. So, this (and the similar places) could be rewritten in a shorter form like this:
```
await expectError(
      async () => await loadConfigResources(projectPath, resolve(projectPath, "missing-type", "garden.yml")),
      { contains: "Error validating module (missing-type/garden.yml): key .type is required" }
)
```
just to avoid the repetition of `stripAnsi` usage.

No worries, that's not critical at all :) We can align the code later.

PR #3533, opened 2023-01-17
Refix action tests

Orzelius:
issue: #3540

some test were broken due to modifications to gerden code.

does not fix all the tests, just the one's modified.

There are a couple tricky ones left (see issue) and a couple that are just yet to be written (marked as todo in code)

### DIFF
--- a/core/test/unit/src/router/run.ts
+++ b/core/test/unit/src/router/run.ts
@@ -94,7 +94,7 @@ describe("run actions", () => {
         () => actionRouter.run.getResult({ log, action, graph }),
         (err) =>
           expect(stripAnsi(err.message)).to.include(
-            "rror validating runtime action outputs from Run 'task-a': key .foo must be a string"
+            "Error validating runtime action outputs from Run 'task-a': key .foo must be a string"
           )
       )
     })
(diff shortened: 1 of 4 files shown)


core/test/unit/src/router/run.ts
```diff
@@ -94,7 +94,7 @@ describe("run actions", () => {
         () => actionRouter.run.getResult({ log, action, graph }),
         (err) =>
           expect(stripAnsi(err.message)).to.include(
-            "Error validating outputs from Run 'task-a': key .foo must be a string"
+            "rror validating runtime action outputs from Run 'task-a': key .foo must be a string"
```
vvagaytsev: Should "rror" be "Error" here?

vvagaytsev:
Looks great! 👍 I've left a minor comment, please check :)

core/test/unit/src/router/run.ts
Walther: Appears to be fixed already in a later commit of this PR 👍

Walther (approved):

vvagaytsev (approved):
LGTM! 👍 Thank you!

core/test/unit/src/router/run.ts
vvagaytsev: Ah, right! :)

Orzelius: yeah sorry messed up the commits a bit. I always try to avoid making later fixes like here to avoid confusion but this one slipped through

PR #3539, opened 2023-01-17
fix(k8s): fix Helm Tests, Runs and dev mode specs

thsig:

Fixed various schema / data shape errors that only surfaced when actually running `garden deploy`, `garden test` and `garden deploy --dev` in the `vote-helm` project.

After this fix, Tests and Runs are now working for `helm` modules.

Also fixed some other issues and cleaned things up along the way.

**IMPORTANT NOTE**

Runs and Tests generated from `helm` modules all have the `kubernetes-pod` type, and don't use the `podSpec` field. Therefore, they include a runtime dependency on their parent module's Deploy. This means that the `helm` Deploy is executed first, and the pod spec for the Test/Run pod runner is read from the cluster.

This behavior is different from 0.12, where the pod spec was read from the output of a dry-run deploy using the `helm` CLI (and did thus not require the deploy to take place first).

I did notice a few more problems with dev mode along the way, but those aren't `helm`-specific and should be addressed in a separate PR (mostly to do with log formatting, and then an issue with `actionSourcesChanged` events being handled even though the source belongs to a Build that's referenced by a dev-mode Deploy).


vvagaytsev (approved):
LGTM 🚀 Thank you! 👍

edvald:
It appears you have an import loop, causing the build to fail.

What you can do to track that down:

```
npm i detect-circular-deps -g
cd <repo root>/core
detect-circular-deps build/src/cli/cli.js
```

Commonly you can address these loops by changing imports to `import type` where possible, and moving exported zero/low-dependency functions to different or separate files.

core/src/plugins/kubernetes/helm/exec.ts
```diff
@@ -26,12 +26,9 @@ export const execInHelmDeploy: DeployActionHandler<"exec", HelmDeployAction> = a
   const defaultTarget = action.getSpec("defaultTarget")
 
   if (!defaultTarget) {
-    throw new ConfigurationError(
-      `${action.longDescription()} does not specify a defaultTarget. Please configure this in order to be able to use this command with.`,
-      {
-        name: action.name,
-      }
-    )
+    throw new ConfigurationError(`${action.longDescription()} does not specify a defaultTarget.`, {
```
edvald, line 29: I suspect this may be a very opaque error message in some scenarios. I think the original text ended with "... hot reloading" or something similar. Might be good to see what the relevant usages are here post refactor.

core/src/plugins/kubernetes/helm/handlers.ts
```diff
@@ -36,6 +37,10 @@ export const helmModuleHandlers: Partial<ModuleActionHandlers<HelmModule>> = {
 
     if (dummyBuild) {
       actions.push(dummyBuild)
+    } else {
+      // We make a dummy build without a `copyFrom` or any build dependencies. This is to ensure there's a build action
+      // for this module if it's used as a base by another Helm module.
+      actions.push(makeDummyBuild({ module, copyFrom: undefined, dependencies: undefined }))
```
edvald, line 43: This is not necessary. There is no notion of a base `helm` Deploy (the `base` feature is only in modules now).

edvald (changes requested):
Some small things I caught (beyond the import loop)

core/src/plugins/kubernetes/helm/handlers.ts
thsig: The `vote-helm` project fails to deploy if I remove this line (which triggers this exception: https://github.com/garden-io/garden/blob/51d027d2ff9db399e8a45106123ec17c26e5bdd2/core/src/graph/actions.ts#L422).

Was the idea instead to ask users to manually migrate their base chart-based modules, and to change the `vote-helm` project to no longer use the `base-chart` structure?

edvald: Then the issue is something more subtle, likely something very specifically to do with the base chart thing. Perhaps the dummy build needs to be injected specifically for that scenario (as opposed to always, which feels excessive to me).

The base chart mechanic should only be in `helm` _modules_ but should still be supported there. Users anyway need to manually migrate from modules to actions, and there is a straightforward alternative approach to "base charts" that doesn't need specific semantics at the action-type level. Namely, you just create multiple `helm` Deploy actions with the same chart path.

edvald (approved):

PR #3541, opened 2023-01-18
chore(server): remove debug logs for ws ping pongs

eysi09:

Previously we were logging websocket ping/pongs at the debug level. A user pointed out that this was adding a lot of noise to their logs. 

This PR removes the logs altogether since they'll add noise at any level. We do log when the connection times out which is based on the ping/pongs and should be enough. 

Fixes #

### DIFF
--- a/core/src/server/server.ts
+++ b/core/src/server/server.ts
@@ -363,11 +363,10 @@ export class GardenServer {
         return
       }
 
-      // Set up heartbeat to detect dead connections
-      let isAlive = true
-
       send("serverReady", { message: "Server ready" })
 
+      // Set up heartbeat to detect dead connections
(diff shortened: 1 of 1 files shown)


edvald (approved):
Looks good, but I'll note that I think we may be over-using the `chore` commit message category :) This is a user-facing change, subtle as it may be, so maybe `improvement` or even `fix` would apply imo.

eysi09:
> Looks good, but I'll note that I think we may be over-using the `chore` commit message category :) This is a user-facing change, subtle as it may be, so maybe `improvement` or even `fix` would apply imo.

That's a great point!

stefreak (approved):
Very good change!

PR #3546, opened 2023-01-18
ci(git): allow using `file` transfer protocol in CI

vvagaytsev:
See https://vielmetti.typepad.com/logbook/2022/10/git-security-fixes-lead-to-fatal-transport-file-not-allowed-error-in-ci-systems-cve-2022-39253.html

Fixes #


Orzelius (approved):
That's good thinking!

Will have to set it personally as well when running the tests, which might confuse future contributors if the vcs tests fail randomly

vvagaytsev:
@Orzelius good point! I added one more item to the TODO list in the description. We should also update the contributing guide. I'll create a separate issue out of the TODO comments above.

PR #3557, opened 2023-01-23
Dev mode fixes

thsig:

**fix(k8s): fixed dev mode for `container` services**

A few changes were needed to make dev mode work for `container` services in 0.13, mostly to use the same helper function to convert from the old-style dev mode sync specs to the new style.

**fix(core): correctly ignore watches in dev mode**

This fixes a regression in 0.13 where an `actionSourcesChanged` event would get fired when sources are changd for a Deploy that's deployed with dev mode enabled. This caused several problems, but is now fixed.

The problem lay in the way we were calculating the paths to skip/ignore, which are passed to the `Watcher` class.

**fix(k8s): fix sync destination formatting issue**

Before this fix, the path in the remote sync desination would get rendered as `[object Object]` - a simple change.

I haven't tested these changes on `kubernetes` modules yet, but I think dev mode should be working there now—and if not, the fixes needed should be quite straightforward at this point.

### DIFF
--- a/core/src/commands/serve.ts
+++ b/core/src/commands/serve.ts
@@ -92,7 +92,7 @@ export class ServeCommand extends Command<Args, Opts> {
     this.server!.setGarden(garden)
 
     const allModules = graph.getModules()
-    const allActions = graph.getActions()
+    const allDeployActions = graph.getActionsByKind("Deploy")
 
     await processActions({
       garden,
@@ -102,7 +102,7 @@ export class ServeCommand extends Command<Args, Opts> {
       watch: true,
       actions: [],
       initialTasks: [],
-      skipWatch: allActions,
+      skipWatch: allDeployActions,
       skipWatchModules: allModules,
       changeHandler: async () => [],
     })
--- a/core/src/garden.ts
+++ b/core/src/garden.ts
@@ -38,6 +38,7 @@ import {
   duplicatesByKey,
   uuidv4,
   getCloudDistributionName,
+  isTruthy,
 } from "./util/util"
 import { ConfigurationError, PluginError, RuntimeError } from "./exceptions"
 import { VcsHandler, ModuleVersion, getModuleVersionString, VcsInfo } from "./vcs/vcs"
@@ -132,6 +133,7 @@ import { actionConfigsToGraph, actionFromConfig, executeAction, resolveAction, r
 import { ActionTypeDefinition } from "./plugin/action-types"
 import { Task } from "./tasks/base"
 import { GraphResultFromTask, GraphResults } from "./graph/results"
+import { DeployAction } from "./actions/deploy"
 
 export interface ActionHandlerMap<T extends keyof ProviderHandlers> {
   [actionName: string]: ProviderHandlers[T]
@@ -430,7 +432,7 @@ export class Garden {
   }: {
     graph: ConfigGraph
     skipModules?: GardenModule[]
-    skipActions?: Action[]
+    skipActions?: DeployAction[]
     bufferInterval?: number
   }) {
     const actions = graph.getActions()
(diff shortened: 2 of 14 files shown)


core/src/garden.ts
```diff
@@ -439,18 +441,11 @@ export class Garden {
...
-    const skipDirectories = uniq([...skipModules.map((m) => m.path), ...skipActions.map((a) => a.basePath())])
-    const configPaths = new Set(
-      [...skipModules.map((m) => m.configPath), ...skipActions.map((a) => a.configPath())].filter(Boolean)
-    )
+    const skipPaths = [
+      ...skipModules.map((m) => m.configPath),
+      ...skipActions.map((a) => a.getBuildAction()?.basePath()),
+    ].filter(isTruthy)
 
-    const skipPaths = flatten(
```
thsig, line 447: It looks to me like this will always return an empty set, since it compares subdirectory abs paths with the config dir abs path (which will never match, since the subdirectory paths are strictly longer).

Was the idea to only ignore subdirectories of the action's directory, so that we're still watching the action config? We could do this using e.g. a prefix match on the directories in the (now removed) `skipDirectories` set.

core/src/garden.ts
```diff
@@ -439,18 +441,11 @@ export class Garden {
...
-      await Bluebird.map(skipDirectories, async (path: string) => {
-        return (await readdir(path))
-          .map((relPath) => resolve(path, relPath))
-          .filter((absPath) => configPaths.has(absPath))
```
thsig, line 451: It looks to me like this will always return an empty set, since it compares subdirectory abs paths with the config dir abs path (which will never match, since the subdirectory paths are strictly longer).

Was the idea to only ignore subdirectories of the action's directory, so that we're still watching the action config? We could do this using e.g. a prefix match on the directories in the (now removed) `skipDirectories` set.

core/src/commands/serve.ts
```diff
@@ -102,7 +102,7 @@ export class ServeCommand extends Command<Args, Opts> {
       watch: true,
       actions: [],
       initialTasks: [],
-      skipWatch: allActions,
+      skipWatch: allDeployActions,
```
edvald, line 105: What's the reason for this change?

core/src/plugins/container/config.ts
```diff
@@ -663,6 +662,7 @@ export interface ContainerCommonDeploySpec extends ContainerCommonRuntimeSpec {
 }
 
 export interface ContainerDeploySpec extends ContainerCommonDeploySpec {
+  devMode?: KubernetesDeployDevModeSpec
```
edvald, line 665: Hmm, this doesn't look right. We shouldn't be pulling the k8s-specific stuff into here.

core/src/garden.ts
edvald: You may well be right. I need to dig a bit into this to wrap my head around it :P

core/src/garden.ts
```diff
@@ -439,18 +441,11 @@ export class Garden {
...
     // For skipped modules/actions (e.g. those with services in dev mode), we skip watching all files and folders in the
     // module/action root except for the config path. This way, we can still react to changes in config files.
-    const skipDirectories = uniq([...skipModules.map((m) => m.path), ...skipActions.map((a) => a.basePath())])
-    const configPaths = new Set(
-      [...skipModules.map((m) => m.configPath), ...skipActions.map((a) => a.configPath())].filter(Boolean)
-    )
+    const skipPaths = [
+      ...skipModules.map((m) => m.configPath),
+      ...skipActions.map((a) => a.getBuildAction()?.basePath()),
+    ].filter(isTruthy)
```
edvald, line 447: FYI, you can also just do `.filter(Boolean)`

core/src/commands/serve.ts
thsig: On second thought, I think this change should be reverted—we don't want to trigger watch events here when sources change for Builds that aren't referenced by a Deploy.

core/src/garden.ts
thsig: Maybe I'm missing some typing, but if I use `.filter(Boolean)`, the type of `skipPaths` comes out as `(string | undefined)[]` (so I used the `isTruthy` type guard instead).

core/src/plugins/container/config.ts
thsig: Aha, I see. I thought it would be simplest to just use the same type for all the dev mode specs (`container`, `kubernetes` and `helm`).

IIRC, the only difference is the optional `containerName` field (which is rather k8s-specific). For comparison, the `volumes[].hostPath?` field, which is also on `ContainerDeploySpex` also feels somewhat k8s-specific.

Do we feel like this is OK as is if we rename `KubernetesDeployDevModeSpec` to `ContainerDeployDevModeSpec`?

Or should `ContainerDevModeSpec` omit the `containerName` field?

edvald: Well, that's not the only difference. There are also the `target` fields with a bunch of k8s-specific stuff, which I don't think makes sense to include. So I'd really like to keep `KubernetesDeployDevModeSpec` as a superset of `ContainerDeployDevModeSpec` and also in general avoid the cross-talk between the plugins.

core/src/garden.ts
edvald: Ah I see. Disregard then.

edvald:
Closing in favor of #3609

PR #3560, opened 2023-01-24
feat(kubernetes): add sync helper commands

eysi09:

This commit adds Kubernetes plugin commands for getting the Mutagen sync status, pausing syncs, and resuming syncs.

I also refactored the Mutagen code to use a class which makes it a little easier to manage some of the daemon process life cycle logic.

The class is currently a singleton but should be refactored after v0.13 into a "normal" class that we can attach to the `garden` instance.

Finally, the new MutagenDaemon class also logs more information about what Mutagen is doing under the hood. This is e.g. useful when Mutagen is syncing large change sets because without the added logs it looks like nothing is happening.

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/mutagen.ts
+++ b/core/src/plugins/kubernetes/mutagen.ts
@@ -6,7 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-const AsyncLock = require("async-lock")
+import AsyncLock from "async-lock"
 import chalk from "chalk"
 import { join } from "path"
 import { chmod, ensureSymlink, mkdirp, pathExists, remove, removeSync, writeFile } from "fs-extra"
@@ -14,7 +14,7 @@ import respawn from "respawn"
 import { LogEntry } from "../../logger/log-entry"
 import { PluginToolSpec } from "../../types/plugin/tools"
 import { PluginTool } from "../../util/ext-tools"
-import { makeTempDir, TempDirectory } from "../../util/fs"
+import { makeTempDir } from "../../util/fs"
 import { registerCleanupFunction, sleep } from "../../util/util"
 import { GardenBaseError } from "../../exceptions"
 import { prepareConnectionOpts } from "./kubectl"
@@ -24,6 +24,9 @@ import { devModeGuideLink } from "./dev-mode"
 import dedent from "dedent"
 import { PluginContext } from "../../plugin-context"
 import Bluebird from "bluebird"
+import { LATEST_MUTAGEN_DATA_DIR_NAME, MUTAGEN_DIR_NAME } from "../../constants"
+import EventEmitter from "events"
+import { ExecaReturnValue } from "execa"
 
 const maxRestarts = 10
 const monitorDelay = 2000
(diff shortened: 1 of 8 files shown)


core/src/plugins/kubernetes/mutagen.ts
```diff
@@ -48,6 +57,28 @@ export const mutagenModeMap = {
   "two-way-resolved": "two-way-resolved",
 }
 
+// This is basically copied from:
+// https://github.com/mutagen-io/mutagen/blob/19e087599f187d85416d453cd50e2a9df1602132/pkg/synchronization/state.go
+// with an updated description to match Garden's context.
+const mutagenStatusDescriptions = {
+  "disconnected": "Sync disconnected",
+  "halted-on-root-emptied": "Sync halted due to one-sided root emptying",
```
edvald, line 65: I wonder if we can use more, ehm, human language here? Would need to think on how exactly though, tbh :P

core/src/plugins/kubernetes/mutagen.ts
```diff
@@ -48,6 +57,28 @@ export const mutagenModeMap = {
...
+  "halted-on-root-deletion": "Sync halted due to root deletion",
+  "halted-on-root-type-change": "Sync halted due to root type change",
+  "connecting-alpha": "Sync connected to alpha",
```
edvald, line 68: Maybe we should swap out (or maybe explain) alpha and beta in more relatable terms for the user.

core/src/plugins/kubernetes/mutagen.ts
```diff
@@ -68,222 +99,569 @@ interface ActiveSync {
...
+      onFailedAttempt: async (err) => {
+        const unableToFlush = err.message.match(/unable to flush session/)
         if (unableToFlush) {
-          log.warn({
+          this.log.warn({
             symbol: "empty",
             section: mutagenLogSection,
-            msg: chalk.gray(`Could not flush mutagen session, retrying (attempt ${loops}/${maxRetries})...`),
+            msg: chalk.gray(
+              `Could not flush mutagen session, retrying (attempt ${err.attemptNumber}/${err.retriesLeft})...`
```
edvald, line 455: Once more, wondering if we can phrase this more clearly. Not sure if "flushing a mutagen session" has any bearing for uninitiated users .)

core/src/plugins/kubernetes/mutagen.ts
```diff
@@ -68,222 +99,569 @@ interface ActiveSync {
...
+            statusMsg = `Sync paused`
+          } else if (session.status && session.status !== activeSync.lastStatus) {
+            statusMsg = mutagenStatusDescriptions[session.status]
+          }
+
+          if (statusMsg) {
+            this.syncStatusLines[sessionName].setState({
+              symbol: "info",
+              section,
+              msg: chalk.gray(statusMsg),
```
edvald, line 635: Should we perhaps add a `[sync] ` prefix here?

core/src/plugins/kubernetes/mutagen.ts
eysi09: Suggestion: "Sync halted because either the source or target directory was emptied"

eysi09: Suggestion for root type change: "Sync halted because either the source or target changed type"

eysi09: Suggestion for root deletion: "Sync halted because either the source or target was deleted"

core/src/plugins/kubernetes/mutagen.ts
eysi09: Suggestion: "Could not flush synchronisation changes..."

edvald (approved):

PR #3561, opened 2023-01-24
feat(commands): allow disabling watch in dev mode

eysi09:

This is an experimental behaviour that allows users to disable watching while in dev mode.

Note that this refers to Garden's watching, Mutagen syncs still work.

This behaviour will change in a future release where we'll make the whole watch mechanism smarter and more consistent.

Fixes #

### DIFF
--- a/core/src/commands/dev.ts
+++ b/core/src/commands/dev.ts
@@ -24,6 +24,7 @@ import {
   getHotReloadServiceNames,
   getMatchingServiceNames,
   validateHotReloadServiceNames,
+  makeSkipWatchErrorMsg,
 } from "./helpers"
 import { startServer } from "../server/server"
 import { BuildTask } from "../tasks/build"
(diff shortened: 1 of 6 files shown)


core/src/commands/dev.ts
```diff
@@ -153,6 +167,7 @@ export class DevCommand extends Command<DevCommandArgs, DevCommandOpts> {
     const modules = graph.getModules()
 
     const skipTests = opts["skip-tests"]
+    const watch = opts.watch
```
vvagaytsev, line 170: Should we check the `devModeServiceNames.length` condition here too? It is still possible to run `garden dev` command for a project that doesn't have any `devMode` configs.

eysi09: Hmm I suppose. 

This current behaviour is a bit confusing. Dev mode services are just all services that don't have the hot reload stanza. Irrespective of whether or not it has `devMode` config. 

So a service with no sync config is still a dev mode service. 

But I suppose you could have a project where all services are hot reload services.

I'd argue that this behaviour is basically broken but this will be fixed with 0.13 any way so I don't want to tamper with it here. 

But to answer your question, yes we can check for the length.

eysi09:
Converting to draft until I wrap this up.

edvald (approved):

PR #3566, opened 2023-01-24
test: attempt to fix test configs

vvagaytsev:

Fixes #

### DIFF
--- a/core/test/unit/src/plugins/exec/exec.ts
+++ b/core/test/unit/src/plugins/exec/exec.ts
@@ -454,7 +454,10 @@ describe("exec plugin", () => {
           silent: false,
           action,
         })
-        expect(res.outputs).to.eql(join(garden.projectRoot, "module-local"))
+        expect(res.state).to.eql("ready")
+        expect(res.detail?.log).to.eql("pwd")
+        // TODO-G2: it looks like there is no path available in the Test result. Does this test case still make sense?
+        // expect(res.outputs).to.eql(join(garden.projectRoot, "module-local"))
       })
 
       it("should receive module version as an env var", async () => {
@@ -488,7 +491,9 @@ describe("exec plugin", () => {
           graph,
           silent: false,
         })
-        expect(res.outputs).to.equal(rawAction.versionString())
+        expect(res.state).to.eql("ready")
+        // TODO-G2: the result contains unresolved env var and looks like `echo $GARDEN_MODULE_VERSION`
+        expect(res.detail?.log).to.eql(`echo ${rawAction.versionString()}`)
       })
     })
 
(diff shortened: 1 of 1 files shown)


core/test/unit/src/plugins/exec/exec.ts
```diff
@@ -435,7 +435,7 @@ describe("exec plugin", () => {
             command: ["pwd"],
           },
           internal: {
-            basePath: "TODO-G2",
+            basePath: "module-local",
```
vvagaytsev, line 438: I'm not sure if this is a valid config.

core/test/unit/src/plugins/exec/exec.ts
```diff
@@ -448,7 +448,10 @@ describe("exec plugin", () => {
         silent: false,
         action,
       })
-      expect(res.outputs).to.eql(join(garden.projectRoot, "module-local"))
+      expect(res.state).to.eql("ready")
+      expect(res.detail?.log).to.eql("pwd")
+      // TODO-G2: it looks like there is no path available in the Test result. Does this test case still make sense?
```
vvagaytsev, line 453: I added an assertion to check the logs. The `TestResult` has no paths. It might make sense to delete this test case. It's called `should run the test command in the module dir if local true` and I'm not sure if there is a way to verify that.

The log output verification is implemented in the test case below.

core/test/unit/src/plugins/exec/exec.ts
```diff
@@ -482,7 +485,9 @@ describe("exec plugin", () => {
         graph,
         silent: false,
       })
-      expect(res.outputs).to.equal(rawAction.versionString())
+      expect(res.state).to.eql("ready")
+      // TODO-G2: the result contains unresolved env var and looks like `echo $GARDEN_MODULE_VERSION`
+      expect(res.detail?.log).to.eql(`echo ${rawAction.versionString()}`)
```
vvagaytsev, line 490: The `res.detail?.log` contains `echo $GARDEN_MODULE_VERSION` instead of the resolved version value. It looks like a bug in the code.

core/test/unit/src/plugins/exec/exec.ts
```diff
@@ -435,7 +435,7 @@ describe("exec plugin", () => {
             command: ["pwd"],
           },
           internal: {
-            basePath: "TODO-G2",
+            basePath: `${testProjectRoot}/module-local`,
```
vvagaytsev, line 438: I'm not sure if this is a valid config.

core/test/unit/src/plugins/exec/exec.ts
```diff
@@ -469,7 +472,7 @@ describe("exec plugin", () => {
             command: ["echo", "$GARDEN_MODULE_VERSION"],
           },
           internal: {
-            basePath: "TODO-G2",
+            basePath: `${testProjectRoot}/module-local`,
```
vvagaytsev, line 475: I'm not sure if this is a valid config.

vvagaytsev:
Superseded by #3689

PR #3568, opened 2023-01-24
fix(kubernetes): make preStop command work with busybox

stefreak:

The pgrep -c option does not exist in busybox. Make the preStop command more portable.

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/container/build/common.ts
+++ b/core/src/plugins/kubernetes/container/build/common.ts
@@ -501,7 +501,7 @@ export function getUtilContainer(authSecretName: string, provider: KubernetesPro
           command: [
             "/bin/sh",
             "-c",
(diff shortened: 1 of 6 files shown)


Walther (approved):

core/src/plugins/kubernetes/container/build/common.ts
```diff
@@ -501,7 +501,7 @@ export function getUtilContainer(authSecretName: string, provider: KubernetesPro
           command: [
             "/bin/sh",
             "-c",
-            "until test $(pgrep -fc '^[^ ]+rsync') = 1; do echo waiting for rsync to finish...; sleep 1; done",
+            "until pgrep -f '^[^ ]+rsync' > /dev/null; do echo waiting for rsync to finish...; sleep 1; done",
```
stefreak, line 504: That doesn't really work, because there's always at least one rsync process. Need to dig into that more

Walther (approved):
nice, thanks! 🚀

PR #3580, opened 2023-01-25
0.13 minor command fixes

vvagaytsev:

Fixes #

### DIFF
--- a/docs/reference/commands.md
+++ b/docs/reference/commands.md
@@ -338,7 +338,9 @@ This can be useful if you find the namespace to be in an inconsistent state, or 
 
 | Argument | Alias | Type | Description |
 | -------- | ----- | ---- | ----------- |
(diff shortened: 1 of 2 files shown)


docs/reference/commands.md
```diff
@@ -372,89 +374,98 @@ providerStatuses:
...
 
-    # Additional detail, specific to the provider.
     detail:
+      # When the service was first deployed by the provider.
+      createdAt:
 
-    # Whether the service was deployed with dev mode enabled.
-    devMode:
+      # Additional detail, specific to the provider.
+      detail:
```
Walther, line 390: just to double check - we have a `detail.detail` structure here?

vvagaytsev: I think so. The top-level structure is [`ActionState`](https://github.com/garden-io/garden/blob/0.13/core/src/actions/types.ts#L110-L118), where the `detail` is an arbitrary object that can be provider specific. I've seen some nested `detail.detail` structures in the tests.

Walther (approved):

PR #3581, opened 2023-01-25
0.13 `DevCommand` fixes

vvagaytsev:
It looks like there is a bug in the [`getDevCommandWatchTasks`](https://github.com/garden-io/garden/blob/0.13/core/src/commands/dev.ts#L282-L283). Both `deploysWatched` and `devModeDeployNames` receive the same value in the [`getActionWatchTasks` function call](https://github.com/garden-io/garden/blob/0.13/core/src/commands/dev.ts#L277). But that function has the condition [`deploysWatched.includes(a.name) && !devModeDeployNames.includes(a.name)`](https://github.com/garden-io/garden/blob/0.13/core/src/tasks/helpers.ts#L105) that will never hold if `deploysWatched` and `devModeDeployNames` get the same value.

Fixes #

### DIFF
--- a/core/test/unit/src/commands/dev.ts
+++ b/core/test/unit/src/commands/dev.ts
@@ -72,7 +72,6 @@ describe("DevCommand", () => {
     const opts = withDefaultGlobalOpts({
       "force-build": false,
       "force": false,
-
       "local-mode": undefined,
       "skip-tests": false,
       "test-names": undefined,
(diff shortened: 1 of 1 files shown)


edvald:
I'll take this one @vvagaytsev! Gonna fold together some stuff I had looked into, plus @thsig's branch.

vvagaytsev:
Thanks @edvald!

PR #3584, opened 2023-01-25
fix(exec): more informative error msg on timeout

stefreak:
- Made the exec module error message on timeout more informative.
- Updated the test

Still working on the kubernetes module, but decided to put this as a separate PR

Fixes #


10ko (approved):
Approving but I'd leave a comment. Would it make sense to add these error messages to a `util` file of some sort in order to reuse it in other plugins?
Or are we ok with duplicating the messages across them in order to maintain isolation?

stefreak:
> Approving but I'd leave a comment. Would it make sense to add these error messages to a `util` file of some sort in order to reuse it in other plugins? Or are we ok with duplicating the messages across them in order to maintain isolation?

Good question, right now this error message exists in only one place (the kubernetes one will be different) – maybe we can extract it to a helper function once we have multiple plugins that throw this error (at least searching for "Timed out" in the code base only revealed exec and kubernetes plugins, and docker swarm but I think that's probably not a supported / tested one?)

10ko:
Ok, gotcha, I was under the impression the error messages would be the same. This might be a conversation worth having at some point though, it would be nice to find a way to render similar error messages across different plugins. I would possibly require a bit of change in conventions/architecture and we can discuss it on a separate issue.

PR #3599, opened 2023-01-26
test: 0.13 refactor test project config creation

vvagaytsev:
This PR makes tests more dryish by introducing more templating for the test project configs.

Fixes #

* Tests in `./core/test/unit/src/garden.ts` haven't been modified yet. There we need some overrides instead of for some config entries. A necessary helper function can be added in a separate PR. 
* I executed all affected tests manually on my local machine and did not find any regression against the `0.13` branch


edvald:
Maybe one suggestion, at least going forward: Use the `test` prefix instead of `refactor(test)`, to avoid having to later filter this out from release notes :)

vvagaytsev:
Thanks, @edvald! I fixed the prefixes in the commit messages.

PR #3601, opened 2023-01-26
0.13 cherrypick logger changes

stefreak:
- fix(framework): increase consistency in verbose plugin logs
- fix(exec): more informative error msg on timeout (#3584)
- fix(events): update log event timestamp type be string or number
- improvement(server): ensure serverReady event has correct type
- fix(helm): stream logs to CLI and Garden Cloud (#3582)

Fixes #


core/src/router/build.ts
```diff
@@ -57,7 +57,11 @@ export const buildRouter = (baseParams: BaseRouterParams) =>
       const moduleVersion = action.moduleVersion().versionString
       const moduleName = action.moduleName()
 
-      params.events.on("log", ({ timestamp, data }) => {
+      params.events.on("log", ({ timestamp, data, origin, log }) => {
+        // stream logs to CLI
+        log.setState(renderOutputStream(data.toString(), origin))
```
stefreak, line 62: For some reason, in 0.13, the "log" handlers are not being called. I tried to understand why, but couldn't find the culprit. My guess is that the `events` instance in the plugin is a different instance than here.

This means that once this PR is merged, some logs won't appear in the CLI anymore until this "unrelated" problem has been fixed.

edvald: It is as you suspect, a different instance (`params.events` vs. `ctx.events`). We may want to drop one of them actually to avoid that confusion going forward. I can take that on.

edvald: I've added a commit that ensures the same event broker is passed along through the plugin router, which should fix this issue.

eysi09 (approved):

core/src/router/build.ts
stefreak: Thank you!

PR #3604, opened 2023-01-26
refactor: remove old `garden dev` command implementation

edvald:
For now, the `dev` command implementation is basically that of the `serve` command, with an added "under construction" message to steer users to the `deploy --dev` command.

Later we will add interactivity on top, hopefully before the 0.13 GA.

BREAKING CHANGE:

The behavior of the `garden dev` command is being fully replaced. It will no longer automatically perform all tests and deployments, but instead connect to the new dashboard, and later feature a command input in the terminal.

### DIFF
--- a/core/src/commands/dev.ts
+++ b/core/src/commands/dev.ts
@@ -6,298 +6,72 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import deline = require("deline")
-import dedent = require("dedent")
+import { CommandResult, CommandParams } from "./base"
+import { dedent } from "../util/string"
+import { ServeCommand, ServeCommandArgs, ServeCommandOpts } from "./serve"
 import chalk from "chalk"
-import { readFile } from "fs-extra"
-import moment = require("moment")
-import { join } from "path"
-
-import { getActionWatchTasks } from "../tasks/helpers"
-import { Command, CommandResult, CommandParams, handleProcessResults, PrepareParams } from "./base"
-import { STATIC_DIR } from "../constants"
-import { processActions } from "../process"
-import { TestTask } from "../tasks/test"
-import { ConfigGraph } from "../graph/config-graph"
-import { getMatchingServiceNames } from "./helpers"
-import { startServer } from "../server/server"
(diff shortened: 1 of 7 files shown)


core/src/commands/dev.ts
```diff
@@ -6,298 +6,70 @@
...
-
-export type DevCommandArgs = typeof devArgs
-export type DevCommandOpts = typeof devOpts
+// const ansiBannerPath = join(STATIC_DIR, "garden-banner-2.txt")
 
 // TODO: allow limiting to certain modules and/or services
-export class DevCommand extends Command<DevCommandArgs, DevCommandOpts> {
+export class DevCommand extends ServeCommand {
   name = "dev"
   help = "Starts the garden development console."
```
eysi09, line 22: Should we also update the help text?

core/src/commands/dev.ts
```diff
@@ -6,298 +6,70 @@
...
-    const deploys = graph.getDeploys({ names: args.deploys })
+  //   this.server = await startServer({ log: footerLog })
+  // }
 
-    const localModeDeployNames = getMatchingServiceNames(opts["local-mode"], graph)
+  async action(params: CommandParams<ServeCommandArgs, ServeCommandOpts>): Promise<CommandResult> {
+    params.log.warn({
+      msg:
+        chalk.bold(dedent`
+          🚧  This command is ${chalk.yellow.bold("under construction")}  🚧
```
eysi09, line 53: Should we explain this a little better? 

A user that is already using the command might be a bit surprised by to see this all of a sudden. 

Maybe add that they're seeing this because they've upgraded to 0.13 and ensure it's clear that they shouldn't downgrade.

eysi09 (commented):
Couple of comments on the text.

core/src/commands/dev.ts
edvald: That's fair. I was imagining this all to be very temporary, even cleared up before 0.13 GA, but I can add some more commentary.

core/src/commands/dev.ts
edvald: Ah yes, missed that. I wonder what makes sense though, any thoughts?

PR #3609, opened 2023-01-27
Dev mode fixes take 2

### LINKED ISSUE #3525: The `--dev` flag for the `garden deploy` command shouldn't automatically imply watch mode for other modules
The `--dev` flag for the `garden deploy` command shouldn't automatically imply watch mode for other modules. Instead, explicitly opt-in to watching non-dev mode modules via the `-w`/`--watch` flag.

No need to change the `garden dev` command.

edvald:
This supersedes #3557, with some commits cherry-picked, some modified and one new commit added that solves a major headache:

[refactor(deploy): remove --skip-watch and disallow --watch with dev/local mode](https://github.com/garden-io/garden/commit/3dfd1b1848b98e58da6839eba35034314ca3f4c1) 

BREAKING CHANGE:

The experimental `--skip-watch` flag has been removed from the `deploy` command.

`garden deploy` no longer implicitly sets `--watch=true` along with dev or
local modes. We also disallow using `--watch` at the same time as `--dev` or
`--local` mode flags.

This avoids a class of issues where automatic actions triggered by Garden's
watcher would conflict with sync or local process activities. Additionally
avoids several scenarios where the file watcher would consume too many
resources, or unwanted actions would be triggered.

Closes #3525


core/src/commands/deploy.ts
```diff
@@ -159,6 +146,13 @@ export class DeployCommand extends Command<Args, Opts> {
   }: CommandParams<Args, Opts>): Promise<CommandResult<ProcessCommandResult>> {
     this.garden = garden
 
+    if (opts.watch && (opts["dev-mode"] || opts["local-mode"])) {
+      throw new ParameterError(
+        `The -w/--watch flag cannot currently be used at the same time as the --local or --dev mode flags. Please refer to documentation for more details.`,
```
eysi09, line 151: Just a nitpick but I'd either skip the "Please refer to..." part of add an actual link.

eysi09 (commented):
Just one small comment, otherwise looks good.

eysi09 (approved):

PR #3610, opened 2023-01-27
fix(install): remove unnecessary rsync dependency in homebrew formula

edvald:
Resolves https://github.com/garden-io/homebrew-garden/issues/8

### DIFF
--- a/support/homebrew-formula.rb
+++ b/support/homebrew-formula.rb
@@ -5,8 +5,6 @@ class GardenCli < Formula
   version "{{version}}"
   sha256 "{{sha256}}"
 
-  depends_on "rsync"
-
   def install
     libexec.install "garden", "fsevents.node", "static", "better_sqlite3.node"
     bin.install_symlink libexec/"garden"
--- a/core/src/build-staging/rsync.ts
+++ b/core/src/build-staging/rsync.ts
@@ -16,23 +16,29 @@ import { validateInstall } from "../util/validateInstall"
 export const minRsyncVersion = "3.1.0"
 const versionRegex = /rsync\s+version\s+v?(\d+.\d+.\d+)/
 
-/**
- * throws if no rsync is installed or version is too old
- */
-export async function validateRsyncInstall() {
-  await validateInstall({
-    minVersion: minRsyncVersion,
-    name: "rsync",
-    versionCommand: { cmd: "rsync", args: ["--version"] },
-    versionRegex,
-  })
-}
+export class BuildStagingRsync extends BuildStaging {
+  private rsyncPath = "rsync"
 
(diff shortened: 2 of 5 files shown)


support/homebrew-formula.rb
```diff
@@ -5,8 +5,6 @@ class GardenCli < Formula
   version "{{version}}"
   sha256 "{{sha256}}"
 
-  depends_on "rsync"
```
stefreak, line 8: @edvald This PR is to `main`; Has the dependency on rsync already been removed in main?

edvald: It has on macOS, yes

edvald: I came across an additional bit of validation that I was unaware of, while working on #3617. I've plucked the relevant commit over to address that.

stefreak: Hi @edvald sorry if I missed something obvious, I just don't understand why it has been required and why it is optional now – what changed? Our docs still list rsync as a requirement: https://docs.garden.io/getting-started/1-installation#requirements

stefreak: I tested garden build and garden deploy with edge on MacOS after brew uninstall rsync, and it works. But I also don't know why the dependency has been added in the first place, as half a year ago I also successfully used Garden without the additional rsync dep.

I'd say at least let's also update the docs so the actual dependencies are documented properly.

edvald: The history is that we _used_ to use rsync for various things that we've since gotten away from. Rsync is unfortunately still needed on Windows (and I _really_ want to get rid of it there as well). 

I'm actually considering just moving this to 0.13 since this is ballooning a bit in scope (docs, validation etc etc.), and there isn't much harm in leaving this in 0.12, so your hunch was likely right there :)

edvald:
Closing, will follow up with PR for 0.13

support/homebrew-formula.rb
stefreak: @edvald  When I tested a build from your branch, garden still used rsync to sync copy the build context for in-cluster building – so it still needs rsync, but just not a super new version of it? 

I basically just asked because I didn't understand if we don't need any rsync anymore at all, or if only the strict "rsync greater than `v3.1.0`" requirement has been lifted. If only updating the docs was left, it might have been really easy as we only mention the dependency in the install guide

edvald: One thing I missed was that we still had the old rsync build syncing enabled by default. I somehow thought we'd flipped the switch to use Mutagen for that. All the more reason to punt to 0.13, and completely remove the old rsync-to-cluster mechanism. I'll add that to a follow-on PR.

PR #3616, opened 2023-01-28
chore: update yarn to berry version

edvald:
Switching to npm turned out to be a whole hassle, and kinda needs to coincide with a broad update of various packages. This is quite a bit simpler to sort out, and hopefully improves stability in CI processes.


edvald:
After some more discussion and probing, I'm closing this. If it had worked smoothly and swiftly, I'd be fine with it, but I'm not really a fan of yarn anymore. We'll just switch to npm when we feel the time is right.

PR #3618, opened 2023-01-28
fix(events): use ISO format date strings in event payload

eysi09:

Previously the event payload type for dates was a JS Date instance. When these events are emitted to other consumers such as Cloud the dates are marshalled to ISO strings.

This commit updates the payload types to strings so that the types reflect what the consumers actually receive and so that the types can be re-used across consumers.

Fixes #

### DIFF
--- a/core/src/actions.ts
+++ b/core/src/actions.ts
@@ -376,7 +376,7 @@ export class ActionRouter implements TypeGuard {
     const actionUid = uuidv4()
     params.events = params.events || new PluginEventBroker()
     let result: BuildResult
-    const startedAt = new Date()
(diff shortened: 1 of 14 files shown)


stefreak (commented):
@edvald @eysi09 should we also change the `log` timestamp to `string`, for consistency?

It's currently

```
timestamp: number | string
```

eysi09:
> @edvald @eysi09 should we also change the `log` timestamp to `string`, for consistency?
> 
> It's currently
> 
> ```
> timestamp: number | string
> ```

Yeah we should. I'll update the PR.

edvald (approved):

PR #3619, opened 2023-01-29
refactor(core): get rid of sqlite and just use file-backed config stores 

edvald:
This substantially cuts down on startup time and general build complexity.

Along the way, we make sure users can have a cloud auth token for each
Garden Cloud domain, which will if nothing else simplify development.

For good measure, this introduces the first bits of zod schemas into our
codebase, which is long overdue.

Best to look at the individual commits to dig into the refactors.


core/src/process.ts
```diff
@@ -288,3 +289,53 @@ async function validateConfigChange(
...
+  // TODO: avoid multiple writes here
+  await Bluebird.map(processes, async (p) => {
+    if (!isRunning(p.pid)) {
+      await globalConfigStore.delete("activeProcesses", String(p.pid))
+    }
+  })
+
+  const pid = process.pid
+
+  const record: GardenProcess = {
```
thsig, line 313: Maybe we should use `processInfo` instead of `processRecord`/`record`? Similar to the naming of the `CommandInfo` interface (and associated local variables).

core/src/config-store/base.ts
```diff
@@ -0,0 +1,175 @@
...
+      const data = await readFile(configPath)
+      parsed = JSON.parse(data.toString())
+    } catch {
+      parsed = await this.initConfig(true)
+    }
+    return this.validate(parsed, "reading")
+  }
+
+  private async writeConfig(config: I<T>) {
+    await writeFile(this.getConfigPath(), JSON.stringify(config))
```
mkhq, line 173: I think this could result in a partial/corrupted config file if there is some error while writing. `writeFile` potentially calls `write` multiple times with the buffer and is not guaranteed to be atomic. One way to work around this and make it more safe would be to first write the data in a temp file and then use `mv` which is atomic on posix filesystems.

core/src/process.ts
edvald: I mean, it's still a record of a process. Don't think it matters too much what this is called in internal interfaces :P

core/src/config-store/base.ts
edvald: I don't believe that's true, given that there is a file lock placed around every execution of this. The lock is intentionally not placed inside this particular method because the lock needs to encapsulate the read and the write at the same time.

edvald: Ah I may have jumped the gun there, I suppose that is indeed possible (if highly unlikely). I don't mind addressing that, to err on the side of caution. My hunch is that there's a library for that .)

edvald: Did that, and added a stricter file mode on the global config to boot, since it will contain auth tokens.

edvald: For future reference: https://github.com/npm/write-file-atomic

core/src/cloud/api.ts
```diff
@@ -294,29 +285,9 @@ export class CloudApi {
...
-      } catch (error) {
-        log.error(`An error occurred while clearing up duplicate client auth tokens:\n${error.message}`)
-      }
-    }
-    log.silly(`Retrieved client auth token from local config db`)
-
-    return token
+  static async getStoredAuthToken(log: LogEntry, globalConfigStore: GlobalConfigStore, domain: string) {
+    log.silly(`Retrieving client auth token config store`)
+    return globalConfigStore.get("clientAuthTokens", domain)
```
eysi09, line 290: Nice :)

core/test/unit/src/process.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright (C) 2018-2022 Garden Technologies, Inc. <info@garden.io>
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+describe("registerProcess", () => {
```
eysi09, line 9: Remove or update?

eysi09 (changes requested):
Looks good but noticed a couple of unfinished tests in there.

core/test/unit/src/process.ts
edvald: Good catch, added the implementations.

eysi09 (approved):

vvagaytsev:
It looks like this affected the [`HideWarninigCommand`](https://github.com/garden-io/garden/blob/main/core/src/commands/util/hide-warning.ts).

So, we should also store warnings in the new FS-based store now. Is that correct, @edvald?

edvald:
We actually do, yes. Let me know if you noticed something wrong about the implementation, perfectly possible I missed something :)

PR #3621, opened 2023-01-30
0.13: implement base router validateActionOutputs tests

Orzelius:


core/test/unit/src/router/base.ts
```diff
@@ -431,20 +438,57 @@ describe("BaseActionRouter", () => {
...
-      throw "TODO"
+      const { router } = await createTestRouter(testPlugins)
+
+      await expectError(async () => await router.validateActionOutputs(resolvedBuildAction, "runtime", { foo: 123 }), {
+        contains: "Error validating runtime action outputs from Build 'module-a': key .foo must be a string.",
+      })
     })
 
     it("throws if no schema is set and a key is set", async () => {
       throw "TODO"
```
Orzelius, line 471: Currently [validateActionOutputs](https://github.com/garden-io/garden/blob/0.13/core/src/router/base.ts#L317) doesn't throw if the schema is not defined and an output is passed. Which is the intended behavior?

Walther: No idea - need to ask someone with more context knowledge 😅

edvald: I believe that is an actual issue, but it's not really a major concern, you can just skip the test.

edvald:
I wonder what's going on with the `test-framework` step here... Those tests are certainly not _actually_ passing.

Issue #3622, opened 2023-01-30
0.13: plugins.ts


Orzelius:

edvald:
Took on the remainder of this @Orzelius

PR #3627, opened 2023-01-31
0.13: implement plugins.ts tests and do a bit of refactoring

Orzelius:
closes: #3622


core/test/unit/src/plugins.ts
```diff
@@ -6,38 +6,160 @@
...
+
+    const dependant = createGardenPlugin({ name: "dependant", base: "base" })
+
+    const result = resolvePlugins(log, { base, dependant }, [{ name: "test" }])
+    const inheritedActionType = result.find((plugin) => plugin.name === "dependant")?.createActionTypes.Build[0]
+    expect(inheritedActionType).to.exist
+    expect(inheritedActionType?.name).to.eql("base")
   })
 
-  context("base is not configured", () => {
```
Orzelius, line 30: The diff is confusing but we deleted [this context](https://github.com/garden-io/garden/blob/0.13/core/test/unit/src/plugins.ts#L30) as it seemed to be the exact same cases as describe above. Please check @edvald

edvald: Actually the idea is to make tests for both scenarios, even though the cases are indeed largely the same. It's to validate that nothing breaks when there are bases on plugins. Might also be missing some more specific cases in that scenario.

edvald (changes requested):
Looks great, just had one comment there (in response to your question). Also noting that the lint step fails—let's try and keep that in order now that we can .)

core/test/unit/src/plugins.ts
Orzelius: I don't really understand that context block in general. It says `base is not configured` but the first test case is `pulls created action type from base`. How can it pull an action type from the base if base is not configured? Also, what would `pull` mean in this context? I'd be happy to implement the tests I just don't quite get what they're supposed to test. cc @vvagaytsev @Walther

vvagaytsev: I didn't get the general context too. We can restore that context, merge the PR, and address the unclear context later.

vvagaytsev (approved):

core/test/unit/src/plugins.ts
Orzelius: Currently I need @edvald's green light to merge as he requested changes

vvagaytsev: Yeah, I see. But if the requested changes were only about this context, then it should be fine to dismiss the review and merge the remained parts of the PR which have been approved (I assume that because there are no other comments with any changes requested).

If we wanna do that, let's just restore the problematic context in a separate commit, and then merge the PR.

Orzelius: Restored the test context here [f74d5c3](https://github.com/garden-io/garden/pull/3627/commits/f74d5c314ee581eaacefa3fdeda98802dd18b138)

edvald: Ok, so, you may sometimes just need to go with your good sense in my absence 😬 All I said in my comment was that we needed tests for both scenarios, not that all the test cases were definitely correct or needed to be unchanged. Seems there was a misplaced test case (or more, not sure) but the intent is just that both scenarios are fairly well tested. If that's the case, all set ✨

PR #3632, opened 2023-01-31
fix(terraform): update condition that triggers `terraform init`

### LINKED ISSUE #3325: terraform garden module using terraform 1.2.9 plugin fails with validation error.
### Current Behavior
When using a garden terraform module type and the terraform uses providers that it needs to fetch from the registry garden will fail with error:

```
Failed getting status for service 'firebase' (from module 'firebase'). Here is the output:
Failed validating Terraform configuration:

Error: missing or corrupted provider plugins:
- registry.terraform.io/hashicorp/kubernetes: there is no package for registry.terraform.io/hashicorp/kubernetes 2.14.0 cached in .terraform/providers
- registry.terraform.io/hashicorp/google: there is no package for registry.terraform.io/hashicorp/google 4.41.0 cached in .terraform/providers
```

### Expected behavior
garden runs terraform init and apply to the module 

### Reproducible example
garden.yml:

```
kind: Module
type: terraform
name: terraform module
allowDestroy: true
autoApply: true
```

main.tf:

```
terraform {
  required_version = ">= 0.13"

  required_providers {
    google = {
      source  = "hashicorp/google"
    }

  }
}
```

### Workaround
Use terraform 1.0.5

### Additional context
between terraform 1.0.5 and terraform 1.2.9 changed the output of `terraform validate`:
terraform 1.0.5:

```
$HOME/.garden/tools/terraform-1-0-5/1d742d164493f796/terraform validate -json
{
  "format_version": "0.1",
  "valid": false,
  "error_count": 1,
  "warning_count": 0,
  "diagnostics": [
    {
      "severity": "error",
      "summary": "Could not load plugin",
      "detail": "\nPlugin reinitialization required. Please run \"terraform init\".\n\nPlugins are external binaries that Terraform uses to access and manipulate\nresources. The configuration provided requires plugins which can't be located,\ndon't satisfy the version constraints, or are otherwise incompatible.\n\nTerraform automatically discovers provider requirements from your\nconfiguration, including providers used in child modules. To see the\nrequirements and constraints, run \"terraform providers\".\n\n2 problems:\n\n- failed to instantiate provider \"registry.terraform.io/hashicorp/google\" to obtain schema: there is no package for registry.terraform.io/hashicorp/google 4.41.0 cached in .terraform/providers\n- failed to instantiate provider \"registry.terraform.io/hashicorp/kubernetes\" to obtain schema: there is no package for registry.terraform.io/hashicorp/kubernetes 2.14.0 cached in .terraform/providers\n"
    }
  ]
}
```

terraform 1.2.9:

```
$HOME/.garden/tools/terraform-1-2-9/bcf17392a2ac5e7c/terraform validate -json
{
  "format_version": "1.0",
  "valid": false,
  "error_count": 1,
  "warning_count": 0,
  "diagnostics": [
    {
      "severity": "error",
      "summary": "missing or corrupted provider plugins:\n  - registry.terraform.io/hashicorp/google: there is no package for registry.terraform.io/hashicorp/google 4.41.0 cached in .terraform/providers\n  - registry.terraform.io/hashicorp/kubernetes: there is no package for registry.terraform.io/hashicorp/kubernetes 2.14.0 cached in .terraform/providers",
      "detail": ""
    }
  ]
}
```

### Your environment

* **OS:** macOS

`garden version`
0.12.46

10ko:
A user was encountering an error when running Garden against a project using the Garden Terraform provider.
More specifically, Garden is supposed to run `terraform init` when a required Terraform provider is missing and we do this by checking the error message returned by the Terraform cli.
It's possible, that in-between releases this error message has changed and we were not running `terraform init` when a provider needed a fresh installation.

Fixes #3325

This is a quick fix for unblocking a user, but theoretically, we'd want to rewrite this logic to be more robust (e.g. does the Terraform cli return machine-readable output and error codes we can parse?).

### DIFF
--- a/plugins/terraform/common.ts
+++ b/plugins/terraform/common.ts
@@ -64,7 +64,8 @@ export async function tfValidate(params: TerraformParams) {
     if (
       reasons.includes("Could not satisfy plugin requirements") ||
       reasons.includes("Module not installed") ||
-      reasons.includes("Could not load plugin")
+      reasons.includes("Could not load plugin") ||
+      reasons.includes("Missing required provider")
     ) {
       // We need to run `terraform init` and retry validation
       log.debug("Initializing Terraform")


twelvemo (approved):

vvagaytsev:
It looks like this relates to #3325.

vvagaytsev (approved):

stefreak:
It would be nice if we could add an integration test for this. I think it could be done using one of terraform's utility providers, e.g. null, local, random etc. https://registry.terraform.io/browse/providers?category=utility

plugins/terraform/common.ts
```diff
@@ -64,7 +64,8 @@ export async function tfValidate(params: TerraformParams) {
     if (
       reasons.includes("Could not satisfy plugin requirements") ||
       reasons.includes("Module not installed") ||
-      reasons.includes("Could not load plugin")
+      reasons.includes("Could not load plugin") ||
```
stefreak, line 67: Is there a reason why we must parse the reasons string? Maybe we could also just try terraform init and re-validate in all cases?

stefreak (dismissed):
I think we can make this more robust by skipping comparing the reasons string, and just trying terraform init in all cases where terraform validate returns `{ valid: false }`, as it doesn't hurt, even if it doesn't help.

stefreak:
I added some tests for different terraform versions in #3633, and they actually fail without this fix. So we have the tests covered there :)

stefreak:
Let's actually do that later and get this merged first

PR #3636, opened 2023-02-01
chore(ci): add automatic edge build for the 0.13 branch

edvald:
Now that we have the binary build working on the 0.13 line, we can add a pre-release and automatic build on every merge to the 0.13 branch.

I wanted to give the release line a codename, instead of just referring to "oh point thirteen" all the time. Some ideas were tossed around and _Bonsai_ became a favorite among the ones I talked to. So the new edge release tag is `edge-bonsai`.

The general idea is that from here on we progress these codenames for each major version line of Garden. Retroactively we should give the 0.12 a name beginning with A, so the 0.13 codename begins with B and so forth.

Please chime in if you don't like the name :)


Orzelius (approved):
Code looks good. I love bonsai! For `A` the following come to mind:
1. Apple, an easy choice
2. Avacado
3. Aronia, tastes funny and causes constipation if eaten in a large quantity. This' my favorite

PR #3637, opened 2023-02-01
refactor(k8s): remove rsync mode for syncing build contexts to in-cluster build

edvald:
We've been testing the mutagen sync mode for a while now, so this feels quite
safe to do. This also means that rsync is no longer necessary on the client
machine except for Windows (which will be addressed separately).

Also removed the rsync dependency from the Homebrew formula for Mac.


vvagaytsev (approved):
🚀

PR #3639, opened 2023-02-01
0.13: implement container module conversion tests

Orzelius:
partially solves #3629

### DIFF
--- a/core/test/unit/src/plugins/container/container.ts
+++ b/core/test/unit/src/plugins/container/container.ts
@@ -8,11 +8,15 @@
 
 import td from "testdouble"
 
-import { Garden } from "../../../../../src/garden"
 import { PluginContext } from "../../../../../src/plugin-context"
-import { gardenPlugin, ContainerProvider } from "../../../../../src/plugins/container/container"
-import { getDataDir, makeTestGarden } from "../../../../helpers"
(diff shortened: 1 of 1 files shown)


core/test/unit/src/plugins/container/container.ts
```diff
@@ -82,20 +87,72 @@ describe("plugins.container", () => {
...
+      const result = await convertContainerModule({ ...getModuleConvertBaseParams(module), dummyBuild })
+      const build = result.group.actions.find((a) => a.kind === "Build")!
+      expect(build).to.exist
+      expect(build.type).to.be.eql("exec")
+      expect(build.name).to.be.eql("dummyBuild")
     })
 
-    it("sets spec.image from module image field", async () => {
-      throw "TODO"
+    it("sets spec.localId from module image field", async () => {
```
Orzelius, line 148: This should be `spec.localId` not `spec.image`, right?
The set is done [here](https://github.com/garden-io/garden/blob/0.13/core/src/plugins/container/container.ts#L303)

vvagaytsev: I think so, looks correct to me.

vvagaytsev (approved):
LGTM, thank you! 👍

Issue #3549, opened 2023-01-18
Fix tests in plugins/exec


vvagaytsev:
Fix specs in `plugins/exec/exec.ts`. The problematic specs are:
* [ ] `should propagate task logs to runtime outputs` -> fails on template string resolution: `Invalid template string (${runtime.tasks.task-a.outputs.log}): Could not find key task-a under runtime.tasks`. **Possible cause:** the [`ActionSpecContext`](https://github.com/garden-io/garden/blob/0.13/core/src/tasks/resolve-action.ts#L146) is not complete. -> to be fixed in #3694 
* [x] `should copy artifacts after test runs` -> fixed test action key in #3551 (https://github.com/garden-io/garden/pull/3551/commits/8b16dac3d877f69e348daf96baad655ee02a4377)
* [x] `should throw if a local exec module has a build.copy spec` -> fixed in #3594 
* [x] `build` -> fixed test assertions in #3551 (https://github.com/garden-io/garden/pull/3551/commits/27e2f8163e46d4bd48ab13806eb1195b1b8c95dc)
* [ ] `testExecModule` -> likely some actions misconfiguration at `internal.basePath: "TODO-G2"` -> draft PR is #3566 
  * [ ] `should run the test command in the module dir if local true`
  * [ ] `should receive module version as an env var`
* [x] `runExecTask` -> fixed test assertions in #3551 (https://github.com/garden-io/garden/pull/3551/commits/bb04022fab4866284dd70250ff2e0b6e77aae407) - see `TODO-G2` comments in the commit
* [ ] `convert` -> implement missing tests
  * -> partially fixed in #3647 
  * -> more fixes were done in #3718 
  * -> the last fixes were done in #3689  
* [x] `services`
  * [x] `deployExecService` -> fixed test assertions in #3551 (https://github.com/garden-io/garden/pull/3551/commits/3f3c997d107653cebfc4fe7b352fb4d7a05ebc5e)
  * [x] `getExecServiceStatus` -> fixed test assertions in #3551 (https://github.com/garden-io/garden/pull/3551/commits/9dc05a5f1843d68ed910d5e1b5b323093f84749e)
  * [x] `deleteExecService` -> fixed in #3551 (https://github.com/garden-io/garden/pull/3551/commits/e18eeb032dd6bcd42790a875cd263e1ebfb0b1fc and https://github.com/garden-io/garden/pull/3551/commits/80b447f86b34a1cd35a52fc4f30b0b473ef19eec)

[...]

PR #3647, opened 2023-02-01
0.13 implement `exec` plugin tests

vvagaytsev:

Partially addresses #3549 

Please note 2 open discussions.

### DIFF
--- a/core/test/unit/src/plugins/exec/exec.ts
+++ b/core/test/unit/src/plugins/exec/exec.ts
@@ -12,7 +12,7 @@ import { join } from "path"
 import psTree from "ps-tree"
 
 import { Garden } from "../../../../../src/garden"
-import { gardenPlugin, getLogFilePath } from "../../../../../src/plugins/exec/exec"
+import { ExecProvider, gardenPlugin, getLogFilePath } from "../../../../../src/plugins/exec/exec"
 import { LogEntry } from "../../../../../src/logger/log-entry"
 import { keyBy } from "lodash"
 import { getDataDir, makeTestModule, expectError, createProjectConfig } from "../../../../helpers"
@@ -28,18 +28,24 @@ import { sleep } from "../../../../../src/util/util"
 import { configureExecModule } from "../../../../../src/plugins/exec/moduleConfig"
 import { actionFromConfig } from "../../../../../src/graph/actions"
 import { TestAction, TestActionConfig } from "../../../../../src/actions/test"
+import { GardenModule } from "../../../../../src/types/module"
+import { PluginContext } from "../../../../../src/plugin-context"
 
 describe("exec plugin", () => {
   const testProjectRoot = getDataDir("test-project-exec")
   const plugin = gardenPlugin()
 
   let garden: Garden
+  let ctx: PluginContext
+  let execProvider: ExecProvider
   let graph: ConfigGraph
   let log: LogEntry
 
   beforeEach(async () => {
     garden = await makeTestGarden(testProjectRoot, { plugins: [plugin] })
     graph = await garden.getConfigGraph({ log: garden.log, emit: false })
+    execProvider = await garden.resolveProvider(garden.log, "exec")
+    ctx = await garden.getPluginContext({ provider: execProvider, templateContext: undefined, events: undefined })
     log = garden.log
     await garden.clearBuilds()
   })
@@ -370,8 +376,6 @@ describe("exec plugin", () => {
         },
         spec: { local: true },
       })
-      const provider = await garden.resolveProvider(garden.log, "test-plugin")
-      const ctx = await garden.getPluginContext({ provider, templateContext: undefined, events: undefined })
       await expectError(async () => await configureExecModule({ ctx, moduleConfig, log }), "configuration")
     })
   })
--- a/core/test/unit/src/plugins/exec/exec.ts
+++ b/core/test/unit/src/plugins/exec/exec.ts
@@ -15,7 +15,14 @@ import { Garden } from "../../../../../src/garden"
 import { ExecProvider, gardenPlugin, getLogFilePath } from "../../../../../src/plugins/exec/exec"
 import { LogEntry } from "../../../../../src/logger/log-entry"
 import { keyBy } from "lodash"
-import { getDataDir, makeTestModule, expectError, createProjectConfig } from "../../../../helpers"
+import {
+  getDataDir,
+  makeTestModule,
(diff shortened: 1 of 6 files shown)


core/test/unit/src/plugins/exec/exec.ts
```diff
@@ -507,17 +522,163 @@ describe("exec plugin", () => {
...
+      const actionsFromModuleB = result.groups.find((g) => g.name === moduleNameB)?.actions! as BuildActionConfig[]
+      expect(actionsFromModuleB).to.exist
+      expect(actionsFromModuleB.length).to.eql(1)
+
+      const buildB = actionsFromModuleB.find((a) => a.kind === "Build")!
+      expect(buildB).to.exist
+      expect(buildB.name).to.eql(moduleNameB)
+      expect(buildB.spec.command).to.eql(buildCommandB)
+      // TODO-G2 this still fails; check if is the test config correct
+      expect(buildB.copyFrom).be.not.empty
```
vvagaytsev, line 681: @edvald does this spec reflect the initial intention? It still fails, so I'm wondering if there is something wrong in the test setup or a bug in the code.

core/test/unit/src/plugins/exec/exec.ts
```diff
@@ -528,17 +697,138 @@ describe("exec plugin", () => {
...
 
     it("correctly maps a serviceConfig to a Deploy with a build", async () => {
-      throw "TODO"
-
       // Dependencies
       // build field
       // timeout
       // service spec
+
+      const moduleNameA = "module-a"
```
vvagaytsev, line 754: @edvald I'm not sure if I got the intention of this spec right. Could you check this, please?

One more question, what is the expected config and behavior difference between `* with a build` and `with no build` specs?

core/test/unit/src/plugins/exec/exec.ts
edvald: As far as I can tell the test code looks correct, might be an actual bug.

core/test/unit/src/plugins/exec/exec.ts
edvald: I think the only thing missing is validating that the `build` field is correctly set on the Deploy action configs that are output from the conversion. Each Deploy should have `build` set to the module name (since that will be the name of the Build action that the conversion returns). 

You could also just have one module in this particular test, although the other bits you're testing for (the service to service dependency) might be good to just peel out to a separate test.

Orzelius (commented):
I really like this no test-project approach! Will use in the future! Had a few minor comments you can take a look at.

core/test/unit/src/plugins/exec/exec.ts
```diff
@@ -507,19 +527,168 @@ describe("exec plugin", () => {
...
+        path: tmpDirResult.path,
+        providers: [{ name: "exec" }],
+      })
+
+      return TestGarden.factory(tmpDirResult.path, { config, plugins: [gardenPlugin()] })
+    }
+
+    let tmpDir: tmp.DirectoryResult
+    let tmpGarden: TestGarden
+    let tmpGraph: ConfigGraph
```
Orzelius, line 559: This var can be removed, right? You need to get a fresh graph after setting the configs anyway so it's not needed.

core/test/unit/src/plugins/exec/exec.ts
```diff
@@ -507,19 +527,168 @@ describe("exec plugin", () => {
...
+            tasks: [
+              {
+                name: "task-a",
+                command: taskCommand,
+              },
+            ],
+          },
+        }),
+      ])
+      tmpGraph = await tmpGarden.getConfigGraph({ log: tmpGarden.log, emit: false })
```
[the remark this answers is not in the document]
vvagaytsev: Good catch! Agree, I'm thinking about moving the whole `convert` spec to the root level `describe`-spec.

core/test/unit/src/plugins/exec/exec.ts
```diff
@@ -507,19 +527,168 @@ describe("exec plugin", () => {
...
+      tmpDir = await tmp.dir({ unsafeCleanup: true })
+      await execa("git", ["init", "--initial-branch=main"], { cwd: tmpDir.path })
+      tmpGarden = await makeGarden(tmpDir)
+    })
+
+    after(async () => {
+      await tmpDir.cleanup()
+    })
+
+    beforeEach(async () => {})
```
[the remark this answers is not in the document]
vvagaytsev: Removed in https://github.com/garden-io/garden/pull/3647/commits/c9f77d248276cfbc1d40a45feb928bd160a501b1

core/test/unit/src/plugins/exec/exec.ts
vvagaytsev: Right, refactored it in https://github.com/garden-io/garden/pull/3647/commits/c9f77d248276cfbc1d40a45feb928bd160a501b1

core/test/unit/src/plugins/exec/exec.ts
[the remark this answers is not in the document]
vvagaytsev: Split the contexts in https://github.com/garden-io/garden/pull/3647/commits/0e51cca64b5185ed81b104319d14abde8a1c0802.

Removed `tmp` prefix in https://github.com/garden-io/garden/pull/3647/commits/0e5d53ecde5be6c34047ae5e2fb44b4344c53908.

core/test/unit/src/plugins/exec/exec.ts
vvagaytsev: @edvald could you check the comment above, please?

Orzelius (approved):
Lovely!

vvagaytsev:
I'll merge this as it is. The [open discussion](https://github.com/garden-io/garden/pull/3647#discussion_r1095813366) can be resolved in a separate PR.

core/test/unit/src/plugins/exec/exec.ts
thsig: Come to think of it, I can't see any reason why we need the `dependencies` field on the `BaseBuildSpec` interface (which is where `ModuleConfig.spec.build.dependencies` comes from).

Should we just get rid of `BaseBuildSpec.dependencies` and remove all references to it in the code? I think that should be relatively straightforward.

vvagaytsev: Thanks, @thsig! I've made the changes in #3686.

PR #3651, opened 2023-02-01
chore: switch from tslint to eslint

### LINKED ISSUE #3648: chore: switch from tslint to eslint
TSLint has been deprecated and using ESLint is the recommended approach.

Some resources:
- https://code.visualstudio.com/api/advanced-topics/tslint-eslint-migration
- https://github.com/typescript-eslint/tslint-to-eslint-config

Walther:

Switches from the deprecated TSLint to ESLint in our main repository.

Fixes #3648 

We should do another pass for simplifying and combining the various `.esllintrc.js` files into one good centralized configuration. This is just the initial pass to get the switch done.

Squash before merging.


Orzelius:
I thought we pushed this into the future as it will make syncing changes with the 0.12 harder? Looking at the changes it doesn't seem to be that big of a deal though.

Walther:
> I thought we pushed this into the future as it will make syncing changes with the 0.12 harder? Looking at the changes it doesn't seem to be that big of a deal though.
Sorry, I got distracted and tried my hand at this to see how much work it would be, and apparently it was not a big change after all. The changes required were all fairly minimal, so I feel like there should be fairly minimal risk.

> It looks like there are some conflicts.
Thanks for noticing! My base branch wasn't quite up to date, so there was one conflict in `test/unit/src/plugins/container/container.ts`. Fixed it.

Of course if we feel like this is not the right time for this PR, we can do this again some other time. Key steps for this project:

- find each directory that has a `tslint.json` file
- run `npx tslint-to-eslint-config --prettier --comments`
- edit `package.json` to remove references to `tslint`
- edit the generated `.eslintrc.json` to remove references to `tslint`
- edit `package.json` to change the `lint` script incantation
- run `yarn run lint`, optionally with `--fix`
- repeat for all relevant directories
- done!

edvald:
Nice! Thanks for taking this on :) I think now is a great time for this, very good to shed more of deprecated packages from the repo and improve our DC along the way.

plugins/conftest-kubernetes/index.ts
```diff
@@ -44,12 +44,12 @@ export const gardenPlugin = () =>
         const allTestNames = new Set(actions.filter((a) => a.kind === "Test").map((m) => m.name))
 
         return {
-          addActions: await Bluebird.filter(actions, async (action) => {
-            return (
```
TimBeyer, line 48: I feel like we should deactivate whatever rule does this.
IMO it's much nicer to require the `return` even if it's just a oneliner since it's more readable and makes it easier to add extra lines to the logic.

Walther:
`check-docs` fails, apparently this [helm user guide](https://docs.garden.io/guides/using-helm-charts) has been removed from docs - but unrelated to this pr

TimBeyer (approved):

edvald (approved):

PR #3662, opened 2023-02-02
0.13: container module build handler tests

Orzelius:
partially fixes https://github.com/garden-io/garden/issues/3629


core/test/unit/src/plugins/container/build.ts
```diff
@@ -6,160 +6,123 @@
...
     })
   })
 
-  describe("build", () => {
+  describe("buildContainer", () => {
     beforeEach(() => {
-      td.replace(helpers, "checkDockerServerVersion", () => null)
-    })
-
-    it("should pull image if image tag is set and the module doesn't container a Dockerfile", async () => {
```
Orzelius: The pull logic seems to have moved somewhere else from the container build handler. I failed to find where it resides so I just removed the test for now.

core/test/unit/src/plugins/container/build.ts
```diff
@@ -6,160 +6,123 @@
...
-      const result = await build({ ctx, log, module })
-
-      expect(result).to.eql({
-        buildLog: "log",
-        fresh: true,
-        details: { identifier: "some/image" },
-      })
-    })
-
-    it("should set build target image parameter if configured", async () => {
```
Orzelius: Also removed this case as I don't quite understand the new container image semantics. Would love an explanation on them and then I can revisit this case.

core/test/unit/src/plugins/container/build.ts
```diff
@@ -6,160 +6,123 @@
...
-      const config = cloneDeep(baseConfig)
-      config.spec.image = "some/image"
-      config.spec.build.targetImage = "foo"
-      const module = td.object(await getTestModule(config))
+    it("should build image if module contains Dockerfile", async () => {
+      const action = await getAction()
 
-      td.replace(helpers, "hasDockerfile", () => true)
-      td.replace(helpers, "imageExistsLocally", async () => false)
+      td.replace(action, "getOutputs", () => ({ localImageId: "some/image" }))
```
Orzelius, line 88: If anybody had a better or a simpler way than a mock I'm all ears

Orzelius (commented):
I thought I'd be easier to review if I copy existing code in one commit and then make the changes in another but the diff is a massive mess so just feel free to click on the "Files changed"

edvald (approved):

PR #3666, opened 2023-02-03
feat(cloud): Dashboard deprecation notice and garden cloud dashboard fallback

mkhq:

This PR clarifies the status of the garden dashboard which will be deprecated in the next major garden release (0.13). As an alternative, we are introducing a cloud version of the dashboard. The Garden Cloud dashboard can be used by logging in via the core command `garden login`. To stop using the dashboard it is possible to logout by using the `garden logout` command.

Fixes #

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -65,7 +65,7 @@ import { JsonFileWriter } from "../logger/writers/json-file-writer"
 import { dedent } from "../util/string"
 import { renderDivider } from "../logger/util"
 import { emoji as nodeEmoji } from "node-emoji"
-import { GlobalConfigStore, LocalConfigStore, RequirementsCheck } from "../config-store"
+import { GlobalConfigStore, RequirementsCheck } from "../config-store"
 
 export async function makeDummyGarden(root: string, gardenOpts: GardenOpts) {
   const environments = gardenOpts.environmentName
@@ -287,13 +287,16 @@ ${renderCommands(commands)}
     let cloudApi: CloudApi | null = null
     if (!command.noProject) {
       const config: ProjectResource | undefined = await this.getProjectConfig(workingDir)
-      const gardenDirPath = resolve(workingDir, DEFAULT_GARDEN_DIR_NAME)
-      const localConfig = await new LocalConfigStore(gardenDirPath).get()
-      const cloudDomain: string | undefined = getGardenCloudDomain(config, localConfig)
 
-      if (cloudDomain) {
-        cloudApi = await CloudApi.factory({ log, cloudDomain })
-      } else {
+      let cloudDomain: string | undefined = undefined
+
+      if (config) {
+        cloudDomain = getGardenCloudDomain(config)
+      }
+
+      cloudApi = await CloudApi.factory({ log, cloudDomain })
+
+      if (!cloudApi) {
         log.debug("Cloud domain not configured, proceeding without a Garden Cloud connection.")
       }
     }
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -45,13 +45,7 @@ import {
   parseEnvironment,
   ProjectResource,
 } from "../config/project"
-import {
-  ERROR_LOG_FILENAME,
-  DEFAULT_API_VERSION,
-  DEFAULT_GARDEN_DIR_NAME,
-  LOGS_DIR_NAME,
-  DEFAULT_GARDEN_CLOUD_DOMAIN,
-} from "../constants"
+import { ERROR_LOG_FILENAME, DEFAULT_API_VERSION, DEFAULT_GARDEN_DIR_NAME, LOGS_DIR_NAME } from "../constants"
 import { generateBasicDebugInfoReport } from "../commands/get/get-debug-info"
 import { AnalyticsHandler } from "../analytics/analytics"
 import { BufferedEventStream, ConnectBufferedEventStreamParams } from "../cloud/buffered-event-stream"
@@ -294,23 +288,9 @@ ${renderCommands(commands)}
     if (!command.noProject) {
       const config: ProjectResource | undefined = await this.getProjectConfig(workingDir)
 
-      let cloudDomain: string | undefined = undefined
-
-      if (config) {
-        cloudDomain = getGardenCloudDomain(config)
-      }
+      let cloudDomain: string = getGardenCloudDomain(config)
 
       cloudApi = await CloudApi.factory({ log, cloudDomain })
-
-      if (!cloudApi) {
-        log.debug("Cloud domain not configured, proceeding without a Garden Cloud connection.")
-
-        const msg = dedent`
-        \n${nodeEmoji.lightning}   ${chalk.cyan.bold(`Sign-up for our beta of Garden Cloud`)}
-        ${nodeEmoji.link}  ${chalk.blueBright.underline(DEFAULT_GARDEN_CLOUD_DOMAIN)}
-      `
-        footerLog.setState(msg)
-      }
     }
 
     // Init event & log streaming.
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -65,7 +65,7 @@ import { JsonFileWriter } from "../logger/writers/json-file-writer"
 import { dedent } from "../util/string"
 import { renderDivider } from "../logger/util"
 import { emoji as nodeEmoji } from "node-emoji"
-import { GlobalConfigStore, RequirementsCheck } from "../config-store"
+import { GlobalConfigStore, LocalConfigStore, RequirementsCheck } from "../config-store"
 
 export async function makeDummyGarden(root: string, gardenOpts: GardenOpts) {
   const environments = gardenOpts.environmentName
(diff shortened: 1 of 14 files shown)


core/src/commands/login.ts
```diff
@@ -38,23 +40,39 @@ export class LoginCommand extends Command {
...
     const projectConfig: ProjectResource | undefined = await cli!.getProjectConfig(currentDirectory)
-    const cloudDomain: string | undefined = getGardenCloudDomain(projectConfig)
+
+    // Garden works by default without Garden Cloud. In order to use cloud, a domain
+    // must be known to cloud for any command needing a logged in user.
+    //
+    // The cloud domain is resolved in the following order:
+    // - 1. GARDEN_CLOUD_DOMAIN config variable
+    // - 2. `domain`-field from the project config
+    // - 3. cloud.domain key in the .garden/local-config.yml
```
eysi09, line 54: As per my earlier comment, can't we just hard code this as opposed to storing in local config?

core/src/commands/logout.ts
```diff
@@ -32,22 +33,27 @@ export class LogOutCommand extends Command {
...
+    const localConfig: LocalConfig | undefined = await garden.configStore.get()
+    const cloudDomain: string | undefined = getGardenCloudDomain(projectConfig, localConfig)
+    const distroName = getCloudDistributionName(cloudDomain || "")
 
-    if (!token) {
-      log.info({ msg: `You're already logged out from ${distroName}.` })
-      return {}
+    if (!cloudDomain) {
+      throw new ConfigurationError(
+        `Could not find a domain, you are either not logged in or has not defined a project domain.`,
```
eysi09, line 43: Again, if we hard code the fallback value we never hit this code path.

core/src/constants.ts
```diff
@@ -50,6 +50,8 @@ export const SEGMENT_PROD_API_KEY = "b6ovUD9A0YjQqT3ZWetWUbuZ9OmGxKMa" // ggigno
 export const DOCS_BASE_URL = "https://docs.garden.io"
 export const VERSION_CHECK_URL = "https://get.garden.io/version"
 
+export const DEFAULT_GARDEN_CLOUD_DOMAIN = "https://cloud.app.garden"
```
eysi09, line 53: Ah right, here it is. Doesn't this eliminate the need for using the local config store?

core/src/cloud/api.ts
```diff
@@ -149,6 +150,9 @@ export function getGardenCloudDomain(projectConfig?: ProjectResource): string |
     cloudDomain = new URL(gardenEnv.GARDEN_CLOUD_DOMAIN).origin
   } else if (projectConfig.domain) {
     cloudDomain = new URL(projectConfig.domain).origin
+  } else if (localConfig && localConfig.cloud) {
```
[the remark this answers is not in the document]
eysi09: So essentially:

If cloud domain is set via env, use that;
else if cloud domain in project config use that;
else use default cloud domain (cloud.app.garden).

eysi09 (commented):
Great work! 

Had some comments on storing the domain in the local config versus just hard coding the fallback value and using the auth token to check if a user is logged in as per usual. 

As a side note, we're doing away with sqlite in v0.13 and making some changes to the local config so there will probably be some conflicts when porting this over.

core/src/cloud/api.ts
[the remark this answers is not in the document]
mkhq: Refactored to depend on the auth token instead, I think the code is much simpler as a result and no dependency on the local config.

core/src/constants.ts
mkhq: Yes, fixed 👍

core/src/cli/cli.ts
```diff
@@ -287,12 +293,23 @@ ${renderCommands(commands)}
...
       const config: ProjectResource | undefined = await this.getProjectConfig(workingDir)
-      const cloudDomain: string | undefined = getGardenCloudDomain(config)
 
-      if (cloudDomain) {
-        cloudApi = await CloudApi.factory({ log, cloudDomain })
-      } else {
-        log.debug("Cloud/Enterprise domain not configured. Aborting.")
+      let cloudDomain: string | undefined = undefined
+
+      if (config) {
```
eysi09, line 299: Can't we skip the check and initialise the Cloud API class directly? Assuming there's always a Cloud domain.

core/src/cli/cli.ts
```diff
@@ -287,12 +293,23 @@ ${renderCommands(commands)}
...
+        cloudDomain = getGardenCloudDomain(config)
+      }
+
+      cloudApi = await CloudApi.factory({ log, cloudDomain })
+
+      if (!cloudApi) {
+        log.debug("Cloud domain not configured, proceeding without a Garden Cloud connection.")
```
eysi09, line 306: There's always a Cloud domain so this just means the user is not logged in. We already log that case in the factory function so perhaps we can just skip this.

core/src/cli/cli.ts
eysi09, line 299: Can't we skip this check and initialise the class directly since there's a always a Cloud domain?

core/src/cli/cli.ts
```diff
@@ -287,12 +293,23 @@ ${renderCommands(commands)}
...
+
+        const msg = dedent`
+        \n${nodeEmoji.lightning}   ${chalk.cyan.bold(`Sign-up for our beta of Garden Cloud`)}
```
eysi09, line 309: This feels a bit too much imo :) 

Some users may decide not to use the Cloud dashboard and it would be pretty annoying to see this message all the time.

core/src/cloud/api.ts
```diff
@@ -138,11 +138,11 @@ function toCloudProject(
  * A helper function to get the cloud domain from a project config. Uses the env var
  * GARDEN_CLOUD_DOMAIN to override a configured domain.
  */
-export function getGardenCloudDomain(projectConfig?: ProjectResource): string | undefined {
-  if (!projectConfig) {
-    return undefined
-  }
+export function getGardenCloudDomainWithFallback(projectConfig: ProjectResource): string {
```
eysi09, line 141: Isn't it enough to just a have single `gardenGardenCloudDomain` function that also returns the fallback?

Or is there a use case where we don't want the fallback at all.

core/src/cloud/api.ts
```diff
@@ -196,18 +196,22 @@ export class CloudApi {
...
-    cloudDomain: string
+    cloudDomain?: string
     skipLogging?: boolean
   }) {
     log.debug("Initializing Garden Cloud API client.")
 
     const token = await CloudApi.getClientAuthTokenFromDb(log)
     if (!token && !gardenEnv.GARDEN_AUTH_TOKEN) {
-      log.debug("User is not logged in. Aborting.")
+      log.debug("User is not logged in, proceeding without a cloud API connection.")
```
eysi09, line 206: Can we change this to:

"No auth token found, proceeding with out access to Garden Cloud. Command results for this command run will not available in Garden Cloud."

core/src/cloud/api.ts
```diff
@@ -196,18 +196,22 @@ export class CloudApi {
...
       return null
     }
 
-    const api = new CloudApi(log, cloudDomain)
+    // The user token is set at this point. If the cloud domain is not set
+    // it means we are logged in to the default domain
+    const fallbackCloudDomain: string = cloudDomain || DEFAULT_GARDEN_CLOUD_DOMAIN
```
eysi09, line 212: As per my suggestion above, shouldn't we just ensure that the cloud domain is passed to class on initialisation? So essentially make the `cloudDomain` param required and ensure that the `getGardenCloudDomain` helper function returns a valid domain with the follow order of precedence env var => config => default domain.

core/src/commands/dashboard.ts
```diff
@@ -33,13 +34,17 @@ type Opts = typeof dashboardOpts
...
   alias = "serve"
-  help = "Starts the Garden dashboard for the current project and environment."
+  help = "[DEPRECATED] Starts the Garden dashboard for the current project and environment."
 
   cliOnly = true
   streamEvents = true
   private garden?: Garden
 
   description = dedent`
+    Deprecation Warning: The Garden Dashboard will be deprecated in the next major Garden release (0.13). As an alternative you can try out a beta version of Garden Cloud at ${DEFAULT_GARDEN_CLOUD_DOMAIN}.
```
eysi09, line 44: Can we change this to: "beta version of the Cloud dashboard"?

core/src/commands/login.ts
```diff
@@ -37,24 +37,40 @@ export class LoginCommand extends Command {
...
+    // must be known to cloud for any command needing a logged in user.
+    //
+    // The cloud domain is resolved in the following order:
+    // - 1. GARDEN_CLOUD_DOMAIN config variable
+    // - 2. `domain`-field from the project config
+    // - 3. fallback to the default garden cloud domain
+    //
+    // If the fallback was used, we rely on the token to decide if the Cloud API instance
+    // should use the default domain or not. The token lifecycle ends on logout.
+    let cloudDomain: string = getGardenCloudDomainWithFallback(projectConfig)
```
eysi09, line 64: See me comment above about just having a single `getGardenCloudDomain` function.

core/src/commands/login.ts
```diff
@@ -37,24 +37,40 @@ export class LoginCommand extends Command {
...
+
     try {
-      const enterpriseApi = await CloudApi.factory({ log, cloudDomain, skipLogging: true })
+      const cloudApi = await CloudApi.factory({ log, cloudDomain, skipLogging: true })
 
-      if (enterpriseApi) {
-        log.info({ msg: `You're already logged in to ${distroName}.` })
-        enterpriseApi.close()
+      if (cloudApi) {
+        log.info({ msg: `You're already logged in to ${distroName} at ${cloudDomain}.` })
```
eysi09, line 72: Do we need to print the domain?

eysi09 (commented):
A few more comments :)

core/test/unit/src/commands/logout.ts
```diff
@@ -74,7 +110,7 @@ describe("LogoutCommand", () => {
     await command.action(makeCommandParams({ cli, garden, args: {}, opts: {} }))
 
     const logOutput = getLogMessages(garden.log, (entry) => entry.level === LogLevel.info).join("\n")
-    expect(logOutput).to.include("You're already logged out from Garden Enterprise.")
+    expect(logOutput).to.include("You're already logged out from http://dummy-domain.com.")
```
edvald, line 113: Maybe make sure @Walther doesn't see this 🙃

eysi09 (approved):

eysi09 (approved):

PR #3668, opened 2023-02-03
ci: use audit egress-policy in the scorecard GH action

vvagaytsev:
This is the continuation of #3664. The [issue](https://github.com/garden-io/garden/actions/runs/4076470288) still happens, but with different addresses.

This PR changes the `egress-policy` to `audit` to avoid maintenance of the `allowed-endpoints`.

Fixes #

### DIFF
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -27,16 +27,19 @@ jobs:
       - name: Harden Runner
         uses: step-security/harden-runner@18bf8ad2ca49c14cbb28b91346d626ccfb00c518
         with:
-          egress-policy: block
-          allowed-endpoints: >
-            api.github.com:443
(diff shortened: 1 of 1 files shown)


Walther:
now that the CI has run with the `audit` setting https://github.com/garden-io/garden/actions/runs/4082821135/jobs/7037614686 can we copy the list from the recommended policies https://app.stepsecurity.io/github/garden-io/garden/actions/runs/4082821135 and set it back to `block`?

stefreak (approved):
👍 

After it ran with audit a couple of times with audit, we can copy the allowed endpoints from the step-security website. The link is in the action output.

The idea is to prevent supply chain attacks by blocking unexpected network access. More information can be found at https://github.com/step-security/harden-runner

PR #3669, opened 2023-02-03
0.13: attempt to fix unit test oom by increasing heap size

Orzelius:


Walther (approved):

Walther:
let's land this on `main` as well for parity

Orzelius:
> let's land this on main as well for parity

I'm still quite sure it's the code that's wrong and main seems to be working fine right now.

PR #3671, opened 2023-02-03
Interactive dev command

edvald:
Ah, it's finally taking shape! Plenty to do still, but the basic mechanics are there, including type-ahead, autocomplete, and the ability to run basically any Garden command.

Kapture 2023-02-07 at 19 29 57

The remaining TODOs below should be addressed in a separate PR, particularly since we're going to need to rework all things related to watching (see #3702 for one, which I'd like to see merged before proceeding).

TODO:

- [ ] Handle multiple watch/dev/sync commands concurrently (i.e. move that logic to the Garden instance)
- [x] Gracefully handle global options
- [x] Print nice intro message on start
- [x] Add reload command to reload config
- [ ] Notify when config changes are detected
- [x] Show active commands in status line
- [ ] Better logging when initial config load doesn't work
- [ ] See why argument suggestions don't give a highlighted result
- [ ] Allow multiple service names in args (both with comma and variadic)


eysi09:
Can we get a screen cap plz? 🙏

core/src/commands/base.ts
```diff
@@ -315,6 +319,14 @@ export abstract class Command<T extends Parameters = {}, U extends Parameters =
   }
 }
 
+export abstract class InteractiveCommand<A extends Parameters = {}, O extends Parameters = {}, R = any> extends Command<
```
Orzelius, line 321: Shouldn't this be parameters and options instead of two parameters? Also, what does `R` stand for?

core/src/commands/dev.tsx
```diff
@@ -113,12 +114,29 @@ Let's get your development environment wired up.
     return super.action(params)
   }
 
+  async reload(log: LogEntry, garden: Garden, commandLine: CommandLine) {
+    try {
+      const newGarden = await Garden.factory(garden.projectRoot, garden.opts)
+      const configDump = await garden.dumpConfig({ log })
```
Orzelius: Shouldn't this also grab the config dump from the new garden?

core/src/commands/serve.ts
```diff
@@ -86,16 +95,148 @@ export class ServeCommand<
...
+        .catch(() => {})
     })
   }
+
+  async reload(log: LogEntry, garden: Garden) {
+    this.commandLine?.disable("Loading Garden project...")
+
+    try {
+      const newGarden = await Garden.factory(garden.projectRoot, garden.opts)
+      const configDump = await garden.dumpConfig({ log })
```
Orzelius, line 125: Same dump config comment here

core/src/cli/command-line.ts
```diff
@@ -505,6 +505,22 @@ ${renderDivider({ width, char, color })}
...
+    const name = command.getFullName()
+
+    if (command.isPersistent(params)) {
+      if ((name === "test" || name === "run") && opts["interactive"]) {
+        // Specific error for interactive commands
+        this.flashError(`Commands cannot be run in interactive mode in the dev console. Please run those separately.`)
+      } else if (name === "dev") {
+        this.flashError(`Nice try :)`)
+      } else if (name === "logs") {
+        this.flashError(`Logs cannot currently be followed in the dev console. Please use a separate sheterminalll.`)
```
Orzelius, line 517:
```suggestion
        this.flashError(`Logs cannot currently be followed in the dev console. Please use a separate terminal.`)
```

Orzelius (commented):
This is very cool but I wonder how big the downside of us having to emulate a terminal like this is. Many users have their own preferences and plugins for their shell and will miss them in our solution. I guess it's a necessary evil.

core/src/logger/writers/file-writer.ts
```diff
@@ -56,7 +56,7 @@ export class FileWriter extends Writer {
   protected fileTransportOptions: FileTransportOptions
 
   constructor(logFilePath: string, config: FileWriterConfig) {
-    super(config.level)
+    super({ level: config.level })
```
vvagaytsev, line 59: This can be simplified. There is a duplicate assignment of `this.level` here and in L62.
The following changes can be done:
1. `FileWriterConfig` can extend `BaseWriterParams`
2. This line can be changed to `super(config)` (no need to create one more object)
3. assignment `this.level = level` can be removed

core/src/cli/command-line.ts
```diff
@@ -0,0 +1,474 @@
...
+        this.renderCommandLine()
+      }
+    })
+
+    this.setKeyHandler("rightArrow", () => {
+      if (this.cursorPosition < this.currentCommand.length) {
+        this.moveCursor(this.cursorPosition + 1)
+        this.renderCommandLine()
+      }
+    })
```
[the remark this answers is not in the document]
edvald: Yeah I've had the same thought 👍

core/src/commands/base.ts
edvald: No, this is correct. See the Command class that's being extended.

core/src/commands/dev.tsx
edvald: It should, yes.

core/src/cli/command-line.ts
```diff
@@ -0,0 +1,474 @@
...
+        this.historyIndex--
+        this.currentCommand = this.commandHistory[this.historyIndex]
+        if (this.cursorPosition > this.currentCommand.length) {
+          this.moveCursor(this.currentCommand.length)
+        }
+        this.renderCommandLine()
+      }
+    })
+
+    this.setKeyHandler("downArrow", () => {
```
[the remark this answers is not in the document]
edvald: Yeah that'd be nice, but left out of scope.

edvald:
> This is very cool but I wonder how big the downside of us having to emulate a terminal like this is. Many users have their own preferences and plugins for their shell and will miss them in our solution. I guess it's a necessary evil.

I do agree, but this is the only viable way to achieve the desired effect. There may be some maintenance to do, hardening etc. but honestly I think the biggest downside was kinda having to make all that low-level stuff in the first place. But now that's done, so... 🤷 

Obviously this is not at all meant to be a full terminal emulation. It is just a single-line command line. Kinda like how you don't expect Spotlight on macOS to be a full-fledged pluggable thing, or the Cmd+K command palette in webapps etc etc.

edvald:
@Orzelius @vvagaytsev updated after review, hoping this is good to go now.

core/src/cli/command-line.ts
```diff
@@ -0,0 +1,589 @@
...
+import { TypedEventEmitter } from "../util/events"
+import { uuidv4 } from "../util/util"
+import { Autocompleter, AutocompleteSuggestion } from "./autocomplete"
+import { parseCliArgs, pickCommand, processCliArgs, renderCommandErrors, renderCommands } from "./helpers"
+import { GlobalOptions, ParameterValues } from "./params"
+
+const defaultMessageDuration = 2000
+const commandLinePrefix = chalk.yellow("🌼  > ")
+const emptyCommandLinePlaceholder = chalk.gray("<enter command> (enter help for more info)")
+const inputChars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789- _*!@$%&/="
```
eysi09, line 27: Looks like `,` is missing here. I wasn't to run e.g. `deploy backend,frontend`

core/src/cli/autocomplete.ts
```diff
@@ -0,0 +1,258 @@
...
+  priority: number
+}
+
+// TODO: there are plenty of optimizations that can be made here
+// TODO: validate the input string (to e.g. highlight invalid names/arguments/flags in red)
+
+interface AutocompleterParams {
+  log: LogEntry
+  commands: Command[]
+  configDump?: ConfigDump
```
eysi09, line 30: Would it make sense to use something more lightweight here in a future version? The config dump can potentially be quite large and clients would need to first obtain it over the network and then send it back, right?

core/src/cli/command-line.ts
```diff
@@ -0,0 +1,589 @@
...
+      if (this.suggestionIndex >= suggestions.length) {
+        this.suggestionIndex = 0
+      }
+      // Pick the suggestion but remember where we are completing from, so we can roll through more suggestions
+      this.currentCommand = suggestions[this.suggestionIndex].line
+      if (this.autocompletingFrom === -1) {
+        this.autocompletingFrom = this.cursorPosition
+      }
+      // Not using this.moveCursor() here so we don't reset the autocomplete state
+      this.cursorPosition = this.currentCommand.length
```
eysi09, line 460: Should this be: `this.currentCommand.length + 1`?

That way you don't need to hit space after autocompleting to add options or arguments.

core/src/cli/command-line.ts
```diff
@@ -0,0 +1,589 @@
...
+        this.flashError(`Logs cannot currently be followed in the dev console. Please use a separate terminal.`)
+      } else {
+        this.flashError(`This command cannot be run in the dev console. Please run it in a separate terminal.`)
+      }
+      return
+    }
+
+    // Execute the command
+    if (!command.isInteractive) {
+      const msg = `Running ${chalk.white.bold(command.getFullName())}...`
```
eysi09, line 533: Nitpick but it's a little odd how it first prints "Running X..." and then "Running X command...".

Screenshot 2023-02-11 at 09 08 56
Screenshot 2023-02-11 at 09 08 59

core/src/commands/cloud/secrets/secrets-create.ts
```diff
@@ -48,7 +48,7 @@ type Opts = typeof secretsCreateOpts
 
 export class SecretsCreateCommand extends Command<Args, Opts> {
   name = "create"
-  help = "Create secrets"
+  help = "Create secrets in Garden Cloud."
```
eysi09, line 51: Is this needed since the command name is `garden cloud secrets create`? 

I'm asking because so far we've been setting this dynamically depending on the distro name (i.e. Garden Cloud / Garden Enterprise / Cloud Dashboard).

core/src/commands/dev.tsx
```diff
@@ -0,0 +1,211 @@
...
+  help = ""
+  hidden = true
+
+  async action({ commandLine }: CommandParams) {
+    commandLine?.showHelp()
+    return {}
+  }
+}
+
+class QuitCommand extends InteractiveCommand {
```
eysi09, line 163: My first reaction was to type "exit". Should we allow that alias as well?

core/src/commands/exec.ts
```diff
@@ -16,9 +16,12 @@ import { ExecInDeployResult, execInDeployResultSchema } from "../plugin/handlers
 import { executeAction } from "../graph/actions"
 
 const execArgs = {
-  service: new StringParameter({
-    help: "The service to exec the command in.",
+  deploy: new StringParameter({
+    help: "The Deploy to exec the command in.",
```
eysi09, line 20: This does read a little awkward :) 

Was the decision to always refer to "Deploys" as the subject to the "Deploy" verb, as opposed to "Service" or "Deployment"?

core/src/commands/run-workflow.ts
```diff
@@ -313,64 +316,40 @@ function printResult({
...
-export async function runStepCommand({
-  cli,
-  garden,
-  bodyLog,
-  footerLog,
-  headerLog,
-  inheritedOpts,
-  step,
-}: RunStepCommandParams): Promise<CommandResult<any>> {
+export async function runStepCommand(params: RunStepCommandParams): Promise<CommandResult<any>> {
```
eysi09, line 319: These changes look good at a glance but we need to make sure we test this properly with Cloud. In particular that logs works as expected.

core/src/cli/autocomplete.ts
```diff
@@ -0,0 +1,258 @@
...
+    const argSpecs = <Parameter<any>[]>Object.values(command.arguments || {})
+    const argSpec = argSpecs[argIndex]
+
+    if (!argSpec) {
+      // No more positional args
+      return []
+    }
+
+    const prefix = rest.length === 0 ? "" : rest[rest.length - 1]
+    const argSuggestions = argSpec.getSuggestions({ configDump })
```
eysi09, line 187: I'm not getting arg suggestions when testing on my end. E.g. if I type `deploy` I don't get suggestions with deploy action names. Is that expected?

eysi09 (changes requested):
Awesome work! 🚀 

Can't wait to start using it day-to-day. 

Left a handful of comments but the only change I would request is supporting comma so you can type `deploy api,web`. 

Also, some other things I noticed that I'll just add here. But all of this can be flagged as future work.

- I hadn't started local K8s when testing things and got an error resolving providers. There was no real actionable hint and I couldn't use a reload command. Wasn't clear whether I should e.g. kill the process or try again, or just start K8s and carry on.  

- Sometimes the suggestion highlights blue, sometimes it doesn't and I don't understand the difference between the states. See screenshot.

Screenshot 2023-02-11 at 09 47 13
Screenshot 2023-02-11 at 09 47 18

- Process crashes if typing a bad command:

```
 🌼  > <enter command> (enter help for more info)


/Users/eysi/code/garden-io/garden/node_modules/yoga-layout-prebuilt/yoga-layout/build/Release/nbind.js:53
        throw ex;
        ^

ParameterError: Unexpected positional argument "backend" (expected only names)
    at processCliArgs (/Users/eysi/code/garden-io/garden/core/src/cli/helpers.ts:277:13)
    at CommandLine.handleReturn (/Users/eysi/code/garden-io/garden/core/src/cli/command-line.ts:491:42)
    at /Users/eysi/code/garden-io/garden/core/src/cli/command-line.ts:256:45
    at CommandLine.keyStroke (/Users/eysi/code/garden-io/garden/core/src/cli/command-line.ts:183:7)
    at /Users/eysi/code/garden-io/garden/core/src/commands/dev.tsx:94:21
    at ReadStream.handleData (/Users/eysi/code/garden-io/garden/node_modules/ink/src/hooks/use-input.ts:185:5)
    at ReadStream.emit (node:events:525:35)
    at ReadStream.emit (node:domain:489:12)
    at addChunk (node:internal/streams/readable:324:12)
    at readableAddChunk (node:internal/streams/readable:293:11)
    at ReadStream.Readable.push (node:internal/streams/readable:234:10)
    at TTY.onStreamRead (node:internal/stream_base_commons:190:23) {
  detail: { expectedKeys: [ 'names' ], extraValue: 'backend' },
  type: 'parameter'
}

Node.js v18.13.0
```

- The old dev command obviously did too much. I think this one may do too little. The majority of our users who use Garden for inner loop dev start the day by running a single command, e.g. `garden deploy --dev`. With this change it's going to be two commands. I think we should support the former workflow. E.g. by deploying by default and adding an option to skip it or allowing users to configure some initial action. 

---

Again, all of this can be addressed later and I'm sure we'll bump into more things around the "unhappy path" that we can fix in follow-up PRs.

core/src/cli/autocomplete.ts
eysi09: Ah right, this happened because resolving the providers failed. See my review comment. 

Feels like we should either exit the process in that case or keep track of the fact that some config hasn't been resolved and try again. 

Or offer a "reload" option.

eysi09: ...I suppose if we can ensure the "reload" option works even if provider resolution fails, this is a non issue.

eysi09:
One other thing I just noticed (which we can also fix in a follow-up):

If I stop the command with a SIGINT signal (e.g. ctrl+c) I get the following error:

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    at Dev (/Users/eysi/code/garden-io/garden/core/build/src/commands/dev.js:91:22)
    at App (/Users/eysi/code/garden-io/garden/node_modules/ink/build/components/App.js:42:9)
```

core/src/cli/autocomplete.ts
edvald: For now this doesn't go over the wire, but yes, we can make a much more trimmed down requirement later.

core/src/cli/command-line.ts
edvald: I get your point but we'd also need to add that space, and I'd want more testing to make sure there aren't unexpected effects. Saving that for later.

core/src/cli/command-line.ts
edvald: Yeah I agree, let's iterate in a future PR.

core/src/commands/cloud/secrets/secrets-create.ts
edvald: Point taken, but I felt this is important just when looking at the command list when running `help` in the dev CLI.

core/src/commands/dev.tsx
edvald: Good call. Done.

core/src/cli/autocomplete.ts
edvald: I've fixed the reload command not showing up or working if initial load fails. But we should definitely give more guidance to the user. Filing that under "a lot of logging improvements to make".

eysi09 (approved):

PR #3684, opened 2023-02-06
chore(lint): improve eslintrc configuration, fix e2e lint errors

Walther:
hoist the ignore patterns to our root eslintrc and make the lint step green again 💚 
(this also fixes the pre-push hook step locally)


Orzelius (approved):
🍀

PR #3686, opened 2023-02-06
0.13 some fixes for `exec` and `container` plugins + some new tests for `exec` plugin

vvagaytsev:
Continuation of #3647.

Partially addresses #3549


core/test/unit/src/plugins/exec/exec.ts
```diff
@@ -1060,26 +1063,29 @@ describe("exec plugin", () => {
...
+                        source: sourcePath,
+                        target: targetPath,
+                      },
+                    ],
+                  },
+                ],
+              },
               spec: {
+                // exec-plugin specific build config defined in the spec
                 build: {
```
edvald, line 1100: Hmm, this doesn't look right. Should just be `spec.command` directly, not `spec.build.command`.

vvagaytsev: @edvald the `spec` field has type [`ExecModuleSpec`](https://github.com/garden-io/garden/blob/0.13/core/src/plugins/exec/moduleConfig.ts) here. How should it look like? Should we replace its `build: ExecBuildSpec` just with `command: string[]` and completely drop the [`ExecBuildSpec`](https://github.com/garden-io/garden/blob/0.13/core/src/plugins/exec/moduleConfig.ts#L225) type definition?

edvald: The confusion here seems to largely stem from the use of "spec" in the context of modules being different from the `spec` field specs in actions. For actions, the exec Build `spec` field should have `ExecBuildActionSpec`  as its interface.

There was no `spec` field on modules, and the term _Spec_ was basically meant to describe the input fields before a resource was parsed and made into a _Config_.

It would make sense to rename `ExecBuildSpec` to e.g. `ExecModuleBuildSpec`, and to make sure that is not accidentally used somewhere for exec Build _actions_.

vvagaytsev: I think the definition of `ExecModuleConfig` is wrong. This PR is a bit clumsy, so I'll close it and will do the changes in smaller portions. Please take a look at #3718

vvagaytsev:
This will be re-done in newer PRs. The first superseding one in #3718.

core/test/unit/src/plugins/exec/exec.ts
vvagaytsev: >There was no spec field on modules, and the term Spec was basically meant to describe the input fields before a resource was parsed and made into a Config.

Actually, there was, see [`spec: M`](https://github.com/garden-io/garden/blob/main/core/src/config/module.ts#L273) in [`ModuleConfig` interface](https://github.com/garden-io/garden/blob/main/core/src/config/module.ts#L257).

In the code above, we create a module **config** with the shape of `ExecModuleConfig`, which, in turn, contains `spec: ExecModuleSpec`. The other tests also use `moduleConfig.spec.build` references. Let's clarify th

I removed some refactoring in the [new PR](https://github.com/garden-io/garden/pull/3718) to make it easier to review.

thsig: Haha yeah, this is all rather confusing until you fully load the context.

vvagaytsev: We can also do some interface renamings that were done here. But let's do it in a dedicated PR, to avoid confusion :) Now #3718 is ready for review :)

PR #3687, opened 2023-02-06
Fix local mode prjc

Orzelius:
https://github.com/garden-io/garden/blob/fix-local-mode-prjc/core/src/plugins/container/config.ts#L224 <- this little thing was added at one point hehe

Incidentally this seems to fix a large portion of k8s integ tests :)


Walther (approved):
good catch! 🎣

vvagaytsev (approved):
Brilliant! :) 👍

PR #3689, opened 2023-02-06
test: exec module conversion spec (module configs with no build)

vvagaytsev:
_Update from @edvald: From this point, all unit/functional tests (i.e. the `test-framework` step) should pass._

🎉


core/test/unit/src/plugins/exec/exec.ts
```diff
@@ -1248,12 +1248,84 @@ describe("exec plugin", () => {
...
+          expect(groupB).to.exist
+
+          const buildB = findActionConfigInGroup(groupB, "Build", moduleNameB)! as BuildActionConfig
+          expect(buildB).to.exist
+          expect(buildB.dependencies).to.eql([{ kind: "Build", name: moduleNameA }])
+
+          const deployB = findActionConfigInGroup(groupB, "Deploy", serviceNameB)! as DeployActionConfig
+          expect(deployB).to.exist
+          expect(deployB.build).to.eql(moduleNameB)
+          expect(deployB.dependencies).to.eql([])
```
vvagaytsev, line 1328: @edvald are these assertions correct? Or should the `deployB.build` field be empty? The spec name is `correctly maps a serviceConfig to a Deploy with no build`, so I wanna confirm what it stands for :)

vvagaytsev: cc @thsig

edvald: The idea with the test case was to validate a conversion of an exec module with just a `serviceConfig`, no build command or build dependencies. This should only result in a single Deploy action, with no Build action.

vvagaytsev: Ah, that was quite simple, thanks! :)

vvagaytsev: Hmm... But the build command is mandatory in the exec module schema, isn't it? Does it make sense to test an incomplete configuration?

We might want to change something in #3686, so let's get back to this later.

core/test/unit/src/plugins/exec/exec.ts
```diff
@@ -1247,12 +1247,61 @@ describe("exec plugin", () => {
...
+                  timeout: 10,
+                },
+              ],
+              tasks: [],
+              tests: [],
+              env: {},
+            },
+          })
+          delete moduleConfigA.spec.build // <--- delete build from the spec to ensure there is no build action
+
```
Walther, line 1282: 🤔  why do we have to create the field first if we immediately delete it below?
If the `makeModuleConfig` function requires the `build` field to be defined, how would an end-user write a no-build module that can be accepted by the function?

vvagaytsev: Because the `build` field is non-optional in the `ExecModuleSpec` interface. Maybe we should make it optional in the interface definition, I'm not sure.

Walther: Yeah - essentially I'm thinking "how would a user get to this situation"? If we want to support users writing exec module configs with no build spec, what would their configuration look like?

Walther: Ah I noticed you've had the same conversation with @edvald above in the other thread as well, my bad!

vvagaytsev: The config might allow the empty values and use empty ones by default. But I need to double-check this. Here we construct the module config objects manually to prepare the test setup. Thanks for raising this question! I'll take a closer look later. Maybe @edvald will share some thoughts on this too.

core/test/unit/src/plugins/exec/exec.ts
edvald: The build command defaults to an empty array in the schema. That effectively amounts to a non-action, which should result in no Build action being generated.

core/src/resolve-module.ts
```diff
@@ -392,11 +392,6 @@ export class ModuleResolver {
       config.build.dependencies = prepareBuildDependencies(config.build.dependencies).filter((dep) => dep.name)
     }
 
-    // We need to refilter the build dependencies on the spec in case one or more dependency names resolved to null.
-    if (config.spec.build && config.spec.build.dependencies) {
-      config.spec.build.dependencies = prepareBuildDependencies(config.spec.build.dependencies)
```
edvald, line 397: Why remove this?

vvagaytsev: I haven't found any usage of this in the code. Also, the `dependencies` field was removed from the `build` field of the module's spec in #3718.

edvald: This is still necessary because of e.g. templating and because the format of the dependencies needs to be canonicalized. I'll restore it in a separate commit.

eysi09 (approved):

PR #3690, opened 2023-02-06
improvement(framework): allow users to specify proxy hostname in project config

### LINKED ISSUE #3667: Error occurred while trying to proxy request on Garden v0.12.50.
### Current Behavior

Proxying a network request through a port forward created by Garden has been reported to fail with the following error after we upgraded to Node 18:

```
[HPM] Error occurred while trying to proxy request *** from localhost:XXXX to localhost:XXXX (ECONNREFUSED) 
```

This seems to happen when the client sending the network request uses Node's `http-proxy-middleware` and the target is localhost. Something like:

```javascript
import { createProxyMiddleware } from "http-proxy-middleware"

const proxy = createProxyMiddlewareI({
  target: "localhost:8080" // <--- Target is localhost
})
```

### Expected behavior

I would expect things to work as before, but maybe this is an issue in Node 18. Here's an [SO thread](https://stackoverflow.com/q/72007843) that looks relevant and the suggestion is to use `127.0.0.1` instead of `localhost`. 

So perhaps this is something that needs be handled by the user in the `createProxyMiddleware` function as opposed to by Garden.

### Workaround

First, set the `GARDEN_PROXY_DEFAULT_ADDRESS` environment variable to 127.0.0.1. E.g.: 

```
GARDEN_PROXY_DEFAULT_ADDRESS=127.0.0.1 garden deploy --dev
```

Then make sure to use `127.0.0.1` as opposed to `localhost` in the `createProxyMiddleware` function as suggested on [StackOveflow](https://stackoverflow.com/q/72007843).

### Your environment

* **OS:** macOS

`garden version` 0.12.50.

eysi09:

After upgrading to Node 18, some users reported issues with port forwarding to "localhost" when using Node's `createProxyMiddleware` package.

This appears to be mostly the [expected behaviour in Node 17+](https://github.com/garden-io/garden/issues/3667#issuecomment-1415665510) and previously the only workaround was to set the proxy URL via the `GARDEN_DEFAULT_PROXY_ADDRESS` env var.

This PR adds a `proxy` field to the project config so that user can explicitly set the proxy address for the project without having to ask everyone to update their environment variables:

```yaml
kind: Project
name: <my-project>
proxy:
  hostname: 127.0.0.1
```

Fixes #3667

### DIFF
--- a/core/src/config/project.ts
+++ b/core/src/config/project.ts
@@ -171,6 +171,10 @@ export interface OutputSpec {
   value: Primitive
 }
 
+export interface ProxyConfig {
+  hostname: string
+}
+
 export interface ProjectConfig {
   apiVersion: string
   kind: "Project"
@@ -179,6 +183,7 @@ export interface ProjectConfig {
   id?: string
   domain?: string
   configPath?: string
+  proxy?: ProxyConfig
   defaultEnvironment: string
   dotIgnoreFiles: string[]
   environments: EnvironmentConfig[]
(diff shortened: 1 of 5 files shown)


edvald (approved):

core/src/config/project.ts
```diff
@@ -319,6 +324,19 @@ export const projectDocsSchema = () =>
       `
         )
         .example([".gardenignore", ".gitignore"]),
+      proxy: joi.object().keys({
+        hostname: joi
```
stefreak, line 328: I think this could easily be confused with a proxy setting for garden itself; `port-forwards: listen-address` would be less confusing to me

PR #3699, opened 2023-02-07
0.13 implement template string tests

### LINKED ISSUE #3698: Implement new tests in `template-string.ts`

vvagaytsev:

Partially addresses #3698

### DIFF
--- a/core/test/unit/src/template-string.ts
+++ b/core/test/unit/src/template-string.ts
@@ -1712,7 +1712,35 @@ describe("collectTemplateReferences", () => {
 describe("getActionTemplateReferences", () => {
   context("action.*", () => {
     it("returns valid action references", () => {
-      throw "TODO"
+      const config = {
+        build: '${action["build"].build-a}',
+        deploy: '${action["deploy"].deploy-a}',
+        run: '${action["run"].run-a}',
+        test: '${action["test"].test-a}',
+      }
+      const actionTemplateReferences = getActionTemplateReferences(config)
+      expect(actionTemplateReferences).to.eql([
+        {
+          kind: "Build",
+          name: "build-a",
+          fullRef: ["action", "build", "build-a"],
+        },
+        {
+          kind: "Deploy",
+          name: "deploy-a",
+          fullRef: ["action", "deploy", "deploy-a"],
+        },
+        {
+          kind: "Run",
+          name: "run-a",
+          fullRef: ["action", "run", "run-a"],
+        },
+        {
+          kind: "Test",
+          name: "test-a",
+          fullRef: ["action", "test", "test-a"],
+        },
+      ])
     })
 
     it("throws if action ref has no kind", () => {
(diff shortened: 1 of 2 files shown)


core/test/unit/src/template-string.ts
```diff
@@ -1710,58 +1710,176 @@ describe("collectTemplateReferences", () => {
...
+      expectError(() => getActionTemplateReferences(config), {
+        contains: "Found invalid action reference (name is not a string)",
+      })
+    })
 
-  it("throws if runtime ref has no name", () => {
-    throw "TODO"
+    it("throws if action name is not resolvable", () => {
+      throw "TODO"
+    })
```
vvagaytsev, line 1802: This has been left as `throws "TODO"`. There is no validation for non-resolvable names yet. Every name is just considered as a string.

@edvald is this related to the [`TODO-G2: Allow such nested references in certain cases, e.g. if resolvable with a ProjectConfigContext.` comment](https://github.com/garden-io/garden/blob/0.13/core/src/template-string/template-string.ts#L429) in [template-string.ts](https://github.com/garden-io/garden/blob/0.13/core/src/template-string/template-string.ts)?

core/test/unit/src/template-string.ts
```diff
@@ -1710,58 +1710,176 @@ describe("collectTemplateReferences", () => {
...
+        foo: '${runtime["tasks"].123}',
+      }
+      expectError(() => getActionTemplateReferences(config), {
+        contains: "Found invalid runtime reference (name is not a string)",
+      })
+    })
+
+    it("throws if runtime ref name is not resolvable", () => {
+      throw "TODO"
+    })
```
vvagaytsev, line 1882: The same question as [above](https://github.com/garden-io/garden/pull/3699/files#r1098647747).

core/test/unit/src/template-string.ts
```diff
@@ -1718,42 +1718,85 @@ describe("getActionTemplateReferences", () => {
...
+      foo: "${action[123]}",
+    }
+    expectError(() => getActionTemplateReferences(config), {
+      contains: "Found invalid action reference (kind is not a string)",
+    })
+  })
+
+  it("throws if action ref has no name", () => {
+    const config = {
+      foo: "${action[\"build\"]}",
```
Orzelius: maybe just using ` would be easier to parse here or does the lint want it this way?
like this:
```
foo: `${action["build"]}`
```

Orzelius: I think `'` would also work the same

Orzelius (approved):
Thank you!

core/test/unit/src/template-string.ts
vvagaytsev: Leaving this discussion open and merging the PR. The remained TODO-tests will be implemented in a separate PR.

edvald: I think this can be removed

core/test/unit/src/template-string.ts
edvald: Same as above, just remove

PR #3701, opened 2023-02-07
chore(test): fix commands/publish.ts

Walther:

Fixes https://github.com/garden-io/garden/issues/3695

### DIFF
--- a/core/test/unit/src/commands/publish.ts
+++ b/core/test/unit/src/commands/publish.ts
@@ -155,7 +155,7 @@ describe("PublishCommand", () => {
       },
       opts: withDefaultGlobalOpts({
         "force-build": false,
-        "tag": tag,
+        tag,
       }),
     })
 
@@ -170,7 +170,7 @@ describe("PublishCommand", () => {
   it("should resolve a templated tag and apply to the builds", async () => {
     const garden = await getTestGarden()
     const log = garden.log
-    const tag = "v1.0-${module.name}-${module.version}"
+    const tag = "v1.0-${module.name}"
 
     const { result } = await command.action({
       garden,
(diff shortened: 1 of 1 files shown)


core/test/unit/src/commands/publish.ts
```diff
@@ -170,7 +170,7 @@ describe("PublishCommand", () => {
   it("should resolve a templated tag and apply to the builds", async () => {
     const garden = await getTestGarden()
     const log = garden.log
-    const tag = "v1.0-${module.name}-${module.version}"
+    const tag = "v1.0-${module.name}"
```
Walther, line 173: Do you think this is sufficient for testing the templating, or should we figure out some more complex logic using some other variables and getters?

edvald: Not quite sure what you mean by variables and getters, but I do think it's important to actually test that templating with a version works as expected.

Walther: In this PR the `v1.0-` part is used in the templating and verified in the test, so at least that part of templating is working as expected.

Previously we also included `module.version` in the template and tested for that, but as discussed on Slack, the `action.versionString()` call will result in a hash that differs from the `module.version` resolved in the template. If we want to test this part too, what would be the correct place to look for the hash?

edvald: The template context is defined here: https://github.com/garden-io/garden/blob/d514b82446cdca1f9de5b0b7376c49bf38f7b6c3/core/src/tasks/publish.ts#L138

In principle the test case _should_ work with the `${module.version}`/`${build.version}` tag, so I'm not quite sure why that's not working right.

Walther:
Closing this PR and delegating https://github.com/garden-io/garden/issues/3695 to @edvald as discussed on Slack.

PR #3702, opened 2023-02-08
Remove watch mode flags

edvald:
_Reviewer note: I did not throw out all of the underlying watching mechanism, because there are plans to use that for e.g. watching configuration files. Also, some discussion is expected here, don't hesitate to chime in!_

BREAKING CHANGE:

The legacy file watching mechanism used to automatically re-build, re-deploy etc. is no longer used in a meaningful manner, and has been known to cause more problems than it's worth.

The -w/--watch flags have now been completely removed, in favor of the more useful and usable _dev mode_ for syncing code to running deployments.

Should there be user-demand, we may later add a separate `watch` command, and support that in the new `garden dev` command.


Orzelius (approved):
There was a large amount of I presume accidental empty lines added instead of the watch options in several files, but that's a pain to fix so I guess let 'em be. Also, good thing we had been putting of fixing the watch tests this entire time.

vvagaytsev (approved):
LGTM! 👍 By the way, now there are some conflicts. Let's resolve 'em and merge this.

edvald:
> There was a large amount of I presume accidental empty lines added instead of the watch options in several files, but that's a pain to fix so I guess let 'em be.

Yeah, it's an artifact of a good ol' find+replace, figured it's fine and we fix it gradually when poking at those files later.

PR #3703, opened 2023-02-08
refactor(k8s): change default kaniko namespace to the project namespace

### LINKED ISSUE #3508: Change default Kaniko namespace to project namespace
See 0.13 TODO comment in https://github.com/garden-io/garden/blob/0.13/core/src/plugins/kubernetes/container/build/kaniko.ts

edvald:
Closes #3508

BREAKING CHANGE:

The default namespace for kaniko build Pods is now the project namespace. Set `kaniko.namespace` in the provider configuration to override to a specific, separate namespace.

### DIFF
--- a/core/src/plugins/kubernetes/config.ts
+++ b/core/src/plugins/kubernetes/config.ts
@@ -540,12 +540,9 @@ export const kubernetesConfigBase = () =>
         namespace: joi
           .string()
           .allow(null)
-          .default(defaultSystemNamespace)
           .description(
             dedent`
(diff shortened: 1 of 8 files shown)


stefreak (approved):
I think I'd not consider this a breaking change per se, because the building would continue to work (just in a different namespace).

But it's also fine to treat it as a breaking change, maybe I'm missing something 👍

edvald:
Thanks! I'm flagging this as breaking because it's quite possible that cluster configuration, policies etc. may need to be updated.

Issue #3499, opened 2023-01-09
[FEATURE]: make log levels easier to use (`-l3` => `-v`, `-l5` => `-vvv`)


stefreak:
## Feature Request
Recently I wanted to see the docker build logs when running `garden build`. First thing I tried is running `garden build -v` (muscle memory from so many other commands I've used in the past).

But -v just prints the version. I had to run `garden build -l3` to get the expected result.

### Background / Motivation
Using garden should be intuitive.

### What should the user be able to do?
I want the `-V` option to print the version, or even just `--version`.

`-v` should be equivalent to `-l3`, `-vv` to `-l4` and `-vvv` to `-l5`.

### Why do they want to do this? What problem does it solve?
This makes choosing the log levels much more intuitive and more similar to other dev tools out there.

### Suggested Implementation(s)
I'm not familiar with the cli argument parsing code but this should be straight forward

### How important is this feature for you/your team?
:rose: It’s a nice to have, but nice things are nice :slightly_smiling_face:

PR #3704, opened 2023-02-08
refactor(cli): make -V instead of -v (lowercase) print Garden version

edvald:
BREAKING CHANGE:

The -v CLI flag (alias for --version) has been replaced by -V (upper-case). This is to make way for #3499.


stefreak (approved):
cool! 👍

PR #3705, opened 2023-02-08
fix(core): error when attempting to emit warning message

### LINKED ISSUE #3688: Fix `HideWarningCommand`
See [`emitWarning`](https://github.com/garden-io/garden/blob/0.13/core/src/warnings.ts#L13) and [`hideWarning`](https://github.com/garden-io/garden/blob/0.13/core/src/warnings.ts#L19) functions.

Those were broken in #3619. Need to preserve the semantics and use file-backed storage instead of DB.

edvald:
Introduced in #3619. Closes #3688.

### DIFF
--- a/cli/src/add-version-files.ts
+++ b/cli/src/add-version-files.ts
@@ -34,7 +34,13 @@ async function addVersionFiles() {
     const path = config.path
     const versionFilePath = resolve(path, GARDEN_VERSIONFILE_NAME)
 
-    const vcsHandler = new GitHandler(STATIC_DIR, garden.gardenDirPath, garden.dotIgnoreFile, new TreeCache())
+    const vcsHandler = new GitHandler({
+      garden,
+      projectRoot: STATIC_DIR,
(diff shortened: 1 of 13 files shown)


Orzelius (approved):
There was a large amount of I presume accidental empty lines added instead of the watch options in several files, but that's a pain to fix so I guess let 'em be. Also, good thing we had been putting of fixing the watch tests this entire time.

Orzelius (approved):
So the bug was that the non-garden class `emitWarning` did not work?

vvagaytsev (approved):

vvagaytsev:
@Orzelius the bug was that the functions `emitWarning` and `hideWarning` defined in the file [warnings.ts](https://github.com/garden-io/garden/blob/8c2fbb7b5d0c5ebdece95cdc3a10a2a62135c918/core/src/warnings.ts) referred the missing DB entity `Warning` via dynamic imports like `const { Warning } = require("./db/entities/warning")`. This caused runtime errors.

Now both functions use file-backed config storage.

Issue #3530, opened 2023-01-17
0.13 fix tests in garden.ts


Walther:
- [x] Some trivial assertion fixes in https://github.com/garden-io/garden/pull/3532
- [x] #3709 
- [x] `resolveModules` errors -> fixed in #3720 
- [x] `getConfigGraph` errors -> fixed in #3710
- [x] `when a provider has an augmentGraph handler` context errors
- [x] `resolveModuleVersion` errors
- [x] fix some schema validation issues
- [x] `warnings` errors -> fixed in #3722

PR #3709, opened 2023-02-08
0.13 fix `resolveProviders` spec in garden unit tests

vvagaytsev:
This fixes only one failing test, but brings some changes to the existing type system to ensure that some custom [mapped types](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html) can also return optional values, i.e. the ones that can be `undefined`.

For example, let's consider the type
```
export type ActionTypeMap<T> = {
  [K in ActionKind]: {
    [type: string]: T
  }
}
```

When an object of the type above is referenced as `typeMap[kind][type]` it still can return an `undefined` value. The first-level sub-map can to forced to never return `undefined` by explicit initialization of `typeMap` for each `kind` like this: `typeMap[kind] = {}`. But `type` is an arbitrary string here.

If there is a better way or TS best practice to handle `undefined` values in the mapped types, I'm all ears.

Maybe, it's better to do some fail-fast validation instead of allowing undefined action-type handlers.

Partially addresses #3530

### DIFF
--- a/core/src/router/base.ts
+++ b/core/src/router/base.ts
@@ -34,7 +34,7 @@ import {
 import { InternalError, ParameterError, PluginError } from "../exceptions"
 import { validateSchema } from "../config/validation"
 import { getActionTypeBases, getPluginBases, getPluginDependencies } from "../plugins"
-import { getNames } from "../util/util"
+import { getNames, MaybeUndefined } from "../util/util"
 import { defaultProvider } from "../config/provider"
 import type { ConfigGraph } from "../graph/config-graph"
 import { ActionConfigContext, ActionSpecContext } from "../config/template-contexts/actions"
@@ -179,6 +179,7 @@ export function createActionRouter<K extends ActionKind>(
   handlers: Omit<ActionRouterHandlers<K>, CommonHandlers>
 ): ActionKindRouter<K> {
   class Router extends BaseActionRouter<K> {}
+
   const router = new Router(kind, baseParams)
 
   const wrapped = mapValues(handlers, (h, key) => {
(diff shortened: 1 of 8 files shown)


core/src/router/base.ts
```diff
@@ -316,9 +317,9 @@ export abstract class BaseActionRouter<K extends ActionKind> extends BaseRouter
 
   async validateActionOutputs<T extends BaseAction>(action: T, type: "static" | "runtime", outputs: any) {
     const actionTypes = await this.garden.getActionTypes()
-    const spec: ActionTypeDefinition<any> = actionTypes[action.kind][action.type]
+    const spec = actionTypes[action.kind][action.type]
 
-    const schema = type === "static" ? spec.staticOutputsSchema : spec.runtimeOutputsSchema
+    const schema = type === "static" ? spec?.staticOutputsSchema : spec?.runtimeOutputsSchema
```
vvagaytsev, line 322: Alternatively, we can check if the spec is defined and throw an error:
```
if (!spec) {
      throw new ConfigurationError(
        `Cannot find handler "${String(
          handlerType
        )}" for action type "${actionType}". Available handlers: [${Object.keys(this.handlers[handlerType]).join(
          ","
        )}]`,
        {
          handlerType,
          actionType,
        }
      )
    }
```

But I'm not sure if that would be expected behavior.

core/src/router/base.ts
```diff
@@ -401,7 +402,7 @@ export abstract class BaseActionRouter<K extends ActionKind> extends BaseRouter
     )
     const spec = this.definitions[actionType]
 
-    if (handlers.length === 0 && spec.base && !pluginName) {
+    if (handlers.length === 0 && spec?.base && !pluginName) {
```
vvagaytsev, line 405: Alternatively, we can check if the spec is defined and throw an error:
```
if (!spec) {
      throw new ConfigurationError(
        `Cannot find handler "${String(
          handlerType
        )}" for action type "${actionType}". Available handlers: [${Object.keys(this.handlers[handlerType]).join(
          ","
        )}]`,
        {
          handlerType,
          actionType,
        }
      )
    }
```

But I'm not sure if that would be expected behavior.

edvald: I think it makes sense to throw here if the spec isn't found

vvagaytsev: Alright, thanks! I'll change this. Should we throw an error in the rest places as well?

vvagaytsev: Ah, the error is thrown later in [L482](https://github.com/garden-io/garden/blob/d63595c3ad9f2ff1065907d9b281b9ac892d4fca/core/src/router/base.ts#L482) is there is no default handler provided. So, maybe it's ok to leave it as it is now, wdyt @edvald ?

edvald (approved):

PR #3722, opened 2023-02-09
test: fix test context init for warning tests

vvagaytsev:
Using the same constant key caused some dirty state leaking across the tests.

Fixes #


Orzelius (approved):
nice catch

Issue #3723, opened 2023-02-09
0.13 test fixes: util/ext-source-util.ts 


Walther:
see commands/link.ts

PR #3724, opened 2023-02-09
test: fix util/ext-source-util.ts assertions

Walther:
Fixes https://github.com/garden-io/garden/issues/3723

I made the distinction between `sourcesObj` and `sourcesList` more explicit, as some of the functions expect a plain list and others expect the object with name-based keys.

This makes the tests on this file pass, however, let me know if this is actually the _intended_ behavior or whether we need to change the underlying code to unify the formats.


Orzelius (approved):
🌻

vvagaytsev (approved):
Looks good! 👍

Issue #3727, opened 2023-02-09
0.13 test fixes: cloud/api.ts


Walther:
-> might be relevant to #3619

PR #3728, opened 2023-02-09
test: fix cloud/api.ts

Walther:
Fixes https://github.com/garden-io/garden/issues/3727

The removed test saved an auth token, which breaks the test right below it as it expects a token to not exist yet.
This test was a duplicate anyway: saving and retrieving a token gets tested one test below as well.


vvagaytsev (approved):
Nice! :+1:

PR #3730, opened 2023-02-09
fix(k8s): more stable & performant log streaming

### LINKED ISSUE #3586: Unexpected upload network traffic while tailing K8s logs
### Current Behavior

While running tasks or building new Docker images, upload bandwidth spikes to 1Mbps (the max for my slow internet connection) and it stays maxed out for the duration of the task execution (and usually for up to a minute after the task container has "completed" as seen via `k9s`). The Garden CLI does not report the task has finished when it actually finishes (as noted, it will sometimes succeed or even report a failure a minute after the task has finished successfully in the remote container) because it is apparently bogged down with the network requests it is making. This is not traffic related to sending build context to the cluster, that traffic actually stays closer to 300KB/s upload. As a side effect of bandwidth being maxed out, I also see failed connections frequently which results in Garden reporting a failure to me even though things were working fine in the cluster (the problem was merely communication _with_ the cluster).

I was able to confirm that for some reason this upload traffic is entirely related to tailing logs by commenting out the following code in a local build of Garden:
```diff
diff --git a/core/src/plugins/kubernetes/logs.ts b/core/src/plugins/kubernetes/logs.ts
index 34e235e17..a32ed241b 100644
--- a/core/src/plugins/kubernetes/logs.ts
+++ b/core/src/plugins/kubernetes/logs.ts
@@ -230,9 +230,9 @@ export class K8sLogFollower<T> {
   public async followLogs(opts: LogOpts) {
     await this.createConnections(opts)
 
-    this.intervalId = setInterval(async () => {
-      await this.createConnections(opts)
-    }, this.retryIntervalMs)
+//    this.intervalId = setInterval(async () => {
+//      await this.createConnections(opts)
+//    }, this.retryIntervalMs)
 
     return new Promise((resolve, _reject) => {
       this.resolve = resolve
```

When the above code has been commented out, upload bandwidth stays at almost 0 during task execution rather than constantly maxed out at 1Mbps. The problem of failed connections also goes away because network contention is no longer a problem.

### Expected behavior

I expect the Garden process of running a task that is entirely contained in the cluster (no transmission of data for the operation of the task once the Docker image has been built) to result in very little upload network bandwidth usage.

### Workaround

Build a custom version of Garden with log tailing commented out. This is what I am working with currently because the experience is so drastically better.

### Your environment

* **OS:** macOS 13.0.1
* **How I'm running Kubernetes:** Azure

`garden version`: 0.12.48

thsig:

Fixes an issue with very high upload bandwidth use when running Kubernetes based tests/tasks that produce a lot of log output.

co-written by @stefreak and @thsig

* Improved connection management and pod lifecycle logic, including more robust connection timeout enforcement.

* Removed keepalive logic, since it doesn't work on all operating systems.

* Improved deduplication logic to generate fewer false positives (and eliminate false-negatives).

* Use `sinceTime` when fetching logs on retry to make sure we don't fetch any unnecessary logs.

* When a runner pod terminates, we make sure to wait until the final logs have been fetched.

* Default to using the `tail` option in conjunction with a "max log lines in memory" setting instead of `limitBytes` to avoid clipping / incomplete log lines while also avoiding the loading of too much log data into memory.

* Only start one connection attempt at a time, to prevent multiple connections to the same container at once.

* Make sure that we only call `createConnections` once it has finished, so that there is only one concurrent instance of the method running per `LogFollower` at a time.

Fixes #3586.

**Notes to reviewer**:

I will use "squash and merge" when merging this PR.

One convenient way to test the reconnect logic in this PR on a macOS machine is by using the Network Link Conditioner: https://nshipster.com/network-link-conditioner/

**Deduplication fix example**:

With a container module + a little demo go program like this:

```go
	for i := 0; i < 10; i++ {
		for x := 0; x < i; x++ {
			fmt.Printf("You should see this line %d times\n", i)
		}
		time.Sleep(1 * time.Second)
	}
```

Difference for `garden test deduplication -l3 -f` before and after the change:
```diff
--- before.log  2023-02-24 15:28:00
+++ after.log   2023-02-24 15:28:18
@@ -1,4 +1,4 @@
-$ garden test deduplication -l3 -f
+$ gdev test deduplication -l3 -f
 Running tests 🌡️ 
 
 
@@ -16,17 +16,52 @@
 ℹ deduplication             → Getting build status for v-a5a55d7920...
 ✔ deduplication             → Getting build status for v-a5a55d7920... → Already built
 ℹ deduplication             → Running test tests
-ℹ deduplication [verbose]   → Starting Pod test-deduplication-test-29ab47 with command 'sh -c ./main 10'
+ℹ deduplication [verbose]   → Starting Pod test-deduplication-test-054b2c with command 'sh -c ./main 10'
 ℹ deduplication [verbose]   → [sh]: You should see this line 1 times
 ℹ deduplication [verbose]   → [sh]: You should see this line 2 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 2 times
 ℹ deduplication [verbose]   → [sh]: You should see this line 3 times
 ℹ deduplication [verbose]   → [sh]: You should see this line 3 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 3 times
 ℹ deduplication [verbose]   → [sh]: You should see this line 4 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 4 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 4 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 4 times
 ℹ deduplication [verbose]   → [sh]: You should see this line 5 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 5 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 5 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 5 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 5 times
 ℹ deduplication [verbose]   → [sh]: You should see this line 6 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 6 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 6 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 6 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 6 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 6 times
 ℹ deduplication [verbose]   → [sh]: You should see this line 7 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 7 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 7 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 7 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 7 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 7 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 7 times
 ℹ deduplication [verbose]   → [sh]: You should see this line 8 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 8 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 8 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 8 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 8 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 8 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 8 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 8 times
 ℹ deduplication [verbose]   → [sh]: You should see this line 9 times
-✔ deduplication             → Running test tests → Success (took 13.3 sec)
+ℹ deduplication [verbose]   → [sh]: You should see this line 9 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 9 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 9 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 9 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 9 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 9 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 9 times
+ℹ deduplication [verbose]   → [sh]: You should see this line 9 times
+✔ deduplication             → Running test tests → Success (took 14.8 sec)
 
 Done! ✔️ 
```

### DIFF
--- a/core/src/plugins/kubernetes/logs.ts
+++ b/core/src/plugins/kubernetes/logs.ts
@@ -11,7 +11,7 @@ import parseDuration from "parse-duration"
 
 import { ServiceLogEntry } from "../../types/plugin/service/getServiceLogs"
 import { KubernetesResource, KubernetesPod, BaseResource } from "./types"
-import { getAllPods } from "./util"
+import { getAllPods, summarize } from "./util"
 import { KubeApi, KubernetesError } from "./api"
 import { GardenService } from "../../types/service"
 import Stream from "ts-stream"
@@ -310,6 +310,13 @@ export class K8sLogFollower<T extends LogEntryBase> {
       this.log.silly(`<Lost connection to ${description}. Reason: ${reason}>`)
     }
 
+    /**
+     * Helper to stop retrying a connection.
+     *
+     * This means that we won't fetch logs again from this container, but createConnections will still
+     * be called and thus we will still notice when new Pods are added to the Deployment, for example when
+     * the user runs `garden deploy`.
+     */
     const stopRetrying = (why: string) => {
       this.log.silly(`<Will stop retrying connecting to ${description}. Reason: ${why}>`)
 
@@ -354,9 +361,6 @@ export class K8sLogFollower<T extends LogEntryBase> {
     })
 
     if (containers.length === 0) {
-      function summarize(resources: KubernetesResource[]) {
-        return resources.map((r) => `${r.kind} ${r.metadata.name}`).join(", ")
-      }
       this.log.debug(
         `<No running containers found for ${summarize(this.resources)}. Will retry in ${
           this.retryIntervalMs / 1000
--- a/core/src/plugins/kubernetes/logs.ts
+++ b/core/src/plugins/kubernetes/logs.ts
@@ -25,6 +25,7 @@ import { splitFirst, isValidDateInstance, sleep } from "../../util/util"
 import { Writable } from "stream"
 import request from "request"
 import { LogLevel } from "../../logger/logger"
+import { clearTimeout } from "timers"
 
 // When not following logs, the entire log is read into memory and sorted.
 // We therefore set a maximum on the number of lines we fetch.
@@ -123,6 +124,7 @@ async function readLogs<T extends LogEntryBase>({
 }
 
 type ConnectionStatus = "connecting" | "connected" | "timed-out" | "error" | "closed"
+const disconnectedStatuses: ConnectionStatus[] = ["timed-out", "error", "closed"]
 
 interface LastLogEntries {
   messages: string[]
@@ -163,7 +165,7 @@ export class K8sLogFollower<T extends LogEntryBase> {
   private log: LogEntry
   private defaultNamespace: string
   private resources: KubernetesResource<BaseResource>[]
-  private intervalId: NodeJS.Timer | null
+  private timeoutId: NodeJS.Timer | null
   private resolve: ((val: unknown) => void) | null
   private retryIntervalMs: number
 
@@ -191,7 +193,7 @@ export class K8sLogFollower<T extends LogEntryBase> {
     this.log = log
     this.defaultNamespace = defaultNamespace
     this.resources = resources
-    this.intervalId = null
+    this.timeoutId = null
     this.resolve = null
     this.retryIntervalMs = retryIntervalMs
   }
@@ -201,11 +203,16 @@ export class K8sLogFollower<T extends LogEntryBase> {
    * until outside code calls the close method.
    */
   public async followLogs(opts: LogOpts) {
-    await this.createConnections(opts)
+    // make sure that createConnections is never called concurrently (wait for it to finish, then wait retryIntervalMs)
+    const followLoop = async () => {
+      try {
+        await this.createConnections(opts)
+      } finally {
+        this.timeoutId = setTimeout(followLoop, this.retryIntervalMs)
+      }
+    }
 
-    this.intervalId = setInterval(async () => {
-      await this.createConnections(opts)
-    }, this.retryIntervalMs)
+    await followLoop()
 
     return new Promise((resolve, _reject) => {
       this.resolve = resolve
@@ -218,7 +225,9 @@ export class K8sLogFollower<T extends LogEntryBase> {
    */
   public close() {
     this.clearConnections()
-    this.resolve && this.resolve({})
+    if (this.resolve) {
+      this.resolve({})
+    }
   }
 
   /**
@@ -232,9 +241,9 @@ export class K8sLogFollower<T extends LogEntryBase> {
 
   private clearConnections() {
     const conns = Object.values(this.connections)
-    if (this.intervalId) {
-      clearInterval(this.intervalId)
-      this.intervalId = null
+    if (this.timeoutId) {
+      clearTimeout(this.timeoutId)
+      this.timeoutId = null
     }
     conns.forEach((conn) => {
       try {
@@ -338,15 +347,16 @@ export class K8sLogFollower<T extends LogEntryBase> {
     await Bluebird.map(containers, async ({ pod, containerName }) => {
       const connection = this.createConnectionIfMissing(pod, containerName)
 
-      if (connection && (connection.shouldRetry === false || connection.status === "connected")) {
-        // Nothing to do
+      if (disconnectedStatuses.includes(connection.status) && connection.shouldRetry) {
+        // The connection has been registered but is not active
+        this.log.silly(
+          `<Connecting to container ${connection.containerName} in Pod ${connection.pod.metadata.name}, because current connection status is ${connection.status}>`
+        )
+        connection.status = "connecting"
+      } else {
+        // nothing to do
         return
       }
-
-      // The connection has been registered but is not active
-      this.log.silly(
-        `<Not connected to container ${connection.containerName} in Pod ${connection.pod.metadata.name}. Connection status is ${connection.status}>`
-      )
 
       let req: request.Request
 
@@ -411,13 +421,7 @@ export class K8sLogFollower<T extends LogEntryBase> {
         })
         this.log.silly(`<Connected to container '${containerName}' in Pod '${pod.metadata.name}'>`)
       } catch (err) {
-        // Log the error and keep trying.
-        // If the error is "HTTP request failed" most likely the pod is just not up yet
-        if (err.message !== "HTTP request failed") {
-          this.log.debug(
-            `<Getting logs for container '${containerName}' in Pod '${pod.metadata.name}' failed with error: ${err?.message}>`
-          )
-        }
+        await this.handleConnectionClose(connection, "error", err.message)
         return
       }
       connection.request = req
@@ -440,8 +444,6 @@ export class K8sLogFollower<T extends LogEntryBase> {
     tail: number
     sinceSeconds?: string
   }) {
-    const logger = this.k8sApi.getLogger()
-
     const opts = {
       follow: true, // only works with follow true, as we receive chunks with multiple messages in the stream otherwise
       pretty: false,
@@ -463,7 +465,9 @@ export class K8sLogFollower<T extends LogEntryBase> {
       opts["sinceSeconds"] = parseDuration(sinceSeconds, "s") || undefined
     }
 
-    return logger.log(connection.namespace, connection.pod.metadata.name, connection.containerName, stream, opts)
+    return this.k8sApi
+      .getLogger()
+      .log(connection.namespace, connection.pod.metadata.name, connection.containerName, stream, opts)
   }
 
   private createConnectionIfMissing(pod: KubernetesPod, containerName: string): LogConnection {
@@ -474,7 +478,7 @@ export class K8sLogFollower<T extends LogEntryBase> {
         namespace: pod.metadata.namespace || this.defaultNamespace,
         pod,
         containerName,
-        status: "connecting",
+        status: "closed",
         shouldRetry: true,
       }
     }
--- a/core/src/plugins/kubernetes/logs.ts
+++ b/core/src/plugins/kubernetes/logs.ts
@@ -7,13 +7,12 @@
  */
 
 import { omit, sortBy } from "lodash"
-import moment from "moment"
 import parseDuration from "parse-duration"
 
 import { ServiceLogEntry } from "../../types/plugin/service/getServiceLogs"
 import { KubernetesResource, KubernetesPod, BaseResource } from "./types"
 import { getAllPods } from "./util"
-import { KubeApi } from "./api"
+import { KubeApi, KubernetesError } from "./api"
 import { GardenService } from "../../types/service"
 import Stream from "ts-stream"
 import { LogEntry } from "../../logger/log-entry"
@@ -21,14 +20,15 @@ import Bluebird from "bluebird"
 import { KubernetesProvider } from "./config"
 import { PluginToolSpec } from "../../types/plugin/tools"
 import { PluginContext } from "../../plugin-context"
-import { getPodLogs } from "./status/pod"
-import { splitFirst, isValidDateInstance } from "../../util/util"
+import { checkPodStatus, getPodLogs } from "./status/pod"
+import { splitFirst, isValidDateInstance, sleep } from "../../util/util"
 import { Writable } from "stream"
 import request from "request"
 import { LogLevel } from "../../logger/logger"
 
 // When not following logs, the entire log is read into memory and sorted.
 // We therefore set a maximum on the number of lines we fetch.
+
 const maxLogLinesInMemory = 100000
 
 interface GetAllLogsParams {
@@ -42,6 +42,11 @@ interface GetAllLogsParams {
   tail?: number
   since?: string
   resources: KubernetesResource[]
+}
+
+export interface LogEntryBase {
+  msg: string
+  timestamp?: Date
 }
 
 /**
@@ -58,7 +63,7 @@ export async function streamK8sLogs(params: GetAllLogsParams) {
       logsFollower.close()
     })
 
-    await logsFollower.followLogs({ tail: params.tail, since: params.since, limitBytes: null })
+    await logsFollower.followLogs({ tail: params.tail, since: params.since })
   } else {
     const pods = await getAllPods(api, params.defaultNamespace, params.resources)
     let tail = params.tail
@@ -70,34 +75,32 @@ export async function streamK8sLogs(params: GetAllLogsParams) {
 
       params.log.debug(`Tail parameter not set explicitly. Setting to ${tail} to prevent log overflow.`)
     }
-    await Bluebird.map(pods, (pod) => readLogs({ ...omit(params, "pods"), entryConverter, pod, tail }))
+    const { stream } = params
+    await Bluebird.map(pods, async (pod) => {
+      const serviceLogEntries = await readLogs({ ...omit(params, "pods", "stream"), entryConverter, pod, tail, api })
+      for (const entry of sortBy(serviceLogEntries, "timestamp")) {
+        void stream.write(entry)
+      }
+    })
   }
   return {}
 }
 
-async function readLogs<T>({
-  log,
-  ctx,
-  provider,
-  stream,
+async function readLogs<T extends LogEntryBase>({
+  api,
   entryConverter,
   tail,
   pod,
   defaultNamespace,
   since,
 }: {
-  log: LogEntry
-  ctx: PluginContext
-  provider: KubernetesProvider
-  stream: Stream<T>
+  api: KubeApi
   entryConverter: PodLogEntryConverter<T>
   tail?: number
   pod: KubernetesPod
   defaultNamespace: string
   since?: string
-}) {
-  const api = await KubeApi.factory(log, ctx, provider)
-
+}): Promise<T[]> {
   const logs = await getPodLogs({
     api,
     namespace: pod.metadata?.namespace || defaultNamespace,
@@ -111,64 +114,40 @@ async function readLogs<T>({
     return _log.split("\n").map((line) => {
       line = line.trimEnd()
       const res = { containerName }
-      try {
-        const [timestampStr, msg] = splitFirst(line, " ")
-        const timestamp = moment(timestampStr).toDate()
-        return entryConverter({ ...res, timestamp, msg })
-      } catch {
-        return entryConverter({ ...res, msg: line })
-      }
+      const { timestamp, msg } = parseTimestampAndMessage(line)
+      return entryConverter({ ...res, timestamp, msg })
     })
   })
 
-  for (const line of sortBy(allLines, "timestamp")) {
-    void stream.write(line)
-  }
+  return sortBy(allLines, "timestamp")
 }
 
-type ConnectionStatus = "connected" | "error" | "closed"
+type ConnectionStatus = "connecting" | "connected" | "timed-out" | "error" | "closed"
 
+interface LastLogEntries {
+  messages: string[]
+  timestamp: Date
+}
 interface LogConnection {
   pod: KubernetesPod
   containerName: string
   namespace: string
-  request: request.Request
   status: ConnectionStatus
+  shouldRetry: boolean
+  request?: request.Request
+  timeout?: NodeJS.Timeout
+
+  // for reconnect & deduplication logic
+  lastLogEntries?: LastLogEntries
+  previousConnectionLastLogEntries?: LastLogEntries
 }
 
 interface LogOpts {
   tail?: number
   since?: string
-  /**
-   * If set to null, does not limit the number of bytes. This parameter is made mandatory, so that the usage site
-   * makes a deliberate and informed choice about it.
-   *
-   * From the k8s javascript client library docs:
-   *
-   * "If set, the number of bytes to read from the server before terminating the log output. This may not display a
-   * complete final line of logging, and may return slightly more or slightly less than the specified limit.""
-   */
-  limitBytes: number | null
 }
 
 const defaultRetryIntervalMs = 10000
-
-/**
- * The maximum number of streamed entries to keep around to compare incoming entries against for deduplication
- * purposes.
- *
- * One such buffer is maintained for each container for each resource the `K8sLogFollower` instance
- * is following (and deduplication is performed separately for each followed container).
- *
- * Deduplication is needed e.g. when the connection with a container is lost and reestablished, and recent logs are
- * re-fetched. Some of those log entries may have the same timestamp and message as recently streamed entries,
- * and not re-streaming them if they match an entry in the deduplication buffer is usually the desired behavior
- * (since it prevents duplicate log lines).
- *
- * The deduplication buffer size should be kept relatively small, since a large buffer adds a slight delay before
- * entries are streamed.
- */
-const defaultDeduplicationBufferSize = 500
 
 /**
  * A helper class for following logs and managing the logs connections.
@@ -176,14 +155,12 @@ const defaultDeduplicationBufferSize = 500
  * The class operates kind of like a control loop, fetching the state of all pods for a given service at
  * an interval, comparing the result against current active connections and attempting re-connects as needed.
  */
-export class K8sLogFollower<T> {
+export class K8sLogFollower<T extends LogEntryBase> {
   private connections: { [key: string]: LogConnection }
   private stream: Stream<T>
   private entryConverter: PodLogEntryConverter<T>
   private k8sApi: KubeApi
   private log: LogEntry
-  private deduplicationBufferSize: number
-  private deduplicationBuffers: { [key: string]: { msg: string; time: number }[] }
   private defaultNamespace: string
   private resources: KubernetesResource<BaseResource>[]
   private intervalId: NodeJS.Timer | null
@@ -196,7 +173,6 @@ export class K8sLogFollower<T> {
     defaultNamespace,
     k8sApi,
     log,
-    deduplicationBufferSize = defaultDeduplicationBufferSize,
     resources,
     retryIntervalMs = defaultRetryIntervalMs,
   }: {
@@ -204,7 +180,6 @@ export class K8sLogFollower<T> {
     entryConverter: PodLogEntryConverter<T>
     k8sApi: KubeApi
     log: LogEntry
-    deduplicationBufferSize?: number
     defaultNamespace: string
     resources: KubernetesResource<BaseResource>[]
     retryIntervalMs?: number
@@ -214,13 +189,11 @@ export class K8sLogFollower<T> {
     this.connections = {}
     this.k8sApi = k8sApi
     this.log = log
-    this.deduplicationBufferSize = deduplicationBufferSize
     this.defaultNamespace = defaultNamespace
     this.resources = resources
     this.intervalId = null
     this.resolve = null
     this.retryIntervalMs = retryIntervalMs
-    this.deduplicationBuffers = {}
   }
 
   /**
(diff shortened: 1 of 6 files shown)


edvald:
Nice 👍  Looks good at a glance, but I'll leave a full review to others.

One thing we should do is to get those we know to have had issues with log outputs to test the edge build, after this is merged.

thsig:
> One thing we should do is to get those we know to have had issues with log outputs to test the edge build, after this is merged.

Definitely!

I'm also not 100% certain that this will fix the issue in question (although I think it's likely)—it's a solid improvement at any rate.

core/src/plugins/kubernetes/logs.ts
```diff
@@ -324,18 +358,9 @@ export class K8sLogFollower<T> {
...
-            const dateInstance = new Date(parts[0])
-            if (isValidDateInstance(dateInstance)) {
-              timestamp = dateInstance
-            }
-            msg = parts[1]
-          } catch {}
-          if (_self.deduplicate({ msg, podName, containerName, timestamp })) {
+          const { timestamp, msg } = parseTimestampAndMessage(line)
+          // If we can't parse the timestamp, we skip the deduplication check and just write.
+          if (!timestamp || _self.deduplicate({ podName, containerName, timestamp, msg })) {
```
stefreak, line 363: In my testing, if the timestamp is missing then the message was not part of the Pod's log output, but an error message from the Kubernetes API.

This is how I tested this:
1. modify modify the vote integ test to run longer and log some output
2. run `gdev test -l3 -f vote` in the vote example.
3. delete the test pod

Output:
```diff
ℹ vote [verbose]            → [npm]: test
ℹ vote [verbose]            → [npm]: test
ℹ vote [verbose]            → [npm]: test
ℹ vote [verbose]            → [npm]: test
+ℹ vote [verbose]            → [npm]: error: code = Unknown desc = Error: No such container: 1e5ca22496f210e089ebda21c160e69c14ee17ec639b176f1916b8b7d1fc5c22
✖ vote                      → Running integ tests → Failed! (took 31.6 sec)

Failed running integ tests in module vote. Here is the output:
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Could not find Pod while waiting for it to complete. The Pod might have been evicted or deleted.
There are several different possible causes for Pod disruptions.

You can read more about the topic in the Kubernetes documentation:
https://kubernetes.io/docs/concepts/workloads/pods/disruptions/

━━━ Events ━━━
Pod test-vote-integ-86ac3d: Scheduled - Successfully assigned vote-default/test-vote-integ-86ac3d to docker-desktop
Pod test-vote-integ-86ac3d: Pulled - Container image "vote:v-6639ee3d75" already present on machine
Pod test-vote-integ-86ac3d: Created - Created container main
Pod test-vote-integ-86ac3d: Started - Started container main
Pod test-vote-integ-86ac3d: Killing - Stopping container main
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
1 test action(s) failed!
```

If I change the code to ignore messages without a timestamp I get the correct result:

```
ℹ vote [verbose]            → [npm]: test
ℹ vote [verbose]            → [npm]: test
ℹ vote [verbose]            → [npm]: test
✖ vote                      → Running integ tests → Failed! (took 24.2 sec)

Failed running integ tests in module vote. Here is the output:
Could not find Pod while waiting for it to complete. The Pod might have been evicted or deleted.
There are several different possible causes for Pod disruptions.

You can read more about the topic in the Kubernetes documentation:
https://kubernetes.io/docs/concepts/workloads/pods/disruptions/

━━━ Events ━━━
Pod test-vote-integ-31d558: Scheduled - Successfully assigned vote-default/test-vote-integ-31d558 to docker-desktop
Pod test-vote-integ-31d558: Pulled - Container image "vote:v-6639ee3d75" already present on machine
Pod test-vote-integ-31d558: Created - Created container main
Pod test-vote-integ-31d558: Started - Started container main
Pod test-vote-integ-31d558: Killing - Stopping container main
1 test action(s) failed!
```

Also, if we change the types so that we always receive a timestamp in the events down the line, some code can be simplified in the PodRunner (`prepareLogsFollower`).

I think instead of including log lines without a timestamp into the application logs, we should emit a warning or a debug message that we received this log line. This also might help reduce some confusion of developers, who might think that this (or a different) error has been emitted by their app, but it actually originates from kubernetes.

core/src/plugins/kubernetes/logs.ts
```diff
@@ -445,29 +470,41 @@ export class K8sLogFollower<T> {
...
+      return true
+    } else {
+      const time = timestamp.getTime()
+      const lastTime = lastEntry.timestamp.getTime()
+      const lastMsg = lastEntry.msg
+      if (time < lastTime) {
+        return false
+      } else {
+        if (time === lastTime && msg === lastMsg) {
+          return false
```
stefreak, line 497: With this logic we might get duplicates if an application outputs multiple different lines of output at the very same time.

Could we change the `deduplicate` function so it takes an array of messages as the first parameter, and returns an array of messages that should be written? Then we could make it more robust:

First, it would filter all the messages that have an older timestamp than the one we remembered in lastLogEntries.

Then it would need to find the index of the message that we remembered in the lastLogEntries. If it doesn't find this message in the array, we remember the last entry of the array as `lastLogEntry` and return the filtered array.

If it finds the message, it removes all the entries before the index of the found message, remembers the new `lastLogEntry` and returns the filtered + deduplicated array.

stefreak: Also it might simplify the calling code as it doesn't need to loop through the log entries

core/src/plugins/kubernetes/logs.ts
```diff
@@ -356,7 +381,7 @@ export class K8sLogFollower<T> {
           limitBytes,
           tail,
           timestamps: true,
-          // If we're retrying, presunmably because the connection was cut, we only want the latest logs.
+          // If we're retrying, presumably because the connection was cut, we only want the latest logs.
           // Otherwise we might end up fetching logs that have already been rendered.
           since: isRetry ? "10s" : since,
```
stefreak, line 386: To make it more robust: What if we never send a `since` argument. Then we would always fetch all the logs on retry, but that's fine because retries are rare and the deduplication logic would now be able to deal with any amount of messages.

stefreak (changes requested):
Awesome, this os definitely an improvement 👍 

I had a couple of comments, what do you think about those ideas and thoughts?

core/src/plugins/kubernetes/logs.ts
stefreak: Consider this one optional, it might not be worth it to be that "paranoid" about missing out on some messages

core/src/plugins/kubernetes/logs.ts
```diff
@@ -70,34 +76,32 @@ export async function streamK8sLogs(params: GetAllLogsParams) {
...
   entryConverter: PodLogEntryConverter<T>
   tail?: number
   pod: KubernetesPod
   defaultNamespace: string
   since?: string
-}) {
-  const api = await KubeApi.factory(log, ctx, provider)
-
+}): Promise<T[]> {
   const logs = await getPodLogs({
```
stefreak, line 105: this function calls an imported `getPodLogs` function, while we also have a `getPodLogs` function defined here in this file (could be called with `this.getPodLogs` – why are we using two different mechanisms of fetching logs in the LogsFollower? Might be confusing for other people reading this code

core/src/plugins/kubernetes/logs.ts
stefreak: Best would be of course, to pass a `sinceTime` argument to kubernetes, based on the last log message that we remembered. 

I did a little experiment, and kubernetes seems to adhere to the `sinceTime` argument, and even typescript doesn't complain:

```
diff --git a/core/src/plugins/kubernetes/logs.ts b/core/src/plugins/kubernetes/logs.ts
--- a/core/src/plugins/kubernetes/logs.ts
+++ b/core/src/plugins/kubernetes/logs.ts
@@ -443,13 +443,12 @@ export class K8sLogFollower<T extends LogEntryBase> {
     since?: string
   }) {
     const logger = this.k8sApi.getLogger()
-    const sinceSeconds = since ? parseDuration(since, "s") || undefined : undefined

     const opts = {
       follow: true,
       pretty: false,
       previous: false,
-      sinceSeconds,
+      sinceTime: "2024-02-07T14:44:14.054636925Z",
       tailLines: tail,
       timestamps,
     }
```

core/src/plugins/kubernetes/logs.ts
thsig: > In my testing, if the timestamp is missing then the message was not part of the Pod's log output, but an error message from the Kubernetes API.

Yeah, good point. I hadn't noticed this—I'll change the logic to skip logging entries with missing timestamps (but log those messages as debug-level logger messages).

core/src/plugins/kubernetes/logs.ts
thsig: Makes sense—could you have a go at making that change and see whether the results look promising, since you've almost written it out in pseudocode here?

core/src/plugins/kubernetes/logs.ts
thsig: Yeah, good point. We should rename the private method here to `streamPodLogs`, since it uses the streaming logs k8s API.

The other one (in `status/pod.ts`) calls the old-fashioned `readNamespacedPodLogs` k8s API endpoint: https://github.com/garden-io/garden/blob/98aeba2b65817c99b1019c25debddeb9a41f005e/core/src/plugins/kubernetes/status/pod.ts#L73.

core/src/plugins/kubernetes/logs.ts
thsig: > To make it more robust: What if we never send a since argument. Then we would always fetch all the logs on retry, but that's fine because retries are rare and the deduplication logic would now be able to deal with any amount of messages.

Yeah, maybe that would be better—my thinking was to avoid fetching all the logs when there's a lot of them (e.g. for very verbose, long-running test suites and the like).

> Should we maybe use sinceTime after all, to only fetch exactly the logs that we need?

Since the k8s Node client library doesn't allow for a `sinceTime` parameter, we'd have to write our own helper to prepare the request parameters directly.

That's totally doable (and not a lot of effort either), but not necessary if we're thinking about just fetching all the pod logs anyway.

I'll have a look at that, and if it ends up looking like too much work, we can keep the current approach (or just fetch all the pod logs as you suggested above).

stefreak: > Since the k8s Node client library doesn't allow for a sinceTime parameter, we'd have to write our own helper to prepare the request parameters directly.

In my testing (see the diff above) adding the sinceTime parameter to opts in `getPodLogs` just works

core/src/plugins/kubernetes/logs.ts
```diff
@@ -499,7 +536,29 @@ export interface PodLogEntryConverterParams {
...
+    msg = parts[1]
+  } catch {}
+  return timestamp ? { msg, timestamp } : { msg }
+}
+
+/**
+ * Returns a list of container names from which to fetch logs. Ignores sidecar containers injected by Garden.
+ */
+function containerNamesForLogging(pod: KubernetesPod): string[] {
+  return pod.spec!.containers.map((c) => c.name).filter((n) => !n.match(/garden-/))
```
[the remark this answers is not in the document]
eysi09: ...I don't think we have containers that match the regex but I wonder if we can make the name of the sidecar containers more specific. Something like `garden-injected-sidecar_mutagen` or similar.

core/src/plugins/kubernetes/run.ts
```diff
@@ -819,7 +819,7 @@ export class PodRunner extends PodRunnerParams {
     })
     return new K8sLogFollower({
       defaultNamespace: this.namespace,
-      retryIntervalMs: 10,
+      retryIntervalMs: 4000,
```
eysi09, line 822: Any specific reason for picking this exact number? 

If so it could be good to add it as a comment.

core/test/integ/src/plugins/kubernetes/container/logs.ts
```diff
@@ -79,7 +79,7 @@ describe("kubernetes", () => {
     describe("K8sLogsFollower", () => {
       let logsFollower: K8sLogFollower<ServiceLogEntry>
 
-      afterEach(() => {
+      afterEach(async () => {
```
eysi09, line 82: Doesn't look like this needs to be async.

core/src/plugins/kubernetes/logs.ts
```diff
@@ -153,37 +150,19 @@ interface LogOpts {
...
-export class K8sLogFollower<T> {
+export class K8sLogFollower<T extends LogEntryBase> {
   private connections: { [key: string]: LogConnection }
   private stream: Stream<T>
   private entryConverter: PodLogEntryConverter<T>
   private k8sApi: KubeApi
   private log: LogEntry
-  private deduplicationBufferSize: number
-  private deduplicationBuffers: { [key: string]: { msg: string; time: number }[] }
+  private lastLogEntries: { [containerKey: string]: { msg: string; timestamp: Date } }
```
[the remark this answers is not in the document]
stefreak: I was trying to change the types (I wanted to be timestamps to be non-optional everywhere in the kubernetes module) But it was a big type mess, as the Stream object with these types is handed in from the framework into the plugin. 

I wonder if timestamp really needs to be optional for some plugins? For the kubernetes plugin it definitely must not...

core/src/plugins/kubernetes/logs.ts
```diff
@@ -445,31 +516,54 @@ export class K8sLogFollower<T> {
...
-      buffer.shift()
+
+    if (time === lastTime) {
+      if (lastMessages.includes(msg)) {
+        return false
+      }
+
+      // This message has not been seen yet + has the same time stamp.
+      // Let's add it to the array of lastLogEntries
+      this.lastLogEntries[key].messages.push(msg)
```
stefreak, line 554:
```suggestion
      lastMessages.push(msg)
```

core/src/plugins/kubernetes/logs.ts
```diff
@@ -240,39 +208,106 @@ export class K8sLogFollower<T> {
...
+      }
+
+      if (allConnections.every((c) => c.status === "closed" && c.shouldRetry === false)) {
+        this.log.debug("flushFinalLogs: all connections were finished. Success!")
+        return
+      }
+      await sleep(1000)
+    }
+
+    this.log.debug("Failed to flush final logs: Timed out after 20 seconds")
```
stefreak, line 262: Should this be warning or error level?

core/src/plugins/kubernetes/logs.ts
```diff
@@ -240,39 +208,106 @@ export class K8sLogFollower<T> {
...
-  private handleConnectionClose(connectionId: string, status: ConnectionStatus, reason: string) {
+  private async flushFinalLogs() {
+    this.log.debug("flushFinalLogs called...")
+
+    // wait max 20 seconds
+    for (let i = 0; i < 20; i++) {
+      const allConnections = Object.keys(this.connections).map((c) => this.connections[c])
+
+      if (allConnections.length === 0) {
+        throw new Error("No connections")
```
stefreak, line 252: This is more or less a sanity-check assertion; Should we remove this so it doesn't cause weird looking error messages some point in the future? (Maybe we convert it to a debug level message)

core/src/plugins/kubernetes/logs.ts
```diff
@@ -240,39 +208,106 @@ export class K8sLogFollower<T> {
...
+  private async handleConnectionClose(connectionId: string, status: ConnectionStatus, reason: string) {
     const conn = this.connections[connectionId]
     const prevStatus = conn.status
+
+    clearTimeout(conn.timeout)
+
+    let retryReason = "Will retry in the background..."
+    let shouldRetry: boolean = true
+    try {
+      const pod = await this.k8sApi.core.readNamespacedPodStatus(conn.pod.metadata.name, conn.namespace)
```
stefreak, line 274: When I test this using Network Link conditioner, we will not see the "Connection timed out" message because readNamespacedPodStatus will re-try effectively forever. Can we somehow disable retry?

Alternatively we could move the decision wether to re-try or not into the createConnections function. That would have the benefit that handleConnectionClose wouldn't need to be async, which is kind of weird.

core/src/plugins/kubernetes/logs.ts
```diff
@@ -283,82 +318,107 @@ export class K8sLogFollower<T> {
...
       const writableStream = new Writable({
         write(chunk, _encoding, next) {
+          // clear the timeout, as we have activity on the socket
+          clearTimeout(conn.timeout)
+          conn.timeout = makeTimeout()
+
           const line = chunk?.toString()?.trimEnd()
 
+          // TODO: isn't it a bug if we don't call next?
           if (!line) {
```
stefreak, line 392: TODO: Do we have to call next() in this case?

core/src/plugins/kubernetes/logs.ts
stefreak: Absolutely, this is a very valid point.

In the current version of the PR, we handle this by enforcing `maxLogLinesInMemory` also for the LogFollower.
@eysi09 Do you think that suffices?

core/src/plugins/kubernetes/run.ts
```diff
@@ -819,7 +819,7 @@ export class PodRunner extends PodRunnerParams {
     })
     return new K8sLogFollower({
       defaultNamespace: this.namespace,
-      retryIntervalMs: 10,
+      retryIntervalMs: 1000,
```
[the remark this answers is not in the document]
stefreak: Should we also hardcode the retry interval, instead of parametrizing?

core/src/plugins/kubernetes/logs.ts
```diff
@@ -240,39 +208,106 @@ export class K8sLogFollower<T> {
...
+  }
+
+  private clearConnections() {
+    const conns = Object.values(this.connections)
     if (this.intervalId) {
       clearInterval(this.intervalId)
       this.intervalId = null
     }
-    Object.values(this.connections).forEach((conn) => {
+    Object.values(conns).forEach((conn) => {
```
stefreak, line 237: `Object.values(Object.values(this.connections))` seems like a bug / mistake

core/src/plugins/kubernetes/logs.ts
```diff
@@ -283,82 +318,107 @@ export class K8sLogFollower<T> {
...
-          } catch {}
-          if (_self.deduplicate({ msg, podName, containerName, timestamp })) {
+          const { timestamp, msg } = parseTimestampAndMessage(line)
+          // If we can't parse the timestamp, we encountered a kubernetes error
+
+          if (!timestamp) {
+            _self.log.debug(
+              `Encountered a log message without timestamp. This is probably an error message from the Kubernetes API: ${line}`
+            )
+          } else if (_self.deduplicate({ podName, containerName, timestamp, msg })) {
```
stefreak, line 403: Add a debug log message when messages have been discarded due to being a duplicate (it should now be rare, usually one message per reconnect)

core/src/plugins/kubernetes/logs.ts
thsig: Yeah, that's unnecessary. The second call to `Object.values` is a no-op there (it will return the same array).

core/src/plugins/kubernetes/logs.ts
thsig: I think converting it to a debug-level log message makes sense.

core/src/plugins/kubernetes/logs.ts
thsig: Warning, I'd say.

core/src/plugins/kubernetes/logs.ts
```diff
@@ -240,39 +208,106 @@ export class K8sLogFollower<T> {
...
-    this.resolve && this.resolve({})
   }
 
-  private handleConnectionClose(connectionId: string, status: ConnectionStatus, reason: string) {
+  private async flushFinalLogs() {
+    this.log.debug("flushFinalLogs called...")
+
+    // wait max 20 seconds
+    for (let i = 0; i < 20; i++) {
+      const allConnections = Object.keys(this.connections).map((c) => this.connections[c])
```
thsig, line 249: This can be replaced with `Object.values(this.connections)`.

core/src/plugins/kubernetes/logs.ts
thsig: > Can we somehow disable retry?

We could do that, e.g. by adding a new method to our `KubeApi` class that allows for optional request parameters (and adding support for a `retry: boolean` option there).

Should we do that, or do you think it could be cleaner to do this in `createConnections`? I'd he happy to make the required changes to `KubeApi` if we'd rather go that route, though.

core/src/plugins/kubernetes/logs.ts
stefreak:
```suggestion
      const allConnections = Object.values(this.connections)
```

core/src/plugins/kubernetes/logs.ts
stefreak:
```suggestion
        this.log.debug("flushFinalLogs: unexpectedly encountered empty list of connections")
```

core/src/plugins/kubernetes/logs.ts
stefreak:
```suggestion
    this.log.warning("Failed to finish streaming logs: Timed out after 20 seconds. Some logs might be missing in the verbose log output, or in Garden Cloud.")
```

core/src/plugins/kubernetes/logs.ts
stefreak: After giving it another thought, I think it's not strictly necessary as the behaviour is fine with or without the retry. Just the debug log is strange, but we can live with that.

core/src/plugins/kubernetes/logs.ts
stefreak:
```suggestion
          } else if (!_self.deduplicate({ podName, containerName, timestamp, msg })) {
              _self.log.debug(`Dropping duplicate log message: ${line}`)
          } else {
```

core/src/plugins/kubernetes/logs.ts
stefreak:
```suggestion
          if (!line) {
```

stefreak, line 392:
```suggestion
          if (!line) {
              next()
```

core/test/integ/src/plugins/kubernetes/container/logs.ts
stefreak:
```suggestion
      afterEach(() => {
```

core/src/plugins/kubernetes/logs.ts
```diff
@@ -240,39 +208,108 @@ export class K8sLogFollower<T> {
...
+  /**
+   * Same as `close`, but also fetches the last several seconds of logs and streams any missing entries
+   * (in case any were missing).
+   */
+  public async closeAndFlush() {
+    await this.flushFinalLogs()
+    this.clearConnections()
+    if (this.resolve) {
+      this.resolve({})
+    }
```
stefreak, line 228: Let's keep this DRY
```suggestion
    this.close()
```

core/src/plugins/kubernetes/logs.ts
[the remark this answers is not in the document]
stefreak: Let's at least narrow it down to container names starting with `garden-` for now

```suggestion
  return pod.spec!.containers.map((c) => c.name).filter((n) => !n.match(/^garden-/))
```

core/src/plugins/kubernetes/run.ts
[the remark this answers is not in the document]
stefreak:
```suggestion
      # We use 1 second in the PodRunner, because the task / test will only finish once the LogFollower finished.
      # If this is too low, we waste resources (network/cpu) – if it's too high we add extra time to the run execution.
      retryIntervalMs: 1000,
```

core/src/plugins/kubernetes/logs.ts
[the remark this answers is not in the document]
stefreak: My suggestion, after multiple attempts of changing this: Let's keep this as-is for now

We might want to add this to the 0.13 TODO-list or have another go at this once 0.13 is finished.

thsig:
@eysi09: This is ready for final review now.

core/src/plugins/kubernetes/logs.ts
```diff
@@ -499,7 +581,29 @@ export interface PodLogEntryConverterParams {
...
+    msg = parts[1]
+  } catch {}
+  return timestamp ? { msg, timestamp } : { msg }
+}
+
+/**
+ * Returns a list of container names from which to fetch logs. Ignores sidecar containers injected by Garden.
+ */
+function containerNamesForLogging(pod: KubernetesPod): string[] {
+  return pod.spec!.containers.map((c) => c.name).filter((n) => !n.match(/^garden-/))
```
eysi09, line 603: Nice to add the `^`

eysi09 (changes requested):
Had a couple of comments. Thanks so much for all the attention to detail on this one!

core/src/plugins/kubernetes/logs.ts
```diff
@@ -240,39 +213,101 @@ export class K8sLogFollower<T> {
...
+
+      connection.shouldRetry = false
+    }
+
+    try {
+      const pod = await this.k8sApi.core.readNamespacedPodStatus(connection.pod.metadata.name, connection.namespace)
+      const podStatus = checkPodStatus(pod)
+
+      const wasError = prevStatus === "error" || status === "error"
+      if (podStatus === "missing" && !wasError) {
```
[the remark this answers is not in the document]
stefreak: No, this wouldn't break that, because every second we are calling `createConnections`, which would call `pods = await getAllPods(this.k8sApi, this.defaultNamespace, this.resources)` (`this.resources` would be, for example, the service resource), and find the new Pod for the service once it's there.

We are just stopping to retry connecting to the old Pod, which doesn't exist anymore and will also never exist again

core/src/plugins/kubernetes/logs.ts
```diff
@@ -7,28 +7,28 @@
...
+import { checkPodStatus, getPodLogs } from "./status/pod"
+import { splitFirst, isValidDateInstance, sleep } from "../../util/util"
 import { Writable } from "stream"
 import request from "request"
 import { LogLevel } from "../../logger/logger"
 
 // When not following logs, the entire log is read into memory and sorted.
 // We therefore set a maximum on the number of lines we fetch.
+
 const maxLogLinesInMemory = 100000
```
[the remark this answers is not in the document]
stefreak: So the garden logs --helps describes the tail option like this:

```
  -t, --tail         Number of lines to show for each service. Defaults to showing all log lines (up to a
                     certain limit). Takes precedence over the `--since` flag if both are set. Note that we
                     don't recommend using a large value here when in follow mode.
```

So if the user calls `garden logs` without the `--tail` option, we'll process up to 100K log lines on startup. Wouldn't it be also okay then to process 100K logs again when we resume the connections, if it was fine on startup?
If the user adds the `--tail` option it will also be honored on the reconnect.

Do you think we should change the limit also for startup, or only for the retry case? if so, why is it okay for startup but not the reconnect case?

core/src/plugins/kubernetes/logs.ts
```diff
@@ -283,82 +318,96 @@ export class K8sLogFollower<T> {
...
         },
       })
 
-      let req: request.Request
       try {
-        req = await this.getPodLogs({
-          namespace,
-          containerName,
-          podName: pod.metadata.name,
+        req = await this.streamPodLogs({
```
stefreak, line 406: If we can't fix the behaviour, I'd like to add a few comment before we merge this that explains that this function might hang forever sometimes, which is why we are not waiting for pending connections to fail but re-try connecting anyway. I think this is extremely confusing behaviour, and if we can't fix it we should at least try to document it

core/src/plugins/kubernetes/logs.ts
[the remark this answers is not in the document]
eysi09: Ah ok, thanks!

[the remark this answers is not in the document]
stefreak: Maybe I can add a comment to the code to clarify that

core/src/plugins/kubernetes/logs.ts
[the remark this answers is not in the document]
stefreak: Yes, I was also talking about the follow case, but now I got lost completely. 

I think you're right and we don't have a problem with memory at all (at least when we're not streaming logs to Cloud), especially now that we eliminated the deduplication buffer. We should therefore rename the misleading `maxLogLinesInMemory` constant.

I think it's more a problem with the amount of work (=time) that is necessary to download millions of lines of logs and print them to the terminal. I think for this reason, the limit definitely makes sense to set a default tail number for both with and without `--follow`.

I have the feeling that as this is nuanced and complicated enough it will be hard to get to a result quickly using the GitHub comments. Do you have time to pair on this?

core/src/plugins/kubernetes/logs.ts
```diff
@@ -214,65 +191,132 @@ export class K8sLogFollower<T> {
...
    * until outside code calls the close method.
    */
   public async followLogs(opts: LogOpts) {
-    await this.createConnections(opts)
+    // make sure that createConnections is never called concurrently (wait for it to finish, then wait retryIntervalMs)
+    const followLoop = async () => {
+      try {
+        await this.createConnections(opts)
+      } finally {
+        this.timeoutId = setTimeout(followLoop, this.retryIntervalMs)
```
stefreak, line 211: if `close()` is called while `createConnections` is still running, the `clearTimeout` in `close()` has no effect. We need to add an `if (this.timeoutId) {` here, to prevent continuation even though close has been called.

core/src/plugins/kubernetes/logs.ts
```diff
@@ -283,191 +327,240 @@ export class K8sLogFollower<T> {
...
+    }
+
+    // Do not pass sinceSeconds parameter if we use sinceTime. The sinceSeconds parameter should only be used for the
+    // first request; For any subsequent re-try we want to get all the missing logs in between the attempts (see above)
+    if (sinceSeconds && !sinceTime) {
+      opts["sinceSeconds"] = parseDuration(sinceSeconds, "s") || undefined
     }
 
-    return logger.log(namespace, podName, containerName, stream, opts)
+    return await logger.log({
```
stefreak, line 470: return is enough

eysi09 (approved):
Whoop whoop 🚀

Issue #3707, opened 2023-02-08
Fix tests in `analytics/analytics.ts`


vvagaytsev:
Need to fix the tests after the changes made in #3619

PR #3734, opened 2023-02-10
test: fix analytics.ts assertions

Walther:
Fixes https://github.com/garden-io/garden/issues/3707

We should probably double-check that all the `optedIn` / `optedOut` logic is correct - the tests pass now, but I would love to have another pair of eyes on them before merging.

### DIFF
--- a/core/test/unit/src/analytics/analytics.ts
+++ b/core/test/unit/src/analytics/analytics.ts
@@ -107,10 +107,8 @@ describe("AnalyticsHandler", () => {
       expect(newConfig.anonymousUserId).to.be.a("string")
       expect(newConfig).to.eql({
         anonymousUserId: newConfig.anonymousUserId,
-        firstRunAt: now.toUTCString(),
-        latestRunAt: now.toUTCString(),
-        optedIn: true,
-        cloudVersion: 0,
+        firstRunAt: now,
+        latestRunAt: now,
         cloudProfileEnabled: false,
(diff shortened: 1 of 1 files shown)


core/test/unit/src/analytics/analytics.ts
```diff
@@ -107,10 +107,8 @@ describe("AnalyticsHandler", () => {
...
       expect(newConfig).to.eql({
         anonymousUserId: newConfig.anonymousUserId,
-        firstRunAt: now.toUTCString(),
-        latestRunAt: now.toUTCString(),
-        optedIn: true,
-        cloudVersion: 0,
+        firstRunAt: now,
+        latestRunAt: now,
         cloudProfileEnabled: false,
       })
```
Orzelius, line 113: Is it expected not to have the `optedOut` field here?

Orzelius (approved):

core/test/unit/src/analytics/analytics.ts
Walther: Seems so? Optional field in the schema

https://github.com/garden-io/garden/blob/c3811d841647894936bc5cce319d4a04d660d72b/core/src/analytics/analytics.ts#L403-L412

https://github.com/garden-io/garden/blob/c3811d841647894936bc5cce319d4a04d660d72b/core/src/config-store/global.ts#L43-L51

PR #3756, opened 2023-02-14
chore(examples): update rust example to use action configs

Walther:

Fixes https://github.com/garden-io/garden/issues/3740

First time converting an example to the new config format.
Is this correct and aligned with the best practices?

### DIFF
--- a/examples/rust/README.md
+++ b/examples/rust/README.md
@@ -5,7 +5,7 @@ Example of using the Rust programming language with Garden.
 Local development with hot reload:
 
 ```bash
-garden dev
+garden deploy --dev
 ```
 
 Local deployment with a release mode build:
--- a/examples/rust/README.md
+++ b/examples/rust/README.md
@@ -2,7 +2,7 @@
 
 Example of using the Rust programming language with Garden.
 
-Local development with hot reload:
+Local development with live code synchronization:
 
 ```bash
 garden deploy --dev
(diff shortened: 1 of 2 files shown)


examples/rust/README.md
```diff
@@ -5,7 +5,7 @@ Example of using the Rust programming language with Garden.
 Local development with hot reload:
```
vvagaytsev, line 5: The term "hot reload" can be confused with the old `--hot-reload` flag. Let's use the "code synchronization" term instead.

examples/rust/backend/backend.garden.yml
```diff
@@ -1,27 +1,34 @@
...
-        containerPort: 3000
-    healthCheck:
-      httpGet:
-        path: /
-        port: http
-    ingresses:
-      - path: /
-        port: http
-        hostname: backend.${var.base-hostname}
+---
```
vvagaytsev, line 8: Nit: an empty line can be placed after the separator :)

Walther: Prettier seems to prefer not having an empty line after the separator. Do we want to change prettier configuration to prefer a newline after the yaml separator?

Walther: The yaml specification also seems to not have a newline after the separator in their examples: https://yaml.org/spec/1.2.2/#22-structures

vvagaytsev: Ah, ok :) Thanks for sharing this!

vvagaytsev (approved):
Looks good, great job! 👍 
If the example works fine with and without `--dev` flag, then it's ready to go! 🚀

Walther:
Thanks for the speedy review!

Issue #3487, opened 2023-01-04
[FEATURE]: More generalized solution for version hash and file path handling


Walther:
### Background / Motivation

We are relaxing the path requirements for `valueFiles` in https://github.com/garden-io/garden/pull/3445. This fixes https://github.com/garden-io/garden/issues/3367 but has a side-effect: the hash of the `valueFiles` might not be taken into account for the module version hash, if the file path is outside the module root.

Based on internal conversations, we're essentially trading issues here: allow users to specify the `valueFiles` in a bit more flexible manner to unblock them, while introducing the slight risk of possible version/cache invalidation issues.

### What should the user be able to do?

The user should be able to use `valueFiles` from other directories, and any changes should be properly reflected in the module version hash changing, resulting in a new deployment when needed.

### Why do they want to do this? What problem does it solve?

Smooths out possible edge cases that could cause confusing situations with cache invalidation.

### Suggested Implementation(s)

Because this would involve significant refactoring, this should be implemented after `0.13` has been merged and released.

### How important is this feature for you/your team?

:rose: It’s a nice to have, but nice things are nice :slightly_smiling_face:

vvagaytsev:
Closing this as stale. Let's reopen if this becomes a practical issue for the Garden users.

PR #3758, opened 2023-02-14
fix: fix source builds processing in build-staging + updated `build-dependencies` example

### LINKED ISSUE #3754: 0.13 example update: build-dependencies

vvagaytsev:

Fixes #3754

### DIFF
--- a/core/src/build-staging/build-staging.ts
+++ b/core/src/build-staging/build-staging.ts
@@ -51,7 +51,7 @@ export class BuildStaging {
     this.createdPaths = new Set()
   }
 
-  async syncFromSrc(action: BuildAction, log: Log) {
+  async syncFromSrc(action: BuildAction, log: Log, withDelete: boolean = true) {
     // We don't sync local exec modules to the build dir
     if (action.getConfig("buildAtSource")) {
       log.silly(`Skipping syncing from source, action ${action.longDescription()} has buildAtSource set to true`)
@@ -67,7 +67,7 @@ export class BuildStaging {
     await this.sync({
       sourceRoot: resolve(this.projectRoot, action.basePath()),
       targetRoot: buildPath,
-      withDelete: true,
+      withDelete,
       log,
       files,
     })
@@ -86,8 +86,6 @@ export class BuildStaging {
         )
       }
 
-      const sourceBuildPath = sourceBuild.getBuildPath()
-
       if (isAbsolute(copy.sourcePath)) {
         throw new ConfigurationError(`Source path in build dependency copy spec must be a relative path`, {
           copySpec: copy,
@@ -100,8 +98,12 @@ export class BuildStaging {
         })
       }
 
+      // init .garden/build directory of the source build before syncing it to the build directory of the target action
+      // here we do not want to remove any existing files produce by the source build action
+      await this.syncFromSrc(sourceBuild, log, false)
+
       return this.sync({
-        sourceRoot: sourceBuildPath,
+        sourceRoot: sourceBuild.getBuildPath(),
         targetRoot: buildPath,
         sourceRelPath: copy.sourcePath,
         targetRelPath: copy.targetPath,
--- a/core/src/build-staging/build-staging.ts
+++ b/core/src/build-staging/build-staging.ts
@@ -61,11 +61,12 @@ export class BuildStaging {
     // Normalize to relative POSIX-style paths
     const files = action.getFullVersion().files.map((f) => normalizeRelativePath(action.basePath(), f))
 
-    await this.ensureDir(action.getBuildPath())
+    const buildPath = action.getBuildPath()
+    await this.ensureDir(buildPath)
 
     await this.sync({
       sourceRoot: resolve(this.projectRoot, action.basePath()),
-      targetRoot: action.getBuildPath(),
+      targetRoot: buildPath,
       withDelete: true,
       log,
       files,
--- a/core/src/build-staging/build-staging.ts
+++ b/core/src/build-staging/build-staging.ts
@@ -51,7 +51,7 @@ export class BuildStaging {
     this.createdPaths = new Set()
   }
 
-  async syncFromSrc(action: BuildAction, log: Log, withDelete: boolean = true) {
+  async syncFromSrc({ action, log, withDelete = true }: { action: BuildAction; log: Log; withDelete?: boolean }) {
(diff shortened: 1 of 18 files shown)


vvagaytsev:
Converting back to draft. The deployment command fails with `-dev` flag.

core/src/plugins/container/config.ts
```diff
@@ -233,8 +233,6 @@ const devModeSyncSchema = () =>
   joi.object().keys({
     source: joi
       .posixPath()
-      .relativeOnly()
-      .subPathOnly()
```
Walther, line 237: does this removal have possible side effects?

i'm reminded of https://github.com/garden-io/garden/pull/3445 and https://github.com/garden-io/garden/issues/3487

vvagaytsev: We don't have these limitations in `0.12`, so I just aligned the validation logic with the `main` branch. @edvald do we want to allow using of the absolute paths as sync sources?

core/src/build-staging/build-staging.ts
```diff
@@ -99,8 +98,11 @@ export class BuildStaging {
         })
       }
 
+      // init .garden/build directory of the source build before syncing it to the build directory of the target action
+      await this.syncFromSrc(sourceBuild, log)
```
vvagaytsev, line 102: This can be avoided if we use `sourceRoot: sourceBuild.basePath(),` in [L105](https://github.com/garden-io/garden/pull/3758/files#diff-9eb4fcb064a98747d3c16f0bc8cb1301149611484b153f332b563716f33402e3R105).

But, if I got it correctly, all build actions' content should be copied to the `build` directory in `.garden/build/{action-name}`.

vvagaytsev:
There are some test failures, moving it back to the draft state.

core/src/build-staging/build-staging.ts
edvald: Could be right but I'd like to look more closely at this.

core/src/plugins/container/config.ts
edvald: Yes, this is okay since these paths don't have bearing on builds or versioning. In fact, this doesn't even have to be a POSIX path, can just be any file path and should work with absolute paths on Windows as well.

core/src/build-staging/build-staging.ts
vvagaytsev: Thanks! It would be nice if you could take a look! By the way, this might cause some [test failures](https://app.circleci.com/pipelines/github/garden-io/garden/15406/workflows/fae39d2c-adf6-419d-8b8f-4f6ca4a19e08/jobs/257392), but I'm not sure. I wasn't able to reproduce the same failures locally.

vvagaytsev: I reproduce the failures locally. Now, the test fails because `action.getFullVersion().files` in [L62](https://github.com/garden-io/garden/pull/3758/files#diff-9eb4fcb064a98747d3c16f0bc8cb1301149611484b153f332b563716f33402e3R62) does not contain the `a.txt` file which is produced by `module-a` Build action.

Still figuring out why that happens. Any help is highly appreciated :)

core/src/plugins/container/config.ts
vvagaytsev: Allowed Windows paths in https://github.com/garden-io/garden/pull/3758/commits/82de6d9d8a9a2ab8a21380238b21e179916931ae

core/src/build-staging/build-staging.ts
```diff
@@ -51,7 +51,7 @@ export class BuildStaging {
     this.createdPaths = new Set()
   }
 
-  async syncFromSrc(action: BuildAction, log: Log) {
+  async syncFromSrc(action: BuildAction, log: Log, withDelete: boolean = true) {
```
Orzelius, line 54: FYI ts will infer the type from assignment so technically `:boolean` is not needed here

Orzelius: But I think adding boolean parameters like this is a bit ugly as it's hard to understand what they're supposed to be by just looking at the function call. Maybe convert to a destructured object instead? Your ide will probably have the option to help out with this.

Orzelius:
Added a small non-blocking comment

core/src/build-staging/build-staging.ts
vvagaytsev: Agree, better to use parameter-object here. I'll refactor this in a separate commit.

vvagaytsev: Refactored this in https://github.com/garden-io/garden/pull/3758/commits/d9f9fc1f820bf2c24979a0282c3994171d0d2f15

vvagaytsev:
@edvald this is ready for review, could you take a look, please?

Orzelius:
Will have a look first thing tomorrow

examples/build-dependencies/frontend/garden.yml
```diff
@@ -1,44 +1,60 @@
...
-    ingresses:
-      - path: /hello-frontend
-        port: http
-      - path: /call-backend
-        port: http
-    dependencies:
-      - backend
-tests:
-  - name: unit
-    args: [npm, test]
```
Walther, line 40: looks like the unit test got dropped, let's add it back as another Test section

examples/build-dependencies/backend/garden.yml
```diff
@@ -1,23 +1,41 @@
...
+
+---
+
+kind: Run
+name: backend
+description: Simple run task
+type: container
+
+spec:
+  command: [ "sh", "-c", "echo task output" ]
```
Walther, line 41: idly pondering if we should keep these kinds of empty tasks and tests that only run `echo` or just remove them

Walther (approved):
two small comments, otherwise looks good to me! 🎉

Orzelius (approved):

examples/build-dependencies/frontend/garden.yml
vvagaytsev: Good catch, thanks!

examples/build-dependencies/backend/garden.yml
vvagaytsev: Yep, maybe we can drop them. But, otoh each that config just demonstrates what Garden can do. So maybe it's fine to keep it :)

PR #3755, opened 2023-02-14
0.13 example update: local-mode-k8s

vvagaytsev:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #3752

**Special notes for your reviewer**:

PR #3760, opened 2023-02-14
0.13 some local mode fixes

vvagaytsev:
Some fixes and refactoring in local mode impl.

This might be a pre-requisite for #3755

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/local-mode.ts
+++ b/core/src/plugins/kubernetes/local-mode.ts
@@ -76,18 +76,20 @@ export const kubernetesLocalModeSchema = () =>
     See the [Local Mode guide](${localModeGuideLink}) for more information.
   `)
 
-interface ConfigureLocalModeParams {
+interface BaseLocalModeParams {
   ctx: PluginContext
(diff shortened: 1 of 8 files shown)


core/src/plugins/kubernetes/local-mode.ts
```diff
@@ -47,18 +47,11 @@ const sshKeystoreAsyncLock = new AsyncLock()
...
   target?: KubernetesTargetResourceSpec
 }
 
 export const kubernetesLocalModeSchema = () =>
   containerLocalModeSchema().keys({
-    containerName: joi
-      .string()
-      .optional()
-      .description(
-        "When using the `defaultTarget` and not specifying `localMode.target`, this field can be used to override the default container name to proxy traffic from."
```
vvagaytsev, line 60: The logic described here has never been implemented. Looks lie we need to complicated conditional logic for this. So, I assumed it could be thrown away in favor of the `containerName` defined in the `target` field below.

edvald (approved):

PR #3764, opened 2023-02-15
Perf improvements and fixes

edvald:
Caught some issues as I was working to accelerate the unit/functional test suite. The tests now run faster, startup time is improved, and some important issues were caught relating to version resolution.

Please review commit by commit.

### DIFF
--- a/core/src/logger/log-entry.ts
+++ b/core/src/logger/log-entry.ts
@@ -32,7 +32,8 @@ export interface TaskMetadata {
   key: string
   status: TaskLogStatus
   uid: string
-  versionString: string
+  inputVersion: string
+  outputVersion?: string
   durationMs?: number
 }
 
--- a/core/src/tasks/base.ts
+++ b/core/src/tasks/base.ts
@@ -33,14 +33,12 @@ export function makeBaseKey(type: string, name: string) {
   return `${type}.${name}`
 }
 
-interface CommonTaskParams {
+export interface CommonTaskParams {
   garden: Garden
   log: LogEntry
   force: boolean
   skipDependencies?: boolean
 }
-
-export interface BaseTaskParams extends CommonTaskParams {}
 
 export interface BaseActionTaskParams<T extends Action = Action> extends CommonTaskParams {
   action: T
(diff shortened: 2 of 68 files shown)


core/src/logger/log-entry.ts
```diff
@@ -32,7 +32,8 @@ export interface TaskMetadata {
   key: string
   status: TaskLogStatus
   uid: string
-  versionString: string
+  inputVersion: string
+  outputVersion?: string
```
Orzelius, line 36: Okay, silly question, but what even are the task input and output versions?

Orzelius (approved):

core/src/vcs/git.ts
```diff
@@ -379,7 +379,7 @@ export class GitHandler extends VcsHandler {
 
     await new Promise<void>((_resolve, _reject) => {
       const proc = lsFiles(this.ignoreFile)
-      proc.on("error", (err: execa.ExecaError) => {
+      void proc.on("error", (err: execa.ExecaError) => {
```
vvagaytsev, line 382: I'm just curious, why do we need `void` here?

core/src/tasks/base.ts
```diff
@@ -40,9 +40,7 @@ interface CommonTaskParams {
   skipDependencies?: boolean
 }
 
-export interface BaseTaskParams extends CommonTaskParams {
-  version: string
-}
+export interface BaseTaskParams extends CommonTaskParams {}
```
vvagaytsev, line 43: This interface now looks redundant, shall we drop it?

vvagaytsev (approved):
LGTM! 🚀 I've left a few non-blocking comments.

core/src/logger/log-entry.ts
edvald: That's actually not at a silly question :) I'll explain this in comments, but basically an "input version" is generally the version of an action before it's resolved, the "output version" is e.g. after it's resolved, built etc. (which may in many cases transform the version).

core/test/unit/src/commands/custom.ts
```diff
@@ -220,7 +220,7 @@ describe("CustomCommandWrapper", () => {
       variables: {
         foo: "test",
       },
-      gardenCommand: ["get", "doddi"],
+      gardenCommand: ["validate"],
```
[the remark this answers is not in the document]
edvald: Well, you can always run the command at your leisure .)

core/src/vcs/git.ts
edvald: I actually think the code was fine before, but the linter complained about a promise not being awaited.

PR #3768, opened 2023-02-15
chore: switch the cleanup command emoji

Walther:
Just a random small thought - maybe some other emoji could be more aligned with our tone and visual style in the cleanup command.

Originally I wanted to use the broom emoji 🧹  but the emoji library we use does not have it available, so I chose the recycling icon instead. Let me know if you have better suggestions or if we should close without merging.


vvagaytsev (approved):
LGTM! A recycle bin looks less aggressive than a skull with crossed bones :)

PR #3772, opened 2023-02-15
chore(examples): update variables example to use action configs

Walther:

Fixes https://github.com/garden-io/garden/issues/3765

I noticed in the backend `main.go` and `Dockerfile` the port `8080` is used, so I edited that into the ingress definition (instead of `80`) and added it to the frontend as well (`const backendServiceEndpoint = 'http://backend:8080/hello-backend'`).
Let me know if there is a better way to fix this - I tried e.g. `containerPort: 8080, localPort: 80` on the backend ingress, but was unsuccessful.

### DIFF
--- a/examples/variables/README.md
+++ b/examples/variables/README.md
@@ -4,8 +4,8 @@ This variant of the basic [demo project](../demo-project/README.md) demonstrate 
 
 ## Project variables
 
-In this example, we set a project variable in the [project config](./garden.yml#L3-L5) called `service-replicas` and
-reference that variable in the module configs[[1](./backend/garden.yml#L18), [2](./frontend/garden.yml#L7)], in this case to set the number of replicas per service.
+In this example, we set a project variable in the [project config](./garden.yml#L5) called `service-replicas` and
+reference that variable in the action configs[[1](./backend/garden.yml#L23), [2](./frontend/garden.yml#L14)], in this case to set the number of replicas per service.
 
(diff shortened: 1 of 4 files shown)


vvagaytsev:
Looks good! 👍 We should also update the `README.md` file, it still uses the term "module".

examples/variables/README.md
```diff
@@ -4,16 +4,16 @@ This variant of the basic [demo project](../demo-project/README.md) demonstrate
...
 
 We also show how you can alternate these variables by the environment you're running, by overriding the default value
 in the `local` environment. In this case, we only want _one_ replica of each service while developing locally, but
 default to three when deploying remotely.
 
 ## Using variables with templating
 
-Also in the [backend module configuration](./backend/garden.yml#L6-L14) two variables are used to store ports configuration for the service and a variable
-is additionally used to detect dev-mode being used. Later in the [service ports configuration](./backend/garden.yml#L19) the variables are dynamically used to enable the development port if dev mode is being used. 
+Also in the [backend action configuration](./backend/garden.yml#L12) two variables are used to store ports configuration for the service and a variable
```
vvagaytsev, line 16: Nit: "backend action" -> "backend deploy action"

Walther: good catch, thanks!

vvagaytsev (approved):
Great job, thank you! 👍

PR #3784, opened 2023-02-16
chore(examples): update vote-helm example to use action configs

Walther:

Fixes https://github.com/garden-io/garden/issues/3767

- `garden validate` passes
- `garden deploy` passes
- `garden run db-init` and `garden run db-clear` pass
- `garden cleanup env` passes

Squash before merging.


core/src/actions/base.ts
```diff
@@ -89,7 +89,7 @@ const actionSourceSpecSchema = () =>
     .xor("path", "repository")
     .meta({ name: "action-source", advanced: true, templateContext: ActionConfigContext })
 
-export const includeExcludeSchema = () => joi.array().items(joi.posixPath().allowGlobs().subPathOnly())
+export const includeExcludeSchema = () => joi.array().items(joi.posixPath().allowGlobs())
```
Walther, line 92: ⚠️  Note this change here

examples/vote-helm/api/garden.yml
```diff
@@ -1,19 +1,19 @@
...
-    tag: ${modules.api-image.version}
-  ingress:
-    enabled: true
-    paths: [/]
-    hosts: ["api.${var.baseHostname}"]
-  healthCheckPath: /api
+  - deploy.redis
+spec:
+  chart:
+    path: ../base-chart/
```
Walther, line 9: ⚠️  Note the change required in order to be able to use relative path here

examples/vote-helm/result/garden.yml
```diff
@@ -1,23 +1,18 @@
...
-tests:
-  - name: integ
-    args: [echo, ok]
-    dependencies:
-      - db-init
\ No newline at end of file
+  - run.db-init
+spec:
+  chart:
+    path: ../base-chart/
```
Walther, line 9: ⚠️  and here

core/src/actions/base.ts
edvald: We can't actually make this change, sorry. This won't play nice with version calculation and build staging. It is also a conscious design choice not to allow pulling files from outside the module root.

vvagaytsev (approved):

PR #3786, opened 2023-02-16
WIP: garden volume mount logic fixes

Orzelius:
WIP


vvagaytsev:
@Orzelius the issue happens [here](https://github.com/garden-io/garden/blob/0.13/core/src/tasks/resolve-action.ts#L217), when a `ResolvedDeployAction` is created.

It looks like the validator applies some transformations defined in the schema and transforms the original object. This might be a side effect of the [`ActionReferenceSchema` implementation](https://github.com/garden-io/garden/blob/0.13/core/src/config/common.ts).

core/src/tasks/deploy.ts
```diff
@@ -56,6 +56,10 @@ export class DeployTask extends ExecuteActionTask<DeployAction, DeployStatus> {
 
     const router = await this.garden.getActionRouter()
 
+    if (!status){
+      console.log("action status was not passed ):")
```
Orzelius, line 60: after my fixes the status is not passed while running the deploy process. 

Can be reproduced by running the [`should deploy a service referencing a volume module`](https://github.com/garden-io/garden/blob/integ-volume-stuff/core/test/integ/src/plugins/kubernetes/container/deployment.ts#L835) integ test.

Here's the log

```log
[silly] Completing node deploy.volume-module:process. aborted=false, error=Cannot read properties of null (reading 'detail')
[silly] Completing node deploy.volume-reference:process. aborted=true, error=processing Deploy type=container name=volume-reference (from module volume-reference) failed: TypeError: Cannot read properties of null (reading 'detail')
[silly] Completing node deploy.volume-reference:request:c784021c-ef07-487d-85d0-e3bc3bb44779. aborted=true, error=Deploy type=container name=volume-reference (from module volume-reference) failed: Error: processing Deploy type=container name=volume-reference (from module volume-reference) failed: TypeError: Cannot read properties of null (reading 'detail')
[silly] GraphSolver: Complete handler for batch c784021c-ef07-487d-85d0-e3bc3bb44779 called with result deploy.volume-reference
[silly] GraphSolver: Complete handler for batch c784021c-ef07-487d-85d0-e3bc3bb44779 matched with request deploy.volume-reference:request:c784021c-ef07-487d-85d0-e3bc3bb44779

Failed to processing Deploy type=persistentvolumeclaim name=volume-module (from module volume-module). Here is the output:
Cannot read properties of null (reading 'detail')
[silly] Full error with stack trace:
TypeError: Cannot read properties of null (reading 'detail')
    at wrapped.process (/Users/orz/work/garden/core/src/tasks/deploy.ts:64:40)
    at wrapped.processNode (/Users/orz/work/garden/core/src/graph/solver.ts:370:29)
[silly] GraphSolver: loop
[silly] Task nodes in progress: (none)
```

Orzelius:
Marking this as ready for review, but please note the bug [described in the comment above](https://github.com/garden-io/garden/pull/3786#discussion_r1109433412). Even if merged as is this will not make things worse as the volume logic doesn't work anyway due the bugs, fixed in the fix commits in this pr.

vvagaytsev (approved):
LGTM! But I'd like to have another pair of eyes on this :)

edvald:
This is okay to merge once that console.log that the linter is complaining about is removed .) The actual issue that's coming up seems entirely unrelated to the fix in the PR.

Orzelius:
Removed the console log commit, please re-approve :)

PR #3788, opened 2023-02-16
chore(examples): update conftest example to use action configs

Walther:

Fixes https://github.com/garden-io/garden/issues/3785

Drawback: helm conftest requires the deployment to happen first before running the linter on the configs

Squash before merging.


plugins/conftest/index.ts
```diff
@@ -210,7 +210,7 @@ export const gardenPlugin = () =>
             configure: async ({ ctx, config }) => {
               let files = config.spec.files
 
-              if (files.length > 0) {
+              if (files && files.length > 0) {
```
vvagaytsev, line 213: I made a similar fix for k8s plugin in https://github.com/garden-io/garden/commit/e2974d9d09db0c1f0c4e2f8b9e7ef619ddf2d0d1 :)

Walther: nice, that seems like a better fix so adapted it to here as well ❤️

PR #3799, opened 2023-02-17
chore(examples): update disabled-configs example to use action configs

### LINKED ISSUE #3796: 0.13 example update: disabled-configs

vvagaytsev:

Fixes #3796

Note, this is a pre-review with some discussions and questions to be resolved.


core/src/graph/actions.ts
```diff
@@ -299,11 +298,17 @@ export async function executeAction<T extends Action>({
...
+    [key: string]: object
+  }[]
+}
+
 const getActionConfigContextKeys = memoize(() => {
   const schema = buildActionConfigSchema()
   const configKeys = schema.describe().keys
   return Object.entries(configKeys)
-    .map(([k, v]) => ((<any>v).meta?.templateContext === ProjectConfigContext ? k : null))
+    .map(([k, v]) => ((<HasMetadata>v).metas.find((m) => m.templateContext === ActionConfigContext) ? k : null))
```
vvagaytsev, line 311: It looks like we need to change his predicate to use `ActionConfigContext` instead of `ProjectConfigContext` because the former one contains more fields that can use template strings.

Also, I changed the assumption on the shape of a value object of `configKeys` map. Introduced a type to ensure type-safety.

examples/disabled-configs/frontend/garden.yml
```diff
@@ -1,28 +1,53 @@
...
-    args: [npm, test]
-  - name: integ
-    args: [npm, run, integ]
-    disabled: ${environment.name == "local"}
-    dependencies:
-      - frontend
+
+build: frontend
+dependencies:
+  - deploy.backend
```
vvagaytsev, line 14: Now, this causes an error:
```
Failed to processing resolve Deploy type=container name=backend. Here is the output:
Build type=container name=backend is disabled.
1 deploy action(s) failed!
```

Should we use conditional dependencies here, or allow dependencies on disabled actions?

vvagaytsev:
@edvald I've left some open discussions, could you check those, please? I need your help.
cc @thsig

examples/disabled-configs/frontend/garden.yml
edvald: We should allow this, same as with modules, services etc

core/src/graph/actions.ts
vvagaytsev: Hmm... Looks like it caused some [test failures](https://app.circleci.com/pipelines/github/garden-io/garden/15555/workflows/2a894228-9a47-4b15-8d2e-f30b5206cd39/jobs/259960), so I'm still not sure if this is the proper way of fixing the issue.

vvagaytsev: Maybe we should check if the type of `m.templateContext` extends `ProjectConfigContext` here.

edvald: The name of the function is perhaps misleading. Notice in the code that uses the outputs, that we actually _only_ want keys from ProjectConfigContext.

To make the function more robust, it would make sense to use this: https://github.com/garden-io/garden/blob/25fe71e02a0b65fea294c83ed4f03a21d9278342/core/src/config/common.ts#L899

vvagaytsev: Ok, thanks! I reverted the template context predicate and rename the function in https://github.com/garden-io/garden/pull/3799/commits/c9ed235f597bd35f376caf70d6a23cf839d858a7

core/src/graph/actions.ts
```diff
@@ -376,7 +376,7 @@ async function preprocessActionConfig({
 
     // Partially resolve other fields
     // TODO-G2: better error messages when something goes wrong here
-    const resolvedOther = resolveTemplateStrings(omit(config, projectContextKeys), builtinFieldContext, {
+    const resolvedOther = resolveTemplateStrings(omit(config, projectConfigContextKeys), builtinFieldContext, {
```
vvagaytsev, line 379: The comment says "Partially resolve **other** fields". Does it mean that we should not use `ProjectConfigContext` here? Should we use `ActionConfigContext` or all possible contexts here? cc @edvald

core/src/graph/actions.ts
```diff
@@ -299,11 +299,11 @@ export async function executeAction<T extends Action>({
   return <Executed<T>>(<unknown>results.results.getResult(task)!.result!.executedAction)
 }
 
-const getActionConfigContextKeys = memoize(() => {
+const getProjectConfigContextKeys = memoize(() => {
   const schema = buildActionConfigSchema()
```
vvagaytsev, line 303: This also looks incorrect. Should we use project config schema instead of `buildActionConfigSchema` @edvald?

vvagaytsev:
Looks great! 👍 Thank you so much @edvald 🚀 

Unfortunately, I can approve this PR because there is one my commit. I'll ask someone else to take a look. The example works fine, all checks passed. So it's ready to go!

edvald (approved):

Issue #3583, opened 2023-01-25
Fix integration tests for `hadolint` plugin


vvagaytsev:
- [x] fix runtime error in the code -> done in #3585
- [x] fix schema validation issues in the `testModule` spec -> in progress in #3628

vvagaytsev:
There are some new failures, reopeninig.

PR #3800, opened 2023-02-17
0.13 fix hadolint plugin

vvagaytsev:
Some fixes in `hadolint` plugin.
...and type-safety for all :)

There was some mess with the `Dockerfile` path processing for different kinds and types of actions. A type-safe approach made it better.

Partially addresses #3583

See individual commits for details, it might be simpler to review it commit-by-commit.


Walther (commented):
thanks for the fixes and refactor! ✨
looks good to me but i'll let another pair of eyes verify as well, as i'm not 100% confident i'm following all the logic here

edvald (approved):

PR #3803, opened 2023-02-17
chore(examples): update kaniko example to use action configs

### LINKED ISSUE #3801: 0.13 example update: kaniko

Walther:

Fixes #3801

### DIFF
--- a/examples/kaniko/README.md
+++ b/examples/kaniko/README.md
@@ -2,7 +2,7 @@
 
 A very basic demo project for Garden showing in-cluster building with [kaniko](https://github.com/GoogleContainerTools/kaniko).
 
-Just run `garden deploy` and see the kaniko builder do its job!
+Just run `garden deploy --env=remote` and see the kaniko builder do its job!
 
 Snippet from [project config](garden.yml)
 ```yml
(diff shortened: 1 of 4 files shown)


examples/kaniko/README.md
```diff
@@ -11,10 +11,16 @@ name: kaniko
 environments:
   - name: local
   - name: remote
-    defaultNamespace: ${var.userId}
 providers:
-  - name: local-kubernetes
-    environments: [local]
```
edvald, line 17: Is there a reason to move this? If there is, best to remove the `local` environment above as well.

examples/kaniko/frontend/garden.yml
```diff
@@ -1,27 +1,43 @@
...
+    - path: /call-backend
+      port: http
+
+---
+kind: Test
+type: container
+build: frontend
+name: frontend-unit
+spec:
+  args: [npm, test]
```
edvald, line 33: Suggest using `command` here instead, seeing as we'd never want this as arguments to an `ENTRYPOINT` defined in the Dockerfile.

examples/kaniko/README.md
vvagaytsev: Ah, yes, actually, we can leave this, just to emphasize that the `local-kubernetes` provider does not use kaniko build. Fixed in the force-pushed commit.

examples/kaniko/frontend/garden.yml
vvagaytsev: Fixed in https://github.com/garden-io/garden/pull/3803/commits/200844d72c10579039ec7dea4ccfa160bf08cd8e

examples/kaniko/README.md
vvagaytsev: Fixed in https://github.com/garden-io/garden/pull/3803/commits/f6699bc78b61c0c0e97bd8c9af2e36a9bd8c12bc

edvald (approved):

PR #3805, opened 2023-02-17
chore(examples): create a modules-to-actions example

Walther:

Fixes https://github.com/garden-io/garden/issues/3804, although the scope has changed.

Will be squashed before merging.


edvald:
I'm sort of thinking this example is redundant now. When it was made, being able to put several modules in one file was a new feature, but it's so basic at this point I think it feels unnecessary :P

Orzelius:
The feature might be obvious to us but newcomers might not think of it.

Also, we should have an example with module and document how they get converted to actions and what limits and problems users might face if they stick with modules. This example could serve that purpose.

edvald:
Sure, I could go with that. In that case I'd suggest renaming the project (I don't know... `legacy-modules`?) and updating the docs accordingly.

vvagaytsev (approved):
Looks good! Thank you! 👍

PR #3814, opened 2023-02-20
fix(cloud): continue when token refresh fails

mkhq:

Improve handling of the case when switching between projects with that uses different cloud API endpoints.

- Continues the command even if the token refresh fails. 
- Clearer error message indicating that a logout is needed before the cloud API can be used for the current project.

Fixes #

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -54,7 +54,7 @@ import type { GardenProcess } from "../db/entities/garden-process"
 import { DashboardEventStream } from "../server/dashboard-event-stream"
 import { GardenPluginReference } from "../types/plugin/plugin"
 import { renderError } from "../logger/renderers"
-import { CloudApi, EnterpriseApiTokenRefreshError, getGardenCloudDomain } from "../cloud/api"
+import { CloudApi, CloudApiTokenRefreshError, getGardenCloudDomain } from "../cloud/api"
 import { findProjectConfig } from "../config/base"
 import { pMemoizeDecorator } from "../lib/p-memoize"
 import { getCustomCommands } from "../commands/custom"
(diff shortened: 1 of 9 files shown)


core/src/cli/cli.ts
```diff
@@ -285,12 +285,29 @@ ${renderCommands(commands)}
...
-      let cloudDomain: string = getGardenCloudDomain(config)
+      const cloudDomain: string = getGardenCloudDomain(config)
+      distroName = getCloudDistributionName(cloudDomain)
 
-      cloudApi = await CloudApi.factory({ log, cloudDomain })
+      try {
+        cloudApi = await CloudApi.factory({ log, cloudDomain })
+      } catch (err) {
+        if (err instanceof EnterpriseApiTokenRefreshError) {
+          log.warn(dedent`
```
stefreak, line 300: Do we also just show a warning in projects where the cloud url is configured? This might lead to confusing error messages, e.g. missing secrets

core/src/cloud/api.ts
```diff
@@ -31,6 +31,7 @@ const gardenClientName = "garden-core"
 const gardenClientVersion = getPackageVersion()
 
 export class EnterpriseApiDuplicateProjectsError extends EnterpriseApiError {}
+export class EnterpriseApiTokenRefreshError extends EnterpriseApiError {}
```
eysi09, line 34: Nit pick but should we just call this `CloudApiTokenRefreshError`? Since we're phasing out `Enterprise*` anyway.

core/src/cli/cli.ts
eysi09: We do. 

I suppose you could argue that if you get a token refresh error and cloud domain is set then it should fail hard like before.

eysi09 (commented):
LGTM. Happy to approve after @stefreak's comment has been addressed.

stefreak (approved):
👍

PR #3816, opened 2023-02-20
fix(k8s): error when getting status of container Deploy in sync mode

edvald:
cc @Walther


Walther (approved):
Thanks for the fix! ✨ 🚀

PR #3833, opened 2023-02-22
chore(examples): update gatsby-code-sync example to use actions

Walther:

Fixes https://github.com/garden-io/garden/issues/3832

Updated the Node engine version and the dependencies while at it - the project had bitrotted to an un-runnable state.


Walther:
😂 what even, how is that lockfile diff so massive

vvagaytsev (approved):

PR #3834, opened 2023-02-22
cloud(fix): Update to use UUIDs

mkhq:

The cloud API types have changed to return strings instead of number. This PR updates to the latest version of api-types with resulting compatibility changes.

Fixes #


10ko (approved):
LGTM

PR #3836, opened 2023-02-23
feat(pulumi): allow CLI options for plugin cmds

thsig:

Any CLI options that aren't used by the Pulumi plugin commands themselves are now passed along to the Pulumi CLI when running plugin commands.

For example, `--skip-dependencies` won't be passed to the Pulumi CLI (since it's a Garden-specific option that doesn't apply to the Pulumi CLI), but the `--same` option would be passed along unchanged to `pulumi preview`.

For example, in `garden plugins pulumi preview -- some-service --same --suppress-outputs`, `--same` and `--suppres-outputs` would be passed as CLI options to `pulumi preview`.

### DIFF
--- a/docs/pulumi-plugin/about.md
+++ b/docs/pulumi-plugin/about.md
@@ -79,6 +79,10 @@ For example:
 ```
 garden plugins pulumi preview -- my-pulumi-module my-other-pulumi-module
 ```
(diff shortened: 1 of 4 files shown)


edvald:
Not sure if you were aware of `args["--"]` which is already passed to commands? I suspect you could simplify some of this by using that.

vvagaytsev:
@thsig do we want to have this in `0.12`? Or should we move it to `0.13`?

PR #3837, opened 2023-02-23
improvement: track the command run result and duration

mkhq:

When the anonymous analytics is enabled, we currently track that a command is run, but not the actual outcome of the command. With this PR, we add an analytics event after a command has finished including the result as success/failure, error type and command duration. 

Fixes #

### DIFF
--- a/core/src/analytics/analytics-types.ts
+++ b/core/src/analytics/analytics-types.ts
@@ -8,8 +8,14 @@
 
 export enum AnalyticsType {
   COMMAND = "Run Command",
+  COMMAND_RESULT = "Run Command Result",
   CALL_API = "Call API",
   MODULE_CONFIG_ERROR = "Module Configuration Error",
   PROJECT_CONFIG_ERROR = "Project Configuration Error",
   VALIDATION_ERROR = "Validation Error",
+}
+
+export enum AnalyticsCommandResult {
+  FAILURE = "failure",
+  SUCCESS = "success",
 }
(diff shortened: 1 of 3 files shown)


core/src/analytics/analytics-types.ts
```diff
@@ -8,8 +8,14 @@
 
 export enum AnalyticsType {
   COMMAND = "Run Command",
+  COMMAND_RESULT = "Run Command Result",
```
eysi09, line 11: This could also just be called "Command Result" imo. Slightly easier to read on Amplitude.

core/src/analytics/analytics-types.ts
```diff
@@ -8,8 +8,14 @@
...
   CALL_API = "Call API",
   MODULE_CONFIG_ERROR = "Module Configuration Error",
   PROJECT_CONFIG_ERROR = "Project Configuration Error",
   VALIDATION_ERROR = "Validation Error",
 }
+
+export enum AnalyticsCommandResult {
```
eysi09, line 18: Can we make this a union type instead of a union? In general we should prefer unions.

core/src/analytics/analytics.ts
```diff
@@ -131,6 +132,16 @@ interface ApiEvent extends EventBase {
   }
 }
 
+interface CommandResultEvent extends EventBase {
+  type: AnalyticsType.COMMAND_RESULT
+  properties: PropertiesBase & {
+    name: string
+    durationMsec: number
```
eysi09, line 139: Nice!

eysi09 (changes requested):
Nice!

mkhq:
These events will only be sent if the analytics handler has been initialized, which in turn depends on a garden instance. We would need a refactor of the init flow to make sure we can catch all potential error cases.

eysi09 (approved):

PR #3856, opened 2023-02-27
Create tar.gz archive container images, which need to be in the Kuber…

OhMyMndy:
…netes cluser, as Github artifacts

Add a `tar.gz` archive with all the container images which need to be in the Kubernetes cluster, so air gapped users can import these container images into their local registry.

Fixes #


vvagaytsev:
Thanks for the PR, @OhMyMndy! Unfortunately, it seems to be out of scope for us. The script can be used locally and does not seem to be a part of a regular Garden CI.

PR #3858, opened 2023-02-28
0.13: module conversion + integ tests

Orzelius:
please make your way to the commits :)

### DIFF
--- a/core/test/integ/src/plugins/kubernetes/kubernetes-module/handlers.ts
+++ b/core/test/integ/src/plugins/kubernetes/kubernetes-module/handlers.ts
@@ -213,9 +213,9 @@ describe("kubernetes-module handlers", () => {
       await kubernetesDeploy(deployParams)
       const res3 = await findDeployedResources(manifests, log)
 
-      expect(res1[0].metadata.annotations![gardenAnnotationKey("sync-mode")]).to.equal("false")
+      expect(res1[0].metadata.annotations![gardenAnnotationKey("sync-mode")]).to.be.undefined
       expect(res2[0].metadata.annotations![gardenAnnotationKey("sync-mode")]).to.equal("true")
-      expect(res3[0].metadata.annotations![gardenAnnotationKey("sync-mode")]).to.equal("false")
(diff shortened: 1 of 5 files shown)


core/test/integ/src/plugins/kubernetes/kubernetes-module/handlers.ts
```diff
@@ -213,9 +213,9 @@ describe("kubernetes-module handlers", () => {
       await kubernetesDeploy(deployParams)
       const res3 = await findDeployedResources(manifests, log)
 
-      expect(res1[0].metadata.annotations![gardenAnnotationKey("sync-mode")]).to.equal("false")
+      expect(res1[0].metadata.annotations![gardenAnnotationKey("sync-mode")]).to.be.undefined
       expect(res2[0].metadata.annotations![gardenAnnotationKey("sync-mode")]).to.equal("true")
-      expect(res3[0].metadata.annotations![gardenAnnotationKey("sync-mode")]).to.equal("false")
+      expect(res3[0].metadata.annotations![gardenAnnotationKey("sync-mode")]).to.be.undefined
```
Orzelius, line 218: for some reason these are just not set instead of false. Probably should look into that at one point

Orzelius:
Need to wait for the ci here as the action name logic most likely will break some unit tests

Orzelius:
Looks like all tests passed, requesting reviews

core/test/integ/src/plugins/kubernetes/kubernetes-module/handlers.ts
Walther: should we use the `isConfiguredForSyncMode` function in these tests to get boolean results instead of querying the annotations manually?

Walther (approved):
one nit, otherwise looks good to me!

the todo item looks important (infinite loop in solver) and we need to remember to get back to it at some point

core/test/integ/src/plugins/kubernetes/kubernetes-module/handlers.ts
Orzelius: I don't know if the missing annotation is an intentional change or on not so I can't say for certain.

PR #3859, opened 2023-02-28
chore(cloud): update action status & log events

thsig:

Here, we udpate and simplify the action status event payload types used to communicate with Cloud.

The event payload types have been streamlined and standardized. We now use `actionState` as the primary state indicator field, which should reduce the special-casing needed on the backend (since we can interpret a common sum type for status events from all the main action kinds).

The action states have also been documented thoroughly, which should help newcomers understand the larger flow more easily.

We now also generate action UIDs for individual task instances, enabling us to use the same UID for the status handler call and the processing handler call.

This way, the API can correlate a status check with its subsequent processing call (if any).

We'll be able to utilize this in the Cloud UI e.g.  for flame charts that capture the full lifecycle of the framework's primary action kinds.

### DIFF
--- a/core/src/actions/types.ts
+++ b/core/src/actions/types.ts
@@ -98,8 +98,23 @@ export interface ActionConfigTypes {
   Test: TestActionConfig<string, any>
 }
 
-// See https://melvingeorge.me/blog/convert-array-into-string-literal-union-type-typescript
-export const actionStateTypes = ["getting-status", "ready", "not-ready", "processing", "failed", "unknown"] as const
+/**
+ * These are the states returned from actions.
+ *
+ * See https://melvingeorge.me/blog/convert-array-into-string-literal-union-type-typescript
+ */
+export const actionStateTypes = ["ready", "not-ready", "processing", "failed", "unknown"] as const
+export type ActionState = typeof actionStateTypes[number]
+
+/**
+ * These are the states emitted in status events. Here, we include additional states to help distinguish status event
+ * emitted around status/cache checks VS statuses emitted around the execution after a failed status check.
+ */
+export const actionStateTypesForEvent = [
+  ...actionStateTypes,
+  "getting-status",
+  "cached",
+] as const
 /**
  * This type represents the lifecycle of an individual action execution.
  *
@@ -110,7 +125,7 @@ export const actionStateTypes = ["getting-status", "ready", "not-ready", "proces
  * initial state: "getting-status"
  * final state: "ready" or "failed"
  *
- * "getting-status" -> "ready" | "not-ready"
+ * "getting-status" -> "cached" | "not-ready"
  * "not-ready" -> "processing"
  * "processing" -> "ready" | "failed"
  * ```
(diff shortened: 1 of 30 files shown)


core/src/actions/types.ts
```diff
@@ -99,7 +99,52 @@ export interface ActionConfigTypes {
...
+ *     build exists for the requested action's version.
+ *   - Or for a Kubernetes deployment, this might involve checking if the requested resources are already live and up
+ *     to date.
+ *
+ * - `"ready"`: An up-to-date result exists for the action.
+ *   - This state can be reached by a status check that returned a successful cache hit (e.g. an up-to-date build
+ *     artifact / deployed resource / test result / run result already exists), or by successfully processing the
+ *     action after getting a `"not-ready"` state from the status check.
+ *
+ * - `"not-ready"`: Only used for actions that have a health state associated with them (currently only Deploys).
```
mkhq, line 132: Reading through the text, it seems that "outdated" and "not-ready" are some form of pre-conditions for the processing that are not yet fulfilled. Is it then a responsibility of the action to ensure the pre-conditions are met or whoever calls the action? E.g. "not-ready" or "outdated" could be a failure state which leads to a retry when the resources are provisioned. A simplified state machine would then be:

(start) -> "getting-status" -> "ready" | "processing"
"processing" -> "ready" | "failed"

We could also introduce a state "provisioning" (or some better term) that can have the status not-ready or outdated. This could be a no-op, e.g. for the build action, but would be needed for deploy.

(start) -> "getting-status" -> "ready" | "provisioning" | "failed" (anything can fail?)
"provisioning" -> "processing" | "failed"
"processing" -> "ready" | "failed"

thsig: How do we feel about the action states and their descriptions now?

I've gotten rid of the `outdated` action state in favour of the `not-ready` state (we still have the `outdated` state on deploy statuses, which is sometimes used internally by plugins).

core/src/router/build.ts
```diff
@@ -65,34 +73,33 @@ export const buildRouter = (baseParams: BaseRouterParams) =>
         garden.events.emit("log", {
```
stefreak, line 73: The `// TODO: consider sending origin as well` should be removed as well (the todo exists in all the different action routers build, deploy, run and test)

core/src/events.ts
```diff
@@ -60,27 +61,12 @@ export class EventBus extends EventEmitter2 {
...
-  /**
-   * ISO format date string
-   */
-  deployStartedAt?: string
-  /**
-   * ISO format date string
-   */
-  deployCompletedAt?: string
-}
+export type DeployStatusForEventPayload = Omit<ServiceStatus, "detail">
```
10ko, line 69: Wouldn't be cleaner to refactor the `ServiceStatus` to be renamed `DeployStatus` and get rid of the `detail` fieald?

core/src/events.ts
```diff
@@ -116,10 +102,21 @@ export function toGraphResultEventPayload(result: GraphResult): GraphResultEvent
   return payload
 }
 
+export interface ActionStatusPayload<S = {}> {
```
10ko, line 105: Should this be a union type insted of `{}`?

core/src/events.ts
thsig: That's true, but I suggest we leave that for another PR—there's a bit of internal terminology cleanup that still remains to be done in Core.

core/src/events.ts
thsig: We could do that, but IMHO we also want to keep this open to new action kinds.

core/src/events.ts
```diff
@@ -212,110 +203,48 @@ export interface Events extends LoggerEvents {
     completedAt: string
   }
   watchingForChanges: {}
+  /**
+   * Line-by-line action log events. These are emitted by the `PluginEventBroker` instance passed to action handlers.
+   *
+   * This is in contrast with the `logEntry` event below, which represents framework-level logs emitted by the logger.
+   *
+   * TODO: Instead of having two event types (`log` and `logEntry`), we may want to unify the two.
```
eysi09, line 211: Yes, in my opinion we should drop this, drop the corresponding task log table in Cloud, and ensure the actual log entries have the metadata needed to tease apart different kinds of logs.

eysi09 (approved):

PR #3860, opened 2023-03-01
chore(examples): update remote-k8s example to use actions

Walther:

Fixes https://github.com/garden-io/garden/issues/3846

I have now tested this enough with a GKE cluster to be personally comfortable with merging this.
Deploying and remote builds work, and the services are accessible via a custom hostname - with some not-committed, local changes required to the configuration, as per instructions.

I have also removed the mention of TLS configuration in the example, to keep this focused on the remote builds and deployments. We have other examples for `cert-manager` use, and using a load balancer for TLS termination is another viable choice for production workloads - either of which should be better than manually setting kube secrets.

Squashing before merging.


vvagaytsev (approved):
Nice! Thank you 👍

PR #3866, opened 2023-03-02
fix(core): remove locks from ConfigStore

thsig:

The lockfile logic in the new `ConfigStore` class became unnecessary after the introduction of the `write-file-atomic` library, which implements its own atomic file writes.

We had sometimes run into errors around the lockfile logic during testing, which prompted us to remove it.


edvald:
Need to close this, the lock is actually important for other reasons.

PR #3868, opened 2023-03-02
Edge bonsai docker release 0.13

worldofgeese:
What this PR does / why we need it:

For testing edge-bonsai it would be good to build and release container images for use where CI depends on a container image e.g. GitLab CI. Users sourcing the Garden binary as part of a multi-stage Dockerfile strategy should also be pleased with the option to have an edge release available.

PR #3872, opened 2023-03-02
fix: edge-bonsai-docker-release-0.13

worldofgeese:
Fixes #3868 by increasing filter to include `build-dist-edge-bonsai`. From [CircleCI's reference](https://circleci.com/docs/configuration-reference/#requires):

> A list of jobs that must succeed for the job to start. Note: When jobs in the current workflow that are listed as dependencies are not executed (due to a filter function for example), their requirement as a dependency for other jobs will be ignored by the requires option. However, if all dependencies of a job are filtered, then that job will not be executed either.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1162,7 +1162,6 @@ workflows:
               only:
                 - main
                 - "0.13"
-              ignore: /.*/
 
   tags:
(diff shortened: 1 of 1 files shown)


Walther (approved):
🎉

PR #3874, opened 2023-03-02
ci: clean-up docker tags

worldofgeese:
This stops git hashes from proliferating in our Docker Hub by removing any step that pushes a git hash. Building is already taken care of by our push-containers.sh scripts which itself calls build-containers.sh.

We want to move to images for clouds (AWS, GCP, etc.) as an image tag for 0.13.  This does not address that. This also tests using `latest` which is not in any way ideal when any number of image builds could be competing for the tag.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -417,10 +417,14 @@ jobs:
             if [ "$CIRCLE_BRANCH" == "0.13" ]; then
               ./scripts/push-containers.sh bonsai-edge
             fi
+            # Get remote image digest of pushed container
+            DIGEST=$(docker inspect --format='{{.RepoDigests}}' gardendev/garden:bonsai-edge-buster | tr -d '[]')
+            echo "export DIGEST=$DIGEST" >> $BASH_ENV
 
   test-docker-gcloud:
     docker:
-      - image: gardendev/garden-gcloud:latest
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -397,114 +397,30 @@ jobs:
...
-          name: Build image and push to registry
+      # TODO: use garden publish here
+      - deploy:
+          name: Release docker images
           command: |
-            TAG=gardendev/garden:${CIRCLE_SHA1}-buster
-            docker build -t ${TAG} -f support/buster.Dockerfile dist/linux-amd64
-            docker push ${TAG}
+            # Push with latest tag for non-pre-release tags
+            if [[ "$TAG" == "$CIRCLE_TAG" ]] && [[ $VERSION != *"-"* ]]; then
```
Walther, line 413: assuming we want to keep `latest` / `edge` based on the actual `main` branch, we probably need to prevent this from running on the 0.13 branch, so that we don't end up with `latest` images built with 0.13 code?

worldofgeese: The last `latest` was pushed two days ago: this conditional appear to already filter out 0.13 releases

Walther (approved):
Looks good to me as far as I can tell 👍

PR #3879, opened 2023-03-03
cherry-pick for 0.13: fix(framework): omit detail property of errors

vvagaytsev:
This cherry-picks https://github.com/garden-io/garden/commit/24f32fad to `0.13`.

Fixes #

### DIFF
--- a/core/src/events.ts
+++ b/core/src/events.ts
@@ -99,7 +99,7 @@ export interface CommandInfoPayload extends CommandInfo {
 export function toGraphResultEventPayload(result: GraphResult): GraphResultEventPayload {
   const payload = sanitizeValue({
     ...omit(result, "dependencyResults", "task"),
-    error: result.error ? String(result.error) : null,
+    error: result.error ? String(omit(result.error, "detail")) : null,
   })
   if (payload.result) {
     // TODO: Use a combined blacklist of fields from all task types instead of hardcoding here.


core/src/events.ts
```diff
@@ -99,7 +99,7 @@ export interface CommandInfoPayload extends CommandInfo {
 export function toGraphResultEventPayload(result: GraphResult): GraphResultEventPayload {
   const payload = sanitizeValue({
     ...omit(result, "dependencyResults", "task"),
-    error: result.error ? String(result.error) : null,
+    error: result.error ? String(omit(result.error, "detail")) : null,
```
stefreak, line 102: I think with `String()` the omit of detail is not necessary anymore.

```
String(new Error("foo"))
'Error: foo'
```

stefreak (changes requested):

stefreak:
@vvagaytsev I think we can close this PR

core/src/events.ts
stefreak: In my experiments, for an object like `String({})` yields `[Object object]` and `String(new Error("message"))` yields `Error: message`

stefreak: In the original commit, it was assigned directly without `String`: https://github.com/garden-io/garden/pull/3623/files – so the change became obsolete

vvagaytsev: Ok, thanks!

PR #3884, opened 2023-03-03
improvement(container): redeploy when spec changes

thsig:

In 0.12, we had wanted to update the `selector.matchLabels` used for `container` deploys, but doing so would require redeploying those resources (since `spec.selector.matchLabels` is an immutable field).

In 0.13, we decided to simply automatically redeploy any `container` deploys if needed (unless the environment in question is a production environment), since we wanted to use different labels (e.g. `action` instead of `service`).

Deleting the resources before redeploying them is necessary in this case, since the `spec.selector` field is immutable and attempting to update it for running resources will simply result in an error).

We don't automatically do this for production environments, except when force-deploying (which is a framework-native way for users to redeploy services with full awareness what they're doing).

Also fixed a bug in the `persistentvolumeclaim` implementation where an unresolved action was returned instead of the resolved one (which led to errors at runtime during deployment).

### DIFF
--- a/core/src/plugins/kubernetes/container/deployment.ts
+++ b/core/src/plugins/kubernetes/container/deployment.ts
@@ -13,7 +13,7 @@ import { ContainerDeployAction, ContainerDeploySpec, ContainerVolumeSpec } from 
 import { createIngressResources } from "./ingress"
 import { createServiceResources } from "./service"
 import { waitForResources } from "../status/status"
-import { apply, deleteObjectsBySelector, KUBECTL_DEFAULT_TIMEOUT } from "../kubectl"
+import { apply, deleteObjectsBySelector, deleteResourceKeys, KUBECTL_DEFAULT_TIMEOUT } from "../kubectl"
 import { getAppNamespace, getAppNamespaceStatus } from "../namespace"
 import { PluginContext } from "../../../plugin-context"
 import { KubeApi } from "../api"
@@ -29,7 +29,7 @@ import { getDeployedImageId, getResourceRequirements, getSecurityContext } from 
 import { configureLocalMode, startServiceInLocalMode } from "../local-mode"
 import { DeployActionHandler, DeployActionParams } from "../../../plugin/action-types"
 import { Resolved } from "../../../actions/types"
-import { ConfigurationError } from "../../../exceptions"
+import { ConfigurationError, DeploymentError } from "../../../exceptions"
 import {
   SyncableKind,
   syncableKinds,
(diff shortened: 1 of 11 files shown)


core/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -49,14 +49,27 @@ export const DEFAULT_MINIMUM_REPLICAS = 1
...
+  const { ctx, action, log, syncMode, localMode, force } = params
   const k8sCtx = <KubernetesPluginContext>ctx
   const { deploymentStrategy } = k8sCtx.provider.config
   const deployWithSyncMode = syncMode && !!action.getSpec("sync")
   const api = await KubeApi.factory(log, k8sCtx, k8sCtx.provider)
 
   const imageId = getDeployedImageId(action, k8sCtx.provider)
 
+  const legacyCheckStatus = await k8sGetContainerDeployStatus(params)
+  if (legacyCheckStatus.detail?.detail.deployedWithLegacySelector) {
```
edvald, line 61: This seems a little overly specific... Can't we just compare the selectors/labels and make this a more generic check?

edvald:
Just one comment, otherwise looks good.

core/src/plugins/kubernetes/container/deployment.ts
thsig: Yeah, I like that idea.

My intent here was to specifically handle these Garden-specific keys, but this behaviour (i.e. delete & redeploy unless in a prod environment and not force-deploying) is a better strategy for immutable field updates in general.

thsig: Ok, this should be good to go now.

core/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -42,38 +42,51 @@ import { k8sGetContainerDeployStatus, ContainerServiceStatus } from "./status"
 import { emitNonRepeatableWarning } from "../../../warnings"
 
 export const DEFAULT_CPU_REQUEST = "10m"
-export const DEFAULT_MEMORY_REQUEST = "90Mi" // This is the minimum in some clusters
+export const DEFAULT_MEMORY_REQUEST = "90Mi" // This is the minimum in some clusters - or so they tell me
```
edvald, line 45: Haha

core/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -743,3 +756,50 @@ export const deleteContainerDeploy: DeployActionHandler<"delete", ContainerDeplo
...
+        deployName: action.name,
+      }
+    )
+  } else {
+    if (production && force) {
+      log.warn(chalk.yellow(`${msgPrefix} Since we're deploying with force = true, we'll now delete it before redeploying.`))
+    } else if (!production) {
+      log.warn(chalk.yellow(`${msgPrefix} Since this environment does not have production = true, we'll now delete it before redeploying.`))
+    }
+    await deleteObjectsBySelector({
```
edvald, line 795: This could in some cases delete more resources than intended

thsig: Yeah, good point. How about changing the flow to call `deleteResources` on the resources that have changed selectors?

That should only delete the minimal required set of resources.

edvald: That would make sense, yes.

thsig: Should be good to go now.

I noticed a bug/limitation in one place (the logic wouldn't get triggered if one or more resources passed to `compareDeployedResources` wasn't already deployed), and improved the tests a bit around this too.

core/src/plugins/kubernetes/container/deployment.ts
thsig: ... I sort of don't want to delete this now. Haha.

core/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -42,38 +42,53 @@ import { k8sGetContainerDeployStatus, ContainerServiceStatus } from "./status"
...
+  const { ctx, action, log, syncMode, localMode, force } = params
   const k8sCtx = <KubernetesPluginContext>ctx
   const { deploymentStrategy } = k8sCtx.provider.config
   const deployWithSyncMode = syncMode && !!action.getSpec("sync")
   const api = await KubeApi.factory(log, k8sCtx, k8sCtx.provider)
 
   const imageId = getDeployedImageId(action, k8sCtx.provider)
 
+  const status = (await k8sGetContainerDeployStatus(params))!
+  const specChangedResourceKeys: string[] = (status.detail?.detail.selectorChangedResourceKeys) || []
```
thsig, line 61: It's a bit awkward that the type of the detail (which is in fact `ContainerServiceStatusDetail`) isn't inferred, but I think we'd have to add a type parameter for the detail type (in addition to the outputs) to do that in a framework-generic way.

Probably out of scope for this PR, but did we make a deliberate choice not to include typing for the deploy status detail in `DeployStatus`?

edvald: It's just something that would take a bit more figuring out in general. We should at least leave it out of scope for this particular PR.

Orzelius (approved):
lgtm. I like the clean integ tests

Issue #3841, opened 2023-02-24
Configure support for depending on pure ESM packages


Walther:
A lot of packages in the Node ecosystem have moved to being pure ECMAScript modules (ESM)
Our current build configurations and scripts do not support depending on ESM.
This is blocking us from updating a lot of our dependencies.

Blocked by `pkg` https://github.com/vercel/pkg/issues/1291

Walther:
@stefreak i have a list of examples mentioned here: https://github.com/garden-io/garden/pull/3838

stefreak:
Still causing issues for us, see also #5182

PR #3838, opened 2023-02-23
chore: update dependencies and remove unused ones

Walther:
**What this PR does / why we need it**:

Did a bit of spring cleaning ✨ 🧹
Used `depcheck` and `ncu` to help with cleanup and updates 

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

Squash before merging.

## Notes

- updating `platform-api-types` skipped, very outdated and causes type errors
- updating `typescript` skipped, newer versions are stricter and catch new errors in existing code
- updating `koa` and its adjacent packages skipped, causes lots of type errors
- updating `testdouble` skipped, results in some type errors
- updating `js-yaml` skipped, large changelog and migration guide
- updating `type-fest` skipped, medium changelog
- updating `yaml-lint` skipped, results in some type errors
- updating `nock` skipped, breaks analytics tests
- a bunch of dependencies have moved to being pure ESM packages, we need to add support for those in order to update them: (https://github.com/garden-io/garden/issues/3841)
  - `chalk`
  - `cli-truncate`
  - `cpy`
  - `crypto-random-string`
  - `execa`
  - `get-port`
  - `glob`
  - `got`
  - `has-ansi`
  - `humanize-string`
  - `indent-string`
  - `inquirer`
  - `log-symbols`
  - `mimimc-fn`
  - `normalize-url`
  - `p-event`
  - `p-retry`
  - `slash`
  - `string-width`
  - `strip-ansi`
  - `tempy`
  - `titleize`
  - `wrap-ansi`
  - `username`

PR #3885, opened 2023-03-03
ci: add dependabot docker and npm ecosystems

stefreak:
This means more dependabot PRs, and hopefully less manual work keeping Dockerfiles up to date.

In some cases careful testing will be required before merging Docker PRs but at least we have a to-do list in the form of dependabot PRs.

With the npm package ecosystem, dependabot will also start suggesting more NPM updates, and not just the security updates (that's why I limited the number of them until 0.13 has landed).

Fixes #

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,9 +5,21 @@
 
 version: 2
 updates:
-  # Maintain dependencies for GitHub Actions
+  # Automatically update GitHub Actions
   - package-ecosystem: "github-actions"
     directory: "/"
-    open-pull-requests-limit: 5
+    open-pull-requests-limit: 10
+    schedule:
+      interval: "weekly"
+  # Automatically update Dockerfile FROM images
+  - package-ecosystem: "docker"
+    directory: "/"
+    open-pull-requests-limit: 10
+    schedule:
+      interval: "daily"
+  # Automatically update NPM packages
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -5,9 +5,21 @@
...
+  # Automatically update Dockerfile FROM images
+  - package-ecosystem: "docker"
+    directory: "/"
+    open-pull-requests-limit: 10
+    schedule:
+      interval: "daily"
+  # Automatically update NPM packages
+  - package-ecosystem: "npm"
+    directory: "/"
+    open-pull-requests-limit: 3 # for now, once 0.13 landed we can increase this
```
Walther, line 23: I'm slightly hesitant on adding the npm updates right now, in part because of the earlier work on https://github.com/garden-io/garden/pull/3838/ and discovering the need to configure our typescript project and its build process to support consuming ECMAScript Modules https://github.com/garden-io/garden/issues/3841. Currently, we are blocked by this on updating a lot of dependencies, we can't use "pure ESM build" dependencies.

stefreak:
```suggestion
```

stefreak: OK, I have commented out the npm ecosystem for now and added a comment that mentions the #3841 ESM packaging issue

PR #3886, opened 2023-03-06
0.13 cherry pick release 0.12.49

vvagaytsev:

Fixes #

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@garden-io/cli",
-  "version": "0.12.48",
+  "version": "0.12.49",
   "description": "Cloud native testing platform for testing and developing container applications on Kubernetes",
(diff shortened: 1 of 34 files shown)


cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@garden-io/cli",
-  "version": "0.12.48",
+  "version": "0.12.49",
```
Walther, line 3: sidenote: we should probably figure out a way for the `garden version` command to return `0.13.x` when running on a version built from that branch. currently, it is difficult to distinguish two garden binaries (or docker images) because both return `0.12.x`

Walther (approved):

cli/package.json
vvagaytsev: Yes, we should do that 👍

PR #3892, opened 2023-03-06
chore: tag packages with version 0.13.0-development

Walther:

Currently, it is difficult to distinguish between two `garden` binaries or containers, as `garden version` will return `0.12.x`.
With this PR, any binary or container built from the 0.13 branch will report the correct version string.

Fixes #

Because `0.13.0-development` is a pre-release, we need to explicitly depend on the pre-release versions in all of our internal dependencies - otherwise `yarn` will not find the packages in the workspace. We should revert this back to the `"*"` version before merging to `main`.


vvagaytsev (changes requested):
Looks good! Please check the [CI check-docs step](https://app.circleci.com/pipelines/github/garden-io/garden/15850/workflows/0977366c-9d30-45a7-9187-4b01c5dd5570/jobs/264751).

It fails saying "generated docs are not up-to-date!".

vvagaytsev:
@Walther can we just use `0.13` as a version? Then links will work because we do have `0.13` branch.

Walther:
Unfortunately that won't work either, as `0.13` is an invalid version for `yarn` (and we cannot use `0.13.0` either as we don't have a tagged release for that).

I'm thinking maybe we should just drop this PR, the 0.13 branch is temporary anyway and we are getting closer and closer to release.

vvagaytsev:
Ok, let's just drop this.

PR #3893, opened 2023-03-06
fix(docs): fix broken links

thsig:

This commit was cherry-picked from `main` to `0.13`, and the resulting conflicts resolved.

This should fix the dead links in the `check-docs` step on 0.13.


vvagaytsev:
Let's merge this after the rest cherry-pick PRs. I'm trying to avoid reordering commits and to keep the `main` and the `0.13` history aligned :)

vvagaytsev (approved):
Thanks for cherry-picking this! 👍

PR #3896, opened 2023-03-07
0.13: a few fixes + remaining k8s run integ tests fixed

Orzelius:
commits contain the keys to the kingdom


Walther (approved):
looks good to me! 🚀  thanks for the good work <3
should we also enable some of the minkube and microk8s tests in CI now?

vvagaytsev (approved):

Orzelius:
> should we also enable some of the minkube and microk8s tests in CI now?

plan to do it by the end of the week. As of right now a big part is still failing but by Friday hopefully I'll have most passing and the rest can be temporarily skipped for a green ci.

PR #3907, opened 2023-03-07
feat(config): add $if/$else/$then conditional objects

edvald:
This is analogous to `$forEach` objects, and addresses a common need (that I happened to run into while working on something else).

From the docs:

---

### If/else conditional objects

You can conditionally set values by specifying an object with `$if`, `$then` and (optionally) `$else` keys. This can in many cases be clearer and easier to work with, compared to specifying values within conditional template strings.

Here's an example:

```yaml
spec:
  command:
    $if: ${this.mode == "sync"}
    $then: [npm, run, watch]
    $else: [npm, start]
```

This sets `spec.command` to `[npm, run, watch]` when the action is in sync mode, otherwise to `[npm, start]`.

You can also skip the `$else` key to default the conditional to _no value_ (i.e. undefined).

---


Orzelius (approved):
Looks good but I wonder what's the use case? I kinda like how the existing ternary operator forces to store objects in variables. Makes it easier to read than a large if else block. Also reminds of terraform which afaik [still only has the short conditional expression](https://developer.hashicorp.com/terraform/language/expressions/conditionals).

thsig:
The use case is to make dynamism on the structural level (i.e. YAML arrays and dictionaries) easier in general.

Variables are still a good idea, but for one-off bits of conditional logic, it's often more readable to just do them in-place (instead of pulling them into a variable).

PR #3908, opened 2023-03-08
fix(core): override Build version with module

thsig:

For `BuildAction`s generated by converting modules, we now override the `versionString` of the full version of `Build` actions with the `versionString` of their parent module.

This fixes an issue where e.g. `helm` modules that referred to `${modules.some-container-module.version}` would fail to deploy, since the version of the `Build` action that `container-module` would get converted to wouldn't match that of the `container` module that it was converted from—this resulted in a different version being built than the Helm chart expected, leading to image pull errors.

The same pattern is also standard when using `kubernetes` and `container` modules in conjunction.

This should ease the transition from 0.12 to 0.13 for many projects.

### DIFF
--- a/core/src/actions/build.ts
+++ b/core/src/actions/build.ts
@@ -36,6 +36,8 @@ import {
   ExecutedActionExtension,
 } from "./base"
 import { ResolvedConfigGraph } from "../graph/config-graph"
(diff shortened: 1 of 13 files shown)


edvald (approved):
Nice!

PR #3912, opened 2023-03-08
docs: add template string reference docs for actions

edvald:
Found that this was missing after the refactors.


trymbill (approved):
LGTM 👍

PR #3917, opened 2023-03-09
chore: fix result filtering logic

thsig:

This fixes an error that would come up when rendering command results for failed commands when the command didn't produce any result.

This is a follow-up/fix to https://github.com/garden-io/garden/pull/3911.


Walther (approved):
always enjoy a good early return! ✨ 🧹

PR #3921, opened 2023-03-10
chore: misc fixes for cloud

thsig:

**fix(container): fix spec.image in container conversion**

When `module.spec.image` is provided, but the value has no tag, we weren't setting the Build/module version as the default tag/suffix to the image ID.

In 0.12, this was done by referencing the deployment image ID from the container build's outputs, which we can't do here. Therefore, we simply call the relevant helper to shim that behaviour here.

**fix: schema fixes around k8s/container conversion**

The `spec.targetImage` field was called `spec.image` on `containerBuildSchema` (i.e. the interface and the schema didn't agree).

Also fixed the spec conversion logic for `kubernetes` modules.

**chore: set default Cloud API auth token validity**

### DIFF
--- a/core/src/cloud/api.ts
+++ b/core/src/cloud/api.ts
@@ -261,10 +261,11 @@ export class CloudApi {
       throw new EnterpriseApiError(errMsg, { tokenResponse })
     }
     try {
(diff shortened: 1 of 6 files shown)


Walther (approved):
🚀

PR #3934, opened 2023-03-13
0.13 enable timeout for hadolint plugin

vvagaytsev:
This enables `timeout` field in the `convert` handler of the `hadolint` plugin.

Fixes #

The original error message was confusing:
```
Error: 
Error parsing config for Test action foo:
Error validating configure handler output from provider hadolint for Test type hadolint: key "timeout" is not allowed at path .config[timeout]. Available keys: config)
    at garden/core/src/graph/actions.ts:108:13
```

It should be fixed as well. Looks like there are some issues in the error description rendering in the [`validateSchema` function](https://github.com/garden-io/garden/blob/0.13/core/src/config/validation.ts#L84). It does not expand the existing `config` field and does not print the correct available keys.

### DIFF
--- a/core/src/actions/deploy.ts
+++ b/core/src/actions/deploy.ts
@@ -16,9 +16,7 @@ import {
 import { Action, BaseActionConfig } from "./types"
 
 export interface DeployActionConfig<N extends string = any, S extends object = any>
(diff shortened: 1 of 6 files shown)


Orzelius (approved):
🌻

PR #3945, opened 2023-03-14
0.13 some refactoring and minor fixes for local mode

vvagaytsev:
This PR **does not** fully fix the local mode integration tests. It refactors the local mode implementation, making it aligned with sync mode code, more readable, and easier to maintain. It also fixes some minor issues, and unifies the way of local mode config and start for all supported providers.

See individual commits for details. The commits be squashed before the merge. 

Fixes #

The local mode works well. Most likely, there is something wrong with the test implementation. That does not block the `0.13` release, and could be fixed in a separate PR.

### DIFF
--- a/core/src/plugins/kubernetes/container/deployment.ts
+++ b/core/src/plugins/kubernetes/container/deployment.ts
@@ -277,6 +277,11 @@ interface CreateDeploymentParams {
   production: boolean
 }
 
+const getDefaultWorkloadTarget = (w: KubernetesResource<V1Deployment | V1DaemonSet>) => ({
+  kind: <SyncableKind>w.kind,
+  name: w.metadata.name,
+})
+
 export async function createWorkloadManifest({
   ctx,
   api,
@@ -493,12 +498,10 @@ export async function createWorkloadManifest({
 
   // Local mode always takes precedence over sync mode
   if (mode === "local" && localModeSpec) {
-    const target = { kind: <SyncableKind>workload.kind, name: workload.metadata.name }
-
     const configured = await configureLocalMode({
       ctx,
       spec: localModeSpec,
-      defaultTarget: target,
+      defaultTarget: getDefaultWorkloadTarget(workload),
       manifest: workload,
       manifests: [workload],
       action,
@@ -508,15 +511,12 @@ export async function createWorkloadManifest({
     workload = <KubernetesResource<V1Deployment | V1DaemonSet>>configured.updated
   } else if (mode === "sync" && syncSpec) {
     log.debug({ section: action.key(), msg: chalk.gray(`-> Configuring in sync mode`) })
-
-    const target = { kind: <SyncableKind>workload.kind, name: workload.metadata.name }
-
     const configured = await configureSyncMode({
       ctx,
       log,
       provider,
       action,
-      defaultTarget: target,
+      defaultTarget: getDefaultWorkloadTarget(workload),
       manifests: [workload],
       spec: syncSpec,
     })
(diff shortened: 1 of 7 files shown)


Orzelius:
While you're at it could you take a look at these two integ tests
1. https://github.com/garden-io/garden/blob/0.13/core/test/integ/src/plugins/kubernetes/helm/deployment.ts#L62
failing, probably because the "deploy is not configured for or does not support local mode" as the console warns
2. https://github.com/garden-io/garden/blob/0.13/core/test/integ/src/plugins/kubernetes/local-mode.ts

vvagaytsev:
> While you're at it could you take a look at these two integ tests
> 
> 1. https://github.com/garden-io/garden/blob/0.13/core/test/integ/src/plugins/kubernetes/helm/deployment.ts#L62
>    failing, probably because the "deploy is not configured for or does not support local mode" as the console warns
> 2. https://github.com/garden-io/garden/blob/0.13/core/test/integ/src/plugins/kubernetes/local-mode.ts

Sure, will do! Now I'm fixing some bugs and implementing missing parts. The fixes will likely address both failures.

core/src/plugins/kubernetes/local-mode.ts
```diff
@@ -49,6 +49,20 @@ export interface KubernetesLocalModeSpec extends ContainerLocalModeSpec {
   target?: KubernetesTargetResourceSpec
 }
 
+export function convertContainerLocalModeSpec(
+  ctx: KubernetesPluginContext,
+  action: Resolved<ContainerDeployAction>
+): KubernetesLocalModeSpec | undefined {
```
Orzelius, line 55: Why is this needed? Is there any difference in kubernetes and container local mode spec? If so wouldn't it be easier to handle this via a base type instead of this conversion.

core/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -188,7 +188,7 @@ export async function startLocalMode({
   await startServiceInLocalMode({
     ctx,
     spec: localModeSpec,
-    targetResource,
+    manifest: targetResource,
```
Orzelius, line 187: What's the motivation for this change? Isn't this parameter still describing the k8s resource that local mode will be "targeting"? For me, manifest implies that something is created based or from it.

Orzelius (approved):
Just two questions

Orzelius:
Also, thanks for the clear commits

core/src/plugins/kubernetes/container/deployment.ts
vvagaytsev: Yes, good point! I was thinking about the same :) Initially, it was renamed to be replaced by a tuple of `manifests` and `defaultTarget` fields in `ConfigureLocalModeParams`. But in `StartLocalModeParams` we still to specify a single target resource explicitly.

Renamed `manifest` back to `targetResource` in https://github.com/garden-io/garden/pull/3945/commits/9738db2478dc626e82d2c6bfdb0a1dec75ac8deb.

It will b changed later when we allow multiple resource configurations in local mode.

core/src/plugins/kubernetes/local-mode.ts
vvagaytsev: This may look unnecessary. Type `KubernetesLocalModeSpec` actually extends `ContainerLocalModeSpec` by adding one optional field `target?: KubernetesTargetResourceSpec`.

So, technically, it is possible to use the base type `ContainerLocalModeSpec` in https://github.com/garden-io/garden/blob/0.13-local-mode-fixes/core/src/plugins/kubernetes/container/deployment.ts#L503
The code compiles.

But, if we add any non-optional fields in `KubernetesLocalModeSpec` or decide to make `target` mandatory, the same code won't compile.

So, despite its triviality, I decided to extract a utility method to do the conversion logic.

PR #3944, opened 2023-03-14
0.13: fix all integ tests

Orzelius:
### ₍꒢  ̣̮꒢₎

We can actually close this once merged https://github.com/garden-io/garden/issues/3113

Probably need to open a separate issue for all the tests skipped or the ones yet to be written.

PR #3947, opened 2023-03-14
improvement(k8s): schedule runners w/o deploying

thsig:

This overcomes a limitation that had been temporarily introduced in 0.13, where the `helm` or `kubernetes` Deploy that a `kubernetes-pod` Run or Test targeted would have to be deployed before the Test/Run could be executed. This was essentially a semantic regression from 0.12.

This was fixed/improved by simply providing manifests to `kubernetes-pod` Runs and Tests in an equivalent way to how `kubernetes` Deploys are configured.

Helm Runs and Tests are now implemented with a new `helm-pod` action definition, which is conceptually similar to the new `kubernetes-pod` implementation, but which render the manifests from a Helm chart instead.

Once the `kubernetes-exec` Run and Test action definitions have been implemented, those will be the generally recommended option unless the semantics of the `kubernetes-pod` or `helm-pod` action types are specifically required.


Orzelius:
Could this wait until https://github.com/garden-io/garden/pull/3944 is merged as I'm having a hard time keeping up with all the features breaking integ tests

core/src/plugins/kubernetes/types.ts
```diff
@@ -98,12 +98,14 @@ export type SyncableResource = KubernetesWorkload | KubernetesPod
...
 
+export type HelmRuntimeAction = HelmDeployAction | HelmPodRunAction | HelmPodTestAction
+
 export type SupportedRuntimeActions =
   | ContainerBuildAction
   | ContainerDeployAction
   | ContainerTestAction
   | ContainerRunAction
-  | HelmDeployAction
+  | HelmRuntimeAction
```
Walther, line 108: non-blocking nit: the other types are not collapsed into a combined `RuntimeAction` type - should we stay consistent and either expand the helm actions here, or collapse the others too?

core/src/plugins/kubernetes/helm/common.ts
```diff
@@ -238,8 +238,8 @@ export function getBaseModule(module: HelmModule): HelmModule | undefined {
 /**
  * Get the full absolute path to the chart, within the action build path, if applicable.
  */
-export async function getChartPath(action: Resolved<HelmDeployAction>) {
-  const chartSpec = action.getSpec("chart") || {}
+export async function getChartPath(action: Resolved<HelmRuntimeAction>) {
```
Walther, line 241: ref. earlier nit, these could also just be expanded like they are in the `kubernetes-type/common.ts`

core/src/plugins/kubernetes/helm/config.ts
```diff
@@ -6,33 +6,44 @@
...
 
 export const defaultHelmTimeout = 300
 export const defaultHelmRepo = "https://charts.helm.sh/stable"
 
+interface HelmChartSpec {
+  name?: string // Formerly `chart` on Helm modules
+  path?: string // Formerly `chartPath`
+  repo?: string // Formerly `repo`
+  url?: string
+  version?: string // Formerly `version`
```
Walther, line 41: nit: should we remove "formerly" comments where the field names are staying the same?

core/src/plugins/kubernetes/helm/helm-pod.ts
```diff
@@ -0,0 +1,196 @@
...
+import { getActionNamespaceStatus } from "../namespace"
+import { k8sGetRunResult, storeRunResult } from "../run-results"
+import { getResourceContainer, getResourcePodSpec, getTargetResource, makePodName } from "../util"
+import { HelmPodRunAction, helmPodRunSchema, HelmPodTestAction } from "./config"
+import { runAndCopy } from "../run"
+import { filterManifests, prepareManifests, prepareTemplates } from "./common"
+import { storeTestResult } from "../test-results"
+
+const helmRunPodOutputsSchema = kubernetesRunOutputsSchema
+const helmTestPodOutputsSchema = helmRunPodOutputsSchema
```
Walther, line 28: (just double-checking in case there's a copy-paste error here: is the `helmTestPodOutputsSchema` intended to be the same as `helmRunPodOutputsSchema`, or should this be `const helmTestPodOutputsSchema = kubernetesTestOutputsSchema` ?)

Walther (commented):
thanks for the great work! 🎉 
left a couple of tiny non-blocking nits and one question, but in general this looks good to me.

i'm letting someone else do another pass though, as i'm not 100% confident on my ability to catch possible nuances in this part of the codebase just yet.

core/src/plugins/kubernetes/types.ts
thsig: That's a good thought, but it might end up not quite working in this case since I'm about to add `kubernetes-exec` actions, which don't have the fields needed for the helpers used in these functions (e.g. `files` and `manifests`).

These sum types also aren't used in that many places, and I think it's sometimes clearer to just list them out in the function definition (since it saves the reader a "go to definition" lookup when reading the code).

core/src/plugins/kubernetes/helm/config.ts
thsig: Yeah, good point—will remove those.

core/src/plugins/kubernetes/helm/helm-pod.ts
thsig: They're all the same! Haha.

core/src/plugins/kubernetes/types.ts
Walther: Agreed, hence I was kind of leaning towards maybe expanding the `HelmRuntimeAction` to the separate `Deploy`, `Run`, and `Test` actions like with the Kubernetes-related actions (and function params)

core/src/plugins/kubernetes/helm/helm-pod.ts
```diff
@@ -0,0 +1,196 @@
...
+
+export const helmPodRunDefinition = (): RunActionDefinition<HelmPodRunAction> => ({
+  name: "helm-pod",
+  docs: dedent`
+    Executes a Run in an ad-hoc instance of a Kubernetes Pod and waits for it to complete.
```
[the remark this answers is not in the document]
thsig: Yeah, agreed.

edvald (approved):

PR #3956, opened 2023-03-15
fix(k8s): retry exec attempts in PodRunner

stefreak:
Before this change, the Kubernetes API request sent as part of PodRunner
exec calls have not been retried.

This change makes sure that we use the existing Kubernetes API retry
implementation and at the same time it makes sure that HttpErrors are
wrapped properly, for improved error messages.

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/api.ts
+++ b/core/src/plugins/kubernetes/api.ts
@@ -789,19 +789,26 @@ export class KubeApi {
       const execWithRetry = () => {
         const execHandler = new Exec(this.config)
 
-        return requestWithRetry(log, "Kubernetes API: exec", () => execHandler.exec(
-          namespace,
-          podName,
-          containerName,
-          command,
-          _stdout,
-          _stderr,
-          stdin || null,
-          tty,
-          (status) => {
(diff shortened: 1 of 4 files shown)


mkhq (approved):

edvald:
Does this just retry the connection, or is it possible that the same exec command executes more than once?

stefreak:
@edvald In theory it's possible that the same exec runs more than once, for example if the successful server response never makes it to the client – though that'll probably be very rare.

My first attempt at this fix, before I discovered `requestWithRetry`, was to use `pRetry` and only retry on 500, 502, 503 and 429 status codes. That would basically guarantee that the same exec is only being executed once.

I think I'll go back to that approach as it would be safer.

edvald:
Yeah, I think we should err on caution here, so that makes sense.

stefreak:
@edvald @mkhq would be glad if you could have another look at this 👍

mkhq (approved):

edvald:
@stefreak would you mind cherry-picking this over to `0.13`?

PR #3959, opened 2023-03-15
Exec Deploy improvements

edvald:
See individual commits:

**[refactor(exec): replace syncMode from exec Deploy with persistent flag](https://github.com/garden-io/garden/commit/5aea03fd105491d88c5a04ceb5642326258752ce)**

With the "dev mode" notion removed, this is a more natural API. For exec
modules, we use conditionals to enable the persistent mode when the Deploy is
in sync mode, such that the old behavior works as before when using
`--sync/--dev`.

**[refactor(exec): split exec.ts module](https://github.com/garden-io/garden/commit/1a8bb0ffc174b03e4dfe4685c99c5007bf085d75)**

**[improvement(exec): run persistent exec processes detached from the Garden proc](https://github.com/garden-io/garden/commit/0a4dde2170011c6074f663f774171f782ee4bf50)**

This makes it easier to work with multiple Garden processes, and imo is a
generally better experience for users. Log following should work just as before.
We use a pid file to track the detached process.

**[refactor(core): explicitly inform framework of persistent attached child process](https://github.com/garden-io/garden/commit/d5f1f12c964c905f97cf81349e536bac12eb35dc)**

This was originially conceived to support persistent exec processes, but is
generally useful without it (since persistent exec processes are no longer
attached to the Garden process). It's been renamed for clarity.


core/src/plugins/exec/exec.ts
```diff
@@ -597,6 +593,30 @@ export async function convertExecModule(params: ConvertModuleParams<ExecModule>)
...
+    let deployCommand = service.spec.deployCommand
+    let statusCommand = service.spec.statusCommand
+
+    if (service.spec.syncMode) {
+      // Maintain compatibility with devMode on exec modules
+      persistent = "${this.mode == 'sync'}"
+
+      if (service.spec.syncMode.command) {
+        deployCommand = <any>{
+          $if: persistent,
```
thsig: Haha. Very cool!

thsig: This is a pretty powerful new pattern that I'm sure we can make use of in more contexts.

thsig (approved):
Looks good to me!

PR #3960, opened 2023-03-15
feat(k8s): kubernetes-exec Runs and Tests

thsig:

> Note: Depends on #3947. Marking this as a draft until that's merged—I'll then rebase this one before marking it as ready.

Here, we introduce new Kubernetes-based action types for Runs and Test (both have the action type `kubernetes-exec`).

These run the provided args in an already deployed resource in the cluster, using \`spec.resource\` to locate the resource.

This is a great fit for inner-loop development, where rebuilding before retesting is not always desirable (e.g. when running tests in conjunction with code syncing), and to speed up the spinning up a new environment by executing Runs in already running services.

We should make a note of documenting this in detail in our guides, and consider converting some of the Test and Run actions in our example projects to use this new action type.


mkhq (approved):
🥇

PR #3966, opened 2023-03-16
fix(solver): ready status was ignored for several task types

edvald:
This would cause unnecessary execution of e.g. builds during deployments etc.

cc @thsig


vvagaytsev:
There are some test failures. Are those related to the changes made in the PR?

edvald:
@vvagaytsev @thsig this should be good to go now. The test failures were actual issues, aside from one bad test case.

vvagaytsev (approved):
👍

PR #3969, opened 2023-03-17
0.13 fix: logger init in emit warning

Orzelius:


core/src/warnings.ts
```diff
@@ -25,7 +25,11 @@ export function emitNonRepeatableWarning(message: string) {
   }
 
   if (!loggerContext.logger) {
-    loggerContext.logger = getLogger()
+    loggerContext.logger = Logger.initialize({
```
edvald, line 28: I really think we should be passing a `Log` instance here instead. This function is a bit unusual to begin with, so I'd suggest addressing that instead.

eysi09: Yeah I agree.

Orzelius: It's sometimes used before the logger is initialized though. I ran into this when trying out cloud custom commands and it [tried to log a warning on the load project config step](https://github.com/garden-io/garden/blob/5a886e65345fe48cf923e03d88b81837310aa003/core/src/cli/cli.ts#L603) at which point the logger is not yet initialized. Shoud I move the initialization to an earlier stage and pass it through the tree all the way to the [call](https://github.com/garden-io/garden/blob/0.13/core/src/config/base.ts#L210)?

Orzelius: Or I could make log an optional parameter and just not pass it if it doesn't exist yet

eysi09: Yeah I think the point stands that if this function uses the logger than it should be used as a parameter, just like anywhere else.

Also shouldn't we be preventing repeatable warnings being executed in the first place, as opposed to hacking around it by swallowing log messages? Feels like this can introduce hard to debug cases where the same logic is running multiple times but we don't see it. 

But if we do need this function, it should receive the logger as a required parameter and we simply need to initialise it sooner. Indeed, if any of the init functionality requires logging it should have access to the logger.

Orzelius: Initialized and passed sooner in [7ef3cb8](https://github.com/garden-io/garden/pull/3969/commits/7ef3cb82b968a9be47c293b3e26b22dbc364837e)

eysi09 (approved):
LGTM!

PR #3972, opened 2023-03-20
fix(k8s): add log streaming for exec Runs/Tests

thsig:

This was a simple oversight in https://github.com/garden-io/garden/pull/3960


vvagaytsev (approved):
Looks good! 👍

PR #3975, opened 2023-03-20
0.13 cherry-pick: fix(k8s): retry exec attempts in PodRunner

vvagaytsev:
* fix(k8s): retry exec attempts in PodRunner

Before this change, the Kubernetes API request sent as part of PodRunner exec calls have not been retried.

This change makes sure that we use the existing Kubernetes API retry implementation and at the same time it makes sure that HttpErrors are wrapped properly, for improved error messages.

* fix(k8s): wrap k8s errors for better error messages

This change makes sure that we wrap exec k8s errors like we do for other Kubernetes API endpoints. This improves the error messages.

* improvement: only retry if the error is recoverable

Only retry on 500, 502, 503 and 429 errors to reduce the risk that the command is being executed twice.

---------

This PR cherry-picks #3956 to `0.13`.

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/api.ts
+++ b/core/src/plugins/kubernetes/api.ts
@@ -30,6 +30,7 @@ import {
   Log as K8sLog,
   NetworkingV1Api,
   ApiextensionsV1Api,
+  HttpError,
 } from "@kubernetes/client-node"
 import AsyncLock = require("async-lock")
 import request = require("request-promise")
@@ -59,6 +60,7 @@ import { getExecExitCode } from "./status/pod"
 import { labelSelectorToString } from "./util"
 import { LogLevel } from "../../logger/logger"
 import { safeDumpYaml } from "../../util/serialization"
+import pRetry = require("p-retry")
 
 interface ApiGroupMap {
   [groupVersion: string]: V1APIGroup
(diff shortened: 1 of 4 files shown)


core/src/plugins/kubernetes/api.ts
```diff
@@ -778,6 +780,55 @@ export class KubeApi {
...
+            {
+              retries: 5,
+              minTimeout: 1000,
+              onFailedAttempt(error) {
+                if (error.cause instanceof HttpError && error.cause.statusCode) {
+                  // only retry if there is no risk that the command will be executed twice.
+                  const recoverableStatusCodes = [500, 502, 503, 429]
+                  if (recoverableStatusCodes.includes(error.cause.statusCode)) {
+                    retryLog = retryLog || log.debug("")
+                    retryLog.error(deline`
```
vvagaytsev, line 815: Replaced `.setState()` with `.error()` here as the new logger does not have the `setState` method.

stefreak: On the new logger, we probably also don't need to assign retryLog, so we could potentially remove the retryLog variable and just use `log.debug`?

stefreak (approved):
Thank you, I wasn't as fast as you were, would have created the cherry-pick myself very soon

🙏

PR #3980, opened 2023-03-22
improvement: add additional info to error message when sh is not found

10ko:
A user is having intermittent errors when running a test with artifacts where Garden responds with the error message:
```
<test-name> specifies artifacts to export, but the image doesn't contain the sh binary. In order to copy artifacts out of Kubernetes containers, both sh and tar need to be installed in the image.
```

They are sure that both `sh` and `tar` are present in the container.

Given the way we determine if `sh` or `tar` are not present and the difficulty in reproducing the error, I propose to print the full error message we receive from the execution, in the hope we can identify the root cause.

Fixes #


vvagaytsev (approved):

vvagaytsev:
Thank you! 👍

vvagaytsev (changes requested):
There are some test failures.

PR #3981, opened 2023-03-22
Action templates

edvald:
This adds support for templating actions and workflows in ConfigTemplate
configs (formerly ModuleTemplate).

The old way of referencing templates (a Module with `type: templated`) is also
superseded by a new `kind: RenderTemplate` config, with largely the same
semantics, except with an added `configs` field that supports actions and
workflows. We retain backwards compatibility though.

The resolution flow is very similar for actions and workflows, as it was (and
remains for now) for modules.

For now we have left out the option to generate files in the source tree, which
is possible with the now-deprecated `generateFiles` field on modules. We may
later consider making that a native feature on `RenderTemplate` configs as
opposed to on the templated configs themselves, but it may be an unnecessary
feature since that can in most cases be handled by `exec` actions.


vvagaytsev:
There are some errors in [e2e-templated-k8s-container-modules](https://app.circleci.com/pipelines/github/garden-io/garden/16083/workflows/d0bbf5bb-998d-432b-a9d2-a09184364c65/jobs/269313):
```
> [error]   Error processing config for Build action my-service-image:
Error validating inputs for container Build my-service-image (from module my-service-image) (module): value at . must have required property 'containerPort', value at . must have required property 'servicePort'
 > [silly]   Error: 
Error processing config for Build action my-service-image:
Error validating inputs for container Build my-service-image (from module my-service-image) (module): value at . must have required property 'containerPort', value at . must have required property 'servicePort'
    at /home/circleci/project/core/src/graph/actions.ts:158:13
```

edvald:
Should be fixed now @vvagaytsev

core/src/config/base.ts
```diff
@@ -39,15 +40,58 @@ The format of the files is determined by the configured file's extension:
...
+    configFilePath: joi.string().optional().meta({ internal: true }),
+    inputs: joi.object().optional().meta({ internal: true }),
+    parentName: joi.string().optional().meta({ internal: true }),
+    templateName: joi.string().optional().meta({ internal: true }),
+  },
+  allowUnknown: true,
+  meta: { internal: true },
+})
+
+joi
```
vvagaytsev, line 72: This looks like an orphan piece of code. Should it be removed?

thsig (approved):
Lovely stuff!

vvagaytsev (approved):
LGTM 👍 Thank you!

PR #3987, opened 2023-03-23
0.13 cherry picks from `main`

vvagaytsev:

Fixes #


plugins/terraform/cli.ts
```diff
@@ -209,12 +208,13 @@ export const terraformCliSpecs: { [version: string]: PluginToolSpec } = {
     name: "terraform-1-2-9",
     description: "The terraform CLI, v1.2.9",
     type: "binary",
+    _includeInGardenImage: true,
```
edvald, line 211: Why was this moved?

vvagaytsev: I'm not sure, it was moved in the [original commit](https://github.com/garden-io/garden/commit/d5441475dccfc37701d096b44c994112b7f2f228). I assume that we apply `_includeInGardenImage: true` only to the latest available tool version, don't we?

vvagaytsev: cc @thsig

edvald (approved):

PR #3990, opened 2023-03-24
Sync commands + sync rework + few small things caught in the process

edvald:
Submitting this for early review and testing.

Several improvements we're making here. The highlights:

**[improvement(k8s): syncs keep running after Garden process exits](https://github.com/garden-io/garden/commit/0345b1ddcc8474fbdebf2fad4e456d3fe758c15d)**

Also, we now only run one daemon per Garden project at any time, with a single
data directory. None of that temp directory business, symlinks and whatnot.

Further, we get updates from Mutagen instantly via a single
`mutagen sync monitor` process instead of execing a `mutagen sync list` process
every 2 seconds.

A side-effect of that is that our logging is maybe a touch noisy. I left a TODO
to address that.

**[feat(cli): add util mutagen command to ease troubleshooting sync issues](https://github.com/garden-io/garden/commit/739c42cceacb2be52492c7dfd0ac16db83feeb24)**

**[feat: add sync start and sync stop commands](https://github.com/garden-io/garden/commit/06a38e0104c0c33f013a546869912881748a6595)**

See the added docs for details.


thsig:
Whoop whoop! I'll review this one tomorrow.

edvald:
FYI @thsig, seems GitHub lists duplicate commits here after my force push. The actual git history seems alright.

core/src/commands/sync/sync-start.ts
```diff
@@ -0,0 +1,241 @@
...
+      const tasks = actions.map((action) => {
+        return new DeployTask({
+          garden,
+          graph,
+          log,
+          action,
+          force: false,
+          forceActions: [],
+          skipRuntimeDependencies: true,
+          startSyncs: false,
```
thsig, line 177: I suggest making this `startSyncs: true`—while this change would currently have no effect, it would make sense if we started e.g. returning an `outdated` status when the sync is not running.

I.e. even when we're processing tasks with `statusOnly: true`, it's probably clearer when the constructor params for the task should "reflect the intent of the task" (and less likely to cause breakages if the task classes change).

core/src/router/base.ts
```diff
@@ -149,13 +149,18 @@ type HandlerParams<K extends ActionKind, H extends keyof ActionTypeClasses<K>> =
...
+  (params: HandlerParams<K, H>): Promise<ActionRouterHandlerOutput<GetActionTypeResults<ActionTypeClasses<K>[H]>>>
 }
 
 export type WrappedActionRouterHandlers<K extends ActionKind> = {
   [H in keyof Omit<ActionTypeClasses<K>, CommonHandlers>]: WrapRouterHandler<K, H>
 }
 
+interface ActionRouterHandlerOutput<R> {
+  ctx: PluginContext
+  result: R
```
thsig, line 161: I'm curious—what was the motivation behind including `ctx` in the return value?

thsig (commented):
Great stuff! Added a couple of minor comments/questions there.

Also, the Git history does include a few duplicate commits (as far as I can tell):

`fix(dev): login and logout commands didn't work in dev command`
and
`feat: add \`sync start\` and \`sync stop\` commands`

The commits themselves aren't the same, but the messages are. So we can either change the messages or amend the original commits to include the patches (which come in the later commits with the same name).

thsig:
Once the commit history has been cleaned up, this is ready to merge.

core/src/router/base.ts
edvald: It'll come to light in the next PR :) It's basically to get access to dynamically generated values, including `ctx.events`.

core/src/commands/sync/sync-start.ts
thsig: ... definitely not a blocker though.

edvald:
Should be good now @thsig

thsig (approved):

PR #3996, opened 2023-03-27
0.13 cloud alignment

TimBeyer:

In order to have Cloud process events from 0.13 correctly, we needed to add the `actionType` key to emitted events which is one of "build", "deploy", "run" or "test".

These are hardcoded here as the property `action.type` available internally contained a different kind of information.

Fixes #

### DIFF
--- a/core/src/events.ts
+++ b/core/src/events.ts
@@ -217,6 +217,7 @@ export interface Events {
     timestamp: string
     actionUid: string
     actionName: string
+    actionType: string
     moduleName: string | null
     origin: string
     data: string
--- a/core/src/router/build.ts
+++ b/core/src/router/build.ts
@@ -14,10 +14,13 @@ import { BaseRouterParams, createActionRouter } from "./base"
 import { ActionState, stateForCacheStatusEvent } from "../actions/types"
 import { PublishActionResult } from "../plugin/handlers/Build/publish"
 
+const API_ACTION_TYPE = "build"
+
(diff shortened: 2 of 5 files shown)


thsig:
While we're at it—do we also need to add a `force: boolean` flag to all these events, given the discussion here?

https://discord.com/channels/817392104711651328/1088822131523080314/1089649413674258664

TimBeyer:
We'd prefer to not do a mapping on the wrapper object to go from `buildStatus` -> `build` etc and have every event payload clearly contain the information we need to identity the event we're dealing with.

I think there was also an issue with the `log` event type which then wouldn't contain any information about the action type.
Due to the potential race condition we do need to create task results from log payloads sometimes and there we need to know the type already, which we couldn't get from the `...Status` key.

After adding it to `log` events it feels a lot more consistent to have it present on every event then and making every event payload fully identify itself.

thsig:
Gotcha, that all makes sense.

thsig (approved):

TimBeyer:
@thsig I don't think Eysi touched anything that would conflict so we should be able to just merge this.

10ko:
@thsig what's the timeline for merging #4002? It looks like there won't be any conflict and this is ready.

thsig:
There are still a few failing tests there, we can merge once they're fixed.

thsig (approved):

PR #3997, opened 2023-03-27
0.13: utilize validation handler

Orzelius:
This will call call the validation handler in the action resolution process.

This doesn't fix the [original issue](https://github.com/orgs/garden-io/projects/5/views/1?pane=issue&itemId=23611004) that brought this to our attention as `garden validate` doesn't actually resolve actions.

Is that desirable? Should `garden validate` also resolve all static fields perhaps?


Walther:
Ran a quick `cd examples; for example in $(ls -1); do pushd $example; echo $example; gdev13 validate; popd; done;` to see if all of our examples pass validation after these changes. Most of them pass fine! 🎉 

There are a couple of validation failures:
- `argocd`: Could not read cluster from kubeconfig for context gke_garden
- `demo-project-start`: kubernetes: must specify deploymentRegistry in config
- `pulumi`: Error validating project environments: value is required
- `terraform-gke`: Error validating project environments: value is required

However, all of these failures also happen outside of this branch, so they are not new / caused by this extra validation ✅ 

Additionally, the `argocd` and `demo-project-start` examples most likely just require additional local configuration before running, so they might not be an issue at all.

This leaves just the `pulumi` and `terraform-gke` examples with validation failures. Here we need to note that those examples have not been converted to 0.13 syntax yet. This introduces the question: does our validation command require 0.13 syntax, and if it does, is that a feature or a bug? I know we plan to have runtime compatibility for 0.12 syntax configuration files in the 0.13 release, but does that extend to the `validate` command - should it too understand both formats, or should it encourage people towards the new one?

Orzelius:
@Walther Like the pr description says, as is, this pr does not change the behavior of the garden validate command.

Walther:
sorry, my bad 🤦‍♂️  should have read the description more carefully. thanks for the clarification!

thsig:
Do you think this is safe to merge before the pre-release, @edvald?

I'm asking because we're adding a validation step here, which might result in new errors in certain places.

edvald:
Yes. Better to get those errors than something more cryptic/unexpected further down.

core/src/garden.ts
```diff
@@ -876,7 +876,7 @@ export class Garden {
     const resolvedProviders = await this.resolveProviders(log)
     const rawModuleConfigs = await this.getRawModuleConfigs()
 
-    const graphLog = log.makeNewLogContext({ section: "graph" }).info(`Resolving actions and modules...`)
+    const graphLog = log.makeNewLogContext({ section: "graph" }).info(`Reading actions and modules...`)
```
vvagaytsev, line 879: It looks like [`resolver.resolveAll()` is called](https://github.com/garden-io/garden/blob/bec3d1ec1520cfd5bb0900d194c06ed8438e6a81/core/src/garden.ts#L890) after this log message. Should we change the wording here?

vvagaytsev: The old message can easily be restored while rebasing this on top of the `0.13` (there are some conflicts).

Orzelius: The modules are resolved, but not the actions as far as I understand.

thsig (approved):

core/src/garden.ts
vvagaytsev: The method's JSDoc says
> Resolve the raw module and action configs and return a new instance of ConfigGraph.

So, should we just keep the unified terminology in all places? We use "resolving" for providers too.

PR #3999, opened 2023-03-27
improvement: more granular version control in `self-update` command

vvagaytsev:

Fixes #

The commits will be squashed before merging.

### DIFF
--- a/core/src/commands/self-update.ts
+++ b/core/src/commands/self-update.ts
@@ -45,21 +45,23 @@ const selfUpdateOpts = {
   }),
   "major": new BooleanParameter({
     defaultValue: false,
+    // TODO Core 1.0 major release: add these notes:
+    //  "Takes precedence over --minor flag if both are defined."
+    //  "The latest patch version will be installed if neither --major nor --minor flags are specified."
     help: dedent`
-    Install the latest major version greater than the current one. Takes precedence over --minor flag if both are defined. Falls back to the current version if the greater major version does not exist.
-
-    The latest patch version will be installed if neither --major nor --minor flags are specified.
+    Install the latest major version greater than the current one. Falls back to the current version if the greater major version does not exist.
 
     Note! If you use a non-stable version (i.e. pre-release, or draft, or edge), then the latest possible major version will be installed.`,
   }),
-  "minor": new BooleanParameter({
-    defaultValue: false,
-    help: dedent`Install the latest minor version greater than the current one. Falls back to the current version if the greater minor version does not exist.
-
-    The latest patch version will be installed if neither --major nor --minor flags are specified.
-
-    Note! If you use a non-stable version (i.e. pre-release, or draft, or edge), then the latest possible major version will be installed.`,
-  }),
+  // TODO Core 1.0 major release: uncomment this:
+  // "minor": new BooleanParameter({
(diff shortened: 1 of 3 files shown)


core/src/commands/self-update.ts
```diff
@@ -151,7 +163,7 @@ export class SelfUpdateCommand extends Command<SelfUpdateArgs, SelfUpdateOpts> {
         result: {
           currentVersion,
           installationDirectory,
-          latestVersion,
+          latestVersion: targetVersion,
```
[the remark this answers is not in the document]
vvagaytsev: I decided to return both `targetVersion` and `latestVersion` in https://github.com/garden-io/garden/pull/3999/commits/75fe9dde530e1faf885ab126031f3076761d18f6

Orzelius (approved):

core/src/commands/self-update.ts
```diff
@@ -70,6 +81,7 @@ export class SelfUpdateCommand extends Command<SelfUpdateArgs, SelfUpdateOpts> {
        garden self-update          # update to the latest Garden CLI version
        garden self-update edge     # switch to the latest edge build (which is created anytime a PR is merged)
        garden self-update 0.12.24  # switch to the 0.12.24 version of the CLI
+       garden self-update --version-scope minor  # install the latest minor version greater than the current one
```
edvald, line 84: To me a more natural API would be e.g. `garden self-update --major` and `garden self-update --minor`. Feels unnecessarily verbose otherwise. wdyt?

vvagaytsev: Sounds good! I'll refactor this.

core/src/commands/self-update.ts
```diff
@@ -262,11 +305,112 @@ export class SelfUpdateCommand extends Command<SelfUpdateArgs, SelfUpdateOpts> {
...
+  }
+
+  private isEdgeVersion(version: string) {
+    return version === "edge" || version.startsWith("edge-")
+  }
+
+  private targetVersionMatches(tagSemVer: semver.SemVer, currentSemVer: semver.SemVer, versionScope: VersionScope) {
+    switch (versionScope) {
+      case "major":
+        return tagSemVer.major >= currentSemVer.major
```
edvald, line 386: I wonder if should consider 0.12->0.13 as major until we hit 1.x. That's basically the semantics we've been following, and it might be confusing for users if `garden self-update --major` _doesn't_ move them from 0.12.x to 0.13.

vvagaytsev: Sounds reasonable 👍  Should we just disable the `--minor` flag then? So it can be either the `--major` upgrade to the latest `0.13.x` or the default patch upgrade to the latest `0.12.x`.

vvagaytsev: Disabled `--minor` flag in https://github.com/garden-io/garden/pull/3999/commits/6ff1a5212ef0b03fe1443cea6e9840fdd22dd309

vvagaytsev:
> Is this ready for merging?

@thsig please check the latest commit, it addresses the latest comment from @edvald. I disabled the `--minor` flag as we discussed. Let's wait for the CI pipeline, and then we can merge it. (If you wanna merge it, please, remember to squash the commits :))

PR #4001, opened 2023-03-27
Add 0.13 issue template

Orzelius:
It was a bit tricky to write as it has to cover both issues and feature request but it'll have to do for now.

### DIFF
--- a/.github/ISSUE_TEMPLATE/0-13-issue-template.md
+++ b/.github/ISSUE_TEMPLATE/0-13-issue-template.md
@@ -1,19 +1,44 @@
 ---
 name: 0.13 issue template
 about: Template for issues related with the 0.13 release
-title: '0.13: [Feature/Bug]: Describe here'
+title: '0.13: [Bug]: Describe here'
 labels: '0.13'
 assignees: ''
 
 ---
 
-This issue relates to the Bonsai (0.13) release of Garden core.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/0-13-issue-template.md
```diff
@@ -0,0 +1,19 @@
...
+
+---
+
+This issue relates to the Bonsai (0.13) release of Garden core.
+
+If it's a bug please describe how to reproduce it or even better link a reproducible example. That would allow as to get to the fix faster. Make sure you describe in detail the current and expected behavior. Also feel free to suggest workaround for the issue.
+
+If it's a feature request please write out the motivation and benefits of the potential feature. We welcome implementation or api ideas. A workaround for people struggling with the lack of this feature would also be of help.
+
+Please fill the following fields if they are relevant:
```
vvagaytsev, line 16: Should we just copy the https://github.com/garden-io/garden/blob/0.13/.github/ISSUE_TEMPLATE/BUG_REPORT.md#your-environment section from the Bug Report template to keep the same formatting styles?

Orzelius: That wouldn't work for the feature request part.

Could copy both and add a message that they should delete the unneeded part.

PR #4002, opened 2023-03-27
Action logs

eysi09:

This PR adds the `ActionLog` class which extends the `CoreLog` and has action related metadata. That goal is to ensure that log entries that are generated by actions and further downstream have the data (namely `actionKind` and `actionName`) attached to them to render the full context. This data is also useful for Cloud. 

The idea is for this context to replace the current log entry sections which are just strings with little consistency and no semantic meaning. 

So instead of an API like this:

```
const log = logger.getLog()
log.info({ section: "api", msg: "hello world" }) // Rendered as api → hello world
```

we have:

```
const actionLog = createLog({ log, actionName: "api", actionKind: "build" })
actionLog.info("hello world") // Rendered as build.api → hello world
```

This allows us to create the action log context once and not have to worry about the setting the section everywhere. 

The idea is to have a corresponding `PluginLog` and maybe even `PluginActionLog`. 

Beyond that, there are some other minor changes:

- Fix bug introduced in #3969.
- Clean up how writers are initialised and discern between terminal writer and file writers.
- Have error and success messages print duration instead of specifying at call site for more consistency.

Partially addresses #3254 

This is not 100% quite yet. First off, the set-up is still a little awkward and I don't love how much testing is needed around adding new Log types, e.g. PluginLog. Beyond that:

- More tests are needed
- Need to add more Log types, like aforementioned PluginLog
- It's a little inconsistent how CoreLog.createLog can overwrite the `context` whereas ActionLog.createLog cannot. It shouldn't be allowed in either case and CoreLogs with a different context should probably be created differently. 

All that being said, this should be enough for the 0.13 pre-release and I'll continue iterating from here.

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -12,11 +12,18 @@ import { resolve, join } from "path"
 import chalk from "chalk"
 import { pathExists } from "fs-extra"
 import { getBuiltinCommands } from "../commands/commands"
-import { shutdown, sleep, getPackageVersion, registerCleanupFunction, getCloudDistributionName } from "../util/util"
+import {
+  shutdown,
+  sleep,
+  getPackageVersion,
+  registerCleanupFunction,
+  getCloudDistributionName,
+  getCloudLogSectionName,
+} from "../util/util"
 import { Command, CommandResult, CommandGroup, BuiltinArgs } from "../commands/base"
 import { PluginError, toGardenError, GardenBaseError } from "../exceptions"
 import { Garden, GardenOpts, DummyGarden } from "../garden"
-import { Logger, LoggerType, LogLevel, parseLogLevel } from "../logger/logger"
+import { getTerminalWriterType, Logger, LogLevel, parseLogLevel } from "../logger/logger"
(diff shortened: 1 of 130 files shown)


core/src/cli/cli.ts
```diff
@@ -255,35 +245,16 @@ ${renderCommands(commands)}
...
-    // Currently we initialise empty placeholder entries and pass those to the
-    // framework as opposed to the logger itself. This is to give better control over where on
-    // the screen the logs are printed.
-    // TODO: Remove header and footer logs. Not needed any more.
-    const headerLog = logger.makeNewLogContext()
-    const log = logger.makeNewLogContext()
-    const footerLog = logger.makeNewLogContext()
+    // Some commands may set their own logger type so we update the logger config here,
+    // once we've resolved the command.
+    const commandLoggerType = command.getTerminalWriterType({ opts: parsedOpts, args: parsedArgs })
```
eysi09, line 250: Note that for the dev command this means we may log some things to the screen before it starts up. 

I'm assuming we clear the screen anyway in between so it shouldn't be an issue.

thsig (approved):

PR #4017, opened 2023-03-28
refactor(logger): use generics instead of multiple class types

eysi09:

Fixes #


edvald:
Any context as to why you're making this change?

PR #4019, opened 2023-03-29
test: fix ingress integ tests

Orzelius:
The failures were caused by https://github.com/garden-io/garden/pull/3997 as the container validation
handler actually modifies the deployment's ingress hostname
to the project default hostname if it's not specified on the deploy
itself.

I couldn't figure out how to fix this in a way where the proper default would be passed from the mocked provider config in the tests so I just manually assigned the hostnames.

### DIFF
--- a/core/test/integ/src/plugins/kubernetes/container/ingress.ts
+++ b/core/test/integ/src/plugins/kubernetes/container/ingress.ts
@@ -401,6 +401,7 @@ describe("createIngressResources", () => {
       annotations: {},
       path: "/",
       port: "http",
+      hostname: "my.domain.com",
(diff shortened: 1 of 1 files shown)


edvald (approved):

edvald:
Hmm actually... Let me take one more look at this in a bit

Orzelius:
@edvald can I merge this for now?

edvald:
My concern there was whether the test was legitimately broken there, and couldn't quite tell from a glance. Maybe double check if the default hostname is correctly set and propagated, that's one of the things that might legitimately have gone wrong along the way.

Issue #4012, opened 2023-03-28
0.13: [Bug]: `garden serve` command doesn't work with `disabled-configs` example


vvagaytsev:
### Current Behavior
Executing `garden serve` from `disabled-config` example fails with the error:
```
Server 📊


ℹ garden                    → Running in namespace default in environment local
ℹ providers                 → Getting status...
✔ providers                 → Cached (in 0.8 sec)
ℹ providers                 → Run with --force-refresh to force a refresh of provider statuses.
ℹ graph                     → Resolving actions and modules...
✔ graph                     → Done (in 0.1 sec)
✖ run.backend               →
Failed processing resolve Run type=container name=backend. Here is the output:
────────────────────────────────────────────────────────────────────────────────
Build type=container name=backend is disabled.
──────────────────────────────────────────────────────────────────────────────── (in 0 sec)
✖ Failed loading the project: Error: Failed to resolve Run type=container name=backend: Error: resolve Run type=container name=backend failed: Error: Build type=container name=backend is disabled.
```

### Expected behavior
No errors.

### Reproducible example
See above.

### Additional context

This might be related to #3799

### Your environment

* **OS:** 
* **How I'm running Kubernetes:**  

`garden version`

Orzelius:
closed with https://github.com/garden-io/garden/pull/4020

PR #4020, opened 2023-03-29
0.13: fix disabled build validation

Orzelius:
closes https://github.com/garden-io/garden/issues/4012

also adds validation to other container runtime actions


Walther (approved):
LGTM as far as i can tell! 🚀 
Verified locally that `gdev serve` no longer fails in the `disabled-configs` example.

vvagaytsev (approved):
Looks good! 👍 Thanks for the fix! 🚀

PR #4022, opened 2023-03-29
feat: support pre-release versions in `self-update` command

vvagaytsev:
To be able to update to the `0.13` pre-releases from `0.12.x` using `garden self-update` command.
This PR also contains some refactoring.

Fixes #

The commits will be squashed before the merge. See individual commits for details.

### DIFF
--- a/core/src/commands/self-update.ts
+++ b/core/src/commands/self-update.ts
@@ -247,15 +247,10 @@ export class SelfUpdateCommand extends Command<SelfUpdateArgs, SelfUpdateOpts> {
 
     try {
       // Fetch the desired version and extract it to a temp directory
+      const { build, filename, extension, url } = this.getReleaseArtifactDetails(platform, desiredVersion)
       if (!platform) {
         platform = getPlatform() === "darwin" ? "macos" : getPlatform()
       }
-      const architecture = "amd64" // getArchitecture()
-      const extension = platform === "windows" ? "zip" : "tar.gz"
-      const build = `${platform}-${architecture}`
-
-      const filename = `garden-${desiredVersion}-${build}.${extension}`
-      const url = `${this._baseReleasesUrl}${desiredVersion}/${filename}`
 
       log.info("")
       log.info(chalk.white(`Downloading version ${chalk.cyan(desiredVersion)} from ${chalk.underline(url)}...`))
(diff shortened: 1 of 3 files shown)


Walther (commented):
I think this looks good but I'll let Albert double-check as well.
Thanks for the refactor and fix!

core/src/commands/self-update.ts
```diff
@@ -351,35 +351,52 @@ export class SelfUpdateCommand extends Command<SelfUpdateArgs, SelfUpdateOpts> {
       )
     }
 
+    const targetRelease = await this.findRelease((release) => {
```
Orzelius: I think this deserves a unit test or two

core/src/commands/self-update.ts
```diff
@@ -25,7 +25,7 @@ import stream from "stream"
 
 const selfUpdateArgs = {
   version: new StringParameter({
-    help: `Specify which version to switch/update to.`,
+    help: `Specify which version to switch/update to. It can be either a stable release version, or a pre-release version, or an edge one.`,
```
Orzelius, line 28:
```suggestion
    help: `Specify which version to switch/update to. It can be either a stable release, a pre-release, or an edge release version.`,
```

maybe also add an example

docs/reference/commands.md
```diff
@@ -3444,7 +3444,7 @@ Examples:
 
 | Argument | Required | Description |
 | -------- | -------- | ----------- |
-  | `version` | No | Specify which version to switch/update to.
```
Orzelius, line 3447: same as above

Orzelius (commented):
a few comments

core/src/commands/self-update.ts
vvagaytsev: Good point! Actually, we need to test the predicate here. The machinery around it is quite trivial.

core/test/unit/src/commands/self-update.ts
```diff
@@ -284,4 +334,94 @@ describe("SelfUpdateCommand", () => {
...
+        expectSkipped({ tag_name: "test", draft: true, prerelease: false }, semver.parse("0.13.0")!, "patch")
+      })
+
+      it("should skip the same pre-release version", () => {
+        expectSkipped({ tag_name: "test", draft: true, prerelease: false }, semver.parse("0.13.0")!, "patch")
+      })
+
+      it("should skip a newer pre-release version", () => {
+        expectSkipped({ tag_name: "test", draft: true, prerelease: false }, semver.parse("0.13.0")!, "patch")
+      })
```
Walther, line 414: possible copy-paste issue - these test cases look similar

vvagaytsev: Thanks! Fixed that in https://github.com/garden-io/garden/pull/4022/commits/144cfee2402b42438306968528a6df9065c9c72a

Walther (approved):
thanks for the good work! looks good to me 🚀

PR #4031, opened 2023-03-31
fix: set defaultEnv default value

Orzelius:
Currently the `garden set default-env ''` command doesn't work due to garden thinking `""` is a non-existing parameter. I think we should implement a `garden clear` instead as that's a more clear api, no pun intended. I can do that if we agree that's the best way to go


edvald:
I take it the description here is unrelated to the fix? That separate issue is fairly easy to fix, much easier than creating a whole new command.

Orzelius:
> I take it the description here is unrelated to the fix? 

yep, just something I noticed while testing the fix

edvald (approved):

PR #4035, opened 2023-04-02
improvement(api): allow unstructured string arguments in HTTP API commands

edvald:
Makes it more straightforward to feed autocomplete suggestions back to the API.

cc @trymbill


trymbill (approved):
LGTM! 👍 

I'll branch off of this branch and attempt to add `stringArguments` to the output of the autocomplete command so that it can be easily passed from autocomplete to command.

PR #4036, opened 2023-04-03
refactor: rename `modules` project config entry to `scan`

### LINKED ISSUE #3512: Rename Project config `modules.*` to `scan.*`
Rename Project config `modules.*` to `scan.*`. Allow the alias with a deprecation notice. Include this in `migrate` command.

vvagaytsev:

Fixes #3512


core/src/config/project.ts
```diff
@@ -187,10 +187,18 @@ export interface ProjectConfig {
   dotIgnoreFile: string
   dotIgnoreFiles?: string[]
   environments: EnvironmentConfig[]
+  /**
+   * Deprecated. Use {@link #scan} instead.
+   * @deprecated remove in 0.14
+   */
   modules?: {
```
edvald, line 194: I think we should just remove this from the internal interface.

core/src/config/base.ts
```diff
@@ -290,11 +290,56 @@ export function prepareProjectResource(log: Log, spec: any): ProjectResource {
...
+    log,
+    "Project configuration field `modules` is deprecated in 0.13 and will be removed in 0.14. Please use the `scan` field instead."
+  )
+
+  if (!projectSpec.scan) {
+    projectSpec.scan = {}
+  }
+
+  if (projectSpec["modules"].include) {
+    const oldSyntaxInclusions = projectSpec["modules"].include || []
```
edvald, line 315: Why not simply use the rename option in the joi schema? You can catch and emit the usage of the old field before validating.

vvagaytsev: That was my initial plan. Should we deny using both fields simultaneously?

vvagaytsev: Simplified this in https://github.com/garden-io/garden/pull/4036/commits/2966c7f86ef5b6f6c7f2c7c6fc428b9c808ba6e0

PR #4037, opened 2023-04-03
improvement(api): return arguments seperately in autocomplete suggestion

trymbill:
Expands on #4035 to return string arguments from argument based autocomplete suggestions in api


trymbill:
Doesn't cover all cases. Closing.

PR #4040, opened 2023-04-03
Fix mutating action versions

edvald:
Plus some other minor issues discovered in the process.

cc @Orzelius

### DIFF
--- a/core/src/actions/base.ts
+++ b/core/src/actions/base.ts
@@ -33,7 +33,7 @@ import { actionOutputsSchema } from "../plugin/handlers/base/base"
 import type { GraphResult, GraphResults } from "../graph/results"
 import type { RunResult } from "../plugin/base"
 import { Memoize } from "typescript-memoize"
-import { flatten, fromPairs, isString, omit, sortBy } from "lodash"
+import { cloneDeep, flatten, fromPairs, isString, omit, sortBy } from "lodash"
 import { ActionConfigContext, ActionSpecContext } from "../config/template-contexts/actions"
 import { relative } from "path"
 import { InternalError } from "../exceptions"
@@ -535,7 +535,7 @@ export abstract class BaseAction<C extends BaseActionConfig = BaseActionConfig, 
   getConfig(): C
   getConfig<K extends keyof C>(key: K): C[K]
   getConfig(key?: keyof C["spec"]) {
-    return key ? this._config[key] : this._config
+    return cloneDeep(key ? this._config[key] : this._config)
   }
 
   /**
@@ -699,7 +699,7 @@ export abstract class ResolvedRuntimeAction<
   getSpec(): Config["spec"]
   getSpec<K extends keyof Config["spec"]>(key: K): Config["spec"][K]
   getSpec(key?: keyof Config["spec"]) {
-    return key ? this._config.spec[key] : this._config.spec
+    return cloneDeep(key ? this._config.spec[key] : this._config.spec)
   }
 
   getOutput<K extends keyof Outputs>(key: K) {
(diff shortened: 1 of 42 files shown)


core/src/plugins/kubernetes/container/exec.ts
```diff
@@ -27,7 +27,7 @@ export const execInContainer: DeployActionHandler<"exec", ContainerDeployAction>
   const namespace = await getAppNamespace(k8sCtx, log, k8sCtx.provider)
 
   // TODO: this check should probably live outside of the plugin
-  if (!status.detail?.detail.workload || !includes(["ready", "outdated"], status.state)) {
+  if (!status.detail?.detail.workload || !includes(["ready", "outdated"], status.detail.state)) {
```
Orzelius, line 30: Could we maybe rename the `detail.state` to something else to reduce confusion, for example `serviceStatus`?

core/src/actions/base.ts
```diff
@@ -536,7 +536,7 @@ export abstract class BaseAction<C extends BaseActionConfig = BaseActionConfig,
   getConfig(): C
   getConfig<K extends keyof C>(key: K): C[K]
   getConfig(key?: keyof C["spec"]) {
-    return key ? this._config[key] : this._config
+    return cloneDeep(key ? this._config[key] : this._config)
```
Orzelius, line 539: Some integration tests used the return to modify action configs, how would this be done now?

Orzelius (approved):
Had some non-blocking comments

core/src/actions/base.ts
thsig: Maybe we'll need to override these methods in the test setup, then?

core/src/plugins/kubernetes/container/exec.ts
Orzelius: Also probably the status two lines below should be changed as well

edvald: Yeah, I have that rename on the TODO. Also getting rid of `.detail.detail` :P

core/src/actions/base.ts
edvald: For testing purposes it's also possible to directly manipulate the config on the action object, instead of after retrieving via `getConfig()`.

vvagaytsev: We also got [one unit test failure](https://app.circleci.com/pipelines/github/garden-io/garden/16332/workflows/70782cff-e9a4-4b79-a33e-2f08afd500a0/jobs/275027) because of this.

We have a similar test case for the `Test` action [here](https://github.com/garden-io/garden/blob/0.13/core/test/unit/src/router/test.ts#L114). It looks like rewriting the `Run` action test, in the same way, will fix the problem.

Also, (not urgent), it might be nice to refactor the [`run.ts`](https://github.com/garden-io/garden/blob/0.13/core/test/unit/src/router/run.ts) and [`test.ts`](https://github.com/garden-io/garden/blob/0.13/core/test/unit/src/router/test.ts) to have the same approach of creating test actions, and, maybe, extract some shared helpers.

core/src/plugins/kubernetes/container/exec.ts
vvagaytsev: @edvald should the same fix been applied to [`helm exec`](https://github.com/garden-io/garden/blob/0.13/core/src/plugins/kubernetes/helm/exec.ts#L61) and [`kubernetes-type exec`](https://github.com/garden-io/garden/blob/0.13/core/src/plugins/kubernetes/kubernetes-type/exec.ts#L52)?

vvagaytsev (changes requested):

edvald:
Updated, should be good now @vvagaytsev

vvagaytsev (approved):
Thanks, @edvald! 👍 
I updated the docs in https://github.com/garden-io/garden/pull/4040/commits/74de54d07f8067b6902b2c9dd6cd20bdf65d1d50

Now it's ready to go! 🚀

PR #4042, opened 2023-04-04
0.13: fix kubernetes type sync

Orzelius:
Works locally but want to make sure this doesn't trip up any integ tests

### DIFF
--- a/core/test/integ/src/plugins/kubernetes/container/ingress.ts
+++ b/core/test/integ/src/plugins/kubernetes/container/ingress.ts
@@ -511,7 +511,7 @@ describe("createIngressResources", () => {
       annotations: {},
       path: "/",
       port: "http",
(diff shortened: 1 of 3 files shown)


edvald (approved):

core/test/integ/src/plugins/kubernetes/container/ingress.ts
```diff
@@ -511,7 +511,7 @@ describe("createIngressResources", () => {
       annotations: {},
       path: "/",
       port: "http",
-      hostname: "hostname.invalid",
```
Orzelius, line 514: cc @Walther, I couldn't figure out why these started failing after the domain change so I just changed it back for now. The same test file has some more usage of such "incorrect" domains.

Orzelius:
aaaand the integ tests have magically started working again, fun

PR #4050, opened 2023-04-05
fix: login/logout using the configured domain

### LINKED ISSUE #4049: [BUG] Garden Login always resolves app.garden even if another domain is configured
When using `garden login` it will always try to login to https://app.garden. The correct behavior is to give precedence to the domain in the configuration or as env variable.

Introduced in https://github.com/garden-io/garden/commit/7f93b908

mkhq:

Ensures that the login/logout commands uses the cloud domain in the project config. Both have the `noProject` flag set which leads to the garden instance not being initialized with a project config. Ensures that this is compatible with `garden dev` and also includes login/logout in the help listing.

Fixes #4049 

@edvald There was an earlier commit which reverted this behavior because of the dev command, https://github.com/garden-io/garden/commit/7f93b90830e1968526fa7f748e6f195c9a6e4421. Could you give some more context on how this relates?

### DIFF
--- a/core/src/cli/command-line.ts
+++ b/core/src/cli/command-line.ts
@@ -50,7 +50,6 @@ const directInputKeys = [
 ]
 
 const hideCommands = ["config analytics-enabled", "tools"]
-const overrideHiddenCommands = ["login", "logout"]
 
 interface CommandLineEvents {
   message: string
@@ -427,14 +426,7 @@ ${renderDivider({ width, char, color })}
   showHelp() {
     // TODO: group commands by category?
     const renderedCommands = renderCommands(
-      this.commands.filter(
-        (c) =>
-          !(
-            (c.hidden && !overrideHiddenCommands.includes(c.getFullName())) ||
-            c instanceof CommandGroup ||
-            hideCommands.includes(c.getFullName())
-          )
-      )
+      this.commands.filter((c) => !(c.hidden || c instanceof CommandGroup || hideCommands.includes(c.getFullName())))
     )
 
     const helpText = `
--- a/core/src/cli/command-line.ts
+++ b/core/src/cli/command-line.ts
@@ -50,6 +50,7 @@ const directInputKeys = [
 ]
 
 const hideCommands = ["config analytics-enabled", "tools"]
+const overrideHiddenCommands = ["login", "logout"]
 
 interface CommandLineEvents {
   message: string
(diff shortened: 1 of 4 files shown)


core/src/commands/login.ts
```diff
@@ -38,26 +39,24 @@ export class LoginCommand extends Command {
...
-          {
-            root: garden.projectRoot,
-          }
-        )
-      }
+    // NOTE: The Cloud API is missing from the Garden class for commands with noProject
+    // so we initialize it here. noProject also make sure that the project config is not
+    // initialized in the garden class, so we need to read it in here to get the cloud
+    // domain.
+    const projectConfig: ProjectResource | undefined = await cli!.getProjectConfig(log, garden.projectRoot)
```
edvald, line 46: There was a reason for changing this. The `cli` parameter is not populated when running from the dev command. This will need to be fixed differently, with that in mind.

edvald (changes requested):

core/src/commands/login.ts
mkhq: `login` and `logout` are currently not listed as commands in the `garden dev` help. Should we make that possible?

edvald: They should be listed imo. I think it's a reasonable thing to do in the console.

core/src/cli/command-line.ts
```diff
@@ -50,6 +50,7 @@ const directInputKeys = [
 ]
 
 const hideCommands = ["config analytics-enabled", "tools"]
+const overrideHiddenCommands = ["login", "logout"]
```
edvald, line 53: A simpler approach would be to just remove the `hidden` flag from the commands :P I don't see any reason not to.

edvald:
One last comment, otherwise seems good to me. Did you test it in the dev command?

core/src/cli/command-line.ts
mkhq: Yeah, this is a bit clunky... I looked into where `hidden` is used and found this for the reference docs: https://github.com/garden-io/garden/blob/0.13/core/src/docs/commands.ts#L26.

mkhq:
> One last comment, otherwise seems good to me. Did you test it in the dev command?

Yes a couple of times :)

core/src/cli/command-line.ts
edvald: Oh I meant specifically for the login/logout commands. The flag is not necessary on them, and I'm not even sure why it's there in the first place.

vvagaytsev:
@edvald is this ready to go?

edvald (approved):

PR #4052, opened 2023-04-05
0.13 fix exec to not deploy anything

vvagaytsev:
This prevents the `garden exec` command from executing any deployments. If the target container is found, the command execution proceeds. Otherwise, an error is thrown.

Fixes #

The commits will be squashed before the merge.

### DIFF
--- a/core/src/commands/exec.ts
+++ b/core/src/commands/exec.ts
@@ -98,7 +98,7 @@ export class ExecCommand extends Command<Args, Opts> {
       case "missing":
       case "stopped":
         throw new NotFoundError(
-          `Cannot execute command in the '${action.name}' service. The target container is ${deployState}.`,
+          `${action.key()} status is ${deployState}. Cannot execute command.`,
           { deployState }
         )
       case "ready":
--- a/core/src/commands/exec.ts
+++ b/core/src/commands/exec.ts
@@ -63,21 +63,21 @@ export class ExecCommand extends Command<Args, Opts> {
   outputsSchema = () => execInDeployResultSchema()
 
   printHeader({ headerLog, args }) {
-    const serviceName = args.deploy
+    const deployName = args.deploy
     const command = this.getCommand(args)
     printHeader(
       headerLog,
-      `Running command ${chalk.cyan(command.join(" "))} in service ${chalk.cyan(serviceName)}`,
+      `Running command ${chalk.cyan(command.join(" "))} in service ${chalk.cyan(deployName)}`,
       "runner"
     )
   }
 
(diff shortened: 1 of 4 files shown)


core/src/commands/exec.ts
```diff
@@ -62,29 +63,57 @@ export class ExecCommand extends Command<Args, Opts> {
...
+    const status = await router.deploy.getStatus({ action: resolved, graph, log })
+    const deployState = status.result.detail?.state
+    switch (deployState) {
+      // Warn if the deployment is not ready yet or unhealthy, but still proceed.
+      case undefined:
+      case "deploying":
+      case "outdated":
+      case "unhealthy":
+      case "unknown":
+        log.warn(chalk.white(`Current state: ${chalk.whiteBright(deployState)}`))
```
edvald, line 95: This log message could be clearer. Suggest making it yellow+bold and say something like `\`The current state of ${action.key()} is ${chalk.whiteBright(deployState)}. If this command fails, you may need to re-deploy it with the ${chalk.whiteBright("deploy")} command\``.

core/src/commands/exec.ts
```diff
@@ -62,29 +63,57 @@ export class ExecCommand extends Command<Args, Opts> {
...
+        break
+      // Only fail if the deployment is missing or stopped.
+      case "missing":
+      case "stopped":
+        throw new NotFoundError(
+          `Cannot execute command in the '${action.name}' service. The target container is ${deployState}.`,
```
edvald, line 101:
```suggestion
          `${action.key()} status is ${deployState}. Cannot execute command.`,
```

core/src/plugin/handlers/Deploy/exec.ts
```diff
@@ -35,7 +35,7 @@ export const execInDeployResultSchema = () =>
 
 export class ExecInDeploy<T extends DeployAction = DeployAction> extends ActionTypeHandlerSpec<
   "Deploy",
-  ExecInDeployParams<Executed<T>>,
+  ExecInDeployParams<Resolved<T>>,
```
edvald, line 38: I don't think this makes sense. The exec plugin handlers will rely on a status being present on the action, which means it needs to be executed. We can approach this differently. I'll add a commit with my suggested approach.

edvald: I've added a commit that you can check out and then squash in if it looks good @vvagaytsev

vvagaytsev: Looks good, thank you!

vvagaytsev: And let's approve this first :)

thsig (approved):

PR #4053, opened 2023-04-05
improvement(k8s): catch internal etcdserver throttling error from API

edvald:
A user came across this nice little Kubernetes API peculiarity. Should help make the API wrapper more resilient to etcd throttling issues.


thsig (approved):
Oh, Kuberneeters.

PR #4066, opened 2023-04-11
chore: set default timeouts

### LINKED ISSUE #3516: Set non-null default values for all timeout fields, and require a number
Need to look through all timeout configs in the project. It might be easier to split it into sub-tasks.

vvagaytsev:
Open questions:
* Should we increase the default timeouts for all actions? Currently, we have 20 min for Builds, 10 min for Runs and Tests, and no limits for Deploys (there are some configurable ones for `exec` Deploys). -> This can be discussed in the code review round.
* For Deploys, it also might make sense to introduce some timeouts. -> discussion is in https://discord.com/channels/817392104711651328/1088795679159222292/1095340640314855554 Many (but not all) Deploy have the `timeout` in the action type-specific `spec` field. We might need to pull it up to the root config level like in Builds, Runs, and Deploys. That will be done in a separate PR.
* ~It might be a good idea to emit a warning message about the default action timeouts on Garden startup.~

Closes #3516

See individual commits for details. **Squash the commits before the merge.**


vvagaytsev:
Let's have a pre-review round @thsig @edvald @Orzelius @stefreak 
I've prepared a list of questions to be discussed in the PR description.

Orzelius:
## my opinions


> Should we increase the default timeouts for all actions?

why? I think these are plenty long enough

> For Deploys, it also might make sense to introduce some timeouts.

I'm thinking between 3 to 5 min, don't see how a normal deploy could take longer than that

> It might be a good idea to emit a warning message about the default action timeouts on Garden startup.

Nah, I don't think so. User's will see the timeout if they ever run into it and most probably won't unless they're doing something wrong or very special

vvagaytsev:
@Orzelius thanks for the feedback!

I've been thinking about this again and have realized that warning messages can be too annoying for users. It's better to emphasize that all actions have timeouts in the release notes. Anyway, it is a breaking change and will be announced accordingly.

On the default values, I'm not sure about deployments. The deployment time depends on the deployment policy and the number of app instances to be deployed.

Theoretically, every action can take a long time. But, of course, the default values should be something sane :)
Maybe keep the current values as they are, and make Deploy timeout the same as Run/Test one, i.e. 10 min.

vvagaytsev:
There are some test failures, I'll fix those soon.

core/src/plugins/kubernetes/container/run.ts
```diff
@@ -32,6 +33,7 @@ export const k8sContainerRun: RunActionHandler<"run", ContainerRunAction> = asyn
     dropCapabilities,
   } = action.getSpec()
 
+  const timeout = action.getConfig("timeout") || DEFAULT_RUN_TIMEOUT_SEC
```
thsig, line 36: Isn't the `|| DEFAULT_RUN_TIMEOUT_SEC` part unnecessary, since the timeout value is always set on the action?

The same question applies elsewhere in this PR.

thsig (commented):
Looks good to me! Mostly had one comment there.

core/src/plugins/kubernetes/container/run.ts
vvagaytsev: @thsig, yes, that's correct. We can safely declare the `timeout` field as non-optional in `BuildActionConfig`, `RunActionConfig`, and `TestActionConfig`.

it would be also nice to pull that field to the base class `BaseActionConfig`, and make it mandatory in `DeployActionConfig` too.

core/src/constants.ts
```diff
@@ -26,8 +26,9 @@ export const DEFAULT_PORT_PROTOCOL = "TCP"
 export const PREVIOUS_API_VERSION = "garden.io/v0"
 export const DEFAULT_API_VERSION = "garden.io/v1"
 
-export const DEFAULT_TEST_TIMEOUT = 60 * 1000
-export const DEFAULT_RUN_TIMEOUT = 60 * 1000
+export const DEFAULT_BUILD_TIMEOUT_SEC = 1200
```
Walther, line 29: (bikeshedding: 20 minutes feels like a pretty long default build timeout, wonder if we could have a lower value here to encourage folks to keep their build times snappy)

core/src/plugin/base.ts
```diff
@@ -114,6 +114,7 @@ export const moduleActionParamsSchema = () =>
 export const runBaseParams = () => ({
   interactive: joi.boolean().description("Whether to run interactively (i.e. attach to the terminal)."),
   silent: joi.boolean().description("Set to false if the output should not be logged to the console."),
+  // TODO-0.13.0: revisit this, the timeout might not need to be optional
```
Walther, line 117: 👍 mandatory timeout (with a default) would be nice

core/test/unit/src/garden.ts
```diff
@@ -4572,9 +4568,9 @@ describe("Garden", () => {
     })
 
     context("test against fixed version hashes", async () => {
-      const moduleAVersionString = "v-0a4dda85e8"
-      const moduleBVersionString = "v-0ab7a050db"
-      const moduleCVersionString = "v-958ee38dd2"
+      const moduleAVersionString = "v-6abf03be92"
+      const moduleBVersionString = "v-13d9d46957"
+      const moduleCVersionString = "v-554616d870"
```
Walther, line 4573: do these hashes change because of the timeout field, or something unrelated?

Walther (dismissed):
Looks good to me as far as I can tell, good work! 🎉

core/test/unit/src/garden.ts
vvagaytsev: I'm checking this.

core/src/constants.ts
vvagaytsev: Let's shrink it to 10 minutes :)

core/test/unit/src/garden.ts
vvagaytsev: Yes, that's because of the default timeout value. I'll comment out this assertion and put a TODO note.

vvagaytsev: Ah, my comment was about a different test case. I'll check this again.

vvagaytsev: Yes, the default timeout value is the cause here.

thsig (approved):
I'll leave it to you to squash and clean up, @vvagaytsev!

core/src/plugin/base.ts
vvagaytsev: Declared it as required in https://github.com/garden-io/garden/pull/4066/commits/f7c1cbf18e5c8e062620f57eda21d74d22cba6c9

This schema is internal, and non-value always comes from outside.

PR #4071, opened 2023-04-12
revert: "refactor: rename deploy state property for clarity"

vvagaytsev:
This reverts the commit https://github.com/garden-io/garden/commit/b8bc083bdd796684186bea5ef9ea09d5b0344422 where the `state` field was renamed to `deployState`.

We need this to unblock Cloud <> Core Bonsai testing. We can revisit this later and do the renamings in both Core and Cloud.

cc @edvald @Orzelius 

Fixes #


vvagaytsev:
This change is irrelevant, there were some issues in the Cloud. Closing this.

PR #4074, opened 2023-04-12
fix(container): fix custom image path

janario:
(open this one but fine to discuss a bit more about it)

Fix when using image path with slashes `/`.

Scenario:

garden.yaml
```
kind: Project
name: my-project

environments:
  - name: dev
     providers:
       - name: kubernetes
          deploymentRegistry:
            hostname: <account_id>.dkr.ecr.<region>.amazonaws.com
            namespace: dev/<my-company>
---
kind: Module
type: container
name: my-app
image: my-app/web

---
kind: Module
type: container
name: my-app2
image: my-app2/web

```

The expected image when building/pushing would be:
* host: <account_id>.dkr.ecr.<region>.amazonaws.com
* namespace: dev/<my-company>
* registry: my-app/web

`<account_id>.dkr.ecr.<region>.amazonaws.com/dev/<my-company>/my-app/web`

But turns out that garden parses the container image resulting it as a namespace and ignoring it.

In this MR we join the full name provided from the container since the host/namespace is always enforced.

Fixes #


janario:
Sharing the PR with some proposals to discuss and hear more about it and what would be the best solution for it

core/src/plugins/container/helpers.ts
```diff
@@ -127,10 +127,11 @@ const helpers = {
 
     const host = registryConfig.port ? `${registryConfig.hostname}:${registryConfig.port}` : registryConfig.hostname
 
+    const repository = [parsedId.host, parsedId.namespace, parsedId.repository].filter((s) => s).join("/")
     return helpers.unparseImageId({
       host,
       namespace: registryConfig.namespace,
```
janario, line 133: Another option is to give precedence to the full name at the `image` property

```
  host: parsedId.host || host,
  namespace: parsedId.host || registryConfig.namespace,
  repository: parsedId.repository,
```

that would require me to add the full name in my module:
```
kind: Module
type: container
name: my-app
image: <account_id>.dkr.ecr.<region>.amazonaws.com/dev/<my-company>/my-app/web

```

vvagaytsev:
Thanks for the PR, @janario!
We're quite busy with the upcoming `0.13` release now. Please, sorry about the late reply and slow progress on this.
We'll get back to this PR soon.

10ko:
Hi @janario, thanks for this and sorry if it took so long to reply.
@vvagaytsev would you have some time to review this next week?

twelvemo:
Hi @janario 
This issue has been resolved in the mean time from our end. With our bonsai release (Garden >= 0.13) the configuration options around publishing images have become more fine grained.
Consider  this build action:
```
kind: Build
name: vote
type: container
include: [.]
spec:
  publishId: 12344556.dkr.ecr.eu-west-1.amazonaws.com/dev/bar/vote:latest
  localId: 12344556.dkr.ecr.eu-west-1.amazonaws.com/dev/bar/vote
```

With `publishId` you can explicitly tell garden which host/repository/imageID and tag to use when using the publish command. The tag can be overridden by the `--tag` option on the publish command. See also https://docs.garden.io/other-plugins/container#publishing-images

With `localId` you get the option to specify how your image should be named locally when building locally. This option does not respect custom tags though, as it requires the Garden version hash as the tag at the end. The whole image would look something like: `12344556.dkr.ecr.eu-west-1.amazonaws.com/dev/bar/vote:v-39c56042bc`

I think especially `publishId` should address your initial problem. Do you think this solves your issue?

stefreak:
Until #5585 Garden did not respect the publishId setting for the publish command. That PR will resolve that problem.

PR #4075, opened 2023-04-13
0.13 fix: use default target in sync overrides

Orzelius:
If a target is not specified for a sync override the default target is used.

Also unified the schema and types between targets for sync and sync overrides.

Changed the behavior to throw if target is not set instead of a warning. I don't think it makes sense to just warn here.

### DIFF
--- a/core/src/plugins/kubernetes/commands/sync.ts
+++ b/core/src/plugins/kubernetes/commands/sync.ts
@@ -30,9 +30,6 @@ export const syncStatus: PluginCommand = {
     if (!(await pathExists(dataDir))) {
       log.info(dedent`
         No active sync session found.
-
-        Garden needs to be running in dev mode in this project for sync statuses to
-        be available.
       `)
 
       logSuccess(log)
--- a/core/src/plugins/kubernetes/sync.ts
+++ b/core/src/plugins/kubernetes/sync.ts
@@ -328,7 +328,8 @@ export async function configureSyncMode({
     const target = override.target || defaultTarget
     if (!target) {
       throw new ConfigurationError(
-        `Sync override configuration on ${action.longDescription()} doesn't specify a target, and none is set as a default.`,
+        dedent`Sync override configuration on ${action.longDescription()} doesn't specify a target, and none is set as a default.
+        Either specify a target via the \`spec.sync.overrides[].target\` or \`spec.defaultTarget\``,
         {
           override,
         }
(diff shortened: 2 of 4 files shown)


core/src/plugins/kubernetes/sync.ts
```diff
@@ -204,7 +193,10 @@ export const kubernetesDeploySyncSchema = () =>
         .array()
         .items(kubernetesDeploySyncPathSchema())
         .description("A list of syncs to start once the Deploy is successfully started."),
-      overrides: joi.array().items(syncModeOverrideSpec()),
+      overrides: joi
+        .array()
+        .items(syncModeOverrideSpec())
+        .description("Overrides for the container command and/or arguments for when in sync mode"),
```
vvagaytsev, line 199:
```suggestion
        .description("Overrides for the container command and/or arguments for when in sync mode"),
        .description("Overrides for the container command and/or arguments when in sync mode."),
```

vvagaytsev: Nit: the period is missing at the end of the sentence :) To sure about "for", maybe it's fine to have it there.

core/src/plugins/kubernetes/commands/sync.ts
```diff
@@ -30,9 +30,6 @@ export const syncStatus: PluginCommand = {
     if (!(await pathExists(dataDir))) {
       log.info(dedent`
         No active sync session found.
-
-        Garden needs to be running in dev mode in this project for sync statuses to
```
vvagaytsev, line 34: I'm not sure if we should remove this completely. Maybe, we should just replace "dev mode" with "sync mode"

edvald: I think this is fine actually.

core/src/plugins/kubernetes/sync.ts
edvald: 👍

core/src/plugins/kubernetes/sync.ts
```diff
@@ -332,7 +325,15 @@ export async function configureSyncMode({
   }
 
   for (const override of spec.overrides || []) {
-    const { target } = override
+    const target = override.target || defaultTarget
+    if (!target) {
+      throw new ConfigurationError(
+        `Sync override configuration on ${action.longDescription()} doesn't specify a target, and none is set as a default.`,
```
edvald, line 331: Maybe we can provide more clear instructions here, i.e. which field needs to be filled in.

edvald (changes requested):
Suggested improving the error message, otherwise looks good!

vvagaytsev:
The [check-docs](https://app.circleci.com/pipelines/github/garden-io/garden/16517/workflows/9e77eead-2805-4403-98b5-3bae6d39fd2c/jobs/279717) step failed. Please, re-generate the docs.

vvagaytsev (approved):

Issue #3521, opened 2023-01-17
Create a new example for certificate generation and management


Walther:
We need a new example for certificate generation and management, as the automatic cert-manager integration is deprecated and deleted in 0.13.

Related to:

* https://github.com/garden-io/garden/issues/2551
* https://github.com/garden-io/garden/pull/3517/

ShankyJS:
Hey @Walther please assign me this one

worldofgeese:
@ShankyJS and I are working on a doc [here](https://youngest-kite-tan-preview.motif.land/cert-manager-install-draft). It's just the commands for now. Exposition to come.

PR #4085, opened 2023-04-17
chore: remove cert-manager support

Walther:

Breaking change for `0.13`.

Removes the deprecated `cert-manager` feature. Followup on https://github.com/garden-io/garden/pull/3517 which marked the feature as deprecated in the `0.12` main branch.

The documentation for the old `cert-manager` is deleted, we have new documentation for the new way of using certs coming soon. The new docs are tracked separately in https://github.com/garden-io/garden/issues/3521.

Fixes https://github.com/garden-io/garden/issues/2551

There might be even more things that could get cleaned up, let me know if I've missed something!


vvagaytsev:
Is there a typo in the message header? Should it be`chore!: ...` -> `chore: ...`?

Walther:
the commit message was based on https://www.conventionalcommits.org/en/v1.0.0/ which i was (mistakenly?) under the impression we were following

> BREAKING CHANGE: a commit that has a footer BREAKING CHANGE:, or appends a ! after the type/scope, introduces a breaking API change (correlating with [MAJOR](http://semver.org/#summary) in Semantic Versioning). A BREAKING CHANGE can be part of commits of any type.

that's why i initially only had the `chore!:` part, but i added the `BREAKING CHANGE` body as well, as you requested

vvagaytsev (approved):
Well done! 🚀 Looks good, thank you!

vvagaytsev:
> the commit message was based on https://www.conventionalcommits.org/en/v1.0.0/ which i was (mistakenly?) under the impression we were following
> 
> > BREAKING CHANGE: a commit that has a footer BREAKING CHANGE:, or appends a ! after the type/scope, introduces a breaking API change (correlating with [MAJOR](http://semver.org/#summary) in Semantic Versioning). A BREAKING CHANGE can be part of commits of any type.
> 
> that's why i initially only had the `chore!:` part, but i added the `BREAKING CHANGE` body as well, as you requested. happy to remove the `!` as well, if you want
> 
> EDIT: did another `commit --amend` to remove it

Ah, sorry, and thanks for changing this! I check the commit history, it looked like we have been preferring explicit "BREAKING CHANGE" instead of the exclamation mark in the header :)

PR #4088, opened 2023-04-17
fix(helm): fix status checks around code syncing

thsig:

Before this fix, we were using locally rendered manifests instead of remotely deployed resources for status checks for helm Deploys.  This caused problems when starting/stopping syncs.

We were also using locally rendered manifests to find target resources
for log streaming for Helm Deploys. While this probably didn't cause problems in most cases, it also didn't make proper use of the retry loop in our Helm logging control flow.

A couple of other fixes were also made along the way:

**fix(sync-mode): ensure project mutagen dir exists**

Before this fix, we'd get Mutagen-related errors when listing syncs if the target resource wasn't deployed with sync mode.

**fix(cli):  correctly handle multiple opt aliases**

This fixes an issue with CLI options with more than one alias. Previously, only the last / highest-indexed alias would actually be applied.

This was fixed by passing an array of alias strings to minimist (instead of a single string).

### DIFF
--- a/core/src/cli/helpers.ts
+++ b/core/src/cli/helpers.ts
@@ -158,9 +158,12 @@ export function prepareMinimistOpts({
     if (!_skipDefault) {
       defaultValues[name] = spec.getDefaultValue(cli)
     }
+    if (!aliases[name]) {
+      aliases[name] = []
+    }
 
     for (const alias of spec.aliases || []) {
-      aliases[name] = alias
+      aliases[name].push(alias)
       if (!_skipDefault) {
         defaultValues[alias] = defaultValues[name]
       }
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -417,7 +417,7 @@ ${renderCommands(commands)}
             const userId = (await cloudApi.getProfile()).id
             const commandResultUrl = cloudApi.getCommandResultUrl({ sessionId, userId }).href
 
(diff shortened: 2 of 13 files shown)


examples/vote-helm/vote/garden.yml
```diff
@@ -16,7 +16,7 @@ spec:
   sync:
     paths:
       - containerPath: /app/src
-        sourcePath: ${action.build.vote-image.sourcePath}/src
+        sourcePath: .
```
Orzelius, line 19: How is this supposed to work? The files are in the vote-image folder so what would it be syncing here?

Orzelius (approved):
looks good but was confused about the change in the example

examples/vote-helm/vote/garden.yml
thsig: Ah, good point—not sure how I missed that. Will drop that commit.

core/src/cli/helpers.ts
```diff
@@ -160,7 +160,10 @@ export function prepareMinimistOpts({
     }
 
     for (const alias of spec.aliases || []) {
-      aliases[name] = alias
+      if (!aliases[name]) {
+        aliases[name] = []
+      }
```
vvagaytsev, line 165: This `if`-check can be done outside the inner `for`-loop.

thsig: Right you are—I've updated the commit.

vvagaytsev (approved):

PR #4094, opened 2023-04-18
fix: use the correct secret in pullFromExternalRegistry in kaniko builds

Walther:

Fixes `garden publish` on our example projects, where we use GCR as the registry and `kaniko` build mode. Without this change, `skopeo` invocation during the image pull was using the wrong secret, with no permissions attached.


srihas-g (approved):
🚀

vvagaytsev:
Thank you for fixing this @Walther! Let's also cherry-pick this fix to `0.13`.

PR #4098, opened 2023-04-19
refactor(config): warn on invalid kind instead of failing

edvald:
This is in preparation for a transition to Bonsai (0.13).


vvagaytsev (approved):
LGTM! Thanks 👍

PR #4101, opened 2023-04-19
chore(examples): use local.demo.garden for local kubernetes examples

twelvemo:
Use local.demo.garden domain for local examples instead of local.app.garden.

Fixes #


Walther (approved):
Thank you! 🎉 
Verified locally with `for example in $(ls -1); do pushd $example; echo $example; gdev deploy && gdev test && gdev delete env; popd; done;` and checking that all examples have ingress controllers referring the url `local.demo.garden`. (There were some failures in some examples, but they were completely unrelated to this PR.)

Walther:
We should also port this to the `0.13` branch

vvagaytsev:
> We should also port this to the `0.13` branch

Sure! Also, we have to update the rest examples to use the new domain. That can be addressed in a separate PR.

vvagaytsev (approved):
Thank you! 🚀

PR #4102, opened 2023-04-20
0.13: Mandatory `apiVersion`-field when using action kinds

### LINKED ISSUE #4089: 0.13: update the apiVersion and make it mandatory
### Background

Currently the project configuration schema has an `apiVersion`-field. This field is not mandatory in 0.12 and is available on both the project and module-level.  In order to provide users of garden a migration path, we propose to make the `apiVersion`-field mandatory at the root of the project configuration and to make that the only place where it is used. Further, for 0.13, we suggest to increase the version to `garden.io/v1`, to clearly indicate a schema breaking change. This will make it possible to fail the garden CLI in a safe way including instructions on how to upgrade for the end-user and simplify the detection of the correct garden runtime in an automation pipeline.

### Changes

Limit the `apiVersion`-field to the project config root, make it mandatory and increase the version to `garden.io/v0.13`.  The value of the version-field is a string ending with `v<major.minor version string>`. Until version 1.0, we will increment the minor version on schema changes. The field will be mandatory for the garden `0.13` CLI runtime and optional in `0.12`.

mkhq:

This PR makes the `apiVersion` mandatory for 0.13 clients when using action kinds. Longer discussion in #4089.

Fixes #4089

### DIFF
--- a/core/src/config/base.ts
+++ b/core/src/config/base.ts
@@ -14,7 +14,7 @@ import { pathExists, readFile } from "fs-extra"
 import { omit, isPlainObject, isArray } from "lodash"
 import { coreModuleSpecSchema, baseModuleSchemaKeys, BuildDependencyConfig, ModuleConfig } from "./module"
 import { ConfigurationError, FilesystemError, ParameterError } from "../exceptions"
-import { DEFAULT_API_VERSION } from "../constants"
+import { DEFAULT_API_VERSION, PREVIOUS_API_VERSION } from "../constants"
 import { ProjectConfig, ProjectResource } from "../config/project"
 import { validateWithPath } from "./validation"
 import { defaultDotIgnoreFile, listDirectory } from "../util/fs"
@@ -194,14 +194,16 @@ export function prepareResource({
 
   // Allow the default api version for all kinds except for Project where we expect it to be defined
   // explicitly from 0.13 and forward.
-  if (!spec.apiVersion && spec.kind !== "Project") {
-    spec.apiVersion = DEFAULT_API_VERSION
-  } else {
+  /*if (!spec.apiVersion && spec.kind === "Project") {
     throw new ConfigurationError(
-      `"apiVersion: garden.io/v0.13" is missing in the Project config. The "apiVersion"-field is required from 0.13 (Bonsai). A detailed migration guide is available at https://docs.garden.io/v/bonsai-release/tutorials/migrating-to-bonsai.`,
+      `"apiVersion" is missing in the Project config. The "apiVersion"-field is required from 0.13 (Bonsai). Use "${PREVIOUS_API_VERSION}" for backwards compatibility with 0.12 or "${DEFAULT_API_VERSION}" for 0.13. A detailed migration guide is available at https://docs.garden.io/v/bonsai-release/tutorials/migrating-to-bonsai.`,
       { spec, path: configFilePath }
     )
   }
+
+  if (!spec.apiVersion) {
+    spec.apiVersion = DEFAULT_API_VERSION
+  }*/
 
   const basePath = dirname(configFilePath)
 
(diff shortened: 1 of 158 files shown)


TimBeyer (approved):

core/src/garden.ts
```diff
@@ -1262,6 +1266,15 @@ export class Garden {
       let actionsCount = 0
 
       for (const kind of actionKinds) {
+        // Verify that the project apiVersion is defined as compatible with action kinds
+        // This is only available with apiVersion `garden.io/v1` or newer.
+        if (this.projectApiVersion !== DEFAULT_API_VERSION) {
+          throw new ConfigurationError(
+            `Action kinds are only supported in project configurations with "apiVersion: ${DEFAULT_API_VERSION}". A detailed migration guide is available at https://docs.garden.io/v/bonsai-release/tutorials/migrating-to-bonsai.`,
```
stefreak, line 1273: I can imagine that this link might break once bonsai goes GA.

Should we maybe link to https://docs.garden.io/tutorials/migrating-to-bonsai and redirect / link to to the bonsai release guide from there?

docs/reference/module-types/terraform.md
```diff
@@ -193,11 +193,11 @@ workspace:
...
 
-The schema version of this config (currently not used).
+The schema version of this config.
 
-| Type     | Allowed Values | Default          | Required |
-| -------- | -------------- | ---------------- | -------- |
-| `string` | "garden.io/v0" | `"garden.io/v0"` | Yes      |
+| Type     | Allowed Values                 | Default          | Required |
+| -------- | ------------------------------ | ---------------- | -------- |
+| `string` | "garden.io/v0", "garden.io/v1" | `"garden.io/v1"` | Yes      |
```
stefreak, line 200: I think the default should be v0 to meet the requirements described in https://github.com/garden-io/garden/issues/4089#issuecomment-1527471349

core/src/config/common.ts
```diff
@@ -856,12 +856,10 @@ export const moduleVersionSchema = createSchema({
...
-export const apiVersionSchema = () =>
-  joi
-    .string()
-    .default(DEFAULT_API_VERSION)
-    .valid(DEFAULT_API_VERSION)
-    .description("The schema version of this config (currently not used).")
+export const apiVersionSchema = () => apiVersionSchemaWithoutDefault().default(DEFAULT_API_VERSION)
+
+export const apiVersionSchemaWithoutDefault = () =>
+  joi.string().valid(PREVIOUS_API_VERSION, DEFAULT_API_VERSION).description("The schema version of this config.")
```
stefreak, line 862: The name is slightly confusing– it's called schema without default, but there is a default

core/src/config/base.ts
```diff
@@ -14,7 +14,7 @@ import { pathExists, readFile } from "fs-extra"
 import { omit, isPlainObject, isArray } from "lodash"
 import { coreModuleSpecSchema, baseModuleSchemaKeys, BuildDependencyConfig, ModuleConfig } from "./module"
 import { ConfigurationError, FilesystemError, ParameterError } from "../exceptions"
-import { DEFAULT_API_VERSION } from "../constants"
+import { DEFAULT_API_VERSION, PREVIOUS_API_VERSION } from "../constants"
```
stefreak, line 17: Instead of constants, maybe we could define an enum and reference it like 

```
enum APIVersion {
  v0 = "garden.io/v0"
  v1 = "garden.io/v1"
}
```

I think that would be slightly more readable 👍

docs/reference/module-types/terraform.md
stefreak: OK, this is the module 🤔 Sorry, didn't notice that

core/src/config/common.ts
stefreak: Ah ok, there is actually no default

core/src/garden.ts
mkhq: Updated the links to https://docs.garden.io/tutorials/migrating-to-bonsai

stefreak (approved):
👍 Let's merge and try it out. We can refine it in following PRs

PR #4129, opened 2023-04-27
feat: `garden community` command

Walther:

Adds a new command `garden community` that opens the Garden Community Discord invite link.

Completes the task defined on the Q2 ideas Miro board.

Sorted the imports in `commands.ts` as a separate commit, as they were _almost_ sorted already.

### DIFF
--- a/core/src/commands/community.ts
+++ b/core/src/commands/community.ts
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) 2018-2023 Garden Technologies, Inc. <info@garden.io>
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { Command, CommandResult } from "./base"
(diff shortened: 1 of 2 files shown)


Orzelius (approved):

core/src/commands/community.ts
```diff
@@ -0,0 +1,38 @@
...
+  loggerType: "basic"
+
+  noProject = true
+
+  printHeader() {}
+
+  async action(): Promise<CommandResult> {
+    const discordInvite = "https://discord.gg/FrmhuUjFs6"
+    // eslint-disable-next-line no-console
+    console.log(discordInvite)
```
vvagaytsev, line 30: A colored message might be a better option here. Something like "Welcome to our community! Join us on Discord: {link}".

Walther: My initial thought was to keep it simple and only output the link in order to make it easier to copy/paste for folks who are not using a terminal emulator that supports clickable link detection. The command is also calling an `open` handler, so the printed line is more of a backup method anyway, and leaves a trail of history for what was opened.

But I have no strong opinions here - we can also tweak this further if we want to, easy changes to make ❤️

Issue #4067, opened 2023-04-11
0.13: Update docs aggregate 


Orzelius:
This is a list of docs that need to be updated. Feel free to put your name behind a doc or a section if you feel like working on any of these.

I will try to take care of the rest.

- [x] advanced
  - [x] [advanced/cert-manager-integration.md](https://github.com/garden-io/garden/tree/0.13/docs/advanced/cert-manager-integration.md) to be removed -> removed in #4085
  - [x] [advanced/custom-commands.md](https://github.com/garden-io/garden/tree/0.13/docs/advanced/custom-commands.md) @Walther  https://github.com/garden-io/garden/pull/4325
    - [x] also remove experimental flag
  - [x] [advanced/using-remote-sources.md](https://github.com/garden-io/garden/tree/0.13/docs/advanced/using-remote-sources.md)
- [x] basics
  - [x] [basics/how-garden-works.md](https://github.com/garden-io/garden/tree/0.13/docs/basics/how-garden-works.md)
  - [x] [basics/quickstart.md](https://github.com/garden-io/garden/tree/0.13/docs/basics/quickstart.md) @ShankyJS
    - [x] also update https://github.com/garden-io/quickstart-example @worldofgeese 
  - [x] [basics/stack-graph.md](https://github.com/garden-io/garden/tree/0.13/docs/basics/stack-graph.md)
- [x] guides
  - [x] [guides/adopting-garden.md](https://github.com/garden-io/garden/tree/0.13/docs/guides/adopting-garden.md)
  - [x] [guides/code-synchronization.md](https://github.com/garden-io/garden/tree/0.13/docs/guides/code-synchronization.md)
  - [x] [guides/hot-reload.md](https://github.com/garden-io/garden/tree/0.13/docs/guides/hot-reload.md) to be removed
  - [x] [guides/migrating-from-docker-compose.md](https://github.com/garden-io/garden/tree/0.13/docs/guides/migrating-from-docker-compose.md) -> https://github.com/garden-io/garden/pull/4235
    - [x] Also update https://github.com/garden-io/garden-docker-compose -> https://github.com/garden-io/garden-docker-compose/pull/2
  - [x] [guides/namespaces.md](https://github.com/garden-io/garden/tree/0.13/docs/guides/namespaces.md)
  - [x] [guides/running-service-in-local-mode.md](https://github.com/garden-io/garden/tree/0.13/docs/guides/running-service-in-local-mode.md) 
- [x] k8s-plugins @Orzelius 
[...]

PR #4130, opened 2023-04-27
docs: update terraform-plugin docs for 0.13

Walther:

Part of https://github.com/garden-io/garden/issues/4067

WIP, some open questions

### DIFF
--- a/docs/terraform-plugin/about.md
+++ b/docs/terraform-plugin/about.md
@@ -11,13 +11,13 @@ It's strongly recommended that you [learn about Terraform](https://www.terraform
 
 ## How it works
 
-Under the hood, Garden simply wraps Terraform, so there's no magic involved. Garden just automates its execution and makes stack outputs available to your Garden providers and modules.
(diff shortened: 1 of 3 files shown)


Orzelius:
about the module question:
Just have to figure out if the command meant terraform modules or garden modules. If garden modules it should be changed to "action" or just make the commands `terraform plan` or `terraform apply`. If tf modules keep as is and add doc that says so.

Walther:
@Orzelius forgot to paste this to the description - it seems that the current commands (and their help descriptions) are mixing module and action terminology

https://github.com/garden-io/garden/blob/719f6c63294d2f1b631ac368a03ddc699d6fc592/plugins/terraform/commands.ts#L71-L77

Orzelius:
The problem is that terraform also has modules of it's own and I'm not sure without digging into the code which ones are meant here

Walther:
Got my branches in a twist while making https://github.com/garden-io/garden/pull/4293
Closing this and making another one, sorry 🙇‍♂️

PR #4137, opened 2023-04-28
chore: update genereated docs

thsig:


Orzelius (approved):

Walther:
thanks for the fix! my bad for missing the `check-docs` step in the previous PR

PR #4139, opened 2023-04-28
0.13 sync mode fixes

vvagaytsev:
Thanks for pairing on this @Walther and @thsig!

Fixes #

### DIFF
--- a/core/src/tasks/delete-deploy.ts
+++ b/core/src/tasks/delete-deploy.ts
@@ -35,6 +35,14 @@ export class DeleteDeployTask extends BaseActionTask<DeployAction, DeployStatus>
     this.deleteDeployNames = params.deleteDeployNames || [params.action.name]
   }
 
+  protected getDependencyParams(): DeleteDeployTaskParams {
+    return {
+      ...super.getDependencyParams(),
+      dependantsFirst: this.dependantsFirst,
+      deleteDeployNames: this.deleteDeployNames,
+    }
+  }
+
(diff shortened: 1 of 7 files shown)


thsig:
I'd like to read this one through properly before merging, will let you know when that's done.

core/src/tasks/delete-deploy.ts
```diff
@@ -52,7 +60,7 @@ export class DeleteDeployTask extends BaseActionTask<DeployAction, DeployStatus>
 
     const depTasks = deps.filter(isDeployAction).map((action) => {
       return new DeleteDeployTask({
-        ...this.getBaseDependencyParams(),
+        ...this.getDependencyParams(),
         action,
         force: this.force,
         deleteDeployNames: this.deleteDeployNames,
```
thsig, line 66: Since we're not changing the value of `deleteDeployNames` from the one set `getDependencyParams`, I think we should skip it here.

I.e. there's no need to override the defaults if we know the value is going to be the same.

thsig (changes requested):
Looks good! Only had one comment there.

core/src/tasks/delete-deploy.ts
vvagaytsev: @thsig yes, but if it's an idempotent change, why don't just keep it? It will allow us to avoid more bugs If we allow changing a subset of action-specific params in the future.

thsig (approved):

PR #4146, opened 2023-05-02
0.13 fix: throw when test or run is not found

Orzelius:


Walther (dismissed):
Thanks for the quick fix!

core/src/commands/run.ts
```diff
@@ -133,10 +133,11 @@ export class RunCommand extends Command<Args, Opts> {
     }
 
     let actions = graph.getActionsByKind("Run", {
-      includeNames,
+      names,
```
edvald, line 136: This approach removes the ability to use wildcards in names, so it won't work as-is. Here's one alternative approach: https://github.com/garden-io/garden/blob/d21b826cf981300778815ebf73260d1f7cfd6857/core/src/commands/sync/sync-start.ts#L112

Overall I'd suggest that we should A) throw if a name _without_ a wildcard is passed in but it does not exist, and B) warn if no actions are matched.

edvald (changes requested):

core/src/commands/run.ts
Orzelius: Yeah that was an oversight on my part. Implemented as suggested in [3765085](https://github.com/garden-io/garden/pull/4146/commits/37650859eb0d8221ca6d31ecb00405a7c71bc650)

edvald (approved):
Looks good!

Orzelius:
I think this is good to go. Integ test failures are unrelated and are being fixed separately.

Issue #3739, opened 2023-02-13
Stack Graph v2 examples update


vvagaytsev:
Goal: update all examples to use new action-based configs.

- [x] #3741
- [x] #3754
- [x] #3815
- [x] #3785
- [x] demo-project -> done in https://github.com/garden-io/garden/commit/a2320fc704b3a9a05b822e865be2620d05e6fa6f
- [x] demo-project-start -> no garden configs, the relevant getting started guide will be updated in a separate PR
- [x] deployment-strategies -> this has been deprecated in #3638, the example was removed in #3825
- [x] #3796
- [x] #3832
- [x] #3882
- [x] #3787
- [x] #3883
- [x] #3782
- [x] #3801
- [x] #3809
- [x] #3819
- [x] #3829
- [x] #3775
- [x] #3750
- [x] #3752
- [x] #3769
- [x] #3780
- [x] #3789
- [x] #3804
- [x] #3826
- [x] #3846
- [x] #3847
- [x] #3848
- [x] #3740
- [x] #3766
- [x] #3849
- [x] #3881
- [x] #3765
- [x] #3757
- [x] #3767
- [x] #4077

vvagaytsev:
🎉

10ko:
Can this be closed?

shumailxyz:
Should have been automatically closed as it was marked done in project. Closing now.

PR #4150, opened 2023-05-02
feat(core): support remote repositories for action sources

edvald:
This also fixes the `source.path` field, which was present in action schemas but not handled in code.

TODO:

- [ ] Update examples to use remote actions instead of modules
- [ ] Update remote sources docs guide


core/src/commands/link/action.ts
```diff
@@ -0,0 +1,94 @@
...
+  help = "Link a remote action to a local directory."
+  arguments = linkActionArguments
+
+  outputsSchema = () =>
+    joi.object().keys({
+      sources: joiArray(linkedActionSchema()).description("A list of all locally linked remote actions."),
+    })
+
+  description = dedent`
+    After linking a remote action, Garden will read the source from the linked local directory instead of from the remote repository. Garden can only link actions that have a remote source, i.e. actions that specifiy a \`source.repository.url\` in their configuration.
```
Orzelius, line 54:
```suggestion
    After linking a remote action, Garden will read the source from the linked local directory instead of the remote repository. Garden can only link actions that have a remote source, i.e. actions that specify a \`source.repository.url\` in their configuration.
```

Orzelius (approved):
I think this can be merged.
The docs update is tracked here: https://github.com/garden-io/garden/issues/4067
And examples here: https://github.com/garden-io/garden/issues/3739

PR #4151, opened 2023-05-02
improvement(k8s): reduce number of k8s API calls during status monitoring

edvald:
This relatively simple change substantially reduces API requests during status polling, particularly when waiting for deployments to complete.


Walther (approved):
nice simplification! ✨ 🧹 
verified locally

PR #4155, opened 2023-05-02
0.13: fix some integ tests

edvald:
See commits. There's one proper bug fix, rest are test fixes.


core/test/integ/src/plugins/container/container.ts
```diff
@@ -51,16 +54,14 @@ describe("plugins.container", () => {
...
-      containerHelpers,
-      getPropertyName(containerHelpers, (h) => h.actionHasDockerfile),
-      () => true
-    )
-    td.replace(
-      containerHelpers,
-      getPropertyName(containerHelpers, (h) => h.dockerCli),
-      () => ({ all: "test log", stdout: "some/image:12345" })
-    )
+    sinon.replace(containerHelpers, "actionHasDockerfile", async () => true)
```
Orzelius, line 57: I see you don't much like the `getPropertyName` helper

Orzelius (approved):
Thank you

core/test/integ/src/plugins/container/container.ts
edvald: It is not an option in this context, the Sinon types only allow constants, and that's in turn type-safe.

(But also, yes, that always felt convoluted haha :P)

PR #4157, opened 2023-05-03
fix osx install deps script and add pointer about Rosetta

shumailxyz:
- Fixes the error regarding brew install `kubernetes-helm` as its now renamed to helm. 
- Adds pointer about Rosetta requirement in installation instructions for Apple Silicon macs. In case Rosetta is not installed, user faces: `bad CPU type in executable` while using `garden` in terminal.

Fixes #

### DIFF
--- a/docs/basics/quickstart.md
+++ b/docs/basics/quickstart.md
@@ -34,6 +34,11 @@ Then install the Garden CLI for your platform:
 ```sh
 brew tap garden-io/garden && brew install garden-cli
 ```
+
+{% hint style="info" %}
+For a Mac computer with Apple silicon, Garden needs [Rosetta](https://support.apple.com/en-us/HT211861).
+{% endhint %}
+
 {% endtab %}
 
 {% tab title="Linux" %}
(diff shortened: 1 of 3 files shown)


Orzelius (approved):

Walther (changes requested):
Thanks for the update!

I'm not 100% sure if we need Rosetta for something - I just got the new work laptop with apple silicon and can run the `garden` binary from a `yarn build` fully natively without Rosetta installed. Let's double-check if we need it

Walther (approved):

stefreak:
> Thanks for the update!
> 
> ~I'm not 100% sure if we need Rosetta for something - I just got the new work laptop with apple silicon and can run the `garden` binary from a `yarn build` fully natively without Rosetta installed. Let's double-check if we need it - ideally we should be providing fully native arm builds soon <3~
> 
> EDIT: scratch that, let's keep the native arm support as a separate task for later

@Walther for yarn dev you are absolutely right and it already runs natively (since we moved to node 18). But if you install a released version of garden, it'll require rosetta at the moment, as we do not publish ARM builds.

shumailxyz:
yeah, it 's only an issue if you install released version. I had fresh laptop, installed using homebrew and ran into error because Rosetta wasn't installed.

vvagaytsev (approved):
Nice! Thank you! 👍

PR #4160, opened 2023-05-03
0.13: improve task logging

Orzelius:
more details in the commits but I made the changes with the goal of making everything as easy to understand as possible.

Feel free to disagree with my changes and lay out your thoughts.

This also fixes the issue where the logs said an action is already complete but then went and executed it anyway when the `--force` flag was provided.

### DIFF
--- a/core/src/tasks/build.ts
+++ b/core/src/tasks/build.ts
@@ -29,8 +29,8 @@ export class BuildTask extends ExecuteActionTask<BuildAction, BuildStatus> {
     const output = await router.build.getStatus({ log: this.log, graph: this.graph, action })
     const status = output.result
 
-    if (status.state === "ready" && !statusOnly) {
-      this.log.info(`${action.longDescription()} already complete, nothing to do.`)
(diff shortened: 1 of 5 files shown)


Walther (commented):
Looks good as far as I can tell, nice cleanup!
I'll let others add their comments too, in case there are differing opinions on the formatting etc

core/src/tasks/build.ts
```diff
@@ -46,18 +46,7 @@ export class BuildTask extends ExecuteActionTask<BuildAction, BuildStatus> {
       )
     }
 
-    const log = this.log.info(`Building version ${action.versionString()}...`)
-    // const newActionLog = log.createLog()
```
eysi09, line 50: I think that's mine. Apologies :)

eysi09 (approved):
Thanks for the update, LGTM! 

The idea behind the `ActionLog` was exactly so that we didn't have to do things like `action.longDescription`, so thanks for cleaning that up! 

And in general, if you see ugly logs, don't hesitate to make pretty.

PR #4163, opened 2023-05-03
docs: add sync mode detection example to migration guide

Orzelius:


docs/tutorials/migrating-to-bonsai.md
```diff
@@ -47,6 +47,19 @@ The general flow of the Garden runtime is as follows:
 - Modules are converted to actions
 - Actions are resolved
 
+In 0.12 often string templating was used to detect sync mode and change behaviour accordingly. This can be done much easier with Bonsai via the
+[`${actions.deploy.<name>.mode}`](https://github.com/garden-io/garden/blob/0.13/docs/reference/action-types/Deploy/container.md#actionsdeploynamemode)
```
Walther, line 51: Should we add a small yaml code block demonstrating how to use this as well?

My thinking: people are used to copypasting configs from code blocks, if they only see the workaround below, they might just use it instead.

We could make it very explicit and clear - here's how to do this in pure 0.13 config / here's how to do this in a mixed 0.12/0.13 environment

docs/tutorials/migrating-to-bonsai.md
```diff
@@ -47,6 +47,19 @@ The general flow of the Garden runtime is as follows:
...
+but that would not be backward compatible with 0.13. Below is an example config that changes the `targetImage` based on the mode the deploy is running in
+that works with both Bonsai and 0.12.
+
+```yml
+variables:
+  sync-mode: ${command.params contains 'sync' && (command.params.sync contains 'api' || isEmpty(command.params.sync))} 
+  dev-mode: ${command.name == 'dev' || (command.params contains 'dev-mode' && (command.params.dev-mode contains 'api' || isEmpty(command.params.dev-mode)))}
+...
+targetImage: "${var.sync-mode || var.dev-mode ? 'dev-mode' : var.target-image}"
+```
```
stefreak, line 61: I'd suggest not using `targetImage` in the example, because it might not be clear at first glance what's the code that is part of detection, and what is just an instruction how to use it.

````suggestion
```

Sync mode or dev mode have been requested if either `var.sync-mode` or `var.dev-mode` are true. You can use a template expression like `${var.sync-mode || var.dev-mode ? 'yes' : 'no'}` to change the behaviour of your actions or modules.
````

Comment (Walther) [reply docs/tutorials/migrating-to-bonsai.md]: Agree on not using a literal called `dev-mode` and a variable `var.target-image` in the conditional.

At least personally, I would still prefer to have some sort of a practical config code block here, and a sentence explaining it - can be a bit easier to parse than full sentences with inline code

File: docs/tutorials/migrating-to-bonsai.md [hunk 3]
```diff
@@ -47,6 +47,21 @@ The general flow of the Garden runtime is as follows:
...
+This can be done much easier with Bonsai via the
+[`${actions.deploy.<name>.mode}`](../reference/action-types/Deploy/container.md#actionsdeploynamemode)
+template string or `${this.mode}` if referenced in the action itself,
+but that would not be backward compatible with 0.13. Below is an example config that changes the `targetImage` based on the mode the deploy is running in
+that works with both Bonsai and 0.12.
+
+```yml
+variables:
+  sync-mode: ${command.params contains 'sync' && (command.params.sync contains 'api' || isEmpty(command.params.sync))} 
+  dev-mode: ${command.name == 'dev' || (command.params contains 'dev-mode' && (command.params.dev-mode contains 'api' || isEmpty(command.params.dev-mode)))}
```
stefreak, line 60: Sorry for the back and forth, but I noticed that it is not super-obvious that `'api'` is the module name.

Also I think, looking at the entire migration guide, this is oddly specific at this point in the document– most people probably don't need or want to know this.

Maybe we can add a new Section at the end of the migration guide, either as "new features" or "FAQ", "Tips and tricks" or something like that?

Orzelius:
rebased based on the latest 0.13

docs/tutorials/migrating-to-bonsai.md
```diff
@@ -310,3 +310,32 @@ spec:
...
 </div>
+
+## FAQ
+
+### 
+
+### Where is the documentation for modules?
+
+The reference documentation can be found [here](../reference/module-types/README.md),
+but all other documentation has been rewriten to be action specific. If you need to keep working
```
vvagaytsev, line 321: Should we add a link to the actions guide here?

vvagaytsev (dismissed):
Left a non-blocking comment. Looks good, thanks for updating this! 👍

stefreak (approved):
LGTM 👍

Orzelius:
I'm sorry but I require an approve once more as there was a conflict with the main branch

vvagaytsev (approved):

PR #4164, opened 2023-05-03
chore(pulumi): make pulumi backendURL and orgName as non-required fields

srihas-g:
Initialize Pulumi's backendURL and orgName with default values. Also make them as non-required config options.

Fixes #

### DIFF
--- a/docs/pulumi-plugin/configure-provider.md
+++ b/docs/pulumi-plugin/configure-provider.md
@@ -13,4 +13,47 @@ providers:
   - name: pulumi # <----
   ...
 ```
+
+In case you want to use different backends for different Garden environments you can configure your provider and modules follows. This example uses two
+different pulumi backends. In the `dev` environment it uses a self-managed state backend, in this case an S3 bucket which is specified
+with the `backendURL`.
+In the `prod` environment it uses pulumi managed state backend, which is the default so we don't need to specify a `backendURL`. 
+
+Note that when you use a self managed state backend, Garden's module level `cacheStatus` needs to be set to `false`, since 
+caching is only available with the pulumi managed state backend. The same applies to `orgName` which only makes sense in the context of the pulumi managed state backend.
+Please ensure that `orgName` is set to `null` or empty string `""` for all the environments that are not using the pulumi managed state backend.
+
+```yaml
+---
+kind: Project
+name: pulumi
+defaultEnvironment: dev
+variables:
+  cacheStatus: true
+environments:
+  - name: dev
+    variables:
+      backendURL: s3://<bucket-name>
+      cacheStatus: false # cacheStatus has to be set to false for self-managed state backends
+  - name: prod
+    variables:
+      orgName: garden
+providers:
+  - name: pulumi
+    environments: [dev, prod]
+    orgName: ${var.orgName || null} # ensure orgName is null or "" for self-managed state backends
+    backendURL: ${var.backendURL || null} # defaults to Pulumi managed state backend if null or ""
+
+---
+kind: Module
+type: pulumi
+name: aws-s3
+stack: ${environment.name}
+createStack: true
+cacheStatus: ${var.cacheStatus} # cacheStatus has to be set to false for self-managed state backends
+description: Creates an s3 bucket
+pulumiVariables:
+  environment: ${environment.name}
+```
+
 There are several configuration options you can set on the provider—see the [reference docs for the pulumi provider](../reference/providers/pulumi.md) for details.
--- a/plugins/pulumi/config.ts
+++ b/plugins/pulumi/config.ts
@@ -47,11 +47,12 @@ export const pulumiProviderConfigSchema = providerConfigBaseSchema()
         This option can be useful when you want to provide a folder of pre-approved pulumi plans to a CI pipeline step.
     `
       ),
-    orgName: joi.string().description(dedent`
+    orgName: joi.string().optional().empty(["", null]).description(dedent`
       The name of the pulumi organization to use. This option can also be set on the module level, in which case it
-      overrides this provider-level option.
+      overrides this provider-level option. Note that setting the organization name is only necessary when using
+      pulumi managed backend with an organization.
     `),
-    backendURL: joi.string().uri().default("https://api.pulumi.com").required().description(dedent`
+    backendURL: joi.string().optional().uri().empty(["", null]).default("https://api.pulumi.com").description(dedent`
       The URL of the state backend endpoint used. This option can also be set on the module level, in which case it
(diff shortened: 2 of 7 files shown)


plugins/pulumi/config.ts
```diff
@@ -17,7 +17,7 @@ import { defaultPulumiVersion, supportedVersions } from "./cli"
 type PulumiProviderConfig = GenericProviderConfig & {
   version: string | null
   previewDir: string | null
-  orgName?: string
+  orgName: string
```
twelvemo, line 20: Shouldn't `orgName` stay optional too?

twelvemo (changes requested):

plugins/pulumi/config.ts
```diff
@@ -47,11 +47,11 @@ export const pulumiProviderConfigSchema = providerConfigBaseSchema()
         This option can be useful when you want to provide a folder of pre-approved pulumi plans to a CI pipeline step.
     `
       ),
-    orgName: joi.string().description(dedent`
+    orgName: joi.string().default(null).allow(null).description(dedent`
```
stefreak, line 50: why null? orgName was already optional, without `default(null).allow(null)`

plugins/pulumi/config.ts
stefreak: If orgName is null or a string (like defined in joi) it should be
```suggestion
  orgName: string | null
```

but I think it would be better to avoid null if possible, and keep using undefined in case it isn't defined (keep `orgName?: string`

plugins/pulumi/helpers.ts
```diff
@@ -435,11 +435,11 @@ export async function selectStack({ module, ctx, provider, log }: PulumiParams)
...
 }
 
-function getOrgName(provider: PulumiProvider, module: PulumiModule): string | null {
-  if (module.spec.orgName || module.spec.orgName === null) {
+function getOrgName(provider: PulumiProvider, module: PulumiModule): string {
+  if (module.spec.orgName) {
     return module.spec.orgName
   } else {
-    return provider.config.orgName || null
+    return provider.config.orgName
```
stefreak, line 442: 👍 This looks much better than before. The code before looks strange actually 👀

plugins/pulumi/config.ts
```diff
@@ -47,11 +47,11 @@ export const pulumiProviderConfigSchema = providerConfigBaseSchema()
...
       The name of the pulumi organization to use. This option can also be set on the module level, in which case it
       overrides this provider-level option.
     `),
-    backendURL: joi.string().uri().default("https://api.pulumi.com").required().description(dedent`
+    backendURL: joi.string().uri().default("https://api.pulumi.com").description(dedent`
```
stefreak, line 54: Ooops, good catch! I'd suggest we add tests for the pulumi config validation. That should be pretty easy. Let's pair on this @srihas-g

stefreak (changes requested):

plugins/pulumi/config.ts
srihas-g: I'd love to do it in another PR and get the breaking changes out of main branch first.

plugins/pulumi/handlers.ts
```diff
@@ -49,28 +49,6 @@ export const configurePulumiModule: ModuleActionHandlers<PulumiModule>["configur
...
-  if (!backendUrl.startsWith("https://")) {
-    if (orgName) {
-      throw new ConfigurationError("Pulumi: orgName is not supported for self-managed backends", {
-        moduleConfig,
-        providerConfig: provider.config,
-      })
-    }
-
-    if (moduleConfig.spec.cacheStatus) {
-      throw new ConfigurationError("Pulumi: `cacheStatus: true` is not supported for self-managed backends", {
```
stefreak, line 67: what happened to these checks? Where do they live now?

stefreak: I think they are important, because otherwise I remember there were some really cryptic error messages when configuring these things

twelvemo (approved):
Looks good to me but since I also worked on this one, would also like to see @stefreak 's review.

docs/pulumi-plugin/configure-provider.md
```diff
@@ -13,4 +13,48 @@ providers:
...
+with the `backendURL`.
+In the `prod` environment it uses pulumi managed state backend, which is the default so we don't need to specify a `backendURL`. 
+
+Note that when you use a self managed state backend, Garden's module level `cacheStatus` needs to be set to `false`, since 
+caching is only available with the pulumi managed state backend. The same applies to `orgName` which only makes sense in the context of the pulumi managed state backend.
+Please ensure that `orgName` is set to `null` or empty string `""` for all the environments that are not using the pulumi managed state backend.
+
+```yaml
+---
+kind: Project
```
stefreak, line 28: I find the example slightly confusing:

Instead of declaring defaults for every variable above, and then overriding e.g. using `orgName: null`, I think it would be easier to read if only non-null variables are declared in every environment, and then you add `|| null` in the provider configuration.

plugins/pulumi/config.ts
```diff
@@ -131,15 +131,10 @@ export const pulumiModuleSchema = () =>
           config. Simply specify all the config variables at the top level.
         `
     ),
-    orgName: joi.string().optional().allow(null).description(dedent`
+    orgName: joi.string().optional().empty(["", null]).description(dedent`
       The name of the pulumi organization to use. Overrides the \`orgName\` set on the pulumi provider (if any).
       To use the default org, set to null.
     `),
-    backendURL: joi.string().uri().optional().allow(null).description(dedent`
```
stefreak, line 138: Why do you remove backendURL here? I looked into the code that executes the pulumi command, and we correctly pass env as a dictionary to the exec function, so there should be no conflict with multiple pulumi processes concurrently.

stefreak (changes requested):
Cool, good work! I think the example code can be slightly improved to make it easier to read and understand. 

Also I'm confused as to why backendURL in the module definition has been removed; I had a look into the code and the exec function should have no issues running multiple processes with different sets of environment variables concurrently.

plugins/pulumi/config.ts
srihas-g: We faced some misbehaviour with Pulumi stack select based on tags when multiple backends are in use, hence we are going ahead with one backend for the provider for the entire env approach first and then get back on this one as an added feature.

docs/pulumi-plugin/configure-provider.md
srihas-g: agreed, updated the example

stefreak (approved):

stefreak (approved):

PR #4168, opened 2023-05-04
test(workflows): add test case for lazy initialization

stefreak:
To make it possible to prepare kubeconfigs and other prerequisites to
provider initialization in a workflow, we want to manifest this
behaviour using tests.

Fixes #


Walther (approved):
Thank you!

PR #4179, opened 2023-05-05
chore: 0.13 analytics improvements

mkhq:

Adapt the analytics to 0.13 by including project stats for action kinds.

Fixes #


core/src/analytics/analytics.ts
```diff
@@ -92,6 +94,11 @@ interface ProjectMetadata {
   servicesCount: number
   testsCount: number
   moduleTypes: string[]
+  actionsCount: number
+  actionBuildsCount: number
```
eysi09, line 98: Nitpick but feels like this should be called `buildActionCount`, `deployActionCount`, etc

core/src/analytics/analytics.ts
```diff
@@ -125,6 +132,7 @@ interface CommandEvent extends EventBase {
   type: "Run Command"
   properties: PropertiesBase & {
     name: string
+    commandIteration: number
```
eysi09, line 135: I think we should skip this for now since we're changing this behaviour in general:

- Introduce the concept of a parent session (with a `parentSessionId`)
- This would be a persistent session like `garden dev`
- Every time you run a command within that session, via the terminal / Cloud / Desktop, that gets counted as a single session within the parent session. 
- With this, a command can't technically be run twice. Each session is exactly a single command but it can have a parent command. Only persistent commands like `garden dev` can be parent commands.

eysi09 (changes requested):
Two comments, otherwise looks great! 🙏

eysi09 (dismissed):

mkhq:
@eysi09 I've re-enabled the failing nock tests, if this turns out to be an issue still we can re-add the skip.

mkhq:
> @eysi09 I've re-enabled the failing nock tests, if this turns out to be an issue still we can re-add the skip.

Added the skip back because those tests keep on failing.

eysi09 (approved):

PR #4180, opened 2023-05-06
feat(core): --logs option for deploy command

thsig:

When used, the `--logs` option streams service logs during deployment, and then keeps the command running until the user exits it.

This is useful for combining deployment with log streaming.

We should only be streaming logs from the current time onwards, but the logs monitor currently doesn't support the `since` option. I didn't add it back in since IIRC @eysi09 is working on a branch that addresses that.

### DIFF
--- a/core/src/commands/deploy.ts
+++ b/core/src/commands/deploy.ts
@@ -33,6 +33,8 @@ import { HandlerMonitor } from "../monitors/handler"
 import { GraphResultFromTask } from "../graph/results"
 import { PortForwardMonitor } from "../monitors/port-forward"
 import { registerCleanupFunction } from "../util/util"
+import { LogMonitor } from "../monitors/logs"
+import { parseLogLevel } from "../logger/logger"
 
 export const deployArgs = {
   names: new StringsParameter({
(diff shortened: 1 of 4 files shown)


core/src/commands/deploy.ts
```diff
@@ -9,6 +9,7 @@
 import deline = require("deline")
 import dedent = require("dedent")
 import chalk = require("chalk")
+// import figures from "figures"
```
eysi09, line 12: Remove?

thsig:
Here's a screenshot:

Screenshot 2023-05-06 at 15 44 40

core/test/unit/src/commands/deploy.ts
```diff
@@ -135,6 +135,8 @@ describe("DeployCommand", () => {
         "skip-dependencies": false,
         "skip-watch": false,
         "forward": false,
+        "logs": false,
```
eysi09, line 138: Should we test the case were `logs=true`?

eysi09: ...or are all of these for non-persistent runs?

eysi09 (changes requested):
Couple of comments, otherwise looks great!

core/test/unit/src/commands/deploy.ts
thsig: These are all for non-persistent runs—I'm not sure how much value there is in testing this functionality specifically (all the interesting parts happen inside the logs monitor and the plugin handlers under that).

eysi09 (approved):

PR #4185, opened 2023-05-07
fix(server): ensure fresh command instance per request

eysi09:

Previously we were re-using the same command instance for every server request (created at start time via `getCoreCommands()`).

This meant that multiple requests would all subscribe to the same instance.

Concretely, if request A, asks for logs for service X and request B, asks for logs for service Y, they would both subscribe to the same logs command.

Now, when service X writes logs, the callbacks for both request A and request B would be called and to the end user it looks like both services X and service Y are writing the logs.

**UPDATE:**

This first one didn't quite do the trick. There's a follow up commit with following commit message:

This is a follow up to: https://github.com/garden-io/garden/commit/ac54b93ef8680f12e86b24b52dcf5221af5e8be7

The previous commit ensured that different ws requests would not
subscribe to the same command instance.

The introduced an issue around monitors: Now we have multiple command
instances but they weren't being set on properly on a given monitor.

The commit fixes that by treating commands as subscribers to monitors
and allowing for multiple command to subscribe to the same monitor at
the same time. 

Fixes #

### DIFF
--- a/core/src/monitors/base.ts
+++ b/core/src/monitors/base.ts
@@ -38,7 +38,7 @@ export abstract class Monitor {
   }
 
   unsubscribe(subscriber: Command) {
-    this.subscribers.filter((sub) => sub === subscriber)
+    this.subscribers = this.subscribers.filter((sub) => sub !== subscriber)
   }
 
   unsubscribeAll() {
--- a/core/src/monitors/base.ts
+++ b/core/src/monitors/base.ts
@@ -12,17 +12,16 @@ import type { Garden } from "../garden"
 
 export interface MonitorBaseParams {
   garden: Garden
-  command: Command
 }
 
 export type MonitorKey = PrimitiveMap
 
 export abstract class Monitor {
-  public command: Command
+  public subscribers: Command[]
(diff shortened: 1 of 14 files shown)


thsig (approved):
Looks good to me! There are some failing unit tests there that we should look into before we merge, though.

core/src/monitors/base.ts
```diff
@@ -34,6 +33,18 @@ export abstract class Monitor {
   abstract start(): Promise<{}>
   abstract stop(): Promise<{}>
 
+  subscribe(subscriber: Command) {
+    this.subscribers.push(subscriber)
+  }
+
+  unsubscribe(subscriber: Command) {
+    this.subscribers.filter((sub) => sub === subscriber)
```
edvald, line 41: Missing an assignment here, right?

edvald: And need to flip the operator ofc.

eysi09: Whoops

edvald (approved):

edvald (approved):

Issue #4166, opened 2023-05-04
0.13: [Bug]: Workflows: Providers should only be initialized once needed (lazily)


stefreak:
### Current Behavior

Garden initialises providers in workflows before running actual garden command steps. 

### Expected behavior

Garden should allow running script steps in workflows before provider initialisation, because they might be used for authentication, e.g. to prepare the kubeconfig.

### Reproducible example

`garden.yml`
```yaml
kind: Project
name: repro
environments:
 - name: default
providers:
 - name: exec
   initScript: exit 1

---

kind: Workflow
name: test
steps:
  - script: echo FOO
    name: setup
  - name: build
    command: [build, "-f"]
```

0.13 behaviour
```shell
steffen@Steffens-MBP repro % garden run workflow test
⚠ Garden Bonsai (0.13) is in alpha. Please report any issues here:
https://github.com/garden-io/garden/issues/new?labels=0.13&template=0-13-issue-template.md&title=0.13%3A+%5BBug%5D%3A
Run 🏃‍♂️

⚠ "apiVersion" is missing in the Project config. Assuming "garden.io/v0" for backwards compatibility with 0.12. The "apiVersion"-field is mandatory when using the new action Kind-configs. A detailed migration guide is available at https://docs.garden.io/tutorials/migrating-to-bonsai
ℹ garden                    → Running in Garden environment default.default
ℹ providers                 → Getting status...
ℹ providers                 → Preparing environment...
ℹ exec                      → Configuring...
ℹ exec                      → Running init script
✖ providers                 →
Failed processing resolve provider exec. Here is the output:
────────────────────────────────────────────────────────────────────────────────
exec provider init script exited with code 1
──────────────────────────────────────────────────────────────────────────────── (in 0.8 sec)
✖ Failed resolving one or more providers:
- exec

See .garden/error.log for detailed error message
```

0.12 behaviour
```shell
steffen@Steffens-MBP incident-repro % garden run workflow test
Running workflow test 🏃‍♂️

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
🌍  Running in namespace default in environment default
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  cloud-dashboard           → Authorizing...
[...]

PR #4187, opened 2023-05-08
refactor: rename `run-workflow` to `workflow`

Walther:

This PR does two things:

- Renames `garden run-workflow` command to a simplified `garden workflow`
- Moves the old-style run command invocation detection earlier (before graph init), to support lazy-init usecases (e.g. workflows)

Fixes https://github.com/garden-io/garden/issues/4166

Not flagging this commit specifically as a breaking change, as the breaking change in behavior has already been flagged earlier when moving from `run workflow` to `run-workflow`. This commit also makes sure all the docs are up to date, so we should be good in terms of communication.

Ship small, ship often. This PR continues the old behavior of throwing a config error in case of the use of old-style commands. If we decide to try to support old-style invocations by only warning and trying to continue execution, let's do that in another PR. We could try to add the invocation logic to `detectOldRunCommand` where we already branch for the different errors and suggestions.

### DIFF
--- a/core/src/commands/commands.ts
+++ b/core/src/commands/commands.ts
@@ -27,7 +27,7 @@ import { OptionsCommand } from "./options"
 import { PluginsCommand } from "./plugins"
 import { PublishCommand } from "./publish"
 import { RunCommand } from "./run"
-import { RunWorkflowCommand } from "./run-workflow"
+import { WorkflowCommand } from "./workflow"
 import { SelfUpdateCommand } from "./self-update"
 import { ServeCommand } from "./serve"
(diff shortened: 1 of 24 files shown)


stefreak (approved):
This is a good change and an important step in the right direction 👍  Thank you for fixing this so quickly @Walther 

Just want to note that additionally, in my opinion, it is critical to keep backwards compatibility for commands that likely are used within workflows or within custom commands, because otherwise mixed 0.12 and 0.13 use is not possible in team settings. This can follow in a later PR.

stefreak:
See also #4187

PR #4191, opened 2023-05-08
0.13 cherry-pick: self managed backends for pulumi stacks

vvagaytsev:
Cherry-picks #4107, #4164, and #4148 to `0.13`.

Fixes #


vvagaytsev:
Converting to draft. There is still some regression to be fixed in the `main` branch. This PR will be updated later.

vvagaytsev:
Waiting for #4164

srihas-g (approved):
I've tested and this works as it does with the same changes in main branch. 
There are few things that need to be addressed, they should be added later to both branches.
Thanks Vova!

PR #4193, opened 2023-05-08
0.13: remote sources fix + example update

Orzelius:
The git diff looks very odd, but basically I set the action base bath to the `remoteSourcePath` which fixes the treeVersion issue. Also updated the example.


shumailxyz (approved):
Looks good, thanks. Could you also take a look at `using-remote-sources` docs page and update that alongside in case it needs any changes/updates?

vvagaytsev (approved):

PR #4196, opened 2023-05-08
`garden up` command + clean up syncs on exit

edvald:
Co-authored by @edvald and @thsig.

See the commit messages for more details.
* Added a `--cmd` option to the `dev` command that executes the requested command (with the provided args & opts) at the start of the interactive session.
    * Newlines in the `--cmd` string result in multiple commands being run in the `dev` command.
* Make the `garden deploy` command run inside the `dev` command whenever it would be put into persistent operation.
* Added a `garden up` command that's an alias for `garden deploy --logs`.
* Any syncs started during an interactive session are now stopped when the command is exited gracefully (via `exit` or CTRL-D).
* Warn the user when exiting the `dev` command with CTRL-C.


core/src/commands/dev.tsx
```diff
@@ -156,7 +156,7 @@ Let's get your development environment wired up.
     }))
 
     function quit() {
-      cl?.disable("🌷  Thanks for stopping by, love you! ❤️")
+      cl?.disable("🌷  Thanks for stopping by! ❤️")
```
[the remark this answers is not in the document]
thsig: I know : (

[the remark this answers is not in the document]
eysi09: 💔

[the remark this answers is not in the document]
thsig: It was suggested that some users might find the old text too sugary.

I think it's fine myself.

core/src/commands/dev.tsx
```diff
@@ -155,16 +155,22 @@ Let's get your development environment wired up.
       history: await garden.localConfigStore.get("devCommandHistory"),
     }))
 
+    function quitWithWarning() {
+      log.warn(chalk.yellow(`Syncs started during this session may still be active when this command terminates. You can run ${chalk.white("garden sync stop '*'")} to stop all code syncs.`))
+      log.warn(chalk.yellow(`Hint: To stop code syncing on exit and allow Garden to exit gracefully, use ${chalk.white("Control-D")} or the ${chalk.white(`exit`)} command to exit ${chalk.white("garden dev")}.`))
+      quit()
+    }
```
Orzelius, line 162: 1. ctrl-c still exits gracefully
2. I think this should be a bit shorter and use `garden.emitwarning` to it can be muted

thsig: > 1. ctrl-c still exits gracefully

It doesn't allow us enough time to guarantee that the syncs are turned off (it often is able to, but not always).

> 2. I think this should be a bit shorter and use garden.emitwarning to it can be muted
> 3. I think this should be a single line/message

Yeah, makes sense. I'll make those changes.

thsig: How about now?

Orzelius: > may still be active

Do we actually attempt to close them? I think it'd be cool if we didn't with ctr-c as that would allow to optionally keep them running.

thsig: I'd be open to having that discussion, but if there are no more comments I think we should merge this one (there are a lot of improvements and new things in there) and add any more tweaks in a follow-up PR.

core/src/commands/dev.tsx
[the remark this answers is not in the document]
thsig: I dropped the commit with this change, we can decide later.

vvagaytsev (approved):
Thank you 👍

PR #4201, opened 2023-05-08
0.13 prettier ws logs

eysi09:

Ensure logs from commands requested via the server are consistent with the dev command logs. 

Fixes #


edvald (approved):

vvagaytsev:
There is one test failure, but it looks irrelevant. It was caused by #4195. I fixed that in a follow-up PR #4204.

PR #4204, opened 2023-05-09
test: fix assertions for `run` command

vvagaytsev:
Garden emits non-repeatable warning "Project is configured with `apiVersion: garden.io/v0`, running with backwards compatibility." Thus, the variable `logMessages` may contain 2 lines. This commit introduces helper function `expectLogsContain` to check if there is at least one entry of a given sample among all log lines.

Patch for #4195

It makes the log-checking unit test assertions more stable by taking into account non-repeatable warnings.

### DIFF
--- a/core/src/util/testing.ts
+++ b/core/src/util/testing.ts
@@ -397,6 +397,10 @@ export function expectFuzzyMatch(str: string, sample: string | string[]) {
   samples.forEach((s) => expect(errorMessageNonAnsi.toLowerCase()).to.contain(s.toLowerCase()))
 }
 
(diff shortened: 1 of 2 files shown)


Walther (approved):
I'm terribly sorry, this was my bad. I added the change for apiVersion in the config files at the last minute in the GitHub UI from a review suggestion, but did not wait for all the tests to complete before merging, as the tests had already passed before. I promise to be more careful in the future. Thank you for the quick fix! 🙇‍♂️

vvagaytsev:
@Walther no worries :) Huge thanks to you and @stefreak for improving UX and ensuring backward compatibility.

PR #4211, opened 2023-05-09
fix: unhandled errors when config store update fails

mkhq:

The lock library we use in the config store throws an uncaught exception when a lock is compromised. We catch the error and log out a warning to avoid the entire garden process to crash. This is a workaround for now until we can reproduce and fix the root cause. 

Fixes #

### DIFF
--- a/core/src/config-store/base.ts
+++ b/core/src/config-store/base.ts
@@ -42,11 +42,7 @@ export abstract class ConfigStore<T extends z.ZodObject<any>> {
         return config[section][key]
       }
     } finally {
-      try {
-        await release()
-      } catch (e) {
-        this.logLockReleaseError()
-      }
+      await release()
     }
   }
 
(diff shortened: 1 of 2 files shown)


core/src/config-store/base.ts
```diff
@@ -42,7 +42,11 @@ export abstract class ConfigStore<T extends z.ZodObject<any>> {
         return config[section][key]
       }
     } finally {
-      await release()
+      try {
```
edvald, line 45: I'd suggest doing wrapping in the `private async lock()` method.

core/src/config-store/base.ts
```diff
@@ -156,7 +172,27 @@ export abstract class ConfigStore<T extends z.ZodObject<any>> {
...
+        .then(resolve)
+        .catch(reject)
+    })
+  }
+
+  private logLockReleaseError() {
+    // eslint-disable-next-line
+    console.log("Warning: Unable to release the lock on the global config store; lock was already released")
+    // eslint-disable-next-line
+    console.log("Warning: If you run into this error repeatedly, please report on our GitHub or Discord")
```
edvald, line 195: Slightly odd to print two consecutive lines starting with "Warning: " .) Maybe just fold into one?

thsig (approved):

PR #4219, opened 2023-05-10
chore: remove harden-runner

stefreak:
harden-runner would be useful in our actual build or release workflows, but has no benefits actually in lint and scorecard.

Fixes #


vvagaytsev (approved):
Thank you! 👍

PR #4220, opened 2023-05-10
docs: update remote-sources guide

Orzelius:
also add back a sweeet line break that was missing from https://github.com/garden-io/garden/pull/4193

### DIFF
--- a/docs/advanced/using-remote-sources.md
+++ b/docs/advanced/using-remote-sources.md
@@ -7,11 +7,11 @@ title: Using Remote Sources
 
 You can import **two** types of remote repositories with Garden:
 
-> **Remote _source_**: A repository that contains one or more Garden modules _and_ their corresponding `garden.yml` config files.
+> **Remote _source_**: A repository that contains one or more Garden modules or actions _and_ their corresponding `garden.yml` config files.
 
-> **Remote _module_**: The source code for a single Garden module. In this case, the `garden.yml` config file is stored in the main project repository while the module code itself is in the remote repository.
+> **Remote _actions_**: The source code for a single Garden action. In this case, the `garden.yml` config file is stored in the main project repositorywhile the action code itself is in the remote repository.
 
-The code examples below are from our [remote sources example](https://github.com/garden-io/garden/tree/0.12.51/examples/remote-sources).
+The code examples below are from our [remote sources example](../../examples/remote-sources/README.md).
 
 ## Importing Remote Repositories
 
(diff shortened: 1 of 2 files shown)


docs/advanced/using-remote-sources.md
```diff
@@ -7,11 +7,11 @@ title: Using Remote Sources
 
 You can import **two** types of remote repositories with Garden:
 
-> **Remote _source_**: A repository that contains one or more Garden modules _and_ their corresponding `garden.yml` config files.
+> **Remote _source_**: A repository that contains one or more Garden modules or actions _and_ their corresponding `garden.yml` config files.
 
-> **Remote _module_**: The source code for a single Garden module. In this case, the `garden.yml` config file is stored in the main project repository while the module code itself is in the remote repository.
+> **Remote _actions_**: The source code for a single Garden action. In this case, the `garden.yml` config file is stored in the main project repositorywhile the action code itself is in the remote repository.
```
vvagaytsev, line 12: Missing whitespace in "repositorywhile"

docs/advanced/using-remote-sources.md
```diff
@@ -101,41 +110,51 @@ $ tree .
...
 
-You can also import sources and modules from your local file system by setting the `repositoryUrl` to a local file path:
+You can also import sources from your local file system by setting the `repositoryUrl` or `source.repository.url` to a local file path:
 
 ```yaml
-repositoryUrl: file:///my/local/project/path#main
+# project configuration (remote source)
+sources:
+  - name: web-services
+    repositoryUrl: file:///my/local/project/path#main
```
vvagaytsev, line 123: Do we plan to use the same new syntax here? E.g. 
```
sources:
  - name: web-services
     repository:
         url: file:///my/local/project/path#main
```

No action is needed now, I'm just curious. Maybe we can deprecate the old syntax somewhere in `0.13.x` and remove it in `0.14`.

vvagaytsev (dismissed):
Looks good! Thank you 👍 

Left a couple of non-blocking comments. Please take a look.

docs/advanced/using-remote-sources.md
Orzelius: I think we decided to keep it as is to avoid introducing a project-level breaking change

Orzelius: Also I don't think the config you provided is valid yaml

docs/advanced/using-remote-sources.md
Orzelius, line 12:
```suggestion
> **Remote _actions_**: The source code for a single Garden action. In this case, the `garden.yml` config file is stored in the main project repository while the action code itself is in the remote repository.
```

vvagaytsev (approved):

vvagaytsev:
Sorry, I thought I approved it along with my comment. Ready to go 🚀

PR #4241, opened 2023-05-11
fix(k8s): fix namespace caching logic

### LINKED ISSUE #4223: 0.13: [Bug]: re-deploying from inside dev mode does not re-create deleted namespaces
### Current Behavior

If a user runs `delete namespace` from inside the REPL followed by `deploy`, Garden will not re-create the namespace before re-deploy, leading to errors.

image

image

### Expected behavior

Garden should create any non-existent namespaces when required.

### Reproducible example

https://gist.github.com/worldofgeese/0c5bac7bb16bbc829a6eb62592ddef1f

### Workaround

Exit REPL with `exit`, then run `garden deploy`

### Your environment

* **OS:** Clear* Linux

`garden version` Bonsai latest as of May 9th, 2023

thsig:

Because the `delete namespace` command can be called interactively (which deletes the default app namespace), and because of the `reload` command, we needed a different approach to caching namespace statuses in the Kubernetes plugin.

The new approach scopes the cache by provider UID, and clears the cache for a given provider UID if the `delete namespace` command is called.

We also introduced an async lock to the `ensureNamespace` helper, since the first `deploy` command after a `delete namespace` command can result in more than one concurrent `ensureNamespace` call for the same namespace.

Fixes #4223.

As noted in one of the comments, we may want to provide a cache to all plugin handlers via the plugin context. We should probably come up with one or two more motivating examples before we dive into that, but it's something to keep in mind for the future.

### DIFF
--- a/core/src/cache.ts
+++ b/core/src/cache.ts
@@ -301,3 +301,47 @@ export function pathToCacheContext(path: string): CacheContext {
   const parsed = parse(normalize(path))
   return ["path", ...parsed.dir.split(sep)]
 }
+
+/**
+ * A simple in-memory cache that prunes older entries once the maximum size (as measured by key count) has been
+ * reached.
+ *
+ * Useful for scenarios where new keys accumulate indefinitely, and there's a need to prevent unbounded memory use.
+ */
+export class BoundedCache<CachedValue> {
+  private cache: {
+    [key: string]: CachedValue
+  } = {}
+  private keys: string[] = []
+
+  constructor(private maxCacheCount = 1000) {}
+
+  get(key: string): CachedValue | null {
+    return this.cache[key] || null
+  }
+
+  set(key: string, val: CachedValue): void {
+    if (this.keys.length >= this.maxCacheCount) {
+      const pruneCount = Math.floor(this.maxCacheCount / 2)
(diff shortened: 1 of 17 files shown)


vvagaytsev (changes requested):
There are some [test failures](https://app.circleci.com/pipelines/github/garden-io/garden/17080/workflows/7b63a3a0-2b60-4bc5-91ce-7ab914107410/jobs/292510), please fix :)

core/src/cache.ts
```diff
@@ -301,3 +301,47 @@ export function pathToCacheContext(path: string): CacheContext {
...
+ *
+ * Useful for scenarios where new keys accumulate indefinitely, and there's a need to prevent unbounded memory use.
+ */
+export class BoundedCache<CachedValue> {
+  private cache: {
+    [key: string]: CachedValue
+  } = {}
+  private keys: string[] = []
+
+  constructor(private maxCacheCount = 1000) {}
```
vvagaytsev, line 317: Nitty: this can be called just `maxSize` or `size`.

vvagaytsev (approved):
Great job! 🚀 
Thank you 👍

PR #4247, opened 2023-05-12
fix(cli): throw if no tests/runs are found

### LINKED ISSUE #4224: 0.13: [Bug]: `garden test` does not fail if no test action has been found, when using wildcards
### Current Behavior

- `garden test foobar` fails with an error `✖ Test action "foobar" was not found.`
- `garden test '*foobar'` succeeds with a warning `⚠ No Test actions found (matching argument(s) '*foobar'). Aborting.`

### Expected behavior

Both commands fail, if no test has been found. Otherwise workflows might look like they succeeded, but then no tests ran.

### Reproducible example

```
kind: Project
name: repro
environments:
 - name: default
providers:
 - name: exec
```
(No tests declared, so `garden test` should fail with any positional arguments)

### Workaround

Unknown

### Suggested solution(s)

Convert the warning to an error

### Your environment

* **OS:** macos

`garden version` latest 0.13 edge

Orzelius:
This unifies the behaviour with build and deploy actions and is the same as in 0.12

closes #4224

### DIFF
--- a/core/src/commands/helpers.ts
+++ b/core/src/commands/helpers.ts
@@ -79,9 +79,7 @@ export async function watchRemovedWarning(garden: Garden, log: Log) {
 }
 
 /**
- * Throws if an action by name is not found.
- * Logs a warning if no actions are found matching wildcard arguments.
- *
+ * Throws if nothing is found matching the search parameters.
  */
 export const validateActionSearchResults = ({
   log,
@@ -95,7 +93,7 @@ export const validateActionSearchResults = ({
   actions: { name: string }[]
   errData: any
   actionKind: ActionKind
-}): { shouldAbort: boolean } => {
+}) => {
   names?.forEach((n) => {
     if (!isGlob(n) && !actions.find((a) => a.name === n)) {
       throw new ParameterError(`${actionKind} action "${n}" was not found.`, { ...errData })
@@ -107,10 +105,6 @@ export const validateActionSearchResults = ({
     if (names) {
       argumentsMsg = ` (matching argument(s) ${naturalList(names.map((n) => `'${n}'`))})`
     }
-    log.warn({
-      msg: `No ${actionKind} actions found${argumentsMsg}. Aborting.`,
-    })
-    return { shouldAbort: true }
+    throw new ParameterError(`No ${actionKind} actions were found${argumentsMsg}.`, { errData })
   }
-  return { shouldAbort: false }
 }
--- a/core/src/commands/helpers.ts
+++ b/core/src/commands/helpers.ts
@@ -79,7 +79,9 @@ export async function watchRemovedWarning(garden: Garden, log: Log) {
 }
 
 /**
- * Throws if nothing is found matching the search parameters.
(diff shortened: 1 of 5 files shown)


core/test/unit/src/commands/helpers.ts
```diff
@@ -59,17 +58,32 @@ describe("command helpers", () => {
...
-    it("should log a warning if no actions are found using wildcards", async () => {
-      const log = logger.createLog()
-      validateActionSearchResults({
-        actionKind: "Build",
-        actions: [],
-        log,
-        names: ["foo*"],
-        errData: {},
-      })
+    it("should throw a warning if no actions are found using wildcards", async () => {
```
shumailxyz, line 61:
```suggestion
    it("should throw an error if no actions are found using wildcards", async () => {
```

shumailxyz (approved):
Looks good, thanks. Just one suggestion.

vvagaytsev (approved):
👍

stefreak (dismissed):
Awesome work! Thank you for the quick fix

Orzelius:
the e2e errors were valid. I forgot that on 0.12 we don't throw if no tasks/tests are defined at all. Made the behavior the same in [54d70b4](https://github.com/garden-io/garden/pull/4247/commits/54d70b49fe21c949021bf03e604c1efbadf89299)

core/src/commands/helpers.ts
```diff
@@ -96,6 +96,11 @@ export const validateActionSearchResults = ({
   errData: any
   actionKind: ActionKind
 }): { shouldAbort: boolean } => {
+  if (actions.length === 0 && (!names || names.length === 0)) {
+    log.warn(`No ${actionKind} actions were found. Aborting.`)
```
stefreak, line 100: I would vote for making this a breaking change in 0.13, and failing here.

I do not see a use case where succeeding with a warning would be a good thing.

stefreak: I don't want this behaviour –
it's very confusing if your CI succeeds when you accidentally made a typo or renamed the tests.

What is the reason that we have the current behaviour? What's the argument against changing this, other than "there are end to end tests that rely on this behaviour" which is the same as "this is the currently implemented behaviour"

stefreak: Ah, sorry for that – so it does fail when running `garden test -n testname` or `garden test '*testmame'` as well?

stefreak: Oops, really sorry that I misunderstood this detail. Good work Albert, as always!

stefreak (approved):
This is awesome, thanks for fixing this so quickly. And sorry that I didn't look closely enough at the code and misunderstood what it does during review. Succeeding when no tests are declared at all I think is totally fine.

core/src/commands/helpers.ts
Orzelius: > does fail when running garden test -n testname or garden test '*testmame' as well?

yes, if no tests are found it fails

stefreak (approved):
LGTM

PR #4256, opened 2023-05-13
improvement(dev): handle config changes

### LINKED ISSUE #4213: 0.13: [Bug]: Changing ingress configuration in a Deploy action from dev mode does not update ingress on fresh `deploy`
### Current Behavior

Changing the ingress configuration for a Deploy action inside dev mode doesn’t get picked up: `garden` doesn’t update the service on a new `deploy`. 

E.g. change `/hello` to `/` while `garden dev` is running. Please see Empathy Session doc (restricted to employees of garden.io) for full context starting from 40:00 in.

image

image

### Expected behavior

Dev mode should automatically update the ingress.

### Reproducible example

https://github.com/shumailxyz/nestjs-typeorm-auth-starter/tree/gardenify

### Workaround

To work around, @shumailxyz  killed `garden dev` then ran `garden deploy`.

### Your environment

* **OS:** macOS Ventura

`garden version` Bonsai latest as of May the 8th.

thsig:

If one or more Garden configs is changed during a `dev` command, we lazily reload the project when the next command is run.

This approach avoids multiple reloads when the user is modifying their Garden configs.

The old Watcher class was reduced down to its bare essentials for this purpose (up to this point, it had been unused in 0.13).

Fixes #4213.

### DIFF
--- a/core/src/cli/command-line.ts
+++ b/core/src/cli/command-line.ts
@@ -23,6 +23,7 @@ import { sleep } from "../util/util"
 import { Autocompleter, AutocompleteSuggestion } from "./autocomplete"
 import { parseCliArgs, pickCommand, processCliArgs, renderCommandErrors, renderCommands } from "./helpers"
 import { GlobalOptions, ParameterValues } from "./params"
+import { ServeCommand } from "../commands/serve"
 
 const defaultMessageDuration = 2000
 const commandLinePrefix = chalk.yellow("🌼  > ")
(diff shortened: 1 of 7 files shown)


edvald (approved):
Nice!

edvald (approved):

Orzelius:
wow good job, I've been pushing this to the future thinking it's too difficult but here's @thsig just making it all work!

worldofgeese:
Glorious ☀️

PR #4257, opened 2023-05-13
Some optimizations

edvald:
Tracked down some quick wins for performance optimization. See individual commits.


thsig:
Good stuff! There are some failing unit tests (and the generated docs are out of date), but once that's taken care of we can merge this.

eysi09 (approved):
LGTM. 

Tested on platform repo and `garden sync status` command went down from an average of 17s to 11s. 

Also noting that @thsig reviewed and gave a thumbs up.

PR #4265, opened 2023-05-14
feat: support multiple envs and projects in dev and serve commands

edvald:
This is a substantial refactor, but important because:

1) We want the dev command to start instantly and support --env flags etc. 
2) The upcoming Garden Desktop client needs that same capability. 
3) This more properly accounts for nested commands for Garden Cloud.


core/src/commands/base.ts
```diff
@@ -163,6 +175,156 @@ export abstract class Command<A extends Parameters = {}, O extends Parameters =
...
+      // We make sure events emitted in the context of the command are forwarded to the parent Garden event bus.
+      garden.events.onAny((name, payload) => {
+        parentEvents.emit(name, payload)
+      })
+    }
+
+    const log = garden.log
+
+    if (garden.cloudApi && garden.projectId && this.streamEvents) {
+      // Note: If a config change during a watch-mode command's execution results in the resolved environment
```
eysi09, line 217: Wouldn't this result in action results being stored in Cloud under the wrong environment? 

Feels like we need to either register a new session here or simply not allow users changing the active environment while in dev mode. 

I suppose something similar could happen if you change the project name.

core/src/commands/base.ts
```diff
@@ -163,6 +175,156 @@ export abstract class Command<A extends Parameters = {}, O extends Parameters =
...
+      // and/or namespace name changing, we don't change the session ID, environment ID or namespace ID used when
+      // streaming events.
+      await garden.cloudApi.registerSession({
```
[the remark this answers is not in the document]
edvald: That would be good, yes, where applicable. I can do that.

core/src/commands/base.ts
eysi09: I took this to mean a change in the actual Garden config. E.g. re-naming an environment.

It's clear to me that you can run "child" commands in a different environment and that those will create a session with the correct env set. 

Assuming my interpretation, it's probably not an issue though unless data is persisted between the config change and the next "child" command run. Which I don't think it is.

edvald:
Note: There are a couple of seemingly flaky integ tests there, not yet sure why they're failing in CI, can't repro when I run them myself and the code seems to be behaving as expected.

eysi09:
Running `garden deploy --dev` from this branch consistently fails with:

```
/Users/eysi/code/garden-io/garden/node_modules/yoga-layout-prebuilt/yoga-layout/build/Release/nbind.js:53
        throw ex;
        ^

Error: listen EADDRINUSE: address already in use ::1:9700
    at Server.setupListenHandle [as _listen2] (node:net:1740:16)
    at listenInCluster (node:net:1788:12)
    at GetAddrInfoReqWrap.doListen (node:net:1937:7)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:109:8)
Emitted 'error' event on WebSocketServer instance at:
    at Server.emit (node:events:513:28)
    at Server.emit (node:domain:489:12)
    at emitErrorNT (node:net:1767:8)
    at processTicksAndRejections (node:internal/process/task_queues:82:21) {
  code: 'EADDRINUSE',
  errno: -48,
  syscall: 'listen',
  address: '::1',
  port: 9700
}
```

The port is not occupied before running the command presumably we're starting the server twice.

edvald:
I'm not seeing the same thing, works fine here... Could you try and debug yourself to see why it's happening?

core/src/garden.ts
```diff
@@ -405,10 +409,23 @@ export class Garden {
...
+  }
+
+  /**
+   * Returns a shallow clone of this instance. Useful if you need to override properties for a specific context.
+   */
+  clone(): Garden {
+    return Object.assign(Object.create(Object.getPrototypeOf(this)), this)
+  }
+
+  needsReload(v?: true) {
```
thsig, line 424: Not a huge deal, but this param seems a bit strange—wouldn't it make more sense to use separate methods to set and get the `needsReload` field?

Also, what does the name `v` refer to?

thsig (commented):
Great stuff! Still reading through this, but submitting a couple of comments ahead of time.

thsig (commented):
Great stuff! Still reading through this, but submitting a couple of comments ahead of time.

core/src/server/helpers.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  const pairs: any = { env }
+
+  // Hash any variable overrides
+  if (!isEmpty(params.variableOverrides)) {
+    pairs.var = serializeObject(params.variableOverrides).slice(0, 8)
+  }
+
+  return Object.entries(pairs)
```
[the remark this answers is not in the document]
edvald: It is unnecessary in this case. Besides, I prefer the order to be `env` first.

core/src/garden.ts
edvald: No strong opinion, I was just moving quickly here. I felt it was quite obvious what `v?: true` refers to, personally, but 🤷‍♂️

PR #4272, opened 2023-05-15
0.13: unknown action type warning

### LINKED ISSUE #4266: 0.13: [BUG] Terraform Deploy action type error
@edvald in that case looks like I may have run into a bug converting the provider to an action

```
kind: Deploy
type: terraform
name: my-terraform
spec: 
  autoApply: true
  variables:
    project_id: devrel-348008
```

```shell
✖ deploy.my-terraform  → 
Failed processing resolve Deploy type=terraform name=my-terraform. Here is the output:
Could not find type definition for Deploy type=terraform name=my-terraform.
This is a bug. Please report it!
──────────────────────────────────────────────────────────────────────────────── (in 0 sec)
ℹ deploy.my-mongodb    → Already deployed
✖ 1 deploy action(s) failed!
```

_Originally posted by @worldofgeese in https://github.com/garden-io/garden/issues/3475#issuecomment-1544184859_

cc @Orzelius

Orzelius:
previously the error was something like this

```
Could not find type definition for Deploy type=terraform name=my-terraform.
This is a bug. Please report it!
```

closes #4266


edvald:
Those tests are all legitimately failing. I think you need to revisit the check there and perhaps throw in a functional test.

worldofgeese:
“Are you missing a provider configuration” is maybe still not clear enough. What about: “You've defined an action without its provider. Ensure you've enabled the associated provider in your Garden project then try again.”

Orzelius:
> “Are you missing a provider configuration” is maybe still not clear enough. What about: “You've defined an action without its provider. Ensure you've enabled the associated provider in your Garden project then try again.”

Actually some providers come built in so that error might be even more confusing. I like the current one a lot tbh.

> It would be useful to tell the user which provider exactly is missing.

Most of the time it will be the `config.type` provider, but in some cases it might be a typo on the user side. Either way we can't really be sure. Also added an error when provider is valid or exists but action kind does not in [0ccd879](https://github.com/garden-io/garden/pull/4272/commits/0ccd87995c6e4ff5b1ff29004f7505d07b0fc160)

vvagaytsev (approved):
LGTM 👍

PR #4276, opened 2023-05-16
fix: fix search predicate in `Run` action detection

vvagaytsev:

Fixes #

### DIFF
--- a/core/src/commands/run.ts
+++ b/core/src/commands/run.ts
@@ -161,7 +161,7 @@ export class RunCommand extends Command<Args, Opts> {
 
     const { shouldAbort } = validateActionSearchResults({
       log,
-      actionKind: "Test",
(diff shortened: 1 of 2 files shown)


core/src/commands/run.ts
```diff
@@ -161,7 +161,7 @@ export class RunCommand extends Command<Args, Opts> {
 
     const { shouldAbort } = validateActionSearchResults({
       log,
-      actionKind: "Test",
+      actionKind: "Run",
```
Orzelius, line 164: image
my bad

Orzelius (approved):

core/src/commands/run.ts
vvagaytsev: No worries :)
We all do copy-pasting errors regularly :D

PR #4282, opened 2023-05-16
docs: 0.13 update for `container` plugin guide

vvagaytsev:

Fixes #

### DIFF
--- a/docs/other-plugins/container.md
+++ b/docs/other-plugins/container.md
@@ -180,7 +180,7 @@ action, so that Garden will make sure the database is deployed before running th
 Since Garden `0.13` `Test` actions can also be dependencies for any other kinds of actions, e.g `Build`, `Deploy`
 and `Run` actions.
 
-One thing to note, is that `Run` actions should in most cases be _idempotent_, meaning that running the same action
+One thing to note, is that `Run` actions should in most cases be _idempotent_, meaning that running the same `Run` action
 multiple times should be safe.
 
 See the [reference](../reference/action-types/Run/container.md#tasks) for all the configurable parameters
--- a/docs/other-plugins/container.md
+++ b/docs/other-plugins/container.md
@@ -160,7 +160,7 @@ across your different action configurations.
 See the [reference](../reference/action-types/Test/container.md) for all the configurable parameters
 for `container Test` actions.
 
-## Running arbitrary scripts and jobs
+## Running arbitrary workloads
 
 `Run` actons are defined very similarly to `Test` actions:
 
--- a/docs/other-plugins/container.md
+++ b/docs/other-plugins/container.md
@@ -175,7 +175,7 @@ spec:
 
 In this example, we define a `db-migrate` `Run` action that executes `rake db:migrate` (which is commonly used for
 database migrations, but you can run anything you like of course). The action has a dependency on `my-database` `Deploy`
-action, so that Garden will make sure the database is up and running before running the migration job.
+action, so that Garden will make sure the database is deployed before running the migration job.
 
 Since Garden `0.13` `Test` actions can also be dependencies for any other kinds of actions, e.g `Build`, `Deploy`
 and `Run` actions.
--- a/docs/other-plugins/container.md
+++ b/docs/other-plugins/container.md
@@ -219,12 +219,12 @@ and [`Test`](../reference/action-types/Test/container.md#outputs) action kinds.
 
 ## Mounting volumes and Kubernetes ConfigMaps
 
-_Runtime_ `container` actions (i.e. `Deploy`, `Run`, and `Test`) can all mount volumes and Kubernetes Configmaps.
+Volumes and ConfigMaps can be mounted in all `Deploy`, `Run`, and `Test` actions of the `container` type.
 
 For mounting volumes, check out our guide on
 the [`persistentvolumeclaim` action type](../k8s-plugins/action-types/persistentvolumeclaim.md), supported by
 the `kubernetes` provider.
 
-And for `ConfigMap`s, check out this guide on the [`configmap` action type](../k8s-plugins/action-types/configmap.md),
+And for ConfigMaps, check out this guide on the [`configmap` action type](../k8s-plugins/action-types/configmap.md),
 also supported by the `kubernetes` provider.
 
--- a/docs/other-plugins/container.md
+++ b/docs/other-plugins/container.md
@@ -177,8 +177,7 @@ spec:
 ```
 
 In this example, we define a `db-migrate` `Run` action that executes `rake db:migrate` (which is commonly used for
-database migrations, but you can run anything you like of course). The action has a dependency on `my-database` `Deploy`
-action, so that Garden will make sure the database is deployed before running the migration job.
+database migrations, but you can run anything you like of course). The action has a dependency on the `my-database` deployment, so that Garden will make sure the database is deployed before running the migration job.
 
 Since Garden `0.13` `Test` actions can also be dependencies for any other kinds of actions, e.g `Build`, `Deploy`
 and `Run` actions.
--- a/docs/other-plugins/container.md
+++ b/docs/other-plugins/container.md
@@ -75,11 +75,6 @@ spec:
   image: redis:5.0.5-alpine   # <- replace with any docker image ID
 ```
 
-{% hint style="warning" %}
-Note that if there is a _Dockerfile_ in the same directory as the action configuration, and you still don't want to
-build it, you have to tell Garden not to pick it up by setting `include: []` in your action configuration.
-{% endhint %}
-
 ## Publishing images
 
 You can publish images that have been built in your cluster using the `garden publish` command.
--- a/docs/other-plugins/container.md
+++ b/docs/other-plugins/container.md
@@ -2,17 +2,25 @@
 title: Container
 order: 1
 ---
+
 # Container
 
-Garden includes a `container` plugin, which provides a high-level abstraction around container-based services, that's easy to understand and use.
+Garden includes a `container` plugin, which provides a high-level abstraction around container-based services, that's
+easy to understand and use.
 
 The plugin is built-in and doesn't require any configuration.
 
-The corresponding `container` module type can be used to just _build_ container images, or it can specify deployable services through the optional `services` key, as well as `tasks` and `tests`. So you might in one scenario use a `container` module to both build and deploy services, and in another you might only build the image using a `container` module, and then refer to that image in a `helm` or `kubernetes` module.
+The corresponding `container` module type can be used to just _build_ container images, or it can specify deployable
+services through the optional `services` key, as well as `tasks` and `tests`. So you might in one scenario use
+a `container` module to both build and deploy services, and in another you might only build the image using
+a `container` module, and then refer to that image in a `helm` or `kubernetes` module.
 
-Below we'll walk through some usage examples. For a full reference of the `container` module type, please take a look at the [reference](../reference/module-types/container.md).
+Below we'll walk through some usage examples. For a full reference of the `container` module type, please take a look at
+the [reference](../reference/module-types/container.md).
 
-_Note: Even though we've spent the most time on supporting Kubernetes, we've tried to design this module type in a way that makes it generically applicable to other container orchestrators as well, such as Docker Swarm, Docker Compose, AWS ECS etc. This will come in handy as we add more providers, that can then use the same module type._
+_Note: Even though we've spent the most time on supporting Kubernetes, we've tried to design this module type in a way
+that makes it generically applicable to other container orchestrators as well, such as Docker Swarm, Docker Compose, AWS
+ECS etc. This will come in handy as we add more providers, that can then use the same module type._
 
 ## Building images
 
@@ -25,11 +33,17 @@ type: container
 name: my-container
 ```
 
-If you have a `Dockerfile` next to this file, this is enough to tell Garden to build it. You can also specify `dockerfile: <path-to-Dockerfile>` if you need to override the Dockerfile name. You might also want to explicitly [include or exclude](../using-garden/configuration-overview.md#includingexcluding-files-and-directories) files in the build context.
+If you have a `Dockerfile` next to this file, this is enough to tell Garden to build it. You can also
+specify `dockerfile: <path-to-Dockerfile>` if you need to override the Dockerfile name. You might also want to
+explicitly [include or exclude](../using-garden/configuration-overview.md#includingexcluding-files-and-directories)
+files in the build context.
 
 ### Build arguments
 
-You can specify [build arguments](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg) using the [`buildArgs`](../reference/module-types/container.md#buildArgs) field. This can be quite handy, especially when e.g. referencing other modules such as build dependencies:
+You can
+specify [build arguments](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg)
+using the [`buildArgs`](../reference/module-types/container.md#buildArgs) field. This can be quite handy, especially
+when e.g. referencing other modules such as build dependencies:
 
 ```yaml
 # garden.yml
@@ -37,16 +51,18 @@ kind: Module
 type: container
 name: my-container
 build:
-  dependencies: [base-image]
+  dependencies: [ base-image ]
 buildArgs:
   baseImageVersion: ${modules.base-image.version}
 ```
 
-Garden will also automatically set `GARDEN_ACTION_VERSION` as a build argument, so that you can reference the version of the action being built.
+Garden will also automatically set `GARDEN_ACTION_VERSION` as a build argument, so that you can reference the version of
+the action being built.
 
 ## Using remote images
 
-If you're not building the container image yourself and just need to deploy an external image, you can skip the Dockerfile and specify the `image` field:
+If you're not building the container image yourself and just need to deploy an external image, you can skip the
+Dockerfile and specify the `image` field:
 
 ```yaml
 # garden.yml
@@ -59,14 +75,17 @@ services:
 ```
 
 {% hint style="warning" %}
-Note that if there is a _Dockerfile_ in the same directory as the module configuration, and you still don't want to build it, you have to tell Garden not to pick it up by setting `include: []` in your module configuration.
+Note that if there is a _Dockerfile_ in the same directory as the module configuration, and you still don't want to
+build it, you have to tell Garden not to pick it up by setting `include: []` in your module configuration.
 {% endhint %}
 
 ## Publishing images
 
 You can publish images that have been built in your cluster using the `garden publish` command.
 
-Unless you're publishing to your configured deployment registry (when using the `kubernetes` provider), you need to specify the `image` field on the `container` module in question to indicate where the image should be published. For example:
+Unless you're publishing to your configured deployment registry (when using the `kubernetes` provider), you need to
+specify the `image` field on the `container` module in question to indicate where the image should be published. For
+example:
 
 ```yaml
 kind: Module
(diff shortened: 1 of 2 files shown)


docs/other-plugins/container.md
```diff
@@ -2,180 +2,233 @@
...
 
-_Note: Even though we've spent the most time on supporting Kubernetes, we've tried to design this module type in a way that makes it generically applicable to other container orchestrators as well, such as Docker Swarm, Docker Compose, AWS ECS etc. This will come in handy as we add more providers, that can then use the same module type._
+_Note: Even though we've spent the most time on supporting Kubernetes, we've tried to design this action type in a way
+that makes it generically applicable to other container orchestrators as well, such as Docker Swarm, Docker Compose, AWS
+ECS etc. This will come in handy as we add more providers, that can then use the same action type._
 
 ## Building images
 
-A bare minimum `container` module just specifies common required fields:
+A bare minimum `container Build` action just specifies common required fields:
```
shumailxyz, line 28:
```suggestion
Following is a bare minimum `container Build` action configuration using the required fields:
```

docs/other-plugins/container.md
```diff
@@ -2,180 +2,233 @@
...
 # garden.yml
-kind: Module
+kind: Build
 type: container
 name: my-container
 ```
 
-If you have a `Dockerfile` next to this file, this is enough to tell Garden to build it. You can also specify `dockerfile: <path-to-Dockerfile>` if you need to override the Dockerfile name. You might also want to explicitly [include or exclude](../using-garden/configuration-overview.md#includingexcluding-files-and-directories) files in the build context.
+If you have a `Dockerfile` next to this file, this is enough to tell Garden to build it. You can also
+specify `spec.dockerfile: <path-to-Dockerfile>` if you need to override the Dockerfile name.
```
shumailxyz, line 38:
```suggestion
If you have a `Dockerfile` next to this file, this is enough to tell Garden to build it. However, you can override the `Dockerfile` name or path by specifying `spec.dockerfile: <path-to-Dockerfile>`.
```

docs/other-plugins/container.md
```diff
@@ -2,180 +2,233 @@
...
+You might also want to
+explicitly [include or exclude](../using-garden/configuration-overview.md#includingexcluding-files-and-directories)
+files in the build context.
 
 ### Build arguments
 
-You can specify [build arguments](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg) using the [`buildArgs`](../reference/module-types/container.md#buildArgs) field. This can be quite handy, especially when e.g. referencing other modules such as build dependencies:
+You can specify
+[build arguments](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg)
```
shumailxyz, line 46:
```suggestion
[build arguments](https://docs.docker.com/engine/reference/commandline/build/#build-arg)
```

The link has been slightly changed to point to correct section.

docs/other-plugins/container.md
```diff
@@ -2,180 +2,233 @@
...
-  baseImageVersion: ${modules.base-image.version}
+dependencies: [ build.base-image ]
+spec:
+  buildArgs:
+    baseImageVersion: ${actions.build.base-image.version}
 ```
 
-Garden will also automatically set `GARDEN_ACTION_VERSION` as a build argument, so that you can reference the version of the action being built.
+Garden will also automatically set `GARDEN_ACTION_VERSION` as a build argument, so that you can reference the version of
+the action being built.
```
shumailxyz, line 62:
```suggestion
Additionally, Garden automatically sets `GARDEN_ACTION_VERSION` as a build argument, which you can use to reference the version of action being built.
```

docs/other-plugins/container.md
```diff
@@ -2,180 +2,233 @@
...
 
 ## Using remote images
 
-If you're not building the container image yourself and just need to deploy an external image, you can skip the Dockerfile and specify the `image` field:
+If you're not building the container image yourself and just need to deploy an external image, you do not need to
+define a `Build` action. You only need to define a `Deploy` action:
```
shumailxyz, line 67:
```suggestion
If you're not building the container image yourself and just need to deploy an external image, you can use the `Deploy` action:
```

I think this part `you do not need to define a Build action` can be skipped here.

docs/other-plugins/container.md
```diff
@@ -2,180 +2,233 @@
...
-services:
-  ...
+spec:
+  image: redis:5.0.5-alpine   # <- replace with any docker image ID
 ```
 
 {% hint style="warning" %}
-Note that if there is a _Dockerfile_ in the same directory as the module configuration, and you still don't want to build it, you have to tell Garden not to pick it up by setting `include: []` in your module configuration.
+Note that if there is a _Dockerfile_ in the same directory as the action configuration, and you still don't want to
+build it, you have to tell Garden not to pick it up by setting `include: []` in your action configuration.
```
shumailxyz, line 80: by setting `include` or `exclude`? 🤔

Orzelius, line 80: I don't think this section is needed anymore as there's no build action to begin with.

docs/other-plugins/container.md
```diff
@@ -2,180 +2,233 @@
...
+kind: Build
+name: my-build
+type: container
+spec:
+  publishId: my-repo/my-image:v1.2.3   # <- if you omit the tag here, the Garden action version will be used by default
 ```
 
-By default, we use the tag specified in the `container` module `image` field, if any. If none is set there, we default to the Garden module version.
+By default, we use the tag specified in the `container` action's `spec.publishId` field, if any. If none is set there,
+we default to the Garden `Build` action version.
```
Orzelius, line 100:
```suggestion
By default, we use the tag specified in the `container` action's `spec.publishId` field. If none is set,
we default to the Garden `Build` action version.
```

docs/other-plugins/container.md
```diff
@@ -2,180 +2,233 @@
...
 
-The Kubernetes plugins (local or remote) can deploy container modules. You'll find the relevant information in [this guide](../k8s-plugins/action-types/container.md) and the full spec in our [reference docs](../reference/module-types/container.md#services).
+The Kubernetes plugins (local or remote) can deploy `container Deploy` actions. You'll find the relevant information
+in [this guide](../k8s-plugins/action-types/container.md) and the full spec in
+our [reference docs](../reference/action-types/Deploy/container.md).
 
 ## Running tests
 
-You can define both tests and tasks as part of any container module. The two are configured in very similar ways, using the `tests` and `tasks` keys, respectively. Here, for example, is a configuration for two different test suites:
+For `container` action type you can define both `Test` and `Run` actions. The two are configured in a very similar way.
```
shumailxyz, line 130:
```suggestion
For `container` type, you can define both `Test` and `Run` actions. The two are configured in a very similar way.
```
IMO `container` action type part is bit confusing in reading when i read the full line.

docs/other-plugins/container.md
```diff
@@ -2,180 +2,233 @@
...
 
-## Deploying services
+## Deploying applications
 
-The Kubernetes plugins (local or remote) can deploy container modules. You'll find the relevant information in [this guide](../k8s-plugins/action-types/container.md) and the full spec in our [reference docs](../reference/module-types/container.md#services).
+The Kubernetes plugins (local or remote) can deploy `container Deploy` actions. You'll find the relevant information
+in [this guide](../k8s-plugins/action-types/container.md) and the full spec in
+our [reference docs](../reference/action-types/Deploy/container.md).
 
 ## Running tests
```
Orzelius, line 128: Maybe add a note that these create a new pod and are not executed in the deploy. That means that the dependency on the deploy is also not needed. Kubernetes-exec tests can be used to execute in the existing deployment.

docs/other-plugins/container.md
```diff
@@ -2,180 +2,233 @@
...
-The names and commands to run are of course completely up to you, but we suggest naming the test suites consistently across your different modules.
+The names and commands to run are of course completely up to you, but we suggest naming the test suites consistently
+across your different action configurations.
 
-See the [reference](../reference/module-types/container.md#tests) for all the configurable parameters for container tests.
+See the [reference](../reference/action-types/Test/container.md) for all the configurable parameters
+for `container Test` actions.
 
-## Running tasks
+## Running arbitrary scripts and jobs
```
shumailxyz, line 170:
```suggestion
## Running arbitrary workloads
```

docs/other-plugins/container.md
```diff
@@ -2,180 +2,233 @@
...
 
-Tasks are defined very similarly to tests:
+`Run` actons are defined very similarly to `Test` actions:
```
shumailxyz, line 172:
```suggestion
To run arbitrary workloads, any scripts or jobs, you can use the `Run` action which is defined similarly to the `Test` action:
```

docs/other-plugins/container.md
```diff
@@ -2,180 +2,233 @@
...
-...
+dependencies: [ deploy.my-database ]
+spec:
+  command: [ rake, db:migrate ]
 ```
 
-In this example, we define a `db-migrate` task that runs `rake db:migrate` (which is commonly used for database migrations, but you can run anything you like of course). The task has a dependency on `my-database`, so that Garden will make sure the database is up and running before running the migration task.
+In this example, we define a `db-migrate` `Run` action that executes `rake db:migrate` (which is commonly used for
+database migrations, but you can run anything you like of course). The action has a dependency on `my-database` `Deploy`
+action, so that Garden will make sure the database is up and running before running the migration job.
```
shumailxyz, line 185:
```suggestion
action, so that Garden will make sure the database is deployed before running the migration job.
```

docs/other-plugins/container.md
```diff
@@ -2,180 +2,233 @@
...
 
-Unlike tests, tasks can also be dependencies for services and other tasks. For example, you might define another task or a service with `db-migrate` as a dependency, so that it only runs after the migrations have been executed.
+Since Garden `0.13` `Test` actions can also be dependencies for any other kinds of actions, e.g `Build`, `Deploy`
+and `Run` actions.
 
-One thing to note, is that tasks should in most cases be _idempotent_, meaning that running the same task multiple times should be safe.
+One thing to note, is that `Run` actions should in most cases be _idempotent_, meaning that running the same action
```
shumailxyz, line 190:
```suggestion
One thing to note, is that `Run` actions should in most cases be _idempotent_, meaning that running the same `Run` action
```

docs/other-plugins/container.md
vvagaytsev: Good catch! Removed it in https://github.com/garden-io/garden/pull/4282/commits/64a85fe9b55051ff10838635a16b9a76e72f994d

docs/other-plugins/container.md
```diff
@@ -2,180 +2,228 @@
...
+spec:
+  values:
+    image:
+      name: ${actions.build.my-image.outputs.deployment-image-name}
+      tag: ${actions.build.my-image.version}
 ```
 
-Here, we declare `my-image` as a dependency for the `my-service` Helm chart. In order for the Helm chart to be able to reference the built container image, we must provide the correct image name and version.
+Here we do not need to declare an explicit _build_ dependency on `my-image` like `dependencies: [ build.my-app ]`.
+Instead, we do it implicitly via the references to the `Build` action outputs in `spec.value.image`.
```
shumailxyz, line 210:
```suggestion
Instead, we do it implicitly via the references to the `Build` action outputs in `spec.values.image`.
```

Orzelius:
wow @shumailxyz that's some good quality review there!

docs/other-plugins/container.md
vvagaytsev: This section was removed in https://github.com/garden-io/garden/commit/64a85fe9b55051ff10838635a16b9a76e72f994d

See the [discussion above](https://github.com/garden-io/garden/pull/4282#discussion_r1196440424).

shumailxyz:
Thanks @vvagaytsev. Good stuff with update! Just some nits here and there to improve the overall copy.

docs/other-plugins/container.md
vvagaytsev: Dependency on `deploy.frontend` is necessary here. The integration tests do some calls to `frontend` endpoints, so the tests will fail if the `frontend` app is not deployed.

Verified that in [`demo-project`](https://github.com/garden-io/garden/tree/0.13/examples/demo-project) example. Tried a cold run
```
gdev run frontend-integ
```
with and without `deploy.frontend` dependency. Got an error in the latter case.

vvagaytsev:
@shumailxyz @Orzelius, thanks for the great review and comments! I've addressed all suggestions and replied to all comments. Please do one more review round :)

Orzelius (dismissed):
Maybe even a third pair of eyes would be of use here to get a fresh perspective on this? This doc is very important to garden in general so maybe we can summon someone?

shumailxyz (approved):

docs/other-plugins/container.md
```diff
@@ -2,180 +2,230 @@
...
 
-_Note: Even though we've spent the most time on supporting Kubernetes, we've tried to design this module type in a way that makes it generically applicable to other container orchestrators as well, such as Docker Swarm, Docker Compose, AWS ECS etc. This will come in handy as we add more providers, that can then use the same module type._
+_Note: Even though we've spent the most time on supporting Kubernetes, we've tried to design this action type in a way
+that makes it generically applicable to other container orchestrators as well, such as Docker Swarm, Docker Compose, AWS
+ECS etc. This will come in handy as we add more providers, that can then use the same action type._
 
 ## Building images
 
-A bare minimum `container` module just specifies common required fields:
+Following is a bare minimum `container Build` action configuration using the required fields:
```
Walther, line 28: nit: this sentence could perhaps be revised - the `container Build` feels a bit odd, considering it is in fixed-width (as if a command or piece of config or code) but we do not use that verbatim in the actual configuration

perhaps we could say e.g. "Following is a bare minimum `Build` action using the `container` type:" or something similar?

shumailxyz: Agree with suggestion. also differentiates between action kind and `type`. Less confusing 👍

docs/other-plugins/container.md
```diff
@@ -2,180 +2,230 @@
...
-  baseImageVersion: ${modules.base-image.version}
+dependencies: [ build.base-image ]
+spec:
+  buildArgs:
+    baseImageVersion: ${actions.build.base-image.version}
 ```
 
-Garden will also automatically set `GARDEN_ACTION_VERSION` as a build argument, so that you can reference the version of the action being built.
+Additionally, Garden automatically sets `GARDEN_ACTION_VERSION` as a build argument, which you can use to reference the
+version of action being built.
```
Walther, line 62: Nit: what is the end-user benefit / feature / capability given here? Maybe add some phrasing (or link to another doc) explaining what this gives you, or how you can use this

docs/other-plugins/container.md
shumailxyz: Quite an improvement already. Thanks :)

docs/other-plugins/container.md
```diff
@@ -2,180 +2,230 @@
...
+- `${build.name}` — the name of the build being tagged
+- `${build.version}` — the full Garden version of the build being tagged, e.g. `v-abcdef1234`
+- `${build.hash}` — the Garden version hash of the build being tagged, e.g. `abcdef1234` (i.e. without the `v-`
+  prefix)
 
-## Deploying services
+## Deploying applications
 
-The Kubernetes plugins (local or remote) can deploy container modules. You'll find the relevant information in [this guide](../k8s-plugins/action-types/container.md) and the full spec in our [reference docs](../reference/module-types/container.md#services).
+The Kubernetes plugins (local or remote) can deploy `container Deploy` actions. You'll find the relevant information
```
Walther: Nit: this sentence is also perhaps a bit too internal-technical. Maybe this could be rephrased to something more user-oriented, e.g.
> After your software has been built, you might also want to deploy it somewhere. For this, check out our guide on deploying to Kubernetes using `Deploy` actions

docs/other-plugins/container.md
```diff
@@ -2,180 +2,230 @@
...
+- `${build.hash}` — the Garden version hash of the build being tagged, e.g. `abcdef1234` (i.e. without the `v-`
+  prefix)
 
-## Deploying services
+## Deploying applications
 
-The Kubernetes plugins (local or remote) can deploy container modules. You'll find the relevant information in [this guide](../k8s-plugins/action-types/container.md) and the full spec in our [reference docs](../reference/module-types/container.md#services).
+The Kubernetes plugins (local or remote) can execute `Deploy` actions of `container` type. You'll find the relevant
+information in [this guide](../k8s-plugins/action-types/container.md) and the full spec in
+our [reference docs](../reference/action-types/Deploy/container.md).
```
Walther, line 121: (repeating the comment as the text changed a bit before i submitted the comment, making it immediately outdated)

Nit: this sentence is also perhaps a bit too internal-technical. Maybe this could be rephrased to something more user-oriented, e.g.
> After your software has been built, you might also want to deploy it somewhere. For this, check out our guide on deploying to Kubernetes using Deploy actions

docs/other-plugins/container.md
```diff
@@ -2,180 +2,230 @@
...
 
 ## Running tests
 
-You can define both tests and tasks as part of any container module. The two are configured in very similar ways, using the `tests` and `tasks` keys, respectively. Here, for example, is a configuration for two different test suites:
+For `container` type, you can define both `Test` and `Run` actions. The two are configured in a very similar way.
```
Walther, line 125: Here we mention both `Test` and `Run` actions and the fact they are configured in a similar way. However, we only give an example of a `Test` action.

Should provide an example of a `Run` action as well, or at least mention some differences between them to explain the motivation?

docs/other-plugins/container.md
```diff
@@ -2,180 +2,230 @@
...
-...
+dependencies: [ deploy.my-database ]
+spec:
+  command: [ rake, db:migrate ]
 ```
 
-In this example, we define a `db-migrate` task that runs `rake db:migrate` (which is commonly used for database migrations, but you can run anything you like of course). The task has a dependency on `my-database`, so that Garden will make sure the database is up and running before running the migration task.
+In this example, we define a `db-migrate` `Run` action that executes `rake db:migrate` (which is commonly used for
+database migrations, but you can run anything you like of course). The action has a dependency on `my-database` `Deploy`
+action, so that Garden will make sure the database is deployed before running the migration job.
```
Walther, line 181: could we simplify this too, maybe
```suggestion
database migrations, but you can run anything you like of course). The action has a dependency on the `my-database` deployment, so that Garden will make sure the database is deployed before running the migration job.
```

docs/other-plugins/container.md
```diff
@@ -2,180 +2,230 @@
...
 
-Unlike tests, tasks can also be dependencies for services and other tasks. For example, you might define another task or a service with `db-migrate` as a dependency, so that it only runs after the migrations have been executed.
+Since Garden `0.13` `Test` actions can also be dependencies for any other kinds of actions, e.g `Build`, `Deploy`
+and `Run` actions.
```
Walther, line 184: On another thought: do we need this sentence at all here? This is in the `Run` section but we are now explaining something about `Test` actions

docs/other-plugins/container.md
```diff
@@ -2,180 +2,230 @@
...
+One thing to note, is that `Run` actions should in most cases be _idempotent_, meaning that running the same `Run`
+action
+multiple times should be safe.
 
-See the [reference](../reference/module-types/container.md#tasks) for all the configurable parameters for container tasks.
+See the [reference](../reference/action-types/Run/container.md#tasks) for all the configurable parameters
+for `Run` actions of `container` type.
 
-## Referencing from other modules
+## Referencing from other actions
```
Walther, line 193: re: previous comment, we could potentially move the sentence here somewhere. Any action kinds can depend on any other actions, etc

docs/other-plugins/container.md
```diff
@@ -2,180 +2,230 @@
...
+the _outputs_ reference
+of [`Build`](../reference/action-types/Build/container.md#outputs),
+[`Deploy`](../reference/action-types/Deploy/container.md#outputs),
+[`Run`](../reference/action-types/Run/container.md#outputs),
+and [`Test`](../reference/action-types/Test/container.md#outputs) action kinds.
 
 ## Mounting volumes and Kubernetes ConfigMaps
 
-`container` services, tasks and tests can all mount volumes and Kubernetes Configmaps.
+_Runtime_ `container` actions (i.e. `Deploy`, `Run`, and `Test`) can all mount volumes and Kubernetes Configmaps.
```
Walther, line 223: Nit: is this sentence a bit too under-the-hood technical as well? Could we say something like

> Volumes and ConfigMaps can be mounted in all `Deploy`, `Run`, and `Test` actions of the `container` type.

docs/other-plugins/container.md
```diff
@@ -2,180 +2,230 @@
...
+  command: [ rake, db:migrate ]
 ```
 
-In this example, we define a `db-migrate` task that runs `rake db:migrate` (which is commonly used for database migrations, but you can run anything you like of course). The task has a dependency on `my-database`, so that Garden will make sure the database is up and running before running the migration task.
+In this example, we define a `db-migrate` `Run` action that executes `rake db:migrate` (which is commonly used for
+database migrations, but you can run anything you like of course). The action has a dependency on `my-database` `Deploy`
+action, so that Garden will make sure the database is deployed before running the migration job.
 
-Unlike tests, tasks can also be dependencies for services and other tasks. For example, you might define another task or a service with `db-migrate` as a dependency, so that it only runs after the migrations have been executed.
+Since Garden `0.13` `Test` actions can also be dependencies for any other kinds of actions, e.g `Build`, `Deploy`
```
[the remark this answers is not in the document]
Walther: actually, see https://github.com/garden-io/garden/pull/4282#discussion_r1196621068

docs/other-plugins/container.md
vvagaytsev: Changed this in https://github.com/garden-io/garden/pull/4282/commits/5159529d7e2685d7720a977f7d3e0d2d4ac86b09

docs/other-plugins/container.md
```diff
@@ -2,180 +2,230 @@
...
 
-See the [reference](../reference/module-types/container.md#tests) for all the configurable parameters for container tests.
+See the [reference](../reference/action-types/Test/container.md) for all the configurable parameters
+for `Test` actions of `container` type.
 
-## Running tasks
+## Running arbitrary workloads
 
-Tasks are defined very similarly to tests:
+To run arbitrary workloads, any scripts or jobs, you can use the `Run` action which is defined similarly to the `Test`
```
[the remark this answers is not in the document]
vvagaytsev: Changed this in https://github.com/garden-io/garden/pull/4282/commits/2e4b6eb4f403570857d7a281c551d8c768a6f00b

docs/other-plugins/container.md
vvagaytsev: Removed this in https://github.com/garden-io/garden/pull/4282/commits/48869558458a375f64b7fcaf5aabfa0f9b773412

docs/other-plugins/container.md
vvagaytsev: Changed this, see https://github.com/garden-io/garden/pull/4282#discussion_r1196670393

vvagaytsev:
@Walther thanks for the detailed comments! I've addressed those. Please, take a look again :) Feel free to resolve the discussions that no longer need any changes :)

vvagaytsev:
Ah, there are a few more open discussions, sorry :) Looking into those now.

docs/other-plugins/container.md
vvagaytsev: I don't know tbh :)

Orzelius: I think it's somewhat obvious: you know what's the version of your app in the build process and can use it internally to set versions, render docs or clear caches. I think adding a link to the docker doc is a quick win here https://docs.docker.com/engine/reference/commandline/build/#build-arg

Orzelius:
This is going to be the best doc ever after all this triage
image

docs/other-plugins/container.md
vvagaytsev: Please check https://github.com/garden-io/garden/pull/4282/commits/07da07b65ab6a0f9081d4264509ed44a675163b9

thsig (approved):

Walther (dismissed):
Big thanks to everyone for making this guide so good! ✨

vvagaytsev:
@Walther @Orzelius I've made some minor changes in https://github.com/garden-io/garden/pull/4282/commits/d2c655516db199bd2a9cdf53251d928a20a53538

Please take a look. Let's get this merged today :)

Orzelius (approved):

PR #4286, opened 2023-05-16
fix: remove duplicate accessModes key from pvc provider module and action

Orzelius:
The schema validation for the jsonSchema generated spec actually doesn't work but I don't want to spend more time on this and k8s validates it on apply anyway


vvagaytsev (approved):
LGTM! Thanks 👍

PR #4293, opened 2023-05-17
0.13 update terraform

Walther:

Adding a new `terraform` version `1.4.6` which is currently the latest stable release.

This is required because some of our example code already depends on `required_version = ">= 1.3.0"`, but we only ship `1.2.9` with garden.

### DIFF
--- a/core/src/plugin/tools.ts
+++ b/core/src/plugin/tools.ts
@@ -29,7 +29,12 @@ const toolBuildSchema = () =>
       .required()
       .example("linux")
       .description("The platform this build is for."),
-    architecture: joi.string().allow("amd64").required().example("amd64").description("The architecture of the build."),
+    architecture: joi
+      .string()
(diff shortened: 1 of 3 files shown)


twelvemo (approved):
Nice!

core/src/plugin/tools.ts
```diff
@@ -29,7 +29,12 @@ const toolBuildSchema = () =>
       .required()
       .example("linux")
       .description("The platform this build is for."),
-    architecture: joi.string().allow("amd64").required().example("amd64").description("The architecture of the build."),
+    architecture: joi
+      .string()
+      .allow("amd64", "arm64")
```
Walther, line 34: Note: interestingly we haven't got an error from here even though we specify `arm64` on some of the tools. Wonder if some validation is going missing / not executed?

PR #4294, opened 2023-05-17
wip: fixing volume stuff

Orzelius:

Fixes #


core/src/config/common.ts
```diff
@@ -682,7 +682,7 @@ joi = joi.extend({
     } catch (err) {
       const error = opts.error("validation")
       error.message = err.message
-      return error
+      return { errors: error }
```
vvagaytsev, line 685: Good catch! 👍

vvagaytsev:
Thanks for the fixes! All look good 👍  I'll take a closer look at the PVC guide update later. It might make sense to split this PR into 2: one for fixes, and another for guide updates. The former would be approved right away :)

docs/k8s-plugins/action-types/persistentvolumeclaim.md
```diff
@@ -5,47 +5,56 @@ order: 5
 
 ## PersistentVolumeClaim
 
-`container` services, tasks and tests can all mount volumes using this module type. To mount a volume, you need to define a volume module, and reference it using the `volumes` key on your services, tasks and/or tests.
+Creates a [PersistentVolumeClaim](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims) in your namespace
+that can be referenced and mounted by other resources and [`container` actions](./container.md).
+
+`container` Deploys, tasks and tests can all mount volumes using this action type. To mount a volume, you need to define a volume Deploy
```
Orzelius, line 11:
```suggestion
`container` Deploys, Runs and Tests can all mount volumes using this action type. To mount a volume, you need to define a volume Deploy
```

docs/k8s-plugins/action-types/persistentvolumeclaim.md
```diff
@@ -5,47 +5,56 @@ order: 5
...
+and reference it using the `volumes` key on your Deploys, Runs or Tests.
```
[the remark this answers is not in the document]
Orzelius: Maybe we should but then we aught to also enforce it

PR #4298, opened 2023-05-17
improvement(core): logging tweaks

thsig:

Here, we make various fixes to rendering & consistency (colors, timestamps, skipping symbols in the command footer separator in the dev command etc).

### DIFF
--- a/core/src/commands/base.ts
+++ b/core/src/commands/base.ts
@@ -572,7 +572,7 @@ export function printResult({
   description: string
 }) {
   const prefix = success ? `${capitalize(description)} output:` : `${capitalize(description)} failed with error:`
-  const msg = renderMessageWithDivider(prefix, result, !success)
+  const msg = renderMessageWithDivider({ prefix, msg: result, isError: !success })
   success ? log.info(chalk.white(msg)) : log.error(msg)
 }
 
--- a/core/src/commands/deploy.ts
+++ b/core/src/commands/deploy.ts
@@ -203,10 +203,10 @@ export class DeployCommand extends Command<Args, Opts> {
     const disabled = deployActions.filter((s) => s.isDisabled()).map((s) => s.name)
 
     if (disabled.length > 0) {
-      const bold = disabled.map((d) => chalk.bold(d))
+      const bold = disabled.map((d) => chalk.white(d))
       const msg =
         disabled.length === 1 ? `Deploy action ${bold} is disabled` : `Deploy actions ${naturalList(bold)} are disabled`
-      log.info(chalk.white(msg))
+      log.info(chalk.gray(msg))
     }
 
     const skipped = opts.skip || []
--- a/core/src/logger/log-entry.ts
+++ b/core/src/logger/log-entry.ts
@@ -17,6 +17,7 @@ import chalk from "chalk"
 import { GardenError } from "../exceptions"
 import hasAnsi from "has-ansi"
 import { omitUndefined } from "../util/objects"
+import { renderDuration } from "./util"
 
 export type LogSymbol = keyof typeof logSymbols | "empty"
 export type TaskLogStatus = "active" | "success" | "error"
@@ -153,7 +154,8 @@ export interface LogEntry<C extends BaseContext = LogContext>
 
 interface LogParams
   extends Pick<LogEntry, "metadata" | "msg" | "symbol" | "data" | "dataFormat" | "error" | "skipEmit">,
-    Pick<LogContext, "origin"> {}
+    Pick<LogContext, "origin">,
+    Pick<LogConfig<LogContext>, "showDuration"> {}
 
 interface CreateLogEntryParams extends LogParams {
   level: LogLevel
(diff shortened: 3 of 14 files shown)


core/src/logger/log-entry.ts
```diff
@@ -273,8 +275,15 @@ export abstract class Log<C extends BaseContext = LogContext> implements LogConf
    * That is, the time from when the log instance got created until now.
    */
   private getMsgWithDuration(params: CreateLogEntryParams) {
-    if (this.showDuration && params.msg) {
-      return params.msg + ` (in ${this.getDuration(1)} sec)`
+    // If params.showDuration is set, it takes precedence over this.duration (since it's set at the call site for the
+    // log line in question).
+    const showDuration = params.showDuration !== undefined
```
vvagaytsev, line 280: Can it be just `const showDuration = params.showDuration || this.showDuration`?

core/src/logger/log-entry.ts
```diff
@@ -273,8 +275,15 @@ export abstract class Log<C extends BaseContext = LogContext> implements LogConf
...
+      ? params.showDuration
+      : this.showDuration
+    if (showDuration && params.msg) {
+      const msg = hasAnsi(params.msg) ? params.msg : chalk.green(params.msg)
+      return msg + " " + chalk.white(renderDuration(this.getDuration(1)))
+      // return msg + " " + chalk.white(`(in ${this.getDuration(1)} sec)`)
```
vvagaytsev, line 286: This can be removed

core/src/tasks/build.ts
```diff
@@ -58,7 +58,7 @@ export class BuildTask extends ExecuteActionTask<BuildAction, BuildStatus> {
       log: log || this.log,
     })
 
-    log.verbose(chalk.green(`Done syncing sources (took ${log.getDuration(1)} sec)`))
+    log.verbose(chalk.green(`Done syncing sources (in ${log.getDuration(1)} sec)`))
```
vvagaytsev, line 61: Can we use `renderDuration` helper here?

core/src/commands/base.ts
```diff
@@ -419,7 +419,7 @@ export function printResult({
   description: string
 }) {
   const prefix = success ? `${capitalize(description)} output:` : `${capitalize(description)} failed with error:`
-  const msg = renderMessageWithDivider(prefix, result, !success)
+  const msg = renderMessageWithDivider({ prefix, msg: result, isError: !success })
```
vvagaytsev, line 422: Nit. The whitespace in `renderMessageWithDivider({ prefix,` is displayed and `NBSP` in WebStorm. Maybe it should just a regular whitespace :)

core/src/commands/deploy.ts
```diff
@@ -196,10 +196,10 @@ export class DeployCommand extends Command<Args, Opts> {
     const disabled = deployActions.filter((s) => s.isDisabled()).map((s) => s.name)
 
     if (disabled.length > 0) {
-      const bold = disabled.map((d) => chalk.bold(d))
+      const bold = disabled.map((d) => chalk.white(d))
       const msg =
         disabled.length === 1 ? `Deploy action ${bold} is disabled` : `Deploy actions ${naturalList(bold)} are disabled`
-      log.info(chalk.white(msg))
+      log.info(chalk.gray(msg))
```
eysi09, line 202: It's kind of awkward that `chalk.gray` is needed here because of the ANSI chars in the msg.

I'm wondering if the render should just always apply `chalk.gray` instead of checking for ANSI codes and only applying if there aren't any.

It's only the text that doesn't have ANSI that the gray color is applied to, so I think it should really break anything, except just color things gray that used to be white for now good reason.

eysi09: ...it's an open question whether gray should be the default color to begin with, but as long as it's consistent it's easy to fix.

core/src/logger/log-entry.ts
```diff
@@ -273,8 +275,15 @@ export abstract class Log<C extends BaseContext = LogContext> implements LogConf
    * That is, the time from when the log instance got created until now.
    */
   private getMsgWithDuration(params: CreateLogEntryParams) {
-    if (this.showDuration && params.msg) {
-      return params.msg + ` (in ${this.getDuration(1)} sec)`
+    // If params.showDuration is set, it takes precedence over this.duration (since it's set at the call site for the
```
eysi09, line 278: 👍

core/src/tasks/build.ts
eysi09, line 61: Should we revert to printing "took" instead of "in". I went for the latter because it was shorter but doesn't always work grammatically. Or a third option if you can think of one. 

We'd essentially need to change it here and in the `renderDuration` function.

core/src/logger/log-entry.ts
thsig: No, because we're distinguishing between `undefined` (the param is not used) and ` false`  (the param is being used, and is asking for the value `false`).

core/src/commands/deploy.ts
thsig: Wouldn't this make it hard to override the color in e.g. `log.info` if that's what we want?

eysi09: I might be wrong but I think it would only be applied to words with no color in the first place. 

Something like `chalk.gray(chalk.white("foo")` is white. 

Where as:
```
const myString = "foo" + chalk.white("bar")
chalk.gray(myString) // gray and white
```

...I think.

eysi09 (approved):

eysi09 (dismissed):

eysi09 (dismissed):

edvald:
Looks good, but there are some failing tests. Can you take a look @thsig?

thsig:
Should be good to go now.

eysi09 (approved):

PR #4299, opened 2023-05-17
improvement(dev): better help text & header

thsig:

Added some more descriptive text to the header about what can be done in the dev command.

Also sorted the command listing rendered by the `help` command by "popularity", with the primary / most commonly used commands appearing closer to the top (and putting less frequently used commands in their own section).

This should make the command listing more readable, especially for new users who aren't familiar with the available commands.

Screenshot 2023-05-17 at 22 55 51

### DIFF
--- a/core/src/cli/command-line.ts
+++ b/core/src/cli/command-line.ts
@@ -27,7 +27,15 @@ import { TypedEventEmitter } from "../util/events"
 import { uuidv4 } from "../util/random"
 import { sleep } from "../util/util"
 import { AutocompleteSuggestion } from "./autocomplete"
-import { parseCliArgs, pickCommand, processCliArgs, renderCommandErrors, renderCommands } from "./helpers"
+import {
+  getOtherCommands,
+  getPopularCommands,
+  parseCliArgs,
+  pickCommand,
+  processCliArgs,
+  renderCommandErrors,
+  renderCommands,
+} from "./helpers"
 import type { GlobalOptions, ParameterValues } from "./params"
 
 const defaultMessageDuration = 3000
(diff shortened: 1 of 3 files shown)


thsig:
Note: The inconsistent colours of the action durations have been fixed in https://github.com/garden-io/garden/pull/4298.

core/src/cli/command-line.ts
```diff
@@ -486,15 +486,50 @@ ${renderDivider({ width, char, color })}
   }
 
   showHelp() {
+    // These commands are rendered first in the help text, since they're more commonly used than the others.
+    const popularCommandFullNames = new Set<string>([
```
vvagaytsev, line 490: This seems to be unused, and it duplicates the function defined in another file.

vvagaytsev (approved):
Left a non-blocking comment, please check it out. There is some dead code :) Otherwise looks good, thank you! 👍

thsig:
Thanks @vvagaytsev—I think this is ready to go now.

vvagaytsev (dismissed):
Thank you! 👍

core/src/commands/dev.tsx
```diff
@@ -60,6 +60,9 @@ export class DevCommand extends ServeCommand<DevCommandArgs, DevCommandOpts> {
 ${renderDivider({ color: chalk.green, title: chalk.green.bold("🌳  garden dev 🌳 "), width })}
 
 ${chalk.bold(`Good ${getGreetingTime()}! Welcome to the Garden interactive development console.`)}
+
+Here, you can ${chalk.white("build")}, ${chalk.white("deploy")}, ${chalk.white("test")} and ${chalk.white("run")} anything in your project, start code syncing, stream live logs and more.
+
```
edvald, line 65: I think it'd be helpful to say a bit more, e.g. tell users about the `help` command, up/down keys to scroll through command history, tab to autocomplete etc etc.

edvald: Merging this, but imo we can improve the dev command intro message further.

edvald (approved):

edvald (approved):

PR #4303, opened 2023-05-18
docs: tweaks to migration guide

thsig:

A few tweaks to the migration guide here (good job on that, by the way!)


Orzelius (approved):
I think it reads very well and has a warm and welcoming feeling to it now

PR #4330, opened 2023-05-19
fix(server): errors in abortCommand and loadConfig requests in cloud projects

edvald:
cc @eysi09


eysi09 (dismissed):
Just tested against my Cloud instance and works like a charm.

eysi09 (approved):

PR #4331, opened 2023-05-19
chore: 0.13 update analytics with new segment library

mkhq:

This uses the new [Segment Analytics Next](https://github.com/segmentio/analytics-next) library which has improved handling for pending events on shutdown and better TypeScript support. This also includes a fix for nock and node 18 by overriding node-fetch in the test setup, details in https://github.com/nock/nock/issues/2336.

Did several rounds of manual testing using the quickstart example to ensure we record events Run Command and Command Result events while in dev console. Also verified the sessionId uses the parent session.

Fixes #


eysi09 (dismissed):
This is awesome! Thank you so much for fixing this and updating Segment. 

Massive improvement.

core/src/analytics/analytics.ts
```diff
@@ -272,15 +273,13 @@ export class AnalyticsHandler {
     cloudUser?: UserResult
     ciInfo: CiInfo
   }) {
-    const segmentClient = require("analytics-node")
-    this.segment = new segmentClient(API_KEY, { flushAt: 20, flushInterval: 300 })
+    this.segment = new Analytics({ writeKey: API_KEY, maxEventsInBatch: 1, flushInterval: 3000 })
```
vvagaytsev: Are these changes intended?
1. increase `flushInterval` from 300 to 300
2. `maxEventsInBatch = 1`

vvagaytsev (approved):
Looks good, thank you!

core/src/analytics/analytics.ts
mkhq: I think this comment was on an older version? There are 5 events in batch and the flushInterval was increased to avoid too frequent checking. The important thing is that analytics are flushed on process close which has been improved with the new library.

PR #4333, opened 2023-05-19
improvement(sync-status): show all syncs

thsig:

We now show a sync status for syncs that aren't currently live (e.g. where the target resource isn't deployed).

### DIFF
--- a/core/src/plugins/kubernetes/sync.ts
+++ b/core/src/plugins/kubernetes/sync.ts
@@ -661,17 +661,16 @@ export async function getSyncStatus(params: GetSyncStatusParams): Promise<GetSyn
       })
     } catch (err) {
       log.debug(`Could not find deployed resource - returning not-active status for sync ${JSON.stringify(s)}.`)
-      const remoteDestination = s.containerPath
-      const { source, target } = orientEndpoints({
+      const oriented = orientEndpoints({
         mode: s.mode,
         localPath: s.sourcePath,
         localPathDescription: s.sourcePath,
-        remoteDestination,
-        remoteDestinationDescription: remoteDestination,
+        remoteDestination: s.containerPath,
+        remoteDestinationDescription: s.containerPath,
       })
       syncStatuses.push({
-        source,
-        target,
+        source: oriented.source,
+        target: oriented.target,
         state: "not-deployed",
         mode: s.mode,
       })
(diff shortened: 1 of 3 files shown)


core/src/plugins/kubernetes/sync.ts
```diff
@@ -639,22 +649,51 @@ export async function getSyncStatus(params: GetSyncStatusParams): Promise<GetSyn
...
       spec: s,
     })
 
-    // Validate the target
-    if (!isConfiguredForSyncMode(target)) {
-      log.debug(chalk.yellow(`Resource ${resourceName} is not deployed in sync mode, cannot start sync.`))
+    if (!isConfiguredForSyncMode(target) || !containerName) {
+      syncStatuses.push({
+        source: sourceDescription,
+        targetDescription: stripAnsi(targetDescription),
```
eysi09, line 689: This is a silly leftover from my changes. 

Should we just have a `targetDescription` variables and a `targetDescriptionLogMsg` so that we don't have to strip the ANSI characters.

core/src/plugins/kubernetes/sync.ts
```diff
@@ -639,22 +649,51 @@ export async function getSyncStatus(params: GetSyncStatusParams): Promise<GetSyn
...
+      target,
       spec: s,
     })
 
-    // Validate the target
-    if (!isConfiguredForSyncMode(target)) {
-      log.debug(chalk.yellow(`Resource ${resourceName} is not deployed in sync mode, cannot start sync.`))
+    if (!isConfiguredForSyncMode(target) || !containerName) {
+      syncStatuses.push({
+        source: sourceDescription,
```
eysi09, line 688: Looks like are we also applying ANSI codes to the source. We should remove those. 

Otherwise it doesn't print right in the terminal and consumers of this API will get a funny looking string.

core/src/plugins/kubernetes/sync.ts
eysi09, line 689: This is a silly leftover from my changes.

Should we just have a `targetDescription` variables and a `targetDescriptionLogMsg` so that we don't have to strip the ANSI characters?

eysi09 (changes requested):
Works like a charm except the ANSI business. Great work!

eysi09 (dismissed):

core/src/plugins/kubernetes/sync.ts
```diff
@@ -661,17 +661,17 @@ export async function getSyncStatus(params: GetSyncStatusParams): Promise<GetSyn
       })
     } catch (err) {
       log.debug(`Could not find deployed resource - returning not-active status for sync ${JSON.stringify(s)}.`)
-      const remoteDestination = s.containerPath
-      const { source, target } = orientEndpoints({
+      // const { source: defaultSource, target: defaultTarget } = orientEndpoints({
```
eysi09, line 664: Need to remove this, right?

thsig:
This should be good to go now (needs to be squashed before merging).

edvald (approved):

PR #4334, opened 2023-05-19
0.13 stream fewer events

eysi09:

Fixes #

### DIFF
--- a/core/src/events.ts
+++ b/core/src/events.ts
@@ -351,11 +351,15 @@ export interface Events {
 
 export type EventName = keyof Events
 
+type GraphEventName = Extract<EventName, "taskCancelled" | "taskComplete" | "taskError" | "taskProcessing">
+type ConfigEventName = Extract<EventName, "configChanged" | "configsScanned">
+
 // These are the events we POST over https via the BufferedEventStream
-export const pipedEventNames: EventName[] = [
+const pipedEventNamesSet = new Set<EventName>([
   "_test",
   "_workflowRunRegistered",
   "configsScanned",
+  "configChanged",
   "sessionCompleted",
   "sessionFailed",
   "sessionCancelled",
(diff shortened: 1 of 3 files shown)


thsig (dismissed):

core/src/events.ts
```diff
@@ -382,5 +382,47 @@ export const pipedEventNames: EventName[] = [
   "workflowStepSkipped",
 ]
 
+type PipedWsEventName = Extract<
+  EventName,
+  "taskCancelled" | "taskComplete" | "taskError" | "taskProcessing"
+>
+
+// These are the events we send over a websocket connection via the Garden server
+export const pipedWsEventNames: PipedWsEventName[] = [
```
edvald, line 391: Two notes:
1) This pattern is much nicer: https://github.com/garden-io/garden/blob/7e73951a789403e197db4d4bb31d53a646de3287/core/src/actions/types.ts#L105-L106
2) We do also need the `configChanged` event here. Adding that back will fix the broken test.

edvald: Sorry, meant `configsScanned`. But both would be useful.

eysi09: > Two notes:
> 
> 1. This pattern is much nicer: https://github.com/garden-io/garden/blob/7e73951a789403e197db4d4bb31d53a646de3287/core/src/actions/types.ts#L105-L106
> 2. We do also need the `configChanged` event here. Adding that back will fix the broken test.

Regarding 1. 

Yeah I thought about that but I wanted to make sure that the `PipedWsEventName` was an actual subset of `EventName`, not sure how to do that with `as const`. 

This is to prevent accidentally having a `PipedWsEventName` like `taskCanceled` (notice the missing `l`). 

So basically just making sure we're always narrowing down from some known initial type.

edvald (approved):

PR #4340, opened 2023-05-20
feat(core): added sync restart command

thsig:

feat(core): added sync restart command

This command stops and starts any active syncs for the relevant Deploy actions.

This is useful for recovering from error states for syncing tools.

Also added diagnostic message logging for failure states to the `sync status` command.

For Mutagen-based syncs, we provide clear information to the user why the sync has been halted, and suggest running the `garden sync reset` command after making sure that the relevant directories are in the expected state.

This addresses sporadic reports we've gotten from our users where code syncing stops working due to a failure condition in Mutagen (e.g. when the user deletes or moves the source dir). Here, both the diagnostics and the call-to-action are made clear and simple.

### DIFF
--- a/core/src/commands/sync/sync-status.ts
+++ b/core/src/commands/sync/sync-status.ts
@@ -90,7 +90,7 @@ export class SyncStatusCommand extends Command<Args> {
     log.info(
       chalk.white(deline`
       Getting sync statuses. For more detailed debug information, run this command with
-      the \`--json\` or \`--yaml\` flags.
+      the \`--output json\` or \`--outputyaml\` flags.
     `)
     )
     log.info("")
--- a/core/src/commands/sync/sync-status.ts
+++ b/core/src/commands/sync/sync-status.ts
@@ -78,8 +78,11 @@ export class SyncStatusCommand extends Command<Args> {
     const router = await garden.getActionRouter()
     const graph = await garden.getResolvedConfigGraph({ log, emit: true })
 
+    // We default to getting the sync status for all actions
+    const names = args.names || ["*"]
+
     const deployActions = graph
-      .getDeploys({ includeDisabled: false, names: args.names })
+      .getDeploys({ includeDisabled: false, names })
       .sort((a, b) => (a.name > b.name ? 1 : -1))
     // This is fairly arbitrary
     const concurrency = 5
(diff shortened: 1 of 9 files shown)


thsig:
Here's a screencap:

https://github.com/garden-io/garden/assets/382137/891e9743-d964-4dc1-be0f-bce2d8259be9

As you can see, I put Mutagen into a "halted due to root deletion" state. The `sync status` command picked this up and provided a diagnostic message. I then followed the instructions and ran `garden sync reset`, and then ran `garden util mutagen sync list` to verify that the sync was healthy again.

thsig:
This one's ready now.

docs/reference/commands.md
```diff
@@ -3510,6 +3510,31 @@ Examples:
...
+
+#### Usage
+
+    garden sync restart <names> 
+
+#### Arguments
+
+| Argument | Required | Description |
+| -------- | -------- | ----------- |
+  | `names` | Yes | The name(s) of one or more Deploy(s) (or services if using modules) whose syncs you want to restart. You may specify multiple names, separated by spaces. To restart all possible syncs, specify &#x27;*&#x27; as an argument.
```
edvald, line 3534: Is this still required? We had a discussion about this a while back and if I recall we wanted to allow an empty list of arguments. We can also do that separately with all the sync commands, but I'm flagging it because the example above contradicts this help text here.

thsig: Ah, I see—must have missed out on that one. I wanted to add support for an empty list of names yesterday, but decided against it to keep things consistent.

I'd be happy to make the relevant changes to the sync commands.

core/src/commands/sync/sync-status.ts
```diff
@@ -90,7 +90,7 @@ export class SyncStatusCommand extends Command<Args> {
     log.info(
       chalk.white(deline`
       Getting sync statuses. For more detailed debug information, run this command with
-      the \`--json\` or \`--yaml\` flags.
+      the \`--output json\` or \`--outputyaml\` flags.
```
vvagaytsev, line 93: The white-spacing looks a bit confusing here. Should the second entry be `--output yaml`?

thsig:
Should be good to go now.

vvagaytsev:
It looks like [docs are not up-to-date](https://app.circleci.com/pipelines/github/garden-io/garden/17635/workflows/94f3c878-1973-47ac-827f-0ca0d4961feb/jobs/306186).

thsig:
Ok, just updated them.

vvagaytsev (approved):
🚀

PR #4345, opened 2023-05-21
perf(solver): fix high CPU usage due to unnecessary event data emission

edvald:
cc @thsig @eysi09


eysi09 (approved):
This looks good to me. 

We need to make sure to test properly against Cloud as well but I honestly don't see anything that would cause issues.

PR #4359, opened 2023-05-22
improvement(commands): add --skip-detail and --only-deploys flags

eysi09:

This commit adds a couple of "internal" flags to the sync status and get status commands that make it easier for Cloud and Desktop to consume their output.

Note that the "garden get status" command will be replaced with a top level "garden status" command and that these changes aren't really meant to be user facing. Rather just to unblock a few things on the Cloud side.

Fixes #

### DIFF
--- a/core/src/commands/get/get-status.ts
+++ b/core/src/commands/get/get-status.ts
@@ -7,7 +7,7 @@
  */
 
 import Bluebird from "bluebird"
-import { fromPairs } from "lodash"
+import { fromPairs, omit } from "lodash"
 import { deepFilter } from "../../util/objects"
 import { Command, CommandResult, CommandParams } from "../base"
 import { ResolvedConfigGraph } from "../../graph/config-graph"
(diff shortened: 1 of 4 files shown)


core/src/commands/get/get-status.ts
```diff
@@ -36,11 +37,30 @@ export interface StatusCommandResult {
   }
 }
 
+const getStatusOpts = {
+  "skip-detail": new BooleanParameter({
+    help: deline`
+      Skip plugin specific details. Only applicable when using the --output=json|yaml option.
```
thsig, line 43:
```suggestion
      Skip plugin-specific details. Only applicable when using the --output=json | yaml option.
```

core/src/commands/get/get-status.ts
```diff
@@ -36,11 +37,30 @@ export interface StatusCommandResult {
...
+      Useful for trimming down the output.
+    `,
+  }),
+  "only-deploys": new BooleanParameter({
+    hidden: true,
+    help: deline`
+      [INTERNAL]: Only return statuses of deploy actions. Currently only used by Cloud and Desktop apps.
+      Will be replaced by a new, top level \`garden status\` command.
```
thsig, line 51:
```suggestion
      Will be replaced by a new, top-level \`garden status\` command.
```

core/src/commands/get/get-status.ts
```diff
@@ -91,6 +123,20 @@ export class GetStatusCommand extends Command {
       }
     }
 
+    // We only skip detail for Deploy actions. Note that this is mostly used internally and that this command
+    // will be replaced by a top-level "garden status" command. For that one we'll probably wan to pass the
```
thsig, line 127:
```suggestion
    // will be replaced by a top-level "garden status" command. For that one we'll probably want to pass the
```

thsig (changes requested):
Looks good! Had a few suggested line edits there.

thsig (approved):

PR #4360, opened 2023-05-22
docs: update stack graph diagrams to use actions

thsig:

Here, we update the diagram on the main README, as well as the top one in the "How Garden Works" guide.

Both now show snippets of action configs (instead of module configs, as they did in the old version).


vvagaytsev (approved):
👍

PR #4375, opened 2023-05-22
fix(exec): do not show outdated warning for in case of sync mode

### LINKED ISSUE #4366: Better warning message when running the exec command when in sync mode
If you run something like `garden exec api /bin/sh` when in sync mode you're almost guaranteed to see the following warning:

Image

Obviously the command did work so the warning isn't necessary. 

Can we skip it if the command doesn't error? Or have the warning be more informative and such that it doesn't "freak out" the user.

shumailxyz:
Avoid showing `outdated` warning for `exec` in case of sync mode.

Fixes #4366


core/src/commands/exec.ts
```diff
@@ -81,13 +81,27 @@ export class ExecCommand extends Command<Args, Opts> {
...
+    // check if deployment is in sync mode
+    const syncStatus = (
+      await router.deploy.getSyncStatus({
+        log: actionLog,
+        action: executed,
+        monitor: false,
+        graph,
+      })
+    ).result
+    const deploySync = syncStatus?.syncs?.[0]
```
vvagaytsev, line 98: Should this be moved to `case "outdated":` block? Looks like it's used only in that scope, so we can avoid unnecessary status check calls for the rest deploy states.

vvagaytsev (approved):
LGTM, thank you! 👍 

This can be squashed and merged :)

PR #4380, opened 2023-05-22
chore: add a per-command flag to disable analytics, starting with autocomplete

mkhq:

This introduces a flag per command to disable analytics. Initially we use this for autocomplete.

Fixes #

There is a temporary fix for an issue with nock and node 18 to make sure we use a compatible fetch. This can be removed when we move to node 20.


vvagaytsev (dismissed):
Thank you! 👍

vvagaytsev (approved):

PR #4381, opened 2023-05-22
test(terraform): run tests for multiple tf versions

stefreak:
- test(terraform): run tests for multiple tf versions
- improvement: remove terraform lock files from git tree

Fixes https://github.com/garden-io/garden/issues/4507

See comment thread below - this turned out to have a bit of complexity in it and spawned some new follow-up issues.


Walther:
All of the terraform-plugin tests pass now locally 🎉 Will look into the CI issue soon.

We should probably do another round of refactoring to clean these tests up (possibly in another PR). They have accumulated incremental changes over time and there's quite a lot of clashing between global and local variables. They were a major contributor to why fixing the tests (and understanding them in their broken state) took such a long time.

Additionally, work on this PR uncovered this issue in our `log` implementation https://github.com/garden-io/garden/issues/4457

Walther:
Blocked by https://github.com/garden-io/garden/issues/4467

Walther:
All of the `terraform` tests now pass locally
```
 70 passing (4m)

✨  Done in 216.82s.
yarn test  113.29s user 32.90s system 67% cpu 3:37.04 total
```

The `test-plugins` step fails in CI though due to https://github.com/garden-io/garden/issues/4467
Additional discussion on that issue on Discord [here](https://discord.com/channels/817392104711651328/1112745715177893969).

I suggest we review and merge this PR as-is, to unblock other work (including other plugin test fixes which depend on this PR) and fix the corruption bug separately. This will sadly cause some redness in our CI, but should be temporary.

plugins/jib/test/util.ts
```diff
@@ -10,7 +10,7 @@ import { expectError } from "@garden-io/sdk/testing"
 import { expect } from "chai"
 import { detectProjectType, getBuildFlags } from "../util"
 
-describe("util", () => {
+describe.skip("util", () => {
```
stefreak, line 13: @Walther There seem to be a couple of left over skips

Walther: Intentional, tracked separately at https://github.com/garden-io/garden/issues/4469

Walther:
I have now added `.skip()` calls to all of the terraform tests due to the executable corruption bug.
Merging this PR fixes https://github.com/garden-io/garden/issues/4507 and unblocks some other PRs.
We have a separate tracking issue for re-enabling these tests here https://github.com/garden-io/garden/issues/4473, to be completed after fixing https://github.com/garden-io/garden/issues/4467

plugins/terraform/test/terraform.ts
```diff
@@ -20,691 +20,711 @@ import { DeployTask } from "@garden-io/core/build/src/tasks/deploy"
...
-      const provider = (await _garden.resolveProvider(garden.log, "terraform")) as TerraformProvider
-      const ctx = await _garden.getPluginContext({ provider, templateContext: undefined, events: undefined })
+      return garden.processTasks({ tasks: [deployTask], throwOnError: true })
+    }
 
-      await setWorkspace({ ctx, provider, root: tfRoot, log: _garden.log, workspace: "default" })
+    async function runTestTask(autoApply: boolean, allowDestroy = false) {
+      await garden.scanAndAddConfigs()
+      garden["actionConfigs"]["Deploy"]["tf"].spec.allowDestroy = allowDestroy
+      garden["actionConfigs"]["Deploy"]["tf"].spec.autoApply = autoApply
```
Orzelius, line 351: These modifications can now be made directly on the action. I think that looks a bit better.
[Example](https://github.com/garden-io/garden/pull/4451/commits/f6d9588a51e7922ddb9cb9cfbacbf034b51caf49#diff-691166936da71fb6c653f86cef0908eb96ba5a7a9f65e8a7ac78e05f4b68ef46R53-R54)

plugins/terraform/test/test-project-action/tf/.terraform.lock.hcl
```diff
@@ -0,0 +1,21 @@
+# This file is maintained automatically by "terraform init".
```
Orzelius, line 1: Think you forgot to remove this file

plugins/terraform/test/test-project-module/garden.yml
```diff
@@ -1,29 +0,0 @@
-kind: Project
```
Orzelius, line 1: Is there a reason for the conversion to actions? Did the module based config not work?

Orzelius (approved):
Just some small comments and questions

plugins/terraform/test/terraform.ts
Walther: Agreed 100%, there's a lot of cleanup and refactoring to do for these tests.

> We should probably do another round of refactoring to clean these tests up (possibly in another PR). They have accumulated incremental changes over time and there's quite a lot of clashing between global and local variables. That was a major contributor to why fixing the tests (and understanding them in their broken state) took such a long time.

I would personally suggest leaving the refactoring (including the direct action modifications you mention above, as well as cleaning up the local vs global scope `garden` instances etc) to a separate PR in order to unblock other, higher-priority work.

What do you think - should this PR be blocked on that cleanup, or should we have it as a separate task?

Orzelius: nah let it roll, I think the number one priority is to get them running in the ci as otherwise they're close to useless

plugins/terraform/test/test-project-action/tf/.terraform.lock.hcl
Walther: Not having the lock file caused some tests to occasionally fail, especially between test runs on the different versions. However that might have been a transient issue while there were still bugs in the test cases and various steps might have been interrupted - I just deleted the local lockfile and re-ran all the tests successfully. Removing and pushing a commit 👍

plugins/terraform/test/test-project-module/garden.yml
Walther: This conversion was mostly to ensure that we have test cases for the new-style configuration, which most new users will be using.

If you want, we could also make sure to have tests for all three - provider configuration, action configuration, module configuration

Walther: I did a `git checkout main test-project-module` to restore the module directory and ran `gdev validate` on it, discovering https://github.com/garden-io/garden/issues/4508 which is most likely the exact same issue as we had on Pulumi modules here https://github.com/garden-io/garden/issues/4468. The fix is likely to be very similar.

I'll create a separate PR for fixing that validation

stefreak: Maybe in that case it makes sense to run tests for both module-style config, and action-style config

stefreak: Sure, works for me

vvagaytsev (approved):
Thanks! 🎉

PR #4389, opened 2023-05-22
fix(core): reintroduce kind-specific cmd outputs

### LINKED ISSUE #4324: 0.13: Update command output format to be closer to 0.12
### Current Behavior

A lot of commands in garden acorn used to have a wide range of useful outputs e.g. the `build` command exposed the versions of the build modules. I am focussing on the `build` command as an example here, but a lot of command outputs are also missing for other commands like `deploy` or `test`.
Outputs of the `build` command in acorn:
```
# A map of all modules that were built (or builds scheduled/attempted for) and information about the builds.
builds:
  <module name>:
    # The full log from the build.
    buildLog:

    # Set to true if the build was fetched from a remote registry.
    fetched:

    # Set to true if the build was performed, false if it was already built, or fetched from a registry
    fresh:

    # Additional information, specific to the provider.
    details:

    # Set to true if the build was not attempted, e.g. if a dependency build failed.
    aborted:

    # The duration of the build in msec, if applicable.
    durationMsec:

    # Whether the build was succeessful.
    success:

    # An error message, if the build failed.
    error:

    # The version of the module, service, task or test.
    version:

# A map of all services that were deployed (or deployment scheduled/attempted for) and the service status.
deployments:
  <service name>:
    # When the service was first deployed by the provider.
    createdAt:

    # Additional detail, specific to the provider.
    detail:

    # Whether the service was deployed with dev mode enabled.
    devMode:

    # Whether the service was deployed with local mode enabled.
    localMode:

    namespaceStatuses:
      - pluginName:

        # Valid RFC1035/RFC1123 (DNS) label (may contain lowercase letters, numbers and dashes, must start with a
        # letter, and cannot end with a dash) and must not be longer than 63 characters.
        namespaceName:

        state:

    # The ID used for the service by the provider (if not the same as the service name).
    externalId:

    # The provider version of the deployed service (if different from the Garden module version.
    externalVersion:

    # A list of ports that can be forwarded to from the Garden agent by the provider.
    forwardablePorts:
      - # A descriptive name for the port. Should correspond to user-configured ports where applicable.
        name:

        # The preferred local port to use for forwarding.
        preferredLocalPort:

        # The protocol of the port.
        protocol:

        # The target name/hostname to forward to (defaults to the service name).
        targetName:

        # The target port on the service.
        targetPort:

        # The protocol to use for URLs pointing at the port. This can be any valid URI protocol.
        urlProtocol:

    # List of currently deployed ingress endpoints for the service.
    ingresses:
      - # The port number that the service is exposed on internally.
        # This defaults to the first specified port for the service.
        port:

        # The ingress path that should be matched to route to this service.
        path:

        # The protocol to use for the ingress.
        protocol:

        # The hostname where the service can be accessed.
        hostname:

    # Latest status message of the service (if any).
    lastMessage:

    # Latest error status message of the service (if any).
    lastError:

    # A map of values output from the service.
    outputs:
      <name>:

    # How many replicas of the service are currently running.
    runningReplicas:

    # The current deployment status of the service.
    state:

    # When the service was last updated by the provider.
    updatedAt:

    # Set to true if the build was not attempted, e.g. if a dependency build failed.
    aborted:

    # The duration of the build in msec, if applicable.
    durationMsec:

    # Whether the build was succeessful.
    success:

    # An error message, if the build failed.
    error:

    # The version of the module, service, task or test.
    version:

# A map of all tests that were run (or scheduled/attempted) and the test results.
tests:
  <test name>:
    # The name of the module that was run.
    moduleName:

    # The command that was run in the module.
    command:

    # The exit code of the run (if applicable).
    exitCode:

    # When the module run was started.
    startedAt:

    # When the module run was completed.
    completedAt:

    # The output log from the run.
    log:

    namespaceStatus:
      pluginName:

      # Valid RFC1035/RFC1123 (DNS) label (may contain lowercase letters, numbers and dashes, must start with a
      # letter, and cannot end with a dash) and must not be longer than 63 characters.
      namespaceName:

      state:

    # A map of primitive values, output from the test.
    outputs:
      # Number, string or boolean
      <name>:

    # The name of the test that was run.
    testName:

    # Set to true if the build was not attempted, e.g. if a dependency build failed.
    aborted:

    # The duration of the build in msec, if applicable.
    durationMsec:

    # Whether the build was succeessful.
    success:

    # An error message, if the build failed.
    error:

    # The version of the module, service, task or test.
    version:

# A map of all raw graph results. Avoid using this programmatically if you can, and use more structured keys instead.
graphResults:
```

Outputs of the `build` command in bonsai:
```
# Set to true if the command execution was aborted.
aborted:

# Set to false if the command execution was unsuccessful.
success:

# A map of all raw graph results. Avoid using this programmatically if you can, and use more structured keys instead.
graphResults:
```

These outputs are especially missed in garden workflows, where you often want one step to use outputs from a previous step. Consider this workflow:
```
kind: Workflow
name: my-workflow
steps:
  - command: [build]
  - script: |
      echo "Built version ${steps.step-1.outputs.builds.my-module.version}"
```
In garden acorn this workflow used to work. One use-case is getting the version of a previously build container to publish the container image with that tag to another registry in a script step.
### Expected behavior

The command outputs should also be there in garden bonsai.

0.13

thsig:

In 0.12, we included keys for build, deploys and tasks in the command result for commnds that used the task graph (e.g. the build & deploy commands).

In 0.13, we initially removed these in favour of the `graphResults` object. Here, we reintroduce them (with most fields still there). This should prevent breakages e.g. in workflows where referencing command outputs is relatively common.

In a future PR, we should get rid of the `graphResults` key entirely from command results. Doing so will require reworking some of our unit tests (since some suites rely on the dependency structure of the results to verify which tasks were actually run).

Fixes #4324

### DIFF
--- a/docs/reference/commands.md
+++ b/docs/reference/commands.md
@@ -74,8 +74,325 @@ aborted:
 # Set to false if the command execution was unsuccessful.
 success:
 
-# A map of all raw graph results. Avoid using this programmatically if you can, and use more structured keys instead.
-graphResults:
+# A map of all executed Builds (or Builds scheduled/attempted) and information about the them.
+build:
+  <Build name>:
+    # The full log from the build.
+    buildLog:
+
+    # Set to true if the build was fetched from a remote registry.
+    fetched:
+
+    # Set to true if the build was performed, false if it was already built, or fetched from a registry
+    fresh:
+
+    # Additional information, specific to the provider.
+    details:
+
+    # Set to true if the action was not attempted, e.g. if a dependency failed.
+    aborted:
+
+    # The duration of the action's execution in msec, if applicable.
+    durationMsec:
+
+    # Whether the action was succeessfully executed.
+    success:
+
+    # An error message, if the action's execution failed.
+    error:
+
+    # The version of the task's inputs, before any resolution or execution happens. For action tasks, this will
+    # generally be the unresolved version.
+    inputVersion:
+
+    # Alias for `inputVersion`. The version of the task's inputs, before any resolution or execution happens. For
+    # action tasks, this will generally be the unresolved version.
+    version:
+
+    # A map of values output from the action's execution.
+    outputs:
+      <name>:
+
+  <Build name>:
+    # The full log from the build.
+    buildLog:
+
+    # Set to true if the build was fetched from a remote registry.
+    fetched:
+
+    # Set to true if the build was performed, false if it was already built, or fetched from a registry
+    fresh:
+
+    # Additional information, specific to the provider.
+    details:
+
+    # Set to true if the action was not attempted, e.g. if a dependency failed.
+    aborted:
+
+    # The duration of the action's execution in msec, if applicable.
+    durationMsec:
+
+    # Whether the action was succeessfully executed.
+    success:
+
+    # An error message, if the action's execution failed.
+    error:
+
+    # The version of the task's inputs, before any resolution or execution happens. For action tasks, this will
+    # generally be the unresolved version.
+    inputVersion:
+
+    # Alias for `inputVersion`. The version of the task's inputs, before any resolution or execution happens. For
+    # action tasks, this will generally be the unresolved version.
+    version:
+
+    # A map of values output from the action's execution.
+    outputs:
+      <name>:
+
+# A map of all executed Deploys (or Deployments scheduled/attempted) and the Deploy status.
+deploy:
+  <Deploy name>:
+    # When the service was first deployed by the provider.
+    createdAt:
+
+    # When the service was first deployed by the provider.
+    updatedAt:
+
+    # The mode the action is deployed in.
+    mode:
+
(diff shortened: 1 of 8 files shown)


docs/reference/commands.md
```diff
@@ -68,14 +68,393 @@ Examples:
 #### Outputs
 
 ```yaml
-# Set to true if the command execution was aborted.
+# Set to true if the action was not attempted, e.g. if a dependency failed or parameters were incorrect.
 aborted:
 
-# Set to false if the command execution was unsuccessful.
+# Whether the action was succeessful.
```
edvald, line 74: Not a big deal but this doesn't refer to a specific action here. Same with `aborted`.

edvald (approved):

edvald (dismissed):

edvald (approved):

PR #4430, opened 2023-05-24
fix: properly wait for k8s deploys to complete

Orzelius:
We were getting the desired replica count from the status API
response [1] which has the current replica count, not the desired amount.

Desired amount is specified in the spec [2].

1. https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/stateful-set-v1/#StatefulSetStatus
2. https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/stateful-set-v1/#StatefulSetSpec

### DIFF
--- a/core/src/plugins/kubernetes/status/workload.ts
+++ b/core/src/plugins/kubernetes/status/workload.ts
@@ -196,7 +196,7 @@ async function getRolloutStatus(workload: Workload) {
     const status = <V1StatefulSetStatus>workload.status
     const workloadSpec = <Required<V1StatefulSetSpec>>workload.spec
 
-    const replicas = status.replicas || 0
+    const replicas = (workload.spec as V1StatefulSetSpec).replicas || 0
     const updated = status.updatedReplicas || 0
     const ready = status.readyReplicas || 0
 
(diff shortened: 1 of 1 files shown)


core/src/plugins/kubernetes/status/workload.ts
```diff
@@ -221,7 +221,7 @@ async function getRolloutStatus(workload: Workload) {
 
     const desired = deploymentSpec.replicas === undefined ? 1 : deploymentSpec.replicas
     const updated = status.updatedReplicas || 0
-    const replicas = status.replicas || 0
+    const replicas = (workload.spec as any).replicas || 0
```
srihas-g, line 224: curious, why is it `any` instead of `V1DeploymentSpec` ?

Orzelius: Good catch, idk what I was thinking. Fixed in [f231320](https://github.com/garden-io/garden/pull/4430/commits/f2313201133312102dc0def335c99260313a1923)

srihas-g (approved):
lgtm, I hope relevant tests are happy!

vvagaytsev (approved):

PR #4431, opened 2023-05-24
Set default value for `apiVersion`

vvagaytsev:
Addresses [this note](https://github.com/garden-io/garden/pull/4297/files#r1202536260) on the `apiVersion` default value.
It also brings some minor refactoring and cleanup.

Fixes #

### DIFF
--- a/core/src/config/common.ts
+++ b/core/src/config/common.ts
@@ -879,6 +879,7 @@ export const unusedApiVersionSchema = () =>
   joi
     .string()
     .valid(GardenApiVersion.v0)
+    .default(GardenApiVersion.v0)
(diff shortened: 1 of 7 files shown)


stefreak (approved):
LGTM 👍

stefreak:
@vvagaytsev Another comment: We only noticed during integration testing with cloud. Should we add a unit test to verify that the default is there?

vvagaytsev:
> @vvagaytsev Another comment: We only noticed this in the cloud e2e tests. Should we add a unit test to verify that the default is there so we notice earlier next time, e.g. when we change to another schema library?

Sure! I'm going to do more changes in the next PR, so I'll add some unit tests too.

Issue #4018, opened 2023-03-29
Garden ignoring nuxt dynamic route files


uni-ravin-singh:
## Bug
We are using nuxt 3 and deploying in k8s using garden. The garden build does not copy the nuxt dynamic routing files. These files are named like below
*[id].vue*
So, whatever file has square bracket in the name garden ignoring those files. 
### Current Behavior

Garden ignoring file with square bracket in the file name.

### Expected behavior

Garden should not ignore files with square bracket in file name

### Workaround

Not able to find any workaround. 

### Your environment

OS: Windows
k8s: EKS

garden version
0.12.51

PR #4434, opened 2023-05-24
fix: escape rsync special characters in filenames on windows

Walther:

Filenames such as `[id].vue` can break our file synchronization (and hence builds and deploys) on Windows, where we use `rsync`

Fixes https://github.com/garden-io/garden/issues/4018
Verified using a [reproducible example](https://github.com/uni-venkateshwar-viswanathan/garden-test/) on a Windows VM on Azure

https://linux.die.net/man/1/rsync section `Include/Exclude Pattern Rules`
> rsync chooses between doing a simple string match and wildcard matching by checking if the pattern contains one of these three wildcard characters: oq*cq, oq?cq, and oq[cq . 

Do we want to backport this to 0.12 (and cut a new release)?

### DIFF
--- a/core/src/util/sync.ts
+++ b/core/src/util/sync.ts
@@ -52,8 +52,13 @@ export async function syncWithOptions({
       // Workaround for this issue: https://stackoverflow.com/questions/1813907
       opts.push("--include-from=-", "--exclude=*", "--delete-excluded")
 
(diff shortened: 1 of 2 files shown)


Orzelius (approved):
You're freakin' legends for doing this windows legwork!

> Do we want to backport this to 0.12 (and cut a new release)?

It shouldn't be too hard as it's a small change but I'll leave it up to you

PR #4439, opened 2023-05-25
fix: render (log) errors

Orzelius:
See commits for details.

A bunch of test assertions might fail, will need to fix separately if so.

This also completes this https://github.com/orgs/garden-io/projects/5/views/1?pane=issue&itemId=28914020 <- was erroring out just as expected, it's just that the error was not shown.

### DIFF
--- a/core/src/logger/renderers.ts
+++ b/core/src/logger/renderers.ts
@@ -50,7 +50,7 @@ export function renderError(entry: LogEntry): string {
   if (error) {
     const noAnsiErr = stripAnsi(error.message || "")
     const noAnsiMsg = stripAnsi(msg || "")
-    // render error only if message doesn't laready contain it
+    // render error only if message doesn't already contain it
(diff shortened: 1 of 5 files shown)


core/src/logger/renderers.ts
```diff
@@ -45,15 +43,17 @@ export function combineRenders(entry: LogEntry, logger: Logger, renderers: Rende
...
+  const { error, msg } = entry
 
   let out = ""
 
   if (error) {
-    if (msg) {
-      out += "\n\n"
+    const noAnsiErr = stripAnsi(error.message || "")
+    const noAnsiMsg = stripAnsi(msg || "")
+    // render error only if message doesn't laready contain it
```
vvagaytsev, line 53:
```suggestion
    // render error only if message doesn't laready contain it
    // render error only if message doesn't already contain it
```

shumailxyz (approved):
Looks good, thank you 👍

vvagaytsev (approved):
👍

Walther (approved):

PR #4449, opened 2023-05-25
feat: add get actions, builds, deploy cmd and update get runs, tests cmd

### LINKED ISSUE #4218: 0.13: Rework the `get` commands
All the get action commands should output a similarly formatted table, something like this:
| Kind   |      Type      |  Name |
|----------|:-------------:|------:|

I think additionally a flag could be passed to show more details like dependencies and dependents, status and if the action is disabled among other possibly useful information.

The following should be implemented:
* `get actions`
* `get deploys`
* `get builds`
* possibly `test-results`

The following needs fixing:
* `get status`

shumailxyz:
Adds following commands:
- `get actions`
- `get deploys`
- `get builds`

Updates following commands:
- `get tests`
- `get runs`

Fixes #4218 

There are breaking changes because the `get tests` and `get runs` commands now behave similar to the get actions command and have different output now.

However, if we want to avoid these breaking changes, i can revert changes to get tests & runs command.


Orzelius:
@shumailxyz if you request review on a wip branch I don't really know if it should we reviewed yet and I get emails about new commits.

shumailxyz:
@Orzelius My bad. I thought i had removed all reviewers.

shumailxyz:
@garden-io/core I'd like to request a pre-review of the WIP PR. Still need to add tests and `get runs`, `get builds` and `get tests` command however would be nice if you could already take a quick look. Thanks.

core/src/commands/get/get-actions.ts
```diff
@@ -0,0 +1,292 @@
...
+import { printHeader } from "../../logger/util"
+import { ActionRouter } from "../../router/router"
+import { sanitizeValue } from "../../util/logging"
+import { deepFilter } from "../../util/objects"
+import { dedent, deline, renderTable } from "../../util/string"
+import { Command, CommandParams, CommandResult } from "../base"
+import { createSchema, joi, joiArray } from "../../config/common"
+import { actionConfigSchema } from "../../actions/helpers"
+
+interface GetActionsCommandResultActionObject {
```
Orzelius, line 30:
```suggestion
interface GetActionsCommandResultItem {
```
I think it reads better like this

core/src/commands/get/get-actions.ts
```diff
@@ -0,0 +1,292 @@
...
+
+export const getActionsCmdOutputSchema = createSchema({
+  name: "get-actions-output",
+  keys: () => ({
+    name: joi.string().required(),
+    kind: joi
+      .string()
+      .required()
+      .allow(...actionKinds)
+      .description(`The kind of action.`),
```
Orzelius, line 58:
```suggestion
      .description(`Action kind (e.g. Build).`),
```

That's how we refer to it elsewhere in the documentation

core/src/commands/get/get-actions.ts
```diff
@@ -0,0 +1,292 @@
...
+    type: joi.string().required().description(`The Type of the action.`),
```
Orzelius, line 59:
```suggestion
    type: joi.string().required().description(`Action Type (e.g. 'container').`),
```

core/src/commands/get/get-actions.ts
```diff
@@ -0,0 +1,292 @@
...
+      .allow(...actionStateTypes)
+      .only()
+      .required()
+      .description("The state of the action."),
+    path: joi.string().required().description("The relative path of the action config file."),
+    disabled: joi.boolean().required().description("Flag to identify if action is disabled."),
+    moduleName: joi
+      .string()
+      .description(
+        "The name of corresponding module name of an action. Only available if action is derived from module."
```
Orzelius, line 71:
```suggestion
        "The name of the module the action is derived from. Only available for converted actions."
```

core/src/commands/get/get-actions.ts
```diff
@@ -0,0 +1,292 @@
...
+    dependents: joiArray(actionConfigSchema()).description("Dependents of the action."),
+  }),
+})
+
+const getActionsArgs = {
+  actions: new StringsParameter({
+    help: deline`
+      Specify action(s) to list. You may specify multiple actions, separated by spaces.
+      Skip to return all actions.
+    `,
```
Orzelius, line 83: I don't think it's clear from this what's the parameter supposed to be. Is it action names, references or something else?

core/src/commands/get/get-actions.ts
```diff
@@ -0,0 +1,292 @@
...
+    getSuggestions: ({ configDump }) => {
+      return Object.keys(configDump.actionConfigs)
+    },
+  }),
+}
+
+const getActionsOpts = {
+  detail: new BooleanParameter({
+    help: deline`
+      Show the detailed info for each action, including state, path, dependencies, dependents, associated module and if action is disabled.
```
Orzelius, line 95:
```suggestion
      Show the detailed info for each action, including state, path, dependencies, dependents, associated module and if the action is disabled.
```

core/src/commands/get/get-actions.ts
```diff
@@ -0,0 +1,292 @@
...
+    const router = await garden.getActionRouter()
+    const graph = await garden.getResolvedConfigGraph({ log, emit: false })
+
+    const kindOpt = opts["kind"]
+    let actions: ResolvedActionWithState[] = []
+
+    switch (kindOpt) {
+      case "build":
+        actions = graph.getBuilds({ names: keys })
+        break
```
Orzelius, line 158: There's a handy `getActionsByKind()` on the graph class that would suit your needs here.

Maybe you could even add a `getActionsByKinds()` to the graph.

core/src/commands/get/get-actions.ts
```diff
@@ -0,0 +1,292 @@
...
+      actions.sort((a, b) => (a.type > b.type ? 1 : -1))
+    } else {
+      actions.sort((a, b) => (a.name > b.name ? 1 : -1))
+    }
+
+    // get state of each action
+    const actionsWithState = await Bluebird.map(actions, async (a) => {
+      a.state = await getActionState(a, router, graph, log)
+      return a
+    })
```
Orzelius, line 185: I don't think you should await the `getActionState()` resolution as that would make then resolve all one by one making it slow if many actions states are resolved. There's a point to be made about running into throttling issues if we fire all of them simultaneously but I say we deal with that problem once it comes our way.

Also, maybe the action state should be behind a flag so it's not resolved by default. I don't think it's needed in most cases and it might just make the command unnecessarily slow.

Orzelius: edit: it definitely shouldn't be resolved if `--detail=false`

Orzelius (commented):
here's a few comments, hope they help. I take that the get deploy command is still under construction so I didn't dig in too deep, but it definitely should reuse more code from the get actions command.

core/src/commands/get/get-actions.ts
shumailxyz: Valid point. I'll move state resolution behind a flag e.g. `--with-state`

core/src/commands/get/get-actions.ts
shumailxyz: its basically the same and calls `getActionsByKind` under the hood with `kind: build`.

core/src/commands/get/get-actions.ts
```diff
@@ -0,0 +1,249 @@
...
+        break
+      case "test":
+        actions = graph.getTests({ names: keys })
+        break
+      default:
+        actions = graph.getActions({ refs: keys })
+        break
+    }
+
+    if (opts["sort"] === "kind") {
```
Orzelius, line 165: I think there should always be a secondary sort by name alphabetically, would make stuff easier to find.

Feel free to skip adding this if you decide it's not important enough though. Stuff like this can be polished indefinitely.

core/src/commands/get/get-builds.ts
```diff
@@ -0,0 +1,90 @@
...
+  options = getBuildsOpts
+
+  outputsSchema = () =>
+    joi.object().keys({
+      actions: joiArray(getActionsCmdOutputSchema()).description("A list of the build actions."),
+    })
+
+  printHeader({ log }) {
+    printHeader(log, "Builds", "📖")
+  }
```
Orzelius, line 75: All of this should not be copy pasted across four commands. Only thing different about these is the action Kind. I would just create a single class and extend it four times. That base class should also reuse most of the logic from the get-actions class.

core/test/unit/src/commands/get/get-builds.ts
```diff
@@ -0,0 +1,62 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { GetBuildsCommand } from "../../../../../src/commands/get/get-builds"
+import { makeTestGarden, withDefaultGlobalOpts, getDataDir } from "../../../../helpers"
+import { expect } from "chai"
+
+describe("GetBuildsCommand", () => {
```
Orzelius, line 13: I think these four test suits should just be a single one looping the same tests. Currently it's not dry enough to my taste. I know loops for tests cases sounds funny at first but I think this is a valid case. I swear we already have a few somewhere in the codebase but I failed to find them.

Orzelius (commented):
I think there's a bit too much of repetition but other than that well done!

core/src/commands/get/get-builds.ts
shumailxyz: Valid point, thanks for pointing. I initially thought that that as most of the stuff is static like description and args etc, it would be better to keep in its own files however now that you commented, i also feel that its quite repetitive so moved to a shared class.

core/src/commands/get/get-actions.ts
shumailxyz: Makes sense. Will add a secondary sort by name.

Orzelius (approved):
Love it!!

core/src/commands/get/get-builds.ts
```diff
@@ -6,85 +6,10 @@
...
-      opts: {
-        ...params.opts,
-        kind: "build",
-      },
-    }
-    const getActionsCmd = new GetActionsCommand()
-    return getActionsCmd.action(getActionCmdParams)
+export class GetBuildsCommand extends GetActionsSubCommand {
+  constructor() {
+    super("Build")
```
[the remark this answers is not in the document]
shumailxyz: yeah, quite minimal 😄 Thanks for the review multiple times :)

core/src/actions/helpers.ts
```diff
@@ -98,3 +101,44 @@ const displayStates = {
...
+      state = (await router.run.getResult({ action: action as ResolvedRunAction, log: actionLog, graph }))?.result
+        ?.state
+      break
+    case "Test":
+      state = (await router.test.getResult({ action: action as ResolvedTestAction, log: actionLog, graph }))?.result
+        ?.state
+      break
+  }
+
+  return state
```
vvagaytsev, line 135: This variable can be removed, you can use immediate `return` statements in the `case` blocks and also skip the `break` operators.

core/src/commands/get/get-actions-subcommand.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+/**
+ * An abstract base class for get actions subcommands
+ * e.g. get builds, deploys, runs, tests.
+ * These commands are same as calling get actions command with option kind
+ */
+export abstract class GetActionsSubCommand extends GetActionsCommand {
+  protected kind: ActionKind
+  protected kindLowercaseString: string
```
vvagaytsev, line 35: These fields could be `private` and `readonly`.

core/src/commands/get/get-actions-subcommand.ts
```diff
@@ -0,0 +1,123 @@
...
+            case "Test":
+              return Object.keys(configDump.actionConfigs.Test)
+            case "Run":
+              return Object.keys(configDump.actionConfigs.Run)
+            case "Deploy":
+              return Object.keys(configDump.actionConfigs.Deploy)
+            default:
+              const _exhaustiveCheck: never = this.kind
+              return _exhaustiveCheck
+          }
```
vvagaytsev, line 66: This makes this class non-abstract :) Because this makes the class aware of all its sub-classes.

I'd prefer to introduce an abstract method here and implement it in every sub-class.

vvagaytsev (changes requested):
Thanks for implementing such a nice feature! 👍 
I've left a few comments, please check :)

core/src/commands/get/get-actions-subcommand.ts
shumailxyz: We could actually simplify it further like

```
getSuggestions: ({ configDump }) => {
  return Object.keys(configDump.actionConfigs[this.kind])
}
```

and i would prefer above instead of abstract method and implementing it in each subclass.

core/src/actions/helpers.ts
shumailxyz: removed

core/src/commands/get/get-actions-subcommand.ts
vvagaytsev: Nice! Elegant solution :)

vvagaytsev (approved):
Thank you! 🚀

Issue #4125, opened 2023-04-25
Update `garden migrate` command


vvagaytsev:
Include the following changes into the `migrate` command:
* #3512
* #3069
* Check workflows
* ... ?

vvagaytsev:
@thsig @eysi09 @Orzelius @Walther let's discuss and collect the list of the changes that must be done in the `migrate` command.

trymbill:
Looks like the description of the command also needs to be updated. It mentions `version 0.12` a couple of times, which I'm assuming is just from when it was used for migrating to `0.12`.

Orzelius:
will not be doing as a bit too tricky and questionable value

PR #4488, opened 2023-05-30
chore: remove garden migrate command

Walther:
In today's planning session, we decided to not proceed with updating the `garden migrate` command to support converting from `0.12` to `0.13`, and closed the relevant issue here https://github.com/garden-io/garden/issues/4125#issuecomment-1568155941

In lieu of updating it, we will remove the command, as it does not do what users might expect it to do these days.


vvagaytsev:
Technically, it's a breaking change. Should we mention that in the commit message body?

Walther:
Amended the commit message to include the breaking header, explanation, and link to the migration guide

vvagaytsev (approved):

Orzelius (approved):

PR #4489, opened 2023-05-30
fix(server): set correct default project root for autocomplete requests

edvald:
Also added some tests to validate.

cc @eysi09 @trymbill


core/src/events.ts
```diff
@@ -393,11 +395,14 @@ const isTaskGraphEvent = (name: string, _payload: any): _payload is Events[Graph
 }
 
 export function shouldStreamWsEvent(name: string, payload: any) {
-  if (isTaskGraphEvent(name, payload) && !skipTaskGraphEventTypes.includes(payload.type)) {
+  const gardenKey = payload?.$context?.gardenKey
+
+  if (isTaskGraphEvent(name, payload) && !skipTaskGraphEventTypes.includes(payload.type) && gardenKey) {
```
eysi09, line 400: Shouldn't we always check for the `gardenKey`, also for config events. 

And out of curiosity, in what cases is it not defined?

core/src/server/commands.ts
```diff
@@ -459,11 +459,15 @@ export async function getGardenForRequest({
...
   sessionId: string
 }) {
-  const cloudApi = await manager.getCloudApi({
-    log,
-    cloudDomain: getGardenCloudDomain(projectConfig.domain),
-    globalConfigStore,
-  })
+  let cloudApi: CloudApi | undefined
+
+  if (!command?.noProject) {
```
eysi09, line 464: This makes sense but I still feel like it's going to break something, considering all the back and forth we've had already. 

Let's make sure we do some manual testing before merging. I can do that on my end.

eysi09 (changes requested):

core/src/events.ts
edvald: No, this is intentional. Config updates may be detected and emitted outside of a particular Garden instance context. I still think we could add more context to those events, but filtering here would basically make them not appear at all on the server.

core/src/server/commands.ts
edvald: Please do, but note that this basically mimics the behavior of the CLI wrapper, so I couldn't see a reason not to make this change. Unifying those flows further is still the idea and trajectory of course, but this addresses the issue.

PR #4491, opened 2023-05-31
Update helm version from `3.7.2` to `3.12.0`

### LINKED ISSUE #3753: [FEATURE]: Helm version >= v3.8.0 / Use local helm installation
### Background / Motivation
Garden is currently shipped with Helm v3.7.2. This is problematic for those of us who use OCI registries (eg. Google Artifact Registry).

Can we please upgrade the bundled Helm binary?

We are using Google Artifact Registry for containers and helm charts. In later versions of Helm, its possible to re-use Docker's credential configuration (`gcloud` credential helper) for these registries. Since Garden runs Helm v3.7.2, I'm unable to authenticate via Helm modules. Container modules work fine for the same registry.

Users should be able to use Google Artifact Registries for both containers and helm charts, with authentication handled by the `gcloud` credential helper in the local Docker config.

It allows full usage for Google Artifact Registries for containers and helm charts; providing a private registry with a simple authentication mechanism.

Bump the bundled Helm binary to v3.11.1

vvagaytsev:
Updates `helm` from the ancient version `3.7.2` to the latest release `3.12.0`.

Fixes #3753

This is a _potentially_ breaking change. There were some breaking changes in helm CLI and SDK between the versions:
* helm 3.11.1 https://github.com/helm/helm/releases/tag/v3.11.1
* helm 3.9.0 https://github.com/helm/helm/releases/tag/v3.9.0

Those changes might not be breaking for Garden helm provider. We use `install`, `upgrade`, and `template` helm commands, but I haven't faced any issues yet. Tested with `garden deploy` in the [`vote-helm` example](https://github.com/garden-io/garden/tree/main/examples/vote-helm).

### DIFF
--- a/core/src/plugins/kubernetes/helm/helm-cli.ts
+++ b/core/src/plugins/kubernetes/helm/helm-cli.ts
@@ -13,16 +13,18 @@ import { PluginToolSpec } from "../../../plugin/tools"
 import split2 from "split2"
 import { Dictionary, pickBy } from "lodash"
 
+export const HELM_VERSION = "3.7.2"
+
 export const helm3Spec: PluginToolSpec = {
   name: "helm",
-  description: "The Helm CLI (version 3.x).",
+  description: `The Helm CLI (version ${HELM_VERSION}).`,
   type: "binary",
   _includeInGardenImage: true,
   builds: [
     {
       platform: "darwin",
       architecture: "amd64",
-      url: "https://get.helm.sh/helm-v3.7.2-darwin-amd64.tar.gz",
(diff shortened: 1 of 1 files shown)


Walther:
Looking at the linked changelog entries, I don't believe these changes are breaking - they seem mostly additive (new extra flags available and so on). In my opinion, we should backport this to 0.12 as well - it is likely that the ancient version of helm is "more breaking" for users in real-world scenarios than the potential changes in the new one.

Walther (approved):
Code changes look good to me, and the CI passes 🎉 
Thanks for the upgrade 🆕

vvagaytsev:
This change will be released in Garden `0.12.57` and `0.13.1`.

PR #4496, opened 2023-05-31
Fetch tools fixes

Orzelius:
please help yourself to the commits, they're rather chonky, sorry

### DIFF
--- a/core/src/commands/util/fetch-tools.ts
+++ b/core/src/commands/util/fetch-tools.ts
@@ -71,7 +71,11 @@ export class FetchToolsCommand extends Command<{}, FetchToolsOpts> {
       }
 
       if (garden instanceof DummyGarden) {
-        garden = await Garden.factory(garden.projectRoot, { ...omit(garden.opts, "config"), log })
+        garden = await Garden.factory(garden.projectRoot, {
+          ...omit(garden.opts, "config", "environmentString"),
+          log,
+          environmentString: opts.env,
+        })
       }
 
       plugins = await garden.getConfiguredPlugins()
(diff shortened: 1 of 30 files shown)


Orzelius:
some tests might fail, I'll fix in a separate pr

Orzelius:
Fixed the tests. Ready for review. @Walther Would this collide with #4381?

core/src/commands/util/fetch-tools.ts
```diff
@@ -90,10 +94,24 @@ export class FetchToolsCommand extends Command<{}, FetchToolsOpts> {
...
     const deduplicated = uniqBy(tools, ({ tool }) => tool["versionPath"])
 
+    const versionedConfigs = garden.getRawProviderConfigs({ names: ["pulumi", "terraform"], allowMissing: true })
+
+    // If the version of the tool is configured on the provider,
+    // download only that version of the tool.
+    const toolsNeeded = deduplicated.filter(tool => {
+      const pluginToolVersion = versionedConfigs.find(p => p.name === tool.plugin.name)?.version
+      const pluginHasVersionConfigured = !!pluginToolVersion
+      if (!pluginHasVersionConfigured) {
```
vvagaytsev, line 104: This looks a bit awkward, how about one line like `if (!pluginToolVersion)`?

core/src/util/util.ts
```diff
@@ -521,11 +521,21 @@ export function pickKeys<T extends object, U extends keyof T>(obj: T, keys: U[],
...
+export function findByNames<T extends ObjectWithName>({
+  names,
+  entries,
+  description,
+  allowMissing = false,
+}: {
+  names: string[]
+  entries: T[]
+  description: string
+  allowMissing?: boolean
```
vvagaytsev, line 533: It looks like `allowMissing` can be non-optional without the default `false` value. The value is always ensured by the `getRawProviderConfigs`.

core/src/commands/util/fetch-tools.ts
```diff
@@ -90,10 +94,24 @@ export class FetchToolsCommand extends Command<{}, FetchToolsOpts> {
     // No need to fetch the same tools multiple times, if they're used in multiple providers
     const deduplicated = uniqBy(tools, ({ tool }) => tool["versionPath"])
 
+    const versionedConfigs = garden.getRawProviderConfigs({ names: ["pulumi", "terraform"], allowMissing: true })
```
vvagaytsev, line 97: Do we need to check only `terraform` and `pulumi` providers here?

vvagaytsev (approved):
Looks good, thank you! 👍
Left a few non-blocking comments, please check.

core/src/commands/util/fetch-tools.ts
Orzelius: You mean the variable name? I chose it especially so it's not confused with the version of the tools available and the version that's configured on the plugin

vvagaytsev: The name is good, I meant the variable itself. But, tbh, the extra variable here brings more clarity 👍

vvagaytsev (approved):
🚢

Walther (approved):
nice work, thank you!

this shouldn't conflict with the other pr (which has been merged now)

Issue #4468, opened 2023-05-30
0.13: [Bug]: Pulumi module with dependencies does not validate


stefreak:
### Current Behavior

Using 0.13.0
```
steffen@MacBook-Pro repro2 % garden validate
Validate ✔️

[...]

foo: Error validating Module 'foo': key "dependencies" is not allowed at path [dependencies], key "build" is not allowed at path [build]. Available keys: allowDestroy, autoApply, createStack, root, pulumiVariables, pulumiVarfiles, orgName, cacheStatus, stackReferences, deployFromPreview, stack)
```

### Expected behavior

Using 0.12.56

```
steffen@MacBook-Pro repro2 % garden validate
Validate ✔️

[...]

OK ✔️
```

### Reproducible example

```
kind: Project
apiVersion: garden.io/v0
name: repro
environments:
 - name: default
providers:
 - name: local-kubernetes
 - name: pulumi

---

kind: Module
type: pulumi
name: foo
stack: "${environment.name}"

createStack: true
cacheStatus: false
allowDestroy: true
autoApply: true
deployFromPreview: false
dependencies:
  - baz

pulumiVarfiles:
  - "../global.yaml"
  - "config/common.yaml"
  - "config/${environment.name}.yaml"
build:
  dependencies:
    - bar

include: []

---
kind: Module
type: exec
name: bar
include: []

---
kind: Module
type: exec
name: baz
tasks:
  - name: baz
    command: [echo]
include: []
```

### Workaround

Workaround unknown so far, other than using Garden 0.12

### Suggested solution(s)

TBD

### Additional context

–

### Your environment

* **OS:** macOs
* **How I'm running Kubernetes:** Docker Desktop

`garden version` 0.13.0

PR #4497, opened 2023-05-31
fix: pulumi module validation

Walther:

Adds a separate spec for Pulumi modules and actions, so that we can validate old configurations correctly.

Fixes https://github.com/garden-io/garden/issues/4468

Taking inspiration from the terraform plugin, and how the module and action validation is handled there.
Pardon the seemingly large changes due to moving the `.keys()` call outside the function for reuse purposes - it is mostly indents.

### DIFF
--- a/plugins/pulumi/index.ts
+++ b/plugins/pulumi/index.ts
@@ -13,7 +13,13 @@ import { getPulumiCommands } from "./commands"
 
 import { joiVariables } from "@garden-io/core/build/src/config/common"
 import { pulumiCliSPecs } from "./cli"
-import { PulumiDeployConfig, pulumiDeploySpecSchema, PulumiModule, pulumiProviderConfigSchema } from "./config"
+import {
+  PulumiDeployConfig,
+  pulumiDeploySchema,
+  PulumiModule,
+  pulumiModuleSchema,
+  pulumiProviderConfigSchema,
+} from "./config"
 import { ExecBuildConfig } from "@garden-io/core/build/src/plugins/exec/config"
 import { join } from "path"
 import { pathExists } from "fs-extra"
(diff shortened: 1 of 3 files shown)


plugins/pulumi/index.ts
```diff
@@ -52,7 +58,7 @@ export const gardenPlugin = () =>
 
           Stack outputs are made available as action outputs. These can then be referenced by other actions under \`${actionOutputsTemplateString}\`. You can template in those values as e.g. command arguments or environment variables for other services.
           `,
-          schema: pulumiDeploySpecSchema(),
+          schema: pulumiDeploySchema(),
```
vvagaytsev, line 61: Should we keep the old name here? This scheme describes the `spec` field of the Pulumi deploy action, right?

vvagaytsev: It would be nice to align the naming convention between the plugins and use the same approach everywhere :)

plugins/pulumi/index.ts
```diff
@@ -101,20 +107,29 @@ export const gardenPlugin = () =>
...
+                allowDestroy: module.spec.allowDestroy || true,
+                autoApply: module.spec.autoApply || true,
+                createStack: module.spec.createStack || false,
+                pulumiVariables: module.spec.pulumiVariables || {},
+                pulumiVarfiles: module.spec.pulumiVarfiles || [],
+                cacheStatus: module.spec.cacheStatus || false,
+                stackReferences: module.spec.stackReferences || [],
+                deployFromPreview: module.spec.deployFromPreview || false,
+                root: module.spec.root || ".",
+                ...omit(module.spec, ["build", "dependencies"]),
```
vvagaytsev, line 128: Will this overwrite the default values defined above? Does it make sense to change the order and keep `...omit(module.spec, ["build", "dependencies"])` as the first statement in the `spec` object creation?

vvagaytsev (commented):
Looks good! 👍  I've left a couple of comments, please check :)

plugins/pulumi/index.ts
Walther: Yeah this entire spec block feels a bit sketchy due to the repetition - I posted a Discord thread for further discussion on this earlier today. The `...omit` seems to make typescript unhappy, and requires this repetition

Walther: Leaving this as something to fix later as we did not find a quick win for this

vvagaytsev (approved):
🚀

PR #4501, opened 2023-05-31
fix(kubernetes): do not fail to deploy List kinds (e.g. ConfigMapList)

### LINKED ISSUE #4500: 0.13: [Bug]: kind=Deploy type=kubernetes fails with `kind: ConfigMapList`
## Garden Bonsai (0.13) Bug
kubernetes Deploys fail with kubernetes manifests of type `kind: ConfigMapList`.

### Current Behavior
```
✖ deploy.test          → Failed processing Deploy type=kubernetes name=test (took 0.58 sec). Here is the output:

Command "/Users/steffen/.garden/tools/kubectl/f8a39834529cd1ed/kubectl --context=docker-desktop apply --output=json -f -" failed with code 1:

error: error validating "STDIN": error validating data: [ValidationError(ConfigMapList.metadata): unknown field "annotations" in io.k8s.apimachinery.pkg.apis.meta.v1.ListMeta, ValidationError(ConfigMapList.metadata): unknown field "labels" in io.k8s.apimachinery.pkg.apis.meta.v1.ListMeta]; if you choose to ignore these errors, turn validation off with --validate=false

error: error validating "STDIN": error validating data: [ValidationError(ConfigMapList.metadata): unknown field "annotations" in io.k8s.apimachinery.pkg.apis.meta.v1.ListMeta, ValidationError(ConfigMapList.metadata): unknown field "labels" in io.k8s.apimachinery.pkg.apis.meta.v1.ListMeta]; if you choose to ignore these errors, turn validation off with --validate=false
1 deploy action(s) failed!
```

### Expected behavior

Deploy succeeded

### Reproducible example

```
kind: Project
apiVersion: garden.io/v1
name: repro
environments:
 - name: default
providers:
 - name: local-kubernetes

---

kind: Deploy
type: kubernetes
name: test
source:
  repository:
    url: https://github.com/prometheus-operator/kube-prometheus.git#0198f98f682f1d7e40b1dd5615437a916107f1aa
spec:
  namespace: monitoring
  files:
    - manifests/grafana-dashboardDefinitions.yaml
```

### Workaround

Do not use `ConfigMapList` but use separete yaml documents with `ConfigMap` objects.

### Suggested solution(s)

Something like this could do the trick:

```
diff --git a/core/src/plugins/kubernetes/kubectl.ts b/core/src/plugins/kubernetes/kubectl.ts
index d15444f4d..05f380d51 100644
--- a/core/src/plugins/kubernetes/kubectl.ts
+++ b/core/src/plugins/kubernetes/kubectl.ts
@@ -71,13 +71,13 @@ export async function apply({
   // because kubectl applies some normalization/transformation that is sometimes difficult to reason about).
   // Hashing the input prevents "Too long annotation..." errors.
   for (const manifest of manifests) {
-    if (!manifest.metadata.annotations) {
-      manifest.metadata.annotations = {}
-    }
-    if (manifest.metadata.annotations[gardenAnnotationKey("manifest-hash")]) {
-      delete manifest.metadata.annotations[gardenAnnotationKey("manifest-hash")]
+    if (manifest.kind === "ConfigMapList") {
+      for (const configMap of manifest.items as KubernetesResource[]) {
+        await annotateResource(configMap)
+      }
+    } else {
+      await annotateResource(manifest)
     }
-    manifest.metadata.annotations[gardenAnnotationKey("manifest-hash")] = await hashManifest(manifest)
   }

   // The `--prune` option for `kubectl apply` currently isn't backwards-compatible, so here, we essentially
@@ -326,3 +326,13 @@ export const kubectlSpec: PluginToolSpec = {
     },
   ],
 }
+
+async function annotateResource(manifest: KubernetesResource) {
+  if (!manifest.metadata.annotations) {
+    manifest.metadata.annotations = {}
+  }
+  if (manifest.metadata.annotations[gardenAnnotationKey("manifest-hash")]) {
+    delete manifest.metadata.annotations[gardenAnnotationKey("manifest-hash")]
+  }
+  manifest.metadata.annotations[gardenAnnotationKey("manifest-hash")] = await hashManifest(manifest)
+}
```

### Your environment

* **OS:** macOS
* **How I'm running Kubernetes:** Docker desktop

`garden version`

stefreak:

Fixes #4500

### DIFF
--- a/core/src/plugins/kubernetes/kubernetes-type/common.ts
+++ b/core/src/plugins/kubernetes/kubernetes-type/common.ts
@@ -52,7 +52,17 @@ export async function getManifests({
   // remove *List objects
   const manifests = rawManifests.flatMap((manifest) => {
     if (manifest.kind.endsWith("List")) {
-      return manifest.items as KubernetesResource[]
+      if (!manifest.items || manifest.items.length === 0) {
+        // empty list
+        return []
+      } else if (manifest.items.length > 0 && manifest.items[0].kind) {
+        // at least the first manifest has a kind: seems to be a valid List
+        return manifest.items as KubernetesResource[]
+      } else {
+        throw new PluginError("Failed to read Kubernetes manifest: Encountered an invalid List manifest", {
+          manifest
+        })
+      }
     }
     return manifest
   })
--- a/core/src/plugins/kubernetes/kubernetes-type/common.ts
+++ b/core/src/plugins/kubernetes/kubernetes-type/common.ts
@@ -47,13 +47,22 @@ export async function getManifests({
   defaultNamespace: string
   readFromSrcDir?: boolean
 }): Promise<KubernetesResource[]> {
-  const manifests = await readManifests(ctx, action, log, readFromSrcDir)
+  const rawManifests = await readManifests(ctx, action, log, readFromSrcDir) as KubernetesResource[]
+
+  // remove *List objects
+  const manifests = rawManifests.flatMap((manifest) => {
+    if (manifest.kind.endsWith("List")) {
+      return manifest.items as KubernetesResource[]
+    }
+    return manifest
+  })
 
   return Bluebird.map(manifests, async (manifest) => {
     // Ensure a namespace is set, if not already set, and if required by the resource type
     if (!manifest.metadata?.namespace) {
       if (!manifest.metadata) {
-        manifest.metadata = {}
+        // TODO: Type system complains that name is missing
+        (manifest as any).metadata = {}
       }
 
       const info = await api.getApiResourceInfo(log, manifest.apiVersion, manifest.kind)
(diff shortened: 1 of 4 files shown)


Orzelius (dismissed):
Thanks for figuring this out!

Walther (approved):
🎉

stefreak:
@Orzelius @Walther @edvald I wonder if this would break if there was a custom resource definition with custom resources ending with `*List`. This code would fail then. Maybe it would be better to hardcode a list of specific kinds of resources instead of just looking for the `*List` ending. What do you think?

stefreak:
Maybe it would be safe enough to only flatMap the List if it actually has `items`– and otherwise keep it as-is

edvald:
I think we should only do this for resources with `apiVersion: v1`, and make sure we gracefully fail on invalid resources. If we do that, it should be all good.

stefreak:
[Kustomize also treats](https://github.com/kubernetes-sigs/kustomize/blob/cf3e81b590ab1fd7fc8f50849535f44bfea0d355/api/resource/factory.go#L550) anything with `*List` as a List, and errors if items is not a list. So I guess this is ok for now.

Orzelius (approved):

edvald:
I still think we should only do this for core API resources, basically just `apiVersion: v1`. Otherwise this may break some CRDs that can be named "SomethingList" and have completely unrelated semantics.

Issue #4508, opened 2023-06-01
0.13: [Bug]: Terraform module with dependencies does not validate


Walther:
This is essentially a copy of https://github.com/garden-io/garden/issues/4468 but for Terraform instead of Pulumi.

Reproducible with `garden/plugins/terraform/test/test-project-module/` and running `garden validate` on `0.13`.

```yaml
kind: Project
name: terraform-provider
environments:
  - name: local
providers:
  - name: terraform
    variables:
      my-variable: base
variables:
  workspace: default
---
kind: Module
type: terraform
name: tf
include: ["*"]
autoApply: true
root: ./tf
workspace: ${var.workspace}
variables:
  my-variable: foo
---
kind: Module
type: exec
name: test
include: ["*"]
tasks:
  - name: test-task
    dependencies: [tf]
    command: ["echo", "${runtime.services.tf.outputs.my-output}"]

```

```
gdev validate
Validate ✔️

"apiVersion" is missing in the Project config. Assuming "garden.io/v0" for backwards compatibility with 0.12. The "apiVersion"-field is mandatory when using the new action Kind-configs. A detailed migration guide is available at https://docs.garden.io/tutorials/migrating-to-bonsai
ℹ garden               → Running in Garden environment local.default
ℹ providers            → Getting status...
✔ providers            → Cached (took 0.3 sec)
ℹ providers            → Run with --force-refresh to force a refresh of provider statuses.
ℹ graph                → Resolving actions and modules...
✔ graph                → Done (took 0 sec)
✖ deploy.tf            → Failed processing resolve Deploy type=terraform name=tf (from module tf) (took 0.01 sec). Here is the output:

────────────────────────────────────────────────────────────────────────────────
Error validating spec for Deploy type=terraform name=tf (from module tf): key "build" is not allowed at path [build], key "dependencies" is not allowed at path [dependencies]. Available keys: allowDestroy, autoApply, root, variables, version, workspace)
────────────────────────────────────────────────────────────────────────────────
Failed to resolve Deploy type=terraform name=tf (from module tf): Error: resolve Deploy type=terraform name=tf (from module tf) failed: Error: Error validating spec for Deploy type=terraform name=tf (from module tf): key "build" is not allowed at path [build], key "dependencies" is not allowed at path [dependencies]. Available keys: allowDestroy, autoApply, root, variables, version, workspace)
```

Walther:
[...]

PR #4509, opened 2023-06-01
fix: terraform module validation

Walther:

Fixes https://github.com/garden-io/garden/issues/4508

We should make a release after merging this PR, as currently terraform module configurations will not work on Bonsai.
We should clean up these tests a bit some other day in a separate PR.

### DIFF
--- a/plugins/terraform/test/terraform.ts
+++ b/plugins/terraform/test/terraform.ts
@@ -727,4 +727,442 @@ for (const terraformVersion of ["0.13.3", defaultTerraformVersion]) {
       })
     })
   })
+
+  // TODO: re-enable after https://github.com/garden-io/garden/issues/4467 has been fixed
+  describe.skip("Terraform module type", () => {
+    const testRoot = join(__dirname, "test-project-module")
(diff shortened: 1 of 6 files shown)


plugins/terraform/test/terraform.ts
```diff
@@ -727,4 +727,442 @@ for (const terraformVersion of ["0.13.3", defaultTerraformVersion]) {
       })
     })
   })
+
+  // TODO: re-enable after https://github.com/garden-io/garden/issues/4467 has been fixed
+  describe.skip("Terraform module type", () => {
```
Walther, line 732: This is a rather ugly copy of the test suite for `Terraform action type` with a couple of changes (e.g. using `garden["moduleConfigs"]["tf"]` instead of `garden["actionConfigs"]["Deploy"]["tf"]`).

These tests could be cleaned up a lot, but we should leave that for later.
These tests pass locally, and they make sure our module-to-action conversion logic works

vvagaytsev (approved):
Thank you! 👍

Issue #4284, opened 2023-05-16
chore: set up linting for the `plugins` directory


Walther:
See https://github.com/garden-io/garden/pull/4281

This should be done on the `0.13` branch, which uses ESLint.

PR #4512, opened 2023-06-01
chore(plugins): include test directories in linting

Walther:

Fixes https://github.com/garden-io/garden/issues/4284


plugins/pulumi/package.json
```diff
@@ -21,10 +21,10 @@
   },
   "scripts": {
     "fix-format": "prettier --write \"**/*.ts\"",
-    "lint": "eslint -c ../../.eslintrc  --ignore-pattern 'test/**' --ext .ts .",
+    "lint": "eslint -c ../../.eslintrc  --ignore-pattern 'test/test-project-k8s/**' --ext .ts .",
```
vvagaytsev, line 24: Why do we need to ignore this project? We have 3 test projects in the `pulumi` plugin, but we don't ignore those.

vvagaytsev (dismissed):
Nice! ❇️

plugins/pulumi/package.json
Walther: Bah, this "cleanup" commit sadly breaks the build. Including the example pulumi project directories in the tsconfig means that we would need to also ensure the pulumi libraries (and their types) get installed, in order to make the typescript compiler happy. Because we use ESLint with the TypeScript plugins, we cannot "just lint" without TS being happy.

Do we want to
- revert this commit and just ignore the `test-project-k8s` directory (including the two test project subdirectories within it),
- or do we want to edit our build script to ensure a `yarn install` within the pulumi examples directory every time you do a build for the main project?
I'm leaning towards the first option, as the example projects are only a few lines of code in two `index.ts` files.

Walther: Reverted, if you feel like this is a good option feel free to approve and I'll do a squash-merge to clean up.
If you feel like we need to solve this in another way, feel free to request changes 🙇‍♂️

shumailxyz: i would also go for option # 1 and ignore the test-project-k8s directory. IMO it's a fair tradeoff instead of editing our build script and increasing the build time.

shumailxyz (approved):

plugins/pulumi/package.json
vvagaytsev: Late to the party, but totally agree with @shumailxyz :)

PR #4514, opened 2023-06-01
Fix some linter warns

vvagaytsev:

Fixes #

### DIFF
--- a/core/test/unit/src/config/template-contexts/base.ts
+++ b/core/test/unit/src/config/template-contexts/base.ts
@@ -59,7 +59,7 @@ describe("ConfigContext", () => {
     context("allowPartial=true", () => {
       it("should throw on missing key when allowPartial=true", async () => {
         const c = new TestContext({})
-        expectError(() => resolveKey(c, ["basic"], { allowPartial: true }), {
+        await expectError(() => resolveKey(c, ["basic"], { allowPartial: true }), {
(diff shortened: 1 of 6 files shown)


core/test/unit/src/config/template-contexts/base.ts
```diff
@@ -59,7 +59,7 @@ describe("ConfigContext", () => {
     context("allowPartial=true", () => {
       it("should throw on missing key when allowPartial=true", async () => {
         const c = new TestContext({})
-        expectError(() => resolveKey(c, ["basic"], { allowPartial: true }), {
+        void expectError(() => resolveKey(c, ["basic"], { allowPartial: true }), {
```
Walther, line 62: Can we use `void` here or should we use `await` on these? I'm not 100% sure how this works with mocha, are the test assertions actually completed for the test cases?

vvagaytsev: Ah, you're right! The function can return a `Promise`, and we use `await` in the other code places.

vvagaytsev: Fixed in https://github.com/garden-io/garden/pull/4514/commits/1ca387cd3417278f78d069246da79dcbb87b2d4f

Walther (approved):
Looks good, thanks for the cleanup! ✨ 🧹

Issue #4504, opened 2023-05-31
Investigate microk8s test failures


Walther:
The `microk8s` tests in our CI are extremely flaky, to the point of _almost always_ failing.

We should figure out why this is happening and solve it properly.

PR #4518, opened 2023-06-02
WIP: experiments with micro-k8s tests

vvagaytsev:
No quick wins in the investigations. I've just increased the timeout for the integration tests. It made things better.

Addresses #4504


Walther (approved):
Thanks for looking into this!

Orzelius:
@vvagaytsev could you replace the .25 and .26 microk8s tests with something else that's stable. I think that's what we agreed to do if we don't figure out what's wrong with them. Having a green ci would be very nice.

PR #4521, opened 2023-06-02
0.12 rename `edge` tag to `edge-acorn`

vvagaytsev:

Fixes #

### DIFF
--- a/scripts/release.ts
+++ b/scripts/release.ts
@@ -342,8 +342,8 @@ async function stripPrereleaseTags(tags: string[], version: string) {
     }
   }
 
-  // We also need to remove the "edge-acorn" tag
-  await execa("git", ["tag", "-d", "edge-acorn"])
+  // We also need to remove the old edge tags
+  await removeTags("edge-acorn", "edge")
 }
 
 (async () => {


vvagaytsev:
Keeping this a draft for now. To avoid accidental premature merge.

stefreak (dismissed):

scripts/release.ts
```diff
@@ -326,8 +326,8 @@ async function stripPrereleaseTags(tags: string[], version: string) {
     }
   }
 
-  // We also need to remove the "edge" tag
-  await execa("git", ["tag", "-d", "edge"])
+  // We also need to remove the "edge-acorn" tag
+  await execa("git", ["tag", "-d", "edge-acorn"])
```
stefreak, line 330: Might make sense to just delete all kinds of edge tags
```suggestion
  await execa("git", ["tag", "-d", "edge"])
  await execa("git", ["tag", "-d", "edge-acorn"])
  await execa("git", ["tag", "-d", "edge-bonsai"])
```

vvagaytsev: Good catch, thanks! I'm not sure about the old `edge` tag. We can do it manually only once. I'm not sure if `execa` will raise an exception if the tag is not found. I'll check that.

vvagaytsev: Actually, `edge-bonsai` should not be deleted here. It belongs to a separate 0.13 release process and must be deleted in the main branch here: https://github.com/garden-io/garden/blob/main/scripts/release.ts#L330

I'll create a separate PR to fix that.

vvagaytsev: Added a [helper function](https://github.com/garden-io/garden/pull/4521/commits/2f813f09b4ea07feb09bb37fdbd920527ec017b3) for fail-safe tags removal, and [removed](https://github.com/garden-io/garden/pull/4521/commits/17b8736b359cad6357a00731554f57e1b36c68d0) both `edge` and `edge-acorn`.

stefreak: Awesome

PR #4523, opened 2023-06-02
fix various pvc issues

Orzelius:
fixes: #4524


core/src/plugins/container/container.ts
```diff
@@ -209,7 +211,7 @@ function convertContainerModuleRuntimeActions(
   convertParams: ConvertModuleParams<ContainerModule>,
   buildAction: ContainerBuildActionConfig | ExecBuildConfig | undefined,
   needsContainerBuild: boolean
-): ContainerActionConfig[] {
+) {
```
Walther, line 214: is the removal of the return type intentional here? (does this end up inferred to the correct type, or does it end up as any?)

core/src/plugins/container/container.ts
```diff
@@ -219,6 +221,24 @@ function convertContainerModuleRuntimeActions(
...
 
+  const volumeModulesReferenced: string[] = []
+  function configureActionVolumes(action: ContainerRuntimeActionConfig, volumeSpec: ContainerModuleVolumeSpec[]) {
+    volumeSpec.forEach((v) => {
+      const referencedPvcAction = v.module ? { kind: <"Deploy">"Deploy", name: v.module } : undefined
+      action.spec.volumes.push({
+        ...omit(v, "module"),
+        action: referencedPvcAction,
+      })
+      if (referencedPvcAction) {
```
Walther, line 232: do we intentionally push `action: undefined` in the case `v.module` is falsy?

core/src/plugins/container/container.ts
```diff
@@ -559,8 +575,7 @@ export const gardenPlugin = () =>
           {
             platform: "windows",
             architecture: "amd64",
-            url:
-              "https://github.com/rgl/docker-ce-windows-binaries-vagrant/releases/download/v20.10.9/docker-20.10.9.zip",
+            url: "https://github.com/rgl/docker-ce-windows-binaries-vagrant/releases/download/v20.10.9/docker-20.10.9.zip",
```
Walther, line 578: (random sidenote: we should probably consider updating the versions here some day - it appears the docker cli version 20.10.9 is almost two years old)

Walther (dismissed):
Looks good as far as I can tell, but others feel free to chime in as well in case you have thoughts 👍

core/src/plugins/container/container.ts
vvagaytsev: The return type gets inferred correctly. It was changed to be `{ volumeModulesReferenced: string[]; actions: ContainerActionConfig[] }`. Explicit anonymous types like that can look awkward.

I guess we are fine with omitting the types that can be inferred correctly :) At least, our codebase has many functions without the explicitly defined return type.

core/src/plugins/container/container.ts
```diff
@@ -324,8 +344,15 @@ export async function convertContainerModule(params: ConvertModuleParams<Contain
...
 
-  const runtimeActions = convertContainerModuleRuntimeActions(params, buildAction, needsContainerBuild)
+  const { actions: runtimeActions, volumeModulesReferenced } = convertContainerModuleRuntimeActions(
+    params,
+    buildAction,
+    needsContainerBuild
+  )
   actions.push(...runtimeActions)
+  if (buildAction) {
+    buildAction.dependencies = buildAction?.dependencies?.filter((d) => !volumeModulesReferenced.includes(d.name))
```
vvagaytsev, line 354: A minor nit :) The statement `buildAction?.dependencies?` in the right-hand side can be simplified to `buildAction.dependencies?` because `buildAction` is known to be non-falsy here.

vvagaytsev (dismissed):
Looks good, thanks for the fixes! 🚀

shumailxyz (approved):

core/src/plugins/container/container.ts
Orzelius: yes

vvagaytsev (approved):

PR #4525, opened 2023-06-02
chore: use `edge-bonsai` tag instead of `edge`

vvagaytsev:

Fixes #

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -512,7 +512,7 @@ jobs:
     steps:
       - *attach-workspace
       - run:
-          name: Publish a pre-release on GitHub with the tag 'edge'
+          name: Publish a pre-release on GitHub with the valid edge tag
           command: |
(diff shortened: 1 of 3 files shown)


stefreak (approved):

vvagaytsev:
@stefreak should we change https://github.com/garden-io/garden/blob/main/.circleci/config.yml#L433? It looks like we should not. It's used for Docker build and there we also have a codename defined. Let's just make sure that the codename is set correctly in `main` and `0.12` branches.

stefreak:
> @stefreak should we change https://github.com/garden-io/garden/blob/main/.circleci/config.yml#L433? It looks like we should not. It's used for Docker build and there we also have a codename defined. Let's just make sure that the codename is set correctly in `main` and `0.12` branches.

Yes correct, this one affects the naming of our images on Dockerhub. We want to keep them as-is. See also https://docs.garden.io/reference/dockerhub-containers

PR #4532, opened 2023-06-05
chore: re-enable pulumi tests

Walther:

Fixes https://github.com/garden-io/garden/issues/4471

Sidenote: we should probably add more Pulumi tests at some point, to increase our confidence and coverage.


plugins/pulumi/test/test-project-k8s/k8s-namespace/garden.yml
```diff
@@ -4,6 +4,7 @@ name: k8s-namespace
 description: Creates a Namespace
 pulumiVarfiles: [varfile1.yaml, varfile2.yaml]
 createStack: true
+cacheStatus: true
```
Walther, line 7: we don't tag the stack with garden version unless cacheStatus is set https://github.com/garden-io/garden/blob/920e0bf775ccc63970ac5deffbb54ed9ba38cc55/plugins/pulumi/handlers.ts#L162-L164
but we expect to find that tag in the test cases

vvagaytsev (approved):
👍

PR #4533, opened 2023-06-05
fix(pulumi): Fix pulumi stack tag rm error

theoribeiro:
Whenever deleting an environment containing a pulumi stack using a self-hosted backend (which doesn't support stack tags), Garden is yielding an error when trying to delete the stack tags and therefore stopping before it deletes dependant modules.  

This PR fixes the propagation of said error by adding a try-catch clause like similar functions in the pulumi helpers module.


vvagaytsev (approved):
Thank you! 👍

Issue #3778, opened 2023-02-16
Switch from `joi` to `zod` in 0.13


Walther:

eysi09:
Are we planning on doing this with the 0.13 release?

PR #4535, opened 2023-06-05
refactor: improved plugin definition SDK interface, phase 1

edvald:
This introduces a more friendly wrapper for defining plugins. One of the key benefits is that plugin authors don't need to think as much about types and implicitly get solid type-safety when defining handlers.

To start, the `hadolint` plugin has been ported to the new SDK.

This will need more iteration, particularly relating to ease of use when defining a plugin and its handlers across multiple modules. A good next step will be to convert the `exec` plugin. Once this is more fleshed out we'll also want to document this more thoroughly.

The implementation involves using Zod schemas instead of the non-typesafe Joi schemas. The underlying framework is unchanged in those terms, and the Zod schemas are wrapped with a custom Joi method. This allows gradual migration.

### DIFF
--- a/core/src/actions/base.ts
+++ b/core/src/actions/base.ts
@@ -51,6 +51,7 @@ import {
   BaseActionConfig,
   ExecutedAction,
   ExecutedActionWrapperParams,
+  GetOutputValueType,
   ResolvedAction,
   ResolvedActionWrapperParams,
 } from "./types"
@@ -316,7 +317,11 @@ export interface ActionDescriptionMap {
   [key: string]: ActionDescription
 }
 
-export abstract class BaseAction<C extends BaseActionConfig = BaseActionConfig, Outputs extends {} = any> {
+export abstract class BaseAction<
+  C extends BaseActionConfig = BaseActionConfig,
+  StaticOutputs extends {} = any,
+  RuntimeOutputs extends {} = any
+> {
   // TODO: figure out why kind and type come out as any types on Action type
   public readonly kind: C["kind"]
   public readonly type: C["type"]
@@ -327,9 +332,8 @@ export abstract class BaseAction<C extends BaseActionConfig = BaseActionConfig, 
 
   // Note: These need to be public because we need to reference the types (a current TS limitation)
   _config: C
-  // TODO: split the typing here
-  _outputs: Outputs
-  protected _staticOutputs: Outputs
+  _staticOutputs: StaticOutputs
+  _runtimeOutputs: RuntimeOutputs
 
   protected readonly baseBuildDirectory: string
   protected readonly compatibleTypes: string[]
@@ -640,8 +644,9 @@ export abstract class BaseAction<C extends BaseActionConfig = BaseActionConfig, 
 
 export abstract class RuntimeAction<
   C extends BaseRuntimeActionConfig = BaseRuntimeActionConfig,
-  Outputs extends {} = any
-> extends BaseAction<C, Outputs> {
+  StaticOutputs extends {} = any,
+  RuntimeOutputs extends {} = any
+> extends BaseAction<C, StaticOutputs, RuntimeOutputs> {
   /**
    * Return the Build action specified on the `build` field if defined, otherwise null
    */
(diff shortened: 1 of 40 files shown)


core/src/actions/base.ts
```diff
@@ -791,12 +807,15 @@ export abstract class ExecutedRuntimeAction<
     return this.status
   }
 
-  getOutput<K extends keyof O>(key: K) {
-    return this.status.outputs[key] || this._staticOutputs[key]
+  getOutput<K extends keyof (StaticOutputs & RuntimeOutputs)>(
+    key: K
+  ): GetOutputValueType<K, StaticOutputs, RuntimeOutputs> {
+    // FIXME: unsure how to avoid the any cast here, but usage is unaffected
+    return <any>(this.status.outputs[<keyof RuntimeOutputs>key] || this._staticOutputs[<keyof StaticOutputs>key])
```
vvagaytsev, line 814: It looks like this cast to `any` is no longer necessary in TypeScript 5.1.3 that has been in use since #4560.

core/src/actions/base.ts
```diff
@@ -770,19 +778,27 @@ export abstract class ResolvedRuntimeAction<
...
-  _Outputs extends {} = any
-> {}
+  StaticOutputs extends {} = any,
+  RuntimeOutputs extends {} = any
+> {
+  getOutput<K extends keyof (StaticOutputs & RuntimeOutputs)>(
+    key: K
+  ): GetOutputValueType<K, StaticOutputs, RuntimeOutputs>
+  getOutputs(): StaticOutputs & RuntimeOutputs
+}
```
vvagaytsev, line 788: This type looks unnecessary because all its descendants already override the same methods defined in `ResolvedBuildAction` or `ResolvedRuntimeAction`.

But maybe it's ok to keep it here, just to emphasize that every `Executed` action must define this pair of methods.

core/src/config-store/global.ts
```diff
@@ -75,7 +75,7 @@ const globalConfigSchema = z.object({
     passed: z.boolean().optional().describe("Whether the last check passed the requirements."),
   }),
 })
-
+type conf = typeof globalConfigSchema
```
vvagaytsev, line 78: Unused declaration. Should we remove this?

core/src/actions/base.ts
edvald: I'll keep the type but can remove redundant methods.

core/src/docs/joi-schema.ts
```diff
@@ -126,6 +127,28 @@ export class JoiKeyDescription extends BaseKeyDescription {
...
+
+        childDescriptions.push(
+          ...Object.entries(jsonSchema["properties"] || {}).map(
+            ([childName, schema]) =>
+              new JsonKeyDescription({
+                schema,
+                name: childName,
+                level: nextLevel,
+                parent,
+                required: jsonSchema["required"] && jsonSchema["required"].includes(childName),
```
vvagaytsev, line 147: Could this be simplified to `jsonSchema["required"]?.includes(childName)`?

edvald: I recall it not working when I attempted it, but forget why.

core/src/plugins/container/publish.ts
```diff
@@ -16,6 +16,7 @@ export const publishContainerBuild: BuildActionHandler<"publish", ContainerBuild
   log,
   tag,
 }) => {
+  const outputs = action.getOutputs()
```
vvagaytsev, line 19: It looks like these outputs should be used in [L42](https://github.com/garden-io/garden/pull/4535/files#diff-7b270d8c42a648cd11deffccf1782cc44121d975f1a30f48430de6c9c4f73013R41), is that correct?

core/src/actions/base.ts
vvagaytsev: Ah, ok, thanks!

core/src/docs/joi-schema.ts
vvagaytsev: ok, nevermind :)

vvagaytsev:
@edvald now there are some merge conflicts. Could you rebase this on top of the `main`, please?

edvald:
@vvagaytsev done 👍

vvagaytsev:
This partially addresses #3778

PR #4539, opened 2023-06-06
fix: show short url for cloud

### LINKED ISSUE #4295: [FEATURE]: Show a shortlink for cloud project URL instead of long links
## Feature Request
When a user connects to cloud using `garden login`, the cloud project link that's provided to user is quite long at the moment and doesn't fit the cli aesthetics. Currently the link looks like this

```
https://app.garden.io/projects/4fc1331d-7742-43a4-9f16-acf90481c68a?sessionId=4fc1331d-7742-43a4-9f16-acf90481c68a&userId=4fc1331d-7742-43a4-9f16-acf90481c68a
```

Would be great aesthetically if a shortlink is provided instead of massive url.

### Background / Motivation

The massive url doesn't look nice, and goes on multiple lines on small window cli atm. A shortlink instead could be much smaller and would fit better aesthetically.

Once session is registered to cloud, core should receive a short link in response from cloud that it shows to the user. The short link could look like: 

```
https://app.garden.io/go/9w5j
```

shumailxyz:
Instead of current lengthy urls in cli, it shows a short url of  format <domain>/go/<shortlink>.

Fixes #4295

### DIFF
--- a/core/src/cloud/api.ts
+++ b/core/src/cloud/api.ts
@@ -94,6 +94,7 @@ export type ApiFetchResponse<T> = T & {
 export interface CloudSessionResponse {
   environmentId: number
   namespaceId: number
+  shortId: string
 }
 
 export interface CloudSession extends CloudSessionResponse {
@@ -693,8 +694,21 @@ export class CloudApi {
     return new URL(`/projects/${projectId}`, this.domain)
   }
 
-  getCommandResultUrl({ projectId, sessionId, userId }: { projectId: string; sessionId: string; userId: string }) {
-    const path = `/projects/${projectId}?sessionId=${sessionId}&userId=${userId}`
+  getCommandResultUrl({
+    projectId,
+    sessionId,
+    userId,
+    shortId,
+  }: {
+    projectId: string
+    sessionId: string
+    userId: string
+    shortId?: string
+  }) {
+    let path = `/projects/${projectId}?sessionId=${sessionId}&userId=${userId}`
+    if (shortId) {
+      path = `/go/${shortId}`
+    }
     return new URL(path, this.domain)
   }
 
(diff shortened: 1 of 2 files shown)


core/src/commands/base.ts
```diff
@@ -225,6 +246,7 @@ export abstract class Command<A extends Parameters = {}, O extends Parameters =
       const distroName = getCloudDistributionName(cloudSession.api.domain)
       const userId = (await cloudSession.api.getProfile()).id
       const commandResultUrl = cloudSession.api.getCommandResultUrl({
+        shortLink: cloudSession.shortLink,
```
shumailxyz, line 249: This is the only change here. Rest of the changes is prettier being prettier.

core/src/commands/base.ts
```diff
@@ -269,13 +290,14 @@ export abstract class Command<A extends Parameters = {}, O extends Parameters =
       const distroName = getCloudDistributionName(cloudSession.api.domain)
       const userId = (await cloudSession.api.getProfile()).id
       const commandResultUrl = cloudSession.api.getCommandResultUrl({
+        shortId: cloudSession.shortId,
```
shumailxyz, line 293: This is the only important change here. Rest of the stuff is prettier being prettier.

core/src/cloud/api.ts
```diff
@@ -693,8 +694,21 @@ export class CloudApi {
...
+    sessionId,
+    userId,
+    shortId,
+  }: {
+    projectId: string
+    sessionId: string
+    userId: string
+    shortId?: string
+  }) {
+    let path = `/projects/${projectId}?sessionId=${sessionId}&userId=${userId}`
```
vvagaytsev, line 708: Nit :) This can be `const` with a ternary operator. No need to create this string object if `shortId` is `true`.

shumailxyz: True. I just find it bit more readable compared to ternary operator here as the string is bit long.

PR #4555, opened 2023-06-06
fix(k8s): handle intermittent socket hang up errors

edvald:
@vvagaytsev could we get this into an 0.12 patch release, and also port to 0.13? This addresses a customer issue.


vvagaytsev (approved):
👍

PR #4560, opened 2023-06-07
chore: update TS to 5.1.3

TimBeyer:
Bumps TS to 5.1.3

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/api.ts
+++ b/core/src/plugins/kubernetes/api.ts
@@ -40,6 +40,7 @@ import { Omit, sleep, StringCollector } from "../../util/util"
 import { flatten, isObject, isPlainObject, keyBy, omitBy } from "lodash"
 import { ConfigurationError, GardenBaseError, RuntimeError } from "../../exceptions"
 import {
+  BaseResource,
   KubernetesList,
   KubernetesPod,
   KubernetesResource,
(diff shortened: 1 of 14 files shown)


core/src/plugins/kubernetes/api.ts
```diff
@@ -147,7 +148,9 @@ interface List {
   items?: Array<any>
 }
 
-type WrappedList<T extends List> = T["items"] extends Array<infer V> ? KubernetesServerList<V> : KubernetesServerList
+type WrappedList<T extends List> = T["items"] extends Array<infer V extends BaseResource | KubernetesObject>
```
eysi09, line 151: Right, I got stuck on this one when testing this earlier. Good job!

core/src/template-string/template-string.ts
```diff
@@ -211,11 +211,19 @@ export function resolveTemplateString(string: string, context: ConfigContext, op
 /**
  * Recursively parses and resolves all templated strings in the given object.
  */
-export const resolveTemplateStrings = profile(function $resolveTemplateStrings<T>(
-  value: T,
+
+// `extends any` here isn't pretty but this function is hard to type correctly
```
eysi09, line 215: Would `unknown` work here? 

If not, `any` is fine since we do a lot of runtime checks. But feels like `unknown` is more correct if it can be used.

eysi09 (approved):
Awesome! Had one question, but approving in case it's not relevant.

Issue #4513, opened 2023-06-01
0.13: [Bug]: Deploy action depends on .garden/build which is entirely absent if the build was skipped


spenczar:
### Current Behavior

Suppose you have a `container` Build action, and a `helm` Deploy action, with the source Dockerfile and Helm chart in the same module. A minimal example:
```yml
kind: Build
type: container
name: bugdemo-image

---

kind: Deploy
name: bugdemo-deploy
type: helm

build: bugdemo-image
dependencies:
  - build.bugdemo-image

spec:
  chart:
    path: ./chart
```
This is in a Garden project that looks like this:
```
├── bugdemo
│   ├── Dockerfile
│   ├── chart
│   │   ├── Chart.yaml
│   │   ├── charts
│   │   ├── templates
│   │   └── values.yaml
│   └── garden.yml
└── project.garden.yml
```

If you `garden build` this project in one context, and then `garden deploy` it in another one (like a different developer's laptop, or CI, or even a different folder on the same machine), then the deploy will fail because it won't find the chart.

### Expected behavior

The deploy should use the helm chart in the module.

### Reproducible example

I made a repository that does it all here: https://github.com/spenczar/garden-debug-missing-build

1. Clone the repo.
2. `garden build`
3. `rm -rf .garden`
4. `garden deploy`

I see this:
```
$ garden deploy --log-level=debug
Deploy 🚀

ℹ cloud-dashboard [debug] → Initializing Garden Cloud API client.
[debug] No auth token found, proceeding without access to Cloud Dashboard. Command results for this command run will not be available in Cloud Dashboard.
[debug] Error reading safe directories from the .gitconfig: Error: Command "git config --get-all safe.directory" failed with code 1:
Garden v0.13 (Bonsai) is a major release with significant changes. Please help us improve it by reporting any issues/bugs here:
https://go.garden.io/report-bonsai
→ Run garden util hide-warning 0.13-bonsai to disable this warning.
ℹ garden               → Running in Garden environment local.default
[debug] Scanning project root at /Users/swnelson/code/scratch/garden-debug-missing-build
→ Includes: **/*garden.y*ml
→ Excludes: .garden/**/*,.git,.gitmodules,.garden/**/*,debug-info*/**,.garden/**/*
[debug] Found 2 files in project root /Users/swnelson/code/scratch/garden-debug-missing-build
[debug] Scanned and found 2 actions, 0 workflows and 0 modules
ℹ providers            → Getting status...
[...]

PR #4561, opened 2023-06-07
fix: ensure build context on ´ready´ status builds

Orzelius:
Ensure build context being available even if the plugin handler returns `ready` status for the action. For example if a container is already built but the build context has additional files needed for the Deploy.

closes: https://github.com/garden-io/garden/issues/4513

### DIFF
--- a/core/src/build-staging/build-staging.ts
+++ b/core/src/build-staging/build-staging.ts
@@ -73,6 +73,11 @@ export class BuildStaging {
     })
   }
 
+  async actionBuildPathExists({ action }: { action: BuildAction }) {
+    const buildPath = action.getBuildPath()
+    return this.createdPaths.has(buildPath) || pathExists(buildPath)
+  }
+
   async syncDependencyProducts(action: BuildAction, log: Log) {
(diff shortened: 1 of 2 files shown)


core/src/build-staging/build-staging.ts
```diff
@@ -73,6 +73,11 @@ export class BuildStaging {
     })
   }
 
+  async actionBuildPathExists({ action }: { action: BuildAction }) {
```
vvagaytsev, line 76: While the [parameter object pattern](https://refactoring.guru/introduce-parameter-object) pattern is good, this looks like the case when we won't have any extra arguments.

So this could be simplified to `actionBuildPathExists(action: BuildAction)`

Orzelius: yeah, [7fed0a8](https://github.com/garden-io/garden/pull/4561/commits/7fed0a8e0372ce2bd68de2204af3b2a2b9661453)

vvagaytsev (approved):
Thanks for fixing this! 👍

Issue #4515, opened 2023-06-01
0.13: [Bug]: garden self-update doesn't update version in folder path


spenczar:
### Current Behavior

I originally installed garden with Homebrew, as instructed here: https://docs.garden.io/guides/installation#step-2-install-garden-macos. That was with version 0.12.48.

I later did `garden self-update 0.13.0`.

I think that worked, but it left some confusing state around. I see things in logs like
```
ℹ garden system [debug] → Initializing...
ℹ garden system [debug] → Scanning project root at /opt/homebrew/Cellar/garden-cli/0.12.48/libexec/static/kubernetes/system
```

What's that `0.12.48` doing in there? Apparently that's just the installation directory:

```
$ garden self-update 0.13.0
Update Garden 🗞️

Checking for target and latest versions...
Installation directory: /opt/homebrew/Cellar/garden-cli/0.12.48/libexec
Current Garden version: 0.13.0
Target Garden version to be installed: 0.13.0
Latest release version: 0.13.0
```

### Expected behavior

I'd expect that `/opt/homebrew/Cellar/garden-cli/0.12.48/libexec` to be `/opt/homebrew/Cellar/garden-cli/0.13.0/libexec`.

### Workaround

n/a

### Suggested solution(s)

--
### Your environment

* **OS:** MacOS
* **How I'm running Kubernetes:** n/a 

`garden version` 0.13.0

stefreak:
@spenczar I understand that you'd like the correct Garden version is in the directory path– in case of installation using homebrew, which behaviour would you prefer after `self-update` detected that Garden has been installed using Homebrew:
1. Run `brew update && brew install garden-cli` automatically instead of self-updating in-place
2. Offer to run `brew update && brew install garden-cli` interactively.
3. Print an error message suggesting to run `brew update && brew install garden-cli`

spenczar:
I think my order of preferences there would be 2, then 3, and lowest preference would be 1.

stefreak:
@spenczar thank you for letting us know! Did you have another solution in mind when writing this, that you would actually prefer?

spenczar:
> @spenczar thank you for letting us know! Did you have another solution in mind when writing this, that you would actually prefer?

Not really. I don't know anything about how garden self-update works under the hood, nor how homebrew works. I just expect the path to reflect the version I have installed, that's really the extent of it. I spent about 20 minutes believing a bug I encountered was possibly because of a stale version, and was confused; it would be good to prevent that confusion for others.

[...]

PR #4562, opened 2023-06-07
chore: clearer logging in the `self-update` command

vvagaytsev:
Clearer logging about the installation directory in the `self-update` command.

This mitigates #4515

I preferred to avoid using any platform-specific code here to infer whether the Garden has been installed with the homebrew on a user's machine.

### DIFF
--- a/core/src/commands/self-update.ts
+++ b/core/src/commands/self-update.ts
@@ -197,12 +197,12 @@ export class SelfUpdateCommand extends Command<SelfUpdateArgs, SelfUpdateOpts> {
     let platform = opts.platform
 
     if (!installationDirectory) {
+      installationDirectory = dirname(process.execPath)
       log.info(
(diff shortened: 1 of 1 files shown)


core/src/commands/self-update.ts
```diff
@@ -197,7 +197,14 @@ export class SelfUpdateCommand extends Command<SelfUpdateArgs, SelfUpdateOpts> {
     let platform = opts.platform
 
     if (!installationDirectory) {
+      log.info(
+        chalk.white(
+          "No installation directory specified via --install-dir option. Garden will be re-installed to the current installation directory: "
+        ) + chalk.cyan(installationDirectory)
```
shumailxyz, line 203: isn't `installationDirectory` undefined here?  🤔

Orzelius (approved):

PR #4574, opened 2023-06-08
0.12 backport chore(deps): update AWS CLI, google cloud CLI and azure CLI to latest versions

### LINKED ISSUE #4571: Bug: Incompatiblity of latest Helm version update in 0.12.57 and the AWS CLI version we ship in 0.12
## Bug

There seems to be an incompatibility with the latest Helm version and the AWS CLI version we ship in the workflow-runner image. (See also https://github.com/helm/helm/issues/10975)

### Current Behavior

When using the 0.12 garden container images starting with 0.12.57 together with AWS  you get the error `Error: Kubernetes cluster unreachable: exec plugin: invalid apiVersion "[client.authentication.k8s.io/v1alpha1](http://client.authentication.k8s.io/v1alpha1)"`

### Expected behavior

No error, it should work

### Reproducible example

We can reproduce this internally

### Workaround

Not really available for users; They can't easily upgrade the AWS CLI to v2 within the container, and they can't choose the helm version in Garden 

### Suggested solution(s)

Either we downgrade helm in 0.12 to 3.8 or we update the AWS CLI. I feel like preferring the latter even though it also has the potential to be breaking for some users. But the list of breaking changes in AWS CLI v1=>v2 is very small and we did not have trouble with it internally at all.

### Your environment

* **OS:** macOS

`garden version` 0.12.57

stefreak:
This backports the new methods of installing the latest versions of AWS CLI, azure CLI and Google Cloud CLI from the main branch (Bonsai).

Fixes #4571 because there is an incompatibility between AWS CLI v1 and the latest helm versions.

### DIFF
--- a/support/alpine.Dockerfile
+++ b/support/alpine.Dockerfile
@@ -1,7 +1,7 @@
 #
-# base
+# garden-base
 #
-FROM node:18-alpine3.17 AS garden-alpine-base
+FROM node:18-alpine@sha256:f41850f74ff16a33daff988e2ea06ef8f5daeb6fb84913c7df09552a98caba09 as garden-alpine-base-root
 
 RUN apk add --no-cache \
   bash \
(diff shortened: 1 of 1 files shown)


stefreak:
@vvagaytsev basics are being tested by [test-dockerhub](https://circleci.com/gh/garden-io/garden/321463) in CircleCI.

But we do not have an integration test with EKS. I think adding that would make a lot of sense, but maybe in another PR.

PR #4575, opened 2023-06-08
Various fixes to action status events

eysi09:

There were several issues around the action status events we send to Cloud. E.g., the types were quite loose and the following would've been a valid event shape:

```typescript
const payload = {
  actionName: "foo",
  actionKind: "build",
  state: "processing",
  completedAt: "2023-06-11T14:39:55.863Z", // <--- It has completed at even though it's processing
  status: {
    state: "built" // <--- Detailed status doesn't match action state
  }
}
```

Other non-sensical payloads like this would also have been valid.  

Also, we now explicitly _pick_ fields to add to the payload type, as opposed to omitting. We should always be explicit about payloads we send "over the wire", prefer picking as opposed to omitting, and thereby ensure it doesn't contain unwanted data. 

For more details, see the full commit message:

```
This commit makes several improvements / fixes to the action status events
sent to Cloud.

A) The types for action status events we emit to Cloud are
narrower. We e.g. ensure that the "completedAt" field is required with
the relevant states, that action state and detailed status are
consistent, add missing fields, and more.

B) Ensure the event emissions are handled in a consistent manner by
introducing a decorator that we call at the task level, not the router
level as before (see also C) below).

C) Ensure superfluous events are not sent when statusOnly=true. This
would previously cause the action graph in Cloud to start spinning when
it shouldn't because we were only checking for statuses, and not
executing anything. To achieve this we now emit events at the task
level, not the action router level.

D) Remove old "taskComplete" et al events which are no longer used.

E) Explicitly _pick_ fields to emit, as opposed to _omitting_ from some
larger interface. The latter is not type safe and can easily cause
unwanted data to be sent "over the wire" as the "parent" interface changes
over time.
```

Fixes #

- Assigning to @thsig for reviewing when you're back online. Not urgent and I can always address any changes you may request in a follow up PR in case we merge before you're back from vacay. 

- Also assigning to @TimBeyer because you're working on a very similar part of the code base and were involved in the original event business around Bonsai.


core/src/tasks/run.ts
```diff
@@ -43,7 +50,14 @@ export class RunTask extends ExecuteActionTask<RunAction, GetRunResult> {
 
       // Should return a null value here if there is no result
       if (status.detail === null) {
```
eysi09, line 52: Not entirely sure what to do here. 

Previously the `getStatus` call would just return `null` if `status.detail === null`.

That doesn't really make sense to me and I fell like we should return a meaningful action state here, even if the detail is null for some reason. 

That being said, it's not clear to be when the `detail` is null so I'm not quite sure what that meaningful action state should be? 

Current options are `ready` | `not-ready` | `cached`. I'm going with `not-ready` here, but again, might be some invariants I'm missing.

eysi09: Also, the solver treats a `not-ready` state and a `null` result the same, so I don't think this breaks anything at that level.

core/src/tasks/test.ts
```diff
@@ -67,20 +68,29 @@ export class TestTask extends ExecuteActionTask<TestAction, GetTestResult> {
...
       }
       return {
         ...status,
-        version: action.versionString(),
-        executedAction: resolvedActionToExecuted(action, { status }),
+        version,
+        executedAction,
+      }
+    } else {
+      return {
```
eysi09, line 84: NOTE: Similar to my comment on the `RunTask` above. 

Previously we returned `null` if the test result was missing or wasn't successful which doesn't sound right. In particular it doesn't account for the scenario where there's a test result but it failed. 

So I'm changing the return value here but let me know if I'm misunderstood something.

eysi09:
One thing I'm noticing is that when we call `garden get status`, the handler for getting the deploy statuses creates tasks and thereby emits events. Whereas the handlers for other actions call the router directly.

Is that intentional @thsig? It kind of makes the behaviour around emitting events inconsistent.

core/src/tasks/run.ts
thsig: > That being said, it's not clear to me when the detail is null so I'm not quite sure what that meaningful action state should be?

I think the field name "detail" is a bit awkward here (I think "result" would be a lot clearer): `status.detail` has the type `RunResult | null` here, with `null` being returned if no result was found.

So I think this is a case of an unfortunate field name making things hard to read.

Your changes totally make sense, though (returning a status with a `not-ready` state instead of just returning `null`).

> Also, the solver treats a `not-ready state` and a `null` result the same, so I don't think this breaks anything at that level.

Yeah, looks good to me too.

thsig (approved):
Good stuff! I didn't test this end-to-end myself, but I assume you've been doing that during development.

Issue #4529, opened 2023-06-04
0.13: [Bug]: All commands fail inside `garden dev` when `terraform` provider enabled


worldofgeese:
### Current Behavior

In the release version of Garden Bonsai, all commands fail inside `garden dev` when the `terraform` provider is enabled. 

### Expected behavior

Commands should run fungibly between the dev console and imperative. 

### Reproducible example

https://github.com/worldofgeese/fastapi-mongodb-garden-example/blob/main/garden.yml

### Workaround

Use the imperative forms of Garden commands.

### Additional context

Interestingly, if the latest `edge-bonsai` release is updated to or installed, Garden will produce a new and different error:

```console
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ Running validate... ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
ℹ providers            → Getting status...


Configured provider 'terraform' has not been registered.
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈

────────────────────────────────────────────────────────── 🌼 🌸 🌷 🌺 🌻  ──────────────────────────────────────────────────────────
```

However, an imperative form of the same command continues to function correctly.

```console
❯ garden validate --force-refresh
Validate 

Garden v0.13 (Bonsai) is a major release with significant changes. Please help us improve it by reporting any issues/bugs here:
https://go.garden.io/report-bonsai
→ Run garden util hide-warning 0.13-bonsai to disable this warning.
[...]

PR #4583, opened 2023-06-09
fix: dev command not working with some plugins

Orzelius:
Dev command wasn't working if a few plugins outside of garden core were initialized (incl. terraform and pulumi).

Fixed by properly setting class plugins and session id params on action call.

closes https://github.com/garden-io/garden/issues/4529

Took me a shameful lot of time to get to the bottom of this but at least I'm more familiar with the `GardenInstanceManager` singleton and `serve` command now.

cc @worldofgeese


vvagaytsev (approved):
Great job! 👍

shumailxyz (approved):
Great work! I can imagine the work behind pinning the bug down to missing props. 👍

PR #4585, opened 2023-06-09
fix(pulumi): remove PULUMI_EXPERIMENTAL flag due to side effects

Walther:

The `PULUMI_EXPERIMENTAL` flag has been originally added in https://github.com/garden-io/garden/commit/9f639397050f1277c93502572ee9c5bcda9ccfb8 as part of implementing the plugin.

However, this can cause side effects, see e.g. https://github.com/pulumi/pulumi/issues/11587 and the slack thread on our #internal-support channel.

Fixes `pulumi up` issues with `violates properties plan`

This needs thorough testing to make sure all of the Pulumi functionality works and to verify that none of the CLI methods we use are no longer considered experimental.

### DIFF
--- a/plugins/pulumi/cli.ts
+++ b/plugins/pulumi/cli.ts
@@ -13,12 +13,7 @@ import { ConfigurationError, RuntimeError } from "@garden-io/sdk/exceptions"
 import { Log, PluginContext, PluginToolSpec } from "@garden-io/sdk/types"
 import { PulumiProvider } from "./config"
 
-/**
- * We're using functionality in the pulumi CLI that's experimental as of February 2022, which is enabled by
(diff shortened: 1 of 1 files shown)


thsig:
At the time, I set the `PULUMI_EXPERIMENTAL` env var to `true` to enable use of the `--plan` option for `pulumi up`.

At a glance, it looks like the `--plan` option is now available for `pulumi up` without setting the env var—if that's the case, then we should be able to stop using this env var.

Orzelius (approved):
lgtm

ishantagarwal-x213235:
I tested and it's working as expected.
Thanks a ton team.

PR #4595, opened 2023-06-09
Update node in examples & some dependencies

shumailxyz:
- Resolves https://github.com/garden-io/garden/security/dependabot/87

- Updated node version in most examples to node 18.
- Manually tested all updated examples and fixed the broken ones:
  - `examples/local-service`
  - `examples/variables`
  - `examples/build-dependencies`
- Removed `request` and `request-promise` dependencies from updated examples as both of those are deprecated and not really needed with availability of `fetch`.

Fixes #

The vote example hasn't been updated as part of this PR because it completely broke with new dependencies and probably would require major tweaking or rewrite of some parts.


Orzelius (approved):
we'll just have to believe you on this one as there are no tests defined for most of these examples

Issue #4580, opened 2023-06-09
0.13: [Bug]: helm chart base not resolved from modules


to266:
### Current Behavior

When a `helm` module has the [`base`](https://docs.garden.io/reference/module-types/helm#base) keyword to reference another module, the templates are not resolved with 0.13, thus deploying a completely empty helm release.

### Expected behavior

Backwards-compatibility with 0.12

### Your environment

* **OS:** any
* **How I'm running Kubernetes:** AKS

`garden version` 0.13.13

Orzelius:
@to266 Thank you for reporting this. Definitely a miss on our part in the module conversion logic. We'll fix it shortly.

to266:
@Orzelius Sorry for taking so long checking this out again - but the PR did not fix this. With 0.13.13 I still get a helm deployment without any manifests

PR #4610, opened 2023-06-13
fix helm and kubernetes module conversion issues

Orzelius:
see commits for details

probably closes [#4580](https://github.com/garden-io/garden/issues/4580)


Orzelius:
This pr probably took 5 years off my lifespan, I'll tell you all about it in the next daily

shumailxyz (approved):
Thanks a lot for this. Looks good to me. 

Should we may be add some tests for this?

PR #4611, opened 2023-06-13
feat(k8s): introduce new flag waitForJobs to wait for k8 jobs

### LINKED ISSUE #4557: [FEATURE]: Kubernetes deploy does should wait for kubernetes jobs
### Current Behavior

Garden deploy does not wait for kubernetes jobs defined in a kubernetes deploy action.

### Expected behavior

I would expect that garden would wait for long running jobs (ideally could would be configurable to wait or not) and also fail the whole deploy if the job fails. 

### Reproducible example

There is a reproducible example for `0.13.1` (`master` branch) and `0.12.57` (`0.12` branch):
https://github.com/pepov/garden-wait

Note: I've used my [asdf plugin](https://github.com/pepov/asdf-garden) to install the correct version of garden

Run `garden deploy` from the repo root on a local kubernetes cluster and check kubernetes pods after garden deploy finished: one of those should be running, the other should have failed.

### Workaround

I'm adding local exec actions to wait for completion, see: https://github.com/pepov/kcp-playground/blob/main/modules.garden.yml#L103-L117

### Suggested solution(s)

According to the garden source code, the way helm implements this should be the example to follow and I agree:
https://github.com/garden-io/garden/blob/main/core/src/plugins/kubernetes/status/status.ts#L70
https://github.com/helm/helm/blob/main/pkg/kube/ready.go#LL225C7-L225C7

### Your environment

* **OS:** Mac OS Ventura 13.4
* **How I'm running Kubernetes:** `minikube start`

```
> garden deploy
Deploy 🚀

ℹ garden               → Running in Garden environment default.default
ℹ cloud-dashboard      → 🌸  Connected to Cloud Dashboard. View logs and command results at:

https://app.garden.io/projects/8b43eb2d-e347-420f-8777-0c563a3c4d82?sessionId=20c8d28e-553f-44a2-b446-9cd6c46906c6&userId=72cd8adc-fd35-4651-9d9e-1a2b6e15479a

✔ providers            → Cached (took 3.6 sec)
ℹ deploy.job-long      → missing
ℹ deploy.job-long      → Deploying version v-ec2bc758d7...
ℹ deploy.job-wrong     → missing
ℹ deploy.job-wrong     → Deploying version v-493f14fe03...
ℹ deploy.job-wrong     → Waiting for resources to be ready...
ℹ deploy.job-long      → Waiting for resources to be ready...
ℹ deploy.job-wrong     → Resources ready
ℹ deploy.job-long      → Resources ready
✔ deploy.job-wrong     → Done (took 2.6 sec)
✔ deploy.job-long      → Done (took 2.6 sec)

Done! ✔️

> kgp
NAME          READY   STATUS              RESTARTS     AGE
long-d9sl8    1/1     Running             0            9s
wrong-77v2l   0/1     RunContainerError   0 (7s ago)   9s
```

shumailxyz:

Fixes #4557 

### DIFF
--- a/core/src/plugins/kubernetes/kubernetes-type/config.ts
+++ b/core/src/plugins/kubernetes/kubernetes-type/config.ts
@@ -85,6 +85,8 @@ export const kubernetesCommonDeploySpecKeys = () => ({
   namespace: namespaceNameSchema(),
   portForwards: portForwardsSchema(),
   timeout: k8sDeploymentTimeoutSchema(),
+  // TODO-0.14: flip this to true and change default behavior to
+  // wait for the jobs
   waitForJobs: joi
     .boolean()
     .optional()
--- a/core/src/plugins/kubernetes/kubernetes-type/config.ts
+++ b/core/src/plugins/kubernetes/kubernetes-type/config.ts
@@ -85,7 +85,11 @@ export const kubernetesCommonDeploySpecKeys = () => ({
   namespace: namespaceNameSchema(),
   portForwards: portForwardsSchema(),
   timeout: k8sDeploymentTimeoutSchema(),
-  waitForJobs: joi.boolean().optional().default(false).description("Wait for batch/v1 jobs to succeed."),
+  waitForJobs: joi
+    .boolean()
+    .optional()
+    .default(false)
+    .description("Wait until the jobs have been completed. Garden will wait for as long as `timeout`."),
 })
 
 export const kubernetesDeploySchema = () =>
--- a/core/src/plugins/kubernetes/kubernetes-type/config.ts
+++ b/core/src/plugins/kubernetes/kubernetes-type/config.ts
@@ -46,6 +46,7 @@ export interface KubernetesDeployActionSpec extends KubernetesTypeCommonDeploySp
   defaultTarget?: KubernetesTargetResourceSpec
   sync?: KubernetesDeploySyncSpec
   localMode?: KubernetesLocalModeSpec
+  waitForJobs?: boolean
 }
 
 export type KubernetesDeployActionConfig = DeployActionConfig<"kubernetes", KubernetesDeployActionSpec>
@@ -84,6 +85,7 @@ export const kubernetesCommonDeploySpecKeys = () => ({
   namespace: namespaceNameSchema(),
   portForwards: portForwardsSchema(),
   timeout: k8sDeploymentTimeoutSchema(),
+  waitForJobs: joi.boolean().optional().default(false).description("Wait for batch/v1 jobs to succeed."),
 })
 
 export const kubernetesDeploySchema = () =>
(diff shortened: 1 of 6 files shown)


edvald:
This change does make sense, but it is technically breaking. I actually think it's still okay given the state of transition to 0.13 across the user base, but we should make sure it's documented as part of 0.13 migration guides and listed as a behavior change.

An alternative would be to add a flag (e.g. `spec.waitForJobs: true`) to control the behavior more explicitly. If we do that, we can make a note to make the breaking change in 0.14, basically flipping the default value of that field.

core/src/plugins/kubernetes/kubernetes-type/config.ts
```diff
@@ -84,6 +85,11 @@ export const kubernetesCommonDeploySpecKeys = () => ({
   namespace: namespaceNameSchema(),
   portForwards: portForwardsSchema(),
   timeout: k8sDeploymentTimeoutSchema(),
+  waitForJobs: joi
```
vvagaytsev, line 88: We can add a todo-comment here like `TODO-0.14: flip this to true`

core/src/plugins/kubernetes/status/status.ts
```diff
@@ -111,6 +113,37 @@ const objHandlers: { [kind: string]: StatusHandler } = {
...
+      resource.spec?.completions &&
+      resource.status?.succeeded &&
+      resource.status?.succeeded < resource.spec.completions
+    ) {
+      // job is not yet completed
+      return { state: "deploying", resource }
+    }
+    // job has succeeded
+    if (resource.status.succeeded) {
+      return { state: "stopped", resource }
```
vvagaytsev, line 136: Should this be `"ready"` instead of `"stopped"`. In [`deployStateMap`](https://github.com/garden-io/garden/blob/main/core/src/plugin/handlers/Deploy/get-status.ts#L29) status `"stopped"` is translated to `"not-ready"`.

vvagaytsev: If this should be changed to `"ready"`, then we might need to change [L145](https://github.com/garden-io/garden/pull/4611/files#diff-24f998c4e8f6dfb59d6ab3922dcc7a3ce8a9cdab363234e51f794537b36611f9R145) to `"unknown"`.

shumailxyz: The thinking behind `stopped` was that it's normal state of pods when job has successfully completed.
But looking at deployStateMap, it seems better to mark these "ready".

Regarding L145, I am not fully sure how "unknown" is handled but we do not want to wait for jobs to complete or timeout if `waitForJobs` isn't set. Also currently, on [L196](https://github.com/garden-io/garden/blob/27012da53e45182102643e7e48880ece915b3e17/core/src/plugins/kubernetes/status/status.ts#L196), state is marked `ready` if there's no explicit handler to check status.

vvagaytsev (approved):
👍

PR #4613, opened 2023-06-13
improvement(server): output more debug info

edvald:
Small improvement to improve working with Desktop.

### DIFF
--- a/core/src/logger/logger.ts
+++ b/core/src/logger/logger.ts
@@ -447,6 +447,7 @@ interface EventLoggerParams extends LoggerConfigBase {
 interface ServerLoggerParams extends LoggerConfigBase {
   defaultOrigin?: string
   rootLogger: Logger
+  terminalLevel?: LogLevel
 }
 
 export interface CreateEventLogParams extends CreateLogParams {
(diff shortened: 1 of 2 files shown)


core/src/server/commands.ts
```diff
@@ -401,7 +403,13 @@ export async function resolveRequest({
     // TODO: Consider using a logger that logs at the silly level but doesn't emit anything.
     serverLogger = new VoidLogger({ level: LogLevel.info })
   } else {
-    serverLogger = command?.getServerLogger() || new ServerLogger({ rootLogger: log.root, level: log.root.level })
+    serverLogger =
+      command?.getServerLogger() ||
+      new ServerLogger({
```
eysi09, line 408: Should we add a comment here or above about the reasoning behind these different levels?

eysi09 (changes requested):
LGTM but might be useful to add a comment explaining the different log props and levels (`commandLogLevel` / `terminalLevel` / `level`).

core/src/server/commands.ts
edvald: I've simplified this after other changes done in the meantime. Pretty trivial now.

eysi09 (approved):

PR #4617, opened 2023-06-13
fix(pulumi): remove PULUMI_EXPERIMENTAL flag due to side effects

TimBeyer:
Backported from #4585 

See #4585 

Fixes #


Walther (approved):
Thanks for the backport!

Issue #4597, opened 2023-06-11
0.13: [Bug]: Graph view on Live page doesn't update correctly


eysi09:
## Garden Bonsai (0.13) Bug

NOTE: Before reading the below, just know that I'm happy to hop on a Discord call to discuss in person. But this took awhile to uncover so I wanted to jot down all the context while it's still fresh in my mind and because some stakeholders are on vacation.

### Current Behavior

The graph view on the Live page doesn't update properly. 

This happens because Garden stops emitting events from a given command that was requested via websocket if another command is also started over websockets before the first one finishes.

I've done some digging around and the problem is pretty deep in the engine. I see what it is, is but I'm not entirely sure how to address it.

I'll list what I know below. 

### Problem

TL;DR: We lookup Garden instances with keys that don't have namespaces but write them with keys that do.   

I suppose there's a bit more to it than the TL;DR suggests. This is the entire flow:

1. **Command A** is requested over websocket. 
2. The instance manager checks whether a valid instance already exists [here](https://github.com/garden-io/garden/blob/master/core/src/server/instance-manager.ts#L134). The namespace is usually not available here so the instance key will be something like `env=dev`. 
3. A Garden instance is created and added to the instance map via `InstanceManager.set()`.
4. In the `set` method we again read the instance key but this time using the `getInstanceKey` method on the newly created Garden instance. See [here](https://github.com/garden-io/garden/blob/master/core/src/server/instance-manager.ts#L217).
5. The Garden instance actually knows the namespace so this time the key will be something like `env=dev-my-namespace`. This is the key we use when writing to the map.
6. We continue processing the command.
7. **Command B** is requested over websocket before **Command A** is done.
8. As in step 2 above, the instance manager checks whether a valid instance already exists. Again it uses the `env=dev` key whereas the instance for **Command A** which is indeed valid is stored via the `env=dev-my-namespace` key. 
9. Because it doesn't find the a valid instance, it again creates a new instance like in step 3 above. 
10. In the `set` method however, it now finds a valid, existing Garden instance because the instance key used there contains the namespace, like in step 4 above. 
11. [This code](https://github.com/garden-io/garden/blob/main/core/src/server/instance-manager.ts#L255-L263) then  unsubscribes the listener for **Command A** and re-subscribes it to the Garden instance for **Command B** which doesn't use the same `EventBus`. 
[...]

PR #4621, opened 2023-06-13
fix(server): inconsistent key set for Garden instances for concurrent requests

edvald:
cc @eysi09, can you take a look to see if this partially addresses #4597?

### DIFF
--- a/core/src/garden.ts
+++ b/core/src/garden.ts
@@ -1506,11 +1506,7 @@ export class Garden {
   public getInstanceKeyParams() {
     let namespace: string | undefined
 
-    let defaultNs = this.getEnvironmentConfig().defaultNamespace
-
-    if (defaultNs === undefined) {
-      defaultNs = defaultNamespace
-    }
+    const defaultNs = this.getEnvironmentConfig().defaultNamespace || defaultNamespace
 
     if (this.namespace !== defaultNs) {
       namespace = this.namespace


eysi09:
I'm still seeing the same behaviour. 

[Here](https://github.com/garden-io/garden/blob/master/core/src/server/instance-manager.ts#L131) the key `env=dev` (for my example project). 

There will never be an instance with this key because when we add it to the map it'll have the key `env=dev-my-namespace`. 

So we always hit the codepath that creates a new Garden instance. That code path also unsubscribes the listener from the previous instance so we drop any further events from that instance. 

Basically what's described here: #4597.

Note also that `defaultNs` resolves to `${var.dev-env-name}` on my project (see my comment on the code).

core/src/garden.ts
```diff
@@ -1506,11 +1506,7 @@ export class Garden {
   public getInstanceKeyParams() {
     let namespace: string | undefined
 
-    let defaultNs = this.getEnvironmentConfig().defaultNamespace
-
-    if (defaultNs === undefined) {
-      defaultNs = defaultNamespace
-    }
+    const defaultNs = this.getEnvironmentConfig().defaultNamespace || defaultNamespace
```
eysi09, line 1509: The template string isn't resolved here so I get `defaultNs=${var.dev-env-name}`.

The returned namespace is still `vote-demo-eysi` because of the line below:

``` 
    if (this.namespace !== defaultNs) {
      namespace = this.namespace
    }
```

eysi09 (changes requested):

eysi09:
...what I don't really understand is why we first read the key from the instance manager params and then again in the `set` from the Garden instance. 

I would've assumed we'd just re-use the original key we get from params and use that in the `set` method.

eysi09:
Some further findings:

When we create the initial DummyGarden for the parent dev command the `gardenKey` we [set on the log context](https://github.com/garden-io/garden/blob/master/core/src/garden.ts#L1855) has a namespace (default) so it becomes `env=default-default`. 

I think that conflicts with this code comment: `// Setting this after resolving the gardenKey above because we don't want the default namespace resolved there`

When we then create a "proper" Garden instance via the instance manager the namespace is missing so the log context gets `gardenKey=env=default`.

edvald:
Closed in favor of #4649

PR #4623, opened 2023-06-14
Fix pixel quantity

ShankyJS:

Fixes #


ShankyJS:
@worldofgeese Gitbook renders images with a different CSS rule; that's why when we set height with the  tag it doesn't work. To mitigate the issue, I uploaded a different picture with 54px of height (To be the same as Cloud shell) and I can see it's rendering fine.

image

PR #4628, opened 2023-06-14
fix(cloud): emit ns statuses for in-cluster builds

### LINKED ISSUE #4609: [Regression]: Garden doesn't send a namespaceStatus event when running build
### Current Behavior

We used to send a `namespaceStatus` event when running the build command in Acorn. These events are used by Cloud to power other features (e.g. AEC), therefore breaking them in certain scenarios.

### Expected behavior

When I run `garden build` against a remote cluster I want the Garden CLI to send a `namespaceStatus` event containing the name of the Kubernetes namespace used to build my images.

### Reproducible example

Ping for reproducing this. 

### Workaround

There is currently not a simple workaround if using AEC, aside implementing the cleanup logic manually, which kinda defeats the purpose of the feature itself.

### Your environment

* **OS:** 

`garden version`

thsig:

We now emit a `namespaceStatus` event whenever one of our internal helpers for checking the status of a Kubernetes namespace is called. This is done whenever we e.g. perform a Deploy, Test or Run in that namespace, or when we perform an in-cluster Build in that namespace (e.g. via BuildKit or Kaniko).

Here, we also remove namespace statuses from the relevant result types (e.g. `ServiceStatus`, `RunStatus` and `EnvironmentStatus` and the associated schemas). These added unwanted noise to the interfaces and schemas.

Instead, we emit the `namespaceStatus` events on the plugin event broker (`ctx.events`) and re-emit them to `garden.events` in the base router.

Previously, automatic environment cleanup would not get triggered for namespaces where only Builds had taken place (i.e. no Deploys, Runs or Tests). This fixes that.

Fixes #4609

I wrote a test case verifying that `namespaceStatus` events get emitted during BuildKit builds, but it kept hanging due to something Mutagen-related. I've verified manually that the events do get emitted (and so long as we're using the standard helpers for getting the Kubernetes namespace status in the BuildKit and Kaniko handlers, we should be good).

We should investigate this further, but I didn't want to get bogged down in debugging a test case that seemed like it might also introduce flakiness into our integ test suite.


vvagaytsev:
@thsig this fixes #4609, right?

thsig:
It does, yeah—forgot to mention that in the description.

core/src/plugins/kubernetes/init.ts
```diff
@@ -187,9 +187,9 @@ export async function prepareEnvironment(
 
   // Prepare system services
   await prepareSystem({ ...params, clusterInit: false })
-  const ns = await getAppNamespaceStatus(k8sCtx, log, k8sCtx.provider)
-
-  return { status: { namespaceStatuses: [ns], ready: true, outputs: status.outputs } }
+  // This call is only to emit a `namespaceStatus` event.
+  await getAppNamespaceStatus(k8sCtx, log, k8sCtx.provider)
```
eysi09, line 191: Wouldn't it be better to just emit the status directly here? 

We could easily refactor or remove this function and that would either break this or produce unwanted side effects.

eysi09 (changes requested):
Had one small comment. 

Otherwise I feel like we need to re-think this in general but don't want to block the fix. 

But shouldn't this be handled in a consistent manner at the action or router level with events that include plugin specific information so that it extends beyond just K8s?

AEC would then be responsible for applying the clean up for different plugins (or using the CLI to do it).

core/src/plugins/kubernetes/init.ts
thsig: Yeah, that's a good point. Will make that change.

thsig:
> But shouldn't this be handled in a consistent manner at the action or router level ...

In this PR, we're catching `namespaceStatus` events from the plugin event broker in one place in the base router. We could do it for each individual router method, but that seems unnecessarily repetitive.

> ... with events that include plugin specific information so that it extends beyond just K8s?

That totally makes sense—we can always change/extend the `namespaceStatus` events to have a `detail` field, for example.

In general, I understand your reservations about this PR. That said, my thinking was that the `namespaceStatuses` field on action results felt bolted-on and not a good pattern, and that this was another "letting Cloud know about what's going on" event (like the old `log` events). So I felt it was less hacky this way.

Also, both before and after this PR, we're relying on integ tests to ensure that we don't accidentally stop sending the events when certain plugin handlers are called. There's currently no framework-level guarantee that any namespace statuses were generated by the plugin handler.

I do think that going forward, it would be nice to have a plugin-level mechanism to register event types for the router to listen for.

eysi09:
Oh I don’t have reservations about the PR per se. Rather this flow in general which needs to be revisited separately imo so we can extend AEC to more plugins.

eysi09 (approved):

core/src/plugins/kubernetes/init.ts
```diff
@@ -187,9 +187,9 @@ export async function prepareEnvironment(
 
   // Prepare system services
   await prepareSystem({ ...params, clusterInit: false })
-  const ns = await getAppNamespaceStatus(k8sCtx, log, k8sCtx.provider)
-
-  return { status: { namespaceStatuses: [ns], ready: true, outputs: status.outputs } }
+  const nsStatus = await getAppNamespaceStatus(k8sCtx, log, k8sCtx.provider)
+  ctx.events.emit("namespaceStatus", nsStatus)
```
vvagaytsev, line 191: It looks like the function `getAppNamesStatus` just delegates to `getNamespaceStatus` without any extra logic. The latter, in turn, also sends the same event. Wouldn't we have duplicate events emitted?

core/src/router/run.ts
```diff
@@ -66,7 +66,6 @@ export const runRouter = (baseParams: BaseRouterParams) =>
         await router.validateActionOutputs(action, "runtime", result.outputs)
 
         // TODO: get this out of the core framework and shift it to the provider
```
vvagaytsev, line 68: Dangling TODO comment here :)

core/src/router/test.ts
```diff
@@ -66,7 +66,6 @@ export const testRouter = (baseParams: BaseRouterParams) =>
         await router.validateActionOutputs(action, "runtime", result.outputs)
 
         // TODO: get this out of the core framework and shift it to the provider
```
vvagaytsev, line 68: [Same](https://github.com/garden-io/garden/pull/4628/files#r1230520095) here

thsig:
Should be good to go now, @vvagaytsev.

vvagaytsev (approved):
Thank you! I've left one more [comment here](https://github.com/garden-io/garden/pull/4628/files#r1230753446). If the duplicate events make no harm, then all is good!

PR #4630, opened 2023-06-14
chore: update `js-yaml` to 4.1.0

### LINKED ISSUE #4633: Update js-yaml and replace all uses of safeLoad functions
In this PR https://github.com/garden-io/garden/pull/4620 the yarnlock was edited and the resolutions for `js-yaml` changed. This caused e.g. Pulumi tests to start failing: https://app.circleci.com/pipelines/github/garden-io/garden/18346/workflows/f87d43b5-a41d-4e29-8650-7e8169da8567/jobs/324282

We should make sure we update the `js-yaml` version across our entire monorepo and remove uses of all `safe` functions, as the new ones are safe by default.

vvagaytsev:

Fixes #4633


yarn.lock
```diff
@@ -6752,7 +6747,7 @@ js-yaml@4.1.0, js-yaml@^4.1.0:
   dependencies:
     argparse "^2.0.1"
 
-js-yaml@^3.10.0, js-yaml@^3.13.1, js-yaml@^3.14.1:
+js-yaml@^3.10.0, js-yaml@^3.13.1:
```
Walther, line 6750: sidenote: wonder why we are still getting `js-yaml` at version(s) `3.x` here 🤔

Walther (approved):
Thanks for the fixes and cleanup ✨ 🧹

plugins/pulumi/.mocharc.yml
```diff
@@ -1,6 +1,6 @@
 reporter: spec
 spec: test/*.js
-timeout: 60000
+timeout: 180000
```
shumailxyz, line 3: just wondering if we do need to increase timeout by 3x here? is this somewhat related to js-yaml?

shumailxyz (approved):
Looks great. Thank you so much. Let's get the CI green 🚀

plugins/pulumi/.mocharc.yml
Walther: this was already bumped from 30s to 60s earlier, i have no strong feelings one way or another here
(although worst case, 180s _per test_ can accumulate to quite a long job total in case of errors that lead to timeouts instead of fast faiures)

vvagaytsev: I got timeouts on `beforeAll` hooks in [test-plugins](https://app.circleci.com/pipelines/github/garden-io/garden/18359/workflows/84dddce4-4ea9-47a2-9cd5-9a3a3c945944/jobs/324739). That hook installs node packages, so I think 1 minute was not enough to set everything up. I used the same timeout as we use for jib modules.

shumailxyz: yeah make sense in case of packages installation.

Issue #4602, opened 2023-06-12
docs.garden.io bug: weird menu behaviour


stefreak:
### Current Behavior

<img width="377" alt="Screenshot 2023-06-12 at 15 25 19" src="https://github.com/garden-io/garden/assets/139624/a062b3ff-afa6-4aae-89a1-98d27323f561">

Nothing happens when you click on "container build" or "configmap deploy" (the first subpage).

Another thing that's slightly strange is that the url of the exec deploy ends with `/container/exec` even though it's unrelated to container, and the same is true for the other categories, e.g. url for container deploy is `/configmap/container` for some reason.

### Expected behavior

I expect to be able to click and successfully navigate to the docs of the first action type of every kind.

### Reproducible example

https://docs.garden.io

### Workaround

Click on the action kind itself, e.g. "Build" or "Deploy" to see the first type.

### Suggested solution(s)

IDK- is this gitbook or us?

### Your environment

* **OS:** macOS, Chrome

PR #4631, opened 2023-06-14
docs: fix action reference docs

Orzelius:
action kind docs were missing readmes which are used for gitbook and
table of content generation

closes https://github.com/garden-io/garden/issues/4602


Walther (approved):
Thanks for the fix ✨ 

_(Would be nice if we could some day have previews for docs PRs to double check that everything will work - now we just have to trust it works and create another PR later in case there's an issue)_

eysi09:
> Thanks for the fix ✨
> 
> _(Would be nice if we could some day have previews for docs PRs to double check that everything will work - now we just have to trust it works and create another PR later in case there's an issue)_

We should. It's a Gitbook feature. 

Perhaps @Orzelius can look into that since you have access.

eysi09:
Oh wait, here it is. Does the link work for you?

Screenshot 2023-06-14 at 18 44 35

eysi09:
Ok gotcha, let’s look into that

PR #4632, opened 2023-06-14
fix: do not mark deploy action outdated unnecessarily

### LINKED ISSUE #4590: Bug: Garden does not print Garden version in verbose log output
### Current Behavior

Garden does not print garden version in verbose logs

### Expected behavior

I expect it to give me useful information to find the issue, so Garden version would be highly useful

### Reproducible example

`garden test -l3`

### Workaround

Never forget to send garden version together with bug reports
`garden --version`

### Suggested solution(s)

Print a verbose log message at startup, with the garden version

### Your environment

`garden version` 0.13.2

### LINKED ISSUE #4618: 0.13: [Bug]: deploy action invalidated when files change that are not included in the build action
## Garden Bonsai (0.13) Bug
When I make changes to files that are not included in the build action, I expect no changes when running garden deploy.

### Current Behavior

Currently, when touching files that are not included in the deploy's build action, the deploy action will be considered 'outdated' and redeployed, even though nothing has changed:

```
ℹ deploy.test          → outdated
ℹ build.test           → Already built
ℹ deploy.test          → Deploying version v-9b6a2216a5...
ℹ deploy.test          → Waiting for resources to be ready...
ℹ deploy.test          → Resources ready
✔ deploy.test          → Done (took 2.7 sec)
```

### Expected behavior

I expect no changes when I touch files that are not included in the build action:
```
ℹ deploy.test          → Already deployed
```

### Reproducible example

`garden.yml`
```
kind: Project
apiVersion: garden.io/v1
name: repro
environments:
 - name: default
providers:
 - name: local-kubernetes
---
kind: Build
type: container
name: test
include:
 - Dockerfile
---
kind: Deploy
type: container
name: test
build: test
```

`Dockerfile`
```
FROM nginx
```

1. `garden deploy` (deploy the service for the first time)
2. `touch someFile` (Change a file that is not `include`d in the build action)
3. `garden deploy` (Deploy should not redeploy, because no files included in the build action changed)

### Workaround

Add `include: []` to the deploy action. This will successfully work around, but I feel that this should be unnecessary because the build action should count here.

### Suggested solution(s)

I don't know where to start with the solution.

### Your environment

* **OS:** macOS

`garden version` 0.13.2

shumailxyz:
Sets `include` of Test/Run/Deploy action to `[]` when `build` is specified and `include` or `exclude` is not set on the config.
~~Compute and update the tree version only for a module or Build action or if any files are `included` explicitly.~~

Previously, adding a new unrelated file to directory or some other change marked the `Deploy` as outdated and deployed unnecessarily. 

This was happening due to the change in tree version which resulted in a change of action version that is included in the manifest. Due to change in the version, comparison of the hashes of new manifest with deployed manifest resulted in different value and state was marked as `outdated`.

Fixes #4618 
Fixes #4590  (was missing version in verbose logging myself so added that as well)

Took me quite some time in digging this down but at least now I am more aware of the vcs handler and comparison checks.


core/src/vcs/vcs.ts
```diff
@@ -225,8 +225,12 @@ export abstract class VcsHandler {
           // Don't include the config file in the file list
           .filter((f) => !configPath || f.path !== configPath)
 
-        result.contentHash = hashStrings(files.map((f) => f.hash))
         result.files = files.map((f) => f.path)
+        if (config.kind === "Build" || config.kind === "Module" || (config.include && config.include.length > 0)) {
+          // only update tree hash in case of module or a build action or
+          // files include is set.
+          result.contentHash = hashStrings(files.map((f) => f.hash))
+        }
```
stefreak, line 233: I'm not 100% sure how to review this because I am very unfamiliar with how this is supposed to work. 
But I feel like this needs a regression test. The minimal example I provided in https://github.com/garden-io/garden/issues/4618 could be used for that

stefreak: One problem with this approach might be that for Deploy actions without `build`, e.g. `kubernetes` type deploys, we do need to take the included files into account as it may contain changed manifests. 
For the `kubernetes` deploy kind we only should include the files that are specified in the spec.

shumailxyz: I used the example you provided as reference and wrote a few other garden files to test it.

Regarding your point about kubernetes deploys, those will be included as this the condition already considers explicitly included files into account.

vvagaytsev (dismissed):
Lgtm! 👍

core/src/vcs/vcs.ts
Orzelius: I agree with Steffen, this should be tested. You can add a test to the [already existing suite](https://github.com/garden-io/garden/blob/main/core/test/unit/src/vcs/vcs.ts#L97) for this.

stefreak: > Regarding your point about kubernetes deploys, those will be included as this the condition already considers explicitly included files into account.
Sounds reasonable– I don't know how this works, but maybe add a test case for a kubernetes or helm deploy that doesn't explicitly set `include`

core/test/unit/src/vcs/vcs.ts
```diff
@@ -95,6 +95,16 @@ describe("VcsHandler", () => {
   })
 
   describe("getTreeVersion", () => {
+    const convertObjToArgs = (args: {
```
Orzelius, line 98: I was thinking a change in the call signature of the original `getTreeVersion` function itself. I think it's okay to have plain params if there's not many of them but at soon as they grow long or there's a bool involved I prefer a [destructured object](https://mariusschulz.com/blog/typing-destructured-object-parameters-in-typescript#typing-immediately-destructured-parameters)

image

Orzelius: @shumailxyz please don't rewrite git history in the review process, it tends to be tricky to see what has changed. Utilize a fixup commit instead. You can squash the fixups into the original commits before the merge if you want to rebase merge or you can do a squash merge.

shumailxyz: I did start with refactoring the call signature of original function but then switched to adding helper for tests as it was being used at some other places too. I do think that 4 args are not that many though however will still update it considering the bool.

Orzelius (approved):
well done! figuring out the issue origin must have taught a bit about garden's internal structures

core/src/cli/cli.ts
```diff
@@ -444,6 +444,7 @@ ${renderCommands(commands)}
     })
 
     const log = logger.createLog()
+    log.verbose(`garden version: ${getPackageVersion()}`)
```
edvald, line 447: This seems a little odd to me to print on every verbose execution. I'd either make this a debug level or reconsider what exactly is printed and where.

edvald (changes requested):
So, I don't think this is the correct solution. Happy to help out with this btw but wanted to at least flag before this is merged.

I can see this create an issue with e.g. an `exec` `Test` action where the version should be based on files related to the action. That will now _only_ work if users explicitly set the `include` field. So, many existing setups will break with this PR as-is, since users are not required to do that. Many similar scenarios would likely have the same issue.

When `build` is specified on Deploy/Run/Test actions, it could make sense to set the `include` field to `[]` unless either of `include` or `exclude` is explicitly set on the config already. This should be checked _after_ the `configure` plugin handler is called, since those handlers often infer or override include/excludes based on the action spec. I in fact thought this was already the case, perhaps it was just discussed at some point and not implemented. At least I can't find it in the action resolution code.

Lastly, I don't think it makes sense to first scan for the files contained in the action and then to conditionally detach the `contentHash` from the file list. In my view, the list of files should always deterministically map to a `contentHash`, and it makes more sense to affect the `include/exclude`. It's more consistent and also avoids unnecessary I/O.

edvald:
That all said, thank you for diving into this @shumailxyz! You're pretty deep in the machine there, amongst the dragons :) And not at all far from the right approach either, this is just one of the more delicate parts of the codebase and quite a bit of nuance to deal with.

shumailxyz:
@edvald Thank you for the feedback. That's a very valid  point and I agree that affecting include/exclude is a better approach. Will update it.

core/src/cli/cli.ts
shumailxyz: This is coming from #4590 and I think it makes sense to print garden version at the start of the execution. There's a point to move it to debug level however I am more inclined towards keeping it at verbose level so if someone runs garden with `-l3` and copies log, the garden version is also copied.

edvald (approved):

PR #4643, opened 2023-06-15
chore: retry pod creation three times

Orzelius:
It seems that one 500ms delay retry [wasn't enough](https://app.circleci.com/pipelines/github/garden-io/garden/18394/workflows/04f7f878-c4b5-4da9-9964-c0e5b32c8725/jobs/325915?invite=true#step-110-535)

### DIFF
--- a/core/src/plugins/kubernetes/api.ts
+++ b/core/src/plugins/kubernetes/api.ts
@@ -831,7 +831,7 @@ export class KubeApi {
    *
    * @throws {KubernetesError}
    */
-  async createPod(namespace: string, pod: KubernetesPod, isRetry = false) {
+  async createPod(namespace: string, pod: KubernetesPod) {
     await pRetry(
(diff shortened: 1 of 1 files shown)


core/src/plugins/kubernetes/api.ts
```diff
@@ -832,25 +832,27 @@ export class KubeApi {
    * @throws {KubernetesError}
    */
   async createPod(namespace: string, pod: KubernetesPod, isRetry = false) {
```
Walther, line 834: can we remove the `isRetry` parameter here, now that everything is wrapped in the retry?

Walther (approved):
Looks good!

---

(Sidenote: at some point in the future, we could potentially adjust other methods to use the same retry mechanisms as well to unify things - currently we have quite a bunch of retry implementations in this file)

PR #4646, opened 2023-06-15
feat(template): allow `this.name` and `this.mode` in action config

### LINKED ISSUE #4616: 0.13: [Bug]: `this.mode` not available in `build` option of `Deploy` action
## Garden Bonsai (0.13) Bug

When choosing the build action for a deploy action, the template context does not include `this` so I cannot differentiate between modes (e.g. sync mode)

I want to use a different build action when in sync mode because in sync mode I need dev dependencies in the Dockerfile, which I do not need for the production container.

### Current Behavior

When using `build: "${this.mode == 'sync' && 'foo' || 'bar' }"` 
```
Error processing config for Deploy action limiter:
Invalid template string (${this.mode == 'sync'}): Could not find key this. Available keys: command, datetime, environment, git, inputs, local, parent, project, secrets, template, var and variables.
```

### Expected behavior

I expect it to successfully choose the correct build action depending on the value of `this.mode`.

### Reproducible example

See this pull request to the quickstart-example: https://github.com/garden-io/quickstart-example/pull/9/files

```
# Production container to be used without sync mode.
kind: Build
type: container
name: limiter-prod
include:
  - src
  - Dockerfile
exclude:
  - src/tmp/*
spec:
  dockerfile: "Dockerfile"

---

# The sync container only provides the tools necessary for building the library
# when using it during inner loop development together with Garden's sync mode.
kind: Build
type: container
name: limiter-sync
include:
  - Dockerfile
spec:
  dockerfile: "Dockerfile.syncmode"

---

kind: Deploy
type: container
name: limiter
build:
  $if: "${this.mode == 'sync'}"
  $then: limiter-sync
  $else: limiter-prod
spec:
  sync:
    command: [air]
    paths:
      - target: /app
        source: ./src
        mode: one-way-replica
        exclude: ["tmp/*/**"]
  ports:
    - name: http
      containerPort: 8080
      servicePort: 80
  ingresses:
    - path: /
      port: http
      hostname: limiter-${var.base-hostname}
  env:
    PROXY_URL: http://api

```

### Workaround

As a workaround it is possible to use 
```
  $if: ${this.mode == 'sync' && (command.params.sync contains 'limiter' || isEmpty(command.params.sync))}
```
Where `limiter` is the name of the deploy action. It is not 100% reliable as you can also use wildcards in the sync params, and you can start / stop syncs using `garden sync`  commands.

### Suggested solution(s)

Make the `this` context available in `build`.

### Additional context
Once this is fixed, also update the code of the `limiter` service in https://github.com/garden-io/quickstart-example

### Your environment

* **OS:** macOS

`garden version` 0.13.2

vvagaytsev:

Fixes #4616

~Still WIP, I need to add some unit tests~
We do not test each individual template string path. Context scanning and template string resolution are covered in https://github.com/garden-io/garden/blob/main/core/test/unit/src/template-string.ts.

### DIFF
--- a/plugins/pulumi/commands.ts
+++ b/plugins/pulumi/commands.ts
@@ -40,6 +40,7 @@ import { TemplatableConfigContext } from "@garden-io/core/build/src/config/templ
 import { ActionTaskProcessParams, ValidResultType } from "@garden-io/core/build/src/tasks/base"
 import { deletePulumiDeploy } from "./handlers"
 import { ActionLog, createActionLog, Log } from "@garden-io/core/build/src/logger/log-entry"
+import { ActionConfigContext } from "@garden-io/core/build/src/config/template-contexts/actions"
 
 type PulumiBaseParams = Omit<PulumiParams, "action">
 
@@ -171,7 +172,10 @@ const pulumiCommandSpecs: PulumiCommandSpec[] = [
 
 const makePluginContextForDeploy = async (params: PulumiParams & { garden: Garden; graph: ConfigGraph }) => {
   const { garden, provider, ctx, action } = params
-  const templateContext = new TemplatableConfigContext(garden, action.getConfig())
+  const templateContext = new ActionConfigContext(garden, action.getConfig(), {
+    name: action.name,
+    mode: action.mode(),
+  })
   const ctxForDeploy = await garden.getPluginContext({ provider, templateContext, events: ctx.events })
   return ctxForDeploy
 }
--- a/plugins/pulumi/commands.ts
+++ b/plugins/pulumi/commands.ts
@@ -36,7 +36,7 @@ import { BooleanParameter, parsePluginCommandArgs } from "@garden-io/sdk/util/cl
 import { copy, emptyDir } from "fs-extra"
 import { join } from "path"
 import { isDeployAction } from "@garden-io/core/build/src/actions/deploy"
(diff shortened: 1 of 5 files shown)


plugins/pulumi/commands.ts
```diff
@@ -171,7 +171,7 @@ const pulumiCommandSpecs: PulumiCommandSpec[] = [
 
 const makePluginContextForDeploy = async (params: PulumiParams & { garden: Garden; graph: ConfigGraph }) => {
   const { garden, provider, ctx, action } = params
-  const templateContext = new ActionConfigContext(garden, action.getConfig())
+  const templateContext = new TemplatableConfigContext(garden, action.getConfig())
```
edvald, line 174: Hmm, I think we should keep ActionConfigContext here. The mode and name should both be available at this point.

edvald (changes requested):
Had one comment, but otherwise looks solid, good work 👍

plugins/pulumi/commands.ts
vvagaytsev: Thanks! Fixed that in https://github.com/garden-io/garden/pull/4646/commits/d62d10f05ec7f3c9c39d81878fc4c4b41a1b11d2

edvald (approved):

PR #4650, opened 2023-06-16
fix custom commands in dev mode

### LINKED ISSUE #4647: custom commands broken in dev console
### Current Behavior

Seeing two issues with custom commands in the dev console:

First, when you first enter the dev console, trying to run a custom command results in the following error:
Screenshot 2023-06-15 at 4 29 57 PM
`Could not find command. Try typing help to see the available commands.`

If you run something else built-in like `help` or `deploy` and then try the custom command, it will now work.

The second issue, however, is that some custom commands fail like this: 
Screenshot 2023-06-15 at 4 36 03 PM
`Missing cli argument in custom command wrapper.
This is a bug. Please report it!`

Both times I've seen this were on commands that use `$concat: ${args.$rest}` to allow args to be supplied to the custom command, but should work without any args.

### Your environment

* **OS:** macOS
` % garden version
0.13.2`

shumailxyz:
- Fixes custom commands not working on first run in dev mode.
- Also uses command artifact for `version` command so it is supported in dev mode and shows up in help.

Fixes #4647 

The second issue with custom gardenCommands reported in #4647 will be address separately. Tracking that with https://github.com/garden-io/garden/issues/4668

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -407,10 +407,6 @@ ${renderCommands(commands)}
       return { argv, code: abortCode, errors, result, consoleOutput }
     }
 
-    if (argv.v || argv.version || argv._[0] === "version") {
-      return done(0, getPackageVersion())
-    }
-
     const workingDir = resolve(cwd || process.cwd(), argv.root || "")
 
     if (!(await pathExists(workingDir))) {
--- a/core/src/commands/version.ts
+++ b/core/src/commands/version.ts
@@ -1,0 +1,30 @@
+/*
+ * Copyright (C) 2018-2023 Garden Technologies, Inc. <info@garden.io>
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { getPackageVersion } from "../util/util"
+import { Command, CommandResult } from "./base"
+
+interface VersionCommandResult {
+  version: string
+}
+
+export class VersionCommand extends Command {
+  name = "version"
+  aliases = ["v", "V"]
+  help = "Shows the current garden version."
+  noProject = true
+
+  async action({ log }): Promise<CommandResult<VersionCommandResult>> {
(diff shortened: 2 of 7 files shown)


core/src/commands/version.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { getPackageVersion } from "../util/util"
+import { Command } from "./base"
+
+export class VersionCommand extends Command {
```
edvald, line 12: You need to add `noProject = true` on this class. This should fix the broken builds/tests.

edvald (commented):
One comment, rest looks solid. Good catch on the `projectRoots` thing, that's a subtle one.

core/src/cli/cli.ts
```diff
@@ -407,9 +407,9 @@ ${renderCommands(commands)}
       return { argv, code: abortCode, errors, result, consoleOutput }
     }
 
-    if (argv.v || argv.version || argv._[0] === "version") {
-      return done(0, getPackageVersion())
-    }
+    // if (argv.v || argv.version || argv._[0] === "version") {
+    //   return done(0, getPackageVersion())
+    // }
```
vvagaytsev, line 412: Let's delete this if we don't need it.

shumailxyz: yeah. it'll all be updated together.

vvagaytsev:
A minor nit. It looks like commits https://github.com/garden-io/garden/pull/4650/commits/bd0e6ce326852045564973828f7a30c60182a2ea and https://github.com/garden-io/garden/pull/4650/commits/91172439fb05c418ebb1241357e12ee16cbe9243 can be squashed.

I'd also consider it as an improvement instead of a bug fix:)

shumailxyz:
@vvagaytsev Thanks. Updated it to reflect improvement and combined both commits into one. Had to force push though.

vvagaytsev (approved):
Great job, thank you! 🚀

PR #4652, opened 2023-06-16
feat(openshift): container Deploy action

### LINKED ISSUE #4624: [FEATURE]: experimental local-openshift container Deploy
Part of the OpenShift epic https://github.com/garden-io/garden/issues/4243

---

Definition of done: users are able to use a Deploy action to deploy an already-built container image, pulled from some registry.

Tasks:

- Create a document listing all the required setup steps and required permissions on OpenShift side
- Create workarounds for the differences between Kubernetes namespaces and OpenShift projects
- Create workarounds / documentation for the requirement of running containers with arbitrary user ids
- Create integration tests using OpenShift Local

Walther:

Fixes #4624 

There is a minimal example project in `core/test/data/openshift/demo-project`.
The README in there contains the required setup instructions for your OpenShift Local installation.
After the setup, you should be able to run `garden deploy`, `garden logs`, and `garden cleanup deploy` (with some caveats).

I have added integration tests and they can be run locally after you have done the setup above.
We should figure out a CI story for these at some point - but let's keep it as a separate PR, as setting up OpenShift in CI might take some time.


core/src/plugins/kubernetes/api.ts
```diff
@@ -464,6 +464,13 @@ export class KubeApi {
           // Then this resource version + kind is not available in the cluster.
           return []
         }
+        // FIXME: OpenShift: developers have more restrictions on what they can list
+        // Ugly workaround right now, basically just shoving the problem under the rug.
+        const openShiftForbiddenList = ["Namespace", "PersistentVolume"]
+        if (err.statusCode === 403 && openShiftForbiddenList.includes(kind)) {
+          log.warn(`No permissions to list resources of kind ${kind}. If you are using OpenShift, ignore this warning.`)
+          return []
+        }
```
Walther, line 473: This has a theoretical chance to show up on users who are _not_ using OpenShift.
However, if they don't have permissions to list these kinds, they would be having issues regardless

core/test/data/openshift/demo-project/README.md
```diff
@@ -0,0 +1,38 @@
...
+socat TCP6-LISTEN:6443,fork,reuseaddr TCP4:127.0.0.1:6443
+```
+
+## Deploy
+
+Ideally, at this point this should work:
+
+```bash
+garden deploy
+open http://hello.local.demo.garden/ # NOTE: this will return 403 as there is no content in the served directory and directory listing is forbidden
```
Walther, line 35: This one should be improved when I add support for the `Build` action in the next PR, and we can have actual content served there.

core/src/plugins/openshift/container.ts
```diff
@@ -0,0 +1,280 @@
...
+) => {
+  const { ctx, action, log } = params
+  // TODO: separate openshift types for these if possible to pass around?
+  // const openshiftCtx = <OpenShiftPluginContext>ctx
+  // const provider = openshiftCtx.provider
+  // const api = await KubeApi.factory(log, ctx, provider)
+
+  const k8sCtx = <KubernetesPluginContext>ctx
+  const provider = k8sCtx.provider
+  const api = await KubeApi.factory(log, ctx, provider)
```
Walther, line 42: As it currently stands, this file has quite a ridiculous amount of copy-paste from the k8s equivalent functions with small tweaks. This file also currently uses some of the k8s types instead of OpenShift specific ones.

My suggestion: let's keep moving forward here to achieve a minimum level of functionality first, and refactor later. Having to repeat things first is a good way to find where the abstractions should be.

core/src/plugins/kubernetes/namespace.ts
```diff
@@ -66,6 +66,8 @@ export async function ensureNamespace(
     const cache = nsCache.get(providerUid) || {}
 
     if (!cache[namespace.name] || namespaceNeedsUpdate(cache[namespace.name].resource!, namespace)) {
+      // FIXME: if passed `namespace: string` this will set cache[undefined]
+      // This seems to require `namespace: { name: foo }` in project config contrary to docs
```
Walther, line 70: This might be worth a separate bugfix, but leaving it out of this PR.

core/test/integ/src/plugins/openshift/tests.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  it("should delete container deploy", async () => {
+    const command = new DeleteEnvironmentCommand()
+    const { result } = await command.action({
+      garden,
+      log,
+      args: {},
+      opts: withDefaultGlobalOpts({ "dependants-first": false }),
+    })
+    expect(result!.deployStatuses["nginx-hello"].state === "ready")
```
shumailxyz, line 82: why this check passes for state = `ready` after deletion?

core/test/integ/src/plugins/openshift/tests.ts
```diff
@@ -0,0 +1,84 @@
...
+        "follow": false,
+        "tag": undefined,
+        "show-tags": false,
+        "timestamps": false,
+        "hide-name": false,
+      }),
+    })
+    // the openshift nginx container image does not produce logs correctly,
+    // but this should ensure the logs command at least runs successfully
+    expect(result).to.deep.eq([])
```
shumailxyz, line 71: interesting. May be we could use some other image and actually test availability of logs?

shumailxyz (commented):
Good stuff. Thank you for digging into this 💯  

i'll have to actually try it out locally and may update afterwards. Leaving some comments for now.

core/test/data/openshift/demo-project/garden.yml
```diff
@@ -6,3 +6,8 @@ environments:
 providers:
   - name: local-openshift
     environments: [local]
+    # NOTE: shorthand `namespace: demo-project` fails because we use `namespace.name` in code
+    namespace:
+      name: demo-project
```
shumailxyz, line 11: instead of NS, should we rather add a field for `project` ?

core/test/integ/src/plugins/openshift/tests.ts
Walther: Absolutely! I was planning to improve this test in the next PR when adding the ability to use a `Build` step - then we can have our own images and make sure to log something properly there.

core/test/data/openshift/demo-project/garden.yml
Walther: Yeah, that's a good question.

We rely on `config.namespace` (and `config.namespace.name`) in a bunch of places, and it might take some refactoring. I'm slightly leaning towards keeping it as-is right now and potentially improving it later - after all, this is mostly a _cosmetic_ difference as it stands, and the functionality should work

core/src/plugins/kubernetes/api.ts
Walther: We should do this later - currently the `OpenShiftPluginContext` is not actually used, we hack it with `KubernetesPluginContext` (and provider)

core/src/plugins/openshift/container.ts
```diff
@@ -0,0 +1,279 @@
...
+}
+
+/**
+ * Returns project/namespace status (which includes the namespace's name).
+ * Based on the kubernetes plugin method `getNamespaceStatus`.
+ * However, this method never attempts to create a missing namespace.
+ * If the expected namespace does not exist, throws an error.
+ *
+ */
+export async function expectNamespaceStatus({
```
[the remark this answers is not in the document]
Walther: yeah - at some point I had this named with `expectProjectStatus` and other variants, naming is hard 😅 
even though openshift uses project terminology, it is kubernetes api compatible and does use namespaces too. this function ends up calling `api.core.readNamespace()` under the hood in the end anyway

core/src/plugins/openshift/container.ts
```diff
@@ -0,0 +1,279 @@
...
+
+export const openshiftContainerDeployExtension = (): DeployActionExtension<ContainerDeployAction> => ({
+  name: "container",
+  handlers: {
+    deploy: openshiftContainerDeploy,
+    delete: openshiftDeleteContainerDeploy,
+    // exec: execInContainer,
+    getLogs: openshiftGetContainerDeployLogs,
+    getPortForward: async (params) => {
+      return getPortForwardHandler({ ...params, namespace: undefined })
```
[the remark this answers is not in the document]
Walther: Good catch! I'll comment this out (or remove it). This skeleton was just copy-pasted from here https://github.com/garden-io/garden/blob/2c0072adfcc2697aa515912e3d9f4032787a3f55/core/src/plugins/kubernetes/container/extensions.ts#L72-L93

I have not verified the port forwarding functionality or added custom handlers for openshift yet.

vvagaytsev (approved):
LGTM! 👍

PR #4654, opened 2023-06-18
improvement(sdk): more iteration on the plugin SDK + migrate exec plugin

edvald:
This gets us closer to fully supporting all plugin features with the new SDK.

The one major thing left to implement is action type extensions. The only plugin that uses that atm is the `kubernetes` plugin, which will take more work to migrate so we'll likely leave it for last.

Note that I'm leaving module type definitions+extensions out of the new SDK
for now, because we're deprecating them and it's not really worth it to
migrate those handlers.

This makes more progress on #3778


thsig (approved):
✨

PR #4660, opened 2023-06-19
fix: update version hash if a file is renamed

### LINKED ISSUE #4661: 0.13: [Bug]: Garden version is not updated when file is renamed
### Current Behavior
Garden version is not updated when file is renamed and Garden doesn't rebuild.

### Expected behavior
Garden should update version when file is renamed.

### Additional context

See slack thread on our #internal-support channel

shumailxyz:
Consider relative file path along with the file hash for version hash calculation. Previously, version was not updated if file is renamed without any change in its content.

Considers relative file path in content hash to cater for path changes (e.g. renaming).
Fixes #4661

This will probably mark all existing caches invalid due to different versions.

### DIFF
--- a/core/src/vcs/vcs.ts
+++ b/core/src/vcs/vcs.ts
@@ -230,7 +230,8 @@ export abstract class VcsHandler {
           // Don't include the config file in the file list
           .filter((f) => !configPath || f.path !== configPath)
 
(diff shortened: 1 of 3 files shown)


vvagaytsev (approved):
Great job, thanks! 🚀 
Let's back-port it to `0.12`.

PR #4664, opened 2023-06-20
feat: open telemetry

TimBeyer:
Adds OTEL tracing support to get deep insights into the build performance.

Fixes #


core/src/commands/dev.tsx
```diff
@@ -122,9 +123,9 @@ Use ${chalk.bold("up/down")} arrow keys to scroll through your command history.
         },
       })
 
-      useInput((input, key) => {
+      useInput(bindActiveContext((input, key) => {
```
eysi09, line 126: Not sure I follow here. Why is this needed?

core/src/commands/dev.tsx
```diff
@@ -231,7 +232,7 @@ Use ${chalk.bold("up/down")} arrow keys to scroll through your command history.
         .finally(() => quit())
     }
 
-    function quit() {
+    async function quit() {
```
eysi09, line 235: Why did we change this? We're not awaiting on anything in the function body.

core/src/index.ts
```diff
@@ -6,4 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+import { initTracing } from "./util/tracing/tracing"
```
eysi09, line 9: Why do we need this in multiple places?

core/src/util/tracing/tracing.ts
```diff
@@ -0,0 +1,62 @@
...
+    return otelSDK
+  }
+}
+
+export function initTracing(): opentelemetry.NodeSDK {
+  if (otelSDK) {
+    return otelSDK
+  }
+
+  if (!gardenEnv.GARDEN_ENABLE_TRACING) {
```
eysi09, line 36: Not super familiar with this flow, but shouldn't we just exit here as opposed to overwriting the env and continuing the function execution.

core/src/garden.ts
```diff
@@ -1108,6 +1119,9 @@ export class Garden {
     })
   }
 
+  @OtelTraced({
```
eysi09, line 1122: Some of this is quite granular.

I think the end user will mostly care about:

- Overall command execution
- Config resolution
- Individual actions

That being said, this could be very useful for our own internal debugging. But wondering if we should discern between that and data that users are interested in.

core/src/plugins/exec/deploy.ts
```diff
@@ -353,6 +354,7 @@ function runPersistent({
     env: {
       ...getDefaultEnvVars(action),
       ...(env ? mapValues(env, (v) => v + "") : {}),
+      ...getTracePropagationEnvVars(),
```
eysi09, line 357: Is this needed here? 

If so, are there other places we could be missing?

eysi09 (commented):
Truly great work! 

I mostly had questions since I'm not super familiar with OTel SDK. 

I would also very much appreciate getting @edvald's eyes on this, since it touches on some critical parts of the code, in particular around the initialisation.

core/src/commands/dev.tsx
TimBeyer: The SDK tracks the current context using [AsyncLocalStorage](https://nodejs.org/api/async_context.html#class-asynclocalstorage).

Since those callbacks get executed at a random time in the future from an event emitter, they lose the automatic binding that would be in place normally. So for that reason, event handlers need to be bound explicitly to keep the context being propagated into the code that gets triggered by the event handler.

core/src/commands/dev.tsx
TimBeyer: That is a leftover of figuring out why the tracing in the dev command didn't work.
It's not needed anymore.

core/src/index.ts
TimBeyer: I added this here so anything that when `garden` is loaded, we immediately install all the hooks.  
It might be redundant here but since the other initialization happened only in `/cli` I wanted to ensure that there's at least also one entry point in `/core` itself.

core/src/util/tracing/tracing.ts
TimBeyer: We're still going to call OTEL SDK functions even if we don't initialize the SDK, the behavior of which we're not entirely sure about if we don't init the SDK.

Thus we went with `OTEL_SDK_DISABLED` which ensures that all operations are no-ops on the OTEL SDK level.

core/src/garden.ts
TimBeyer: I think we can build upon this and maybe have different detail levels.  
I would do that as a follow up though.

core/src/plugins/exec/deploy.ts
TimBeyer: This ensures that in the - admittedly very specific - case of someone calling a `garden` command from within the `exec` command, but not using `gardenCommand` we still trace things through.

Also in theory any other piece of software that we spawn there and that follows the environment variable spec will be able to pick up the traces.

I didn't find any other obvious spots to add the propagation vars, but I may have missed some.
I looked for all `execa` calls to see if they do something that might require the headers.

eysi09:
Thanks for the response! All of that make sense to me. 

I'd still very much like to get @edvald take before we merge. In particular since we're planning on a release on Monday. 

If Jón is super busy, I'm happy to merge on Monday after the release and just dog food this some.

eysi09 (approved):

cli/src/cli.ts
```diff
@@ -35,11 +38,19 @@ export async function runCli({
...
 
   try {
-    if (!cli) {
-      cli = new GardenCli({ plugins: getBundledPlugins(), initLogger })
-    }
-    // Note: We slice off the binary/script name from argv.
-    result = await cli.run({ args, exitOnError })
+    // initialize the tracing to capture the full cli execution
+    result = await withContextFromEnv(() =>
+      wrapActiveSpan("garden", async (span) => {
```
edvald, line 43: Not a major thing, but stylistically I'd tend to prefer wrapping these functions in an object/namespace, e.g. `tracing.withContextFromEnv` and `tracing.wrapActiveSpan`. That's what I've been doing with the new plugin SDK for example, and it makes it a bit more obvious what these functions are about when reading code.

core/src/util/tracing/spans.ts
```diff
@@ -0,0 +1,50 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import * as opentelemetry from "@opentelemetry/sdk-node"
+import { tracer } from "./tracing"
+import { getSessionContext } from "./context"
+import { prefixWithGardenNamespace } from "./util"
+
+export function startActiveSpan<T>(name: string, fn: (span: opentelemetry.api.Span) => Promise<T>): Promise<T> {
```
edvald, line 14: It would be really good to get comments on each of these public functions. I for example am a bit unclear on the difference between them and how/where exactly to use each.

cli/src/cli.ts
TimBeyer: In that case we'd have to import and re-export all the different functions from the different submodules since we have for example `withContextFromEnv` in one module and `wrapActiveSpan` in another.

I feel like then it would be preferable if we used an `index.ts` that groups all the things together so they lie under `util/tracing` directly. Then in there we have to group those all into an object since we can't really enforce the `import * as tracing from "tracing"` syntax we could use if we simply just re-export everything.

To me that seems to introduce different redundant ways of importing those modules which can't really be enforced, or discourage modularization in order to prevent that redundancy.

I'm not sure that's worth it. Maybe what would be better is to use clear naming everywhere so it's always obvious it's `withTracingContextFromEnv` or something like that.

core/src/util/tracing/spans.ts
TimBeyer: I've documented all the functions

PR #4666, opened 2023-06-20
fix(sync): fix status bug for `kubernetes` deploys

### LINKED ISSUE #4511: 0.13: [Bug]: sync status not correct
### Current Behavior

The new sync status command doesn't seem to be working as expected. When running it in a project with a whole bunch of syncs configured, after starting the syncs, they all show as "not-active". But if I run `garden plugins kubernetes sync-status` they all show as `status: watching`

### Your environment

* **OS:** macOS

`% garden version`
`0.13.0`

thsig:

Before this fix, the `sync status` command would always show the status of active syncs for `kubernetes` Deploys as `not-active`.

This was because we were passing deployed resources for `container` and `helm` deploys when starting syncs and getting sync statuses, but passing local manifests for `kubernetes` deploys.

Fixes #4511


vvagaytsev (approved):
👍

PR #4673, opened 2023-06-20
docs(readme): update readme with config examples

eysi09:

Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,59 +54,91 @@ Click a button to start your Killercoda or Google Cloud Shell environment 👇�
 
 If you find any bugs 🐛 or have suggestions to improve our labs please don't hesitate to reach out by creating an [issue here](https://github.com/garden-io/garden-interactive-environments) or by asking in our [Discord Community](https://go.garden.io/discord)🌸
 
-## Usage
+## Using Garden
 
-> Make sure you have Garden installed and Kubernetes running locally (e.g. with Minikube or Docker for Desktop) before deploying the project.
+### Configuration
 
-If you have a `garden.yml` file in your project, you can run `garden` commands from the root of your project. If you don't have a `garden.yml` file, clone the quickstart project:
+A Garden "project" is configured via `garden.yml` configuration files. In them, you describe the parts of your system in terms of the following **four actions** and their dependencies on one another:
 
-```sh
-git clone https://github.com/garden-io/garden-quickstart.git
+- **Build actions**
+- **Deploy actions**
+- **Test actions**
+- **Run actions** (for running ad-hoc tasks)
+
+Here's a simple example for a three trier web application:
+
+```yaml
+# This config is in a single file for convenience.
+# You can split it into multiple files and even across repositories!
+kind: Deploy
+name: db
+type: helm
+spec:
+  chart:
+    repo: https://charts.bitnami.com/bitnami
+---
+kind: Run
+name: db-init
+type: container
+dependencies: [deploy.db]
+---
+kind: Deploy
+name: api
+type: kubernetes
+build: api
+dependencies:
+  - run.db-init
+spec:
+  files: [api/manifests]
+---
+kind: Deploy
+name: web
+type: kubernetes
+build: api
+dependencies:
+  - deploy.api
+spec:
+  files: [web/manifests]
+---
+kind: Test
+name: e2e
+type: kubernetes-exec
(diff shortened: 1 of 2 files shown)


worldofgeese:
I have reservations about inlining big blocks of configuration into the README: the intention of having folks clone the quickstart and running `garden dev` is to give people a rapid launch. The problem of complexity is a big one we face and we really just have an elevator pitch of time to get people inside and able before they nope out and move on. Users already attracted to the proposition will visit the documentation for anything more complex than the quickstart.

We do highlight requirements at the top of Usage for getting going.

eysi09:
Yeah I hear you. 

> the intention of having folks clone the quickstart and running garden dev is to give people a rapid launch

Thing is that doesn't work. It's going to fail if you follow the steps in the README as opposed to the actual steps from the docs which include installing Garden, installing local K8s, etc. 

So that's not a good experience imo. 

> really just have an elevator pitch of time to get people inside and able before they nope out and move on

I guess I'm arguing that this piece of YAML is our elevator pitch. Obviously I won't make it far in business, pitching YAML. 

Jokes aside though, a lot of people say that their "aha moment" with Garden is when they see the config and realise how it's put together.

Would it make more sense to only show a smaller piece (we're already skipping builds)?

---

Also happy to skip the YAML altogether, but I do think we need to fix the current "Usage" section so as not to mislead people.

eysi09:
The YAML could also be under an expandable. 

Or we have a short sneak preview and then you can expand for the whole enchilada.

eysi09:
Hey, wanted to quickly revisit this. Any thoughts on my suggestions above, @worldofgeese?

worldofgeese:
@eysi09 

> Thing is that doesn't work. It's going to fail if you follow the steps in the README as opposed to the actual steps from the docs which include installing Garden, installing local K8s, etc.

Ok so we do have 

> Make sure you have Garden installed and Kubernetes running locally (e.g. with Minikube or Docker for Desktop) before deploying the project.

then we ask them to git clone the quickstart and go. So to me this feels complete.

> The YAML could also be under an expandable.

This is a good compromise, I think. Continuing with the quickstart example, we could "macro-expand" each part of the commands, where under the expandable we isolate the section that's being acted on and explain how the config matches the imperative i.e.

> **Note**
> Click "Details" beneath each command for a cross-section decomposition of Garden's YAML configuration 

`build`
<details>
The `build` command executes these `Build` blocks across all your `garden.yml` files in the project. Explanation continues... 
```yaml
kind: Build
type: container
name: api-build
description: The backend build for the voting UI
```
</preview>

And we'd have these expandables under each imperative, like this. What do you think?

eysi09:
Thanks for the reply! 

> So to me this feels complete.

Maybe this is just me and my limited attention span, but I would not noticed the greyed out text at the top mentioning the requirements. In fact I'd probably clone and run the command and then realise I wasn't actually installing Garden. 

I guess my problem is that it _reads_ like a getting started guide (clone something then run commands) but it is actually a usage guide for those that already have installed Garden and set up local Kubernetes. The number of people who do all that and then come to the GitHub README for the next steps is probably very low. 

> Continuing with the quickstart example, we could "macro-expand" each part of the command

Yeah, that could be nice. In that context, I'm wondering whether we skip the build command since it's usually implicit with the deploy. Otherwise we find ourselves explaining that you actually don't really need the build etc etc.

worldofgeese:
> Maybe this is just me and my limited attention span, but I would not noticed the greyed out text at the top mentioning the requirements. In fact I'd probably clone and run the command and then realise I wasn't actually installing Garden.

Ok, that's no problem! GitHub Flavored Markdown ships with support for warnings:

> **Warning**
> Make sure you have Garden installed and Kubernetes running locally (e.g. with Minikube or Docker for Desktop) before deploying the project.

> I guess my problem is that it reads like a getting started guide (clone something then run commands) but it is actually a usage guide for those that already have installed Garden and set up local Kubernetes. The number of people who do all that and then come to the GitHub README for the next steps is probably very low.

The first step of the README is to curl install Garden so I think we have that covered. In the warning, we could link to that headline should users need a reminder.

Once we ship support for ephemeral clusters then the only thing we'll expect of them is Garden installed, which is covered in that first step.

worldofgeese:
Alternatively, we could ship support for the expandables now and make certain fields more prominent to assist in shepherding user attention.

PR #4675, opened 2023-06-20
fix(cloud): properly handle dev delegation

thsig:

When a command is run with an option that results in it being run inside the `dev` command instead (e.g. `garden deploy --sync`), we now register the initial command session with Cloud as a `dev` command with the `--cmd` option set appropriately. This is useful for Cloud when listing command sessions created this way.

For good measure, we include an explicit `isDevCommand` parameter with the session registration request to Cloud to make this setup more robust against future changes (instead of inferring whether the command is a `dev` command from the `commandInfo` map).

I've verified that including the `isDevCommand` param in the registration request doesn't result in errors (it will be ignored by the API until implemented there). So there shouldn't be any compatibility problems in the meantime.

### DIFF
--- a/core/src/cloud/api.ts
+++ b/core/src/cloud/api.ts
@@ -586,6 +586,7 @@ export class CloudApi {
     localServerPort,
     environment,
     namespace,
+    isDevCommand,
   }: {
     parentSessionId: string | undefined
     sessionId: string
@@ -594,6 +595,7 @@ export class CloudApi {
     localServerPort?: number
     environment: string
     namespace: string
+    isDevCommand?: boolean
   }): Promise<CloudSession | undefined> {
     let session = this.registeredSessions.get(sessionId)
 
@@ -610,6 +612,7 @@ export class CloudApi {
         projectUid: projectId,
         environment,
         namespace,
+        isDevCommand,
       }
       this.log.debug(`Registering session with ${this.distroName} for ${projectId} in ${environment}/${namespace}.`)
       const res: CloudSessionResponse = await this.post("sessions", {
--- a/core/src/cloud/api.ts
+++ b/core/src/cloud/api.ts
@@ -595,7 +595,7 @@ export class CloudApi {
     localServerPort?: number
     environment: string
     namespace: string
-    isDevCommand?: boolean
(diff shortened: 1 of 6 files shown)


core/src/cloud/api.ts
```diff
@@ -594,6 +595,7 @@ export class CloudApi {
     localServerPort?: number
     environment: string
     namespace: string
+    isDevCommand?: boolean
```
vvagaytsev, line 598: Should we make this parameter non-optional? Sometimes it's easy to forget to pass an optional parameter, so having it passed explicitly might be a benefit for bug safety :)

Now we call this function only in 2 places, so converting it to the required one won't touch that many places in the code :)

core/src/commands/base.ts
```diff
@@ -284,6 +286,17 @@ export abstract class Command<A extends Parameters = {}, O extends Parameters =
...
+    // going into a subcommand in the `GardenCli` class where the `commandInfo` object is first created.
+    if (!["dev", "serve"].includes(this.name) && this.maybePersistent(params) && !params.parentCommand) {
+      // Then this command will be starting a `dev` command and then running itself from there, so we update
+      // the `commandInfo` accordingly.
+      // Example: `garden deploy --sync`.
+      const outerName = this.name
+      garden.commandInfo.name = "dev"
+      garden.commandInfo.opts.cmd = getCmdOptionForDev(outerName, params)
+    }
+
```
vvagaytsev, line 299: This can be extracted to a function, and the heading comment will become its jsdoc :)

vvagaytsev (approved):
@thsig thanks for fixing this! I've left a few non-blocking comments.

core/src/commands/base.ts
thsig: IMHO it's fine as is—we're only doing this once at this central location (in the base command).

core/src/cloud/api.ts
thsig: I guess we could, but I thought this seemed like a less commonly used opt-in param that's only used in very specific circumstances.

thsig: Yeah, good point—I've made that change now.

PR #4676, opened 2023-06-21
feat(cli): add `get files` command to see files included in actions

edvald:
Came up in a customer conversation. Quite basic but also handy.

### DIFF
--- a/core/src/commands/get/get-files.ts
+++ b/core/src/commands/get/get-files.ts
@@ -1,0 +1,72 @@
+/*
+ * Copyright (C) 2018-2023 Garden Technologies, Inc. <info@garden.io>
+ *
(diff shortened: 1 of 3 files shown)


core/src/commands/get/get-files.ts
```diff
@@ -0,0 +1,72 @@
...
+
+type Args = typeof getFilesArgs
+type Opts = {}
+
+interface Result {
+  [key: string]: string[]
+}
+
+/**
+ * An abstract base class for get actions subcommands
```
shumailxyz, line 31: This comment seems off and can be removed.

shumailxyz (approved):
one minor nit. rest LGTM.

PR #4680, opened 2023-06-21
feat(server): add internal/experimental cmd for getting action statuses

eysi09:

This is an internal/experimental cmd that'll only be used by the web dashboard for now.

Fixes #

### DIFF
--- a/core/src/actions/helpers.ts
+++ b/core/src/actions/helpers.ts
@@ -6,8 +6,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+import Bluebird from "bluebird"
 import chalk from "chalk"
-import { memoize } from "lodash"
+import { fromPairs, memoize } from "lodash"
 import { joi } from "../config/common"
 import { Garden } from "../garden"
 import { createActionLog, Log } from "../logger/log-entry"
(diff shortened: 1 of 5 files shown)


core/src/server/commands.ts
```diff
@@ -250,7 +256,42 @@ export class _GetDeployStatusCommand extends ConsoleCommand {
...
+  streamEvents = false
+
+  outputsSchema = () => joi.object()
+
+  async action({ garden, log }: CommandParams): Promise<CommandResult<GetActionStatusesCommandResult>> {
+
+    const router = await garden.getActionRouter()
+    const graph = await garden.getResolvedConfigGraph({ log, emit: true })
+
+    const actions = await Bluebird.props({
```
thsig, line 286: Maybe `statuses` would be a better name here?

core/src/actions/helpers.ts
```diff
@@ -140,3 +146,90 @@ export function getRelativeActionConfigPath(projectRoot: string, action: Action)
   const relPath = relative(projectRoot, action.configPath() ?? "")
   return relPath.startsWith("..") ? relPath : "." + sep + relPath
 }
+
+export async function getDeployStatuses(router: ActionRouter, graph: ResolvedConfigGraph, log: Log) {
```
thsig, line 150: How about renaming these to `getDeployStatusPayloads`/`getBuildStatusPayloads` etc. to more clearly distinguish this from the other `getDeployStatuses` function?

thsig (approved):
Looks good! Had a couple of non-blocking comments there, I'll leave it up to you whether to accept/address them or not.

core/src/server/commands.ts
eysi09: It's an action map though that we're returning. And it jives with the result interface.

Issue #4683, opened 2023-06-21
[FEATURE]: Add link shortener for command results for the web dashboard


eysi09:
### Background / Motivation

There are currently _two_ cases where we want to print links to the Cloud / web dashboard in the console.

Case 1: When running the `garden dev` command we want to print a shortened link to the live page. We already have this (although the link is printed in the wrong place, I'm working on a fix).

Case 2: When running most other Garden commands we want to print a shortened link to the command results page. E.g. if you run `garden build`. Similarly if you run `build` from within the dev command. See examples below.  

When running Garden dev we want links for the live page _and_ the results of the individual sub commands (build in this case) 👇 

```
──────────────────────────────────────────────────────────────────── 🌳  garden dev 🌳  ────────────────────────────────────────────────────────────────────

Good afternoon! Welcome to the Garden interactive development console.

Here, you can build, deploy, test and run anything in your project, start code syncing, stream live logs and more.

Use the command line below to enter Garden commands. Type help to get a full list of commands.
Use up/down arrow keys to scroll through your command history.
    
🌸 Connected to Garden Dashboard 🌸 
Follow the link below to stream logs, run commands, and more from your web dashboard 👇  

https://app.garden.io/go/<some-id> // <--- This is good!


┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ Running build... ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
ℹ garden-dashboard    → View command results at: https://app.garden.io/go/commands/<some-id> // <--- This is missing, should link to command results 

ℹ providers            → Getting status...
ℹ providers            → Preparing environment...
ℹ exec                 → Configuring...
ℹ exec                 → Running init script
ℹ exec                 → Fetching cluster endpoint and auth data.
ℹ exec                 → kubeconfig entry generated ...
```

When running non-dev commands we want to print a link to the command results page 👇 

```
❯ garden build
Build 🔨 

[...]

PR #4685, opened 2023-06-21
fix(cli): print correct link to web dashboard

eysi09:

From commit message:

Before this fix we were printing links to the live page for command
results. This fixes that but in turn we need to deal with the
unshortened link until we add link shortening support for command
results page.

See also: https://github.com/garden-io/garden/issues/4683

Fixes #

### DIFF
--- a/core/src/cloud/api.ts
+++ b/core/src/cloud/api.ts
@@ -447,8 +447,7 @@ export class CloudApi {
       this.log.debug({ msg: `Failed to refresh the token.` })
       const detail = is401Error(err) ? { statusCode: err.response.statusCode } : {}
       throw new CloudApiTokenRefreshError(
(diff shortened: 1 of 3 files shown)


core/src/cloud/api.ts
```diff
@@ -447,8 +447,7 @@ export class CloudApi {
       this.log.debug({ msg: `Failed to refresh the token.` })
       const detail = is401Error(err) ? { statusCode: err.response.statusCode } : {}
       throw new CloudApiTokenRefreshError(
-        `An error occurred while verifying client auth token with ${getCloudDistributionName(this.domain)}: ${
-          err.message
+        `An error occurred while verifying client auth token with ${getCloudDistributionName(this.domain)}: ${err.message
```
eysi09, line 450: This is just auto formatting, looks like we need to fix that so it's consistent for all of us.

thsig (approved):
LGTM

PR #4686, opened 2023-06-21
fix(k8s) Update default Kaniko version

mattpolzin:
**What this PR does / why we need it:**
This PR updates the default version of Kaniko used by Garden. Of course, anyone can use a non-default version of Kaniko regardless, but newer versions of K8s cause older versions of Kaniko to error out with the mistaken message that "Kaniko must be run inside a container" when, of course, it is running inside a container. Newer versions of Kaniko should work for older versions of K8s, but I can confirm they also fix the aforementioned bug.

**Special notes for your reviewer:**

I have only tested this change against my current deploy's K8s version (`1.26.3`). It would seem reasonable to want to validate the newer Kaniko version works against select other versions of K8s.


vvagaytsev:
@mattpolzin, huge thanks for your contribution! 👍 
The [check-docs](https://app.circleci.com/pipelines/github/garden-io/garden/18540/workflows/0bac3123-5f4a-4b38-a776-0372711bcf2c/jobs/331481) CI step failed. That happened because the generate docs weren't up-to-date.

Could you re-generate the docs, please? In order to do that, just run `yarn build` locally and commit the changes. Thank you!

mattpolzin:
@vvagaytsev Done. Let me know if I can do anything else!

shumailxyz (approved):
Thank you so much 👍

vvagaytsev (approved):

PR #4689, opened 2023-06-21
fix(k8s): invalid session name error with e.g. underscore in username/path

edvald:
This gets rid of our custom sanitization function and also trims the source path to a relative path to make the key cleaner+shorter.


vvagaytsev (approved):
👍

PR #4690, opened 2023-06-21
fix(dev): running command status message could become inconsistent

edvald:
Basically commands could in certain instances get stuck, after some failure modes etc.


shumailxyz (approved):
Many thanks for this 👍

PR #4692, opened 2023-06-21
fix(template-strings): do not apply helper functions on unresolved string

### LINKED ISSUE #4614: Template string variables unresolved in some circumstances
## Bug

When setting up a kubernetes module, in some circumstances variables in template strings are not resolved unless the expression somehow mentions another context to resolve variables from before the string to be templated.

### Current Behavior

```
# This sets it to the string "${environment.namespace}"
TEST: "base64Encode(\"${environment.namespace}\")"
```

```
# This sets it to the string "${environment.namespace}"
TEST: "${true == true ? base64Encode(\"${environment.namespace}\") : false }"
```

```
# This resolves the template correctly
TEST: "${environment && base64Encode(\"${environment.namespace}\")}"
```

### Expected behavior

The variable should always be resolved and replaced

### Reproducible example

```
apiVersion: garden.io/v0
kind: Project
name: bug-repro
environments:
  - name: local
providers:
  - name: local-kubernetes
    environments: [local]
    namespace: ${project.name}-testing-${var.userId}
variables:
  userId: ${local.username}

---

kind: Module
description: Secret
type: kubernetes
name: secret
manifests:
  - apiVersion: v1
    kind: Secret
    data:
      # DOES NOT WORK
      TEST_A: "${base64Encode(\"test-${environment.namespace}-data\")}"
      TEST_B: "${true == true && base64Encode(\"test-${environment.namespace}-data\")}"
      # WORKS
      TEST_C: "${environment && base64Encode(\"test-${environment.namespace}-data\")}"
      TEST_D: "${var.userId && base64Encode(\"test-${environment.namespace}-data\")}"
    metadata:
      name: credentials
```

### Workaround

Using `"${environment && base64Encode(\"${environment.namespace}\")}"`

### Your environment

* **OS:** macOS 13.4
* **How I'm running Kubernetes:** docker

```
garden version
0.12.58
0.13.1
```

shumailxyz:
Before this fix, helper functions were being applied even on resolved string and the unsubstituted string was considered the string itself for helper function. 

Basically, in the [first resolveTemplateStrings of module](https://github.com/garden-io/garden/blob/main/core/src/resolve-module.ts#L358), when only `inputs` are available and keys like `environment` are missing, unresolved strings are fine as `allowPartial` is set to `true`. However, the helper functions were still being applied.

For example, for a template string like  
```
TEST:  "base64Encode(\"${environment.namespace}\")"`
```
base64Encode was being called for `environment.namespace` instead of the actual value and resulted in wrong values. So before the [2nd resolveTemplateStrings of module](https://github.com/garden-io/garden/blob/main/core/src/resolve-module.ts#L363), when the module context is available, there was no `base64Encode(\"${environment.namespace}\")` left to be resolved.

The workaround mentioned in original issue #4614 
`TEST: "${environment && base64Encode(\"${environment.namespace}\")}"` 
works because of the `environment &&`. Even though the `base64Encode` is applied on `environment.namespace`, the overall expression fails due to `environment` missing in first pass.

I added a check before helper functions are applied to not apply helper on unresolved string. Had to wrap `text` in `${}` so it's returned as an expression to be resolved later. In case, `allowPartial` is set to `false`, it returns error. 

Fixes #4614 

I am not sure if we can avoid this check altogether and instead extend the actual parser to address this issue. 🤔

### DIFF
--- a/core/src/template-string/functions.ts
+++ b/core/src/template-string/functions.ts
@@ -16,6 +16,7 @@ import { validateSchema } from "../config/validation"
 import { load, loadAll } from "js-yaml"
 import { safeDumpYaml } from "../util/serialization"
 import indentString from "indent-string"
+import { maybeTemplateString } from "./template-string"
 
 interface ExampleArgument {
   input: any[]
@@ -508,6 +509,19 @@ export function callHelperFunction({
         context: `argument '${argName}' for ${functionName} helper function`,
         ErrorClass: TemplateStringError,
       })
+
+      // do not apply helper function for an unresolved template string
+      if (maybeTemplateString(value)) {
+        const _error = new TemplateStringError(`Function '${functionName}' cannot be applied on unresolved string`, {
+          functionName,
+          text,
+        })
+        if (allowPartial) {
+          return { resolved: "${" + text + "}" }
+        } else {
+          return { _error }
+        }
+      }
     } catch (_error) {
       if (allowPartial) {
         return { resolved: text }
--- a/core/src/template-string/functions.ts
+++ b/core/src/template-string/functions.ts
@@ -512,13 +512,13 @@ export function callHelperFunction({
 
       // do not apply helper function for an unresolved template string
       if (maybeTemplateString(value)) {
-        const _error = new TemplateStringError(`Function '${functionName}' cannot be applied on unresolved string`, {
-          functionName,
-          text,
-        })
         if (allowPartial) {
           return { resolved: "${" + text + "}" }
         } else {
+          const _error = new TemplateStringError(`Function '${functionName}' cannot be applied on unresolved string`, {
+            functionName,
+            text,
+          })
           return { _error }
         }
       }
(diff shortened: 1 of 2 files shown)


core/src/template-string/functions.ts
```diff
@@ -508,6 +509,19 @@ export function callHelperFunction({
         context: `argument '${argName}' for ${functionName} helper function`,
         ErrorClass: TemplateStringError,
       })
+
+      // do not apply helper function for an unresolved template string
+      if (maybeTemplateString(value)) {
+        const _error = new TemplateStringError(`Function '${functionName}' cannot be applied on unresolved string`, {
```
vvagaytsev, line 515: Nit. The error can be created inside the `else`-block below.

vvagaytsev:
LGTM, but I'd like @edvald to take a look too. Can we add some test cases? The repro examples from the original issue can be used as a base.

edvald:
Fix looks great (and very well caught @shumailxyz!) but agree that a test case would be helpful.

core/src/template-string/functions.ts
shumailxyz: moved below.

vvagaytsev (approved):

PR #4695, opened 2023-06-22
improvement(jib): support concurrent builds in maven

vvagaytsev:

Fixes #


shumailxyz (approved):
Looks good to me. Thank you so much. 💯 

on a side note, i would suggest adding some tests as well for this.

shumailxyz:
Indeed that can be tricky. Let's add those separately and merge this.

PR #4699, opened 2023-06-23
chore(analytics): error stack trace metadata error handling and wrapped errors in analytics

mkhq:

Updates error handling in getStackTraceMetadata and introduces support for wrappedErrors.

Fixes #

### DIFF
--- a/core/src/exceptions.ts
+++ b/core/src/exceptions.ts
@@ -19,6 +19,10 @@ export interface GardenError<D extends object = any> extends Error {
   wrappedErrors?: GardenError[]
 }
 
+export function isGardenError(err: any): err is GardenError {
+  return "type" in err && "message" in err
+}
+
 export type StackTraceMetadata = {
   functionName: string
   relativeFileName?: string
--- a/core/src/exceptions.ts
+++ b/core/src/exceptions.ts
@@ -20,13 +20,13 @@ export interface GardenError<D extends object = any> extends Error {
 
 export interface GardenErrorStackTrace {
   functionName: string
-  relativeFileName: string
+  relativeFileName?: string
 }
 
 export abstract class GardenBaseError<D extends object = any> extends Error implements GardenError<D> {
   abstract type: string
 
-  constructor(message: string, public readonly detail: D, public readonly wrappedError?: Error) {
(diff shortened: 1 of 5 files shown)


core/src/exceptions.ts
```diff
@@ -183,29 +198,35 @@ export function formatGardenErrorWithDetail(error: GardenError) {
...
+    const filePath = atLine[2]
+    let lastSrcFilePos = -1
+    let tmpPos = -1
+
+    if ((tmpPos = filePath.lastIndexOf("src")) > -1) {
+      lastSrcFilePos = tmpPos + 4
+    } else if ((tmpPos = filePath.lastIndexOf("node_modules")) > -1) {
+      lastSrcFilePos = tmpPos + 13
+    } else if ((tmpPos = filePath.lastIndexOf("node:internal")) > -1) {
+      lastSrcFilePos = tmpPos + 14
```
shumailxyz, line 222: Some comments using constant values e.g. 4,13,14 here would be nice.

would be handy to read in future :)

core/src/exceptions.ts
```diff
@@ -214,8 +235,25 @@ export function getStackTraceMetadata(error: GardenError): GardenErrorStackTrace
       },
     ]
   })
+}
+
+export function getStackTraceMetadata(error: GardenError): GardenErrorStackTrace {
+  if (!error.stack && !error.wrappedErrors) {
+    return { metadata: [], wrappedMetadata: undefined }
```
shumailxyz, line 242: I feel like we should wrap the following block in try/catch as well and just return like this in catch
```
return { metadata: [], wrappedMetadata: undefined }
```

shumailxyz (approved):
some non blocking nits that would be nice to address.

core/src/exceptions.ts
vvagaytsev: Agree, also we can log something if an error occurs in the `try`-block.

mkhq: Wrapped the call points in analytics.ts with try catch since we can log the errors there.

shumailxyz (approved):
Looks good to me ✅ Thank you.

shumailxyz:
Also, lets squash while merging.

PR #4706, opened 2023-06-25
fix(cloud): emit stable action uids across phases

thsig:

Before this fix, `log` events emitted by router methods wouldn't include the same `actionUid` as the status events emitted by the `emitGetStatusEvents` and `emitProcessingEvents` helpers. This led to those action logs not loading in Cloud.

This was because the memoized `getUid()` instance method on the `BaseTask` didn't result in the UID being passed on to the resolved (and later, the executed) action, since those are initalized with the same constructor params as their parent action.

This was fixed by explicitly including the `uid` in the base action constructor params, which means that it gets passed on to the resolved and executed phases of the action (if any).


eysi09 (approved):
LGTM!

PR #4707, opened 2023-06-25
chore(analytics): improve error reporting structure

mkhq:

Garden errors are aggregated in several steps when executing a graph. This PR simplifies the error reporting in analytics by only including the `lastError`. This is the last error added to the list of errors in the results of a graph execution. The struct reported back in analytics has detailed data on three levels:

- `error`, which refers to the top-most error
- `wrapped`, the first wrapped error in the top-most error
- `leaf`, the first leaf found when recursively iterating through wrapped errors. This is the most specific error and should provide a good indication of the error cause.

In addition to the new structure, we also started on an error context struct which can contain specific fields that would help identify task-types or plugins that typically lead to a failed execution.

Additional care has been taken to avoid reporting back any sensitive data. E.g. we opt-out of including task names which could simplify analysis.

Fixes #

- This changes the graph export to include wrapped errors recursively, its not entirely clear to me what the implications can be?

### DIFF
--- a/core/src/analytics/analytics.ts
+++ b/core/src/analytics/analytics.ts
@@ -23,7 +23,13 @@ import { Profile } from "../util/profiling"
 import { ModuleConfig } from "../config/module"
 import { UserResult } from "@garden-io/platform-api-types"
 import { uuidv4 } from "../util/random"
-import { GardenBaseError, StackTraceMetadata, getStackTraceMetadata } from "../exceptions"
+import {
+  GardenBaseError,
+  GardenError,
+  GardenErrorContext,
+  StackTraceMetadata,
+  getStackTraceMetadata,
+} from "../exceptions"
 import { ActionConfigMap } from "../actions/types"
 import { actionKinds } from "../actions/types"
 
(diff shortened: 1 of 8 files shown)


vvagaytsev (changes requested):
There are some merge conflicts, let's rebase this.

core/test/unit/src/analytics/analytics.ts
```diff
@@ -702,6 +704,269 @@ describe("AnalyticsHandler", () => {
...
+
+      const startTime = new Date()
+      timekeeper.freeze(startTime)
+
+      analytics = await AnalyticsHandler.factory({ garden, log: garden.log, ciInfo })
+
+      timekeeper.travel(startTime.getTime() + 60000)
+      const event = analytics.trackCommandResult("testCommand", [], startTime, 0)
+
+      expect(event).to.eql({
```
Orzelius, line 733: I think it'd be cleaner if we had only one of these snapshot tests and the other cases just validated only what's needed. That way I'd be dry-er and would decrease line count by a good bit.

core/src/analytics/analytics.ts
```diff
@@ -600,6 +612,59 @@ export class AnalyticsHandler {
...
+          // the first wrapped error
+          wrapped: e.wrappedErrors?.at(0) ? getErrorDetail(e.wrappedErrors?.at(0)!) : undefined,
+          // recursively get all the leaf errors and select the first one
+          leaf: e.wrappedErrors ? getLeafErrors(e).at(0) : undefined,
+        }
+      } catch (err) {
+        this.log.silly(`Failed to get analytics error detail from ${e}, ${e.wrappedErrors?.at(0)}, ${err}`)
+        return []
+      }
+    })
```
vvagaytsev, line 652: It looks like these 3 functions are re-created on every call of `resultErrorProperties`. Should we declare those outside the class?

mkhq: Broke out into a separate helpers file to clean up this function/file a bit. It's only called once at the end of a command run.

vvagaytsev (approved):
LGTM, thanks! 👍

PR #4724, opened 2023-06-27
refactor: garden error takes named params instead of positional

mkhq:

This changes the `GardenBaseError` constructor to take named params instead of positional. Increases type safety and will make future refactorings/improvements easier. The important change is in `core/src/exceptions.ts`, the rest is just fall-out from that.

Fixes #

### DIFF
--- a/core/src/exceptions.ts
+++ b/core/src/exceptions.ts
@@ -49,7 +49,7 @@ export abstract class GardenBaseError<D extends object = any> extends Error impl
   constructor({ message, detail, stack, wrappedErrors }: GardenErrorParams<D>) {
     super(message)
     this.detail = detail
-    this.stack = stack
+    this.stack = stack || this.stack
     this.wrappedErrors = wrappedErrors
   }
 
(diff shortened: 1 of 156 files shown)


core/src/exceptions.ts
```diff
@@ -33,12 +33,23 @@ export type GardenErrorStackTrace = {
...
   abstract type: string
+  public message: string
+  public detail?: D
+  public wrappedErrors?: GardenError<any>[]
 
-  constructor(message: string, public readonly detail: D, public readonly wrappedErrors?: GardenError[]) {
+  constructor({ message, detail, stack, wrappedErrors }: GardenErrorParams<D>) {
     super(message)
     this.detail = detail
+    this.stack = stack || this.stack
```
vvagaytsev, line 52: Isn't `this.stack` in the right-hand side always undefined at this point?

vvagaytsev (approved):
LGTM, thank you! 👍 
I left one non-blocking comment.

core/src/exceptions.ts
mkhq: It's actually set by the underlying error. We can pass in a stack to override if needed.

PR #4728, opened 2023-06-28
feat(server): add internal _shell command and new WS endpoint

edvald:
This is to support desktop shell commands. The former is a bit crude and can't really support xterm.js, the latter is likely what we'll want once we set up the terminal emulator on the other side.

cc @trymbill


trymbill (dismissed):
LGTM 👍

vvagaytsev (changes requested):
There are some failures that look relevant in the [build-dist](https://app.circleci.com/pipelines/github/garden-io/garden/18709/workflows/188f5497-e5f9-4801-9f19-1714dc80d1e6/jobs/336426) CI step. Please take a look.

edvald:
Ugh yeah, looks like there isn't a prebuild for Windows there. Will need to revisit.

thsig (approved):

PR #4730, opened 2023-06-30
feat(openshift): Run and Test actions

### LINKED ISSUE #4626: [FEATURE]: experimental local-openshift container Run action
Part of the OpenShift epic https://github.com/garden-io/garden/issues/4243

---

Definition of done: users are able to use a Run action to run a command within the OpenShift context.

Tasks:

- Figure out the best plan for this. Should the `kind` be `container` or a dedicated `openshift-*` like with `kubernetes-pod` and `kubernetes-exec`?
- Create integration tests using OpenShift Local

### LINKED ISSUE #4627: [FEATURE]: experimental local-openshift container Test action
Part of the OpenShift epic https://github.com/garden-io/garden/issues/4243

---

Definition of done: users are able to use a Test action to run tests within the OpenShift context.

Tasks:

- Create support for running a unit test within an OpenShift container
- Create support for running an integration test across OpenShift containers / services
- Create integration tests using OpenShift Local

Walther:

Fixes #4626 
Fixes #4627 

### DIFF
--- a/core/src/plugins/kubernetes/namespace.ts
+++ b/core/src/plugins/kubernetes/namespace.ts
@@ -233,6 +233,11 @@ export async function getSystemNamespace(
   api?: KubeApi
 ): Promise<string> {
   const namespace = { name: provider.config.gardenSystemNamespace }
+  // HACK: in OpenShift, we work in only one namespace, the one assigned to the developer
+  if (!namespace.name) {
+    log.warn("No system namespace found, using the current namespace. If you are using OpenShift, ignore this warning.")
+    namespace.name = provider.config.namespace!.name
+  }
 
   if (!api) {
     api = await KubeApi.factory(log, ctx, provider)
--- a/core/src/plugins/openshift/openshift.ts
+++ b/core/src/plugins/openshift/openshift.ts
@@ -8,6 +8,11 @@
 
 import { ConfigureProviderParams } from "../../plugin/handlers/Provider/configureProvider"
 import { createGardenPlugin } from "../../plugin/plugin"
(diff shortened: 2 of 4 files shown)


core/src/plugins/kubernetes/namespace.ts
```diff
@@ -233,6 +233,11 @@ export async function getSystemNamespace(
   api?: KubeApi
 ): Promise<string> {
   const namespace = { name: provider.config.gardenSystemNamespace }
+  // HACK: in OpenShift, we work in only one namespace, the one assigned to the developer
+  if (!namespace.name) {
+    log.warn("No system namespace found, using the current namespace. If you are using OpenShift, ignore this warning.")
```
Orzelius, line 238: Is there a reason to print the message in the first place for openshift users?

shumailxyz (approved):
LGTM, Thank you 💯

core/src/plugins/kubernetes/namespace.ts
Walther: This is within the generic kubernetes code, and we don't necessarily know here if we are running OpenShift or not. Something to fix later when we have the OpenShfit provider in a better shape and can propagate the knowledge around

Orzelius: I think the ctx passed here has that data though

Walther: Currently, we don't have a _proper_ OpenShift provider context - it is using the KubernetesProvider and its context under the hood. Definitely something we need to improve later!

PR #4731, opened 2023-06-30
improvement(k8s): retry on config retrieval failure

### LINKED ISSUE #4591: Investigate retries for kubectl commands
Currently, we can sometimes see errors like

```
2023-06-09T09:09:25.3573989Z error: error when retrieving current configuration of:
2023-06-09T09:09:25.3574834Z Resource: "apps/v1, Resource=deployments", GroupVersionKind: "apps/v1, Kind=Deployment"
2023-06-09T09:09:25.3634918Z Name: "redacted", Namespace: "redacted"
2023-06-09T09:09:25.3636527Z from server for: "STDIN": Get "redacted": dial tcp redacted_ip:443: i/o timeout
```

We should investigate whether we have logic set up for retries - either in our own code, or by `kubectl`'s own features, possibly depending on `kubectl` version as well, etc

vvagaytsev:
Wrap `kubectl` call into existing retry-machinery. Add a way to avoid `shouldRetry` checker call,
because sometimes the error messages are not well-known in advance.

Use retry machinery when retrieve Kubernetes configs using `kubectl`.

Attempts to address #4591

### DIFF
--- a/core/src/plugins/kubernetes/api.ts
+++ b/core/src/plugins/kubernetes/api.ts
@@ -946,7 +946,16 @@ export async function getKubeConfig(log: Log, ctx: PluginContext, provider: Kube
       kubeConfigStr = (await readFile(provider.config.kubeconfig)).toString()
     } else {
       // We use kubectl for this, to support merging multiple paths in the KUBECONFIG env var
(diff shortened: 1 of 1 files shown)


shumailxyz (approved):
Thanks a lot for implementing this 👍

PR #4735, opened 2023-07-03
improvement: allow to deploy all actions with --skip-dependencies

### LINKED ISSUE #4417: feat: allow garden deploy skip-dependencies without specifying a deploy action
For larger stacks with many deploys that depend on each other it doesn't always make sense to wait for each deploy to be complete before moving on to the next one in the dependency tree.

For test environments just deploying the most vital infra and then all of the Deploys at once will most likely suffice. All garden has to do is wait for all the deploys to be ready and them mark the command complete. Of course this means the Deploys have to be configured in a way where they don't go into crashloop if a dependency is not yet up.

Deploys with output dependencies will of course have to wait, but most of these are infra related anyways.

This would greatly increase the startup times of environments that get deployed from the zero point but has the downside of making the api more confusing and people misusing the feature for production.

Orzelius:
closes #4417


core/src/commands/deploy.ts
```diff
@@ -87,7 +87,7 @@ export const deployOpts = {
   }),
   "skip-dependencies": new BooleanParameter({
     help: deline`
-    Deploy the specified actions, but don't build, deploy or run any dependencies. This option can only be used when a list of Deploy names is passed as CLI arguments.
+    Deploy the specified actions, but don't build, deploy or run any dependencies.
```
vvagaytsev, line 90: I'm not sure if the old documentation was accurate enough here. The `BaseActionTask.skipRuntimeDependencies` flag is used only in one place here: https://github.com/garden-io/garden/blob/main/core/src/tasks/base.ts#L283.

It looks like the Build actions can still be performed. Have you tested that?

core/src/commands/deploy.ts
```diff
@@ -87,8 +87,9 @@ export const deployOpts = {
   }),
   "skip-dependencies": new BooleanParameter({
     help: deline`
-    Deploy the specified actions, but don't build, deploy or run any dependencies. This option can only be used when a list of Deploy names is passed as CLI arguments.
-    This can be useful e.g. when your stack has already been deployed, and you want to run specific Deploys in sync mode without building, deploying or running dependencies that may have changed since you last deployed.
+    Deploy the specified actions, but don't deploy or run any dependencies.
```
shumailxyz, line 90:
```suggestion
    Deploy the specified or all actions, but don't deploy or run any dependencies.
```

shumailxyz (dismissed):
Thank you 👍

Orzelius:
Did re-enable the dismiss reviews after a new commit has been pushed thing?

shumailxyz:
@Orzelius Yes, it has been re-enabled.

core/src/commands/deploy.ts
```diff
@@ -87,8 +87,8 @@ export const deployOpts = {
   }),
   "skip-dependencies": new BooleanParameter({
     help: deline`
-    Deploy the specified actions, but don't build, deploy or run any dependencies. This option can only be used when a list of Deploy names is passed as CLI arguments.
-    This can be useful e.g. when your stack has already been deployed, and you want to run specific Deploys in sync mode without building, deploying or running dependencies that may have changed since you last deployed.
+    Skip deploy and run dependencies. Build dependencies and runtime output reference dependencies are not skipped.
```
vvagaytsev, line 90: Test dependencies are also skipped, right? Should we mention it here?

Orzelius: Good catch, usually deploys don't depend on tests but it's worth adding in anyway. [2edbeea](https://github.com/garden-io/garden/pull/4735/commits/2edbeea1a29637921b37632e554d04099e096b76)

vvagaytsev (dismissed):
LGTM 🚀

Orzelius:
I need one more approve on this bad boy

vvagaytsev (approved):
Let this bad boy be good :)

PR #4737, opened 2023-07-03
fix: don't use var file when path is unresolved

### LINKED ISSUE #4702: ${environment.name} is not interpolated appropriately on deploy action
## Bug
I have the following  Deploy action that sets up some kubernetes secrets that utilizes ${environment.name} to select the correct varfile to merge into the deployment but it does not work. Note previously this worked with Modules Here is an example of the Module that worked before
```yaml
kind: Module
name: secrets
type: kubernetes
manifests:
  $forEach: ${var.secrets}
  $return:
    apiVersion: v1
    kind: Secret
    metadata:
      name: ${item.key}
    type: Opaque
    stringData:
      $merge: ${item.value}
varfile: secrets.${environment.name}.yaml
```

### Current Behavior
```yaml
kind: Deploy
name: secrets
type: kubernetes
spec:
  manifests:
    $forEach: ${var.secrets}
    $return:
      apiVersion: v1
      kind: Secret
      metadata:
        name: ${item.key}
      type: Opaque
      stringData:
        $merge: ${item.value}
varfiles: 
  - ./secrets.${environment.name}.yaml
  ```
  
I get the following error when I run garden deploy
```
Error processing config for Deploy action secrets:
Could not find varfile at path './secrets.${environment.name}.yaml'
```
If i hard code it to the environment e.g. "local" it works.

### Expected behavior
It should interpolate the environment name as part of the filepath

### Reproducible example

git clone https://github.com/KoduIsGreat/garden-bug
cd to garden-bug
garden deploy -l5

### Workaround

N/a

### Additional context
Discussed this in discord with worldofgeese

### Your environment

* **OS:** M1 Mac OS Monteray
* I'm using kind for kubernetes

garden version: 0.13.3

shumailxyz:
in action pre-processing, only use the var files that are not template strings as those are resolved and populated later.

Fixes #4702

### DIFF
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -462,10 +462,12 @@ export const preprocessActionConfig = profileAsync(async function preprocessActi
   const description = describeActionConfig(config)
   const templateName = config.internal.templateName
 
+  // in pre-processing, only use varfiles that are not template strings
+  const resolvedVarFiles = config.varfiles?.filter((f) => !maybeTemplateString(f))
   const variables = await mergeVariables({
     basePath: config.internal.basePath,
     variables: config.variables,
-    varfiles: config.varfiles,
+    varfiles: resolvedVarFiles,
   })
   const resolvedVariables = resolveTemplateStrings(
     variables,


vvagaytsev (approved):
Nice fix, thank you! 👍

Issue #4095, opened 2023-04-18
garden publish does not respect deploymentRegistry


Walther:
### Current Behavior

Running `garden publish` in `examples/demo-project`, which uses the GCR image registry, results in an attempted push to the `docker.io` registry

![Screenshot 2023-04-18 at 15 59 32](https://user-images.githubusercontent.com/2943750/232788391-dcd56ce6-c0ae-4b63-b472-4426326160f6.png)

### Expected behavior

`garden publish` should respect the `deploymentRegistry` configuration and publish the image to the appropriate registry.

### Reproducible example

examples/demo-project

### Additional context

Originally reported by a customer in a support conversation, more details in internal Slack.

### Your environment

* **OS:** 
* **How I'm running Kubernetes:**  

`garden version`

jalper:
One customer affected by this bug currently is using a workaround where instead of using garden publish they've got a custom shell script that calls AWS commands directly to publish the image tag. This also results in it being much faster since it doesn't need to pull the image down to their machine locally and then re-upload it.

worldofgeese:
@Walther If you scroll down to *Building and pushing images to multiple registries from CI/CD*
 of [this PoC doc](https://garden-preview.motif.land/flux-fastapi-mongodb-garden), that should function as a solid reproducible example for you 😺

vvagaytsev:
This is re-opened by #6052 that reverted #4740.

vvagaytsev:
The `publish` command should not respect the deployment registry defined in the project-level provider configuration.

The full public image identifier (registry, image name and version tag) should be defined in `spec.publishId` of the Build-action. This behavior was broken prior to #6050 and was fixed in #6052.

Closing this as not planned.

PR #4740, opened 2023-07-04
fix: respect deploymentRegistry in the garden publish command

Walther:

Fixes https://github.com/garden-io/garden/issues/4095

### DIFF
--- a/core/src/plugins/kubernetes/container/publish.ts
+++ b/core/src/plugins/kubernetes/container/publish.ts
@@ -8,14 +8,19 @@
 
 import { ContainerBuildAction } from "../../container/moduleConfig"
 import { KubernetesPluginContext } from "../config"
-import { publishContainerBuild } from "../../container/publish"
 import { pullBuild } from "../commands/pull-image"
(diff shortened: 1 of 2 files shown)


core/src/plugins/kubernetes/container/publish.ts
```diff
@@ -25,10 +30,17 @@ export const k8sPublishContainerBuild: BuildActionHandler<"publish", ContainerBu
...
+  log.info({ msg: `Publishing image ${remoteId}...` })
+  // TODO: stream output to log if at debug log level
+  await containerHelpers.dockerCli({ cwd: action.getBuildPath(), args: ["push", remoteId], log, ctx })
+
+  return {
+    state: "ready",
+    detail: { published: true, message: `Published ${remoteId}` },
+    // TODO-0.13.1
+    outputs: {},
+  }
```
Walther, line 45: copied as-is from the container plugin's `publish` method

vvagaytsev (approved):
Thanks for the fix! 👍

PR #4745, opened 2023-07-04
ci: fix e2e vote-module test

shumailxyz:
An attempt to resolve the failing e2e vote-module tests. Updates the node version in vote-module test project. 

The test seemed to be failing due to some dependencies issue in the vote `build` and since the dependencies layer was cached by kaniko, every build of `vote` resulted in bad dependencies.

Updating node version should result in invalidating of the kaniko dependencies layer cache and fetch dependencies again.


Orzelius (approved):
good idea

vvagaytsev (approved):

PR #4747, opened 2023-07-04
chore(test): fix remote source repo link

vvagaytsev:
Use main branch that contains the action-based configs and the updated Node vesrion.

Fixes #

### DIFF
--- a/examples/remote-sources/garden.yml
+++ b/examples/remote-sources/garden.yml
@@ -4,9 +4,11 @@ name: remote-sources
 sources:
   - name: web-services
     # use #your-branch to specify a branch, #v0.3.0 for a tag or a full length commit SHA1
+    # This tests branch/tag. If you use tag, remember to update it if the target repo changes.
     repositoryUrl: https://github.com/garden-io/garden-example-remote-sources-web-services.git#main
   - name: db-services
-    repositoryUrl: https://github.com/garden-io/garden-example-remote-sources-db-services.git#main
+    # This tests SHA1 commit hash, please use commit hashes if the target repo is updated.
+    repositoryUrl: https://github.com/garden-io/garden-example-remote-sources-db-services.git#66903d73fbca3daa6ed53d04a4b03ddd9dafea43
 environments:
   - name: local
     variables:
--- a/examples/remote-sources/garden.yml
+++ b/examples/remote-sources/garden.yml
@@ -4,7 +4,7 @@ name: remote-sources
 sources:
   - name: web-services
     # use #your-branch to specify a branch, #v0.3.0 for a tag or a full length commit SHA1
-    repositoryUrl: https://github.com/garden-io/garden-example-remote-sources-web-services.git#0.13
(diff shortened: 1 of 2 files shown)


examples/remote-sources/garden.yml
```diff
@@ -4,7 +4,7 @@ name: remote-sources
 sources:
   - name: web-services
     # use #your-branch to specify a branch, #v0.3.0 for a tag or a full length commit SHA1
-    repositoryUrl: https://github.com/garden-io/garden-example-remote-sources-web-services.git#0.13
+    repositoryUrl: https://github.com/garden-io/garden-example-remote-sources-web-services.git#main
```
Orzelius, line 7: This should have worked fine as the 0.13 branch exists and has action configs on the repo
https://github.com/garden-io/garden-example-remote-sources-web-services/tree/0.13

I'd like to have both a branch and a tag based links here as the logic for fetching is completely different and both should be tested

Orzelius (changes requested):

examples/remote-sources/garden.yml
vvagaytsev: It stopped working with `0.13` branch because it did have undefined versions for Python dependencies. So I updated the `main` branch in the target repo and switched the link to point to the `main` branch. Just to avoid maintaining one more branch in the example repo.

The previous version of this config also used 2 URLs with branches. Do we have different code paths to clone tags and branches? I haven't found any difference yet.

We have function [`parseGitUrl`](https://github.com/garden-io/garden/blob/main/core/src/vcs/git.ts#L51) that checks URL validity and returns URL and "hash" (a hash can be either sha1 commit hash or branch or tag).
If it's not a sha1 hash, we just clone the branch, see https://github.com/garden-io/garden/blob/main/core/src/vcs/git.ts#L537

So, it looks like we handle tag and branch URLs in the same way. Just tested it, and it works with both branches and tags. If it stops working at some point, it will be caused by a breaking change in Git.

We can still have one tag URL here. I'd like to keep this and https://github.com/garden-io/garden/blob/main/e2e/projects/remote-sources-modules/garden.yml aligned. So, I'll make a new tag on the latest revision of the `main` branch in https://github.com/garden-io/garden-example-remote-sources-db-services.git and will change the URL of `db-services` here.

vvagaytsev: Changed the second link to point to a Git tag in https://github.com/garden-io/garden/pull/4747/commits/e56ee90d8cfceb760881b052a800008ec67bdf7c

Orzelius: ahh yep, I mixed up the commit SHA and tag cloning logic, my bad. Do we test the SHA clone anywhere?

vvagaytsev: Added SHA1 URLs in https://github.com/garden-io/garden/pull/4747/commits/f41daea7b1d8be8caa90f532245f68c17a8a72b0

Orzelius (approved):
thanks

shumailxyz (approved):
Awesome 💯  Thank you.

PR #4759, opened 2023-07-06
fix: task description conversion

### LINKED ISSUE #4757: 0.13: [Bug]: module to action conversion: 'description' field mapped incorrectly
### Current Behavior

I'm trying to upgrade our garden stack to bonsai.

In our stack we have the following module:
```
kind: Module
type: container
name: db-extensions
image: ${var.postgresql-image}
tasks:
  - name: ${this.name}
    description: Set up postgres extensions
    dependencies:
      - postgresql-credentials
    command:
      - bash
    args:
   ....
```

When running `garden deploy`, the module-to-action auto-conversion fails:
```
Error validating spec for Run type=container name=db-extensions (from module db-extensions) (modules/db-extensions): key "description" is not allowed at path [description]. Available keys: command, args, env, cpu, memory, volumes, privileged, addCapabilities, dropCapabilities, tty, deploymentStrategy, artifacts, image, cacheResult)
```

It looks like the `tasks[0].description` field of the container module is mapped to `spec.description` of the Run action.

### Expected behavior

According to the [docs](https://docs.garden.io/reference/action-types/run/container#description-1) the field should be mapped to `description` instead of `spec.description`.

### Your environment

* **OS:** arch linux

`garden version`
0.13.6

Orzelius:
fixes #4757

Turns out tests and services didn't even have a description field


shumailxyz (approved):
Thank you for getting to it promptly 💯

Issue #4763, opened 2023-07-06
0.13: [Bug]: stack graph resolution performance degradation


hnicke:
### Current Behavior

I'm about to upgrade our stack to bonsai.
We are seeing a drastic performance degradation in the module / action resolution step.

On v12, `garden validate` takes ~6 seconds.
On v13, `garden validate` takes ~51 seconds.

At first I was suspecting that the module-to-action conversion takes so much longer, but according to the logs this conversion only takes ~3 seconds.
However, it takes about ~44 seconds to perform the module scanning.
Please do note that the number of `Found xx files` in the logs are the same in v12 and v13.

PS: After further debugging I found out that the `exclude` action / module configuration is extremely slow in v0.13.
After migrating to `.gardenignore` files the performance issue is gone.

### Expected behavior

I'd expect `garden validate` to take <10 seconds.

### Workaround

Use `.gardenignore` files instead of `exclude` in action / module config.

### Additional context

FYI, most of our modules reside in remote sources.

### Your environment

* **OS:** arch linux
* **How I'm running Kubernetes:** GKE

`garden version`
0.12.61
0.13.6

edvald:
You're definitely running into some sort of a performance bug here. Typically it's quite the opposite when moving over. It might be best if we troubleshoot this over in our discord where you could perhaps send detailed logs with timestamps. Would you mind posting there? Someone on our team (even myself) will then help out. Whatever we find we'll then post back here for posterity.

hnicke:
Great, thank you! discord thread: https://discord.com/channels/817392104711651328/1126473787693604884

vvagaytsev:
@hnicke the fix has been released in [0.13.9](https://github.com/garden-io/garden/releases/tag/0.13.9), please try it out.

PR #4768, opened 2023-07-07
file inclusion improvement + docs

Orzelius:
see commits

related to #4763

### DIFF
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -284,7 +284,7 @@ export const actionFromConfig = profileAsync(async function actionFromConfig({
   const dependencies = dependenciesFromActionConfig(log, config, configsByKey, definition, templateContext)
 
   if (config.exclude?.includes("**/*")) {
-    if (config.include?.length !== 0) {
+    if (config.include && config.include.length !== 0) {
       throw new ConfigurationError({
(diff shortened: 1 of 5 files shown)


core/src/graph/actions.ts
```diff
@@ -282,6 +283,18 @@ export const actionFromConfig = profileAsync(async function actionFromConfig({
 
   const dependencies = dependenciesFromActionConfig(log, config, configsByKey, definition, templateContext)
 
+  if (config.exclude?.includes("**/*")) {
```
Walther, line 286: should we have equals here instead of includes, in case of something like `src/test/**/*`?

shumailxyz: also, can we also add a test for this?

Orzelius: `config.exclude` is an array of strings. It's checking if the array includes the exact match of "**/*"

Orzelius: Writing tests was a good idea, found a bug in my original logic and removed some unwanted mutation
[e4a9dce](https://github.com/garden-io/garden/pull/4768/commits/e4a9dcef26e256511af177244f7c34b3d76c034e)

shumailxyz (approved):
Nice, thank you 👍

Walther (approved):
nice work! 🚢

PR #4769, opened 2023-07-07
feat: otel collector integration

TimBeyer:
This PR enables us to locally deploy an [otel collector](https://github.com/open-telemetry/opentelemetry-collector) by configuring it as a provider.

The provider once configured finds an open port, creates a configuration file on the fly and then spawns the collector on the local machine. It then reconfigures the existing otel exporter to target the new collector, which then proceeds to export the data to the desired destination.

The collector gracefully shuts down at the end of the garden session after attempting to flush all remaining data.

An example project config could look like this:

```
apiVersion: garden.io/v0
kind: Project
name: otel-collector-test
environments:
  - name: local
    defaultNamespace: ${var.dev-env-name}
providers:
  - name: local-kubernetes
    environments: [local]
    namespace: ${environment.namespace}
  - name: otel-collector
    exporters:
      - name: otlphttp
        enabled: true
        endpoint: http://localhost:4318
      - name: newrelic
        enabled: true
        apiKey: ${secrets.NR_API_KEY}
      - name: honeycomb
        enabled: true
        apiKey: ${secrets.HONEYCOMB_API_KEY}
        dataset: ${secrets.HONEYCOMB_DATASET}
      - name: datadog
        enabled: true
        site: ${secrets.DD_EXPORTER_SITE}
        apiKey: ${secrets.DD_EXPORTER_API_KEY}
variables:
  dev-env-name: ${project.name}-testing-${local.username}
```

Currently supported are a generic `otlphttp` type for any HTTP OTLP endpoint, and configs for `newrelic`, `honeycomb` and `datadog`.

The environment variable `GARDEN_ENABLE_TRACING` is now `true` by default, allowing just the override to disable it.  
When not configured however, it sets up a "no op" exporter so that we're not leaking memory.
By default no data is sent anywhere without explicit configuration, this is just so that there's still an override for disabling it but there's no need to manually set it to enable the collector provider to work.

If a user would like to override the exporter at the lowest level, skipping the provider based setup and just using an environment variable, that is still possible via setting `OTEL_TRACES_EXPORTER` based on the options documented in https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-sdk-node/README.md#configure-trace-exporter-from-environment

For example to trace to a JSON HTTP OTLP endpoint on localhost, one could use

```
OTEL_TRACES_EXPORTER=otlp OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318/ OTEL_EXPORTER_OTLP_PROTOCOL=http/json garden deploy
```

In the majority of cases that should not be necessary, but it still does exist in case a user has their own OTEL setup and does not want to deploy an additional collector via the provider.

Fixes #


core/src/util/process.ts
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  const logger = ctx.log.createLog({
+    name,
+  })
+
+  logStream.on("data", (line: Buffer) => {
+    const logLine = line.toString()
+    ctx.events.emit("log", { timestamp: new Date().toISOString(), msg: logLine, ...logEventContext })
+    logger.silly(logLine)
```
TimBeyer, line 46: If `streamLogs` is used we now send all logs from subprocesses also to the CLI log output using the silly loglevel.
This is quite useful in order to debug issues with tools we call from garden.

Orzelius (changes requested):
I think the formatter might have been misconfigured. There are many styling changes which seem to be unrelated and make it hard to review

core/src/garden.ts
```diff
@@ -1279,7 +1298,7 @@ export class Garden {
         this.configPaths.add(path)
       }
 
-      const allResources = flatten(
+      let allResources = flatten(
```
Orzelius, line 1301: This change doesn't seem to be needed

core/src/plugins/otel-collector/config/logging.ts
```diff
@@ -0,0 +1,61 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { inferType } from "../../../config/zod"
+import { sdk } from "../../../plugin/sdk"
+import { baseValidator } from "./base"
+
+export type LoggingExporterVerbosityLevel = "detailed" | "normal" | "basic"
```
Orzelius, line 13: "basic" doesn't sound like a verbosity level to me. Is it some terminology coming from elsewhere that I'm unaware of? Why not the usual "info", "verbose" and "trace" or "all"?

core/src/plugins/otel-collector/otel-collector.ts
```diff
@@ -0,0 +1,191 @@
...
+const OTEL_CONFIG_NAME = "otel-config.yaml"
+
+export interface OtelCollectorProviderConfig extends GenericProviderConfig {}
+export type OtelCollectorProvider = Provider<OtelCollectorProviderConfig>
+
+export const gardenPlugin = sdk.createGardenPlugin({
+  name: "otel-collector",
+  dependencies: [{ name: "exec" }],
+  docs: dedent`
+  Otel Collector
```
Orzelius, line 37: Will the docs be added later?

core/src/plugins/otel-collector/otel-collector.ts
```diff
@@ -0,0 +1,191 @@
...
+const providerConfigSchema = s.object({
+  exporters: s.array(
+    s.union([loggingValidator, otlpHttpValidator, newRelicValidator, dataDogValidator, honeycombValidator])
+  ),
+})
+
+export const provider = gardenPlugin.createProvider({ configSchema: providerConfigSchema, outputsSchema: s.object({}) })
+
+provider.addHandler("getEnvironmentStatus", async ({ ctx }) => {
+  return { ready: false, outputs: {} }
```
Orzelius, line 102: Is the hardcoded `ready: false` intentional?

core/src/plugins/otel-collector/otel-collector.ts
```diff
@@ -0,0 +1,191 @@
...
+      await processExited
+      scopedLog.debug("Process exited")
+    })
+
+    return { status: { ready: true, disableCache: true, outputs: {} } }
+  } catch (err) {
+    // TODO: We might not want to fail if the collector didn't initialize correctly
+    scopedLog.error("otel-collector failed to initialize")
+    return { status: { ready: false, disableCache: true, outputs: {} } }
+  }
```
Orzelius, line 190: I think the error should be logged as well

docs/reference/providers/otel-collector.md
```diff
@@ -0,0 +1,116 @@
...
+    #
+    # Example: `["dev","stage"]`
+    environments:
+
+    exporters:
+      - name:
+
+        enabled:
+
+        verbosity: normal
```
Orzelius, line 41: Will the docs come later?

Orzelius (commented):
Thank you! Incredibly clean and easy to read code! I've posted some comments and am wondering about the lack of tests and docs, but other than that it looks sweet. Sorry for the long wait for the review, I just forgot about the pr.

core/src/garden.ts
TimBeyer: You're right. I have no idea where this change came from.
Shouldn't our linter have `prefer-const` enabled and catch this?

Orzelius: Seems like it's not configured to do so properly

core/src/plugins/otel-collector/otel-collector.ts
TimBeyer: I believe it is. If I recall this correctly, garden would check first if the environment is ready, and if so it would not call `prepareEnvironment` again.
We need to call `prepareEnvironment` every time, so we always say it's not ready.

Even worse is that if you just once leave out this hook, the status is assumed to be complete with cache enabled, and the cached state is written to the filesystem making it that without a `--force` the provider will never initialize again.

We also found that returning `disableCache: true` in `getEnvironmentStatus` does not actually disable the cache and will still proceed caching the state.

This it is required to return `ready: false` here and then later on return `disableCache: true` from the `prepareEnvironment` handler.

docs/reference/providers/otel-collector.md
TimBeyer: The problem here is that `oneOf` and `anyOf` schemas [aren't correctly converted into docs yet](https://github.com/garden-io/garden/blob/de50fcc7ccfa29b520facbe83bee8fdb31caea8b/core/src/docs/json-schema.ts#L37-L45).

I think we should fix this soon since it makes the docs incomplete, but since it's not strictly related and once fixed the docs will show up for this as well, I decided to not make it part of this PR.

shumailxyz:
Great work @TimBeyer 💯 

Just read through the PR. Wondering the same what @Orzelius mentioned about lack of tests and docs.

TimBeyer:
@shumailxyz @Orzelius I thought about how to test this but couldn't come up with a good way.
If we unit test this we need to mock almost everything that makes this work in the first place leading to not very solid tests.
Probably e2e tests would be best, but then we'd have to create an OTLP HTTP endpoint and verify that we get all the data, which is difficult since we're sending quite a lot of data that can also change at any time if new spans are added.
The schemas for the config generation could be tested, but they are very simple and the typesystem already makes sure that inputs and outputs are correct.
That just leaves the `ReconfigurableExporter` and the function to find a string in a childprocess output.
I can add tests for those if you prefer.
Also if you have a good idea for how to tests the overall feature without having to mock a lot of things and that will always work even if we change the exact things we trace, I'm happy to take a look at that too.

Orzelius:
I think a e2e test is the way to go here. It would serve as a smoke test for if we make a fatal mistake that would kill the functionality completely and would double as an example configuration for when we need to do further work on the functionality.

We already have some clumsy e2e tests that do a lot, but that's the business we're in

core/src/util/process.ts
stefreak: This is slightly weird now, as we also send the messages emitted as `log` event to `ctx.events` to our logger with `verbose` log level.
Do we need duplicate logs both on silly and verbose?

core/src/util/util.ts
```diff
@@ -90,13 +90,18 @@ export async function shutdown(code?: number) {
       // eslint-disable-next-line no-console
       console.log(getDefaultProfiler().report())
     }
-    process.exit(code)
+    gracefulExit(code)
```
stefreak, line 93: 🚀 Nice that's much better

core/src/util/process.ts
TimBeyer: The way I understood it is that ctx.events only goes to cloud and isn't actually logged. That's why I added the extra log statement so it shows up in core as well.

stefreak: The action router also logs these: https://github.com/garden-io/garden/blob/main/core/src/router/build.ts#L44C12-L44C12

Do you see the logs only once or twice in manual testing?

stefreak: It's a little surprising maybe, but I added that log statement there so that we always send logs to both Cloud and the CLI

stefreak: For tests and tasks it does that too, but there only if the interactive flag is false: https://github.com/garden-io/garden/blob/main/core/src/router/test.ts#L49

That's because logs get streamed to stdout in that case

TimBeyer: I see the logs only once

stefreak: Interesting, that's strange can we have a look at this together?

TimBeyer:
@shumailxyz @Orzelius Added some e2e smoke tests for the feature.

core/src/util/process.ts
TimBeyer: I fixed this so we shouldn't see duplicate logs from the places where you already added the logging

Orzelius (commented):
Looks good, but what about docs 👉👈?

Orzelius (approved):
@TimBeyer and I decided that the provider reference docs should suffice. This is good to merge from my side. cc @garden-io/core

Issue #3873, opened 2023-03-02
Can not supply nested value via --var cli option


hnicke:
### Current Behavior

Currently it is not possible to supply a nested value via `--var` cli option.

Running...
```
garden deploy --var foo.bar=baz

```
...results in the following variable:
```yaml
"foo.bar": baz
```

### Expected behavior

It should result in the following variables instead:
```yaml
foo:
  bar:  baz
```

### Workaround

Don't use nested variables. Seems like a sad workaround though.

### Your environment

* **OS:** arch linux
* **How I'm running Kubernetes:** GKE

`garden version`
0.12.52

stefreak:
Workaround: it is possible to use project-level varfiles to override yaml structures

hnicke:
This doesn't feel like a good workaround for ad hoc overrides.

PR #4773, opened 2023-07-07
fix: override action variables via --var cli flag

### LINKED ISSUE #4772: 0.13: [Bug]: variables supplied via --var cli flag has no effect on action level variables
### Current Behavior

Variables supplied via `--var` cli flag only apply on the project level and have no effect on the variables defined on action level. So, it's not possible to override action level variables using `--var` cli flag.

### Expected behavior

Variables supplied via `--var` cli flag should also apply on action variables.

### Reproducible example

Consider the following garden config for a `backend` service.

project.garden.yml
```yaml
apiVersion: garden.io/v1
kind: Project
name: repro-action-vars
environments:
  - name: local
providers:
  - name: local-kubernetes
    environments: [local]
    namespace: ${project.name}-testing-${var.userId}
```

garden.yml

```yaml
kind: Build
name: backend
description: Backend service container image
type: container

---

kind: Deploy
name: backend
description: Backend service container
type: container

build: backend

# You can specify variables here at the action level
variables:
  ingressPath: /hello
  nestedmap:
    foo: bar

spec:
  healthCheck:
    httpGet:
      path: ${var.ingressPath}
      port: http
  ports:
    - name: http
      containerPort: 8080
      # Maps service:80 -> container:8080
      servicePort: 80
  ingresses:
    - path: ${var.ingressPath}
      port: http
  env:
    USER: ${var.nestedmap.foo}
    INGRESS_PATH: ${var.ingressPath}

---
```
and `backend` is simply a go file. 

``` go
// main.go
package main

import (
	"fmt"
	"net/http"
	"os"
)

	fmt.Fprint(w, "Backend says: Hello!", os.Getenv("USER"))
}

func main() {
	http.HandleFunc(os.Getenv("INGRESS_PATH"), handler)
	fmt.Println("Server running...")

	err := http.ListenAndServe(":8080", nil)
	if err != nil {
		panic(err)
	}
}

```

Deploy this using this command:

```
garden deploy --force --force-build  --var ingressPath="/welcome"
```

Note the ingress url. The url will be 
```
http://repro-action-vars.local.demo.garden/hello
``` 
whereas it should in fact be 
```
http://repro-action-vars.local.demo.garden/hello
``` 
as we override the `ingressPath` variable and set it to `/welcome` via `--var` cli flag.

### Your environment

* **OS:** 

`garden version`: 0.13.6

shumailxyz:
A follow up PR to #3873 for addressing #4772. Previously, variables set using the --var cli flag only had impact on the
project level variables. 

Fixes #4772

### DIFF
--- a/core/src/garden.ts
+++ b/core/src/garden.ts
@@ -1996,7 +1996,7 @@ export const resolveGardenParams = profileAsync(async function _resolveGardenPar
 })
 
 // Override variables, also allows to override nested variables using dot notation
-function overrideVariables(variables: DeepPrimitiveMap, overrideVariables: DeepPrimitiveMap): DeepPrimitiveMap {
+export function overrideVariables(variables: DeepPrimitiveMap, overrideVariables: DeepPrimitiveMap): DeepPrimitiveMap {
   let objNew = cloneDeep(variables)
   Object.keys(overrideVariables).forEach((key) => {
     if (objNew.hasOwnProperty(key)) {
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -39,7 +39,7 @@ import type { GroupConfig } from "../config/group"
 import { ActionConfigContext } from "../config/template-contexts/actions"
 import { validateWithPath } from "../config/validation"
 import { ConfigurationError, InternalError, PluginError, ValidationError } from "../exceptions"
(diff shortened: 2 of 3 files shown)


edvald (approved):

edvald:
Good catch!

PR #4780, opened 2023-07-10
chore(examples): fix octal numbers in manifest

thsig:

Before this fix, the `kubernetes-deploy` example failed to deploy because of a problem with the way octal numbers in manifests were parsed.


Walther (approved):
thanks for the quick fix! ✨

PR #4786, opened 2023-07-10
feat: add `with-dependants` deploy command flag

### LINKED ISSUE #2759: [FEATURE]: conditionally deploy services that have a dependency on another service when it is deployed
### Background / Motivation

we have service-b that has a dependencies on service-a and also consumes an output from service a.. for example a dns entry. when we change service-a's outputs and re-deploy service-a, service-b has stale values causing it to not be able to connect to service-a. we need to then redeploy service-b to pick up the new output values.

automatically deploy services that have a dependency on the service that is being deployed. should be able to specify if you only want to deploy if the dependent service is already running or deploy if its missing.

not having to redeploy dependent services yourself

a couple of options added to the deploy command: --with-dependants and --update-only, the rormer willatter indicating that missing services should not be deployed.

Orzelius:
closes #2759

### DIFF
--- a/core/src/commands/deploy.ts
+++ b/core/src/commands/deploy.ts
@@ -36,6 +36,7 @@ import { LogMonitor } from "../monitors/logs"
 import { LoggerType, parseLogLevel } from "../logger/logger"
 import { serveOpts } from "./serve"
 import { gardenEnv } from "../constants"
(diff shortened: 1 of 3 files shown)


shumailxyz (approved):
Thanks, very nice! 👍

PR #4787, opened 2023-07-10
fix(cloud): send "sessionFailed" event if result has errors

eysi09:

Previously we'd send a "sessionCompleted" event which caused Cloud to display the wrong command status.

This is really just a quickfix since it's urgent but we should add some tests here.  

Fixes #

### DIFF
--- a/core/src/cloud/api.ts
+++ b/core/src/cloud/api.ts
@@ -94,8 +94,8 @@ export type ApiFetchResponse<T> = T & {
 
 // TODO: Read this from the `api-types` package once the session registration logic has been released in Cloud.
 export interface CloudSessionResponse {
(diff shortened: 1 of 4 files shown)


core/src/cloud/api.ts
```diff
@@ -94,8 +94,8 @@ export type ApiFetchResponse<T> = T & {
 
 // TODO: Read this from the `api-types` package once the session registration logic has been released in Cloud.
 export interface CloudSessionResponse {
-  environmentId: number
-  namespaceId: number
+  environmentId: string
```
eysi09, line 97: This types here were wrong.

vvagaytsev (approved):

Issue #4748, opened 2023-07-05
garden sync status --output flag ignored


andreavocado:
### Current Behavior

Running `garden sync status -o yaml` outputs `Command includes global options that will be ignored: --output` while it gives a hint to use that: `Getting sync statuses. For more detailed debug information, run this command with the `--output json` or `--output yaml` flags.`

### Expected behavior

Detailed output **OR** remove the hint the command can be executed with this flag.

### Reproducible example

Run `garden sync status -o yaml` in your garden dev console

### Further Information

![mutagen](https://github.com/garden-io/garden/assets/85166619/cc6af37a-68e2-415a-aad4-b6dbf47a43ba)

### Your environment

* **OS:** MacOS 13.4.1
* **How I'm running Kubernetes:** Google cloud 
* **garden version**: 0.13.5

Walther:
Thank you for reporting this UX issue! The hint message has been removed in https://github.com/garden-io/garden/pull/4791.

PR #4791, opened 2023-07-11
chore: remove output flag hint from sync status

Walther:

Fixes https://github.com/garden-io/garden/issues/4748

Added a breakpoint in the debugger and looked around, did not find any obvious variables/flags denoting that we would be within an interactive `garden dev` repl. Removing the hint message entirely, instead of only in dev mode.


vvagaytsev (approved):
Thanks! 👍

PR #4801, opened 2023-07-11
chore(example): remove `imagePullPolicy` from `templated-k8s-container`

worldofgeese:

This is a very simple PR to fix Garden attempting to pull from a non-existent Docker registry. Garden should always build and push our templated image.

The `imagePullPolicy: "Always"` setting in Kubernetes means that it will always attempt to pull the image from a Docker registry.

Since we're using the templated `Build` image for our manifest, I've removed this line, which automatically sets `ImagePullPolicy: IfNotPresent`, our desired behavior.


vvagaytsev (approved):
Thanks for fixing this! 👍

PR #4804, opened 2023-07-11
feat(cli): add cloud secrets update command

### LINKED ISSUE #3079: [FEATURE]: Ability to perform UPSERT garden secrets
### Background / Motivation

At our company, we have a strict policy that secrets must be provisioned into an Azure Key Vault (AKV). We found alignment with our security team that secrets can be synchronized from AKV into garden enterprise secrets.  So, our team has constructed a way to read all secrets from AKV and execute `garden cloud secrets create [secrets]` for each secret to obtain parity.  

Unfortunately, now when a secret changes (as in the case of a password rotation), our Azure Key Vault is updated, but the only way to update the corresponding secret in garden enterprise is to manually delete it, and then re-create it.  This has rendered our synchronization routine useless.

```sh
garden cloud secrets set [secret_name]=[secret_value]
```

This command should update a secret, and if it doesn't exist, it should create it.  Effectively, an "upsert."

This would allow us to synchronize secrets into garden secrets in an idempotent way.  That is, being able to UPSERT a secret would allow us to load secrets en-masse and have everything in sync with external syste.

I could see this feature revealing itself in two ways:
- As mentioned above, a way to use the command line to UPSERT secrets would be immensely helpful.  I chose the word `set` as it felt more flexible than `update` and `upsert` sounded kind-of awkward.
- On the garden enterprise secrets page, it would be nice to be able to click on a secret and be able to edit its value (this is not necessary for our team, but it does kind of fit in with this feature request).

We're able to get by with out it, as we can delete and re-create secrets, but it leads to situations where passwords are updated in one place, and not another, resulting in un-expected behavior.

shumailxyz:
Adds a new command to update secrets in cloud. 
```
garden cloud secrets update
```

Fixes #3079

- Added a flag `upsert` to perform upsert and create missing secrets while updating existing secrets.
- Added a flag `update-by-id` to update secrets by id.
- As there can be multiple secrets with same name across different env/users, added flags `scope-to-env` and `scope-to-user-id` to perform scoped filtering.
- An error is thrown if update is performed by name and after applying all filters, multiple secrets with same name are found.
- Added some unit tests however integration tests are missing for this new command as well as other secrets commands. Would be added separately for all the secrets operations together.


core/src/cloud/api.ts
```diff
@@ -591,6 +591,18 @@ export class CloudApi {
     })
   }
 
+  async put<T>(path: string, opts: ApiFetchOptions & { body?: any } = {}) {
+    const { body, headers, retry, retryDescription, maxRetries } = opts
+    return this.apiFetch<T>(path, {
+      method: "PUT",
+      body: body || {},
+      headers: headers || {},
+      retry: retry === true ? true : false, // defaults to false unless true is explicitly passed
```
Orzelius, line 600: There must be a cleaner way to write this 🤔

How about
```suggestion
      retry: !!retry
```

Orzelius (changes requested):
I think the documentation should also mention that secrets will be created if they don't exist.

If you want to make the creation of secrets default behavior the command should be upsert not update. If you want to keep it update I think that the creation of new secrets should be behind a flag, something like `--upsert`.

Although we decided that we won't be testing this e2e with cloud I still think this code should be split into smaller testable parts and unit tests should be written to ensure quality and make future developments easier.

core/src/commands/cloud/secrets/secrets-update.ts
```diff
@@ -0,0 +1,314 @@
...
+  args,
+}: {
+  allSecrets: SecretResult[]
+  envName?: string
+  userId?: string
+  secretsToUpdateArgs: StringMap
+  log: Log
+  args: T
+}): Promise<Array<UpdateSecretBody>> {
+  let tmp = sortBy(allSecrets, "name")
```
eysi09, line 42: Kind of a nitpick but I'd prefer something more descriptive like `sortedSecrets` instead of `tmp`

core/src/commands/cloud/secrets/secrets-update.ts
```diff
@@ -0,0 +1,314 @@
...
+  if (envName) {
+    tmp = tmp.filter((secret) => secret.environment?.name === envName)
+  }
+  if (userId) {
+    tmp = tmp.filter((secret) => secret.user?.id === userId)
+  }
+  tmp = tmp.filter((secret) => Object.keys(secretsToUpdateArgs).includes(secret.name))
```
eysi09, line 49: How about chaining all these together so that you don't have to constantly re-assign the `tmp` variable?

Something like:

```
const filteredSecrets = sortBy(allSecrets, "name")
  .filter((s) => envName ? s.environment.name === envName : true)
  .filter((s) => userId ? ...)
  .filter((s) => ...)
```

core/src/commands/cloud/secrets/secrets-update.ts
```diff
@@ -0,0 +1,314 @@
...
+  secretNamesOrIds: new StringsParameter({
+    help: deline`The name(s) or ID(s) of the secrets to update along with the new values, separated by '='.
+      You may specify multiple secret id/value pairs, separated by spaces. `,
+    spread: true,
+  }),
+}
+
+export const secretsUpdateOpts = {
+  "upsert": new BooleanParameter({
+    help: "Set this flag to upsert secrets instead of update i.e. existing secrets are updated while missing secrets are created.",
```
eysi09, line 98:
```suggestion
    help: "Set this flag to upsert secrets instead of updating them. I.e., existing secrets will be updated while missing secrets will be created.",
```

core/src/commands/cloud/secrets/secrets-update.ts
```diff
@@ -0,0 +1,314 @@
...
+    defaultValue: false,
+  }),
+  "update-by-id": new BooleanParameter({
+    help: deline`Update the secret(s) by providing the ID(s) of secrets.
```
eysi09, line 102:
```suggestion
    help: deline`Update secret(s) by secret ID(s).
```

core/test/unit/src/commands/cloud/secrets/secrets-update.ts
```diff
@@ -0,0 +1,180 @@
...
+import {
+  SecretsUpdateCommand,
+  getSecretsToCreate,
+  getSecretsToUpdateByName,
+} from "../../../../../../src/commands/cloud/secrets/secrets-update"
+import { StringMap } from "../../../../../../src/config/common"
+import { deline } from "../../../../../../src/util/string"
+import { expectError, getDataDir, makeTestGarden } from "../../../../../helpers"
+
+describe("SecretsUpdateCommand", () => {
```
eysi09, line 20: Maybe out of scope for now since this PR is already pretty old but moving forward I think we should be testing the cloud commands with `nock` and just examining the request body. 

That way we can just create the Garden command instance, test all the different args and opts and examine what the POST body looks like. For this case we can also mock the get request.

I'm typically not a fan of mocking but these kind of tests are an exception.

eysi09 (commented):
Great work! 

Left a few comments but overall looks good.

core/src/commands/cloud/secrets/secrets-update.ts
shumailxyz: Sure, updated in https://github.com/garden-io/garden/pull/4804/commits/5b7c5a057f86a0ba4307de8c5b57448efe6597e7

core/test/unit/src/commands/cloud/secrets/secrets-update.ts
shumailxyz: yeah makes sense. May be we can create some kind of package that provides mock for our api client and can be reused whenever required

eysi09 (approved):
Thanks for addressing the PR comments. Looks good to me!

PR #4810, opened 2023-07-12
fix(template): allow `null` as a valid argument in helper functions and fix `isEmpty()` helpers

### LINKED ISSUE #3398: isEmpty() computes wrong value with null values
### Current Behavior
With a variable defined as null...

```
variables:
  my-value: null
```

...the following template...

```
${isEmpty(var.my-value)}
```

...evaluates to false.

The following snippet in the context of a kubernets manifest of a kubernetes module...

```
${isEmpty(null)}
```

causes this exception:

```
Failed deploying service 'tls-cert' (from module 'tls-cert'). Here is the output:
Invalid template string (apiVersion: cert-manager.io/v1
kin…): Cannot read property '_error' of null
1 deploy action(s) failed!
```

I find this slightly disturbing.

### Expected behavior
As documented here: https://docs.garden.io/reference/template-strings/functions#isempty

### Your environment

* **OS:** arch linux

`garden version`
0.12.46

vvagaytsev:
Ports #4809 to the `main` branch.

Fixes #3398

### DIFF
--- a/core/src/template-string/functions.ts
+++ b/core/src/template-string/functions.ts
@@ -9,7 +9,7 @@
 import { v4 as uuidv4 } from "uuid"
 import { createHash } from "crypto"
 import { TemplateStringError } from "../exceptions"
-import { keyBy, mapValues, escapeRegExp, trim, isEmpty, camelCase, kebabCase, isArrayLike, isString } from "lodash"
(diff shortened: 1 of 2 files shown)


shumailxyz (approved):
Looks good. Thank you for the fix and porting it to the main branch 👍

PR #4822, opened 2023-07-13
fix(serve): connect to Cloud if process is started outside of project dir

edvald:
This came up in the Desktop client because it'd start the serve command outside of any project root, and thus not connect to Cloud at all for the live view etc. to work.

cc @eysi09


vvagaytsev:
@eysi09 is this ready to do? LGTM, but I might miss some context.

eysi09:
> @eysi09 is this ready to do? LGTM, but I might miss some context.

I'm not sure. Maybe I'm missing something, but when we connect to Cloud we need to do that in the context of some specific project. What would that be in the case where Garden is started outside of a given project?

trymbill:
@eysi09 and I just verified this works now for Desktop. So LGTM! 👍

vvagaytsev (approved):

eysi09:
Just to follow up. 

I think we just need to make sure we only set the `localServerPort` for the `dev` or `serve` command. So when registering a session for, say, a `deploy` command, we need to make sure that the local server port isn't set in the payload. 

This hasn't been an issue until the change in this issue was introduced. Basically the added `registerSession` call appears to sometime set a server port when it shouldn't.

core/src/server/instance-manager.ts
```diff
@@ -403,7 +403,8 @@ export class GardenInstanceManager {
         // Use the process (i.e. parent command) session ID for the serve/dev command session
         sessionId: this.sessionId,
         commandInfo: garden.commandInfo,
-        localServerPort: this.serveCommand.server.port,
+        // set localServerPort only for dev/serve commands
+        localServerPort: ["dev", "serve"].includes(garden.commandInfo.name) ? this.serveCommand.server.port : undefined,
```
eysi09, line 407: This is called in more places and we need to update it all of them. 

Can I also suggest that in the `registerSession` function we set the `localServerPort` type to `number | null` so that it's explicit and we won't forget to add it at other call sites.

eysi09 (changes requested):

core/src/server/instance-manager.ts
shumailxyz: I think we don't need this on other places. We basically have 3 calls of `registerSession` including this one. The other 2 calls are in `base.ts` and  `serve.ts`.

The `base.ts` call already has a check and skips the registration in case of dev/serve.
https://github.com/garden-io/garden/blob/8490aacf9f1e29278212f91ac3f3af31eab38b88/core/src/commands/base.ts#L311-L315

And in [`serve.ts`](https://github.com/garden-io/garden/blob/8490aacf9f1e29278212f91ac3f3af31eab38b88/core/src/commands/serve.ts#L179), we do not need this check as it only registers session for dev/serve.

eysi09: Aha ok. Makes sense.

eysi09 (approved):

Issue #4783, opened 2023-07-10
new VCS repo scanning: file exclusion not working properly


hnicke:
### Current Behavior
I have tried the new VCS repo scanning algorithm (`scan.git.mode: repo`) that ships with v0.13.7 (#4642) since I hoped it would help with the stack resolution performance problems described in #4763.

Something is off with the file exclusion.
E.g., one module with ~1100 files (without new algorithm) now has around ~10000, according to the log.

Another module (with node_modules folder) normally has ~1000 files (without new algorithm), with new algorithm log reports there are ~200000 files.

I have defined all exclusions via module config `excludes` array.
It seems they don't work properly.

Therefore, with the new vcs algorithm `garden validate` takes >5 minutes.

### Expected behavior

Both repo scanning algorithms detect the same amount of files for each module.

### Workaround 

Use `.gardenignore` files for file exclusions instead of `exclude` in module /action configuration.

### Your environment

* **OS:** arch linux
* **How I'm running Kubernetes:** GKE

`garden version`
0.13.7

@edvald might be interesting for you

hnicke:
I have migrated our stack from using `exclude` in module config to using `.gardenignore` files which fixed the issue for me.

stefreak:
Good catch @hnicke – thanks for the detailed report! 👍 
I am adding this as candidate to our weekly board so someone will have a look at this soon.

hnicke:
I currently don't have the time to provide a full example, but here's the gist:

Given the `node_modules` folder is huge as almost always the case.

The project config is configured to ignore node_modules:
```
scan:
  git:
    mode: repo
  include:
    - "*.garden.yml"
  exclude:
    - "**/node_modules/**/*"
```

---

The following module (without further ignorefiles) will take ages to scan (i.e., `garden validate`):
```yaml
# example.garden.yml
kind: Module
type: container
name: example
exclude:
  - node_modules/**
```
The logs (`.garden/logs/validate.debug..`) will show that the exclusion didn't work: `...Found <huge number> files in module example...`.

---

However, when changing the module like this, the perfomance issue is gone:
```yaml
# example.garden.yml
kind: Module
type: container
name: example
```
```yaml
# .gardenignore
node_modules
```
---

By the way - in my case the module is located in a remote source - not sure if that is a factor that has to be taken into account.

stefreak:
Thank you for the more detailed information

I tried to reproduce this and saw the following behaviour

# Steps to reproduce
1. Create `garden.yml`
[...]

PR #4823, opened 2023-07-13
perf(git): optimize git scan when exclude but no include filter is set

### LINKED ISSUE #4763: 0.13: [Bug]: stack graph resolution performance degradation

edvald:
This can be a major optimization in certain cases.

The reason this wasn't done previously was due to a git ls-files issue but we missed out on this path when no includes are used, in which case the issue doesn't come up.

Usually includes are much more specific than excludes in configs, so this approach should cover a lot of user scenarios.

It is still more optimal to use `.gardenignore` files over exclude patterns where possible, but using excludes in configs is obviously  a common and supported use case, and not always possible to use the ignore files (say, if many actions/modules share a directory and need different exclusions).

Fixes #4763

### DIFF
--- a/core/src/vcs/git.ts
+++ b/core/src/vcs/git.ts
@@ -300,17 +300,40 @@ export class GitHandler extends VcsHandler {
 
     const absExcludes = exclude.map((p) => resolve(path, p))
 
-    // Apply the include patterns to the ls-files queries. We use the 'glob' "magic word" (in git parlance)
+    // Apply the include patterns to the ls-files queries. We use the --glob-pathspecs flag
     // to make sure the path handling is consistent with normal POSIX-style globs used generally by Garden.
-    // Note: We unfortunately can't exclude at this level because it simply doesn't work in git, for reasons unknown.
-    const patterns = [...(include || []).map((p) => ":(glob)" + p)]
+
+    // Due to an issue in git, we can unfortunately only use _either_ include or exclude patterns in the
+    // ls-files commands, but not both. Trying both just ignores the exclude patterns.
+
+    if (include?.includes("**/*")) {
+      // This is redundant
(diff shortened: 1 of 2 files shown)


Orzelius:
I think this closes https://github.com/garden-io/garden/issues/4763?

edvald:
I believe so, yes.

stefreak:
Does it also fix #4783? If yes it would be great if we add a test case that catches the issue described in https://github.com/garden-io/garden/issues/4783#issuecomment-1630607457 to make this code more safe to change in the future 👍

edvald:
It does not. Would be great to address that but it's a separate issue, and I wasn't able to repro it. That one might have something to do with submodules or at least some specific case that I'm not quite capturing.

shumailxyz (approved):
Looks good, thanks

PR #4824, opened 2023-07-13
fix: mutagen default permissions are too restrictive

stefreak:
Especially for in-cluster building, the file permissions mutagen
defaults to are too restrictive.

Git adheres to the system umask by default, which usually is set to 022.
That leads to file permissions 644 and directory permissions 755, which
means group and world readable. I would expect this to be the default on
any system.

Dockerfile's COPY directive by default copies files including the
permissions.

In Garden 0.12 we used rsync by default to transfer the build context for in-cluster builds, and thus files had the correct permissions as rsync also transferred the permissions.

The new default in 0.13, Mutagen, does not transfer the file permissions
by default, and thus when using in-cluster building together with the
COPY directive, the file permissions in the resulting container image
will differ compared to using a local Docker daemon.

This commit changes the default file permissions for mutagen to 644 and
the default directory permissions to 755. This is usually a good
default, and if users need different permissions they can use the
`--chmod` option in Dockerfile's `COPY` directive or the respective Garden sync mode settings.

Sorry, I was too lazy to open a bug report for this one. Instead here's the fix and a description of why we need this above :)

This is not a breaking change, because the default permissions become more permissive than they were before this change, when using mutagen. This leads effectively to the same permissions of transferred files as with the rsync context transfer mode + umask 022 (which is usually the default)– that's why I consider this a bug fix for a regression introduced in 0.13 (we changed the default build context transfer subroutines from rsync to mutagen in 0.13)


core/src/plugins/container/config.ts
```diff
@@ -194,8 +194,9 @@ export const syncDefaultFileModeSchema = memoize(() =>
     .number()
     .min(0)
     .max(777)
+    .default(644)
     .description(
-      "The default permission bits, specified as an octal, to set on files at the sync target. Defaults to 0600 (user read/write). " +
+      "The default permission bits, specified as an octal, to set on files at the sync target. Defaults to 0600 (user can read/write, everyone else can read). " +
```
vvagaytsev, line 199: It defaults to `0644`, not to `0600`, right?

stefreak: Thank you, corrected that and generated docs.

stefreak:
@vvagaytsev I could use some help for writing an e2e-test for this. Would you be available for a pairing session?

vvagaytsev (dismissed):
Good job! But there are some failed assertions in the tests, please check and fix :)

worldofgeese:
In our Discord, a user [notes this is a breaking change](https://discord.com/channels/817392104711651328/1126240052830277722/1130396805515976755) for users migrating from Acorn to Bonsai. Quote,

> I managed to solve issue with wrong permissions by using following Mutagen config at `~/.mutagen.yml`:
> ```yaml
> sync:
>   defaults:
>     permissions:
>       defaultFileMode: 0644
>       defaultDirectoryMode: 0755
> ```
> However, I think, it can't be considered as a proper solution, because permissions are hardcoded for all docker images and are not defined within Garden project folder (not portable). Ideally, permissions must be copied over with files to `util` container in `garden-buildkit` pod as is. Alternatively, we can define permission for each docker image in corresponding `Dockerfile`, but in this case I would like you to confirm that current behavior won't change in future.

Please see the Discord thread for further context.

stefreak:
@worldofgeese Thank you for letting me know. This PR will solve this issue 👍

core/src/plugins/container/config.ts
```diff
@@ -204,10 +204,10 @@ export const syncDefaultFileModeSchema = memoize(() =>
     .number()
     .min(0o0)
     .max(0o777)
-    .default(0o644)
+    .default(() => 0o644)
```
Walther, line 207: (tiny nit: considering this didn't affect the table rendering issue, we cna probably switch this back to an octal literal instead of a function)

core/src/plugins/container/config.ts
```diff
@@ -217,10 +217,10 @@ export const syncDefaultDirectoryModeSchema = memoize(() =>
     .number()
     .min(0o0)
     .max(0o777)
-    .default(0o755)
+    .default(() => 0o755)
```
Walther, line 220: (ditto)

core/src/plugins/container/config.ts
stefreak: I agree, will roll this back

docs/reference/action-types/Deploy/helm.md
```diff
@@ -273,13 +273,13 @@ spec:
       # `.git` directories and `.garden` directories are always ignored.
       exclude:
 
-      # The default permission bits, specified as an octal, to set on files at the sync target. Defaults to 0644 (user
-      # can read/write, everyone else can read). See the [Mutagen
+      # The default permission bits, specified as an octal, to set on files at the sync target. Defaults to 0o644
+      # (user can read/write, everyone else can read). See the [Mutagen
       # docs](https://mutagen.io/documentation/synchronization/permissions#permissions) for more information.
       fileMode: 420
```
Walther: note for future reference, hoisting from our pairing call: unfortunately our documentation generation converts to decimal instead of keeping it as octal. ideally, we should fix this at some point, but it is not critical enough to warrant a deeper investigation right now.

stefreak: Now implemented in https://github.com/garden-io/garden/pull/4824/commits/2c7cda1042b87dd8d6ad455804c0843fe52383a3 thanks to AI

Walther (dismissed):

docs/reference/action-types/Deploy/container.md
```diff
@@ -266,15 +266,15 @@ spec:
         # guide](https://docs.garden.io/guides/code-synchronization) for details.
         mode: one-way-safe
 
-        # The default permission bits, specified as an octal, to set on files at the sync target. Defaults to 0600
-        # (user read/write). See the [Mutagen
+        # The default permission bits, specified as an octal, to set on files at the sync target. Defaults to 0o644
+        # (user can read/write, everyone else can read). See the [Mutagen
         # docs](https://mutagen.io/documentation/synchronization/permissions#permissions) for more information.
-        defaultFileMode:
+        defaultFileMode: 420
```
Walther, line 272: interestingly, the number shows up as decimal here, but correctly as octal in the table below

stefreak: Good catch. I noticed that too, and tried to look into why and tried to fix that, but then I noticed that fixing the number here would require a pretty hefty amount of refactoring in the docs generator. The code that generates the yaml representation is extremely brittle, and making tiny changes here causes other configs to change in ways we don't want them to.

Walther: Yep, let's leave it. The docs should be clear enough as-is 👍

Walther (approved):
Thank you for all your great work here ✨

porterchris:
> This is not a breaking change
This is a breaking change.

A segment of our Acorn based config:
```
    devMode:
      sync:
        - source: ./src
          target: ./target
          mode: one-way-replica
          defaultFileMode: 0666
```

An error message from one of our developers using Acorn:
> (module-name): Error validating Module (module0name)’ (../src): Mode permission bits out of range. Please specify a number between 0 and 0o777 in octal representation. The number needs to be prefixed with ‘0o’ in YAML, e.g. 0o777.. Available keys: build, buildArgs, extraFlags, image, include, hotReload, dockerfile, services, tests, tasks)

There is no mention of devMode > sync > defaultFile Mode changes in the Bonsai migration docs:
https://docs.garden.io/tutorials/migrating-to-bonsai

There are mentions of using Acorn based configuration with Bonsai, and infers that `apiVersion: garden.io/v0` will be used to denote an Acorn based configuration format when the change is breaking. This method was not used in your change to the file mode configuration format.

Your documentation still references the use of decimal based values:
https://docs.garden.io/guides/code-synchronization#permissions-and-ownership

> This is not a breaking change
This is a breaking change.

stefreak:
Hi @porterchris this was definitely not intended to be a breaking change, we're having a look at this immediately

stefreak:
@porterchris this is a pitfall with the YAML standards version 1.1 (Where the octal notation is `0666`) and version 1.2 (Where the octal notation changed to `0o666`); See also https://github.com/garden-io/garden/issues/4929

Really sorry for this unintentional breaking change. We will look into this, but in the meantime I suggest using the newer YAML 1.2-style notation. If we can't solve this problem in a backwards-compatible way, we'll add it to the migration guide of course.

PR #4827, opened 2023-07-13
chore: fix changelog generation

vvagaytsev:
Aligned with `draft-release-notes.ts` script.
Used explicit version range instead of experimental `--next-tag` flag.

Fixes #

To be ported to `0.12` to fix the changelogs computed against the latest 0.13 version. This change has no visible effect in 0.13, but it must be applied as it is the semantically correct approach.


shumailxyz (approved):
Thanks for fixing this 👍

PR #4829, opened 2023-07-13
fix: restore azure devops support for dockerhub images

### LINKED ISSUE #4818: [FEATURE]: Publish Docker images compatible with Azure Pipelines
### Background / Motivation

In conversation with @ITHedgeHog, it came to light that we don't publish container images compatible with Azure Pipelines. Given our focus to improve CI/CD in Q3, I believe this is a great candidate.

Users of Azure DevOps should be able to use our published images without maintaining their own.

Azure Pipelines [do not support `ENTRYPOINT`](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/container-phases?view=azure-devops#linux-based-containers).

I understand we previously had an `azure.Dockerfile` contributed by @ITHedgeHog #2509. We may consider resurrecting this image and using `CMD` instead of `ENTRYPOINT` for its images. I'd be particularly interested in feedback from @stefreak who has been greatly involved in the creation of our fantastic new container images.

stefreak:
We do not need an ENTRYPOINT, and azure devops only supports images
without it.

Fixes #4818

### DIFF
--- a/support/buster.Dockerfile
+++ b/support/buster.Dockerfile
@@ -28,7 +28,9 @@ RUN set -ex; \
 ENV USER=root
 ENV HOME=/root
 
-ENTRYPOINT ["/garden/garden"]
+# We do not set an entrypoint here for compatibility with Azure DevOps pipelines.
+# See also https://learn.microsoft.com/en-us/azure/devops/pipelines/process/container-phases?view=azure-devops#linux-based-containers
+ENTRYPOINT []
 
 FROM buster-base-root as buster-base-rootless
 
--- a/support/alpine.Dockerfile
+++ b/support/alpine.Dockerfile
@@ -28,6 +28,11 @@ RUN apk add --no-cache \
   groff \
   py3-crcmod
 
+# Add tools required for Azure DevOps. See also https://github.com/microsoft/azure-pipelines-agent/blob/master/docs/design/non-glibc-containers.md
+RUN apk add --no-cache --virtual .pipeline-deps readline linux-pam  \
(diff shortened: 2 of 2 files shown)


stefreak:
@ITHedgeHog is there anything else that comes to your mind that could cause incompatibilities with Azure DevOps, or should this change suffice?
It should make all images compatible with Azure Devops now, not just the garden-azure image.

ITHedgeHog:
That looks fine to me, we also used to have to add the following lines to our custom Garden image:

```
RUN apk add --no-cache --virtual .pipeline-deps readline linux-pam  \
  && apk add bash sudo shadow \
  && apk del .pipeline-deps
``` 

I've no idea if they're still required for the image to work on Azure DevOps - just providing them here in case they turn out to be necessary.

support/buster.Dockerfile
```diff
@@ -28,7 +28,9 @@ RUN set -ex; \
 ENV USER=root
 ENV HOME=/root
 
-ENTRYPOINT ["/garden/garden"]
+# We do not set an entrypoint here for compatibility with Azure DevOps pipelines.
+# See also https://learn.microsoft.com/en-us/azure/devops/pipelines/process/container-phases?view=azure-devops#linux-based-containers
+ENTRYPOINT []
```
worldofgeese, line 33: The linked doc specifies, `Doesn't define an ENTRYPOINT` as a requirement. Why are we defining an empty entrypoint? Shouldn't it be removed completely?

worldofgeese (commented):
Do we want to add a default `CMD`?

Issue #4816, opened 2023-07-13
0.13: [Bug]: ${parent.name} not available in Module templates


Orzelius:
## Error

```
An error occurred while reading pulumi varfiles for action my-deploy: Unable to load varfile at '***/path/pulumi/config/global.yaml': Error: Invalid template string (# App name refers to the Deploymen…): Invalid template string (${parent.name}-publish-image): Could not find key parent. Available keys: actions, command, datetime, environment, git, inputs, local, modules, project, providers, runtime, secrets, this, var and variables
```

Our doc says

> The module names can reference the # ${inputs.*}, ${parent.name} and ${template.name} keys

ishantagarwal-x213235:
we are using pulumi module so can you please check with below configuration using pulumi provider

```modules:
  - type: pulumi
    name: ${parent.name}-service # ${var.test}-${parent.name} #${parent.name}-module
    description: common to all REST services
    root: ${inputs.deploymentFolder}/pulumi
    stack: "${environment.name}"

    variables:
      excludeDefault: ["${parent.name}"]
      pulumiProjectName: ${var.gardenProjectName}-${var.projectname}-${parent.name}

    createStack: true
    autoApply: true

    deployFromPreview: ${var.withPlan}

    pulumiVarfiles: 
       - "${local.projectPath}/deployment-common/pulumi/config/global.yaml"
       - "${inputs.deploymentFolder}/config/common.yaml"
       - "${inputs.deploymentFolder}/config/${environment.name}.yaml"

PR #4836, opened 2023-07-14
fix: allow to use `parent.name` template in templated modules

Orzelius:
closes https://github.com/garden-io/garden/issues/4816


vvagaytsev (approved):
Nice! Thanks for the fix! ✨

shumailxyz (approved):
nice! 🌟

PR #4838, opened 2023-07-14
introduce --target option for exec command to target specific container in pod

### LINKED ISSUE #3039: `garden exec` and `garden logs` wrongly target sidecar containers
### Current Behavior

`garden exec` and `garden logs` always target the first container configured in a pod. Sometimes, sidecar containers are auto-injected. When this happens, `garden exec` and `garden logs` become useless because they target the sidecar container instead of the main application container.

Since Kubernetes 1.18, we can use the `kubectl.kubernetes.io/default-container` annotation to configure this behavior for `kubectl exec` and `kubectl logs`. However, garden does not respect this annotation.

### Expected behavior

`garden exec` and `garden logs` always target the application container. 

### Reproducible example

1. Deploy a container module
2. Have some other service inject a sidecar container (I am experimenting with osiris)
3. Run `garden exec` or `garden logs`
4. Observe that commands target the sidecar container instead of the container managed by Garden

### Workaround

None

### Your environment

* **OS:** macOS
`garden version` 0.12.42

shumailxyz:
Introduces a flag `--target` in `exec` command. This is useful when a Deploy action consists of multiple underlying components and user wants to target a specific one.

At the moment, this flag is only supported for kubernetes-exec and container-exec. When a pod has multiple containers, user can use `target` flag to exec into a specific container. Additionally, if no container name is provided, it respects the annotation `kubectl.kubernetes.io/default-container` and defaults to that container. The existing behavior is still the default: it picks the first container in Pod.

This PR also adds the default container annotation in the manifests generated with `container` type.

Fixes #3039

### DIFF
--- a/core/src/commands/exec.ts
+++ b/core/src/commands/exec.ts
@@ -16,6 +16,7 @@ import { executeAction } from "../graph/actions"
 import { NotFoundError } from "../exceptions"
 import { DeployStatus } from "../plugin/handlers/Deploy/get-status"
 import { createActionLog } from "../logger/log-entry"
+import { K8_POD_DEFAULT_CONTAINER_ANNOTATION_KEY } from "../plugins/kubernetes/run"
 
 const execArgs = {
   deploy: new StringParameter({
@@ -39,6 +40,12 @@ const execOpts = {
     defaultValue: false,
     cliDefault: true,
     cliOnly: true,
+  }),
+  target: new StringParameter({
+    help: `Specify name of the target if a Deploy action consists of multiple components. _NOTE: This option is only relevant in certain scenarios and will be ignored otherwise._ For Kubernetes deploy actions, this is useful if a Deployment includes multiple containers, such as sidecar containers. By default, the container with \`${K8_POD_DEFAULT_CONTAINER_ANNOTATION_KEY}\` annotation or the first container is picked.`,
+    cliOnly: true,
+    defaultValue: undefined,
+    required: false,
   }),
 }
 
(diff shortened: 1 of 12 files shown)


core/src/commands/exec.ts
```diff
@@ -33,13 +33,20 @@ const execArgs = {
...
-  interactive: new BooleanParameter({
+  "interactive": new BooleanParameter({
     help: "Set to false to skip interactive mode and just output the command result",
     aliases: ["i"],
     defaultValue: false,
     cliDefault: true,
     cliOnly: true,
   }),
+  "container-name": new StringParameter({
+    help: "Specify name of specific container for exec if the Pod has multiple containers",
```
Orzelius, line 44: Maybe add here that "By default the container with `kubectl.kubernetes.io/default-container` annotation is picked"

core/src/plugins/kubernetes/run.ts
```diff
@@ -1047,7 +1050,33 @@ export class PodRunner extends PodRunnerParams {
...
+          )
+        }
+      } else {
+        containerName = this.pod.spec.containers[0].name
+        if (this.pod.spec.containers.length > 1) {
+          const allContainerNames = this.pod.spec.containers.map((c) => c.name)
+          log.info(
+            `Defaulted container ${containerName} out of: ${allContainerNames.join(
+              ", "
+            )}. Use the --container-name flag to target specific container in the Pod.`
```
Orzelius, line 1075: This is not always applicable as this logic is also used for kubernetes-exec action types and maybe even something else.

core/src/commands/exec.ts
edvald: Hmm. So, we have to be conscious of the fact that all of this is in plugin territory. Garden itself doesn't know what a container is or Kubernetes for that matter. I think it's fine to mention specifics as examples ("e.g. for kubernetes actions, this will...") but please don't make it seem native to the framework.

With that in mind, perhaps a less specific flag name would make sense. Something like `--target`, which might apply in various scenarios, outside of the very specific case this PR adds support for. It's also important to note that the flag only has meaning in certain scenarios and will otherwise be ignored.

shumailxyz: Good point, a generic name definitely makes more sense here.

Orzelius (approved):
Good Job!  🚢

PR #4842, opened 2023-07-18
fix: enoent when version set to null on tf and pulumi

### LINKED ISSUE #4832: 0.13: [Bug]: Setting the version null in pulumi provider does not work
## Garden Bonsai (0.13) Bug : Garden is not able identify the local path of pulumi when set the version to null

garden is not able to identify the pulumi when set the version to null in ```project.garden.yaml```
Below is the configuration in project.garden.yaml file,

```providers:
  - name: pulumi
    environments: ["dev"]
    # The version of pulumi to use. Set to `null` to use whichever version of `pulumi` is on your PATH.
    version: null
```

Reported Error
```An error occurred while trying to run 'pulumi/pulumi' (spawn pulumi/pulumi ENOENT). Please make sure 'pulumi/pulumi' is installed and in the $PATH.```

**With version 0.12, it detected the correct path of pulumi**
```
:information_source: x-service [debug] → Spawning '/home/workstation/.pulumi/bin/pulumi stack select telushealth/dev --create' in /home/workstation/thx/workspace/services/21062023/thxcs-shared-services/services/x/app/deployment/pulumi
```

**While with 0.13.7 version, it is not detecting the correct path of pulumi**
```
:information_source: deploy.x-service [debug] → Spawning 'pulumi/pulumi stack select telushealth/dev --create' in /home/workstation/thx/workspace/services/02062023/thxcs-shared-services/services/x/app/deployment/pulumi
```

### PR DESCRIPTION
**Pull request by Orzelius**
was caused by me in https://github.com/garden-io/garden/pull/4496

closes #4832

### DIFF
--- a/core/src/util/ext-tools.ts
+++ b/core/src/util/ext-tools.ts
@@ -54,17 +54,17 @@ export class CliWrapper {
   name: string
   protected toolPath: string
 
-  constructor(name: string, path: string) {
+  constructor({ name, path }: { name: string; path: string }) {
     this.name = name
     this.toolPath = path
   }
 
-  async ensurePath(_: Log) {
+  async getPath(_: Log) {
     return this.toolPath
   }
 
   async exec({ args, cwd, env, log, timeoutSec, input, ignoreError, stdout, stderr }: ExecParams) {
-    const path = await this.ensurePath(log)
+    const path = await this.getPath(log)
 
     if (!args) {
       args = []
@@ -105,7 +105,7 @@ export class CliWrapper {
   }
 
   async spawn({ args, cwd, env, log }: SpawnParams) {
-    const path = await this.ensurePath(log)
+    const path = await this.getPath(log)
 
     if (!args) {
       args = []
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: plugins/pulumi/cli.ts [hunk 1]
```diff
@@ -44,7 +44,7 @@ export function pulumi(ctx: PluginContext, provider: PulumiProvider) {
 
 export class GlobalPulumi extends CliWrapper {
   constructor() {
-    super("pulumi", "pulumi/pulumi")
+    super({ name: "pulumi", path: "pulumi" })
```
shumailxyz, line 47: just wondering why the change in path from `"pulumi/pulumi"` to `pulumi` unless the path was wrong earlier.

shumailxyz (approved):
Thanks for fixing this 🌟

core/src/util/ext-tools.ts
```diff
@@ -54,17 +54,17 @@ export class CliWrapper {
...
   protected toolPath: string
 
-  constructor(name: string, path: string) {
+  constructor({ name, path }: { name: string; path: string }) {
     this.name = name
     this.toolPath = path
   }
 
-  async ensurePath(_: Log) {
+  async getPath(_: Log) {
```
vvagaytsev, line 62: Looks like it doesn't have to be async.

vvagaytsev: Ah, actually, it does. The overridden method in a sub-class is async :)

vvagaytsev (approved):

PR #4845, opened 2023-07-18
chore(openshift): ensure garden works with openshift cloud

Walther:

Fixes https://github.com/orgs/garden-io/projects/7/views/1?pane=issue&itemId=32592909


core/src/plugins/kubernetes/container/build/common.ts
```diff
@@ -247,11 +245,19 @@ export async function skopeoBuildStatus({
...
+ Useful for e.g. when getting the build status for an image that has never been pushed before.
+ */
 export function skopeoManifestUnknown(errMsg: string | null | undefined): boolean {
   if (!errMsg) {
     return false
   }
-  return errMsg.includes("manifest unknown") || /(artifact|repository) [^ ]+ not found/.test(errMsg)
+  return (
+    errMsg.includes("manifest unknown") ||
+    errMsg.includes("name unknown") ||
```
Walther, line 258: OpenShift registry returns `name unknown` when the image name does not exist yet in the registry

core/test/data/openshift/demo-project/README.md
```diff
@@ -105,3 +105,79 @@ This will delete the VM running OpenShift, including any configuration done with
...
+- `vote` which can be copied from examples and made to work with slightly modified project configuration.
+
+Copy the vote example to the `vote` directory, and remove the project configuration included with it:
+
+```bash
+cp -r ../../../../../examples/vote/ vote/
+rm vote/garden.yml
+```
+
+NOTE: in case you run into issues, you might need to edit the `Build` actions to have `buildAtSource: true` and/or specify the target platform as `linux/amd64` in additional build arguments.
```
Walther, line 160: We probably have a core bug here somewhere. The `api` of `vote` fails to build and deploy correctly _even in the kaniko in-cluster build` with an exec format error, unless you set `buildAtSource: true`. Which is also weird because that shouldn't affect the executable format.

Issue #4811, opened 2023-07-12
0.13: [Bug]: Deploy cached even though there were changes


stefreak:
### Current Behavior

First deploy (as expected):
```
ℹ deploy.postgres      → missing
ℹ deploy.postgres      → Deploying version v-5f8da6486d...
```

Second deploy (as expected):
```
ℹ deploy.postgres      → Already deployed
```

Third deploy with changed password (unexpected!):
```
ℹ deploy.postgres      → Already deployed
```

### Expected behavior

I expect the postgres module to calculate a different version hash in the third deploy, because the postgres password changed.

### Reproducible example

It happens both with kubernetes modules and native action kinds

- Check out [the `kubernetes-module` example](https://github.com/garden-io/garden/tree/0.12/examples/kubernetes-module) from the 0.12 branch
- Alternatively, check out the [the `kubernetes-deploy` example](https://github.com/garden-io/garden/tree/main/examples/kubernetes-deploy) from the main branch

1. Run `garden deploy postgres`
2. Run `garden deploy postgres`
3. Run `garden deploy postgres --var postgres-password=$(echo somethingelse | base64)`

In step three, it is not important that the manifest changes due to the variable parameter. It also reproduces this bug when changing the value of the `postgres-password` variable in `garden.yml`.

It even reproduces when simply changing the manifest directly in `postgres/garden.yml`

### Workaround

Downgrade to Garden 0.13.6

### Suggested solution(s)

I don't know

### Additional context

This is a regression that has been introduced in 0.13.7.
I suspect it might have been introduced in the bigger changes around https://github.com/garden-io/garden/pull/4516/files

@twelvemo and me discovered this one because internal deploy pipelines stopped working.

### Your environment

* **OS:** macOS
* **How I'm running Kubernetes:** Docker Desktop

`garden version` 0.13.7

stefreak:
Confirmed with @vvagaytsev that this regression has been introduced in #4516

Opened #4812 with a revert

TODO: this needs an integration test

stefreak:
If we find a fix really quickly we can also skip reverting of course 👍

PR #4846, opened 2023-07-18
fix(k8s): correctly resolve manifests when `build` is set

### LINKED ISSUE #4505: 0.13: [Bug]: Empty source folder for build dependencies when running `garden deploy`, but not `garden build`
### Current Behavior

I'm not 100% certain of the cause, but while upgrading Modules to Actions, I have repeatedly encountered error messages indicating an empty build directory. I can confirm that `.garden/sources/actions/<my-action>` is created as an empty directory in these cases.

Before migration, an example particular Module looked like the following. It has a module for building a container, and a second module for deploying a helm chart for the container. It uses a `build.dependencies[]` and references the build module's output:
```yml
kind: Module
type: container
name: gateway-cache-image
description: Image used by gateway-cache
repositoryUrl: git@github.com:B612-Asteroid-Institute/gateway.git#main
dockerfile: ./Dockerfile
services: []
exclude: []
---
kind: Module
type: helm
name: gateway-cache
description: Deploys the simple caching service
repositoryUrl: git@github.com:B612-Asteroid-Institute/gateway.git#main
chartPath: ./chart
exclude: []
build:
  dependencies:
    - gateway-cache-image
values:
  image:
    repository: ${modules.gateway-cache-image.outputs.deployment-image-name}
    tag: ${modules.gateway-cache-image.version}
  hostname: gateway-cache.${var.hostname}
```

I rewrote that to this:
```yml
kind: Build
name: gateway-cache
description: Image used by gateway-cache
type: container
source:
  repository:
    url: git@github.com:B612-Asteroid-Institute/gateway.git#main
---

kind: Deploy
name: gateway-cache
description: Deploys the simple caching service
type: helm
build: gateway-cache
source:
  repository:
    url: git@github.com:B612-Asteroid-Institute/candidate_review.git#main
spec:
  chart:
    path: ./chart
  values:
    image:
      repository: ${actions.build.gateway-cache.outputs.deployment-image-name}
      tag: ${actions.build.gateway-cache.version}
    hostname: gateway-cache.${var.hostname}
```

Then I ran `garden deploy gateway-cache`, and I got the following _build_ error:

```
$ garden deploy gateway-cache
Deploy 🚀

ℹ garden               → Running in Garden environment dev.ai-swnelson
✔ providers            → Cached (took 0.6 sec)
✔ graph                → Done (took 1.1 sec)
✔ build.gateway-cache  → Connected to sync target Build sync Pod
ℹ build.gateway-cache  → [sync]: Sync disconnected
ℹ build.gateway-cache  → [sync]: Scanning files to sync
✔ build.gateway-cache  → [sync]: Completed initial sync from Module gateway-cache build path to Build sync Pod
ℹ build.gateway-cache  → [sync]: Watching for changes
ℹ build.gateway-cache  → [sync]: Scanning files to sync
ℹ build.gateway-cache  → [sync]: Synchronized from Module gateway-cache build path to Build sync Pod at 2:02:53 PM
ℹ build.gateway-cache  → [sync]: Scanning files to sync
ℹ build.gateway-cache  → [sync]: Synchronized from Module gateway-cache build path to Build sync Pod at 2:02:53 PM
ℹ build.gateway-cache  → [sync]: Scanning files to sync
ℹ build.gateway-cache  → [sync]: Synchronized from Module gateway-cache build path to Build sync Pod at 2:02:53 PM
ℹ build.gateway-cache  → File sync to cluster complete
ℹ build.gateway-cache  → Buildkit Building image gateway-cache:v-89d107b987...
✖ build.gateway-cache  → Build failed (took 11.5 sec)
✖ build.gateway-cache  → Failed processing Build type=container name=gateway-cache (took 11.53 sec). Here is the output:

Command exited with code 1:
#1 [internal] load .dockerignore
#1 transferring context: 2B done
#1 DONE 0.0s

#2 [internal] load build definition from Dockerfile
#2 transferring dockerfile: 2B done
#2 DONE 0.0s
error: failed to solve: failed to read dockerfile: open /tmp/buildkit-mount350800148/Dockerfile: no such file or directory
1 deploy action(s) failed!
```

This was confusing, but I eventually figured out that the `.garden/sources/action/build.gateway-cache--*` directory was totally empty:

```
ls -alh .garden/sources/action/build.gateway-cache--f30ff7e3aa
total 0
drwxr-xr-x  2 swnelson  staff    64B May 31 14:02 .
drwxr-xr-x  7 swnelson  staff   224B May 31 14:02 ..
```

Then I ran `garden build gateway-cache` to _explicitly_ build it, and it worked!:
```
$ garden build gateway-cache
Build 🔨

ℹ garden               → Running in Garden environment dev.ai-swnelson
✔ providers            → Cached (took 0.7 sec)
✔ graph                → Done (took 1.1 sec)
✔ build.gateway-cache  → Connected to sync target Build sync Pod
ℹ build.gateway-cache  → [sync]: Scanning files to sync
ℹ build.gateway-cache  → [sync]: Reconciling sync changes
ℹ build.gateway-cache  → [sync]: Staging files to sync in target
ℹ build.gateway-cache  → [sync]: Syncing changes...
ℹ build.gateway-cache  → [sync]: Saving sync archive
✔ build.gateway-cache  → [sync]: Completed initial sync from Module gateway-cache build path to Build sync Pod
ℹ build.gateway-cache  → [sync]: Watching for changes
ℹ build.gateway-cache  → [sync]: Scanning files to sync
ℹ build.gateway-cache  → [sync]: Reconciling sync changes
ℹ build.gateway-cache  → [sync]: Synchronized from Module gateway-cache build path to Build sync Pod at 2:08:18 PM
ℹ build.gateway-cache  → File sync to cluster complete
ℹ build.gateway-cache  → Buildkit Building image gateway-cache:v-fc955ff417...
✔ build.gateway-cache  → Done (took 27.7 sec)
```

Was the original source folder populated? No, but a sibling with a different hash _was_ populated:
```
$ ls -alh .garden/sources/action/build.gateway-cache--f30ff7e3aa
total 0
drwxr-xr-x  2 swnelson  staff    64B May 31 14:02 .
drwxr-xr-x  8 swnelson  staff   256B May 31 14:07 ..

$ ls -alh .garden/sources/action/build.gateway-cache--390814b07d
total 64
drwxr-xr-x  13 swnelson  staff   416B May 31 14:07 .
drwxr-xr-x   8 swnelson  staff   256B May 31 14:07 ..
drwxr-xr-x  13 swnelson  staff   416B May 31 14:07 .git
-rw-r--r--   1 swnelson  staff   857B May 31 14:07 .pre-commit-config.yaml
-rw-r--r--   1 swnelson  staff   565B May 31 14:07 Dockerfile
-rw-r--r--   1 swnelson  staff   397B May 31 14:07 README.md
drwxr-xr-x   5 swnelson  staff   160B May 31 14:07 chart
drwxr-xr-x   6 swnelson  staff   192B May 31 14:07 gateway
-rw-r--r--   1 swnelson  staff   113B May 31 14:07 pyproject.toml
-rw-r--r--   1 swnelson  staff   1.1K May 31 14:07 requirements.txt
-rw-r--r--   1 swnelson  staff   583B May 31 14:07 setup.cfg
-rw-r--r--   1 swnelson  staff    38B May 31 14:07 setup.py
-rw-r--r--   1 swnelson  staff    86B May 31 14:07 uwsgi.ini
```

### Expected behavior
I expect that when I run `garden deploy`, if it has any explicit build dependencies which depend on remote sources, those remote sources will get populated in exactly the same way as with `garden build`. 

### Reproducible example

I'm not sure I have a minimal reproducer, but the above may help.

### Workaround

Run `garden build` explicitly before every deploy.

### Suggested solution(s)

I don't know

### Additional context

I think this is only happening with

### Your environment

* **OS:** macOS

$ garden version
0.13.0

### LINKED ISSUE #4506: 0.13: [Bug]: `build` flag of deploy action has no effect
### Current Behavior

When running `garden deploy`
```
✖ deploy.deploy        → Failed resolving status for Deploy type=kubernetes name=deploy (took 0.03 sec). Here is the output:

ENOENT: no such file or directory, open '/Users/steffen/repro/generated-manifest.yml'
```

- Build action has not been triggered (directory `.garden/build/generate/` does not exist)
- It looks for `generated-manifest.yml` in the wrong directory

When running `garden build` it actually generates the file
```
✔ build.generate       → Done (took 0 sec)

Done! ✔️
```
```
% ls .garden/build/generate/
generated-manifest.yml
```

### Expected behavior

1. it should execute the build action when running `garden deploy`, as it is listed in `dependencies`
2. Successful deployment of the generated manifest

### Reproducible example

```
kind: Project
apiVersion: garden.io/v1
name: repro
environments:
 - name: default
providers:
 - name: local-kubernetes
---
kind: Build
type: exec
name: generate
spec:
  shell: true
  command: ["echo -n 'apiVersion: v1\nkind: ConfigMapList' > generated-manifest.yml"]

---
kind: Deploy
type: kubernetes
name: deploy
dependencies:
 - build.generate
build: generate
spec:
  files:
    - generated-manifest.yml
```

### Workaround

Unknown

### Suggested solution(s)

Use the original solution (https://github.com/garden-io/garden/pull/4516/commits/a5f5092953b70a4d09fbcfddb6c43f4d8f2bb8a7) from #4516 and make sure there is no regression like we had in #4811. Here is the patch with the original changes:

<details>

```
Index: core/src/plugins/kubernetes/status/status.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/core/src/plugins/kubernetes/status/status.ts b/core/src/plugins/kubernetes/status/status.ts
--- a/core/src/plugins/kubernetes/status/status.ts	(revision 409899599c81409852f08b4de1e4f44acce1e84f)
+++ b/core/src/plugins/kubernetes/status/status.ts	(date 1689325155940)
@@ -12,31 +12,36 @@
 import { PluginContext } from "../../../plugin-context"
 import { KubeApi } from "../api"
 import { getAppNamespace } from "../namespace"
-import { KubernetesResource, KubernetesServerResource, BaseResource, KubernetesWorkload } from "../types"
-import { zip, isArray, isPlainObject, pickBy, mapValues, flatten, cloneDeep, omit, isEqual, keyBy } from "lodash"
-import { KubernetesProvider, KubernetesPluginContext } from "../config"
+import {
+  BaseResource,
+  KubernetesResource,
+  KubernetesServerResource,
+  KubernetesWorkload,
+  SyncableResource,
+} from "../types"
+import { cloneDeep, flatten, isArray, isEqual, isPlainObject, keyBy, mapValues, omit, pickBy } from "lodash"
+import { KubernetesPluginContext, KubernetesProvider } from "../config"
 import { isSubset } from "../../../util/is-subset"
 import { Log } from "../../../logger/log-entry"
 import {
-  V1ReplicationController,
-  V1ReplicaSet,
-  V1Pod,
-  V1PersistentVolumeClaim,
-  V1Service,
-  V1Container,
   KubernetesObject,
+  V1Container,
   V1Job,
+  V1PersistentVolumeClaim,
+  V1Pod,
+  V1ReplicaSet,
+  V1ReplicationController,
+  V1Service,
 } from "@kubernetes/client-node"
-import dedent = require("dedent")
 import { getPods, getResourceKey, hashManifest } from "../util"
 import { checkWorkloadStatus } from "./workload"
 import { checkWorkloadPodStatus } from "./pod"
 import { deline, gardenAnnotationKey, stableStringify } from "../../../util/string"
-import { SyncableResource } from "../types"
 import { ActionMode } from "../../../actions/types"
 import { deepMap } from "../../../util/objects"
-import { DeployState, combineStates } from "../../../types/service"
+import { combineStates, DeployState } from "../../../types/service"
 import { isTruthy, sleep } from "../../../util/util"
 
 export interface ResourceStatus<T extends BaseResource | KubernetesObject = BaseResource> {
   state: DeployState
@@ -174,18 +179,14 @@
   log: Log
   waitForJobs?: boolean
 }) {
-  const handler = objHandlers[manifest.kind]
-
   if (manifest.metadata?.namespace) {
     namespace = manifest.metadata.namespace
   }
 
   let resource: KubernetesServerResource
-  let resourceVersion: number | undefined
 
   try {
     resource = await api.readBySpec({ namespace, manifest, log })
-    resourceVersion = parseInt(resource.metadata.resourceVersion!, 10)
   } catch (err) {
     if (err.statusCode === 404) {
       return { state: <DeployState>"missing", resource: manifest }
@@ -194,15 +195,41 @@
     }
   }
 
-  let status: ResourceStatus
+  return resolveResourceStatus({ api, namespace, resource, log, waitForJobs })
+}
+
+export async function resolveResourceStatus(
+  params: Omit<StatusHandlerParams, "resourceVersion">
+): Promise<ResourceStatus> {
+  const handler = objHandlers[params.resource.kind]
+
   if (handler) {
-    status = await handler({ api, namespace, resource, log, resourceVersion, waitForJobs })
+    const resourceVersion = parseInt(params.resource.metadata.resourceVersion!, 10)
+    return handler({ ...params, resourceVersion })
   } else {
     // if there is no explicit handler to check the status, we assume there's no rollout phase to wait for
-    status = { state: "ready", resource: manifest }
+    return { state: "ready", resource: params.resource }
   }
+}
 
-  return status
+export function resolveResourceStatuses(log: Log, statuses: ResourceStatus[]) {
+  const deployedStates = statuses.map((s) => s.state)
+  const state = combineStates(deployedStates)
+
+  if (state !== "ready") {
+    const descriptions = statuses
+      .filter((s) => s.state !== "ready")
+      .map((s) => `${getResourceKey(s.resource)}: "${s.state}"`)
+      .join("\n")
+
+    log.silly(
+      dedent`
+      Resource(s) with non-ready status found in the cluster:
+
+      ${descriptions}` + "\n"
+    )
+  }
+  return state
 }
 
 interface WaitParams {
@@ -370,10 +397,7 @@
   manifests = flatten(manifests.map((r: any) => (r.apiVersion === "v1" && r.kind === "List" ? r.items : [r])))
 
   // Check if any resources are missing from the cluster.
-  const maybeDeployedObjects = await Bluebird.map(manifests, (resource) =>
-    getDeployedResource(ctx, ctx.provider, resource, log)
-  )
-  const deployedResources = <KubernetesResource[]>maybeDeployedObjects.filter((o) => o !== null)
+  const deployedResources = await getDeployedResources({ ctx, log, manifests })
   const manifestsMap = keyBy(manifests, (m) => getResourceKey(m))
   const manifestKeys = Object.keys(manifestsMap)
   const deployedMap = keyBy(deployedResources, (m) => getResourceKey(m))
@@ -405,24 +429,13 @@
   log.debug(`Getting currently deployed resource statuses...`)
 
   const deployedObjectStatuses: ResourceStatus[] = await Bluebird.map(deployedResources, async (resource) =>
-    checkResourceStatus({ api, namespace, manifest: resource, log })
+    resolveResourceStatus({ api, namespace, resource, log })
   )
 
-  const deployedStates = deployedObjectStatuses.map((s) => s.state)
-  if (deployedStates.find((s) => s !== "ready")) {
-    const descriptions = zip(deployedResources, deployedStates)
-      .filter(([_, s]) => s !== "ready")
-      .map(([o, s]) => `${logDescription(o!)}: "${s}"`)
-      .join("\n")
-
-    log.silly(
-      dedent`
-      Resource(s) with non-ready status found in the cluster:
+  const resolvedState = resolveResourceStatuses(log, deployedObjectStatuses)
 
-      ${descriptions}` + "\n"
-    )
-
-    result.state = combineStates(deployedStates)
+  if (resolvedState !== "ready") {
+    result.state = resolvedState
     return result
   }
 
@@ -595,9 +608,6 @@
   }
 }
 
-/**
- * Fetches matching deployed resources from the cluster for the provided array of manifests.
- */
 export async function getDeployedResources<ResourceKind extends KubernetesObject>({
   ctx,
   manifests,
@@ -606,8 +616,8 @@
   ctx: KubernetesPluginContext
   manifests: KubernetesResource<ResourceKind>[]
   log: Log
-}): Promise<KubernetesResource<ResourceKind>[]> {
-  const maybeDeployedObjects = await Bluebird.map(manifests, async (resource) =>
+}) {
+  const maybeDeployedObjects = await Bluebird.map(manifests, (resource) =>
     getDeployedResource(ctx, ctx.provider, resource, log)
   )
   return maybeDeployedObjects.filter(isTruthy)
Index: core/src/plugins/kubernetes/kubernetes-type/handlers.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/core/src/plugins/kubernetes/kubernetes-type/handlers.ts b/core/src/plugins/kubernetes/kubernetes-type/handlers.ts
--- a/core/src/plugins/kubernetes/kubernetes-type/handlers.ts	(revision 409899599c81409852f08b4de1e4f44acce1e84f)
+++ b/core/src/plugins/kubernetes/kubernetes-type/handlers.ts	(date 1689324546625)
@@ -9,7 +9,7 @@
 import Bluebird from "bluebird"
 import { isEmpty, omit, partition, uniq } from "lodash"
 import type { ModuleActionHandlers } from "../../../plugin/plugin"
-import { ServiceStatus } from "../../../types/service"
+import { DeployState, ForwardablePort, ServiceStatus } from "../../../types/service"
 import { gardenAnnotationKey } from "../../../util/string"
 import { KubeApi } from "../api"
 import type { KubernetesPluginContext } from "../config"
@@ -19,16 +19,28 @@
 import { getActionNamespace, getActionNamespaceStatus } from "../namespace"
 import { getForwardablePorts, killPortForwards } from "../port-forward"
 import { getK8sIngresses } from "../status/ingress"
-import { compareDeployedResources, waitForResources } from "../status/status"
+import {
+  getDeployedResource,
+  resolveResourceStatus,
+  resolveResourceStatuses,
+  ResourceStatus,
+  waitForResources,
+} from "../status/status"
 import type { BaseResource, KubernetesResource, KubernetesServerResource, SyncableResource } from "../types"
-import { convertServiceResource, gardenNamespaceAnnotationValue, getManifests } from "./common"
+import {
+  convertServiceResource,
+  gardenNamespaceAnnotationValue,
+  getManifests,
+  getMetadataManifest,
+  parseMetadataResource,
+} from "./common"
 import { configureKubernetesModule, KubernetesModule } from "./module-config"
 import { configureLocalMode, startServiceInLocalMode } from "../local-mode"
 import type { ExecBuildConfig } from "../../exec/build"
 import type { KubernetesActionConfig, KubernetesDeployAction, KubernetesDeployActionConfig } from "./config"
 import type { DeployActionHandler } from "../../../plugin/action-types"
 import type { ActionLog } from "../../../logger/log-entry"
-import type { Resolved } from "../../../actions/types"
+import type { ActionMode, Resolved } from "../../../actions/types"
 import { deployStateToActionState } from "../../../plugin/handlers/Deploy/get-status"
 
 export const kubernetesHandlers: Partial<ModuleActionHandlers<KubernetesModule>> = {
@@ -169,42 +181,69 @@
     provider,
     skipCreate: true,
   })
-  const namespace = namespaceStatus.namespaceName
+  const defaultNamespace = namespaceStatus.namespaceName
   const api = await KubeApi.factory(log, ctx, k8sCtx.provider)
 
-  // FIXME: We're currently reading the manifests from the module source dir (instead of build dir)
-  // because the build may not have been staged.
-  // This means that manifests added via the `build.dependencies[].copy` field will not be included.
-  const manifests = await getManifests({ ctx, api, log, action, defaultNamespace: namespace, readFromSrcDir: true })
-  const prepareResult = await configureSpecialModesForManifests({
-    ctx: k8sCtx,
-    log,
-    action,
-    manifests,
-  })
-  const preparedManifests = prepareResult.manifests
+  // Note: This is analogous to how we version check Helm charts, i.e. we don't check every resource individually.
+  // Users can always force deploy, much like with Helm Deploys.
+  const metadataManifest = getMetadataManifest(action, defaultNamespace, [])
 
-  let {
-    state,
-    remoteResources,
-    mode: deployedMode,
-  } = await compareDeployedResources({
-    ctx: k8sCtx,
-    api,
-    namespace,
-    manifests: preparedManifests,
-    log,
-  })
+  let deployedMode: ActionMode = "default"
+  let state: DeployState = "ready"
+  let remoteResources: KubernetesResource[] = []
+  let forwardablePorts: ForwardablePort[] | undefined
 
-  // Local mode has its own port-forwarding configuration
-  const forwardablePorts = deployedMode === "local" ? [] : getForwardablePorts(remoteResources, action)
+  const remoteMetadataResource = await getDeployedResource(ctx, provider, metadataManifest, log)
 
-  if (state === "ready") {
-    // Local mode always takes precedence over sync mode
-    if (mode === "local" && spec.localMode && deployedMode !== "local") {
+  if (!remoteMetadataResource) {
+    state = "missing"
+  } else {
+    const deployedMetadata = parseMetadataResource(log, remoteMetadataResource)
+    deployedMode = deployedMetadata.mode
+
+    if (deployedMetadata.resolvedVersion !== action.versionString()) {
+      state = "outdated"
+    } else if (mode === "local" && spec.localMode && deployedMode !== "local") {
       state = "outdated"
     } else if (mode === "sync" && spec.sync?.paths && deployedMode !== "sync") {
       state = "outdated"
+    } else if (mode === "default" && deployedMode !== mode) {
+      state = "outdated"
+    }
+
+    const manifestMetadata = Object.values(deployedMetadata.manifestMetadata)
+
+    if (manifestMetadata.length > 0) {
+      try {
+        const maybeDeployedResources = await Bluebird.map(manifestMetadata, async (m) => {
+          return [m, await api.readOrNull({ log, ...m })]
+        })
+
+        const statuses: ResourceStatus[] = await Bluebird.map(maybeDeployedResources, async ([m, resource]) => {
+          if (!resource) {
+            return {
+              state: "missing" as const,
+              resource: { apiVersion: m.apiVersion, kind: m.kind, metadata: { name: m.name, namespace: m.namespace } },
+            }
+          }
+          remoteResources.push(resource)
+          return resolveResourceStatus({ api, namespace: defaultNamespace, resource, log })
+        })
+
+        state = resolveResourceStatuses(log, statuses)
+      } catch (error) {
+        log.debug({ msg: `Failed querying for remote resources: ${error.message}`, error })
+        state = "unknown"
+      }
+    }
+
+    // Note: Local mode has its own port-forwarding configuration
+    if (deployedMode !== "local" && remoteResources && remoteResources.length > 0) {
+      try {
+        forwardablePorts = getForwardablePorts(remoteResources, action)
+      } catch (error) {
+        log.debug({ msg: `Unable to extract forwardable ports: ${error.message}`, error })
+      }
     }
   }
 
@@ -216,7 +255,7 @@
       version: state === "ready" ? action.versionString() : undefined,
       detail: { remoteResources },
       mode: deployedMode,
-      ingresses: getK8sIngresses(remoteResources),
+      ingresses: getK8sIngresses(remoteResources || []),
     },
     // TODO-0.13.1
     outputs: {},
@@ -397,7 +436,7 @@
   }
 
   return {
-    state: "ready",
+    state: "not-ready",
     detail: status,
     outputs: {},
   }
Index: core/src/plugins/kubernetes/kubernetes-type/common.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/core/src/plugins/kubernetes/kubernetes-type/common.ts b/core/src/plugins/kubernetes/kubernetes-type/common.ts
--- a/core/src/plugins/kubernetes/kubernetes-type/common.ts	(revision 409899599c81409852f08b4de1e4f44acce1e84f)
+++ b/core/src/plugins/kubernetes/kubernetes-type/common.ts	(date 1689324546621)
@@ -9,13 +9,13 @@
 import { resolve } from "path"
 import { readFile } from "fs-extra"
 import Bluebird from "bluebird"
-import { flatten, set } from "lodash"
+import { flatten, keyBy, set } from "lodash"
 import { loadAll } from "js-yaml"
 
 import { KubernetesModule } from "./module-config"
 import { KubernetesResource } from "../types"
 import { KubeApi } from "../api"
-import { gardenAnnotationKey } from "../../../util/string"
+import { gardenAnnotationKey, stableStringify } from "../../../util/string"
 import { Log } from "../../../logger/log-entry"
 import { PluginContext } from "../../../plugin-context"
 import { ConfigurationError, PluginError } from "../../../exceptions"
@@ -24,9 +24,10 @@
 import { KubernetesDeployAction } from "./config"
 import { CommonRunParams } from "../../../plugin/handlers/Run/run"
 import { runAndCopy } from "../run"
-import { getTargetResource, getResourcePodSpec, getResourceContainer, makePodName } from "../util"
-import { Resolved } from "../../../actions/types"
+import { getTargetResource, getResourcePodSpec, getResourceContainer, makePodName, getResourceKey } from "../util"
+import { ActionMode, Resolved } from "../../../actions/types"
 import { KubernetesPodRunAction, KubernetesPodTestAction } from "./kubernetes-pod"
+import { V1ConfigMap } from "@kubernetes/client-node"
 import { glob } from "glob"
 
 /**
@@ -39,16 +40,14 @@
   log,
   action,
   defaultNamespace,
-  readFromSrcDir = false,
 }: {
   ctx: PluginContext
   api: KubeApi
   log: Log
   action: Resolved<KubernetesDeployAction | KubernetesPodRunAction | KubernetesPodTestAction>
   defaultNamespace: string
-  readFromSrcDir?: boolean
 }): Promise<KubernetesResource[]> {
-  const rawManifests = (await readManifests(ctx, action, log, readFromSrcDir)) as KubernetesResource[]
+  const rawManifests = (await readManifests(ctx, action, log)) as KubernetesResource[]
 
   // remove *List objects
   const manifests = rawManifests.flatMap((manifest) => {
@@ -71,6 +70,11 @@
     return manifest
   })
 
+  if (action.kind === "Deploy") {
+    // Add metadata ConfigMap to aid quick status check
+    manifests.push(getMetadataManifest(action, defaultNamespace, manifests))
+  }
+
   return Bluebird.map(manifests, async (manifest) => {
     // Ensure a namespace is set, if not already set, and if required by the resource type
     if (!manifest.metadata?.namespace) {
@@ -106,26 +110,80 @@
   })
 }
 
+export interface ManifestMetadata {
+  key: string
+  apiVersion: string
+  kind: string
+  name: string
+  namespace: string
+}
+
+export interface ParsedMetadataManifestData {
+  resolvedVersion: string
+  mode: ActionMode
+  manifestMetadata: { [key: string]: ManifestMetadata }
+}
+
+export function getMetadataManifest(
+  action: Resolved<KubernetesDeployAction>,
+  defaultNamespace: string,
+  manifests: KubernetesResource[]
+): KubernetesResource<V1ConfigMap> {
+  const manifestMetadata: ManifestMetadata[] = manifests.map((m) => ({
+    key: getResourceKey(m),
+    apiVersion: m.apiVersion,
+    kind: m.kind,
+    name: m.metadata.name,
+    namespace: m.metadata.namespace || defaultNamespace,
+  }))
+
+  return {
+    apiVersion: "v1",
+    kind: "ConfigMap",
+    metadata: {
+      name: `garden-meta-${action.kind.toLowerCase()}-${action.name}`,
+    },
+    data: {
+      resolvedVersion: action.versionString(),
+      mode: action.mode(),
+      manifestMetadata: stableStringify(keyBy(manifestMetadata, "key")),
+    },
+  }
+}
+
+export function parseMetadataResource(log: Log, resource: KubernetesResource<V1ConfigMap>): ParsedMetadataManifestData {
+  // TODO: validate schema here
+  const output: ParsedMetadataManifestData = {
+    resolvedVersion: resource.data?.resolvedVersion || "",
+    mode: (resource.data?.mode || "default") as ActionMode,
+    manifestMetadata: {},
+  }
+
+  const manifestMetadata = resource.data?.manifestMetadata
+
+  if (manifestMetadata) {
+    try {
+      // TODO: validate by schema
+      output.manifestMetadata = JSON.parse(manifestMetadata)
+    } catch (error) {
+      log.debug({ msg: `Failed querying for remote resources: ${error.message}`, error })
+    }
+  }
+
+  return output
+}
+
 const disallowedKustomizeArgs = ["-o", "--output", "-h", "--help"]
 
 /**
  * Read the manifests from the module config, as well as any referenced files in the config.
- *
- * @param module The kubernetes module to read manifests for.
- * @param readFromSrcDir Whether or not to read the manifests from the module build dir or from the module source dir.
- * In general we want to read from the build dir to ensure that manifests added via the `build.dependencies[].copy`
- * field will be included. However, in some cases, e.g. when getting the service status, we can't be certain that
- * the build has been staged and we therefore read the manifests from the source.
- *
- * TODO: Remove this once we're checking for kubernetes module service statuses with version hashes.
  */
 export async function readManifests(
   ctx: PluginContext,
   action: Resolved<KubernetesDeployAction | KubernetesPodRunAction | KubernetesPodTestAction>,
-  log: Log,
-  readFromSrcDir = false
+  log: Log
 ) {
-  const manifestPath = readFromSrcDir ? action.basePath() : action.getBuildPath()
+  const manifestPath = action.getBuildPath()
 
   const spec = action.getSpec()
 
Index: core/test/integ/src/plugins/kubernetes/kubernetes-type/handlers.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/core/test/integ/src/plugins/kubernetes/kubernetes-type/handlers.ts b/core/test/integ/src/plugins/kubernetes/kubernetes-type/handlers.ts
--- a/core/test/integ/src/plugins/kubernetes/kubernetes-type/handlers.ts	(revision 409899599c81409852f08b4de1e4f44acce1e84f)
+++ b/core/test/integ/src/plugins/kubernetes/kubernetes-type/handlers.ts	(date 1689324546649)
@@ -14,7 +14,11 @@
 import { TestGarden } from "../../../../../helpers"
 import { getKubernetesTestGarden } from "./common"
 import { DeployTask } from "../../../../../../src/tasks/deploy"
-import { getManifests } from "../../../../../../src/plugins/kubernetes/kubernetes-type/common"
+import {
+  getManifests,
+  getMetadataManifest,
+  parseMetadataResource,
+} from "../../../../../../src/plugins/kubernetes/kubernetes-type/common"
 import { KubeApi } from "../../../../../../src/plugins/kubernetes/api"
 import { ActionLog, createActionLog, Log } from "../../../../../../src/logger/log-entry"
 import { KubernetesPluginContext, KubernetesProvider } from "../../../../../../src/plugins/kubernetes/config"
@@ -127,7 +131,143 @@
     }
   })
 
-  describe("getServiceStatus", () => {
+  describe("getKubernetesDeployStatus", () => {
+    it("returns ready status and correct detail for a ready deployment", async () => {
+      const graph = await garden.getConfigGraph({ log: garden.log, emit: false })
+      const action = await garden.resolveAction<KubernetesDeployAction>({
+        action: graph.getDeploy("module-simple"),
+        log: garden.log,
+        graph,
+      })
+      const deployParams = {
+        ctx,
+        log: actionLog,
+        action,
+        force: false,
+      }
+
+      await kubernetesDeploy(deployParams)
+
+      const status = await getKubernetesDeployStatus(deployParams)
+      expect(status.state).to.equal("ready")
+
+      expect(status.detail?.mode).to.equal("default")
+      expect(status.detail?.forwardablePorts).to.eql([
+        {
+          name: undefined,
+          protocol: "TCP",
+          targetName: "Deployment/busybox-deployment",
+          targetPort: 80,
+        },
+      ])
+
+      const remoteResources = status.detail?.detail?.remoteResources
+      expect(remoteResources).to.exist
+      expect(remoteResources.length).to.equal(1)
+      expect(remoteResources[0].kind).to.equal("Deployment")
+    })
+
+    it("should return missing status when metadata ConfigMap is missing", async () => {
+      const graph = await garden.getConfigGraph({ log: garden.log, emit: false })
+      const action = await garden.resolveAction<KubernetesDeployAction>({
+        action: graph.getDeploy("module-simple"),
+        log: garden.log,
+        graph,
+      })
+      const deployParams = {
+        ctx,
+        log: actionLog,
+        action,
+        force: false,
+      }
+
+      await kubernetesDeploy(deployParams)
+
+      const namespace = await getActionNamespace({
+        ctx,
+        log,
+        action,
+        provider: ctx.provider,
+        skipCreate: true,
+      })
+
+      const metadataManifest = getMetadataManifest(action, namespace, [])
+
+      await api.deleteBySpec({ log, namespace, manifest: metadataManifest })
+
+      const status = await getKubernetesDeployStatus(deployParams)
+      expect(status.state).to.equal("not-ready")
+      expect(status.detail?.state).to.equal("missing")
+    })
+
+    it("should return outdated status when metadata ConfigMap has different version", async () => {
+      const graph = await garden.getConfigGraph({ log: garden.log, emit: false })
+      const action = await garden.resolveAction<KubernetesDeployAction>({
+        action: graph.getDeploy("module-simple"),
+        log: garden.log,
+        graph,
+      })
+      const deployParams = {
+        ctx,
+        log: actionLog,
+        action,
+        force: false,
+      }
+
+      await kubernetesDeploy(deployParams)
+
+      const namespace = await getActionNamespace({
+        ctx,
+        log,
+        action,
+        provider: ctx.provider,
+        skipCreate: true,
+      })
+
+      const metadataManifest = getMetadataManifest(action, namespace, [])
+      metadataManifest.data!.resolvedVersion = "v-foo"
+
+      await api.replace({ log, namespace, resource: metadataManifest })
+
+      const status = await getKubernetesDeployStatus(deployParams)
+      expect(status.state).to.equal("not-ready")
+      expect(status.detail?.state).to.equal("outdated")
+    })
+
+    it("should return outdated status when metadata ConfigMap has different mode", async () => {
+      const graph = await garden.getConfigGraph({ log: garden.log, emit: false })
+      const action = await garden.resolveAction<KubernetesDeployAction>({
+        action: graph.getDeploy("module-simple"),
+        log: garden.log,
+        graph,
+      })
+      const deployParams = {
+        ctx,
+        log: actionLog,
+        action,
+        force: false,
+      }
+
+      await kubernetesDeploy(deployParams)
+
+      const namespace = await getActionNamespace({
+        ctx,
+        log,
+        action,
+        provider: ctx.provider,
+        skipCreate: true,
+      })
+
+      const metadataManifest = getMetadataManifest(action, namespace, [])
+      metadataManifest.data!.mode = "sync"
+
+      await api.replace({ log, namespace, resource: metadataManifest })
+
+      const status = await getKubernetesDeployStatus(deployParams)
+      expect(status.state).to.equal("not-ready")
+      expect(status.detail?.state).to.equal("outdated")
+    })
+
     it("should return not-ready status for a manifest with a missing resource type", async () => {
       const graph = await garden.getConfigGraph({ log: garden.log, emit: false })
       const action = await garden.resolveAction<KubernetesDeployAction>({
@@ -183,11 +323,30 @@
         log,
         action: resolvedAction,
         defaultNamespace: namespace,
-        readFromSrcDir: true,
       })
       return { deployParams, manifests }
     }
 
+    it("gets the correct manifests when `build` is set", async () => {
+      const graph = await garden.getConfigGraph({ log: garden.log, emit: false })
+      const action = graph.getDeploy("with-build-action")
+      const deployParams = {
+        ctx,
+        log: actionLog,
+        action: await garden.resolveAction<KubernetesDeployAction>({ action, log: garden.log, graph }),
+        force: false,
+      }
+
+      const status = await kubernetesDeploy(deployParams)
+      expect(status.state).to.eql("ready")
+
+      const remoteResources = status.detail?.detail?.remoteResources
+      expect(remoteResources).to.exist
+      expect(remoteResources.length).to.equal(1)
+      expect(remoteResources[0].kind).to.equal("Deployment")
+      expect(remoteResources[0].metadata?.name).to.equal("busybox-deployment")
+    })
+
     it("should successfully deploy when serviceResource doesn't have a containerModule", async () => {
       const graph = await garden.getConfigGraph({ log: garden.log, emit: false })
       const action = graph.getDeploy("module-simple")
@@ -207,6 +366,48 @@
       expect(namespaceStatus!.namespaceName).to.eql("kubernetes-type-test-default")
     })
 
+    it("creates a metadata ConfigMap describing what was last deployed", async () => {
+      const graph = await garden.getConfigGraph({ log: garden.log, emit: false })
+      const action = graph.getDeploy("module-simple")
+      const resolvedAction = await garden.resolveAction<KubernetesDeployAction>({ action, log: garden.log, graph })
+      const deployParams = {
+        ctx,
+        log: actionLog,
+        action: resolvedAction,
+        force: false,
+      }
+      const status = await kubernetesDeploy(deployParams)
+      expect(status.state).to.eql("ready")
+
+      const namespace = await getActionNamespace({
+        ctx,
+        log,
+        action: resolvedAction,
+        provider: ctx.provider,
+        skipCreate: true,
+      })
+
+      const emptyManifest = getMetadataManifest(resolvedAction, namespace, [])
+
+      const metadataResource = await api.readBySpec({ log, namespace, manifest: emptyManifest })
+
+      expect(metadataResource).to.exist
+
+      const parsedMetadata = parseMetadataResource(log, metadataResource)
+
+      expect(parsedMetadata.resolvedVersion).to.equal(resolvedAction.versionString())
+      expect(parsedMetadata.mode).to.equal("default")
+      expect(parsedMetadata.manifestMetadata).to.eql({
+        "Deployment/busybox-deployment": {
+          apiVersion: "apps/v1",
+          key: "Deployment/busybox-deployment",
+          kind: "Deployment",
+          name: "busybox-deployment",
+          namespace: "kubernetes-type-test-default",
+        },
+      })
+    })
+
     it("should toggle sync mode", async () => {
       const syncData = await getTestData("with-source-module", {
         sync: ["deploy.with-source-module"],
@@ -322,8 +523,9 @@
     it("should successfully deploy List manifest kinds", async () => {
       const configMapList = await getTestData("config-map-list", {})
 
-      // this should be 3, and not 1, as we transform *List objects to separate manifests
-      expect(configMapList.manifests.length).to.be.equal(3)
+      // this should be 4, and not 2 (including the metadata manifest),
+      // as we transform *List objects to separate manifests
+      expect(configMapList.manifests.length).to.be.equal(4)
 
       // test successful deploy
       await kubernetesDeploy(configMapList.deployParams)
Index: core/test/integ/src/plugins/kubernetes/kubernetes-type/common.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/core/test/integ/src/plugins/kubernetes/kubernetes-type/common.ts b/core/test/integ/src/plugins/kubernetes/kubernetes-type/common.ts
--- a/core/test/integ/src/plugins/kubernetes/kubernetes-type/common.ts	(revision 409899599c81409852f08b4de1e4f44acce1e84f)
+++ b/core/test/integ/src/plugins/kubernetes/kubernetes-type/common.ts	(date 1689324546646)
@@ -60,7 +60,7 @@
       action["_config"].spec.kustomize!.extraArgs = ["--output", "foo"]
 
       await expectError(
-        () => readManifests(ctx, action, garden.log, false),
+        () => readManifests(ctx, action, garden.log),
         (err) => expect(err.message).to.equal(expectedErr)
       )
     })
@@ -69,7 +69,7 @@
       action["_config"].spec.kustomize!.extraArgs = ["-o", "foo"]
 
       await expectError(
-        () => readManifests(ctx, action, garden.log, false),
+        () => readManifests(ctx, action, garden.log),
         (err) => expect(err.message).to.equal(expectedErr)
       )
     })
@@ -78,7 +78,7 @@
       action["_config"].spec.kustomize!.extraArgs = ["-h"]
 
       await expectError(
-        () => readManifests(ctx, action, garden.log, false),
+        () => readManifests(ctx, action, garden.log),
         (err) => expect(err.message).to.equal(expectedErr)
       )
     })
@@ -87,20 +87,20 @@
       action["_config"].spec.kustomize!.extraArgs = ["--help"]
 
       await expectError(
-        () => readManifests(ctx, action, garden.log, false),
+        () => readManifests(ctx, action, garden.log),
         (err) => expect(err.message).to.equal(expectedErr)
       )
     })
 
     it("runs kustomize build in the given path", async () => {
-      const result = await readManifests(ctx, action, garden.log, true)
+      const result = await readManifests(ctx, action, garden.log)
       const kinds = result.map((r) => r.kind)
       expect(kinds).to.have.members(["ConfigMap", "Service", "Deployment"])
     })
 
     it("adds extraArgs if specified to the build command", async () => {
       action["_config"].spec.kustomize!.extraArgs = ["--reorder", "none"]
-      const result = await readManifests(ctx, action, garden.log, true)
+      const result = await readManifests(ctx, action, garden.log)
       const kinds = result.map((r) => r.kind)
       expect(kinds).to.eql(["Deployment", "Service", "ConfigMap"])
     })
```

</details>

### Additional context

–

### Your environment

* **OS:** macOS
* **How I'm running Kubernetes:** Docker desktop

`garden version` 0.13.0

### LINKED ISSUE #4931: 0.13: [Bug]: Failure to use specified `build` action in Kubernetes `deploy` action
### Current Behavior

When specifying a `build` action under the [build key](https://docs.garden.io/reference/action-types/deploy/kubernetes#build), the commands ran do not utilize the `build` directory, but instead use the source directory.

### Expected behavior

When I specify the build action, the build directory should be used.

### Reproducible example

in this example, we generate an example Kubernetes manifest and attempt to deploy it.

First, run this command:
`helm repo add examples https://helm.github.io/examples`

Next, use these modules:
```yaml
type: exec
kind: "Build"
name: infra-python-starter-generate-config
description: Istio, DNS, CA certs, and other config
include:
  - asdf
spec:
  command:
    - helm
    - template
    - --name-template=${project.name}
    - --output-dir=${this.buildPath}
    - examples/hello-world
---
type: kubernetes
name: infra-python-starter-apply-config
kind: "Deploy"
dependencies:
  - build.infra-python-starter-generate-config
build: infra-python-starter-generate-config
description: Istio, DNS, CA certs, and other config
spec:
  files:
    - ./hello-world/templates/*.yaml
  timeout: 2400
  namespace: default
---
```

Building works in creating a `.garden/build/infra-python-starter-generate-config/hello-world/templates` folder,
however, deployment fails with this error:
```bash
Invalid manifest file path(s) in Deploy action 'infra-python-starter-apply-config'. Cannot find manifest file(s) at ./hello-world/templates/ in /Users/aron/Workspace/Pachama/infra/infra-python-starter directory.
```

### Workaround

A workaround would be to create a `exec` `deploy` action, which does have access to the `build` folder contents.
But, this doesn't give us the natively supported features of the `kubernetes` action, like syncing folders to the deployment, cleanup, or status commands.
Like this:
```yaml
type: exec
kind: "Build"
name: infra-python-starter-generate-config
description: Istio, DNS, CA certs, and other config
include:
  - asdf
spec:
  command:
    - helm
    - template
    - --name-template=${project.name}
    - --output-dir=${this.buildPath}
    - examples/hello-world
---
type: exec
name: infra-python-starter-apply-config
kind: "Deploy"
dependencies:
  - build.infra-python-starter-generate-config
build: infra-python-starter-generate-config
include:
  - .garden/build/infra-python-starter-generate-config
description: Istio, DNS, CA certs, and other config
spec:
  deployCommand:
    - kubectl
    - apply
    - -f
    - hello-world/templates
---
```

### Suggested solution(s)

There appears to be an issue with the 0.13 Kubernetes deploy action source code.

### Additional context

This is a blocker for migrating to version 13

### Your environment

* **OS:** macOS
* **Garden version:** 0.13.10

vvagaytsev:

Possibly fixes #4505
Fixes #4506
Fixes #4931

There are many commits, but those are incremental ones.
Recommend to review commit-by-commit.

Most part of the code changes is taken by refactoring.


core/src/plugins/kubernetes/kubernetes-type/handlers.ts
```diff
@@ -169,42 +181,71 @@ export const getKubernetesDeployStatus: DeployActionHandler<"getStatus", Kuberne
...
+              resource: { apiVersion: m.apiVersion, kind: m.kind, metadata: { name: m.name, namespace: m.namespace } },
+            }
+          }
+          remoteResources.push(resource)
+          return resolveResourceStatus({ api, namespace: defaultNamespace, resource, log })
+        })
+
+        if (state !== "outdated") {
+          state = resolveResourceStatuses(log, statuses)
+        }
```
vvagaytsev, line 235: [This quickfix](https://github.com/garden-io/garden/pull/4846/commits/53fdc0219a520c89baa8ad6b41abe3be47283ef7) avoids resetting the previously calculated `"outdated"` status to `"ready"`.

That can happen because some k8s resource kinds [do not have handlers](https://github.com/garden-io/garden/pull/4846/files#diff-24f998c4e8f6dfb59d6ab3922dcc7a3ce8a9cdab363234e51f794537b36611f9R211-R218) and return the `"ready"` status by default. Before https://github.com/garden-io/garden/pull/4846/commits/fa1bb4d545447fb3fee0255849141a8af18a98f7 we generated manifests from the action config and called [`compareDeployedResources`](https://github.com/garden-io/garden/blob/main/core/src/plugins/kubernetes/status/status.ts#L356).

Now we rely on the status handlers defined in [`objHandlers`](https://github.com/garden-io/garden/blob/main/core/src/plugins/kubernetes/status/status.ts#L73), The lack of the status handler for `Secret` kind of k8s resource caused the regression described in #4811.

Probably, the proper fix would be to implement [status handler for `Secret` k8s resource](https://github.com/garden-io/garden/pull/4846/files#diff-24f998c4e8f6dfb59d6ab3922dcc7a3ce8a9cdab363234e51f794537b36611f9R105-R110), see https://github.com/garden-io/garden/pull/4846/commits/0ff268c9fce5740dfe00f98ba3abd96b36468854.

That requires some extension of the `StatusHandlerParams` interface and some extra calculations in the `Secret` handler because we need to compare the `Secret` manifest generated from the local config against the deployed resource.

So, it looks like we have to preserve the old scenario to detect secrets update:
1. Make the manifest out of the action config
2. Get deployed resource
3. Compare the manifest against the deployed resource

@edvald @garden-io/core wdyt?

core/test/integ/src/plugins/kubernetes/kubernetes-type/common.ts
```diff
@@ -89,20 +89,20 @@ describe("readManifests", () => {
...
     it("runs kustomize build in the given path", async () => {
-      const result = await readManifests(ctx, action, garden.log, true)
+      const result = await readManifests(ctx, action, garden.log)
       const kinds = result.map((r) => r.kind)
       expect(kinds).to.have.members(["ConfigMap", "Service", "Deployment"])
     })
 
     it("adds extraArgs if specified to the build command", async () => {
       action["_config"].spec.kustomize!.extraArgs = ["--reorder", "none"]
-      const result = await readManifests(ctx, action, garden.log, true)
```
edvald, line 105: Bit of a side-note, but this function looks primed for destructured params.

vvagaytsev: Refactored it in 05827818a92be7658c8dbade3fedb4986c91a5a4

vvagaytsev:
I need a pre-review for this, I'll add some tests tomorrow.

vvagaytsev:
@edvald @thsig can you take a look, please?

thsig (commented):
Good stuff!

See my comment about going back to using `compareDeployedResources` (but passing in the manifests from the metadata ConfigMap for comparing against the deployed resources).

If we think that's a good idea, we could either merge this and implement that approach in a follow-up PR, or do it in this one before merging.

core/src/plugins/kubernetes/kubernetes-type/handlers.ts
thsig: Would also be great to add an integ test case that verifies that updating a secret to change its value does indeed make the Deploy status `outdated`.

core/src/plugins/kubernetes/status/status.ts
```diff
@@ -95,6 +101,12 @@ const objHandlers: { [kind: string]: StatusHandler } = {
     return checkWorkloadPodStatus(resource, await getPods(api, namespace, resource.spec!.selector!))
   },
 
+  Secret: async (params: StatusHandlerParams<V1Secret>) => {
+    const { resource } = params
+    // TODO: compare the `resource` against the manifest generated from the action?
+    return { state: "ready", resource }
+  },
+
```
edvald, line 109: I'm not sure we should have this in here. This is going to break certain cases for sure, and I'm unclear on why we need a specific handler for the Secret kind.

thsig: This is essentially the same as the default handler used in `resolveResourceStatus`, which also always returns a `ready` state.

https://github.com/garden-io/garden/pull/4846/files/c7c125601e326ad10649255994b0e2de0977376c#diff-24f998c4e8f6dfb59d6ab3922dcc7a3ce8a9cdab363234e51f794537b36611f9R219

The `objHandlers` map looks like it was designed for waiting for the rollout status of a deploy that's in-progress. For that use-case, it makes sense to return `ready` when it's not clear how to wait for the resource to fully come up.

However, that does mean that `ready` states can be returned for outdated resources when these handlers are repurposed for status checking.

I didn't closely read the original PR for this, but are we sure we want to move away from `compareDeployedResources`? In an earlier comment, I suggested simply passing the manifests from the metadata ConfigMap to `compareDeployedResources` (for comparing against the remotely deployed ones).

Or is that approach not viable for some reason?

vvagaytsev: @thsig right, the current `Secret` handler stub behaves exactly the same as the old code.

The [quickfix](https://github.com/garden-io/garden/pull/4846/commits/63581f469eece33f0f64637d161350d990f55405) was made to ensure that we'd never fall through the `"outdated"` state coming from the version hash comparison.

So, if the version hashes differ, we would never reach any `objHandlers`. It looks like for the `kubernetes` type we only allow sub-paths for `files` and `kustomize` config options. Thus, in theory, we should not get into a case when secrets are stored in some external files that are not taken into account while the version hash calculation.

We relaxed path validation requirements for `helm` type in https://github.com/garden-io/garden/pull/3445, but not for `kubernetes` type.

So, it should be safe to rely on the [quickfix](https://github.com/garden-io/garden/pull/4846/commits/63581f469eece33f0f64637d161350d990f55405) because it should ensure that `"outdated"` state will never cause any `objHandler` invocation.

Or am I missing something? Can you see any way to change secrets without the action's version hash change?

edvald (dismissed):

core/src/plugins/kubernetes/status/status.ts
thsig: I think the TL;DR is that we should write a few more integ test cases to express & test our expectations for how this logic should handle the various cases we're discussing here.

CC @shumailxyz @vvagaytsev

stefreak:
Who is working on the necessary changes in the Garden Cloud AEC "pause" feature? This definitely impacts Garden cloud and that should be considered before merging or releasing this. See also https://github.com/garden-io/garden/pull/4516#discussion_r1214604254

core/test/integ/src/plugins/kubernetes/kubernetes-type/handlers.ts
```diff
@@ -128,7 +132,143 @@ describe("kubernetes-type handlers", () => {
     }
   })
 
-  describe("getServiceStatus", () => {
+  describe("getKubernetesDeployStatus", () => {
```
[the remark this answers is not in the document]
vvagaytsev: Oh, have no calls to `deleteKubernetesDeploy` handler in our tests at all. We should definitely add some tests for that handler to ensure that no dirty state is left in the target runtime.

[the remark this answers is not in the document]
vvagaytsev: Ah, we're doing some tests over the `DeleteDeployTask`.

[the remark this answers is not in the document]
vvagaytsev: Added a test case in adf847a400d9ab48488fa5f3783e9fc6c26073e0

core/src/plugins/kubernetes/status/status.ts
vvagaytsev: Actually, that's not necessarily a `Secret` resource that can reproduce the original issue. The bug was in the ability to overwrite an already calculated `outdated` state with `ready`.

We do not need a dedicated `objHandler` for `Secret` resource. I removed that commit from the changeset.

Also, I refactored the code in `getKubernetesDeployStatus` to have less local mutable state and to be more linear and straightforward.

Now I only need to add a test case that will cover the original bug. We have one called `should return outdated status when metadata ConfigMap has different version`. But, it doesn't hit the code block that examines the remote resources, because in the test we use an empty metadata object in the metadata `ConfigMap`.

I'm gonna try to change the test data to have something in the config map metadata.

core/src/plugins/kubernetes/kubernetes-type/handlers.ts
vvagaytsev: I amended an existing one in 27996da8efe243fb2e63921c60227c95efff51cd

core/src/plugins/kubernetes/status/status.ts
vvagaytsev: Fixed an existing test in 27996da8efe243fb2e63921c60227c95efff51cd

shumailxyz (approved):
Awesome, thanks @vvagaytsev. Let's get this merged 💯

Issue #4833, opened 2023-07-13
0.13: [Bug]: ModuleTemplate - Module is not able to find the build dependency


ishantagarwal-x213235:
## Garden Bonsai (0.13) Bug - Ignore build.dependencies mentioned in pulumi module

Garden is not able to find the build dependency mentioned in ModuleTemplate. Below is the configuration

```
kind: ModuleTemplate
name: spring-service-module-template
inputsSchemaPath: rest.schema.json
modules:
  - type: pulumi
    name: ${parent.name}-service # ${var.test}-${parent.name} #${parent.name}-module
    description: common to all REST services
    root: ${inputs.deploymentFolder}/pulumi
    stack: "${environment.name}"

    variables:
      excludeDefault: ["${parent.name}"]
      pulumiProjectName: ${var.gardenProjectName}-${var.projectname}-${parent.name}

    createStack: true
    autoApply: true

    deployFromPreview: ${var.withPlan}

    pulumiVarfiles: 
       - "${local.projectPath}/deployment-common/pulumi/config/global.yaml"
       - "${inputs.deploymentFolder}/config/common.yaml"
       - "${inputs.deploymentFolder}/config/${environment.name}.yaml"

      
    generateFiles:
      - sourcePath: ../pulumi/Pulumi.yaml
        targetPath: ${inputs.deploymentFolder}/pulumi/Pulumi.yaml
      - sourcePath: ../pulumi/index.ts
        targetPath: ${inputs.deploymentFolder}/pulumi/default-index.ts
    build:
      dependencies: [x-publish-image]


  - type: jib-container
    name: x-publish-image
    description: Backend service

    varfile: "dependencies.yaml"

    include: 
      - ./**/src/**/*
      - pom.xml
      - ./**/pom.xml
    image: ${parent.name}
    extraFlags: 
    - -Djib.useOnlyProjectCache=false
    - "-Djib.from.image=${var.dockerRepo}/${inputs.jdkVersion == '11' ? var.baseImageJava11 : var.baseImageJava8}"
    - "-nsu"
    - "-Dtransitive=false"

    build: 
      jdkVersion: ${inputs.jdkVersion} 
      projectType: maven
      dependencies: ${var.dependencies}?
      mavenPhases: [ "clean", "package" ]
```

When I trigger the command ```garden deploy x-service```

**Expected Output**
garden first execute the x-publish-image and then it should execute the x-service.

**Actual Output**
garden only execute the x-service and ignore the build-dependency

[...]

PR #4848, opened 2023-07-18
fix: properly handle build deps for pulumi and tf modules

Orzelius:
fixes: https://github.com/garden-io/garden/issues/4833

### DIFF
--- a/core/test/unit/src/config/render-template.ts
+++ b/core/test/unit/src/config/render-template.ts
@@ -80,14 +80,14 @@ describe("config templates", () => {
       })
     })
 
-    it("defaults to an empty object schema for inputs", async () => {
+    it("defaults to an object with any properties for schema", async () => {
       const config: ConfigTemplateResource = {
         ...defaults,
       }
       const resolved = await resolveConfigTemplate(garden, config)
       expect((<any>resolved.inputsSchema)._rules[0].args.jsonSchema.schema).to.eql({
         type: "object",
-        additionalProperties: false,
+        additionalProperties: true,
       })
     })
 
--- a/plugins/pulumi/index.ts
+++ b/plugins/pulumi/index.ts
@@ -103,7 +103,7 @@ export const gardenPlugin = () =>
           configure: configurePulumiModule,
 
           async convert(params: ConvertModuleParams<PulumiModule>) {
-            const { module, dummyBuild, prepareRuntimeDependencies } = params
+            const { module, dummyBuild, convertBuildDependency, prepareRuntimeDependencies } = params
             const actions: (ExecBuildConfig | PulumiDeployConfig)[] = []
 
             if (dummyBuild) {
@@ -115,7 +115,12 @@ export const gardenPlugin = () =>
               type: "pulumi",
               name: module.name,
               ...params.baseFields,
-              dependencies: prepareRuntimeDependencies(module.spec.dependencies, dummyBuild),
+
+              build: dummyBuild?.name,
+              dependencies: [
+                ...module.build.dependencies.map(convertBuildDependency),
+                ...prepareRuntimeDependencies(module.spec.dependencies, dummyBuild),
+              ],
 
               timeout: defaultPulumiTimeoutSec,
               spec: {
--- a/plugins/terraform/index.ts
+++ b/plugins/terraform/index.ts
@@ -188,7 +188,7 @@ export const gardenPlugin = () =>
         needsBuild: false,
         handlers: {
           async convert(params: ConvertModuleParams<TerraformModule>) {
-            const { module, dummyBuild, prepareRuntimeDependencies } = params
+            const { module, dummyBuild, convertBuildDependency, prepareRuntimeDependencies } = params
             const actions: (ExecBuildConfig | TerraformDeployConfig)[] = []
 
             if (dummyBuild) {
@@ -202,7 +202,10 @@ export const gardenPlugin = () =>
               ...params.baseFields,
 
               build: dummyBuild?.name,
-              dependencies: prepareRuntimeDependencies(module.spec.dependencies, dummyBuild),
+              dependencies: [
+                ...module.build.dependencies.map(convertBuildDependency),
+                ...prepareRuntimeDependencies(module.spec.dependencies, dummyBuild),
+              ],
 
               timeout: defaultTerraformTimeoutSec,
               spec: {
--- a/core/src/config/config-template.ts
+++ b/core/src/config/config-template.ts
@@ -75,10 +75,10 @@ export async function resolveConfigTemplate(
   })
 
   // Read and validate the JSON schema, if specified
-  // -> default to object with no properties
+  // -> default to any object
   let inputsJsonSchema = {
     type: "object",
-    additionalProperties: false,
+    additionalProperties: true,
   }
 
   const configDir = configPath ? dirname(configPath) : resource.internal.basePath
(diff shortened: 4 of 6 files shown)


Orzelius:
Didn't have time to fix the test, but you can do so by changing it to
```ts
    it("defaults to an object with any properties for schema", async () => {
      const config: ConfigTemplateResource = {
        ...defaults,
      }
      const resolved = await resolveConfigTemplate(garden, config)
      expect((<any>resolved.inputsSchema)._rules[0].args.jsonSchema.schema).to.eql({
        type: "object",
        additionalProperties: true,
      })
    })
```

core/test/unit/src/config/render-template.ts
```diff
@@ -87,7 +87,7 @@ describe("config templates", () => {
       const resolved = await resolveConfigTemplate(garden, config)
       expect((<any>resolved.inputsSchema)._rules[0].args.jsonSchema.schema).to.eql({
         type: "object",
-        additionalProperties: false,
+        additionalProperties: true,
```
Orzelius, line 90: You also need to change the test case description. Feel free to just copy my rec from above

vvagaytsev: Thanks!

vvagaytsev: Renamed it in https://github.com/garden-io/garden/pull/4848/commits/8ec919b3bdb6c39dc57d0c4bb3965cae7429e429

plugins/pulumi/index.ts
```diff
@@ -131,6 +131,10 @@ export const gardenPlugin = () =>
                 ...omit(module.spec, ["build", "dependencies"]),
               },
             }
+            const buildDependencies = module.build.dependencies
+            if (buildDependencies.length > 0) {
+              deployAction.build = params.convertBuildDependency(buildDependencies[0]).name
```
Walther, line 136: maybe i'm misunderstanding, but is it enough to only supply the first buildDependency here? should we have a for-loop or map here?

plugins/terraform/index.ts
```diff
@@ -214,6 +213,10 @@ export const gardenPlugin = () =>
                 ...omit(module.spec, ["build", "dependencies"]),
               },
             }
+            const buildDependencies = module.build.dependencies
+            if (buildDependencies.length > 0) {
+              deployAction.build = params.convertBuildDependency(buildDependencies[0]).name
```
Walther, line 218: same here

plugins/pulumi/index.ts
vvagaytsev: We have 2 root-level config fields in the action configuration:

 - `build` - a single build dependency:
> Specify a _Build_ action, and resolve this action from the context of that Build.
For example, you might create an `exec` Build which prepares some manifests, and then reference that in a `kubernetes` _Deploy_ action, and the resulting manifests from the Build.
This would mean that instead of looking for manifest files relative to this action's location in your project structure, the output directory for the referenced `exec` Build would be the source.

 - `dependencies`:
> A list of other actions that this action depends on, and should be built, deployed or run (depending on the action type) before processing this action.
Each dependency should generally be expressed as a `"<kind>.<name>"` string, where _<kind>_ is one of `build`, `deploy`, `run` or `test`, and _<name>_ is the name of the action to depend on.
You may also optionally specify a dependency as an object, e.g. `{ kind: "Build", name: "some-image" }`.
Any empty values (i.e. null or empty strings) are ignored, so that you can conditionally add in a dependency via template expressions.

I'm wondering if we should use the same approach as we use in [exec module converter](https://github.com/garden-io/garden/blob/main/core/src/plugins/exec/convert.ts#L62) in all module type converters.

cc @thsig

thsig: > I'm wondering if we should use the same approach as we use in the [exec module converter](https://github.com/garden-io/garden/blob/main/core/src/plugins/exec/convert.ts#L62) in all module type converters.

In short: We should only use the `build` field when one or more runtime plugin handlers (e.g. for deploys or status checks) need to be run _from the build path_ of the dependency (the docstring you referenced above for the `build` key names an example of this).

I don't think that's the case for the Pulumi or Terraform plugins, for example.

For all other cases, just adding the build dependencies to the `dependencies` array is the way to go.

vvagaytsev: Thanks, @thsig! And should we use set `build: dummyBuild` if the dummy one is presented?

vvagaytsev: I've changed the dependencies conversion in https://github.com/garden-io/garden/pull/4848/commits/98177aa971aed82ec686dfb975d001a9d568133d

cc @Walther @thsig

Walther (dismissed):
looks good to me as far as i can tell - thank you both for the fixes! ❤️

plugins/pulumi/index.ts
```diff
@@ -116,6 +116,7 @@ export const gardenPlugin = () =>
               name: module.name,
               ...params.baseFields,
               dependencies: prepareRuntimeDependencies(module.spec.dependencies, dummyBuild),
+              build: params.convertBuildDependency(module.build.dependencies[0]).name,
```
shumailxyz: I would suggest to please add a comment here about `[0]` and the single root level build dependency, so it's more readable in future.

plugins/pulumi/index.ts
```diff
@@ -115,7 +115,10 @@ export const gardenPlugin = () =>
               type: "pulumi",
               name: module.name,
               ...params.baseFields,
-              dependencies: prepareRuntimeDependencies(module.spec.dependencies, dummyBuild),
+              dependencies: [
+                ...module.build.dependencies.map(convertBuildDependency),
+                ...prepareRuntimeDependencies(module.spec.dependencies, dummyBuild),
```
vvagaytsev, line 120: @thsig @shumailxyz @Walther I'm wondering if we still need to include `dummyBuild` here? Should it also b added as a root-level config entry, i.e. `build: dummyBuild`?

vvagaytsev: @Walther `build: dummyBuild?.name` was removed in https://github.com/garden-io/garden/pull/4497/commits/9fe9e808b84088cb3982f1db417cae26c1130888#diff-def0851e81190cb5605afe8d5779115f595aaa677399e2d72cdfb791fb766fe6. Do you recall what was the reason of doing that? Did it cause any issues?

Walther: My memory is a bit hazy on this one, but if I recall correctly, the entire `build` part was removed from the user-facing Pulumi config schema for some reason. Might need to dig more into old PRs and Slack threads 🤔

Walther: Here - the Module schema still has the `build` field, but the Deploy schema doesn't
https://github.com/garden-io/garden/pull/4497/files#diff-a238a97496c93fa58298f0275ae13a57ac515f758978fbd45deebc846f68c3cfR197-R206

vvagaytsev: It's Deploy's _spec_ schema, isn't it? The root-level `build` field is defined in the `BaseAction` and is still documented in the https://github.com/garden-io/garden/blob/0.13/docs/reference/action-types/Deploy/pulumi.md

Walther: Ah, thanks for the clarification! (I should get more coffee)

vvagaytsev:
@shumailxyz @Walther now this is ready for review. Let's see if the CI passes (it should, I've just pushed some squashed commits, the previous CI was ok).

PR #4853, opened 2023-07-19
test(analytics): ensure that DEV_ANALYTICS is used for tests

mkhq:

Selecting the segment key was done on `process.env` and not `gardenEnv`. Moved the logic into the constructor to make sure tests are using the correct value.

Fixes #


vvagaytsev (approved):
LGTM! 👍

PR #4865, opened 2023-07-19
fix(k8s): install `microk8s` addons sequentially

### LINKED ISSUE #4860: Enabling multiple addons on MicroK8s in one invocation may lead to unreachable API server
### Current Behavior

When running `microk8s enable addons dns storage registry ingress`, a `WARNING` is printed:

```python
            "WARNING: Do not enable or disable multiple addons in one command.\n"
            "         This form of chained operations on addons will be DEPRECATED in the future.\n"
            f"         Please, {action} one addon at a time: 'microk8s {action} <addon>'"
```

The responsible line of code is [linked here](https://github.com/canonical/microk8s/blob/07e7ed6f9aa83b986cebc3667588435463dab13d/scripts/wrappers/common/utils.py#L531).

### Expected behavior

Garden should enable addons single invocation by single invocation.

Impacted code in our CircleCI [config.yml](https://github.com/garden-io/garden/blob/69f22b7a972f6f6a327080e947520f655eeacf8b/.circleci/config.yml#L635) and our [config.ts](https://github.com/garden-io/garden/blob/69f22b7a972f6f6a327080e947520f655eeacf8b/core/src/plugins/kubernetes/local/config.ts#L137).
 
### Reproducible example

Use https://github.com/garden-io/quickstart-example

### Workaround

Manually enable addons as user before running first `garden deploy`.

### Additional context

Following `garden deploy` auto-enabling all addons on Windows 21H2, the MicroK8s API server became unreachable, requiring a complete VM delete and re-creation. I manually enabled the required addons one-by-one before running `garden deploy` again, at which point MicroK8s was accessible.

[garden-microk8s-error-log.log](https://github.com/garden-io/garden/files/12093271/garden-microk8s-error-log.log)

### Your environment

* **OS:** Windows Enterprise LTSC 21H2
* **How I'm running Kubernetes:** MicroK8s

`garden version` 0.13.8

vvagaytsev:

Fixes #4860

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -632,9 +632,13 @@ jobs:
             sudo microk8s.status --wait-ready --timeout=300
 
             # Workaround for some issue with snap/microk8s
-            sudo microk8s.enable storage registry dns || true
+            sudo microk8s.enable storage || true
+            sudo microk8s.enable registry || true
+            sudo microk8s.enable dns || true
             sleep 10
-            sudo microk8s.enable storage registry dns
+            sudo microk8s.enable storage
+            sudo microk8s.enable registry
+            sudo microk8s.enable dns
 
             echo "********* microk8s ready! ************"
 
--- a/core/src/plugins/kubernetes/local/microk8s.ts
+++ b/core/src/plugins/kubernetes/local/microk8s.ts
@@ -13,7 +13,7 @@ import { exec } from "../../../util/util"
 import { containerHelpers } from "../../container/helpers"
 import { ContainerBuildAction } from "../../container/moduleConfig"
 import chalk from "chalk"
-import { naturalList, deline } from "../../../util/string"
+import { deline, naturalList } from "../../../util/string"
 import { ExecaReturnValue } from "execa"
 import { PluginContext } from "../../../plugin-context"
 import { parse as parsePath } from "path"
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -632,9 +632,13 @@ jobs:
             sudo microk8s.status --wait-ready --timeout=300
 
             # Workaround for some issue with snap/microk8s
-            sudo microk8s.enable storage registry dns || true
-            sleep 10
-            sudo microk8s.enable storage registry dns
+            # sudo microk8s.enable storage || true
+            # sudo microk8s.enable registry || true
+            # sudo microk8s.enable dns || true
+            # sleep 10
```
shumailxyz, line 638: Are these comments intentional otherwise we can remove those.

Also, i would suggest to keep sleep 10 here for add-ons to be enabled.

vvagaytsev: Intentional and temporary, that was an experimental change, see https://github.com/garden-io/garden/pull/4865/commits/6d84f0fdfebc5d129d18d56196decb3889ae533f.

Let's see how CI goes (I'll try a few reruns) and either enable those back or at least preserve the sleep pause.

vvagaytsev: By the way, is `microk8s.enable` a non-blocking operation? I mean should we wait after its completion?

shumailxyz: okay, then makes sense. Let's see how the CI goes.

shumailxyz: Not sure about microk8s.enable, couldn't find much in quick search. Could be that it's a blocking call and waits already for the addons to be enabled.

vvagaytsev: I reverted https://github.com/garden-io/garden/commit/6d84f0fdfebc5d129d18d56196decb3889ae533f because after that I've started seeing an [error like this](https://app.circleci.com/pipelines/github/garden-io/garden/19145/workflows/e2c9bd95-7629-4d7a-b058-e700b99c6b8b/jobs/349731) more often.

vvagaytsev:
This is ready for review now.

Walther (approved):
Thanks for the fix! Let's hope the CI is a bit more stable now ✨

PR #4868, opened 2023-07-19
ci(dockerhub): fix garden version checks

vvagaytsev:
Use new `garden version` command instead of discontinued `garden --version`.

Fixes the way of checking the Garden version in the `docker-bake-test.sh` script.

Without this PR the command fails silently with OK exit status like this:

<details>

```console
USAGE
  garden <command> [options]

COMMANDS
  build          Perform your Builds.                                                     
  cleanup        Clean up resources.                                                      
  cloud          Manage Garden Cloud/Enterprise resources such as users, groups and       
                 secrets.                                                                 
  community      Join our community Discord to chat with us!                              
  config         Configure user and project settings.                                     
  create         Create new project or module.                                            
  deploy         Deploy actions to your environment.                                      
  dev            Starts the Garden interactive development console.                       
  exec           Executes a command (such as an interactive shell) in a running service.  
  get            Retrieve and output data and objects, e.g. secrets, status info etc.     
  link           Link a remote source or module to a local path.                          
  login          Log in to Garden Cloud.                                                  
  logout         Log out of Garden Cloud.                                                 
  logs           Retrieves the most recent logs for the specified Deploy(s).              
  options        Print global options.                                                    
  plugins        Plugin-specific commands.                                                
  publish        Build and publish artifacts (e.g. container images) to a remote          
                 registry.                                                                
  run            Perform one or more Run actions                                          
  self-update    Update the Garden CLI.                                                   
  serve          Starts the Garden Core API server for the current project and            
                 environment.                                                             
  set            Set or modify data and configuration values.                             
  sync           Manage synchronization to running actions.                               
  test           Run all or specified Test actions in the project.                        
  tools          Access tools included by providers.                                      
  unlink         Unlink a remote source or module from its local path.                    
  up             Spin up your stack with the dev console and streaming logs.              
  update-remote  Pulls the latest version of remote sources, actions or modules from      
                 their repository.                                                        
  util           Misc utility commands.                                                   
  validate       Check your garden configuration for errors.                              
  version        Shows the current garden version.                                        
  workflow       Run a Workflow.                                                          
    
OK: command run_binary garden --version gardendev/garden:bonsai-buster succeeded as expected
```
</details>

This PR replaces the old-fashioned discontinued `garden --version` with the new `garden version` command.

Fixes #

### DIFF
--- a/support/docker-bake-test.sh
+++ b/support/docker-bake-test.sh
@@ -103,18 +103,18 @@ TEST "run all binaries"
   for variant in bonsai-buster{,-rootless} bonsai-alpine{,-rootless}
     do
     # Garden on vanilla images
-    should_succeed garden --version gardendev/garden:$variant
-    should_succeed garden --version gardendev/garden:$variant
+    should_succeed garden version gardendev/garden:$variant
+    should_succeed garden version gardendev/garden:$variant
   done
 
   for variant in bonsai-alpine{,-rootless}
     do
     # garden
-    should_succeed garden --version gardendev/garden-aws-gcloud-azure:$variant
-    should_succeed garden --version gardendev/garden-aws-gcloud:$variant
-    should_succeed garden --version gardendev/garden-aws:$variant
-    should_succeed garden --version gardendev/garden-gcloud:$variant
-    should_succeed garden --version gardendev/garden-azure:$variant
+    should_succeed garden version gardendev/garden-aws-gcloud-azure:$variant
+    should_succeed garden version gardendev/garden-aws-gcloud:$variant
+    should_succeed garden version gardendev/garden-aws:$variant
+    should_succeed garden version gardendev/garden-gcloud:$variant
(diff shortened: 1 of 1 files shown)


Walther (approved):
Nice catch! 🎉

shumailxyz (approved):
Good catch 🌟

stefreak:
I think there's another issue here: Why does garden exit with code 0 when passing an unrecognized option? @vvagaytsev

vvagaytsev:
> I think there's another issue here: Why does garden exit with code 0 when passing an unrecognized option? @vvagaytsev

Yes, I think that should be changed too. It looks like the `0` error code comes from this block: https://github.com/garden-io/garden/blob/main/core/src/cli/cli.ts#L470-L474

stefreak:
I agree this should be fixed in a patch release, IMO this is not a breaking change but a bug!

vvagaytsev:
Created #4911.

PR #4881, opened 2023-07-21
fix(k8s): allow null in spec.files for deploy config

### LINKED ISSUE #4878: 0.13: [Bug]: kubernetes deploy action: crash when encountering falsy file entries
### Current Behavior

Garden crashes when encoutering a falsy (i.e., `""` or `null`) entry in the `spec.files` array of a kubernetes deploy action.

### Expected behavior

Garden should ignore the file.
This is a regression - it used to work in <0.13

### Reproducible example

```
# example.garden.yml
kind: Deploy
type: kubernetes
name: example
spec:
  files:
    - null
```

```
$ garden deploy example

Error processing config for Deploy action example:
Error validating configure handler output from provider kubernetes for Deploy type kubernetes: key .config does not match any of the allowed types. Available keys: config, supportedModes)
```

### Workaround

Not known.

### Additional context

This is required when using templating to determine whether a file is supposed to be included or not.
For example: 
```
kind: Deploy
type: kubernetes
name: example
spec:
  files:
    - "${var.ingress-enabled ? 'ingress.yaml' : null}"
```

### Your environment

* **OS:** arch linux

`garden version`
0.13.9

shumailxyz:
ignore null values from spec.files in configure handler. Useful when some manifest files are to be used conditionally based on variable/template string resolution.

Fixes #4878


Walther (commented):
Nice work!
Could we have a regression test for this please 👉 👈

core/src/plugins/kubernetes/kubernetes-type/deploy.ts
```diff
@@ -44,7 +44,8 @@ export const kubernetesDeployDefinition = (): DeployActionDefinition<KubernetesD
     configure: async ({ ctx, config }) => {
       // The `spec` field hasn't been validated here yet,
       // so the value of `files` might be `undefined` instead of a default empty array
-      let files = config.spec.files || []
+      // also we discard any falsy file values in files array
+      let files = (config.spec.files || []).filter((f) => !!f)
```
vvagaytsev, line 48: We can use our fancy `isTruthy` helper here :)
```
let files = (config.spec.files || []).filter(isTruthy)
```

shumailxyz: nice. I wasn't aware that this existed.

shumailxyz: Updated.

vvagaytsev (approved):
Great, thanks for fixing this! 💯

hnicke:
I'm rather impatiently waiting for this fix to get released :smiley:

vvagaytsev:
@hnicke we have released [0.13.10](https://github.com/garden-io/garden/releases/tag/0.13.10)! :)

hnicke:
great. thank you!

PR #4889, opened 2023-07-25
feat(config): better error messages around schema validation

edvald:
This attaches the YAML parse results to configs when loading, and leverages that when rendering validation error messages.

```yaml
...
3  | spec:
4  |   foo: 123
-------^
"foo" should be a string
```

This should be a useful step for an IDE highlighting integration as well.


vvagaytsev (changes requested):
There are some test failures, please take a look.

TimBeyer:
Does this work as expected with YAML files that contain multiple docs?
When I was trying this out you needed to do something like

```ts
const lineCounter = new LineCounter()
const parser = new Parser(lineCounter.addNewLine)

const tokens = parser.parse(content)
const docsGenerator = new Composer().compose(tokens)

const docs = Array.from(docsGenerator) as Document.Parsed<ParsedNode>[]
```

to get the docs and then use `lineCounter.linePos` to get the correct line in the input file.

edvald:
I hadn't tested that specifically, good catch. Not sure I fully follow your code there though. Happy to take commits on this PR btw, it's very much a draft still, bunch of broken tests etc.

thsig:
Btw, I think it's acceptable if we just show the YAML doc that the error came from, even it it's a sub-doc of a larger one—we just need to help the user quickly pinpoint the source location.

core/src/config/validation.ts
```diff
@@ -82,78 +94,147 @@ export const validateWithPath = profile(function $validateWithPath<T>({
...
+    return result.value
+  }
+
+  const description = schema.describe()
+
+  const yamlDoc = source?.yamlDoc
+  const rawYaml = yamlDoc?.source
+  const yamlBasePath = source?.basePath || []
+
+  const errorDetails = error.details.map((e) => {
```
vvagaytsev, line 140: This could be extracted as a separate function that does the error rendering.
Also, there are some index-based array operations. Would it be safer to wrap them in a `try/catch` just in case?

core/src/tasks/resolve-provider.ts
```diff
@@ -174,7 +174,24 @@ export class ResolveProviderTask extends BaseTask<Provider> {
...
+    const yamlDoc = projectConfig.internal.yamlDoc
+    let yamlDocBasePath: (string | number)[] = []
+
+    if (yamlDoc) {
+      projectConfig.providers.forEach((p, i) => {
+        if (p.name === this.config.name) {
+          yamlDocBasePath = ["providers", i]
+          return false
+        }
+        return true
```
vvagaytsev, line 188: Does this statement always break the loop after the first iteration? Was it done intentionally? It looks like the returned `boolean` value is never used, do we need it?

vvagaytsev:
I left a few comments and questions, please check.
Otherwise looks good, thank you! 👍

core/src/config/validation.ts
edvald: A try/catch definitely makes sense. Will refactor a bit.

core/src/tasks/resolve-provider.ts
edvald: That's intentional. The boolean needs to be there, linter complains if we only return it to break.

core/src/config/validation.ts
edvald: Done 👍

core/src/tasks/resolve-provider.ts
vvagaytsev: Ok, got it, thanks! It may make sense to leave a short comment on that :)

vvagaytsev (dismissed):
LGTM! 👍  Thanks for the great improvement! 💯

edvald:
Just did that actually, good timing :) Just grappling with my merge commit now.

edvald:
Ok, should be ready now, pending tests passing.

vvagaytsev:
Now there are more conflicts to be resolved.

edvald:
Updated again.

core/src/template-string/template-string.ts
```diff
@@ -80,6 +80,23 @@ function getValue(v: Primitive | undefined | ResolvedClause) {
...
 }
 
+type ObjectPath = (string | number)[]
+
+export class TemplateError extends GardenError {
+  type = "template"
+
+  path: ObjectPath | undefined
+  value: any
+  resolved: any
```
vvagaytsev, line 90: It looks like these parameters are never rendered. Do we need this specific kind of error? Or should we just use `ConfigurationError` class instead?
@stefreak @TimBeyer wdyt?

edvald: I opted to keep this even though the typed detail object was removed, since we are likely to use this (and potentially more) later for e.g. a language server implementation.

vvagaytsev (approved):
Thank you! 💯

PR #4891, opened 2023-07-25
fix(config): throw error if multiple project configs are found

### LINKED ISSUE #3210: BUG: no error with two garden projects in the same directory
### Current Behavior
If you have two`kind: project` definitions in different files there is no error message

### Expected behavior
I expect an error, otherwise it can be very confusing if user edits the "wrong" file (that garden does not use)

### Reproducible example
```
mkdir repro
cd repro
git init
garden create project
mv garden.project.yml garden.yml
garden create project
garden deploy
```

### Workaround
Unknown

### Suggested solution(s)
Garden should validate project definitions. There should be only one. Otherwise the user needs to be warned

### Additional context
Just had this issue and it was very confusing, until I actually found out there was accidentally two project definitions in the same dir.

### Your environment

* **OS:** macos

`garden version`
0.12.42

shumailxyz:
throw an error if multiple project configs are found across multiple files in a dir (e.g. in garden.yml and project.garden.yml)

Fixes #3210

### DIFF
--- a/core/src/config/base.ts
+++ b/core/src/config/base.ts
@@ -457,24 +457,38 @@ export async function findProjectConfig({
 }): Promise<ProjectResource | undefined> {
   let sepCount = path.split(sep).length - 1
 
+  let allProjectSpecs: GardenResource[] = []
+
   for (let i = 0; i < sepCount; i++) {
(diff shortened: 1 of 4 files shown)


Walther (approved):
Thanks for the fix! ✨

PR #4894, opened 2023-07-26
improvement(k8s): less verbose Run/Test errors

thsig:

We now only log the full pod status on the debug log level and above when a Test or Run fails.

This makes e.g. logs for failing Tests much less verbose by default, which should be a UX improvement.

### DIFF
--- a/core/src/plugins/kubernetes/run.ts
+++ b/core/src/plugins/kubernetes/run.ts
@@ -1236,6 +1236,14 @@ export class PodRunner extends PodRunnerParams {
       throw err
     }
 
+    function renderDiagnosticErrorMessage(error: KnownError): string | undefined {
+      if (error.type === "pod-runner" && error.detail.podStatus) {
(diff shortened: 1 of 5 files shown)


core/src/plugins/kubernetes/run.ts
```diff
@@ -1305,16 +1309,17 @@ export class PodRunner extends PodRunnerParams {
...
+      log: renderError(err),
+      diagnosticErrorMsg: renderDiagnosticErrorMessage(err),
       moduleName,
       version,
       success: false,
       startedAt,
       completedAt: new Date(),
       command,
       exitCode: err.detail.exitCode,
+      errorDetail: err.detail,
```
vvagaytsev, line 1322: We don't seem to read this explicitly in the caller-side code. But this will be printed to the error.log if the `RunResult` is not successful, right? Then it's definitely good to have more details in the logs.

vvagaytsev (approved):
LGTM! 👍 Thank you!

It looks like #4904 duplicates this.

PR #4901, opened 2023-07-31
fix(docs): update migration guide

### LINKED ISSUE #4890: 0.13: [Bug?]: Task caching behavior changes
### Current Behavior

I've noticed that starting with v0.13, the task caching  behavior changed quite a bit.
When directly running a task in v0.12 (`garden run task <task-name>`), the task will never hit the cache and hence would always run.
In contrast when directly running a task in v0.13 (`garden run <task-name>`), the task can hit the cache (`Already complete`).

This change of behavior broke most of our garden workflows.
I'm aware of the fact that I can disable the cache with the `--force` flag, which will be my workaround.

Worth mentioning is that we're still haven't migrated to actions yet - all tasks stem from modules.

Is this a bug or a feature?
If it's a feature, I suggest mentioning this breaking change in the migration guide (https://docs.garden.io/tutorials/migrating-to-bonsai).

### Expected behavior
I'd expect that the task caching behaves just as in v0.12.

### Workaround
Use `--force` flag:

`garden run --force <task-name>`

### Your environment

* **OS:** arch linux
* **How I'm running Kubernetes:** KE

`garden version`
0.13.9

vvagaytsev:
Document the breaking change in the Run command caching. Now the caching is enabled by default.

Fixes #4890


Walther (approved):
Nice clarification!

PR #4903, opened 2023-07-31
fix(k8s): regression in globs in k8s manifest files

### LINKED ISSUE #4900: [FEATURE]: Print a warning or something if a kubernetes deploy action specifies files which are not found
### Background / Motivation

In #4899, I got pretty confused. I had a kubernetes deploy action which specified a file named `jaeger-operator.yml`. When I ran the deploy, nothing happened. That's because the file was actually named `jaeger-oeprator.yaml` - `yaml` vs `yml`.

Get warned when a file is specified that isn't actually found. Maybe throw an error and abort, even.

It's confusing for the tool to proceed happily even if a specified file is missing. It can lead to the false belief that a deploy succeeded.

Dunno! At least warning is good. A fatal error would make sense to me too, but maybe there's some reason to proceed? I don't really know.

vvagaytsev:
The regression was introduced in #4516. After that PR, it was possible to silently skip all non-existing files specified in in `spec.files` of a Deploy action of type `kubernetes`.

In `0.13.6` that would fail with error like this:
```console
ENOENT: no such file or directory, open '~/garden/core/test/data/test-projects/kubernetes-type/with-build-action/deployment-action1.yaml'
```

This PR implements some explicit checks for path existence and makes the output a bit more readable:
```console
Invalid manifest file path(s) in Deploy action 'with-build-action'. Cannot find manifest file(s) at ~/garden/core/test/data/test-projects/kubernetes-type/.garden/build/exec-build/deployment-action1.yaml
```

Issues #4505 and #4506 still exist. That's why the deployment can fail even with the existing manifest file. The build directory appears to be empty. It will be fixed in a separate PR.

Fixes #4900

You can find more detail and explanation in the commit messages.

### DIFF
--- a/core/src/plugins/kubernetes/kubernetes-type/common.ts
+++ b/core/src/plugins/kubernetes/kubernetes-type/common.ts
@@ -7,7 +7,7 @@
  */
 
 import { resolve } from "path"
-import { readFile } from "fs-extra"
+import { pathExists, readFile } from "fs-extra"
 import Bluebird from "bluebird"
 import { flatten, set } from "lodash"
 import { loadAll } from "js-yaml"
@@ -15,7 +15,7 @@ import { loadAll } from "js-yaml"
 import { KubernetesModule } from "./module-config"
 import { KubernetesResource } from "../types"
 import { KubeApi } from "../api"
-import { gardenAnnotationKey } from "../../../util/string"
+import { gardenAnnotationKey, naturalList } from "../../../util/string"
 import { Log } from "../../../logger/log-entry"
 import { PluginContext } from "../../../plugin-context"
 import { ConfigurationError, PluginError } from "../../../exceptions"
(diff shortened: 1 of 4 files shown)


core/src/plugins/kubernetes/kubernetes-type/common.ts
```diff
@@ -128,11 +129,53 @@ export async function readManifests(
...
+  const specFiles = spec.files
 
-  const files = await glob(spec.files, { cwd: manifestPath })
+  const regularPaths = specFiles.filter((f) => !isGlob(f)).map((path) => resolve(manifestPath, path))
+  const missingPaths = await Bluebird.filter(regularPaths, async (regularPath) => {
+    return !(await pathExists(regularPath))
+  })
+  if (missingPaths.length) {
+    throw new ConfigurationError({
+      message: `Invalid manifest file path(s) in ${action.kind} action '${
```
thsig, line 140: Since we're adding a nice error message here, let's go the extra mile and include the missing (relative) paths in the error message.

thsig (changes requested):
Great stuff! Really appreciate the added test cases.

I had one suggestion to improve the new error message, otherwise we're good to go here.

core/src/plugins/kubernetes/kubernetes-type/common.ts
vvagaytsev: Changed that in dad43954aff3ab8a554e1973e8c950a3da1ce866.

Walther (approved):
Looks good, thanks for the fix! ✨

PR #4906, opened 2023-07-31
chore: add code coverage for core

### LINKED ISSUE #4545: Investigate code test coverage tools
Currently, we have a lot of code that is not tested. Unfortunately, we also do not have easy visibility into _which parts_ are untested.

Let's investigate various tools, e.g.
- codecov
- coveralls
- others?

This would be a tool for figuring out where we are lacking coverage and help us improve things.
This would **not** be about trying to achieve some specific percentage of coverage as an absolute value.

Walther:

Fixes #4545 


vvagaytsev:
Closing this as we have closed the original issue.

PR #4907, opened 2023-07-31
fix(template): inputs processing in module config resolution

### LINKED ISSUE #4882: 0.13: [Bug]: ModuleTemplate failed to identify variable with default value pass in schema.json file
### Current Behavior

Garden fails with the error:

```console
Failed resolving one or more modules:

name: Invalid template string (-Djib.from.image=${var.dockerRepo}…): Could not find key jdkVersion under inputs. Available keys: (none).
```

### Expected behavior

No errors, validation should pass.

### Reproducible example

Run `garden validate` with the configuration below.

**Garden configuration.**
```yaml
kind: Project
apiVersion: garden.io/v1
name: repro
environments:
  - name: default
providers:
  - name: local-kubernetes
    environments: [ default ]
variables:
  dockerRepo: "hub.docker.io"
  baseImageJava8: eclipse-temurin:8u345-b01-jre-ubuntu23
  baseImageJava11: distroless-os-images:debian11-java11.0.18

---
kind: Module
type: templated
template: k8s-container
name: name

---
kind: ModuleTemplate
name: k8s-container
inputsSchemaPath: schema.json

modules:
  - kind: Module
    type: container
    name: ${parent.name}  # Now possible as it doesn't conflict with the RenderTemplate type
    extraFlags:
      - "-Djib.from.image=${var.dockerRepo}/${inputs.jdkVersion == '11' ? var.baseImageJava11 : var.baseImageJava8}"
```

**`schema.json`**
```yaml
{
    "type": "object",
    "properties": {
      "excludeModules": {
        "type": "array",
        "items": {
            "type": "string"
        }
      },
      "jdkVersion": {
        "type": "string",
        "default": "8"
      }
    }
}
```

**Dummy `Dockerfile`**
```dockerfile
FROM busybox:1.36.0
ENTRYPOINT ["tail", "-f", "/dev/null"]
```

### Workaround

Use action-based template:

```yaml
---
kind: ModuleTemplate
name: k8s-container
inputsSchemaPath: schema.json

configs:
  - kind: Build
    type: container
    name: ${parent.name}  # Now possible as it doesn't conflict with the RenderTemplate type
    spec:
      extraFlags:
        - "-Djib.from.image=${var.dockerRepo}/${inputs.jdkVersion == '11' ? var.baseImageJava11 : var.baseImageJava8}"
```

vvagaytsev:

Fixes #4882

### DIFF
--- a/core/src/resolve-module.ts
+++ b/core/src/resolve-module.ts
@@ -342,8 +342,12 @@ export class ModuleResolver {
     if (templateName) {
       const template = this.garden.configTemplates[templateName]
 
-      // All inputs need to be resolvable
-      inputs = resolveTemplateStrings(inputs, new ModuleConfigContext(templateContextParams), { allowPartial: false })
+      inputs = resolveTemplateStrings(
+        inputs,
+        new ModuleConfigContext(templateContextParams),
+        // Not all inputs may need to be resolvable
+        { allowPartial: true }
+      )
 
       inputs = validateWithPath({
         config: cloneDeep(config.inputs || {}),
--- a/core/src/resolve-module.ts
+++ b/core/src/resolve-module.ts
@@ -9,23 +9,23 @@
 import { cloneDeep, isArray, isString, keyBy } from "lodash"
 import { validateWithPath } from "./config/validation"
 import {
-  resolveTemplateStrings,
   getModuleTemplateReferences,
-  resolveTemplateString,
   mayContainTemplateString,
+  resolveTemplateString,
+  resolveTemplateStrings,
 } from "./template-string/template-string"
 import { ContextResolveOpts, GenericContext } from "./config/template-contexts/base"
-import { relative, resolve, posix, dirname } from "path"
+import { dirname, posix, relative, resolve } from "path"
 import type { Garden } from "./garden"
 import { ConfigurationError, FilesystemError, PluginError } from "./exceptions"
-import { deline, dedent } from "./util/string"
+import { dedent, deline } from "./util/string"
 import {
-  ModuleConfigMap,
   GardenModule,
-  ModuleMap,
-  moduleFromConfig,
-  ModuleTypeMap,
   getModuleTypeBases,
+  ModuleConfigMap,
+  moduleFromConfig,
+  ModuleMap,
+  ModuleTypeMap,
 } from "./types/module"
 import { BuildDependencyConfig, ModuleConfig, moduleConfigSchema } from "./config/module"
 import { Profile, profileAsync } from "./util/profiling"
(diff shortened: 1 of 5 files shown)


vvagaytsev:
Hmm.. I think this is wrong. As [this comment](https://github.com/garden-io/garden/pull/4907/files#diff-e13fa0f921a945a48a4e813bcf47059e53a6f9d1025d975a29cf19a79ffd3505R347) says, `Not all inputs may need to be resolvable`. We can't resolve runtime inputs in advance. But an actions's `spec` still can reference the inputs.

vvagaytsev:
Actually, the fix I've made might be ok. The [Module `inputs` docs](https://docs.garden.io/reference/module-types/templated#inputs) say:
> Note: You can use template strings for the inputs, but be aware that inputs that are used to generate the resulting config names and other top-level identifiers must be resolvable when scanning for configs, and thus cannot reference other actions, modules or runtime variables. See the [environment configuration context reference](https://github.com/reference/template-strings/environments) to see template strings that are safe to use for inputs used to generate config identifiers.

core/test/data/test-projects/template-configs/garden.yml
```diff
@@ -0,0 +1,43 @@
...
+name: template-configs
+environments:
+  - name: default
+providers:
+  - name: test-plugin
+    environments: [ default ]
+variables:
+  dockerRepo: "hub.docker.io"
+  baseImageJava8: eclipse-temurin:8u345-b01-jre-ubuntu23
+  baseImageJava11: distroless-os-images:debian11-java11.0.18
```
edvald, line 12: I guess that works but these strike me as oddly specific for the context :)

core/src/resolve-module.ts
```diff
@@ -343,12 +342,8 @@ export class ModuleResolver {
...
       const template = this.garden.configTemplates[templateName]
 
-      inputs = resolveTemplateStrings(
-        inputs,
-        new ModuleConfigContext(templateContextParams),
-        // Not all inputs may need to be resolvable
-        { allowPartial: true }
-      )
+      // All inputs need to be resolvable
+      inputs = resolveTemplateStrings(inputs, new ModuleConfigContext(templateContextParams), { allowPartial: false })
```
edvald, line 346: I'm _fairly_ sure this would constitute a breaking change in some scenarios. Is this actually relevant for the issue?

vvagaytsev: Thanks! That's not necessary to fix the issue. Let's revert this.

vvagaytsev: @edvald what are the scenarios when we can have partially resolved inputs here? Inputs can contain template strings but only from the environment config context. Can that context be unresolved at this point?

vvagaytsev: Ah, ok. Thank you! Reverted this in e910bb196b045e68d010a35b4bdb6199022635f7.

core/test/data/test-projects/template-configs/garden.yml
vvagaytsev: Made it less exotic in 5d30a5a4eae2f06f1465a7556a6ec07f709651df :)

Walther (approved):
Looks good to me - thanks for the fix! ✨

PR #4915, opened 2023-08-02
chore(analytics): include project and user metadata when running commands with noProject

mkhq:

Optionally includes project and user metadata when running a command with `noProject = true`. Internally, we instantiate a DummyGarden instance which is not having project or user metadata available. In addition, we only set the `enterpriseDomain` field if its defined in the project config.

To achieve this there are a couple of notable changes:

- The global store is extended with a cloud user profile including `userId` and `organizationName` for each associated domain
- A profile request is made when logging in to a cloud domain to ensure the existence of the user and to store the metadata in the global store upon login
- When analytics is called with a `DummyGarden`-instance, it tries to fetch the project name and cloud domain from the config + does a lookup of the cloud user profile 
- Mocha tests are called with an additional flag to handle nock and fetch compatibility (https://github.com/nock/nock/issues/2397)
- Version checks are disabled in tests unless explicitly part of the tests. This was always the case for circle CI which is why the nocked version check service tests always failed.
- Extends the `pendingRequests` for identify analytics events to make sending them more robust 
- Only sets the `cloudDomain` when logged in
- Only sets the `enterpriseDomain` and `enterpriseProjectId` when defined in the project config

Fixes #

- Auth token validity check is best effort, opted for that approach now instead of pulling in a JWT library


core/src/cloud/api.ts
```diff
@@ -257,7 +264,8 @@ export class CloudApi {
     log: Log,
     globalConfigStore: GlobalConfigStore,
     tokenResponse: AuthTokenResponse,
-    domain: string
+    domain: string,
+    userProfile: CloudUserProfile | undefined = undefined
```
vvagaytsev, line 268: It might be nicer to have a parameter object instead of 5 positional args here.

core/src/analytics/analytics.ts
```diff
@@ -445,15 +458,58 @@ export class AnalyticsHandler {
...
         log.debug(`Getting profile from API failed with error: ${err}`)
       }
     }
 
+    // best effort load the project if this is a dummy garden instance
+    let projectName = garden.projectName
+
+    let projectConfig: ProjectConfig | undefined
+
+    if (garden instanceof DummyGarden) {
```
eysi09, line 483: Can we extract this to a separate function and update the control flow so that it doesn't have all these nested if statements?

We can instead return early with something like:

```
if (!projectConfig) {
  return
}

// proceed...
```

eysi09 (changes requested):
Great stuff! 

I had a couple of comments but otherwise looks good.

vvagaytsev:
Closing as not planned for now. We'll re-think this and create a new PR if necessary.

Issue #4880, opened 2023-07-21
Non-existing variable breaks the behaviour of `$merge` operator


vvagaytsev:
### Current Behavior
When `$merge` operator encounters an undefined variable as an argument it never gets resolved properly. It is rendered as JSON object instead.

Assume we have variables defined in the project-level config:
```yaml
variables:
  example-env:
    FIELD1: hi
    FIELD2: bye
```

If we use `$merge` operator like this
```
  env:
    $merge: ${var.example-env}
    HELLO: GOODBYE
```
then we get a valid output:
```yaml
          env:
            - name: FIELD1
              value: hi
            - name: FIELD2
              value: bye
            - name: HELLO
              value: GOODBYE
```

But if we change `$merge: ${var.example-env}` to ``$merge: ${var.empty || var.example-env}` then the output is invalid:
```yaml
          env:
            - name: $merge
              value:
                FIELD1: hi
                FIELD2: bye
            - name: HELLO
              value: GOODBYE
```
Here `$merge` is not considered a valid helper function and is interpreted as a JSON object key.

### Expected behavior

The missing variable can be interpreted as an empty one to produce a valid output:
```yaml
          env:
            - name: FIELD1
              value: hi
            - name: FIELD2
              value: bye
            - name: HELLO
              value: GOODBYE
```

Or an error can be thrown because of the reference to a missing variable. To be confirmed. Need to check the expected behavior in the docs.

### Reproducible example

<details>

1. Create the project from the files below:

<details>

`bug-demo.garden.yml`
```yaml
---
kind: Module
type: container
name: bug-module
description: Container Module
---
kind: Module
type: templated
template: manifests-local
name: manifests
inputs:
  env:
    $merge: ${var.empty || var.example-env}
    HELLO: GOODBYE
---
kind: ModuleTemplate
name: manifests-local
inputsSchemaPath: schema.json
modules:
  - type: kubernetes
    name: bug-service
    description: ${parent.name} manifests
    build:
      dependencies:
        - bug-module
    variables:
      env:
        # put non-secrets into the correct format.
        - $concat:
            $forEach: ${inputs.env || []}
            $return:
              name: ${item.key}
              value: ${item.value}
    files: [.manifests-template.yml]
    generateFiles:
      - sourcePath: manifests-template.yml
        targetPath: .manifests-template.yml
```

`project.garden.yml`
```yaml
kind: Project
name: example-project
variables:
  example-env:
    FIELD1: hi
    FIELD2: bye
[...]

PR #4932, opened 2023-08-09
fix: templated module templating

TimBeyer:
Fixes an issue where in Action / Module templates there would be variable used to iterate over using `$foreach` which cannot be resolved in partial application using `allowPartial: true` but still gets returned as an object.

We now check if something returned to iterate over still includes unresolved 'magic' keys and in case that is the case, it tries to resolve them again in non-partial mode, or just returns the original unresolved value in partial mode.

That ensures that we do not iterate over those until the final non-partial application of the templating.

Fixes https://github.com/garden-io/garden/issues/4880

### DIFF
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -7,7 +7,7 @@
  */
 
 import Bluebird from "bluebird"
-import { cloneDeep, isEqual, mapValues, memoize, pick, uniq } from "lodash"
+import { cloneDeep, isEqual, mapValues, memoize, omit, pick, uniq } from "lodash"
 import {
   Action,
   ActionConfig,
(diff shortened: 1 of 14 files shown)


core/src/graph/actions.ts
```diff
@@ -609,10 +609,11 @@ export const preprocessActionConfig = profileAsync(async function preprocessActi
 
     // Partially resolve other fields
     // TODO-0.13.1: better error messages when something goes wrong here (missing inputs for example)
+
     const resolvedOther = resolveTemplateStrings(omit(config, builtinConfigKeys), builtinFieldContext, {
       allowPartial: true,
     })
-    config = { ...config, ...resolvedOther }
+    config = { ...config }
```
edvald, line 616: Is this necessary along with the other fix? I can imagine some side-effects here.

TimBeyer: Nope, that was an oversight, thanks for spotting it 👍

vvagaytsev (approved):
Great! ✨ Thanks for fixing this!

PR #4933, opened 2023-08-09
fix: delete pvc on namespace cleanup

### LINKED ISSUE #1740: `garden del env` can't finish
reporting a bug as suggested in #1734 

## Bug

`garden del env` stopped working for me as soon I started using the persistenvolumeclaim, I can see during deleting `No delete service handler available for module type persistentvolumeclaim` and the PersistentVolume I have declared as k8 yml (module type kubernetes) just hangs indefinitely presumably because the volume is still claimed. After `kubectl delete pvc my-pvc` another try of garden del env passes.

### Current Behavior

`garden del env` hangs indefinitely.

### Expected behavior

`garden del env`, deletes the PVC and allows PV to be deleted as well.

### Reproducible example

You just need to declare a PV in k8. I can create a small repro repo if needed.

### Workaround

`kubectl delete pvc <my_pvc>` before `garden del env`

### Suggested solution(s)

Not a k8 expert but probably PVC should be deleted during `del env`

### Your environment

OSX
garden/kubectl/docker at newest

shumailxyz:
Add delete handler for pvc

Fixes #1740

### DIFF
--- a/core/src/plugins/kubernetes/volumes/persistentvolumeclaim.ts
+++ b/core/src/plugins/kubernetes/volumes/persistentvolumeclaim.ts
@@ -107,11 +107,11 @@ export const persistentvolumeclaimDeployDefinition = (): DeployActionDefinition<
     delete: async (params) => {
       const result = await deleteKubernetesDeploy({
         ...(<any>params),
-        action: getKubernetesAction(params.action)
+        action: getKubernetesAction(params.action),
       })
 
       return { ...result, outputs: {} }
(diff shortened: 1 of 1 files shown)


vvagaytsev (changes requested):
Thanks for fixing this!
LGTM, but there are some lint errors, please double-check and fix 'em.

vvagaytsev (approved):
Thanks for fixing this! 👍

PR #4945, opened 2023-08-15
improvement(dev): show spinner when cmd is running

thsig:

We now show a spinner while one or more commands is being run in the dev console.

We've gotten the feedback that when individual actions take a while (e.g. long builds) and there isn't any new log output for several seconds,, users want visual feedback that the command/actions are still being processed.

I'm assigning this to @TimBeyer, since he was the last one to modify the relevant files (around the recent OTEL work), but feel free to reassign this as is relevant.


core/src/cli/command-line.ts
```diff
@@ -746,11 +743,10 @@ ${chalk.white.underline("Keys:")}
             )
           } catch (error) {
             this.flashError(getCmdFailMsg(name))
-            this.log.error({ error })
-            return
-          } finally {
             delete this.runningCommands[id]
```
TimBeyer, line 746: Without the `finally`, won't we never clean up the running command then?

thsig: We clean it up a bit further down in this method:

https://github.com/garden-io/garden/pull/4945/files#diff-fd1ffc6f1b7bbbe7ff38de11288a20940c7cb69cd52bda986990c086e122d123R782

The reason I moved this from the `finally` to the `catch` is that before, we stopped showing the spinner and the "Running command" status text too early in the flow.

TimBeyer: Oh I see, we just grabbed the child garden object and we immediately removed the running command from the list. Makes sense

TimBeyer (approved):

TimBeyer (approved):

PR #4952, opened 2023-08-16
build: fix edge release

TimBeyer:
Makes sure that we update the tag for `edge-bonsai` when we make a prerelease for it.
Also updates the release to point to the new edge tag.

That way the edge release should always be up to date with the correct pointer and be easily findable in GH.

Fixes #


vvagaytsev (approved):
Thanks! 👍

stefreak (approved):
Thank you for fixing this! Very good change

PR #4957, opened 2023-08-17
fix(core): improve dependency logic for tests

thsig:

Before this fix, changes to transitive (i.e. not direct) runtime dependencies of a test action would not result in the test action being re-run if the action handler of the intermediating runtime dependency can return `ready` even when input versions change. This is the case for `kubernetes` and `helm` Deploys (but not for `container` Deploys).

Note that this only applies when outdated components have already been deployed in the environment/namespace.

To illustrate, let's say `T`, `D1`, `D2` and `B2` are a Test, two Deploys and a Build (where `B2` is the build for `D2`).

If `B2` is changed in such a way that might cause `T` to fail when run against an up-to-date version of `D1` and `D2` unless the action handler for `D2` always returns a non-ready status when the input version changes, then under those circumstances `T` would erroneously show a cached and successful result.

This was fixed by defaulting to recursively including all runtime dependencies for test tasks.

Also added a solver test case for this (and set up some test data & helpers that should be useful for writing more tests that need to use the semantics of the action-based task classes).


edvald:
Can you provide an example where `B2` doesn't implicitly affect the shape and thus status of `D2`?

thsig:
The problem is that e.g. the status handlers for `kubernetes` and `helm` Deploys return ready whenever the manifests don't change.

So before this fix, the solver would start at `T` and then not look past `D2` (since the status there comes out as `ready`).

I.e. the versions change, but that's not enough to trigger reprocessing of nested dependencies if there's a `ready` node in between.

core/test/unit/src/graph/solver.ts
```diff
@@ -338,106 +542,6 @@ describe("GraphSolver", () => {
   //   )
   // })
 
-  // it("should emit events when processing and completing a task", async () => {
```
edvald, line 341: Is this intentional?

thsig: Ah, on second thought I'll undo this part.

thsig: Yeah, since we're no longer emitting these events (and the caching semantics that are being tested date to the 0.12-era TaskGraph too).

I thought I'd just clean up here while I was modifying the file, but I'm happy to undo this deletion too.

thsig: Thoughts on this? Unless there are more comments/questions (or if we'd prefer another solution to the problem this PR addresses), I think this should be ready to go.

thsig: Any thoughts on this, @edvald? Unless there are more comments/questions (or if we'd prefer another solution to the problem this PR addresses), I think this should be ready to go.

thsig:
As of the changes in https://github.com/garden-io/garden/pull/4846, I think `kubernetes` Deploys no longer have this problem.

thsig:
The deployment & status checking semantics for `kubernetes` and `helm` Deploys have been updated/fixed since this PR was written, so it's no longer necessary.

Issue #4921, opened 2023-08-04
generateFiles in a shared ModuleTemplate generates the files in the calling repository instead of the build directory


BillRamsey-minted:
## Bug
generateFiles in a shared ModuleTemplate generates the files in the calling repository instead of the build directory and also if called twice, it over-writes each other.

### Current Behavior

in my project, I include a source for the module template:
ie:
```
sources:
  - name: templates
    repositoryUrl: https://github.com/BillRamsey-minted/garden-template-example#main
    
---
kind: Module
type: container
name: bug-module
description: Container Module
---
kind: Module
type: templated
template: my-template
name: service-1
inputs:
  replicas: 1
  image: ${modules.bug-module.outputs.deployment-image-id}
  service-name: s1
---
kind: Module
type: templated
template: my-template
name: service-2
inputs:
  replicas: 1
  image: ${modules.bug-module.outputs.deployment-image-id}
  service-name: s2
```

the ModuleTemplate has a generateFiles:
```
    files: [".manifests/.manifests-template.yml"]
    generateFiles:
      - sourcePath: manifests-template.yml
        targetPath: ".manifests/.manifests-template.yml"
```

when I do a `garden build` that file shows up in the calling project in three places:
top level:
.manifests/.manifests-template.yml
.garden/build/s1/.manifests/.manifests-template.yml
.garden/build/s2/.manifests/.manifests-template.yml

However, both the build manifests in s1/ and s2/ show the same service name of s2.
### Expected behavior

I would expect it not to create the file at the top level of the calling repo at all,
and the build/s1 and build/s2 to each be different manifests instead of both using the same service name.

### Reproducible example

[project.garden.txt](https://github.com/garden-io/garden/files/12263674/project.garden.txt)

### Workaround

I include the service name in the manifests:
```
    files: ['.manifests/.${inputs.service-name}-secrets.yml']
    generateFiles:
      - sourcePath: 'manifest-templates/secrets.yml'
        targetPath: '.manifests/.${inputs.service-name}-secrets.yml'
```

### Suggested solution(s)

unknown

### Your environment

EKS and docker desktop

`garden version`
0.12.63

vvagaytsev:
@BillRamsey-minted thanks for reporting this!

It looks like the current behavior is formally correct, despite it looking confusing.
Let's take a closer look at the `ModuleTemplate` config:

```yaml
    files: [".manifests/.manifests-template.yml"]
    generateFiles:
      - sourcePath: manifests-template.yml
[...]

PR #4958, opened 2023-08-18
0.12: fix: detect overlapping `targetPath` in `generateFiles`

vvagaytsev:
This PR implements some additional rules to the module overlap detection logic.
Now, if any modules have the same resolved `generateFiles[].targetPath`, a configuration error will be thrown.
Otherwise, the same target file will be overwritten multiple times with different contents.

Addresses #4921 for `0.12`.

This PR must be ported to the `main` branch to fix the issue above in `0.13`.

It might be easier to review it commit-by-commit. Some refactoring was done in separate commits.
All commits will be squashed before merging.


core/src/util/module-overlap.ts
```diff
@@ -0,0 +1,239 @@
...
+      if (!compare.generateFiles) {
+        return []
+      }
+      const compareTargetPaths = resolveGenerateFilesTargetPaths(compare.path, compare.generateFiles)
+      return intersection(targetPaths, compareTargetPaths)
+    }
+
+    const overlaps: ModuleConfig[] = []
+    const generateFilesOverlaps: string[] = []
+    for (const enabledModule of enabledModules) {
```
thsig, line 102: Since this is `O(#modules^2)`, a simple optimization to halve the number of comparisons would be to only look for overlaps between two modules if the second module's name comes after the first.

Currently, we're doing duplicate comparisons (we're looking for overlaps between every `a, b` and `b, a`, but we only need to do one of those comparisons).

Not sure how much of a performance impact this would have, but some of our users have a _lot_ of modules.

thsig (approved):
Great stuff, appreciate the detailed tests there!

Added a suggested performance optimization there and suggested some textual changes to user-facing messages, but otherwise we're good to go.

thsig:
Btw, once you've addressed the PR comments, you can apply them to the 0.13 version of the PR and I'll approve that one too.

thsig (dismissed):

thsig (approved):

PR #4959, opened 2023-08-18
perf: Performance improvements

TimBeyer:

This PR adds a bunch of performance improvements, and replaces Bluebird with native promises.

## Bluebird
Let's start off with Bluebird: The bluebird repo itself [recommends moving to native promises](https://github.com/petkaantonov/bluebird#%EF%B8%8Fnote%EF%B8%8F) and bluebird is no longer seeing any changes. Meanwhile native promises have seen big performance improvements and also usually show up better in stack traces and in performance profiles.

## Repo scan tree indexing
The main performance improvement comes for `GARDEN_GIT_SCAN_MODE=repo`.
It now indexes the list of files into a tree structure for faster access yielding large improvements in graph resolution speeds for large projects.

Here's a comparison between the current `main` version and the updated one, run on an M2 mac on a repo of 11032 with 500 modules.

Old:
```
time GARDEN_GIT_SCAN_MODE=repo gdev validate
Validate ✔️

Project is configured with `apiVersion: garden.io/v0`, running with backwards compatibility.
ℹ garden               → Running in Garden environment dev.default
Scanning repository at /Users/tim/Development/garden/garden-large-repo-generator
✔ providers            → Cached (took 2.1 sec)
ℹ graph                → Scanning repository at /Users/tim/Development/garden/garden-large-repo-generator
✔ graph                → Done (took 16.5 sec)

OK ✔️
GARDEN_GIT_SCAN_MODE=repo ~/Development/garden/garden/bin/garden validate  76.32s user 7.61s system 133% cpu 1:02.88 total
```

New:
```
time GARDEN_GIT_SCAN_MODE=repo gdev validate
Validate ✔️

Project is configured with `apiVersion: garden.io/v0`, running with backwards compatibility.
ℹ garden               → Running in Garden environment dev.default
Scanning repository at /Users/tim/Development/garden/garden-large-repo-generator
✔ providers            → Cached (took 2.1 sec)
ℹ graph                → Scanning repository at /Users/tim/Development/garden/garden-large-repo-generator
✔ graph                → Done (took 8 sec)

OK ✔️
GARDEN_GIT_SCAN_MODE=repo ~/Development/garden/garden/bin/garden validate  64.68s user 6.62s system 136% cpu 52.410 total
```

As you can see the overall performance improvements are still moderate, but the graph resolution is almost 2x faster.

## GC Improvements
We increase `max-semi-space` to 64M which should lead to less GC pauses and better performance at the cost of slightly higher memory consumption.

See https://github.com/nodejs/node/blob/main/doc/api/cli.md#useful-v8-options
Also see https://www.alibabacloud.com/blog/better-node-application-performance-through-gc-optimization_595119 and https://github.com/nodejs/node/issues/42511 for some details on impact.

Fixes #


core/src/vcs/file-tree.ts
```diff
@@ -0,0 +1,120 @@
...
+  addFile(file: VcsFile): boolean {
+    if (!file.path.startsWith(`/${this.ownPath}`)) {
+      return false
+    }
+
+    this.files.push(file)
+
+    const relativePath = file.path.slice(this.ownPath.length)
+    // We use absolute paths so the first part of the split is always an empty string
+    const [_, subpathSegment, nextSegment] = relativePath.split(path.sep)
```
TimBeyer, line 34: Here I wasn't sure if we normalize everything to POSIX paths also for glob compatibility, or if we do need to use `path.sep` so that windows paths get split correctly as well.

That probably would also impact the `.startsWith` part above.

vvagaytsev:
Wow! This is super awesome! 🚀 Huge thanks for such a great job! ✨ 
I need more time to take a closer look in review it in detail :)

core/src/commands/custom.ts
```diff
@@ -242,7 +241,7 @@ export async function getCustomCommands(log: Log, projectRoot: string) {
   const rootFiles = await listDirectory(projectRoot, { recursive: false })
   const paths = rootFiles.filter(isConfigFilename).map((p) => join(projectRoot, p))
 
-  const resources = flatten(await Bluebird.map(paths, (path) => loadConfigResources(log, projectRoot, path)))
+  const resources = flatten(await Promise.all(paths.map((path) => loadConfigResources(log, projectRoot, path))))
```
vvagaytsev, line 244: We don't need explicit `async` inside the `map` function here, right? By the way, we use explicit `async` in the other code places. Should we use the same style everywhere?

core/src/commands/workflow.ts
```diff
@@ -90,7 +89,7 @@ export class WorkflowCommand extends Command<Args, {}> {
     const files = resolveTemplateStrings(workflow.files || [], templateContext)
 
     // Write all the configured files for the workflow
-    await Bluebird.map(files, (file) => writeWorkflowFile(garden, file))
+    await Promise.all(files.map((file) => writeWorkflowFile(garden, file)))
```
vvagaytsev, line 92: [Same](https://github.com/garden-io/garden/pull/4959/files#r1299966786) here.

core/src/garden.ts
```diff
@@ -580,7 +579,7 @@ export class Garden {
   }
 
   async getRegisteredPlugins(): Promise<GardenPluginSpec[]> {
-    return Bluebird.map(this.registeredPlugins, (p) => loadPlugin(this.log, this.projectRoot, p))
+    return Promise.all(this.registeredPlugins.map((p) => loadPlugin(this.log, this.projectRoot, p)))
```
vvagaytsev, line 582: [Same](https://github.com/garden-io/garden/pull/4959/files#r1299966786) here.

vvagaytsev: Ah, there are many other places like this. So I think we can just ignore it, and use both styles.

core/src/commands/cloud/users/users-create.ts
```diff
@@ -144,39 +144,40 @@ export class UsersCreateCommand extends Command<Args, Opts> {
...
-        if (asyncBatch > currentAsyncBatch) {
-          currentAsyncBatch = asyncBatch
-          cmdLog.info({ msg: `Creating users... → Batch ${currentAsyncBatch}/${nAsyncBatches}` })
-        }
-        count++
-        try {
-          const body: CreateUserBulkRequest = {
-            users: userBatch,
-            addToGroups,
+    const limit = pLimit(concurrency)
```
vvagaytsev, line 147: The same can be done with `pMap`, right? I'm just curious. Both approaches look good and can co-exist :)

core/src/commands/custom.ts
TimBeyer: Some need an explicit async when you actually use await inside but here we just return a promise so it's not needed.

vvagaytsev (approved):
Amazing work! Thank you! I've left a few non-blocking comments, LGTM! 💯

core/src/commands/cloud/users/users-create.ts
TimBeyer: Yes this was done before I found pMap. We could rewrite it though.

vvagaytsev: Nice! Feel free to rewrite this :) Pls ping me again when you need another approval :)

vvagaytsev (approved):
💯 🚀

PR #4965, opened 2023-08-22
fix: correct otel collector binary definition for arm64 linux

vvagaytsev:

Fixes #

### DIFF
--- a/core/src/plugins/otel-collector/otel-collector.ts
+++ b/core/src/plugins/otel-collector/otel-collector.ts
@@ -86,7 +86,7 @@ gardenPlugin.addTool({
     },
     {
       platform: "linux",
-      architecture: "arm4",
+      architecture: "arm64",
       url: "https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.80.0/otelcol-contrib_0.80.0_linux_arm64.tar.gz",
       sha256: "19d878166dbc39821f11b6a7c2ed896726c8d5ac6c15108b66d8e874efa8db85",
       extract: {


TimBeyer (approved):

shumailxyz (approved):
Good catch @stefreak. Thanks for fixing @vvagaytsev :)

core/src/plugins/otel-collector/otel-collector.ts
```diff
@@ -86,7 +86,7 @@ gardenPlugin.addTool({
     },
     {
       platform: "linux",
-      architecture: "arm4",
+      architecture: "arm64",
```
stefreak, line 89: This is a good fix, but IMO for every change we do, we need to also take measures to prevent a similar mistake in the future, for example adding a test.

In this case it would be enough to change the type of the `ToolBuildSpec` to allow only valid architectures: https://github.com/garden-io/garden/blob/main/core/src/plugin/tools.ts#L15C1-L16C1

We could also reject platform/architecture duplicates and have similar sanity checks in the `gardenPlugin.addTool` if not already there.

vvagaytsev: Great point! We should do that at some point. Can you create a GH task for that, pls?

stefreak: @vvagaytsev I created a PR instead: https://github.com/garden-io/garden/pull/4968

On the way I found another binary with wrong architecture (aarch64 in maven, instead of amd64) – this has been fixed along the way and now this shouldn't happen anymore. Another class of bugs gone for good 🥳

vvagaytsev: Thank you!

PR #4967, opened 2023-08-22
fix: checkout repo before prerelease

TimBeyer:
We didn't checkout the code for the prerelease step so the added tagging now fails.
This should fix it.

Fixes #


vvagaytsev (approved):

stefreak (approved):
Oops, nice catch 👍

PR #4975, opened 2023-08-23
fix: avoid action execution for the static outputs references of implicit dependencies.

### LINKED ISSUE #4920: [BUG]: Garden is building ALL Docker images on namespace cleanup in `v0.13`
### Current Behavior

As title says, Garden is building ALL Docker images from every submodule before removing deployed services on namespace cleanup in `v0.13`.

### Expected behavior

No docker images are built when cleaning up namespace, just like in `v0.12`.

### Reproducible example

See `cleanup-ns-bug` branch:
https://github.com/xdvpser/bonsai-bug-repro/tree/cleanup-ns-bug

### Additional context
Logs:
```
$ garden cleanup ns --dependants-first
Cleanup namespace 

ℹ garden               → Running in Garden environment local.default
✔ providers            → Cached (took 1.1 sec)
Deleting deployments...
✔ build.my-container   → Connected to sync target Build sync Pod
ℹ build.my-container   → [sync]: Sync disconnected
ℹ build.my-container   → [sync]: Scanning files to sync
✔ build.my-container   → [sync]: Completed initial sync from Build my-container build path to Build sync Pod
ℹ build.my-container   → [sync]: Watching for changes
ℹ build.my-container   → File sync to cluster complete
ℹ build.my-container   → Buildkit Building image my-container:v-c0b412250f...
✔ build.my-container   → Done (took 10.1 sec)
ℹ deploy.helloworld    → Cleaning up...
✔ deploy.helloworld    → Service deleted (took 0.3 sec)
✔ deploy.helloworld    → Done (took 0.3 sec)
ℹ deploy.helloworld    → Done!
Done

Cleaning up environments...
ℹ kubernetes           → Deleting namespace root-default (this may take a while)
✔ kubernetes           → Done
Done

Done!
```

### Your environment

* **OS:** Ubuntu 22.04
* **How I'm running Kubernetes:** Self-managed setup in AWS using Kops. Bug is also appearing with local-kubernetes plugin using minikube.

`garden version: 0.13.9`

shumailxyz:
With this PR, we avoid marking an action's implicit dependency as needing execution if a static output of dependency is referenced.

Fixes #4920

### DIFF
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -742,13 +742,13 @@ function dependenciesFromActionConfig(
     // also avoid execution when referencing the static output keys of the ref action type.
     // e.g. a helm deploy referencing container build static output deploymentImageName
     // ${actions.build.my-container.outputs.deploymentImageName}
-    const actionRef = { name: ref.name, kind: ref.kind }
-    const refActionType = configsByKey[actionReferenceToString(actionRef)]?.type
+    const actionRefKey = actionReferenceToString(ref)
+    const refActionType = configsByKey[actionRefKey]?.type
     let staticOutputKeysForRef: string[] = []
     if (refActionType) {
-      const refDefinition = actionTypes[ref.kind][refActionType]?.spec
-      staticOutputKeysForRef = refDefinition?.staticOutputsSchema
-        ? describeSchema(refDefinition.staticOutputsSchema).keys
+      const refActionSpec = actionTypes[ref.kind][refActionType]?.spec
+      staticOutputKeysForRef = refActionSpec?.staticOutputsSchema
+        ? describeSchema(refActionSpec.staticOutputsSchema).keys
         : []
     }
 
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -742,7 +742,7 @@ function dependenciesFromActionConfig(
     // also avoid execution when referencing the static output keys of the ref action type.
     // e.g. a helm deploy referencing container build static output deploymentImageName
     // ${actions.build.my-container.outputs.deploymentImageName}
-    const actionRef = { name: ref.name, kind: ref.kind}
+    const actionRef = { name: ref.name, kind: ref.kind }
     const refActionType = configsByKey[actionReferenceToString(actionRef)]?.type
     let staticOutputKeysForRef: string[] = []
     if (refActionType) {
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -42,7 +42,7 @@ import { ConfigurationError, InternalError, PluginError, ValidationError } from 
 import { overrideVariables, type Garden } from "../garden"
 import type { Log } from "../logger/log-entry"
 import type { ActionTypeDefinition } from "../plugin/action-types"
-import { getActionTypeBases } from "../plugins"
+import { ActionDefinitionMap, getActionTypeBases } from "../plugins"
 import type { ActionRouter } from "../router/router"
 import { getExecuteTaskForAction } from "../tasks/helpers"
 import { ResolveActionTask } from "../tasks/resolve-action"
@@ -297,7 +297,7 @@ export const actionFromConfig = profileAsync(async function actionFromConfig({
     })
   }
 
-  const dependencies = dependenciesFromActionConfig(log, config, configsByKey, definition, templateContext)
+  const dependencies = dependenciesFromActionConfig(log, config, configsByKey, definition, templateContext, actionTypes)
 
   if (config.exclude?.includes("**/*")) {
     if (config.include && config.include.length !== 0) {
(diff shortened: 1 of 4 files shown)


core/src/graph/actions.ts
```diff
@@ -738,8 +739,24 @@ function dependenciesFromActionConfig(
...
+        ? describeSchema(refDefinition.staticOutputsSchema).keys
+        : []
+    }
 
-    if (ref.fullRef[3] === "outputs" && outputKey && !staticKeys?.includes(<string>outputKey)) {
+    if (
+      ref.fullRef[3] === "outputs" &&
+      outputKey &&
+      !staticKeys?.includes(<string>outputKey) &&
+      !staticOutputKeysForRef?.includes(<string>outputKey)
```
vvagaytsev, line 758: A nit on the code-style. Repeated cast to `string` can be avoided if we declare `output` key in L730 as follows:
```
const outputKey = ref.fullRef[4] as string
```

core/src/graph/actions.ts
```diff
@@ -738,8 +739,24 @@ function dependenciesFromActionConfig(
         continue
       }
     }
+    // also avoid execution when referencing the static output keys of the ref action type.
+    // e.g. if a helm deploy is referencing container build static output
+    // deploymentImageName: ${actions.build.my-container.outputs.deploymentImageName}
+    const refActionType = configsByKey[`${ref.kind.toLowerCase()}.${ref.name.toLowerCase()}`]?.type
```
vvagaytsev, line 745: Do we need to lowercase both `ref.kind` and `ref.name` here? It looks like we use `actionReferenceToString` helper to do other lookups to `configsByKey` dictionary here.

Also, there are 2 `actionReferenceToString` helpers defined in the project, and one of them (from `types.ts`) is not used. Let's remove the unused duplicate.

shumailxyz: Good point, i'll simplify this.

shumailxyz: - unused `actionReferenceToString` removed in https://github.com/garden-io/garden/pull/4975/commits/f37e7b6327f85d711684e533311f165ef39019e8
- refactored to use helper in https://github.com/garden-io/garden/pull/4975/commits/de295ecad344cc8cb73ffbccc71f28fe1f398813

core/src/graph/actions.ts
```diff
@@ -738,8 +739,25 @@ function dependenciesFromActionConfig(
         continue
       }
     }
+    // also avoid execution when referencing the static output keys of the ref action type.
+    // e.g. a helm deploy referencing container build static output deploymentImageName
+    // ${actions.build.my-container.outputs.deploymentImageName}
+    const actionRef = { name: ref.name, kind: ref.kind}
```
vvagaytsev: It looks like this is unnecessary; `ref` has type `ActionReferenceTemplate` that extends `ActionReference`. So, the following line can be rewritten as `const refActionType = configsByKey[actionReferenceToString(ref)]?.type`.

I'm not sure if it's better or not to extract local var `const refKey = actionReferenceToString(ref)` here. The inlined version also looks good.

core/src/graph/actions.ts
```diff
@@ -738,8 +739,25 @@ function dependenciesFromActionConfig(
...
+    const refActionType = configsByKey[actionReferenceToString(actionRef)]?.type
+    let staticOutputKeysForRef: string[] = []
+    if (refActionType) {
+      const refDefinition = actionTypes[ref.kind][refActionType]?.spec
```
vvagaytsev, line 749: Minor nit. This can be named `refActionSpec`, just for slightly better clarity :)

shumailxyz: Sure, addressed in https://github.com/garden-io/garden/pull/4975/commits/24c661d96517ae4817a1d620389dc12c7bd70126

core/src/graph/actions.ts
shumailxyz: Thanks, addressed in https://github.com/garden-io/garden/pull/4975/commits/24c661d96517ae4817a1d620389dc12c7bd70126

core/src/graph/actions.ts
```diff
@@ -738,8 +739,25 @@ function dependenciesFromActionConfig(
...
+        ? describeSchema(refActionSpec.staticOutputsSchema).keys
+        : []
+    }
 
-    if (ref.fullRef[3] === "outputs" && outputKey && !staticKeys?.includes(<string>outputKey)) {
+    if (
+      ref.fullRef[3] === "outputs" &&
+      outputKey &&
+      !staticKeys?.includes(outputKey) &&
+      !staticOutputKeysForRef?.includes(outputKey)
```
vvagaytsev, line 759: Ah, sorry about too-many-comments :D Just one more minor thing I've just noticed. Both `staticKeys` and `staticOutputKeysForRef` are always defined as arrays, and cannot be `null` or `undefined`, right? In that case, we can omit the null-safe checker `?`.

We can also unify the naming for both vars. something like `staticOutputKeys` and `refStaticOutputKeys`. Or just `staticKeys` and `refStaticKeys`. But that's totally optional.

shumailxyz: All good :D Thanks in fact for checking so meticulously :)

Yeah, those can't be null/undefined. Addressed that in https://github.com/garden-io/garden/pull/4975/commits/97761b4915bcb5c6400df9e60e9dce95db3e96a5

vvagaytsev (approved):
LGTM! 🚀 
Thanks for fixing this! 👍

PR #4981, opened 2023-08-24
chore: remove the create module command

stefreak:
This prevents new users getting confused and working with modules,
instead of actions as we prefer.

The alternative would have been to implement a "create actions" command
and deprecating the "create module" command; but we think the create
actions command can be implemented later as it's only nice to have.

Co-authored-by: Tim <tim@garden.io>

Fixes #


vvagaytsev (changes requested):
Well done!
I've found some leftovers, please check:
* https://github.com/garden-io/garden/blob/remove-create-module-command/.circleci/config.yml#L548
* https://github.com/garden-io/garden/blob/remove-create-module-command/core/src/cli/helpers.ts#L474
* https://github.com/garden-io/garden/blob/remove-create-module-command/docs/using-garden/using-the-cli.md?plain=1#L244

stefreak:
> Well done! I've found some leftovers, please check:
> 
> * https://github.com/garden-io/garden/blob/remove-create-module-command/.circleci/config.yml#L548
> * https://github.com/garden-io/garden/blob/remove-create-module-command/core/src/cli/helpers.ts#L474
> * https://github.com/garden-io/garden/blob/remove-create-module-command/docs/using-garden/using-the-cli.md?plain=1#L244

@vvagaytsev Good catch! Hm, great that the tests passed while this was still left over. Is this due to the wrong exit codes for unrecognized commands?

vvagaytsev:
> > Well done! I've found some leftovers, please check:
> > 
> > * https://github.com/garden-io/garden/blob/remove-create-module-command/.circleci/config.yml#L548
> > * https://github.com/garden-io/garden/blob/remove-create-module-command/core/src/cli/helpers.ts#L474
> > * https://github.com/garden-io/garden/blob/remove-create-module-command/docs/using-garden/using-the-cli.md?plain=1#L244
> 
> @vvagaytsev Good catch! Hm, great that the tests passed while this was still left over. Is this due to the wrong exit codes for unrecognized commands?

Probably, yes. Can this change break any existing workflows? `garden create module` is an interactive command (i.. can be used in automated workflows), but if all the necessary params are supplied it might not ask for any input.

We should emphasize that braking change in the release notes.

stefreak:
I agree, but I believe it's unlikely that people are using this command in scripts or workflows as it does not do any heavy lifting, it just writes a simple yaml file.

twelvemo (approved):

PR #4985, opened 2023-08-25
improvement: more useful "create project" yaml

stefreak:

The previous yaml generated by the docs-generator looked like this:

```yaml
# Documentation about Garden projects can be found at https://docs.garden.io/using-garden/projects
# Reference for Garden projects can be found at https://docs.garden.io/reference/project-config

# The Garden apiVersion for this project.
#
# The value garden.io/v0 is the default for backwards compatibility with
# Garden Acorn (0.12) when not explicitly specified.
#
# Configuring garden.io/v1 explicitly in your project configuration allows
# you to start using the new Action configs introduced in Garden Bonsai (0.13).
#
# Note that the value garden.io/v1 will break compatibility of your project
# with Garden Acorn (0.12).
apiVersion: garden.io/v1

# Indicate what kind of config this is.
kind: Project

# The name of the project.
name: foobar

# A list of environments to configure for the project.
environments:
  - # The name of the environment.
    name: default

# A list of providers that should be used for this project, and their configuration. Please refer to individual
# plugins/providers for details on how to configure them.
providers:
  - # The name of the provider plugin to use.
    name: local-kubernetes
```

While it is cool that we are re-using the descriptions from our docs,
there are some problems with this yaml code:
- The comments often do not add any valuable information. For example
  it's already clear that `name: foobar` is the name of the project.
- Another problem is that the description in the reference docs is not
  necessarily the information a user needs when getting started. This
becomes apparent in the comment above `apiVersion`, for example.

This commit simplifies the `create project` command implementation to
just write a file with hardcoded yaml. The comments are hand-crafted for
a good getting-started experience.

This commit also updates the docs that refer to the "create project"
command.

The new yaml looks like this:
```yaml
# Documentation about Garden projects can be found at https://docs.garden.io/using-garden/projects
# Reference for Garden projects can be found at https://docs.garden.io/reference/project-config

apiVersion: garden.io/v1
kind: Project
name: test

defaultEnvironment: local

# Environments typically represent different stages of your development and deployment process.
environments:
  # Use this environment to develop in your local Kubernetes solution of choice.
  # Installation instructions and list of supported local Kubernetes environments: https://docs.garden.io/kubernetes-plugins/local-k8s/install
  - name: local

  # Use this environment to develop in remote, production-like environments that scale with your stack.
  # This means you don't need any dependencies on your local machine, even the builds can be performed remotely.
  # It enables sharing build and test caches with your entire team, which can significantly speed up pipelines and development.
  - name: remote
    defaultNamespace: garden-remote-${local.username}

  - name: staging
    # Ask before performing potentially destructive commands like "deploy".
    production: true
    defaultNamespace: staging

# Providers make action types available in your Garden configuration and tell Garden how to connect with your infrastructure.
# For example the kubernetes and local-kubernetes providers allow you to use the container, helm and kubernetes action types.
# All available providers and their configuration options are listed in the reference docs: https://docs.garden.io/reference/providers
providers:
  - name: local-kubernetes
    environments:
      - local

  # To configure the remote kubernetes providers, follow the steps at https://docs.garden.io/kubernetes-plugins/remote-k8s
  - name: kubernetes
    environments:
      - remote
    context: # ... your remote development kubecontext here
  - name: kubernetes
    environments:
      - staging
    context: # ... your staging kubecontext here

# Next step: Define actions to tell Garden how to build, test and deploy your code.
# You can find out more here: https://docs.garden.io/using-garden/actions
```

Fixes #

### DIFF
--- a/core/src/commands/create/create-project.ts
+++ b/core/src/commands/create/create-project.ts
@@ -15,7 +15,6 @@ import { isDirectory } from "../../util/fs"
 import { loadConfigResources } from "../../config/base"
 import { resolve, basename, relative, join } from "path"
 import { GardenBaseError, ParameterError } from "../../exceptions"
-import { renderProjectConfigReference } from "../../docs/config"
 import { addConfig } from "./helpers"
 import { wordWrap } from "../../util/string"
 import { PathParameter, StringParameter, BooleanParameter, StringOption } from "../../cli/params"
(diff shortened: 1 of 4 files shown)


core/src/commands/create/create-project.ts
```diff
@@ -138,30 +137,56 @@ export class CreateProjectCommand extends Command<CreateProjectArgs, CreateProje
...
+    # Environments typically represent different stages of your development and deployment process.
+    environments:
+      - name: local
+        defaultNamespace: garden-local
+
+      # Use this environment to develop in remote, production-like environments that scale with your stack.
+      # This means you don't need any dependencies on your local machine, even the builds can be performed remotely.
+      # It enables sharing build and test caches with your entire team, which can significantly speed up pipelines and development.
+      - name: remote
+        defaultNamespace: garden-remote-\${local.username}
```
stefreak, line 164: The backslash seems to end up in the final yaml file. We need to fix that before merging.

TimBeyer: Is that true? In theory that should just escape the `$` so it doesn't get templated in JS.

TimBeyer: image

TimBeyer (approved):

core/src/commands/create/create-project.ts
stefreak: I'm puzzled about this too. But I ran the command "garden create project" with the dev build, and the backslash was there. 🤔

stefreak: @TimBeyer I now worked around this mystery and updated the tests.

core/src/commands/create/create-project.ts
```diff
@@ -138,30 +137,56 @@ export class CreateProjectCommand extends Command<CreateProjectArgs, CreateProje
...
+    # Reference for Garden projects can be found at ${projectReferenceURL}
+
+    apiVersion: ${GardenApiVersion.v1}
+    kind: Project
+    name: ${name}
+
+    defaultEnvironment: local
+
+    # Environments typically represent different stages of your development and deployment process.
+    environments:
```
twelvemo, line 156: I would love to also have a quick description of the local environment so it is more consistent e.g.:
```suggestion
    environments:
    
    # Use this environment to develop in your local Kubernetes solution of choice. For more information about
    # installation instructions and against which local Kubernetes solutions Garden is tested see:
    # ${localKubernetesInstallationURL}
```

core/src/commands/create/create-project.ts
```diff
@@ -138,30 +137,56 @@ export class CreateProjectCommand extends Command<CreateProjectArgs, CreateProje
...
+      - name: local
+        defaultNamespace: garden-local
+
+      # Use this environment to develop in remote, production-like environments that scale with your stack.
+      # This means you don't need any dependencies on your local machine, even the builds can be performed remotely.
+      # It enables sharing build and test caches with your entire team, which can significantly speed up pipelines and development.
+      - name: remote
+        defaultNamespace: garden-remote-${"${local.username}"}
+
```
twelvemo, line 165:
```suggestion
    # You could also define a staging environment, which is shared by several developers.
```

core/src/commands/create/create-project.ts
```diff
@@ -138,30 +137,56 @@ export class CreateProjectCommand extends Command<CreateProjectArgs, CreateProje
...
+      - name: staging
+        # Ask before performing potentially destructive commands like "deploy".
+        production: true
+        defaultNamespace: staging
+
+    # Available providers are listed in the reference docs: ${providersReferenceURL}
```
twelvemo, line 171:
```suggestion
    # With Providers you tell Garden a few common things like for example against which Kubernetes context to
    # run your deploys for your environments or how to build your container images.
    # Available providers and their configuration options are listed in the reference docs: 
    # ${providersReferenceURL}
```

core/src/commands/create/create-project.ts
```diff
@@ -138,30 +137,56 @@ export class CreateProjectCommand extends Command<CreateProjectArgs, CreateProje
...
-      },
-    })
-
     const projectDocURL = `${DOCS_BASE_URL}/using-garden/projects`
     const projectReferenceURL = `${DOCS_BASE_URL}/reference/project-config`
-    yaml =
-      dedent`
+    const providersReferenceURL = `${DOCS_BASE_URL}/reference/providers`
+    const remoteK8sReferenceURL = `${DOCS_BASE_URL}/kubernetes-plugins/remote-k8s`
+    const actionsGettingStartedURL = `${DOCS_BASE_URL}/using-garden/actions`
```
twelvemo, line 144: If you like my suggestion on local Kubernetes below, we'd need to add this. Can't make a suggestion, because i can't simply add a line and not change one.
```
    const localKubernetesInstallationURL = `${DOCS_BASE_URL}/kubernetes-plugins/local-k8s/install`
```

twelvemo (approved):
I think this is great! I have a few suggestions around comments, feel free to adept or discard.

core/src/commands/create/create-project.ts
stefreak: I decided not to incorporate this one, because it does not add a lot of information– Here I'd value conciseness and usefulness over consistency

core/src/commands/create/create-project.ts
stefreak: I iterated on your suggestions and amended the commit, what do you think about the current version?

PR #4986, opened 2023-08-25
improvement: always offer -f as alias with --force

stefreak:
`build` offered the `-f` alias for `--force`, but all other commands did
not. I found myself trying to run `garden deploy -f` multiple times and
it failed, so let's just make it consistent.

Fixes #


shumailxyz (approved):
Totally agreed. Thanks for adding the alias 👍

Issue #4729, opened 2023-06-30
[FEATURE]: kubernetes-container module


ShankyJS:
### Background / Motivation

The `container`  type started as a basic implementation that could serve multiple providers instead of tailoring it exclusively to Kubernetes.

As time passed, Garden users continued using this module as it's simple and doesn't require too much management, providing a cleaner syntax with easier adaptability. However, we have reached a point where we can no longer extend the container type, leaving users with open features requests pending.
 
This is why, as part of this [discussion](https://github.com/garden-io/garden/issues/4034), we are requesting the creation of a `kubernetes-container` type that expands the simple `container` one.

### What should the user be able to do?

Users should be able to use this new `kubernetes-container` type but have access to more in-deep customization of different resources, creating a `production` ready type that is war-tested, has lots of built-in capabilities, and should be extensible. 

### Why do they want to do this? What problem does it solve?

By removing the complexity of Helm, or Kubernetes manifests, users could only focus on building their `kubernetes-container` type once and then concentrate exclusively on coding their applications. 

Also, this new type could have features like horizontalPodAutoscaler, VPAs, NetworkPolicies, and other specific K8s resources that could make this type even more powerful.

### Suggested Implementation(s)

I'm not sure how we are templating our resources into Kubernetes, but I would build a powerful Helm Chart (Like Helmet), for example, with everything a user might need to deploy.

- Deployment
- Service
- Ingress
- PVC/PV
- HPA
- VPA
- NetworkPolicies
- ConfigMaps
- Secrets
- Etc.

Example of a strong [Helm Chart](https://github.com/companyinfo/helm-charts/tree/main/charts/helmet) that offers multiple features.

And then template it into Garden with the details the user will provide.

### How important is this feature for you/your team?

🌹 It's a nice to have, but nice things are nice 🙂

chulkilee:
What would be added to / removed from k8s manifest attributes when using garden's `kubernetes-container`? Probably templating, dependencies / some controls (e.g. disable)?

I'm wondering how it would remove the K8s manifest complexity too.

porterchris:
@eysi09 I'm curious how you've used/re-used action templates, pulled in shared manifest files, etc. Do you have an example project or could you put one together and push it up to show how you've handled the `kubernetes` type? We've found that the jump from `container` to `kubernetes` is quite far and sometimes frustrating.

eysi09:
[...]

Issue #1199, opened 2019-09-17
Allow relative paths outside of the `garden.yaml` root


kvokka:
## Feature request

Facing the #1190 issue, I do not have `dirty option`. Lack of this ability block the entire integration. In the future a developer may find some other situations, when it is required.

All usage of `../` in path is prohibited, even when i know, what and why i'm doing it.

I think some project-wide flag for this may be very handy. Of course, you will write that it is a bad practice and developer should avoid it, but it may keep "show go on".

edvald:
Understood. The reasons for not allowing paths outside the project are indeed partly to avoid "bad practices" but are also technical. We need to be able to assert that any referenced paths are within git repos, and in most cases in the same repo as the project. This becomes much more difficult when we allow arbitrary paths. For modules, staging builds also becomes difficult if one can reference arbitrary paths.

Which field(s) were you trying to set this way?

kvokka:
I understand, that allowing this is bad practice from the very beginning. But in other hand, developer need some `last resort` tool.

Given the project root is `dev/garden.yaml`
And service1 root is `dev/core/garden.yaml`
When `dev/support/core/garden.yaml` with:

```yaml
---
kind: Module
description: Image for core
type: container
name: poker-core
dockerfile: ../../docker/development/Dockerfile
---
kind: Module
description: Chart for core
type: helm
name: core
chartPath: ../../charts/core
...
```

I tried to use some external configuration path, which is actually impossible.

Please, understand me right, I'm totally agree that it is a bad practice. But uglifying is better than blocking.

Maybe switch to #1201 ? This should allow to add some configs to the root. Not very pretty, but should work for a while.

edvald:
I can look into both options. There are several edge-cases we need to consider, especially as they relate to in-cluster building, but either or both should be feasible.

ox-michaelradionov:
I am not sure if I have the same issue, but I've just started trying out Garden and have some issues with relative paths. I have a set of microservices, each in it's own repository, and a separate repository with Kubernetes manifests. All of them are fetched to the same directory resulting in the following dir structure:

```
- manifests/
- microservice1/
- microservice2/
```

If I am following Garden requirements correctly, I must have a Project which can then reference Modules. I've created a "garden.yml" with Project in "manifests/" folder, as it used to be my root for deploying the application, and also created "garden.yml" for each microservice of "Module" kind.

```
- manifests/
   - garden.yml             # kind: Project
- microservice1/
[...]

Issue #853, opened 2019-06-17
Allow specifying a different docker build context


diasjorge:
### Current Behavior

The Dockerfiles must be at the root of the project

### Expected behavior

A user is able to specify a context pointing to another directory like with docker-compose

I can't change our project layout to have different structure like in your examples.

I'd like to be able to specify all garden configuration in a subdirectory like:

garden/garden.yml
garden/module1/garden.yml
garden/module2/garden.yml

And have module1 and module2 build from a parent directory.

Currently with docker-compose we specify the context like: ../../

### Your environment

`garden version`

0.9.12

`kubectl version`

Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.3", GitCommit:"5e53fd6bc17c0dec8434817e69b04a25d8ae0ff0", GitTreeState:"clean", BuildDate:"2019-06-07T09:55:27Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"12+", GitVersion:"v1.12.7-gke.10", GitCommit:"8d9b8641e72cf7c96efa61421e87f96387242ba1", GitTreeState:"clean", BuildDate:"2019-04-12T22:59:24Z", GoVersion:"go1.10.8b4", Compiler:"gc", Platform:"linux/amd64"}

`docker version`

Client: Docker Engine - Community
 Version:           18.09.2
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        6247962
 Built:             Sun Feb 10 04:12:39 2019
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.2
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.6
  Git commit:       6247962
  Built:            Sun Feb 10 04:13:06 2019
  OS/Arch:          linux/amd64
  Experimental:     false

edvald:
Hey @diasjorge! That's good feedback. We've done some work to make placement of code+configs more flexible, but we do currently assume source files for a module to be within the directory where the config is. The reason we do that is that we "stage" builds by copying source files into a temporary directory, and allowing arbitrary locations makes that quite a bit more complex to do.

We have wanted to add a `buildRoot` (or similar) field to the `container` module type, but we'd likely still need it to be within the module path.

[...]

PR #4992, opened 2023-08-29
docs: add k8s Deploy action example with shared manifests

eysi09:

From the commit message:

This commit adds an example that uses the `kubernetes` deploy action and shares the same manifests between a couple of actions. This is achieved by using Garden template strings in the manifests.

The example is a three tier web app version of the vote example with "web", "API", and "DB" components that I think we should consider using by default.

It's not as complex as the main vote example which tends to draw attention away from the functionality being demonstrated, and not as trivial as the basic "demo-project".

Three tier web apps are also a common pattern that a lot of people will be familiar with.

And lastly, this is also the same example we use in the "How Garden Works" video series so it will become familiar to users eventually.

Fixes #

Added this example following the conversation here: https://github.com/garden-io/garden/issues/4729

We'll also be adding actual docs on the topic, this example is just the first step and can be used as a reference in those docs.


worldofgeese:
I have now tested this locally to confirm everything runs and is accessible.

I see this as part of a larger initiative to directly address needs in our docs. Currently we suggest the `container` module for a simplified abstract for users. I think there's a space in our docs for positioning this Config Templates example and a Helm library chart example as methods to provide helpful abstractions to downstream (of the Platform Team) consumers.

E.g. a docs page, even a Seed, that walks users through generating shared manifests. This could even be turned into a native `garden` command such as, `garden generate configtemplates` similar to `helm create`.

Your PR mentions,

> - Actions cannot reference files in parent directories. That's why we "hoist" the Garden config for the web and API components to the root (where the manifests reside) and name the config files `api.garden.yml` and `web.garden.yml`.
> - You can also set non primitive values like maps and arrays by
using the `jsonEncode` templating function.

Your first bullet sounds like something we should support in `garden` natively. Do we have a feature request for it? Or any reason why we wouldn't want to implement a feature like this?

For your `jsonEncode` example, could you expand on this with a post-transform example? So users can directly see.

worldofgeese (commented):
See comment above 👀

eysi09:
> I see this as part of a larger initiative to directly address needs in our docs

Absolutely, it is. I just started with the examples because they had been requested. But the overall goal is to update the docs and reference the examples from there. 

> Your first bullet sounds like something we should support in garden natively. Do we have a feature request for it? Or any reason why we wouldn't want to implement a feature like this?

Yes, we do but they're stale. My plan was to open a new one with up to date context.

- https://github.com/garden-io/garden/issues/853
- https://github.com/garden-io/garden/issues/1199

I feel quite strongly that we should support this. The current workaround is to hoist everything to the top which is not nice.

> For your jsonEncode example, could you expand on this with a post-transform example? So users can directly see.

Will do.

eysi09:
@worldofgeese, I added a new commit addressing your comments.

worldofgeese (dismissed):
Looks good to me! 🥇

worldofgeese (approved):
Great shot of clarity and taught me a new key to use for rewriting source path ⚡

PR #4993, opened 2023-08-29
fix(k8s): detect duplicate manifest declarations

stefreak:

Detect duplicate manifest declarations to prevent caching issues due to
accidental duplicate manifest declarations.

For example, take the following deployment action:

```
kind: Deploy
type: kubernetes
name: silly-demo
spec:
  manifests:
  - apiVersion: apps/v1
    kind: Deployment
    metadata:
      name: silly-demo
  files:
  - k8s/deployment.yaml
```

If the declared file contains a Deployment with the same name, Garden
will behave very strangely.

In the first deploy, garden will correctly surface the issue of missing
required values in the inline manifest:

```
Command "/Users/steffen/.garden/tools/kubectl/f8a39834529cd1ed/kubectl --context=docker-desktop apply --output=json -f -" failed with code 1:

The Deployment "silly-demo" is invalid:
* spec.selector: Required value
* spec.template.metadata.labels: Invalid value: map[string]string(nil): `selector` does not match template `labels`
* spec.template.spec.containers: Required value
* spec.selector: Invalid value: "null": field is immutable

The Deployment "silly-demo" is invalid:
* spec.selector: Required value
* spec.template.metadata.labels: Invalid value: map[string]string(nil): `selector` does not match template `labels`
* spec.template.spec.containers: Required value
* spec.selector: Invalid value: "null": field is immutable
```

But the second time we deploy this Garden configuration, it appears to
work fine:
```
ℹ deploy.silly-demo    → Already deployed
```

This is because the other deployment declared in the file was valid,
and kubectl apply has actually applied that configuration. The second
time we run `garden deploy`, Garden picks up that it has already
been deployed and does not bother to continue.

This PR introduces a validation step, where we detect duplicates on the
action-level to prevent issues with our caching system further down the
line. With this PR the configuration above would throw the following
error:

```
✖ deploy.silly-demo    → Failed resolving status for Deploy type=kubernetes name=silly-demo (from module silly-demo) (took 0.04 sec). Here is the output:

────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Duplicate manifest definition: Deployment named silly-demo is declared more than once:

- Deployment silly-demo declared in the file /Users/steffen/viktor/garden-demo/k8s/deployment.yaml (index: 0)
- Deployment silly-demo declared inline in the Garden configuration (filename: /Users/steffen/viktor/garden-demo/garden.yml, index: 0)
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

1 deploy action(s) failed!
```

This is one of the issues @vfarcic discovered in his YouTube video about
Garden (https://youtu.be/BUlrbSxpRTs?t=1124). Thank you for finding the issue
and providing your configuration in a GitHub repo
(https://github.com/vfarcic/garden-demo).


thsig (approved):
Great stuff! Better validation & error messages are a win-win.

PR #4995, opened 2023-08-29
fix(core): input validation for module templates

### LINKED ISSUE #4946: 0.13: [Bug]: Conditional disabled flag is not working with default value
I have a conditional flag "disabled" which needs to pass as an input but only with few services. Here is my config for two services x and y
**Use Case**
X should be disabled in dev environment and Y should work for all envs.

```
kind: RenderTemplate
template: spring-service-module-template
name: x
inputs:
   disabled: ${environment.name == 'dev'}
```

```
kind: RenderTemplate
template: spring-service-module-template
name: y

---
kind: ConfigTemplate
name: spring-service-module-template
inputsSchemaPath: schema.json

configs:
  - type: pulumi
    name: ${parent.name}-service
    disabled: {inputs.disabled}
```

```
schema.json
{
    "type": "object",
    "properties": {
      "excludeModules": {
        "type": "array",
        "items": {
            "type": "string"
        }
      },
      "jdkVersion": {
        "type": "string",
        "default": "8"
      },
      "disabled":{
	"type": "boolean",
	"default":false		
      }
    }
}
```

Following error is coming 
**x-service: Error validating inputs for x-service (services/x/garden.yaml): value at ./disabled must be boolean**

thsig:

Before this fix, we weren't performing a partial resolution pass on module template inputs before validating them against the input schema. This looks like it was a simple oversight (there's definitely no point in partially resolving an always-empty object).

This would result in validation errors e.g. when the schema declares a boolean field but the unresolved template is still a string value.

Fixes #4946.

While looking at the code around input resolution and validation, I noticed that we never apply the input schema (as defined in e.g. `schema.json`) against the fully resolved inputs.

It looks like the idea here is to allow the user to pass through template strings that reference e.g. runtime values, which get resolved later in the flow.

If that's the case, then this may cause some surprising behaviour, since a template string for an input value that's supposed to resolve e.g. to a number or a boolean will fail validation because it's still a (template) string. CC @edvald 

If this is the way we intend it to be, we should mention this wrinkle in our guide for config templates.

### DIFF
--- a/core/src/config/render-template.ts
+++ b/core/src/config/render-template.ts
@@ -114,7 +114,7 @@ export async function renderConfigTemplate({
   templates: { [name: string]: ConfigTemplateConfig }
 }): Promise<RenderConfigTemplateResult> {
   // Resolve template strings for fields. Note that inputs are partially resolved, and will be fully resolved later
-  // when resolving the resolving the resulting modules. Inputs that are used in module names must however be resolvable
+  // when resolving the resulting modules. Inputs that are used in module names must however be resolvable
   // immediately.
   const loggedIn = garden.isLoggedIn()
   const enterpriseDomain = garden.cloudApi?.domain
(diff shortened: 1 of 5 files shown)


thsig:
Btw, thanks @shumailxyz for the excellent repro you posted on the issue!

shumailxyz (approved):
@thsig  Awesome! thanks for fixing this 👍

PR #4996, opened 2023-08-29
fix: move back to js-yaml 3.14.1 for compatibility

### LINKED ISSUE #4929: 0.13: [Bug]: Octal configuration values interpreted as decimal in Bonsai
### Current Behavior

[YAML 1.1](https://yaml.org/spec/1.1/#id862121)-style octal values (e.g. `0777`) are interpreted as decimal in Bonsai configuration.

### Expected behavior

Garden should either mention changes in octal value parsing in the Bonsai migration guide, or interpret both `0777` and `0o777` as octal values.

### Reproducible example

In the `main` branch, apply this patch:
```
diff --git a/examples/build-dependencies/frontend/garden.yml b/examples/build-dependencies/frontend/garden.yml
index 5a8d6aa3f..ffdb086f0 100644
--- a/examples/build-dependencies/frontend/garden.yml
+++ b/examples/build-dependencies/frontend/garden.yml
@@ -35,6 +35,7 @@ spec:
         target: /app/config/
         exclude: [ garden.yml ]
         mode: one-way-replica
+        defaultFileMode: 0666
   ports:
     - name: http
       containerPort: 8080
```

then, `cd examples/build-dependencies/frontend` and run `garden validate`:

```
steffen@MacBook-Pro-3 frontend % garden validate
Validate ✔️

ℹ garden               → Running in Garden environment local.default
✔ providers            → Cached (took 1.3 sec)
✔ graph                → Done (took 0.3 sec)
✖ deploy.frontend      → Failed processing resolve Deploy type=container name=frontend (took 0.17 sec). Here is the output:

────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Error validating spec for Deploy type=container name=frontend (frontend): Mode permission bits out of range. Please specify a number between 0 and 0o777 in octal representation. The number needs to be prefixed with '0o' in YAML, e.g. 0o777.. Available keys: command, args, env, cpu, memory, volumes, privileged, addCapabilities, dropCapabilities, tty, deploymentStrategy, annotations, daemon, sync, localMode, image, ingresses, healthCheck, hotReload, timeout, limits, ports, replicas)
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────


Failed to resolve Deploy type=container name=frontend: Error: resolve Deploy type=container name=frontend failed: Error: Error validating spec for Deploy type=container name=frontend (frontend): Mode permission bits out of range. Please specify a number between 0 and 0o777 in octal representation. The number needs to be prefixed with '0o' in YAML, e.g. 0o777.. Available keys: command, args, env, cpu, memory, volumes, privileged, addCapabilities, dropCapabilities, tty, deploymentStrategy, annotations, daemon, sync, localMode, image, ingresses, healthCheck, hotReload, timeout, limits, ports, replicas)

```

### Workaround

Use [YAML 1.2](https://yaml.org/spec/1.2.2/#24-tags)-style octal notation (e.g. `0o777`)

### Suggested solution(s)

I'm not sure; Did we update our YAML library in the mean time and it now uses the updated 1.2 spec? Can we instruct it to continue to interpret the outdated notation as octal numbers?

### Additional context

This issue surfaced with the stricter validation in https://github.com/garden-io/garden/pull/4824

### Your environment

* **OS:** macOS

`garden version`

thsig:

The parsing of octal values changed in version 4.0 of `js-yaml`, the YAML parsing library we use to load configs. Previously, values like `0777` would be parsed as octal numbers, but in 4.0 they're parsed as decimal numbers.

Since some users want to maintain backwards-compatibility with Garden 0.12, we're moving back to `js-yaml`  3.14.1 (the last 3.x release) to get back to the old octal value parsing behavior.

Fixes #4929.


thsig:
Some context: This is the PR where we updated to `js-yaml` 4.0.

https://github.com/garden-io/garden/pull/4630

It's not quite clear to me what the motivation was behind the update—it's also totally reasonable to update libraries from time to time (this octal problem wasn't obvious at the time).

stefreak:
It looks like the update PR  #4630 resolved the issue #4633

Also, does this mean that we cannot update js-yaml even when there are upstream security fixes or bug fixes? 

Have you considered the alternative to include the issue with octal numbers in the 0.13 migration guide?
If the old version used by 0.12 can handle the YAMLv2 notation (`0o777`), then mentioning this in the 0.13 migration guide  might be the more sustainable solution.

stefreak:
If the old version can't handle the new notation, then I'm actually not sure what to do 😭 
Maybe the upstream library should support reading both old and new-style octals?

thsig:
> It looks like the update PR https://github.com/garden-io/garden/pull/4630 resolved the issue https://github.com/garden-io/garden/issues/4633

That's right, but I'm not sure what problem that issue was addressing (other than being a general TODO item relating to updating Core's dependencies around the Bonsai push—I'm not aware there was any bug or issue relating to the old version).

> Also, does this mean that we cannot update js-yaml even when there are upstream security fixes or bug fixes?

Looks like it, yeah. If those do come up and there are no patches for the 3.x line, we'll have to fork it.

> Have you considered the alternative to include the issue with octal numbers in the 0.13 migration guide?
> If the old version used by 0.12 can handle the YAMLv2 notation (0o777), then mentioning this in the 0.13 migration guide might be the more sustainable solution.

Yeah, that was my suggestion to a user that reported this a couple of weeks ago, but this user has specifically asked for backwards-compatible parsing of these octal values:

https://github.com/garden-io/garden/issues/4929#issuecomment-1675496791

I think that's a valid concern, at least for the time being. We could always make this a breaking change in the next minor release of 0.13.

How should we proceed here? Should we fork the library now and see if we can gracefully handle both formats, or should we aim to merge this PR?

Either way, since I'm going offline for a while soon, I'll probably need to hand this off to someone else now.

jalper:
> Have you considered the alternative to include the issue with octal numbers in the 0.13 migration guide? If the old version used by 0.12 can handle the YAMLv2 notation (`0o777`), then mentioning this in the 0.13 migration guide might be the more sustainable solution.

The issue with doing this is that we've got multiple customers that are experiencing this issue while trying to migrate to 0.13 who need to be able to gradually migrate all their users, rather than forcing them all to move at once. Backwards compatibility is quite important here.

stefreak:
I would prefer in this partucular case to make 0.12 forward-compatible, which is straight forward to do with the old yaml library because the new syntax (`0o123`) will be parsed as a string with the old library.

This way users can migrate like this:
1. to the latest 0.12 version that supports the new YAML syntax, and the old one at the same time
2. gradually to the latest 0.13 version, person by person (as the config is now compatible with both 0.12 and 0.12)

@mkhq @jalper I know this is not perfect, but this is much better than downgrading the YAML library version and sticking with an old version forever. I think this would be an acceptable solution too.

stefreak:
The solution we went with in the end was switching to a different YAML library that supports document markers for YAML 1.1 compatibility: https://github.com/garden-io/garden/issues/4929#issuecomment-1747066510

PR #4997, opened 2023-08-29
improvement(plugins): debug log manifest dump

eysi09:
Before this we were printing all the manifests we were applying at the verbose level which imo is too low a level and can make the logs very noisy.

Noticed this e.g. when streaming logs via the dashboard.

Fixes #

Assigning @stefreak since you added this. Not sure if it was intentional or not, but don't you agree that this should be debug as opposed to verbose?

### DIFF
--- a/core/src/plugins/kubernetes/kubectl.ts
+++ b/core/src/plugins/kubernetes/kubectl.ts
@@ -111,7 +111,7 @@ export async function apply({
 
   const input = Buffer.from(encodeYamlMulti(manifests))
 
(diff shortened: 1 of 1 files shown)


stefreak (approved):
@eysi09 it was intentional, but I am also ok with the debug log level. Probably better, you're right.

The idea was that users can see what exactly has been applied when kubectl returns errors. I wonder if the manifests are part of the error details? Currently the error.log does not work, which is a bug I'll probably look into next

PR #5001, opened 2023-08-30
fix(framework): several error handling improvements

stefreak:
- Made sure that InternalError is only used for errors that "should not happen" and
  also to wrap unhandled errors that originate from libraries or the NodeJS standard library.

- Stop masking leaf errors (leaf means basically the root cause).
  One of the most occuring errors on Amplitude is "graph", but that's a
  wrapping error. This will give us better insights in what error types
  actually causing command failures.

- Fixed a bug where garden would print endless amounts of yaml with the --output option in case of simple errors. You can reproduce the bug by running `garden deploy -o yaml` in the Garden core repository root directory. The bug has been fixed by filtering error details early to avoid issues later.

- Prevent infinite recursion by making sure not to call `sanitizeValue` from `toSanitizedValue` and `toJSON`

- Removed dead code in the GraphNodeError.

- Consistently handle unexpected errors and format them equally on all
  levels.

- Format InternalErrors as crashes with an explanation
  and a call to action to report the bug on all levels

- Remove the GardenError interface in favor of just a GardenError abstract base class. This makes GardenError recognition less error-prone because there is only one way left of recognizing a GardenError (`instanceof`)

- Treat uncaught errors on the CLI level (cli.ts) as crashes, even if
  they are GardenError instances other than InternalError, because the GardenCli class should not
  throw but return an exit code.

- Make sure that in case of crashes we print all relevant information to the terminal. Yes, that makes the log output less concise, but it increases the probability that users report the errors in a way where we can understand what happened and fix the bug.

- Add a new CRASH issue template, and add a link to reporting crashes on GitHub in the error message.

Crashes that happen in an action handler look like this:
Screenshot 2023-08-30 at 23 43 38

Before it looked like this:
Screenshot 2023-08-31 at 01 04 55

Fixes #


core/src/actions/base.ts
```diff
@@ -731,7 +731,7 @@ export abstract class ResolvedRuntimeAction<
       const buildAction = this.getResolvedDependencies().find((a) => a.kind === "Build" && a.name === buildName)
 
       if (!buildAction) {
-        throw new InternalError({
+        throw new ConfigurationError({
```
stefreak, line 734: TODO: test if my assumption was correct that this isn't a "this should not happen" case.

core/src/cli/cli.ts
```diff
@@ -443,14 +448,19 @@ ${renderCommands(commands)}
...
+      logger = RootLogger.initialize({
+        level: parseLogLevel(logLevelStr),
+        storeEntries: false,
+        displayWriterType: getTerminalWriterType({ silent, output, loggerTypeOpt, commandLoggerType: null }),
+        useEmoji: emoji,
+        showTimestamps,
+        force: this.initLogger,
+      })
+    } catch (error) {
+      return done(1, explainGardenError(error, "Failed to initialize logger"))
```
stefreak, line 462: missing `toGardenError`

core/src/graph/config-graph.ts
```diff
@@ -55,6 +55,7 @@ interface GetActionsParams extends GetActionOpts {
   excludeNames?: string[] // Glob patterns to exclude. An action is returned if its name matches none of these.
 }
 
+// TODO: There are two different GraphError classes
```
stefreak, line 58: TODO: Why do we have two different GraphError classes? I think that's a mistake

core/src/graph/results.ts
```diff
@@ -258,9 +258,23 @@ function filterErrorForExport(error: any) {
     }
   })
 
-  return {
+  // we can't just return a plain object here, because then it wouldn't be
+  // recognizable anymore as a GardenError further down the line and handled as an unexpected crash
+  return new FilteredError({
```
stefreak, line 263: TODO: add a test that makes sure that graph result errors are instanceof GardenBaseError

core/src/plugins/kubernetes/kubernetes.ts
```diff
@@ -77,7 +77,7 @@ export async function configureProvider({
 
   if (config.name !== "local-kubernetes" && !config.deploymentRegistry) {
     throw new ConfigurationError({
-      message: `kubernetes: must specify deploymentRegistry in config`,
+      message: `Must specify deploymentRegistry in provider configuration`,
```
stefreak, line 80: Maybe we roll this one back

core/src/logger/renderers.ts
```diff
@@ -48,7 +48,9 @@ export function renderError(entry: LogEntry): string {
 
   let out = ""
 
-  if (error) {
+  if (!msg && error) {
+    out = explainGardenError(error)
```
stefreak, line 52: missing toGardenError

core/src/analytics/analytics.ts
```diff
@@ -163,7 +163,7 @@ export interface CommandResultEvent extends EventBase {
     name: string
     durationMsec: number
     result: AnalyticsCommandResult
-    errors: string[] // list of GardenBaseError types
+    errors: string[] // list of GardenError types
```
stefreak, line 166: Wait, why is this of type string and the comment says it's a list of GardenError? 🤔

core/src/cli/cli.ts
```diff
@@ -402,9 +402,14 @@ ${renderCommands(commands)}
     processRecord?: GardenProcess
     cwd?: string
   }): Promise<RunOutput> {
+    // for testing
+    if (args.includes("--test-cli-error")) {
+      throw new Error("Test")
```
stefreak, line 407: TODO: implement tests. This seems like a stupid approach to test this, but maybe it's ok?

core/src/exceptions.ts
```diff
@@ -94,124 +85,173 @@ export abstract class GardenBaseError<D extends object = any> extends Error impl
...
+  static wrapError(error: Error | string | any, detail?: unknown, prefix?: string): InternalError {
+    let message: string
+    let stack: string | undefined
+
+    if (error instanceof Error) {
+      message = error.message
+      stack = error.stack
+    } else if (isString(error)) {
+      message = error
+    } else {
```
stefreak, line 200:
```suggestion
    } else if (error) {
```

core/src/logger/logger.ts
```diff
@@ -51,7 +51,8 @@ export function parseLogLevel(level: string): LogLevel {
     lvl = LogLevel[level]
   }
   if (!getNumericLogLevels().includes(lvl)) {
-    throw new InternalError({
+    // This should be validated on a different level
+    throw new ParameterError({
```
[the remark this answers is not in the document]
stefreak: Otherwise `garden --log-level doesnotexist` will be rendered as a crash.

edvald:
This is great stuff. Haven't yet done a full review, but really appreciate this effort @stefreak and I like the overall approach ✨

core/src/exceptions.ts
```diff
@@ -94,124 +85,186 @@ export abstract class GardenBaseError<D extends object = any> extends Error impl
...
+  // If this is an unexpected error, we want to output more details by default and provide some guidance for the user.
+  if (error instanceof InternalError) {
+    let bugReportInformation = formatGardenErrorWithDetail(error)
+
+    if (context) {
+      bugReportInformation = `${stripAnsi(context)}\n${bugReportInformation}`
+    }
+
+    return chalk.red(dedent`
+    ${chalk.bold("Encountered an unexpected Garden error. We are sorry for this. This is likely a bug 🍂")}
```
edvald, line 246:
```suggestion
    ${chalk.bold("Encountered an unexpected Garden error. This is likely a bug 🍂")}
```
I appreciate the tone, but I personally think the apology is maybe unnecessary, plus "we" is sort of undefined for an open-source project :) If we want to include one, we should at least change the phrasing slightly. _Maybe_ "Sorry about that!" but it's also fine imo to just skip.

core/src/exceptions.ts
```diff
@@ -46,17 +36,17 @@ export interface GardenErrorParams<D extends object = any> {
 export type GardenErrorContext = {
   taskType?: string
 }
-
-export abstract class GardenBaseError<D extends object = any> extends Error implements GardenError<D> {
+export abstract class GardenError<D extends object = any | undefined> extends Error {
   abstract type: string
   public override message: string
-  public detail?: D
+  public detail: D
```
stefreak, line 42: This is deceptive: the type is being sanitized in the constructor so the types are only correct if they are base types and non-circular. Maybe we can craft a TypeScript type that allows only base types without circular references? 

On the other hand, the rabbit hole was already quite deep and maybe we can get away with a comment here.

core/src/plugins/kubernetes/container/build/common.ts
```diff
@@ -230,20 +230,25 @@ export async function skopeoBuildStatus({
...
-      const output = res.allLogs || err.message
-
-      throw new RuntimeError({
-        message: `Unable to query registry for image status: ${output}`,
-        detail: {
-          command: skopeoCommand,
-          output,
-        },
-      })
+    if (err instanceof PodRunnerError) {
```
stefreak, line 233: Forgot to throw if it's not a PodRunnerError

.github/ISSUE_TEMPLATE/CRASH.md
```diff
@@ -0,0 +1,41 @@
...
+about: Report crashes
+title: ''
+labels: 'bug,crash'
+assignees: ''
+
+---
+
+## Crash report
+
+<!-- First of all: Thank you for helping us improve Garden! 💚 And sorry. 🍂 -->
```
stefreak, line 12: remove sorry here

core/src/exceptions.ts
stefreak: @TimBeyer and me tried to solve this problem properly, but fell into a deep rabbit hole. Our conclusion: The detail object should be eliminated. But this is out of scope for this PR. Let's merge this for now.

core/src/exceptions.ts
stefreak: OK! O felt it was necessary because the user cannot do anything about these errors and you can feel quite helpless in such a situation. I removed it because it doesn't really help with the helplessness anyway.

TimBeyer (approved):

PR #5023, opened 2023-09-05
test: share less state between semantically independent test contexts

vvagaytsev:
* Introduce helper function to deploy in namespaces
* DRY to avoid repetition and too complex local state
* Less shared global state between different test contexts
* Avoid dependencies between tests and reliance on the execution order
* Ability to run individual tests locally

Improves the quality of the integration tests.

Initially, this was planned to be done as a part of #4846.
I decided to ship it separately to avoid having overly large changes in #4846. That will simplify the review and make the rebasing easier.

### DIFF
--- a/core/test/integ/src/plugins/kubernetes/kubernetes-type/handlers.ts
+++ b/core/test/integ/src/plugins/kubernetes/kubernetes-type/handlers.ts
@@ -53,14 +53,14 @@ describe("kubernetes-type handlers", () => {
   let moduleConfigBackup: ModuleConfig[]
   let nsModuleConfig: ModuleConfig
 
-  const withNamespace = (moduleConfig: ModuleConfig, nsName: string): ModuleConfig => {
+  function withNamespace(moduleConfig: ModuleConfig, nsName: string): ModuleConfig {
     const cloned = cloneDeep(moduleConfig)
     cloned.spec.manifests[0].metadata.name = nsName
     cloned.spec.manifests[0].metadata.labels.name = nsName
     return cloned
   }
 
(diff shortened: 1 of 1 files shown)


core/test/integ/src/plugins/kubernetes/kubernetes-type/handlers.ts
```diff
@@ -53,14 +53,14 @@ describe("kubernetes-type handlers", () => {
...
-  const withNamespace = (moduleConfig: ModuleConfig, nsName: string): ModuleConfig => {
+  function withNamespace(moduleConfig: ModuleConfig, nsName: string): ModuleConfig {
     const cloned = cloneDeep(moduleConfig)
     cloned.spec.manifests[0].metadata.name = nsName
     cloned.spec.manifests[0].metadata.labels.name = nsName
     return cloned
   }
 
-  const findDeployedResources = async (manifests: KubernetesResource<BaseResource>[], logCtx: Log) => {
+  async function findDeployedResources(manifests: KubernetesResource<BaseResource>[], logCtx: Log) {
```
shumailxyz, line 63: just wondering if arrow functions were resulting in some issue here?

shumailxyz (approved):
Looks good, thanks!

core/test/integ/src/plugins/kubernetes/kubernetes-type/handlers.ts
vvagaytsev: No, I converted those to make the code more readable. I'd prefer to keep arrow functions short (like one- or few-liners). Here I just decided to convert all 3 to functions instead of having 1 arrow and 2 functions :)

vvagaytsev:
I'll bring it back to #4846. The merge queue needs some config tuning.

PR #5034, opened 2023-09-06
fix: use shell matching platform when running scripts

twelvemo:
The `runScript()` function that is used to run script steps in workflows as well as the `initScript` was directly shelling out via `/bin/bash -s` which causes the script to be run in WSL on Windows. This can cause confusing behaviour because the commands needed in the `initScript` might not be available in WSL e.g. clis to authenticate to cloud providers.
This PR uses the `exec` wrapper and sets `shell: true` which causes execa to dynamically pick the right shell for the platform.

Fixes #


mkhq (approved):

core/src/commands/workflow.ts
```diff
@@ -376,28 +381,26 @@ export async function runStepScript({ garden, bodyLog, step }: RunStepParams): P
...
     // Unexpected error (failed to execute script, as opposed to script returning an error code)
-    if (!error.exitCode) {
-      throw error
+    if (!(err instanceof ChildProcessError)) {
+      throw err
     }
 
     const scriptError = new WorkflowScriptError({
-      message: `Script exited with code ${error.exitCode}`,
+      message: `Script exited with code ${err.detail.code}`,
```
mkhq, line 391: The `detail` field is currently being removed, but while that is underway I think using it is ok. /cc @stefreak @TimBeyer

mkhq (approved):

Issue #5035, opened 2023-09-06
0.13: [Bug]: update-remote: --parallel flag broken


hnicke:
### Current Behavior
 
After updating from garden v0.13.12 to v0.13.13, we are experiencing the following problem:

```
$ garden remote-update all --parallel
...
Expected concurrency to be an integer from 1 and up or Infinity, got 0 (number)
```

### Expected behavior

The remote-update command should not crash.

### Workaround
Don't use the `--parallel` flag for now.

### Suggested solution(s)

Might be a regression introduced by [switching from Bluebird to native promises](https://github.com/garden-io/garden/pull/4959).

### Your environment

* **OS:** OSX, linux

`garden version`
v0.13.13

PR #5036, opened 2023-09-06
fix: do not error if actionSources are empty

TimBeyer:
When there were no `actionSources` during a remote update, `pMap` was still called but the concurrency would have been set to 0.
That would cause an error and lead to it failing.
Now we only call the map if the `actionSources` are actually there.

fixes https://github.com/garden-io/garden/issues/5035


vvagaytsev (approved):
Thanks for the fast fix! 👍

PR #5039, opened 2023-09-07
refactor: remove error details

stefreak:

Remove the `detail` property and constructor argument from `GardenError`.

We did this because there were several issues with the `detail` property:
- When instantiating a `GardenError`, the `detail` property misled developers in some cases to conclude that the error message has enough information for users: The contents of the `detail` property are not printed to the screen when an error occurs, and sometimes useful information has not been exposed through the message.
- The `detail` property often used to include complex objects, like instances of `Garden` or the `Solver`. The error details sometimes have been exposed to the user through the `-o yaml` and `-o json` output flags, or serialized to JSON either accidentally or on purpose through the WebSocket interface or the Garden Cloud API. This led to issues ranging from bad user experience because of huge amounts of YAML, to OOM crashes when serializing huge amounts of JSON. This in turn led to very hard to maintain work-arounds like `sanitizeValue`. One particular big problem with `sanitizeValue` is that the return type is only known at runtime so we cannot define a correct TypeScript return type.
- The `detail` property is of type `any` by default but there was still a lot of error handling code that relied on particular details to be defined. Due to this issue there were no type errors when changing the details in any way, and that could lead to really hard to understand issues.

We addressed these issues with the following changes:
- Remove the `detail` property and constructor argument from the `GardenError` ABC.
- In cases where error details were needed to be accessed programmatically, we added the exact type definitions needed in a particular subclass of `GardenError`. In the error handling code we used `instanceof` to narrow the type of the error to that subclass, and only then tried accessing the error details.
- In cases where crucial details were missing from the error message itself, we added them to the error message.

Fixes #


core/test/unit/src/commands/delete.ts
```diff
@@ -69,7 +69,7 @@ const moduleConfigs: ModuleConfig[] = [
 
 const missingDeployStatus: ActionStatus = {
   state: "not-ready",
-  detail: { state: "missing", forwardablePorts: [], outputs: {}, detail: {}, mode: "default" },
```
stefreak, line 72: oops, these kinda details we want to keep I guess. But I believe we'll find these through type errors.

vvagaytsev:
There is [one failing test](https://app.circleci.com/jobs/github/garden-io/garden/377481) in the CI.

core/src/plugins/kubernetes/container/ingress.ts
```diff
@@ -260,11 +260,9 @@ async function getCertificateHostnames(api: KubeApi, cert: IngressTlsCertificate
       return getHostnamesFromPem(crtData)
     } catch (error) {
       throw new ConfigurationError({
-        message: `Unable to parse Secret '${cert.secretRef.name}' as a valid TLS certificate`,
-        detail: {
-          ...cert,
-          error,
-        },
+        message: `Unable to parse Secret '${cert.secretRef.name}' as a valid TLS certificate: ${
+          error.message || error
```
vvagaytsev, line 264: Will `error` be rendered as `[object Object]` here?

core/src/plugins/kubernetes/api.ts
```diff
@@ -604,10 +607,7 @@ export class KubeApi {
 
     if (!namespace) {
       throw new KubernetesError({
-        message: `Missing namespace on resource and no namespace specified`,
-        detail: {
-          manifest,
-        },
+        message: `Missing namespace on on ${manifest.kind} resource named ${manifest.metadata.name} and no namespace specified`,
```
vvagaytsev, line 610: Typo: duplicate "on" :)

core/src/util/ext-tools.ts
```diff
@@ -287,9 +283,9 @@ export class PluginTool extends CliWrapper {
         await this.fetch(tmpPath, log)
 
         if (this.buildSpec.extract && !(await pathExists(targetAbsPath))) {
-          throw new ConfigurationError({
-            message: `Archive ${this.buildSpec.url} does not contain a file or directory at ${this.targetSubpath}`,
-            detail: { name: this.name, spec: this.spec },
+          // if this happens, it's a bug!
```
vvagaytsev, line 286: Should we include this line in the error message below?

vvagaytsev: Same question for lines 339 and 368.

vvagaytsev: Ah, no need to do it for `IntrenalError`. It's already implemented in the class.

core/src/plugins/kubernetes/container/ingress.ts
stefreak: No, `Error` implements `toString` which returns the message by default.

core/src/plugins/kubernetes/api.ts
stefreak: good catch!

core/src/util/ext-tools.ts
stefreak: Exactly 👍  It's being done in the `explain` method now. We can also extend the `explain` method of other errors in the future, e.g. ConfigurationError to make sure all syntax errors are being rendered in the same format.

core/src/plugins/kubernetes/run.ts
```diff
@@ -930,11 +989,9 @@ export class PodRunner extends PodRunnerParams {
         if (e instanceof KubernetesError) {
           // if the pod has been deleted during execution we might run into a 404 error.
           // Convert it to Garden NotFoundError and fetch the logs for more details.
-          if (e.statusCode === 404) {
-            throw new NotFoundError({
-              message:
-                "Could not find Pod while waiting for it to complete. The Pod might have been evicted or deleted.",
-              detail: await notFoundErrorDetails(),
+          if (e.responseStatusCode === 404) {
```
vvagaytsev, line 992: The comment looks outdated, we now create an instance of `PodRunnerNotFoundError`.

vvagaytsev (approved):
Huge thanks for doing this PR! 👍  :)

PR #5067, opened 2023-09-12
docs: add a 'Use Cases' page

eysi09:

Adds a "Use Cases" page to the "Basics" section of our docs. 

The end goal is to have a link to a case study and a detailed how to guide for each of the use cases but we just don't have them for all of these at the moment, so instead it's more of a high level overview as things stand. 

The idea was also to split this up into individual pages so that they're visible from the front page, but if it's all the same I'd like to do that in a follow up PR.

We're making quite a few changes to the docs and it'll be easier to split this up and cross link once all those changes have been merged. Otherwise we'll just end up wasting time moving links around as we iterate on this. 

Fixes #

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -8,6 +8,7 @@
 * [Quickstart Guide](./basics/quickstart.md)
 * [Core Concepts](./basics/core-concepts.md)
 * [Garden vs Other Tools](./basics/garden-vs-other-tools.md)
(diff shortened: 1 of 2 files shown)


worldofgeese (approved):
Approving with the aim of reorganizing and editing in "post".

vvagaytsev (approved):

PR #5085, opened 2023-09-13
feat: support for Linux-homebrew installation

stefreak:

Fixes https://github.com/garden-io/homebrew-garden/issues/2

### DIFF
--- a/.github/actions/checksum/action.yml
+++ b/.github/actions/checksum/action.yml
@@ -1,8 +1,8 @@
 name: 'Checksum'
 description: 'Calculate checksum for URL'
 inputs:
-  archive-url:
-    description: 'URL of archive to checksum'
+  url:
+    description: 'URL to checksum'
     required: true
 outputs:
   sha256:
@@ -15,13 +15,13 @@ runs:
       run: |
         set -o pipefail
 
-        if ! checksum=$(curl -sSL --fail "${{ env.ARCHIVE_URL }}" | shasum -a 256 | cut -d ' ' -f 1);
+        if ! checksum=$(curl -sSL --fail "${{ env.URL }}" | shasum -a 256 | cut -d ' ' -f 1);
         then
-          echo "Failed to fetch binary from ${{ env.ARCHIVE_URL }}"
+          echo "Failed to fetch URL ${{ env.URL }}"
           exit 1
         fi
 
         echo "sha256=$checksum" >> "$GITHUB_OUTPUT"
       shell: bash
       env:
-        ARCHIVE_URL: ${{ inputs.archive-url }}
+        URL: ${{ inputs.url }}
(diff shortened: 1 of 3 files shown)


.github/actions/checksum/action.yml
```diff
@@ -0,0 +1,27 @@
+name: 'Checksum'
+description: 'Calculate checksum for URL'
+inputs:
+  archive-url:
```
TimBeyer, line 4: "archive" seems to be the wrong term here since we will hash anything here - even a 404 error response.

TimBeyer: Well, in this case we fail on 404 but you know what I mean :D

stefreak: Because of the `--fail` flag for curl, it will not hash 404 error responses. But you're right, it might respond with something else than an archive. We're only using it for archives though– but I can rename this of course.

TimBeyer: I'd just call it `url`

stefreak: makes sense, done 👍

.github/actions/checksum/action.yml
```diff
@@ -0,0 +1,27 @@
+name: 'Checksum'
+description: 'Calculate checksum for URL'
+inputs:
+  url:
+    description: 'URL of archive to checksum'
```
TimBeyer, line 5: Still mentioning "archive" here

TimBeyer (approved):

support/homebrew-formula.rb.j2
```diff
@@ -6,14 +6,24 @@ class GardenCli < Formula
...
+      url "{{macosArmTarballUrl}}"
+      sha256 "{{macosArmSha256}}"
+    else
+      url "{{macosAmdTarballUrl}}"
+      sha256 "{{macosAmdSha256}}"
+    end
+  end 
 
   def install
     libexec.install "garden", "fsevents.node", "static"
```
stefreak, line 29: The formula doesn't work on linux, I am getting a weird error:

```
steffen@lima-default:~$ brew install garden-cli
Running `brew update --auto-update`...
==> Fetching garden-io/garden/garden-cli
==> Downloading https://download.garden.io/core/0.13.13/garden-0.13.13-linux-amd64.tar.gz
Already downloaded: /home/steffen.linux/.cache/Homebrew/downloads/f852a6a846822bbdd4f0f27cb714d1d92ed7f36400a52f3a54cb61178e746b06--garden-0.13.13-linux-amd64.tar.gz
==> Installing garden-cli from garden-io/garden
Error: The following formula cannot be installed from bottle and must be
built from source.
  garden-cli
Install Clang or run `brew install gcc`.
```

When debugging I noticed that the linux archive does not contain an `fsevents.node` file, but even if I do not copy that file I still get this error. I got stuck there and gave up for now.

PR #5086, opened 2023-09-13
fix: remove rsync dependency from homebrew formula

stefreak:
According to our install guide we only depend on rsync on Windows.

See also https://docs.garden.io/guides/installation

Fixes https://github.com/garden-io/homebrew-garden/issues/8

### DIFF
--- a/scripts/install-osx-dependencies.sh
+++ b/scripts/install-osx-dependencies.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # install/update homebrew dependencies
-BREW_DEPS="jq cmake git kubectl helm rsync icu4c pkg-config dep git-chglog parallel"
+BREW_DEPS="jq cmake git kubectl helm icu4c pkg-config dep git-chglog parallel"
 
 brew update
 brew tap git-chglog/git-chglog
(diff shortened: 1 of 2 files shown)


support/homebrew-formula.rb.j2
```diff
@@ -4,8 +4,6 @@ class GardenCli < Formula
 
   version "{{version}}"
 
-  depends_on "rsync"
```
stefreak, line 7: Note to the reviewers: Updating this jinja2 file will automatically update the homebrew-garden repository upon the next release since https://github.com/garden-io/garden/pull/5062

scripts/install-osx-dependencies.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # install/update homebrew dependencies
-BREW_DEPS="jq cmake git kubectl helm rsync icu4c pkg-config dep git-chglog parallel"
+BREW_DEPS="jq cmake git kubectl helm icu4c pkg-config dep git-chglog parallel"
```
vvagaytsev, line 4: Do we still need `dep`? I got this error when I tried it last time:
```
Error: dep has been disabled because it has an archived upstream repository! (macOS and Linux)
Error: dep not installed
```

That's not a critical one. The installation continues and everything seems to go well.

stefreak:
```suggestion
BREW_DEPS="jq cmake git kubectl helm icu4c pkg-config git-chglog parallel"
```

vvagaytsev (approved):

Issue #4477, opened 2023-05-30
Get rid of .git directory in static directory


Orzelius:
will likely be fixed with https://github.com/garden-io/garden/pull/4047

Issue #4935, opened 2023-08-10
[FEATURE]: Add package for Nix/NixOS


sagikazarmark:
### Background / Motivation

I use Nix as my package manager for macOS and Linux.

### What should the user be able to do?

Install garden using Nix.

### Why do they want to do this? What problem does it solve?

Nix is particularly useful in creating portable development environments. Since garden is often used for development, integrating it into such environments is crucial.

### Suggested Implementation(s)

Create a Nix package.

### How important is this feature for you/your team?

🌵 Not having this feature makes using Garden painful

### Additional information

There is already a closed issue asking for a Nix package: #2785

I decided to open a new issue, because I'd be happy to do the work, but I'd need some assistance because I'm not intimately familiar how Node CLIs are packaged (particularly garden seems to require some additional files for it to run).

What I need to be able to create a Nix package:

- a build command that creates a script/artifact I can run
- list of additional files required for said artifact

I tried following the contributing guide, but it's not obvious what the output of `yarn build` is.

Any assistance would be greatly appreciated.

Thanks!

sagikazarmark:
Here is an initial version using the binary distribution for now: https://github.com/sagikazarmark/nix-garden

stefreak:
@sagikazarmark Cool, thank you for working on this!

In our contributing guide we have a [section about how to build the release binaries and docker containers]().

In NixOS you use glibc by default, right? 

Basically we need the built pkg vercel nodejs binary, and a "static" directory (that needs to be initialized as a git repository unfortunately right now).

See also the debian Dockerfile: https://github.com/garden-io/garden/blob/main/support/buster.Dockerfile#L47C2-L47C2
The build context root for that Dockerfile dist/linux-amd64 (This directory exists after running yarn dist)

Happy to help if you run into problems.

stefreak:
Unfortunately we have no official ARM binaries yet, which should be straightforward to change but we are tracking a separate issue for that: #1547

sagikazarmark:
@stefreak thanks for getting back to me.

As far as I understand the value behind building the vercel pkg binaries is to distribute garden as a single executable. I'm not sure the Nix package needs that as Nix is perfectly capable of installing Node as a dependency as well.

The repo I linked above already uses the binary distribution which is a good start IMO, but I'd like to play a bit with skipping the final vercel build and just use ... the artifact that comes before that. A script file? Not sure yet.

[...]

Issue #5030, opened 2023-09-05
garden self-update painful on windows


kroher:
### Current Behavior

When I run garden self-update in windows(usually running with a version), the next time i run garden it complains/errors 

`Path C:\Users\usrdir\.garden\bin\static is not in a git repository root. Garden must be run from within a git repo. Please run `git init` if you're starting a new project and repository, or move the project to an existing repository, and try again.`

after i go into the directory and run git init garden works fine.

doesnt do this on macOs

### Expected behavior

i shouldnt have to manually run git init in the user/.garden directory at all to start garden

### Reproducible example

clean install of garden via installer on windows
garden dev and things stand up fine on vote
garden self-update 0.13.12
garden dev
error message

### Your environment

windows 11
gke

0.12.63, 0.12.64, 0.13.12, 0.13.13

vvagaytsev:
Thanks for reposting this @kroher!
This might have the same root cause as #2614.

We have a draft PR #4047 to solve the issue.

Issue #2614, opened 2021-10-07
Running garden self-update on Windows 10 leaves garden non-functional


ITHedgeHog:
### Current Behavior

If I issue the command garden self-update in an administrative or user session of PowerShell, I end up with Garden non-functional until I reinstall via the PowerShell script
### Expected behavior

I expect to be able to issue a command such as garden deploy without failure

### Reproducible example
```
PS C:\git\uk.energy> garden self-update
Update Garden

Running in namespace ...

Checking for latest version...
Installation directory: C:\Users\Dan Taylor\.garden\bin
Current Garden version: 0.12.26
Latest release version: 0.12.27

Downloading version 0.12.27 from
https://github.com/garden-io/garden/releases/download/0.12.27/garden-0.12.27-windows-amd64.zip...
Backing up prior installation to C:\Users\Dan Taylor\.garden\bin\.backup\0.12.26...
Extracting to installation directory C:\Users\Dan Taylor\.garden\bin...

Done!
PS C:\git\uk.energy> garden version
0.12.27
PS C:\git\uk.energy> garden --env=systest plugins kubernetes cluster-init
Plugins


Running in namespace default

Initializing/updating cluster-wide services for systest environment

ℹ providers                 → Getting status...

Failed resolving provider kubernetes. Here is the output:
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Path C:\Users\Dan Taylor\.garden\bin\static\kubernetes\system is not in a git repository root. Garden must be run from within a git repo. Please run `git init` if you're starting a new project and repository, or move the project to an existing repository, and try again.
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Failed resolving one or more providers:
- kubernetes

See .garden/error.log for detailed error message
```
[error.log](https://github.com/garden-io/garden/files/7301471/error.log)

### Your environment

* **OS:** Windows 10 Professional
* **How I'm running Kubernetes:** AKS

`garden version`
```
PS C:\git\uk.energy> garden version
0.12.27
```

ITHedgeHog:
Just adding, same behaviour on Windows 11.

mcanaleta:
[...]

PR #5087, opened 2023-09-13
feat: NPM registry package for `garden`

worldofgeese:

This adds an NPM registry package to be published by our GitHub Actions workflow for `garden`.

Unblocks #4935.

There's a chicken and egg problem here you may run into when testing the GitHub Actions publish workflow, which is that the top-level `yarn && yarn build` expects these packages to exist but only `core` is published. NPM can also derive these packages locally so it may be a non-issue but it's something to watch out for.

I have tested a version of this (and you can too) by publishing the CLI package in my own NPM registry namespace. Run `npx @worldofgeese/cli`. It has been my daily driver for a week now. The difference between this and the `@garden-io/cli` is the [`package.json`](https://www.npmjs.com/package/@worldofgeese/cli?activeTab=code) file. You'll notice in my test package I've stripped out a lot of `@garden-io/cli`'s dependencies but in this PR I've modified `@garden-io/cli` directly to offer a more elegant solution. We need to test that these modifications work as well as they did in my own test package.


cli/bin/garden
```diff
@@ -1,5 +1,24 @@
 #!/usr/bin/env node
+const path = require('path');
 
+process.env.NODE_OPTIONS = "max-old-space-size=4096 max-semi-space-size=64";
```
stefreak, line 4: The NODE_OPTIONS environment variable has no effect if you only set it after the nodejs process booted up.
We need to pass the options earlier:
```suggestion
#!/usr/bin/env node --max-old-space-size=4096 --max-semi-space-size=64
const path = require('path');

```

cli/bin/garden
```diff
@@ -1,5 +1,24 @@
...
+process.env.STATIC_DIR = path.join(__dirname, '..', 'static');
+
+const fs = require('fs');
+const exec = require('child_process').exec;
+
+const gitDir = path.join(process.env.STATIC_DIR, '.git');
+
+if (!fs.existsSync(gitDir)) {
+  exec(`git init ${process.env.STATIC_DIR}`, (error, stdout, stderr) => {
```
stefreak, line 12: I think this is a little early in the code to call git init; Also if we generally call git init on the static directory if it isn't a git repository yet (until we fixed #4477) we also fix a windows issue (#5030 and #2614)

.github/workflows/publish-release.yml
```diff
@@ -33,13 +34,13 @@ jobs:
...
-          npm version "$CURRENT_CORE_VERSION-edge.$SHORT_SHA"
-          npm publish --tag edge
+          lerna version "$CURRENT_CORE_VERSION-edge.$SHORT_SHA" --no-git-tag-version --no-push --yes
+          lerna publish -from-package --dist-tag edge --yes
       - name: Publish stable version if normal release
         if: github.event.release.prerelease != true
         working-directory: ./core
         run: |
-          npm publish
+          lerna publish from-package --yes
```
TimBeyer, line 43: Could we add some documentation how exactly the process works that increments the package version to the correct version number?

worldofgeese: @TimBeyer sure! In a code comment?

lerna.json
```diff
@@ -0,0 +1,16 @@
+{
+    "$schema": "node_modules/lerna/schemas/lerna-schema.json",
+    "version": "0.13.13",
```
TimBeyer, line 3: Will this version get updated correctly with every release publishing step?

worldofgeese:
In the build logs, we're getting build failures due to `patch-package` acting on  a missing `pkg` in `node_modules`. This wasn't an issue in my test package because I created a slim version of the CLI that didn't include `pkg`. I don't know how to resolve this. I've tried to remove `nohoist` but that didn't solve it.

Any suggestions @TimBeyer or @stefreak?

.github/workflows/publish-release.yml
TimBeyer: I think you could leave a few short comments in here and then more detailed in https://github.com/garden-io/garden/blob/main/RELEASE_PROCESS.md

lerna.json
worldofgeese: Yes! `lerna` can automatically bump versions according to one of a [number of strategies](https://lerna.js.org/docs/features/version-and-publish#versioning-strategies).

TimBeyer: What I mean is, does this number here get updated and committed back or is this just a placeholder of sorts?

TimBeyer:
> In the build logs, we're getting build failures due to `patch-package` acting on a missing `pkg` in `node_modules`. This wasn't an issue in my test package because I created a slim version of the CLI that didn't include `pkg`. I don't know how to resolve this. I've tried to remove `nohoist` but that didn't solve it.
> 
> ```
> error /home/circleci/project/node_modules/@garden-io/cli: Command failed.
> Exit code: 1
> Command: patch-package
> Arguments: 
> Directory: /home/circleci/project/node_modules/@garden-io/cli
> Output:
> patch-package 6.5.1
> Applying patches...
> Error: Patch file found for package pkg which is not present at node_modules/pkg
> ---
> patch-package finished with 1 error(s).
> info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
> 
> Exited with code exit status 1
> ```
> 
> Any suggestions @TimBeyer or @stefreak?

I'm wondering if we need to restructure things a tiny bit.
Currently `cli` both does the bundling and docs generation, and is the actual CLI code that runs.
Probably we should move the release parts out into a `private` package and only call that in CI so that we don't really need to install a big tool like `pkg` just to run garden via `npm` / `npx`.

stefreak:
A big problem with this approach is that dependencies will not be pinned (`npm install @garden/cli` will ignore the `yarn.lock` file we have in place).

If we'd be using NPM, the solution would be to switch from using `package-lock.json` to using the `npm-shrinkwrap.json`.

But [Yarn does not support `npm-shrinkwrap.json`](https://classic.yarnpkg.com/lang/en/docs/migrating-from-npm/) – and we can't switch to NPM, because [the PKG-built binary does not work anymore correctly with the `node_modules` installed by NPM](https://github.com/garden-io/garden/pull/3919).

Using both `yarn.lock` and `npm-shrinkwrap.json` we'd need to solve the problem of keeping both files in sync, or otherwise our unit, integration and e2e tests might not be covering exactly the dependencies that end up in Garden that ships through NPM / NPX.

cli/bin/garden
worldofgeese: @stefreak, where would you move this call?

worldofgeese:
Hey @stefreak and @TimBeyer thank you both for your feedback and great work in making this PR possible! I will continue working on this. I just synced with Tim and collected the steps still left before this is merge-ready.  Can you both look this over and let me know if there's anything missing? Thank you! 

- [ ] make a new `builder` package folder
- [ ] add a `prepublishOnly` life cycle script to shrinkwrap
-`[ ] waiting on Vova's PR to land that removes `static`
- [ ] remove `static` dir `git init` step
- [ ] test how `lerna` bumps package versions on my own branch: does it commit back?

vvagaytsev:
Closing this as not planned.

PR #5092, opened 2023-09-15
refactor: enable "useUnknownInCatchVariables" in tsconfig

### LINKED ISSUE #5097: Crash: toKubernetesError encountered an unknown error unexpectedly during Kub
## Crash report
`garden deploy` crashed for `deploy.api` and `deploy.version-check`

### Error message

```
✖ deploy.api           → Failed resolving status for Deploy type=kubernetes name=api (from module api) (took 17.69 sec). This is what happened:

──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

You can help by reporting this on GitHub: https://github.com/garden-io/garden/issues/new?labels=bug,crash&template=CRASH.md&title=Crash%3A%20toKubernetesError%20encountered%20an%20unknown%20error%20unexpectedly%20during%20Kub...

Failed resolving status for Deploy type=kubernetes name=api (from module api) (took 17.69 sec). This is what happened:
RequestError: Error: connect ETIMEDOUT 18.194.93.210:443
    at new RequestError (/snapshot/project/tmp/pkg/cli/node_modules/request-promise-core/lib/errors.js:14:15)
    at plumbing.callback (/snapshot/project/tmp/pkg/cli/node_modules/request-promise-core/lib/plumbing.js:87:29)
    at Request.RP$callback [as _callback] (/snapshot/project/tmp/pkg/cli/node_modules/request-promise-core/lib/plumbing.js:46:31)
    at self.callback (/snapshot/project/tmp/pkg/cli/node_modules/request/request.js:185:22)
    at Request.emit (node:events:537:28)
    at Request.emit (node:domain:482:12)
    at Request.onRequestError (/snapshot/project/tmp/pkg/cli/node_modules/request/request.js:877:8)
    at /snapshot/project/tmp/pkg/cli/node_modules/@opentelemetry/context-async-hooks/build/src/AbstractAsyncHooksContextManager.js:50:55
    at AsyncLocalStorage.run (node:async_hooks:330:14)
    at AsyncLocalStorageContextManager.with (/snapshot/project/tmp/pkg/cli/node_modules/@opentelemetry/context-async-hooks/build/src/AsyncLocalStorageContextManager.js:33:40)
    at ClientRequest.contextWrapper (/snapshot/project/tmp/pkg/cli/node_modules/@opentelemetry/context-async-hooks/build/src/AbstractAsyncHooksContextManager.js:50:32)
    at ClientRequest.emit (node:events:537:28)
    at ClientRequest.emit (node:domain:482:12)
    at TLSSocket.socketErrorListener (node:_http_client:465:9)
    at TLSSocket.emit (node:events:537:28)
    at TLSSocket.emit (node:domain:482:12)
    at emitErrorNT (node:internal/streams/destroy:151:8)
    at emitErrorCloseNT (node:internal/streams/destroy:116:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

toKubernetesError encountered an unknown error unexpectedly during Kubernetes API: /api/v1/namespaces/ge-ulianasirotina/serviceaccounts/api-serviceaccount: toKubernetesError encountered an unknown error unexpectedly during Kubernetes API: /api/v1/namespaces/ge-ulianasirotina/serviceaccounts/api-serviceaccount: Error: connect ETIMEDOUT 18.194.93.210:443

```

### What did you do?

```
garden delete environment
garden deploy
```

### Your environment

* **OS:** macOS 13.5.2 
* **How I'm running Kubernetes:** 🤷‍♀️
* **Garden version:** garden version: 0.13.13-edge-bonsai-1f585c672

### Frequency

Keeps happening 

### Workaround

```
garden deploy --force
```

stefreak:
This forces us to narrow types in the error handling code.

Fixes #5097

### DIFF
--- a/core/src/commands/cloud/secrets/secrets-create.ts
+++ b/core/src/commands/cloud/secrets/secrets-create.ts
@@ -6,7 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import { CommandError, ConfigurationError, CloudApiError } from "../../../exceptions"
+import { CommandError, ConfigurationError, CloudApiError, GardenError } from "../../../exceptions"
 import { CreateSecretResponse } from "@garden-io/platform-api-types"
 import { readFile } from "fs-extra"
 
@@ -93,7 +93,7 @@ export class SecretsCreateCommand extends Command<Args, Opts> {
         secrets = dotenv.parse(await readFile(fromFile))
       } catch (err) {
         throw new CommandError({
-          message: `Unable to read secrets from file at path ${fromFile}: ${err.message}`,
+          message: `Unable to read secrets from file at path ${fromFile}: ${err}`,
         })
       }
     } else if (args.secrets) {
@@ -104,7 +104,7 @@ export class SecretsCreateCommand extends Command<Args, Opts> {
           return acc
         } catch (err) {
           throw new CommandError({
-            message: `Unable to read secret from argument ${keyValPair}: ${err.message}`,
+            message: `Unable to read secret from argument ${keyValPair}: ${err}`,
           })
         }
       }, {})
@@ -174,9 +174,12 @@ export class SecretsCreateCommand extends Command<Args, Opts> {
         const res = await api.post<CreateSecretResponse>(`/secrets`, { body })
         results.push(makeSecretFromResponse(res.data))
       } catch (err) {
+        if (!(err instanceof GardenError)) {
+          throw err
+        }
         errors.push({
           identifier: name,
-          message: err?.response?.body?.message || err.messsage,
+          message: err.message,
         })
       }
     }
--- a/core/src/exceptions.ts
+++ b/core/src/exceptions.ts
@@ -137,6 +137,8 @@ export abstract class GardenError extends Error {
   toJSON() {
     return {
       type: this.type,
+      taskType: this.taskType,
+      code: this.code,
       message: this.message,
       stack: this.stack,
       wrappedErrors: this.wrappedErrors,
@@ -339,8 +341,13 @@ export class InternalError extends GardenError {
   static wrapError(error: Error | string | any, prefix?: string): InternalError {
     let message: string | undefined
     let stack: string | undefined
+    let code: NodeJSErrnoErrorCodes | undefined
 
-    if (error instanceof Error) {
+    if (isErrnoException(error)) {
+      message = error.message
+      stack = error.stack
+      code = error.code
+    } else if (error instanceof Error) {
       message = error.message
       stack = error.stack
     } else if (isString(error)) {
@@ -352,7 +359,7 @@ export class InternalError extends GardenError {
 
     message = message ? stripAnsi(message) : ""
 
-    return new InternalError({ message: prefix ? `${stripAnsi(prefix)}: ${message}` : message, stack })
+    return new InternalError({ message: prefix ? `${stripAnsi(prefix)}: ${message}` : message, stack, code })
   }
 
   override explain(context?: string): string {
--- a/core/src/exceptions.ts
+++ b/core/src/exceptions.ts
@@ -14,6 +14,17 @@ import stripAnsi from "strip-ansi"
 import { Cycle } from "./graph/common"
 import indentString from "indent-string"
 
+/**
+ * NodeJS native errors with a code property.
+ */
+export type OSError = {
+  code: string
+} & Error
+
+export function isOSError(err: any): err is OSError {
+  return typeof err?.code === "string"
+}
+
 export type StackTraceMetadata = {
   functionName: string
   relativeFileName?: string
--- a/core/src/exceptions.ts
+++ b/core/src/exceptions.ts
@@ -17,7 +17,7 @@ import { constants } from "os"
 
 // See https://nodejs.org/api/os.html#error-constants
 type NodeJSErrnoErrors = typeof constants.errno
-type NodeJSErrnoErrorCodes = keyof NodeJSErrnoErrors
+export type NodeJSErrnoErrorCodes = keyof NodeJSErrnoErrors
 
 const errnoErrorCodeSet = new Set(Object.keys(constants.errno))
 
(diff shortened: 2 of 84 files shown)


core/src/exceptions.ts
```diff
@@ -14,6 +14,17 @@ import stripAnsi from "strip-ansi"
...
 
+/**
+ * NodeJS native errors with a code property.
+ */
+export type OSError = {
+  code: string
+} & Error
+
+export function isOSError(err: any): err is OSError {
+  return typeof err?.code === "string"
```
TimBeyer, line 25: I wonder if we need to be more explicit here.
We could grab the list from https://nodejs.org/api/errors.html#common-system-errors and check against that explicitly.

stefreak: I like how the OpenTelemetry package does this: 
```
interface ExceptionWithCode {
    code: string | number;
    name?: string;
    message?: string;
    stack?: string;
}
interface ExceptionWithMessage {
    code?: string | number;
    message: string;
    name?: string;
    stack?: string;
}
interface ExceptionWithName {
    code?: string | number;
    message?: string;
    name: string;
    stack?: string;
}
/**
 * Defines Exception.
 *
 * string or an object with one of (message or name or code) and optional stack
 */
export declare type Exception = ExceptionWithCode | ExceptionWithMessage | ExceptionWithName | string;
```

WDYT?

core/src/commands/cloud/secrets/secrets-create.ts
```diff
@@ -174,9 +174,12 @@ export class SecretsCreateCommand extends Command<Args, Opts> {
         const res = await api.post<CreateSecretResponse>(`/secrets`, { body })
         results.push(makeSecretFromResponse(res.data))
       } catch (err) {
+        if (!(err instanceof GardenError)) {
+          throw err
+        }
         errors.push({
           identifier: name,
-          message: err?.response?.body?.message || err.messsage,
+          message: err.message,
```
TimBeyer, line 182: `messsage` 😄

core/src/exceptions.ts
TimBeyer: You're calling this OSError though which implies a certain kind of error with specific codes. Anyone can make any error have a code property so for this we should really pin it down to the correct error codes so we don't accidentally make wrong assumptions.

stefreak: I'm not sure if that list will be maintainable, because it's platform-specific and probably also depends on the NodeJS version. If the NodeJS standard library has such a type I'm all for it, but maintaining it ourselves not so much

TimBeyer: Then we should rename this since it's not accurate. But I think using that list and maintaining it manually also wouldn't be the worst idea. I haven't seen those get updated very frequently if at all.

core/src/logger/logger.ts
```diff
@@ -363,6 +363,9 @@ export class RootLogger extends LoggerBase {
       try {
         config.level = parseLogLevel(gardenEnv.GARDEN_LOG_LEVEL)
       } catch (err) {
+        if (!(err instanceof GardenError)) {
```
TimBeyer, line 366: `parseLogLevel` throws a `ParameterError` so maybe it makes sense to be explicit here.
I'm also wondering if we need the duplication of messages since the `ParameterError` already logs 
```
`Unexpected log level, expected one of ${getLogLevelChoices().join(", ")}, got ${level}`
```

but I guess it's not too bad.

core/src/plugins/kubernetes/helm/logs.ts
```diff
@@ -42,7 +42,7 @@ export const getHelmDeployLogs: DeployActionHandler<"getLogs", HelmDeployAction>
         break
       } catch (err) {
         log.debug(`Failed getting deployed resources. Retrying...`)
-        log.silly(err.message)
+        log.silly(`Failed getting deployed resources: ${err}. Retrying...`)
```
TimBeyer, line 45: I'm not sure if this is better than before since now silly will log twice which might lead one to think it retried once already. Maybe we should just log `err` directly?

core/src/plugins/kubernetes/run.ts
```diff
@@ -664,6 +664,9 @@ async function runWithArtifacts({
       try {
         await copy(tmpDir.path, artifactsPath, { filter: (f) => !f.endsWith(".garden-placeholder") })
       } catch (err) {
+        if (!(err instanceof Error)) {
+          throw err
```
TimBeyer, line 668: I wonder if it makes sense for those sort of cases to throw a `NonErrorThrownError` or something like that

core/src/plugins/kubernetes/run.ts
```diff
@@ -1238,9 +1241,7 @@ export class PodRunner extends PodRunnerParams {
       const containerLogs = allLogs.find((l) => l.containerName === mainContainerName)?.log?.trim()
       return containerLogs || ""
     } catch (err) {
-      return `[Could not retrieve logs for container '${mainContainerName}': ${
-        err.message || "unknown error occurred"
-      }]`
+      return `[Could not retrieve logs for container '${mainContainerName}': ${err}]`
```
TimBeyer, line 1244: I wonder if that `"unknown error occurred"` comes from some actual occurrence of that or is just a precaution. If something throws an error without a message, does it make sense to surface that somehow differently?

stefreak: is it even possible to throw undefined?

core/src/plugins/kubernetes/run.ts
stefreak: In those cases we transform these into `InternalError` already later down the line; Here I just want to make the `message` property accessible.

TimBeyer (dismissed):

TimBeyer (dismissed):

TimBeyer (approved):

PR #5095, opened 2023-09-15
chore: cleanup and code inspection-based fixes

vvagaytsev:

Fixes #


stefreak (approved):
Awesome! This looks good to me. There is a failing test though (I'm not sure if it's a flake).

@vvagaytsev How did you find the typos and the unused code? If you used a tool for that, we should add it to the CI / our linter configuration as well.

core/src/commands/get/get-eysi.ts
```diff
@@ -21,8 +21,6 @@ export class GetEysiCommand extends Command {
     Just try it.
   `
 
-  loggerType: "basic"
```
TimBeyer, line 24: I believe these are actually bugs where we needed to use `=` but accidentally used `:`

vvagaytsev:
@stefreak I used the code inspection tool in WebStorm. There should be some tools that we can embed into the CI pipeline.

core/src/commands/get/get-eysi.ts
vvagaytsev: Yes, looks like `=` should be used. But this field is no longer used in the code. So I removed it :)

stefreak:
@vvagaytsev  I think for CI it would be paid: https://www.jetbrains.com/qodana/buy/?billing=yearly

PR #5099, opened 2023-09-18
chore: add debian family support images

twelvemo:
Adds debian family garden support images with the same toolset as the alpine based images. Creates images for debian buster, bullseye and bookworm.

Fixes #


support/debian.Dockerfile
```diff
@@ -0,0 +1,188 @@
...
+WORKDIR /project
+
+#
+# garden-aws-base
+#
+FROM garden-base-root as garden-aws-base
+ENV AWSCLI_VERSION=2.13.15
+ENV AWSCLI_SHA256="ac63e8f42c7f8775edccdc004921420159420de9185cf011952dba8fda5895ff"
+
+RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-${AWSCLI_VERSION}.zip" -o "awscliv2.zip"
```
stefreak, line 80: We need to test the sha hash after curl.

support/debian.Dockerfile
```diff
@@ -0,0 +1,188 @@
...
+RUN unzip awscliv2.zip
+RUN ./aws/install
+
+#
+# garden-gcloud-base
+#
+FROM garden-base as garden-gcloud-base
+ENV GCLOUD_VERSION=444.0.0
+
+RUN curl -O https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-cli-${GCLOUD_VERSION}-linux-x86_64.tar.gz
```
stefreak, line 90: We need to test the sha hash after curl.

support/debian.Dockerfile
```diff
@@ -0,0 +1,188 @@
...
+
+#
+# garden-aws-gloud
+#
+FROM garden-base as garden-aws-gcloud
+
+# Copy aws cli
+COPY --chown=$USER:root --from=garden-aws-base /usr/local/aws-cli /usr/local/aws-cli
+RUN mkdir -p ${HOME}/bin
+RUN ln -s /usr/local/aws-cli/v2/current/bin/aws ${HOME}/bin/aws
```
stefreak, line 156: Why are we linking, instead of copying the `/usr/local/aws-cli/v2/current` directory?

support/docker-bake.hcl
```diff
@@ -18,6 +18,14 @@ variable "PRERELEASE" {
   default = ""
 }
 
+variable "ALL_SUPPORTED_DISTROS" {
+  default = ["alpine", "buster", "bullseye", "bookworm"]
```
stefreak, line 22: Would this also work, so we don't repeat ourselves?
```suggestion
  default = concat(["alpine"], DEBIAN_DISTROS)
```

support/docker-bake.hcl
```diff
@@ -102,100 +164,139 @@ group "alpine" {
...
-##
-
 target "alpine-base" {
   dockerfile = "../../support/alpine.Dockerfile"
-  target     = "garden-alpine-base"
+  target     = "garden-base"
   platforms  = ["linux/amd64"]
   context    = "dist/alpine-amd64"
-  tags       = repository("gardendev/garden", withLatest(tags("alpine")))
+  tags       = repository("gardendev/garden", tags("alpine"))
```
stefreak, line 190: why did we remove withLatest here? I think we want to continue using the `latest` tag with the alpine flavor

stefreak:
```suggestion
  tags       = repository("gardendev/garden", withLatest(true, tags("alpine")))
```

stefreak (changes requested):
This looks good already!

There is a lot more complexity with handling multiple versions of Debian. Are we sure we want to support more than the LTS versions of debian and ubuntu or something like that? If yes that's ok for me.

What's still missing overall is:
- Update the docker-bake-test.sh script to reflect and test the changes
- Update the docs at https://docs.garden.io/reference/dockerhub-containers to reflect the new distros we support. Maybe it should be a separate "distros" section

support/debian.Dockerfile
twelvemo: aws cli comes with a couple of symlinks when installed and some of them are inside of the current directory. Changed it in a way, so we don't need the symlinks.

support/docker-bake.hcl
twelvemo: As discussed only using bookworm, so this won't be necessary.

support/docker-bake.hcl
twelvemo: Good catch!

twelvemo:
As for docs: We never made it clear in the docs that only a subset of the images are available for buster, so i think it actually makes more sense now since all variants have a bookworm version. However i am wondering if it could be potentially disrupting if we remove the buster version of the garden base images with this PR since some folks might rely on these images in their CIs. It would probably be better to announce that we are deprecating the buster base images in favor of the bookworm images for all variants but keep building them for a bit. What do you think @stefreak ?

twelvemo:
The buildx version that ships with circlecis most up-to-date remote-docker executor does not support the matrix notation:
```
support/docker-bake.hcl:138
--------------------
 136 |     target "aws" {
 137 |       name = "${distro}-aws"
 138 | >>>   matrix = {
 139 |         distro = ["bookworm", "alpine"]
 140 |       }
--------------------
ERROR: support/docker-bake.hcl:138,3-9: Unsupported argument; An argument named "matrix" is not expected here., and 75 other diagnostic(s)
```

stefreak:
> As for docs: We never made it clear in the docs that only a subset of the images are available for buster, so i think it actually makes more sense now since all variants have a bookworm version. However i am wondering if it could be potentially disrupting if we remove the buster version of the garden base images with this PR since some folks might rely on these images in their CIs. It would probably be better to announce that we are deprecating the buster base images in favor of the bookworm images for all variants but keep building them for a bit. What do you think @stefreak ?

As buster is LTS I think it would make sense to continue supporting it (until it's EOL). I think it is ok to drop support for certain distros, because the old images will continue to work, so it's not a breaking change per se. The lifetime of distros can differ from the lifetime of Garden major releases. Announcing deprecations doesn't hurt though.

stefreak:
> The buildx version that ships with circlecis most up-to-date remote-docker executor does not support the matrix notation:
> 
> ```
> support/docker-bake.hcl:138
> --------------------
>  136 |     target "aws" {
>  137 |       name = "${distro}-aws"
>  138 | >>>   matrix = {
>  139 |         distro = ["bookworm", "alpine"]
>  140 |       }
> --------------------
> ERROR: support/docker-bake.hcl:138,3-9: Unsupported argument; An argument named "matrix" is not expected here., and 75 other diagnostic(s)
> ```

Is it maybe possible to add the latest version to our [circleci-runner](https://github.com/garden-io/garden/blob/main/images/circleci-runner/Dockerfile) image? Or are we not using that for some reason?

stefreak (changes requested):
Looks good to me, except for some minor comments. We can weed out the remaining blockers together next week, if you want!

support/debian.Dockerfile
```diff
@@ -0,0 +1,181 @@
...
+COPY --chown=$USER:root --from=garden-azure-base /usr/bin/az /usr/bin/az
+COPY --chown=$USER:root --from=garden-azure-base /opt/az /opt/az
+COPY --chown=$USER:root --from=garden-azure-base /usr/local/bin/kubectl /usr/local/bin/kubectl
+COPY --chown=$USER:root --from=garden-azure-base /usr/local/bin/kubelogin /usr/local/bin/kubelogin
+
+#
+# garden-aws
+#
+FROM garden-base as garden-aws
+ENV AWSCLI_VERSION=2.13.15
```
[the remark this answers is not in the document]
twelvemo: The bin dir in aws-cli is in a dir with the current version. Current however is just a symlink to that versioned dir and points to the binaries at install time. We could also solve this with setting a couple of symlinks. 
```
ls -la /root/aws-cli/v2/current 
lrwxrwxrwx 1 root root 29 Sep 20 11:00 /root/aws-cli/v2/current -> /usr/local/aws-cli/v2/2.13.15
```

worldofgeese (approved):

stefreak (approved):
Nice work!

PR #5100, opened 2023-09-18
chore: rename Basics/basics to Overview/overview

worldofgeese:


eysi09:
Note that the `docs/README` file is auto generated from the directory structure. 

You can't just rename its contents to `Overview`. You rather have to re-name the `basics` directory with `git mv docs/basis docs/overview` and then run `yarn run genereate-docs`.

worldofgeese:
> Note that the `docs/README` file is auto generated from the directory structure.
> 
> You can't just rename its contents to `Overview`. You rather have to re-name the `basics` directory with `git mv docs/basis docs/overview` and then run `yarn run genereate-docs`.

Didn't know this but it looks like it did it anyway?

eysi09:
> My tests are failing so I do need the generate-docs step looks like

Yeah you definitely need to do that. I'm not sure whether you need to rename the actual dir or not, or whether it's enough to just update the `title` field on the frontmatter.

eysi09:
That being said, it wouldn't hurt to just rename the dir as well so it's consistent with the other and easy to find in the code.

worldofgeese:
@eysi09 what am I doing wrong here that `generate-docs` continues to want to rename a few links back to `basics`?

image
image
image
image

eysi09:
Ah right, these lines are actually from the source code proper. 

For example here: https://github.com/garden-io/garden/blob/main/core/src/plugins/kubernetes/ephemeral/ephemeral.ts#L32

You need to update that line of code and change it from `basic` to `overview`. 

When Garden generates the reference docs it uses these schemas which are defined "in-code".

worldofgeese:
> The other two instances are here:
> 
> https://github.com/garden-io/garden/blob/main/core/src/plugins/kubernetes/kubernetes.ts#L152
> 
> ...and here:
> 
> https://github.com/garden-io/garden/blob/main/core/src/plugins/kubernetes/local/local.ts#L23

All done and ready for review. Thanks! 🙏🏼

.gitbook.yaml
```diff
@@ -7,8 +7,8 @@ structure:
 redirects:
   installation: ./getting-started/1-installation.md
   quick-start: ./getting-started/0-introduction.md
-  basics/installation: ./getting-started/1-installation.md
```
eysi09, line 10: This should change. 

This describes redirects so if we e.g. have a link to `https://docs.garden.io/basics/installation` in other content (e.g. website, blog) it'll redirect to `https://docs.garden.io/getting-started/1-installation`. 

But we don't need to change that in this case.

.gitbook.yaml
```diff
@@ -41,10 +41,10 @@ redirects:
   advanced/pulumi: ./pulumi-plugin/about.md
   advanced/rbac-config: ./k8s-plugins/advanced/rbac-config.md
   getting-started: ./tutorials/your-first-project.md
-  getting-started/0-introduction: ./basics/quickstart
+  getting-started/0-introduction: ./overview/quickstart
   getting-started/1-installation: ./guides/installation.md
   getting-started/2-connect-to-a-cluster: ./tutorials/your-first-project/2-connect-to-a-cluster.md
-  basics/5-min-quickstart: ./basics/quickstart.md
```
eysi09, line 47: Here we shouldn't change the left hand side. Presumably this is for a link in some past content. 

The right hand side should point the new `getting-started` section. 

So it should be: `  basics/5-min-quickstart: ./getting-started/quickstart.md`

README.md
```diff
@@ -6,7 +6,7 @@ _If you love Garden, please ★ star this repository to show your support :green
   <img src="https://github.com/garden-io/garden/assets/59834693/f62a04cb-44bc-4dd4-8426-398b6cd846fd" align="center">
 </p>
 <div align="center">
-  <a href="https://docs.garden.io/basics/5-min-quickstart/?utm_source=github">Quickstart</a>
+  <a href="https://docs.garden.io/overview/5-min-quickstart/?utm_source=github">Quickstart</a>
```
eysi09, line 9: This should point to `getting-stared/quickstart`

README.md
```diff
@@ -30,15 +30,15 @@ With Garden you can:
 
 ## Installation
 
-The fastest way to get started with Garden is by following our [quickstart guide](https://docs.garden.io/basics/quickstart).
+The fastest way to get started with Garden is by following our [quickstart guide](https://docs.garden.io/overview/quickstart).
```
eysi09, line 33: Same here.

README.md
```diff
@@ -30,15 +30,15 @@ With Garden you can:
...
 
 Otherwise:
 
 ```sh
 curl -sL https://get.garden.io/install.sh | bash
 ```
 
-For more installation options, see the [installation guide](https://docs.garden.io/basics/quickstart#step-1-install-garden).
+For more installation options, see the [installation guide](https://docs.garden.io/overview/quickstart#step-1-install-garden).
```
eysi09, line 41: I think we should move `guides/installation.md` to `getting-started/installation.md` and update this link so that it points to `getting-started/installation`.

core/src/plugins/kubernetes/kubernetes.ts
```diff
@@ -149,7 +149,7 @@ export const gardenPlugin = () => {
 
     For usage information, please refer to the [guides section](../../guides). A good place to start is
     the [Remote Kubernetes guide](${makeDocsLink`k8s-plugins/remote-k8s/README`}) guide if you're connecting to remote clusters.
-    The [Quickstart guide](${makeDocsLink`basics/quickstart`}) guide is also helpful as an introduction.
+    The [Quickstart guide](${makeDocsLink`overview/quickstart`}) guide is also helpful as an introduction.
```
eysi09, line 152: This should be `getting-started`

eysi09 (changes requested):

worldofgeese:
I should have addressed all your asks!

This has convinced me we need a docs contribution guide that guides a contributor through a docs change example step-by-step, like my own. Should also help @valeriegarden get going.

eysi09 (dismissed):

eysi09 (dismissed):

eysi09 (dismissed):

worldofgeese:
@eysi09 can you re-run please? Looks like a flaky test.

PR #5103, opened 2023-09-18
test: enable remote tests

stefreak:

We enable the remote tests to run in CI. This includes integration tests that are requiring some form of remote service access. In order to get these tests to work again, we have had to:

- Replace the tests using a remote container registry. Previously we accessed docker hub with encrypted credentials in the repo. After this PR we instead use Google Artifact Registry. Access is based on a google cloud user group which retrieves the credentials needed for both local and remote (k8s container) registry operations.
- Change test initialization to use the new credentials to be accessible both locally and in k8s containers
- Make sure tests are independent(*) and can run with a clean initial state
- Add/rewrite utility functions related to accessing and deleting images in the GAR and the local docker image cache
- Remove dead code related k8s build status and running k8s builds

Fixes #

There are a couple of test suites that are disabled:

- With the remote registry we have introduced shared state and cannot run tests concurrently.
- Buildkit acquires a lock to ensure the container exists, this lock is sometimes causing the entire test suite to block for up to a minute.
- Local mode tests that were disabled before we started this work

Tests in the disabled suites have been passing locally when we have been able to ensure there is no concurrent access to the registry. We plan to fix issues with concurrent and slow running tests in an upcoming PR. The affected tests are mainly related to local-mode and in-cluster building with kaniko and buildkit.


core/src/plugins/kubernetes/container/build/common.ts
```diff
@@ -327,6 +326,12 @@ export async function ensureUtilDeployment({
 
     // Check status of the util deployment
     const { deployment, service } = getUtilManifests(provider, authSecret.metadata.name, imagePullSecrets)
+
+    // TODO: think about this, it's ugly
```
stefreak, line 330: Likely this can be solved by using the Deployment strategy "recreate". If that works, we can remove this code

stefreak: Done, see https://github.com/garden-io/garden/pull/5125

core/test/data/test-projects/container/garden.yml
```diff
@@ -42,6 +43,10 @@ providers:
     kaniko:
       image: gcr.io/kaniko-project/executor:debug-perf
     deploymentRegistry: *deploymentRegistry
+  - <<: *base
+    environments: [kaniko-remote-registry]
+    buildMode: kaniko
+    deploymentRegistry: *deploymentRegistry
```
stefreak, line 49: this is the same as just `kaniko`– instead of having two identical environments with different names, could we instead call `init("kaniko", true)` instead of `init("kaniko-remote-registry")`

core/test/integ/helpers.ts
```diff
@@ -11,36 +11,106 @@ import { expect } from "chai"
...
+export async function deleteGoogleArtifactImage(packageName: string): Promise<void> {
+  const client = await getArtifactRegistryClient()
+
+  const fullName = `projects/garden-ci/locations/europe-west3/repositories/garden-integ-tests/packages/${packageName}`
+
+  try {
+    await client.deletePackage({ name: fullName })
+  } catch (err: unknown) {
+    if (isGCloudServiceError(err)) {
+      const gcloudErr = err as GCloudServiceError
```
stefreak, line 96: The cast is not necessary: isGCloudServiceError already narrows the type

core/src/plugins/container/helpers.ts
```diff
@@ -233,15 +233,46 @@ const helpers = {
...
+      return undefined
+    }
+
+    const imageIds = result.stdout.split("\n")
+    return { identifier, imageIds }
+  },
+
+  // Remove all images for a given identifier
+  async removeLocalImage(identifier: string, log: Log, ctx: PluginContext) {
+    const { identifier: _, imageIds } = (await containerHelpers.imageExistsLocally(identifier, log, ctx)) || {}
```
TimBeyer, line 258: We can just remove `identifier` here, right?

mkhq (approved):

mkhq (approved):

PR #5105, opened 2023-09-18
improvement(build): update moby/buildkit to v0.12.2

highb:

This PR upgrades the moby/buildkit to [v0.12.2](https://github.com/moby/buildkit/releases/tag/v0.12.2), which, among other improvements, adds the `--secret` flag which will allow secrets to be mounted into buildkit builds without including them in the final image layers. This is a security improvement, as it helps prevent secrets from leaking out in the image layers, if used properly by end users.

Technically, the secrets flag was added in [v0.12.0](https://github.com/moby/buildkit/releases/tag/v0.12.0), but I am assuming that it would be a better use of everyone's time if we went directly to the most recent Z release.

Fixes #

### DIFF
--- a/images/buildkit/Dockerfile
+++ b/images/buildkit/Dockerfile
@@ -1,4 +1,4 @@
-FROM moby/buildkit:v0.10.5@sha256:ca9d86324a000a0cc6d93ae9d0d5a9df750a194d0d43644c3a71fc6230ceba44 as buildkit
+FROM moby/buildkit:v0.12.2@sha256:8ea9857f95c2a0402c245bb0e94f36e2b5b4a1cb05e7ed322c213ed50607ce62 as buildkit
 
 RUN apk add --no-cache wget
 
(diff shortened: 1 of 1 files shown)


images/buildkit/Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN wget "https://github.com/GoogleCloudPlatform/docker-credential-gcr/releases/
   chmod +x /usr/local/bin/docker-credential-gcr && \
   rm docker-credential-gcr_linux_amd64-2.0.1.tar.gz
 
-FROM moby/buildkit:v0.10.5-rootless@sha256:3a5eca9b8d5d0e6cdcd0e756d607bf7386cd1b61950daf63afadee79b43ba8bf as buildkit-rootless
+FROM moby/buildkit:v0.12.2-rootless@sha256:0919807170af622451887366c17408dc9a946d04c6fe4fcca3071f9637f8598f as buildkit-rootless
```
shumailxyz, line 19: This only updates the buildkit-rootless. Should also update the buildkit version which is specified at [line # 1 of this file. ](https://github.com/garden-io/garden/blob/main/images/buildkit/Dockerfile#L1)

shumailxyz (commented):
Thank you for submitting the PR. ✨ 
Left a comment about updating the buildkit version in addition to updating the buildkit-rootless version.

images/buildkit/Dockerfile
highb: Ah, I missed that. I'll get it updated today.

highb: Done!

shumailxyz (approved):
Thank you so much @highb ✨

PR #4047, opened 2023-04-04
refactor(k8s): remove old system garden and improve local-k8s startup time

edvald:
This also gets rid of an annoying issue where the static directory needed to be in a git repo root.

We may want to hold off on merging this until we have our next RC out, since this will take a bit of manual testing.

PR #5120, opened 2023-09-21
improvement(git): don't require Garden `static` dir to be a Git repo

vvagaytsev:
This is a subset of the changes from #4047. The original PR is quite old and requires some conflict resolution.
This PR brings a quick win to improve the usability of Garden.

This PR should fix the following issues:
* #5030
* #2614


cli/src/add-version-files.ts
```diff
@@ -1,58 +0,0 @@
...
-  const moduleConfigs = await garden.getRawModuleConfigs()
-
-  return Promise.all(
-    moduleConfigs.map(async (config) => {
-      const path = config.path
-      const versionFilePath = resolve(path, GARDEN_VERSIONFILE_NAME)
-
-      const vcsHandler = new GitHandler({
-        garden,
-        projectRoot: STATIC_DIR,
```
vvagaytsev, line 37: This was the only place in the code where we used the `static` directory as a `projectRoot` of the `GitHandler` class instance.

stefreak (changes requested):
This is great! 🥳 

Can you remove `git init` in the following places in this PR, to verify that garden really works without git in the static directory in our e2e tests:
- https://github.com/garden-io/garden/blob/6c737a9056ea05df634438af50e3e7e49b2696d1/support/alpine.Dockerfile#L58
- https://github.com/garden-io/garden/blob/6c737a9056ea05df634438af50e3e7e49b2696d1/cli/src/build-pkg.ts#L89
- https://github.com/garden-io/garden/blob/6c737a9056ea05df634438af50e3e7e49b2696d1/support/buster.Dockerfile#L49
(Found those quickly scanning the code, but there might be more)

vvagaytsev:
Good catch, @stefreak! Thank you so much! I'll double-check the rest pieces of the existing code.

vvagaytsev:
@stefreak I've addressed the comments above in a686596a2a313cae8c415eb24e6474ddb78d8e32. Please check this again. I haven't found any other places where we init git repo for `static` dirs.

cli/src/add-version-files.ts
stefreak: How does that remove the git init requirement, if the only place we called `addVersionFiles` was in our yarn `build` script? What are these version files there for, and where did we read them?

vvagaytsev: It looks like these files were only used in the function [`resolveTreeVersion`](https://github.com/garden-io/garden/pull/5120/files#diff-5f17dbb72e9d791ba8b214f07c86f9d435613e277ebf9c65b4399283efd73b32L259) which is no longer used in 0.13.

It was used in 0.12 in the function [`resolveModuleVersion`](https://github.com/garden-io/garden/blob/0.12/core/src/garden.ts#L911).

The key change in this PR is the removal of `git init` command call in the `static` dir in [`build-pkg.ts`](https://github.com/garden-io/garden/pull/5120/files#diff-ccbc11fc635d77546db33d202b9adb7c4c1c4e759642d8c575de56aeeea8cbdcL95).

vvagaytsev: Ah, actually at least one necessary change was missing in this PR. I made it in 53efde570e8140f929b946bfe8534070295508fe.

vvagaytsev: Now I'm getting the same error for for `static/kubernetes/system` directory:
```console
Path ~/Downloads/garden-tmp/macos-arm64/static/kubernetes/system is not in a git repository root. Garden must be run from within a git repo. Please run `git init` if you're starting a new project and repository, or move the project to an existing repository, and try again.
```

It looks like I need to get some changes from the original PR #4047.

vvagaytsev: The rest part of the changes will be done in #5136.

stefreak: @vvagaytsev I'd suggest we add a test that triggers the error you described in https://github.com/garden-io/garden/pull/5120#discussion_r1336831047

This would be to ensure that the ingress controller deployments in the static directory continues to work when we make changes in the future.

Right now all tests passed, but real users would see this error, right?

vvagaytsev: @stefreak, right. We can add tests for ingress controllers in #5136. It looks like we don't have any tests for ingress controllers now.

vvagaytsev: The tests were added in #5136.

stefreak (approved):
Awesome 🥳

Issue #3438, opened 2022-12-13
garden deploy: deployments are not being scaled up to 1 replica


hnicke:
### Current Behavior
Since v0.12.47 we see a new behaviour:
A k8s deployment belonging to a garden kubernetes module with `spec.replicas: 1` which got manually scaled to `0` is not being scaled up when deploy via `garden deploy`.
The output is: `Already deployed`.

### Expected behavior
`garden deploy` should scale up the deployment to the specified replica count.
This was the behavior before 0.12.47.

### Reproducible example
* use garden v0.12.47
* use any k8s module with a k8s deployment
* deploy it using `garden deploy`
* scale it down manually: `kubectl scale deploy ... --replicas 0`
* deploy again using `garden deploy`.
  The output will say: `Already deployed`.
* we still have 0 pods.

#### != 0.13.7
Upon scaling down the deployment, overwrite its `garden.io/manifest-hash` annotation:

```
kubectl annotate --overwrite deploy ... garden.io/manifest-hash=scaleddown
```

Full example for a `scale-down` task:

```
kind: Module
type: exec
name: scale-down
description: |
  The 'scale-down' task completely scales down a system.
  After the task completed, there should be no running pods in the system.
tasks:
  - name: scale-down
    command:
      - bash
      - -c
      - |
        'set -eu
        kubectl="kubectl --context ${providers.kubernetes.config.context} --namespace '${providers.kubernetes.outputs.app-namespace}'"
        for resource_type in deployment statefulset; do
          if [ $($kubectl get $resource_type --no-headers | wc -l) -gt 0 ]; then
            $kubectl scale $resource_type --all --replicas 0
            # see https://github.com/garden-io/garden/issues/3438#issuecomment-1349264690
            $kubectl annotate $resource_type --overwrite --all garden.io/manifest-hash=scaledown
          fi
        done
        echo Successfully scaled down system'
```

#### == 0.13.7
Upon scaling down the deployment, delete the related garden-meta configmap:
```
kubectl delete configmap garden-meta-<garden-deploy-name>
```

Full example for a `scale-down` task:

```
kind: Module
type: exec
name: scale-down
description: |
  The 'scale-down' task completely scales down a system.
  After the task completed, there should be no running pods in the system.
tasks:
  - name: scale-down
    command:
      - bash
      - -c
      - |
        'set -eu
        kubectl="kubectl --context ${providers.kubernetes.config.context} --namespace '${providers.kubernetes.outputs.app-namespace}'"
[...]

PR #5122, opened 2023-09-21
fix(k8s): handle AEC-paused resources properly

vvagaytsev:
A follow-up fix for #4846 to address @stefreak 's [comment on the AEC](https://github.com/garden-io/garden/pull/4846#issuecomment-1695938718).

Fixes #

The commit-by-commit review might be easier :)

### DIFF
--- a/core/src/plugins/kubernetes/kubernetes-type/handlers.ts
+++ b/core/src/plugins/kubernetes/kubernetes-type/handlers.ts
@@ -45,6 +45,7 @@ import type { ActionLog } from "../../../logger/log-entry"
 import type { ActionMode, Resolved } from "../../../actions/types"
 import { deployStateToActionState } from "../../../plugin/handlers/Deploy/get-status"
 import { ResolvedDeployAction } from "../../../actions/deploy"
+import { isSha256 } from "../../../util/hashing"
 
 export const kubernetesHandlers: Partial<ModuleActionHandlers<KubernetesModule>> = {
   configure: configureKubernetesModule,
@@ -256,10 +257,12 @@ async function getResourceStatuses({
         return { resource: missingResource, state: "missing" } as ResourceStatus
       }
 
-      // Check if AEC has paused the resource
-      // TODO: remove this quickfix once we have implemented generic manifests/resources comparison
+      // TODO: consider removing this quickfix once we have implemented generic manifests/resources comparison
+      // Check if the "garden.io/manifest-hash" annotation is a valid sha256 hash.
+      // If it's not, consider the remote resource as outdated.
+      // AEC feature uses a dummy non-sha256 value to ensure the outdated state.
       const manifestHash = resource.metadata?.annotations?.[k8sManifestHashAnnotationKey]
-      if (manifestHash === "paused-by-aec") {
+      if (manifestHash && !isSha256(manifestHash)) {
         return { resource, state: "outdated" } as ResourceStatus
       }
 
--- a/core/src/plugins/kubernetes/kubernetes-type/handlers.ts
+++ b/core/src/plugins/kubernetes/kubernetes-type/handlers.ts
@@ -238,7 +238,7 @@ async function getResourceStatuses({
     return []
   }
 
-  const maybeDeployedResources: [ManifestMetadata, any][] = await Promise.all(
+  const maybeDeployedResources: [ManifestMetadata, KubernetesResource | null][] = await Promise.all(
     manifestMetadata.map(async (m) => {
       return [m, await api.readOrNull({ log, ...m })]
     })
--- a/core/src/plugins/kubernetes/kubernetes-type/handlers.ts
+++ b/core/src/plugins/kubernetes/kubernetes-type/handlers.ts
@@ -256,6 +256,7 @@ async function getResourceStatuses({
       }
 
       // Check if AEC has paused the resource
+      // TODO: remove this quickfix once we have implemented generic manifests/resources comparison
       const manifestHashAnnotationKey = gardenAnnotationKey("manifest-hash")
       const manifestHash = resource.metadata?.annotations?.[manifestHashAnnotationKey]
       if (manifestHash === "paused-by-aec") {
--- a/core/src/plugins/kubernetes/kubernetes-type/handlers.ts
+++ b/core/src/plugins/kubernetes/kubernetes-type/handlers.ts
@@ -253,9 +253,16 @@ async function getResourceStatuses({
           metadata: { name: m.name, namespace: m.namespace },
         }
         return { resource: missingResource, state: "missing" } as ResourceStatus
-      } else {
-        return await resolveResourceStatus({ api, namespace, resource, log })
       }
+
+      // Check if AEC has paused the resource
+      const manifestHashAnnotationKey = gardenAnnotationKey("manifest-hash")
+      const manifestHash = resource.metadata?.annotations?.[manifestHashAnnotationKey]
+      if (manifestHash === "paused-by-aec") {
+        return { resource, state: "outdated" } as ResourceStatus
+      }
+
+      return await resolveResourceStatus({ api, namespace, resource, log })
     })
   )
 }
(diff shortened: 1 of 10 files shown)


core/src/plugins/kubernetes/kubernetes-type/handlers.ts
```diff
@@ -253,9 +253,16 @@ async function getResourceStatuses({
...
         }
         return { resource: missingResource, state: "missing" } as ResourceStatus
-      } else {
-        return await resolveResourceStatus({ api, namespace, resource, log })
       }
+
+      // Check if AEC has paused the resource
+      const manifestHashAnnotationKey = gardenAnnotationKey("manifest-hash")
+      const manifestHash = resource.metadata?.annotations?.[manifestHashAnnotationKey]
+      if (manifestHash === "paused-by-aec") {
```
shumailxyz, line 261: should we may be create a constant for this like [we do in case of helm](https://github.com/garden-io/garden/blob/feat-update-secret-cmd/core/src/plugins/kubernetes/helm/status.ts#L28)?

shumailxyz: also, consider adding a todo to remove it once we implement the generic manifests comparison

vvagaytsev: Makes sense, thank you!

vvagaytsev: Addressed the items above in the recent commits, please check.

core/src/plugins/kubernetes/kubernetes-type/handlers.ts
```diff
@@ -253,9 +254,16 @@ async function getResourceStatuses({
           metadata: { name: m.name, namespace: m.namespace },
         }
         return { resource: missingResource, state: "missing" } as ResourceStatus
-      } else {
-        return await resolveResourceStatus({ api, namespace, resource, log })
       }
+
+      // Check if AEC has paused the resource
+      // TODO: remove this quickfix once we have implemented generic manifests/resources comparison
+      const manifestHash = resource.metadata?.annotations?.[k8sManifestHashAnnotationKey]
```
twelvemo, line 261: Isn't `k8sManifestHashAnnotationKey = "manifest-hash"` but we want to look for the annotation key `"aec-status"` here? The value of an annotation with the key `"manifest-hash"` would never be `"paused-by-aec"` right?

twelvemo (changes requested):
Overall looks good, but i think we might be using the wrong annotation key.

core/src/plugins/kubernetes/kubernetes-type/handlers.ts
shumailxyz: I think aec overrides the value of `manifest-hash` annotation

twelvemo: Oh i see, yes you are right of course. I was just getting confused because for helm we look for another annotation key.

shumailxyz: yes, for helm it's different.

shumailxyz (approved):

shumailxyz:
`e2e-run-actions` is failing but i think it's a flake

twelvemo (approved):
Super quick turnaround!

core/src/plugins/kubernetes/kubernetes-type/handlers.ts
vvagaytsev: Hmm.. It looks like Cloud sets `garden.io/manifest-hash` to `paused-by-aec`, see https://github.com/garden-io/garden-platform/blob/main/workflow-runner/src/runners/cleanup-runner/index.ts#L258-L260

The `garden.io/aec-status` seems to be used only if the resource in question is managed by Helm: https://github.com/garden-io/garden-platform/blob/main/workflow-runner/src/runners/cleanup-runner/index.ts#L262-L268

core/src/plugins/kubernetes/kubernetes-type/handlers.ts
```diff
@@ -253,9 +254,16 @@ async function getResourceStatuses({
...
+      if (manifestHash === "paused-by-aec") {
```
stefreak, line 262: Some people are using another value here, it would be good if we either also supported a generic value like just "paused", or if we check for the length of the string (=if it's not a hash)

See also: https://github.com/garden-io/garden/issues/3438#issuecomment-1349264690

stefreak: Should we maybe add a test for this behaviour, so we don't miss it next time? It was quite time-consuming to make sure that we do not regress here. A test might save a lot of time in the future when we make changes to this functionality again. Also it'd ensure that this change actually works

stefreak (dismissed):
Great work! 👍 

I think we need to still make sure that `deploy -f` always deploys even when the deploy is `ready` already; But that can be done in a separate PR.

shumailxyz:
@vvagaytsev Some of the tests are failing. I wonder if those are due to the manifest hash check you added.

vvagaytsev:
@shumailxyz thanks! I see, that's because of the annotation change. The tests share the same in-cluster resource. After the new test is executed, I'll ensure that everything is reverted back to the normal state.

@stefreak thanks! Yes, let's address the `--force` flag in a separate PR/

core/test/integ/src/plugins/kubernetes/kubernetes-type/handlers.ts
```diff
@@ -262,6 +265,41 @@ describe("kubernetes-type handlers", () => {
...
+        remoteDeploymentResource.metadata.annotations![k8sManifestHashAnnotationKey] = "non-sha256-hash"
+        await api.replace({ log, namespace, resource: remoteDeploymentResource })
+
+        const status = await getKubernetesDeployStatus(deployParams)
+        expect(status.state).to.equal("not-ready")
+        expect(status.detail?.state).to.equal("outdated")
+      } finally {
+        // Redeploy the resource to restore the original value of `k8sManifestHashAnnotationKey` annotation,
+        // because the remote resource state is used by the other tests.
+        await kubernetesDeploy(deployParams)
```
stefreak, line 299: I understand if that's a bigger change and you want to do it separately; But would it be better to run `kubernetesDeploy` in a `beforeEach` in a group for all the tests that rely on it being deployed?

stefreak: seeing `try {} finally {}` in a test is strange when mocha provides `before` and `after` primitives for these kinds of things

vvagaytsev: That requires more changes and re-grouping of the tests. We can do it later if we have more tests that modify remote resources that should be re-deployed on the test completion.

vvagaytsev: I added a dedicated config for this test case in 75c1b2461f9df6e98e2599f358e522564bdf642b.
Now we don't need to restore the remote resource state used by the other tests.

stefreak (dismissed):
👍

core/test/integ/src/plugins/kubernetes/kubernetes-type/handlers.ts
```diff
@@ -262,6 +265,38 @@ describe("kubernetes-type handlers", () => {
       expect(status.detail?.state).to.equal("outdated")
     })
 
+    it("should return outdated status when k8s remote resource version is not a valid sha256 hash", async () => {
+      // We use a standalone module here, because we need to modify the state of the remote resource to perform the test.
+      // Using of the `module-simple` will require remote resource redeployment in many other test cases here,
+      // and that qill slow down the test execution.
```
shumailxyz, line 271:
```suggestion
      // and that will slow down the test execution.
```

shumailxyz (dismissed):
Awesome 💯 

Just a typo that i found.

shumailxyz (approved):

PR #5124, opened 2023-09-21
refactor: strict typescript

TimBeyer:
This PR puts TS into strict mode.
I had to refactor a whole bunch of types, especially since we're using properties on classes as a way to pass type information along. I manage to get rid of that in most but not all cases.

Fixes #

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -31,7 +31,7 @@ import {
   renderCommandErrors,
   cliStyles,
 } from "./helpers"
-import { Parameters, globalOptions, OUTPUT_RENDERERS, GlobalOptions, ParameterValues } from "./params"
+import { ParameterObject, globalOptions, OUTPUT_RENDERERS, GlobalOptions, ParameterValues } from "./params"
 import { ProjectResource } from "../config/project"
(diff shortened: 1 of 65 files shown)


stefreak (commented):
Only one small comment, otherwise looks awesome! ❤️

core/src/cli/cli.ts
```diff
@@ -91,12 +91,12 @@ export class GardenCli {
       .sort()
       .filter((cmd) => cmd.getPath().length === 1)
 
-    let msg = `
-${cliStyles.heading("USAGE")}
-  garden ${cliStyles.commandPlaceholder()} ${cliStyles.optionsPlaceholder()}
+    let msg = dedent`
+      ${cliStyles.heading("USAGE")}
```
stefreak, line 95: Oh, maybe this was not indented due to the ANSI code bug in dedent?

TimBeyer: Oh that makes sense. Damn

stefreak (approved):
Wow, how did you do manage to push through and enable strict mode? Awesome work 🥇 ❤️ 
LGTM

PR #5128, opened 2023-09-22
chore: replace yarn with npm

TimBeyer:
This was a big one.
We had a huge ton of dependencies that were never explicitly declared and only worked because they were transitive dependencies. I had to find them all and list them explicitly.

I configured `npm` to not do any dependency hoisting for now since I wasn't sure if it's safe, so there is some redundancy in all the dependencies that are installed. At the same time it ensures that there's really no accidental usage of transitive dependencies anywhere.

I moved all the source files in the plugins and sdk into the `src` folder because there was a lot of implicitly working stuff that only worked that way because the output files were at the root of the package instead of a subdirectory. Also the directories kept getting polluted with compiled outputs making it hard to keep an overview. That's why now some imports from the SDK are a bit uglier, but they are a bit closer to the "truth".

The `pkg` bundling process now uses the same package-lock as the rest of the project and it simply installs from there which makes the install quite fast.

Sadly due to no hoisting the binary size became a bit larger, but I believe now that everything works we can try as a next step to re-enable the hoisting. I just didn't want to go down that path right away since this has already cost me a bit of my sanity.

There's probably some other changes in here as well that I forgot about, but in general it's now better aligned and structured with all dependencies explicitly listed, and installing packages now works just fine again.

Fixes #


stefreak (approved):
Wow, amazing work! 👏

vvagaytsev:
@TimBeyer the [`CONTRIBUTING.md guide`](https://github.com/garden-io/garden/blob/main/CONTRIBUTING.md) still mentions `yarn`. It would be nice to update that guide as well.

PR #5133, opened 2023-09-25
fix: add fsevents as a dependency

mkhq:

Add fsevents to the dependencies. It's loaded dynamically if the platform is darwin in [core/src/watch.ts ](https://github.com/garden-io/garden/blob/c500c2d71a709eb32c0a36eacbe10ce48c2e59e7/core/src/watch.ts#L55).

Fixes #


TimBeyer:
Don't we have our own version of `fsevents` in `core/lib`?
Also is it really optional if it is required for darwin?

mkhq:
> > Don't we have our own version of fsevents in core/lib?

> Yes, honestly not sure if we need that? I've removed it since now we include as an optional dependency.

We use it when packaging the macos build.

vvagaytsev (approved):
LGTM! 👍 Thank you!

PR #5136, opened 2023-09-26
improvement(k8s): remove old system garden and improve local-k8s startup time

vvagaytsev:
This and #5120 replace #4047.
This PR contains most part of the changes from #4047 - removal of the Garden system services..

This PR enables us to merge #5120 which in turn fixes issues for windows users.


static/kubernetes/nginx-kind-old.yaml
```diff
@@ -0,0 +1,326 @@
...
+      nodeSelector:
+        kubernetes.io/os: linux
+        ingress-ready: "true" # <- kind customization
+      tolerations: # <- kind customization
+        - key: node-role.kubernetes.io/master
+          operator: Equal
+          effect: NoSchedule
+      containers:
+        - name: nginx-ingress-controller
+          image: quay.io/kubernetes-ingress-controller/nginx-ingress-controller:0.30.0
```
stefreak, line 252: missing sha256 hash. it's important to add it to every single image we reference from the garden binary or the static directory, to make sure that docker containers that ship with garden do not change after the garden binary shipped (which increases security by preventing supply-chain attacks).

vvagaytsev: Thanks! I fixed that in e665126a9edeee78ac8b08ccbd00b2b9fa0ec1e1.

core/src/plugins/kubernetes/commands/cluster-init.ts
```diff
@@ -7,12 +7,13 @@
...
 import { PluginCommand } from "../../../plugin/command"
-import { prepareSystem, getEnvironmentStatus } from "../init"
+import { prepareEnvironment, getEnvironmentStatus } from "../init"
 import chalk from "chalk"
 
+// TODO: remove in 0.14
 export const clusterInit: PluginCommand = {
   name: "cluster-init",
-  description: "Initialize or update cluster-wide Garden services.",
+  description: "[DEPRECATED] Initialize or update cluster-wide Garden services.",
```
stefreak, line 16: The docs need to be still updated to reflect that this command is not needed anymore.

core/src/plugins/kubernetes/local/local.ts
```diff
@@ -64,6 +65,8 @@ async function prepareEnvironment(
 
   const setupIngressController = config.setupIngressController
 
+  const systemNamespace = await getSystemNamespace(ctx, provider, log)
```
vvagaytsev, line 68: @twelvemo do we need to calculate the `systemNamespace` here? We do not use this variable in the code but use `config.gardenSystemNamespace` instead.

vvagaytsev: Ah, we need that to ensure the system namespace existence.

core/src/plugins/kubernetes/integrations/nginx.ts
```diff
@@ -41,8 +49,8 @@ export async function helmNginxStatus(ctx: KubernetesPluginContext, log: Log): P
     log.debug(chalk.yellow(`Helm release status for ${HELM_INGRESS_NGINX_RELEASE_NAME}: ${status}`))
     return helmStatusMap[status] || "unknown"
   } catch (error) {
-    log.warn(chalk.yellow(`Unable to get helm status for ${HELM_INGRESS_NGINX_RELEASE_NAME} release: ${error}`))
-    return "unknown"
+    log.debug(chalk.yellow(`Helm release ${HELM_INGRESS_NGINX_RELEASE_NAME} missing.`))
+    return "missing"
```
vvagaytsev, line 53: Does the error here always mean that the helm release is missing? Should we keep it `unknown` as before?

twelvemo: I would say it should always implies missing, because we only reach this line if the helm command failed. Helm has a couple of statuses itself which we cover before. So we should only print this line if helm could not find the release at all.

vvagaytsev: Ok, thanks for the clarification!

core/src/plugins/kubernetes/ephemeral/config.ts
```diff
@@ -163,6 +165,9 @@ export async function configureProvider(params: ConfigureProviderParams<Kubernet
   }
   const { config: updatedConfig } = await base!(kubernetesPluginConfig)
 
+  // make sure that the system namespace exists
+  await getSystemNamespace(ctx, ctx.provider, log)
```
twelvemo, line 169: We need to make sure for this provider again that the namespace actually exists.

core/src/plugins/kubernetes/ephemeral/config.ts
```diff
@@ -143,6 +139,8 @@ export async function configureProvider(params: ConfigureProviderParams<Kubernet
   ]
   // set build mode to kaniko
   baseConfig.buildMode = "kaniko"
+  // set resource requests and limits defaults for builder, sync and util
+  baseConfig.resources = defaultResources
```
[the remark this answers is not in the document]
vvagaytsev: Good catch, thank you! 👍

static/kubernetes/nginx-ingress-class.yaml
twelvemo, line 1: Do we still need this file? I think we have a function for deploying the `ingressClass` now.

static/kubernetes/nginx-kind.yaml
twelvemo, line 1: We should be able to get rid of this file as well.

core/src/plugins/kubernetes/commands/uninstall-garden-services.ts
```diff
@@ -22,25 +21,11 @@ export const uninstallGardenServices: PluginCommand = {
...
-
-    log.info("")
-
-    const deployNames = deploys.map((s) => s.name)
-    const statuses = await actions.deleteDeploys({ graph, log, names: deployNames })
-
-    log.info("")
-
-    const environmentStatuses = await actions.provider.cleanupAll(log)
+    await ingressControllerUninstall(k8sCtx, log)
```
twelvemo, line 25:
```suggestion
     if (k8sCtx.provider.config.setupIngressController === "nginx") {
      await ingressControllerUninstall(k8sCtx, log)
    }
```
We should only remove it if it was installed by Garden or this specific Garden project in the first place.

static/kubernetes/nginx-ingress-class.yaml
vvagaytsev: Yes, we have had the function called `ingressClassCreate` since 13720e91c881e15dd496f8218c9e6ea7182e18e0. But it has not been used yet. I changed that in 7fdd8fa1ba1ff9b2656e0d60e052609c73ba5324, please take a look.

The file [`nginx-ingress-class.yaml`](https://github.com/garden-io/garden/pull/5136/files/9399f5d744c2b74792ed98cd70b9af57f48a9a6f#diff-64fb5b2e763ad016dab2de57687b03c7125ffa0ff7ec5cafe31ca002403a2f5e) was used only in microk8s ingress controller installation in [`local.ts`](https://github.com/garden-io/garden/pull/5136/commits/13720e91c881e15dd496f8218c9e6ea7182e18e0#diff-b2d8fd1b3e1cf8d96418cb3bf6a6a15efa3c4521bdb47de856aefaf4927adb7cL111).

Do we need to create the `IngressClass` explicitly only for microk8s? OR for any other cluster types as well?

static/kubernetes/nginx-kind.yaml
vvagaytsev: Good catch! Removed this in b172ef6f86344832b08ee55a9c630c8b422f6fc7.

core/src/plugins/kubernetes/commands/uninstall-garden-services.ts
vvagaytsev: Right! Good catch, thank you!

static/kubernetes/nginx-ingress-class.yaml
twelvemo: > Do we need to create the IngressClass explicitly only for microk8s? OR for any other cluster types as well?

You are right, we don't need to do that at all. If we use the helm chart, the `ingressClass` is deployed via the helm chart.
For microk8s and minikube, the addons already include the nginx `ingressClass`. This probably changed since the last time this was revisited.

And for kind clusters, we use the custom manifests that also include the `ingressClass`.  So we can get rid of  the `ingressClassCreate` function and remove it from microk8s nginx setup.

vvagaytsev:
Now this is ready for review. There is still one unresolved comment by @stefreak. Let's do a full review round, and then address all comments at once :)

core/src/plugins/kubernetes/nginx/default-backend.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  const provider = ctx.provider
+  const config = provider.config
+  const namespace = config.gardenSystemNamespace
+  const api = await KubeApi.factory(log, ctx, provider)
+  await api.deleteBySpec({ namespace, manifest: service, log })
+  await api.deleteBySpec({ namespace, manifest: deployment, log })
+}
+
+const GARDEN_DEFAULT_BACKEND_IMAGE =
```
stefreak, line 70: I believe it would be good if we declare all Docker images in `constants.ts` and use the `DockerImageWithDigest` type to ensure that we don't miss the sha256 when updating the image in the future.

stefreak (changes requested):
Great job 👍 
Let's hope that we can merge the PR to remove the .git repository in the static directory soon.

Moving all docker image references to `constants.ts` and declaring them using the type `DockerImageWithDigest` is important to me, the other comments are optional.

core/src/plugins/kubernetes/nginx/ingress-controller.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  if (clusterType === "kind") {
+    await kindNginxInstall(ctx, log)
+  } else if (clusterType === "microk8s") {
+    await microk8sNginxInstall(ctx, log)
+  } else if (clusterType === "minikube") {
+    await minikubeNginxInstall(ctx, log)
+  } else if (clusterType === "k3s") {
+    await helmNginxInstall(ctx, log, getK3sNginxHelmValues)
```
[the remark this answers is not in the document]
twelvemo: Really like that idea!

docs/k8s-plugins/advanced/rbac-config.md
```diff
@@ -87,14 +87,6 @@ metadata:
   namespace: garden-system
   name: user-<username>-common
 rules:
-  # Allow port forward to build-sync services
```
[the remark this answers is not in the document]
vvagaytsev: That was cherry-picked from the original PR. I don't know why it was changed. @edvald do you remember why it was changed?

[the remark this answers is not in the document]
edvald: I do. There were several rules there that we no longer need by default, since we’re not using the `garden-system` namespace as much. Probably good to review this closely, but looking at it again quickly it seems like the changes still make sense.

core/src/plugins/kubernetes/nginx/default-backend.ts
vvagaytsev: Done in 23774d16a10b43a5fe70c2569790418c1da0336e.

stefreak: @vvagaytsev There are still some image digests hidden in some of the manifests, e.g. core/src/plugins/kubernetes/nginx/nginx-kind-manifests.ts – maybe you can uncover more using the search

vvagaytsev: Thanks! I did the full-text search and fixed the remained code places in 57d56af93a35c7e394bcbb4b026d2feeea43f6a5.

stefreak: Love this!

stefreak (approved):
LGTM :shipit: 
image

core/src/plugins/kubernetes/nginx/ingress-controller.ts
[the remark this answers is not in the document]
vvagaytsev: @stefreak I've made the refactoring in https://github.com/garden-io/garden/compare/57d56af93a35c7e394bcbb4b026d2feeea43f6a5..2ea6202c39cf8afd185a6cb26311bb80ca5e3ade

Please review it again :)

core/src/plugins/kubernetes/commands/cluster-init.ts
vvagaytsev: Sure, I'll update the docs soon.

stefreak (dismissed):
Great additional changes! 👍 

I already meant to approve in my last review, but I accidentally only commented. But here's the approval again :)

core/src/plugins/kubernetes/nginx/ingress-controller.ts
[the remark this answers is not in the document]
stefreak: Looked good to me already, looks even better now :) 👍

core/src/plugins/kubernetes/commands/cluster-init.ts
stefreak: Should we also emit a deprecation warning of we want to officially deprecate this command?

vvagaytsev: Yes, I think so. I'l push a few more commits and will ask for another approval :)

vvagaytsev: Added a deprecation warning in 509d490a8f790dc4a3732ee4e5e1ec464a0c7f40.

vvagaytsev: > The docs need to be still updated to reflect that this command is not needed anymore.

@stefreak this command is still functional, we'll remove it in `0.14`. So, is it fine to keep the documentation as it is for now? Wdyt? Feel free to suggest/push a documentation update to this PR if necessary :)

stefreak: OK 👍

Issue #4638, opened 2023-06-14
0.13: [Bug]: JavaScript heap out of memory on `garden validate`


mdragilev:
### Current Behavior

After upgrading from 0.12.53 -> 0.13.[1|2], we are running into the an out of memory error.  Other commands like deploys run ok.  We use the validate command in our pull requests that impact garden resource definitions.

```
garden validate
Validate 

"apiVersion" is missing in the Project config. Assuming "garden.io/v0" for backwards compatibility with 0.12. The "apiVersion"-field is mandatory when using the new action Kind-configs. A detailed migration guide is available at https://docs.garden.io/tutorials/migrating-to-bonsai
Project configuration field `modules` is deprecated in 0.13 and will be removed in 0.14. Please use the `scan` field instead.
Garden v0.13 (Bonsai) is a major release with significant changes. Please help us improve it by reporting any issues/bugs here:
https://go.garden.io/report-bonsai
→ Run garden util hide-warning 0.13-bonsai to disable this warning.
ℹ garden               → Running in Garden environment thor.affirm-garden-affirm
ℹ providers            → Getting status...
ℹ providers            → Getting status...
✔ providers            → Done (took 0.2 sec)
ℹ graph                → Resolving actions and modules...
✔ graph                → Done (took 0.2 sec)
ℹ providers            → Preparing environment...
ℹ local-kubernetes     → Configuring...
ℹ providers            → Getting status...
✔ providers            → Cached (took 0.2 sec)
ℹ providers            → Run with --force-refresh to force a refresh of provider statuses.
ℹ providers            → Getting status...
✔ providers            → Cached (took 0 sec)
ℹ providers            → Run with --force-refresh to force a refresh of provider statuses.
ℹ graph                → Resolving actions and modules...
✔ graph                → Done (took 0.3 sec)
ℹ deploy.nginx-kind-new → outdated
ℹ deploy.nginx-kind-new → Deploying version v-7cc78df9b7...
ℹ deploy.nginx-kind-new → Waiting for resources to be ready...
ℹ deploy.nginx-kind-new → Resources ready
ℹ deploy.nginx-kind-new → Waiting for resources to be ready...
⚠ deploy.nginx-kind-new → Deployment/ingress-nginx-controller: MountVolume.SetUp failed for volume "webhook-cert" : secret "ingress-nginx-admission" not found
⚠ deploy.nginx-kind-new → Deployment/ingress-nginx-controller: RELOAD NGINX reload triggered due to a change in configuration
[...]

PR #5137, opened 2023-09-27
fix: memoize memory leak

TimBeyer:
Memoize was using an object as a cache key whose identity kept changing. Thus it never returned a cache hit and kept storing new objects in cache. With this fixed, memory consumption becomes a lot more reasonable, plus the cache actually yields the desired performance benefits.

Fixes https://github.com/garden-io/garden/issues/4638


shumailxyz (approved):
Awesome 💯

PR #5141, opened 2023-09-27
fix: otel collector enabled detection

TimBeyer:
I noticed that we were just checking if the array exists, which it always does.
Instead we need to check if it's empty.

Also removed the yarn release which we don't need anymore.

Fixes #


shumailxyz:
Woahhh  `−147,530`

shumailxyz (approved):

Issue #4158, opened 2023-05-03
[FEATURE]: Upgrade to Node 20


Walther:
### Background / Motivation

Node 20 has been released. It is the "current" release and will be upgraded to LTS status later this year.

### What should the user be able to do?

Run `garden` that has been built with the Node version 20.

### Why do they want to do this? What problem does it solve?

This has the potential to directly solve some issues present on older runtime versions.

### Suggested Implementation(s)

Read the release notes of both Node 19 and 20 for breaking changes, and then attempt the upgrade in a branch. Note to also update the necessary base images.

We should probably do this for the `0.13` branch / Bonsai release, as the `0.12` branch and release is already feature-frozen.

### How important is this feature for you/your team?

🌹 It’s a nice to have, but nice things are nice 🙂

vvagaytsev:
Remember about #4467.

PR #5161, opened 2023-09-29
fix: replace unzipper to resolve corrupt binaries on node > v18.15

### LINKED ISSUE #4467: Bug: plugin / tool installation can result in corrupt binaries
### Current Behavior

Currently, tools downloaded by Garden might be unzipped wrong, resulting in a corrupt executable. This affects at least `terraform`, but might also affect e.g. `helm`, `mutagen`, and so on.

Example output locally when running the `terraform-plugin` tests:

```
 16) Terraform action type
       apply-action command
         calls terraform apply for the action:
     Error: Command "/Users/walther/.garden/tools/terraform-0-13-3/692e8c519d913625/terraform init" failed with code Unknown system error -8:
      at ChildProcess.spawn (node:internal/child_process:413:11)
      at Object.spawn (node:child_process:757:9)
      at execa (/Users/walther/git/garden/node_modules/execa/index.js:80:26)
      at exec (/Users/walther/git/garden/core/src/util/util.ts:216:16)
      at PluginTool.exec (/Users/walther/git/garden/core/src/util/ext-tools.ts:78:16)
      at getWorkspaces (common.ts:262:3)
      at setWorkspace (common.ts:297:36)
      at Object.handler (commands.ts:95:7)
      at Context.<anonymous> (test/terraform.ts:379:9)
```

Example of corrupt executables in our CI: https://app.circleci.com/pipelines/github/garden-io/garden/17991/workflows/f023b92f-2435-443c-94fd-1bd0b57cf084/jobs/314086

### Expected behavior

Tools get downloaded and extracted correctly, resulting in executables that work.

### Reproducible example

```
rm -rf ~/.garden/tools/terraform-*
cd garden/plugins/terraform/test
yarn run test
# all tests fail

ls -halF ~/.garden/tools/terraform-*/*/terraform
file ~/.garden/tools/terraform-*/*/terraform 
# the files are "binary data", not recognized as valid executables
```

### Workaround

Manually download the relevant .zip or .tar.gz, manually extract it, and move the executable to the location garden expects it within `~/.garden/tools/*`

### Suggested solution(s)

- Wait for a fix in the upstream `unzipper` library
- Debug the library to find a fix, submit PR
- Switch to another library or libraries (we use both .zip and .tar.gz formats, some libraries only support one or the other)

### Additional context

This is an adverse interaction between the `unzipper` library and Node version `>= 18.16.0` which we use.

See the upstream issue here: https://github.com/ZJONSSON/node-unzipper/issues/271
The issue description also has a minimal repro for local testing.

### Your environment

This happens both on local macOS as well as in our CI on Linux.
This affects both `0.13` as well as `0.12` versions released after Jan 30th.

shumailxyz:
Replaces `unzipper` with `unzip-stream` due to upstream issue https://github.com/ZJONSSON/node-unzipper/issues/271 not being fixed yet. And that was blocking us from updating our node version from v18.15.

Replaced with `unzip-stream` as the api is almost the same and it does resolve the issue. 

Also a step forward for https://github.com/garden-io/garden/issues/4158

Fixes #4467

#4467 was previously fixed by pinning the node version to 18.15.

- I built and published the [runner image to dockerhub](https://hub.docker.com/layers/gardendev/circleci-runner/18.18.0/images/sha256-687ab1e9f59050898d21b7531410cc449ba9acd1ab14ce5f642f4cd91745f7fe?context=explore) already and the CI uses the new image for this PR.
 
- Also looked into the [yauzl](https://www.npmjs.com/package/yauzl) as that seems to be most popular package for dealing with zip files however yauzl doesn't support streams.

- Please also test it by running and using `garden util fetch-tools`.


shumailxyz:
Example, from few months back, of the failing CI step due to corrupt binaries issue before we downgraded to node v18.15.

https://app.circleci.com/pipelines/github/garden-io/garden/17991/workflows/f023b92f-2435-443c-94fd-1bd0b57cf084/jobs/314086

TimBeyer (commented):
That's great. Should we also load [`pkg`](https://github.com/vercel/pkg) from github so that it uses the newest `pgk-fetch` version and updates the CLI node version then?

shumailxyz:
@TimBeyer  We could, my only concern is about the stability if `pkg` if we fetch the `main` branch as that could be unstable. Although, it doesn't seem to be actively developed anymore so the risk is low.

TimBeyer:
We also should pin it to a specific commit and then revisit that every once in a while.

shumailxyz:
How about this approach and overriding the `pkg-fetch`?

https://github.com/vercel/pkg/issues/1998#issuecomment-1736417818

TimBeyer:
Actually let's try this instead: https://github.com/vercel/pkg/issues/1998#issuecomment-1736417818

TimBeyer:
Haha exactly

TimBeyer:
Package overrides are somehow fundamentally broken in `npm` so we went with the github install instead.

TimBeyer:
`pkg-fetch` in the newest version doesn't have prebuilt binaries for `macos-arm64` and would need them to be compiled during the build. That fails since we're not building on macos. 
Will skip this for now and come back to this some time in the future.

TimBeyer (approved):

vvagaytsev (changes requested):
There are some test failures, please check.

stefreak:
This has been superseded by #esm (We had to cherry-pick this in the ESM branch)

PR #4984, opened 2023-08-25
fix: ensure using next free port for dev console

vvagaytsev:
**What this PR does / why we need it**:
A quickfix. Error handling still requires some refactoring.

**Which issue(s) this PR fixes**:

Fixes #4969

**Special notes for your reviewer**:

PR #5163, opened 2023-09-29
fix(server): ensure using a free port for dev console

### LINKED ISSUE #5158: Garden server fails silently on port conflicts
### Current Behavior

If I run `garden dev` in a given terminal window it starts a server on the default port, `9777`. 

Now, I I run another `garden dev` command in a different window it also the picks the same port. It attempts to start the server at that port which fails silently. 

The error is never logged and Garden continues running, so to the user, everything looks okay. However, in this state, the web dashboard won't work.

### Expected behavior

Firstly, that Garden picks the next available port if the current one is in use. The library we use for picking ports doesn't appear to do that and we may need to implement this on our end.

Second, that Garden doesn't just swallow these errors. Here's the offending code: https://github.com/garden-io/garden/blob/main/core/src/server/server.ts#L184

If we were to log the error there, we'd see something like:

```console
Error Error: listen EADDRINUSE: address already in use ::1:9777
    at Server.setupListenHandle [as _listen2] (node:net:1740:16)
    at listenInCluster (node:net:1788:12)
    at GetAddrInfoReqWrap.doListen [as callback] (node:net:1937:7)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:109:8)
    at GetAddrInfoReqWrap.callbackTrampoline (node:internal/async_hooks:130:17) {
  code: 'EADDRINUSE',
  errno: -48,
  syscall: 'listen',
  address: '::1',
  port: 9777
}
``` 

Instead this information is just swallowed and the user thinks that Garden is running as expected. 

### Reproducible example

- Add a console.log to [this callback](https://github.com/garden-io/garden/blob/main/core/src/server/server.ts#L184) that logs the error
- Run `garden dev` in one terminal window and observe how it starts up without isses
- Run `garden dev` in a second terminal window and observe how the error is logged but that Garden otherwise continues running and that you wouldn't know the server had issues without that console.log.

### Workaround

Manually set the port when multiple Garden dev commands. You can e.g. run the first one with `garden dev`, second one with `garden dev --port 9778`, etc. 

### Suggested solution(s)

See "Expected behaviour" above.

### Your environment

- macOS M1, Monterey
- Latest main

vvagaytsev:

Patches #4984
Fixes #5158

### DIFF
--- a/core/src/server/server.ts
+++ b/core/src/server/server.ts
@@ -204,9 +204,28 @@ export class GardenServer extends EventEmitter {
       })
     }
 
-    if (this.port) {
-      await _start()
+    if (this.port !== undefined) {
+      try {
+        await _start()
+      } catch (err) {
+        // Fail if the explicitly specified port is already in use.
+        if (isEAddrInUseException(err)) {
+          throw new ParameterError({
+            message: dedent`
+            Port ${this.port} is already in use, possibly by another Garden server process.
+            Either terminate the other process, or choose another port using the --port parameter.
+          `,
+          })
+        } else if (isErrnoException(err)) {
+          throw new CommandError({
+            message: `Unable to start server: ${err.message}`,
+            code: err.code,
+          })
+        }
+        throw err
+      }
     } else {
+      let serverStarted = false
       do {
         try {
           this.port = await getPort({
@@ -216,25 +235,21 @@ export class GardenServer extends EventEmitter {
             alternativePortRange: [defaultWatchServerPort - 1, defaultWatchServerPort - 50],
           })
           await _start()
+          serverStarted = true
         } catch (err) {
           if (isEAddrInUseException(err)) {
-            throw new ParameterError({
-              message: dedent`
-            Port ${this.port} is already in use, possibly by another Garden server process.
-            Either terminate the other process, or choose another port using the --port parameter.
-          `,
-            })
+            this.log.debug(`Unable to start server. Port ${this.port} is already in use. Retrying with another port...`)
           } else if (isErrnoException(err)) {
             throw new CommandError({
               message: `Unable to start server: ${err.message}`,
               code: err.code,
             })
           }
-
           throw err
         }
-      } while (!this.server)
+      } while (!serverStarted)
     }
+    this.log.info(chalk.white(`Garden server has successfully started at port ${chalk.bold(this.port)}.\n`))
 
     const processRecord = await this.globalConfigStore.get("activeProcesses", String(process.pid))
 
--- a/core/src/server/server.ts
+++ b/core/src/server/server.ts
@@ -228,8 +228,7 @@ export class GardenServer extends EventEmitter {
           this.port = await getPort({
             host: hostname,
             port: defaultServerPort,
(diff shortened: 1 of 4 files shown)


vvagaytsev:
@eysi09 should we use [`defaultWatchServerPort = 9777`](https://github.com/garden-io/garden/blob/main/core/src/server/server.ts#L52-L54) or [`defaultServerPort = 9700`](https://github.com/garden-io/garden/blob/main/core/src/commands/serve.ts#L26) as a default port value for `serve` and `dev` commands?

The `--help` documentation says it should be `9700`. So I'll change it to be `defaultServerPort = 9700`.

TimBeyer:
How does the front end know to connect to this port? Do we store that info on the session? Or is this another different server?

vvagaytsev:
@TimBeyer we set it as `localServerPort` when we call [`CloudApi.registerSession`](https://github.com/garden-io/garden/blob/main/core/src/cloud/api.ts#L620).

core/src/server/server.ts
```diff
@@ -197,20 +201,52 @@ export class GardenServer extends EventEmitter {
...
+            this.log.debug(`Unable to start server. Port ${this.port} is already in use. Retrying with another port...`)
+          } else if (isErrnoException(err)) {
+            throw new CommandError({
+              message: `Unable to start server: ${err.message}`,
+              code: err.code,
+            })
+          }
+          throw err
+        }
+      } while (!serverStarted)
```
TimBeyer, line 247: `getPort` already tries all the ports.
Won't this loop be potentially endless then?
Also I think in that case the error message is misleading because effectively the entire port range would be already in use, not just `this.port`.

TimBeyer: Oh I see, this is because it might be that a port is assigned and then the server starts and the port might be unassigned again, so we try again.

TimBeyer (approved):

core/src/server/server.ts
vvagaytsev: Actually, alternativePortRange can be omitted, it shouldn't be used here, see the [usage docs](https://github.com/unjs/get-port-please):

> Alternative port range to check as fallback when none of the ports are available. Default is `[3000, 3100]` (only when `port` is unspecified.)

If none of the ports from the range are available, then a random port will be picked. I tested that locally.

vvagaytsev: I've removed `alternativePortRange` in f28d0b367846059fa4e7141b7077d6d373cfded2.

TimBeyer (approved):

TimBeyer (approved):

vvagaytsev:
Now the tests are green

TimBeyer (approved):

eysi09:
cc @trymbill 

This has already been merged but wondering if the work you've been doing makes assumptions about ports?

trymbill:
> [...] if the work you've been doing makes assumptions about ports?

Desktop handles grabbing a free port by itself (see: https://github.com/garden-io/garden-platform/blob/61a0d724716a5324400bbe903af912aa4c8b3dd8/desktop/src-tauri/src/garden_core_process_manager.rs#L89-L95), so this shouldn't change anything on that end.

PR #5182, opened 2023-10-04
chore: bump dependencies

TimBeyer:
Removes gulp and replaces the header check with eslint.
Bumps nodemon and otel dependencies.

This fixes a few security notifications, but definitely not all of them.
Sadly especially the transitive dependencies from modules that now only get released as ESM cannot easily be bumped without us switching to ESM. So eventually we may have to tackle this.

Fixes #


stefreak (approved):
Awesome 🥳

vvagaytsev:
It looks like the pre-push hooks are broken now. I'm getting this error:
```console
husky > pre-push (node v18.15.0)

> garden@0.13.16 check-pre-push
> npm run check-package-lock && npm run check-licenses && npm run lint


> garden@0.13.16 check-package-lock
> git diff --quiet HEAD -- package-lock.json || (echo 'package-lock.json is dirty!' && exit 1)

npm ERR! Missing script: "check-licenses"
npm ERR! 
npm ERR! Did you mean this?
npm ERR!     npm run add-licenses # run the "add-licenses" package script
npm ERR! 
npm ERR! To see a list of scripts, run:
npm ERR!   npm run

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/vladimirvagaytsev/.npm/_logs/2023-10-04T20_38_21_774Z-debug-0.log
husky > pre-push hook failed (add --no-verify to bypass)
error: failed to push some refs to 'https://github.com/garden-io/garden.git'
```

Should I make any local changes to reconfigure the hooks?

vvagaytsev:
I tried to change `npm run check-licenses` to `npm run add-licenses` in `package.json` ([L70](https://github.com/garden-io/garden/blob/main/package.json#L70) and [L74](https://github.com/garden-io/garden/blob/main/package.json#L74)) but it failed with another error:
```console
husky > pre-push (node v18.15.0)

> garden@0.13.16 check-pre-push
> npm run check-package-lock && npm run add-licenses && npm run lint


> garden@0.13.16 check-package-lock
> git diff --quiet HEAD -- package-lock.json || (echo 'package-lock.json is dirty!' && exit 1)


> garden@0.13.16 add-licenses
> copyright-header --license-file support/license-header.txt --syntax support/license-syntax.yaml --add-path core/src:core/test --output-dir .

sh: copyright-header: command not found
husky > pre-push hook failed (add --no-verify to bypass)
```

It looks like `copyright-header` script is missing in the repo.

stefreak:
Oh god, I'd personally prefer if we removed these hooks and instead relied on the CI for this. WDYT?
Or alternatively used something that only checks the files you have actually touched and completes super quickly, for example https://pre-commit.com/

vvagaytsev:
It's nice to have hooks to catch linting issues early. Pre-commit sounds good. Agree, only the changes files can be inspected, the rest part van be checked in CI.

TimBeyer:
I'd prefer not to bring a python dependency into the project (pre-commit).
I'm a bit confused why this fails now though, and not before.
We didn't touch any of those dependencies and only removed the gulp script and gulp.
Unless this was an accidental transitive dependency, this shouldn't have had that impact.
If it was, then we just need to add it back explicitly.

TimBeyer:
Ah I see, the hook still called `npm run check-licenses`. 
That's no longer needed because it's part of `lint` now which happens right after.
I simply forgot to remove them from `"check-all"` and `"check-pre-push"`

PR #5187, opened 2023-10-04
feat(k8s): add support for patching manifests

### LINKED ISSUE #5082: [FEATURE]: Re-use existing Kubernetes manifests without modifying them
# Feature Request

## Background / Motivation

One of Garden's philosophies is _automation over abstraction_. It should be able to take existing config (e.g. Dockerfiles and K8s manifests) and add the automation that empowers users to re-use it across all stages of software delivery. E.g. for end-to-end testing or for developing in production like environments.

This currently works great with Helm but we're missing a proper story for projects that already have K8s manifests but aren't using Helm. 

In that case, you _should_ be able to use the `kubernetes` Deploy action and just point Garden to the manifests files, something like:

```
kind: Build
name: api
type: container
---
kind: Deploy
name: api
type: kubernetes
spec:
  files: [./my-manifests.yml]
```  

The problem however is that you need to reference the image build in the Pod spec for the manifest. This means you need to modify every single K8s manifest that references a image from your source code and set the following:

``` 
# For a given container stanza in a K8s Pod spec 
name: api
image: ${actions.build.api.outputs.deployment-image-id}
```

At this point, these manifests aren't valid K8s manifests anymore and you've broken the rest of your toolchain.

It simply doesn't work and you either need to duplicate config or change tooling.

## What should the user be able to do?

Drop Garden into an existing K8s project with minimal Garden config and zero modification to existing code / config.

## Why do they want to do this? What problem does it solve?

To be able to use Garden to add automation to their software delivery workflows without breaking other tooling.

## Suggested Implementation(s)

We have a few options here. In my opinion we should be optimising for something very simple here. It's not power user stuff, rather something that will feature in most getting started / onboarding docs and we really shouldn't have to go out of our way to explain.

### Option 1

Add a basic field to the `kubernetes` Deploy spec such as: 

```
overwriteImage:
  kind: Deployment
  name: api
  containerName: api
  buildAction: api # <--- The Garden build Action
```

**Pros:**
* Relatively simple syntax

**Cons:**
* Very limited and users may want to overwrite other fields as well.

### Option 2

Add a `patchResources` field to the spec and that allows the user to identify the resource they want to patch and define the patch proper which would be applied per the [json merge patch spec](https://datatracker.ietf.org/doc/html/rfc7386).  Note that we'd need to handle `containers` specifically or require users to list all the containers in the patch because it just overwrites arrays. 

Example:

```
kind: Deploy
name: api
kind: kubernetes
spec:
  files: [./my-manifests.yaml]
  patchResources:
    - kind: Deployment # <--- Could also be Pod or even ReplicaSet 
      name: api
      patch: # General purpose patch, applied via the JSON merge patch spec
        spec:
          replicas: 1 
      imagePatches: # <--- Specifically patch a given container image. Could also be more flexible `containerPatches`  
        - containerName: api
           newImage:  ${actions.api.outputs.deployment-image-id}
           # <--- Optionally support just setting image or image tag?
```
 
See inline comments for notes on syntax which is very much up for debate.

**Pros:**
- Flexible

**Cons:**
- More verbose. But if we imagine the user being a K8s user already (given they have all the manifests) it should feel familiar.

I'm personally partial to this approach or some version thereof.

### Option 3

Same as Option 1 but use Kustomize syntax and nest under existing `kustomize` field:

```
kind: Deploy
name: api
kind: kubernetes
spec:
  kustomize:
    spec: # <--- Basically what you'd put in kustomize.yaml
      - baseResources: ...
         images: # ...
```

**Pros:**
- Leaning on (somewhat) known concepts

**Cons:**
- Kustomize doesn't support passing args to the CLI so we'd to quite some work in terms of parsing the config; finding the base resources; staging to a build dir; creating a legit kustomize file; applying `kustomize build`
- Confusing for those not familiar with `kustomize`

### Option 4

Support inline templates in kustomize files. 

**Pros:**
- Familiar to existing kusomtize users

**Cons:**
-  You're back to breaking config unless you have a dedicated Garden overlay
- Also need to render the manifests and write back to the file system before doing `kustomize build` due its limitations
- Confusing for those not familiar with `kustomize`

## How important is this feature for you/your team?

The reason for that choice is that I was working on "Gardenifying" the Google [microservices demo](https://github.com/GoogleCloudPlatform/microservices-demo/tree/main/kubernetes-manifests) and wanted to submit it to the repo. 

I can't do that without modifying manifests which would never get approved so Garden can't really be used.

eysi09:

From the commit message:

```console
This change introduces a \`patchResources\` field to the \`kubernetes\`
Deploy action that allows users to overwrite the values of manifests in
Garden config without modifying the manifest itself.

This enables users to drop Garden into existing K8s projects without
making any other changes to code or config.

Under the hood, Garden uses the `kubectl patch` command. The behaviour
is documented here:

https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/
```

I also added an example project that demonstrates this functionality and updated the [k8s-deploy-config-templates](https://github.com/garden-io/garden/tree/main/examples/k8s-deploy-config-templates) example to use the same pattern.

To do:

- [ ] ~Update docs on Kubernetes action types (we can do that in a follow up PR if needed)~

I started writing the docs on a different branch to keep things simple.

Fixes #5082

The large number of line changes is due to the example projects. The main feature is pretty simple.


core/src/plugins/kubernetes/types.ts
```diff
@@ -52,6 +52,14 @@ export type KubernetesResource<T extends BaseResource | KubernetesObject = BaseR
       [P in Extract<keyof T, "spec">]: Exclude<T[P], undefined>
     }
 
+export interface KubernetesPatchResource {
+  name: string
+  // TODO @eysi: Can we get a better type here?
+  kind: string
```
eysi09, line 58: I just went for a string type here instead of listing all K8s object types.

core/src/plugins/kubernetes/kubernetes-type/config.ts
```diff
@@ -68,6 +70,27 @@ const kubernetesResourceSchema = () =>
...
+        dedent`
+        The patch strategy to use. One of 'json', 'merge', or 'strategic'. Defaults to 'strategic'.
+
+        You can read more about the different strategies in the offical Kubernetes documentation at:
+        https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/
+        `
+      )
+      .default("strategic")
+      .optional(),
+    patch: joi.object().required().description("The patch to apply").unknown(true),
```
twelvemo, line 91:
```suggestion
    patch: joi.object().required().description("The patch to apply.").unknown(true),
```

twelvemo (approved):
Great job! The examples are also super sound, nothing to improve imho. One nit-pick about a missing dot at the end of the sentence but that's it. Leaving it up to you to if you want to add it, otherwise approving.

twelvemo (dismissed):

twelvemo (approved):

PR #5190, opened 2023-10-05
fix(helm): use build path for charts for helm modules converted to ac…

twelvemo:
…tions

Fixes a compatibility issue when using Helm modules. In Garden 0.12.x Helm modules always had an implicit build and charts and their dependenies were resolved in the build directory under `.garden/build/<module-name>`. However Helm deploy actions and their dependencies are resolved in the action directory. When converting from modules to actions, we were creating a `dummyBuild` which is a build of type exec for some scenarios e.g. when a `base` chart was explicitly set in the Helm module. When this was not the case the converted Helm deploy action would use the action directory for all Helm operations. This leads to a compatibility issue e.g. when referencing local dependency charts with relative paths. This PR creates a `dummyBuild` for every converted Helm module, so that every converted Helm module still uses the build directory.
It also fixes a bug, where the `module.path` was referencing the path of the node module and not the garden module, which is `moduleConfig.path` in the context of this function. This resulted in only a part of the Helm chart to be copied to the build directory.
Adds tests to check if the correct base directory is used for both converted modules and native actions.

Fixes #


shumailxyz:
@twelvemo The tests are failing. Can you please check those?

twelvemo:
> @twelvemo The tests are failing. Can you please check those?
Just did. The failing integ tests about includes in helm charts was actually wrong. It was changed [here](https://github.com/garden-io/garden/pull/3944/files#diff-cab361ce677f2c0000f1dee600d40da78bb4cde51693ba4ce4400213cbc7dc6bL102) to match the behavior resulting from the bug of comparing the node module path for helm charts and setting the includes according to it. Fixed the test.

core/src/plugins/kubernetes/helm/handlers.ts
```diff
@@ -31,28 +31,30 @@ export const helmModuleHandlers: Partial<ModuleActionHandlers<HelmModule>> = {
...
-
     if (dummyBuild) {
       actions.push(dummyBuild)
     } else {
-      // We make a dummy build without a `copyFrom` or any build dependencies. This is to ensure there's a build action
-      // for this module if it's used as a base by another Helm module.
-      actions.push(makeDummyBuild({ module, copyFrom: undefined, dependencies: undefined }))
+      // We create a dummy build without a `copyFrom` or any build dependencies, to ensure there's a build action
+      // for this module. This is needed for compatibility reasions e.g. if there was a `base` field on the module
+      // or if a helm chart referencences dependent local charts relative to the modules build directory.
```
shumailxyz, line 45:
```suggestion
      // or if a helm chart references dependent local charts relative to the modules build directory.
```

shumailxyz (approved):
Thank you @twelvemo. Great work ✨

core/src/plugins/kubernetes/helm/module-config.ts
```diff
@@ -210,7 +210,7 @@ export async function configureHelmModule({
     ]
   }
 
-  const yamlPath = join(module.path, chartPath, helmChartYamlFilename)
+  const yamlPath = join(moduleConfig.path, chartPath, helmChartYamlFilename)
```
shumailxyz, line 213: Oopss :D

PR #5216, opened 2023-10-09
Docs k8s action types nested by kind

eysi09:

This main goal was updating the docs on the `kubernetes` Deploy action and clarifying best practices. 

As I was doing that I realised that it's a bit awkward to stuff all of this content under the action _type_ so I added a second level of nesting for the action kind. 

This is consistent with the user journey. You start by picking your plugin, in this case Kubernetes. You then start defining the builds, deploys, etc of your system and then add the corresponding Garden action. 

That is, you think to yourself: "I need to deploy something to Kubernetes" so you look for Kubernetes then Deploy and then see the types available to you (e.g. `helm`, `kubernetes`, etc). 

This also makes the content for each page a lot more manageable. 

Finally, I also used the chance to update the `exec` plugin doc. You can argue that we should split that into subpages as well but we can do that in a follow up commit if needed.

Before:

Screenshot 2023-12-01 at 11 30 12

After:

Screenshot 2023-12-01 at 11 30 34

Fixes #


thsig (dismissed):
Good stuff! Left a few optional comments there.

thsig (approved):

PR #5225, opened 2023-10-10
fix(cloud): send correct action version in status event payload

shumailxyz:
Use resolved action for version calculation when sending the get status results event to cloud. Before this, we were not using the resolved action when preparing the payload, and there was the version mismatch between the action version in the cloud event and the action version stored in k8s Configmap.

The status handlers for all action types ([test](https://github.com/garden-io/garden/blob/main/core/src/tasks/test.ts#L76), [build](https://github.com/garden-io/garden/blob/main/core/src/tasks/build.ts#L50), [deploy](https://github.com/garden-io/garden/blob/main/core/src/tasks/deploy.ts#L79), [run](https://github.com/garden-io/garden/blob/main/core/src/tasks/run.ts#L58)) already use the resolved action.

Fixes #

### DIFF
--- a/core/src/tasks/base.ts
+++ b/core/src/tasks/base.ts
@@ -432,7 +432,10 @@ export function emitGetStatusEvents<
       const donePayload = makeActionCompletePayload({
         result,
         startedAt,
-        action: this.action,
+        // Use executedAction instead of this.action here for correct version
+        // as executedAction is the resolved action and this.action is unresolved.
+        // The action is resolved in method.apply and getStatus handlers use the resolved action.
+        action: result.executedAction,
         operation: methodName,
         force: this.force,
         sessionId: this.garden.sessionId,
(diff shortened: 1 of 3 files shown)


core/src/tasks/base.ts
```diff
@@ -432,7 +432,10 @@ export function emitGetStatusEvents<
       const donePayload = makeActionCompletePayload({
         result,
         startedAt,
-        action: this.action,
+        // Use executedAction instead of this.action here for correct version
+        // as executedAction is the resolved action and this.action is unresolved.
+        // The action is resolved in method.apply and getStatus handlers use the resolved action.
+        action: result.executedAction,
```
stefreak, line 438: Would it make sense to change the type signature of `makeActionCompletePayload` to expect `Resolved<Action>` instead of just `Action` to use the type system to prevent the same mistake in the future?

PR #5226, opened 2023-10-10
chore: stricter linter

TimBeyer:

Makes the linter stricter, partly for better code, partly to ensure we eventually have a smoother time changing to ESM.
Part of this required changing all dynamic imports.
There are still a whole bunch of `import foo = require("foo")` style imports around which we need to get rid of next.

The settings around `any` use are violated too many times to do this in a single PR so I made them warnings only.
We should start improving those step by step.

For reviewing this I'd recommend skipping the `linter autofixes` commit since we trust the linter there, and then review the rest step by step.

Fixes #


core/package.json
```diff
@@ -67,7 +67,6 @@
     "execa": "^4.1.0",
     "fast-copy": "^3.0.1",
     "fs-extra": "^11.1.0",
-    "fsevents": "^2.3.3",
```
stefreak, line 70: Was fsevents supposed to be removed here?

TimBeyer: It's an `optionalDependency` and I guess was added manually as such with it showing up twice. 
I ran `npm install` which probably removed it from the `dependencies` section and kept it only in `optionalDependencies`.

stefreak (approved):
🚀

PR #5229, opened 2023-10-11
improvement: using a new release endpoint for self-update

### LINKED ISSUE #4893: [Bug]: github API rate-limiting can occur in `garden self-update`
### Current Behavior

If github is seeing too many requests to the API from your IP address, for example if you're working in an office or using a company VPN with NAT, it'll rate limit your API calls.

In garden self-update this can look like this:
```
% garden self-update
Update Garden 🗞️

No installation directory specified via --install-dir option. Garden will be re-installed to the current installation directory: /opt/homebrew/Cellar/garden-cli/0.13.9/libexec

Response code 403 (rate limit exceeded)
```

### Expected behavior

I would expect garden self-update to "just work" even when using IP addresses that GitHub is rate limiting, or alternatively the error message should contain instructions how to update Garden manually or how to supply Github access token to work around the rate limit.

### Reproducible example

Not easy to reproduce,  but we know this can happen with unauthenticated requests to the GitHub api, e.g. https://github.com/garden-io/garden/blob/main/core/src/commands/self-update.ts#L98

Use the following example at your own risk (successfully reproduces the issue for me :))

```
#!/bin/bash
while true; do
  garden self-update &
  sleep 0.1
done
```

### Workaround

You can use the methods described in https://docs.garden.io/guides/installation to install the latest version of garden manually as a workaround.

### Suggested solution(s)

- Use the version-check service that we host server side, to provide latest version information to users. This service can cache the version information to reduce calls to the GitHub API and use an auth token to authenticate requests.
- Alternativels, improve the error message to instruct users how to update manually or allow users to specify their own github auth token (But I'd strongly vote for the first option instead of this)

### Your environment

* **OS:** all OSes (I'm using macos)

`garden version` 0.13.9

mkhq:

Update to use a new endpoint to check for releases.

Fixes #4893 


vvagaytsev:
Would this PR fix #4893?

TimBeyer (dismissed):

core/test/unit/src/commands/self-update.ts
```diff
@@ -124,102 +125,166 @@ describe("SelfUpdateCommand", () => {
...
     expect(result?.installedVersion).to.be.undefined
     expect(result?.abortReason).to.equal("Not running from binary installation")
+
+    expect(scope.isDone()).to.be.true
   })
 
   it(`aborts cleanly if desired version isn't found`, async () => {
+    const scope = nock("https://get.garden.io")
+    scope.get("/releases/latest").reply(200, { tag_name: "0.13.0" })
+
```
twelvemo, line 259: Why would 0.13.0 not exist?

twelvemo (commented):
Good stuff! Just had one question about a test.

core/test/unit/src/commands/self-update.ts
mkhq: The test is for checking that `foo` cannot be downloaded. It resolves to a local server with the files in the `data/self-update` directory.

twelvemo (approved):

PR #5235, opened 2023-10-13
fix(cli): exit code 1 on unknown commands, sub-commands and flags

### LINKED ISSUE #4911: [Bug] Garden exists with `0` exit code on unsupported options and sub-commands
### Current Behavior

**Case 1.**
When `garden` is executed with an unrecognized _option_ and without any command (e.g. `garden --bad`) it prints the usage help and exists with code `0`.

**Case 2.**
When `garden` is executed with an unrecognized _sub-command_ (e.g. `garden run bad`) it prints the usage help and exists with code `0`.

### Expected behavior

In both cases above `garden` should exit with code `1` if an unrecognized option or sub-command is passed.

### Reproducible example

Run these commands:

```console
# Unrecognized option:
> garden --bad
# usage output omitted
# ...
> echo $?
0   # <- this is wrong

# Unrecognized sub-command:
> garden delete bad
# usage output omitted
# ...
> echo $?
0   # <- this is wrong
```

### Workaround

No

### Suggested solution(s)

Implement checks for `garden`'s options inspection and exit with code `1` on unrecognized options.

**Note!** The old behaviour for other testing scenarios is correct and must be kept:

```console
# Unrecognized (invalid) command:
> garden bad
# usage output omitted
# ...
> echo $?
1

# Unrecognized (invalid) _command_ option:
> garden deploy --bad
# usage output omitted
# ...
> echo $?
1

# Recognized (valid) option:
> garden --help
# usage output omitted
# ...
> echo $?
0
```

**Make sure we have the tests for the scenarios above. Please implement the missing test scenarios if any.**

### Additional context

This comes from the discussion started in #4868.

### Your environment

* **OS:** all
* **How I'm running Kubernetes:** does not matter

`garden version 0.13.9`
`garden version 0.12.63`

vvagaytsev:
This PR brings strict consistency into Garden command exit codes and fixes the possible CI problems like the ones we got in #4868.

Fixes #4911

This PR has been done in TDD mode :)
The test specs were written first.

~To be back-ported to `0.12`.~

TODOs (to be addressed in a separate PR):
* ensure the error message is always printed on unrecognized flag
* ensure the error message is always printed on unrecognized command or sub-command

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -12,7 +12,7 @@ import chalk from "chalk"
 import { pathExists } from "fs-extra"
 import { getBuiltinCommands } from "../commands/commands"
 import { shutdown, getPackageVersion, getCloudDistributionName } from "../util/util"
-import { Command, CommandResult, CommandGroup, BuiltinArgs } from "../commands/base"
+import { Command, CommandResult, BuiltinArgs, CommandGroup } from "../commands/base"
 import { PluginError, toGardenError, GardenError } from "../exceptions"
 import { Garden, GardenOpts, makeDummyGarden } from "../garden"
 import { getRootLogger, getTerminalWriterType, LogLevel, parseLogLevel, RootLogger } from "../logger/logger"
@@ -51,6 +51,7 @@ import { uuidv4 } from "../util/random"
 import { withSessionContext } from "../util/open-telemetry/context"
 import { wrapActiveSpan } from "../util/open-telemetry/spans"
 import { JsonFileWriter } from "../logger/writers/json-file-writer"
+import minimist from "minimist"
 
 export interface RunOutput {
   argv: any
@@ -65,6 +66,10 @@ export interface GardenCliParams {
   plugins?: GardenPluginReference[]
   initLogger?: boolean
   cloudApiFactory?: CloudApiFactory
+}
+
+function hasHelpFlag(argv: minimist.ParsedArgs) {
+  return argv.h || argv.help
 }
 
 // TODO: this is used in more contexts now, should rename to GardenCommandRunner or something like that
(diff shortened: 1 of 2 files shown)


core/src/cli/cli.ts
```diff
@@ -493,21 +498,54 @@ ${renderCommands(commands)}
...
+
       return done(0, command.renderHelp())
     }
 
+    // Handle incomplete subcommand listings.
+    // A complete sub-command won't be recognized as CommandGroup
+    if (command instanceof CommandGroup) {
+      const subCommandName = rest[0]
+      if (subCommandName === undefined) {
+        // exit code 0 if sub-command is not specific and --help flag is passed, e.g. garden get --help
```
stefreak, line 532: this comment seems wrong/ out of date, as the line below it exits with code 1

stefreak (dismissed):
Great work! I have one minor comment on a comment. 

I like the tests a lot ❤️

stefreak:
One comment on the plan to backport this to 0.12: Do we have to do that? I think it might not be important enough to justify the effort (But I might be wrong, you might have good reasons).

core/src/cli/cli.ts
vvagaytsev: Good catch, thanks! I fixed that in e474f896582f3b337c318fff1aa2515a9d193e51,

vvagaytsev:
> One comment on the plan to backport this to 0.12: Do we have to do that? I think it might not be important enough to justify the effort (But I might be wrong, you might have good reasons).

It's not super critical, but it might be better to align the exit code behavior between 0.12 and 0.13. The back-port should be a big effort. We're going to do the next `0.12` release soon, so we can include this fix too.

stefreak (approved):

PR #5236, opened 2023-10-15
0.12 backport: improvement: using new release endpoint for self-update

### LINKED ISSUE #4893: [Bug]: github API rate-limiting can occur in `garden self-update`

mkhq:

This is a 0.12 backport of #5229. Tests and code are ported over. ARM support was excluded from the backport since its only available from 0.13.12.

Fixes #4893 

### DIFF
--- a/core/src/constants.ts
+++ b/core/src/constants.ts
@@ -47,8 +47,7 @@ export const SUPPORTED_ARCHITECTURES: SupportedArchitecture[] = ["x64", "arm64"]
 export const SEGMENT_DEV_API_KEY = "D3DUZ3lBSDO3krnuIO7eYDdtlDAjooKW" // ggignore
 export const SEGMENT_PROD_API_KEY = "b6ovUD9A0YjQqT3ZWetWUbuZ9OmGxKMa" // ggignore
 
-export const DOCS_BASE_URL = "https://docs.garden.io/v/acorn-0.12"
-export const VERSION_CHECK_URL = "https://get.garden.io/version"
+export const DOCS_BASE_URL = "https://docs.garden.io"
 
 export const DEFAULT_GARDEN_CLOUD_DOMAIN = "https://app.garden.io"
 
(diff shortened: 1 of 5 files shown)


vvagaytsev:
@mkhq thanks for back-porting this! The docs need to be re-generated. The rest test failures and known and expected.

core/src/constants.ts
```diff
@@ -47,8 +47,7 @@ export const SUPPORTED_ARCHITECTURES: SupportedArchitecture[] = ["x64", "arm64"]
 export const SEGMENT_DEV_API_KEY = "D3DUZ3lBSDO3krnuIO7eYDdtlDAjooKW" // ggignore
 export const SEGMENT_PROD_API_KEY = "b6ovUD9A0YjQqT3ZWetWUbuZ9OmGxKMa" // ggignore
 
-export const DOCS_BASE_URL = "https://docs.garden.io/v/acorn-0.12"
-export const VERSION_CHECK_URL = "https://get.garden.io/version"
+export const DOCS_BASE_URL = "https://docs.garden.io"
```
vvagaytsev, line 50: This must be `DOCS_BASE_URL = "https://docs.garden.io/v/acorn-0.12"` in the `0.12` branch.

vvagaytsev (approved):
Thank you! 👍

PR #5271, opened 2023-10-19
test: make testFlags.expandErrors work with mocha error renderer

stefreak:
Changes GardenError to expand the stack in the initializer already, so
that errors are always rendered expanded in mocha tests.

This greatly eases test development.

Fixes #

### DIFF
--- a/core/src/exceptions.ts
+++ b/core/src/exceptions.ts
@@ -92,7 +92,6 @@ export abstract class GardenError extends Error {
 
   constructor({ message, wrappedErrors, taskType, code }: GardenErrorParams) {
     super(message.trim())
-    this.stack = this.stack
     this.wrappedErrors = wrappedErrors
     this.taskType = taskType
     this.code = code
--- a/core/src/exceptions.ts
+++ b/core/src/exceptions.ts
@@ -57,7 +57,6 @@ export type GardenErrorStackTrace = {
 
 export interface GardenErrorParams {
   message: string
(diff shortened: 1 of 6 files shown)


core/src/exceptions.ts
```diff
@@ -86,25 +85,41 @@ export abstract class GardenError extends Error {
...
-  constructor({ message, stack, wrappedErrors, taskType, code }: GardenErrorParams) {
+  /**
+   * Necessary to make testFlags.expandErrors work.
+   */
+  private unexpandedStack: string | undefined
+
+  constructor({ message, wrappedErrors, taskType, code }: GardenErrorParams) {
     super(message.trim())
-    this.stack = stack || this.stack
+    this.stack = this.stack
```
vvagaytsev, line 95: Will the self-assignment `this.stack = this.stack` always leave the `this.stack` undefined?

vvagaytsev: If so, then we can remove this assignment :)

stefreak: No, it does not because stack gets set automagically, but the line of code is definitely unnecessary!

```suggestion
```

stefreak: Good catch @vvagaytsev !

shumailxyz (approved):
This is so great! Thanks @stefreak ✨

vvagaytsev (approved):
💯

PR #5272, opened 2023-10-19
fix(k8s): correct `0.12 => 0.13` service resource conversion

### LINKED ISSUE #5269: 0.13: [Bug]: Module task with `podSelector` does not work with 0.13
### Current Behavior

If a `task` in a `module` specifies `podSelector`, it doesn't work in bonsai with the following error message:

```
Error validating spec for Run type=kubernetes-pod name=task-foo (from module test-tasks):
"podSelector" can't be specified with "kind"
object at spec.resource can only contain one of [kind, podSelector]
```

### Expected behavior

Task should work fine.

### Reproducible example

Run `garden run task-foo` with the following project and module:

project.garden.yml
```
apiVersion: garden.io/v1
kind: Project
name: repro-task-to-run
environments:
  - name: local
providers:
  - name: local-kubernetes
    environments: [local]

```

garden.yml
```
kind: Module
name: test-tasks
description: task bug in bonsai
type: kubernetes
allowPublish: false
include: []
timeout: 120
manifests:
  - apiVersion: v1
    kind: Pod
    metadata:
      name: test-app
      labels:
        app: test-app
    spec:
      containers:
        - name: test-app
          image: nginx
      restartPolicy: Never
tasks:
  - name: task-foo
    description: This will take a backup of assets + database
    timeout: 1200
    resource:
      podSelector:
        app: test-app
    dependencies:
      - test-tasks
    command:
      - '/bin/bash'
      - '-c'
      - 'echo hello'

```

### Workaround

Don't know of any.

### Additional context

Could be an issue with conversion of tasks to run.

### Your environment

* **OS:** mac

`garden version`
0.13.17

vvagaytsev:
This PR ensures the the module conversion respects the 0.13 schema validation rules, which are now stricter than in 0.12.

Fixes #5269

A new function `convertServiceResourceSpec` was extracted to take more specific input parameters and to do the spec-to-spec conversion. This also allowed making the unit test data preparation way easier.

### DIFF
--- a/core/src/plugins/kubernetes/kubernetes-type/common.ts
+++ b/core/src/plugins/kubernetes/kubernetes-type/common.ts
@@ -8,7 +8,7 @@
 
 import { basename, dirname, join, resolve } from "path"
 import { pathExists, readFile } from "fs-extra"
-import { flatten, keyBy, set } from "lodash"
+import { flatten, isEmpty, keyBy, set } from "lodash"
 
 import { KubernetesModule } from "./module-config"
 import { KubernetesResource } from "../types"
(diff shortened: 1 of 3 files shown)


twelvemo (dismissed):
Looks really good to me!

shumailxyz (changes requested):
Thanks @vvagaytsev Fix looks great, but the tests are failing. Please check that.

twelvemo (commented):
One test failures around expecting `containerName`.

twelvemo (approved):
Tests looking good now too! Thanks so much for the fast fix!

Issue #5275, opened 2023-10-19
0.13: [Bug]: Garden does not error when running in an environment that doesn't exist.


eysi09:
### Current Behavior

Garden does not error when running in an environment that doesn't exist. 

### Expected behavior

I'd expect it to fail with the message that the environment doesn't exist since otherwise it ca cause cascading downstream issues that are hard to debug.

### Reproducible example
1. Run `garden login` (This only happens when logged in to Garden Cloud)
2. Run `garden validate --env does-not-exist` in any example project.

`garden version` Latest main as of writing.

vvagaytsev:
@eysi09 which example did you try to reproduce this?

I've tried the [demo-project](https://github.com/garden-io/garden/tree/main/examples/demo-project) and got the error in 0.12.66:
```console
Project demo-project does not specify environment does-not-exist
```
and in 0.13.18:
```console
Project demo-project does not specify environment does-not-exist (Available environments: local and remote)
```

eysi09:
Aha ok. This appears to only happen if you're logged in. 

So apparently we call the API before validating the environment. And that call will use a bad environment and fail but with a confusing error. 

So I suppose we need to hoist the validation to before we at least fetch secrets. I _think_ prior API calls should work. But you can verify if you test this while logged in.

vvagaytsev:
Hmm... I've tried to log in and run the validation, and it failed because of an unknown env again. Tried with demo-project, vote, and cloud repo.

PR #5277, opened 2023-10-19
fix(api): visible and better error when fetching secrets 404s

eysi09:

Before this fix, 404 errors that happened when fetching secrets were logged at a debug level and not visible to users.

This fixes that and improves the error message. 

 Related to #5275 because fixing that will reduce the chance of this error surfacing in the first place. 

Fixes #


vvagaytsev (approved):
LGTM, thanks! 👍

PR #5278, opened 2023-10-20
Store action type in cloud

eysi09:

There's a bit of history here. 

Garden has a concept of action _kinds_ and action _types_, both of which we want to store in display in the dashboard UI. 

Due to some legacy issues, were confusing the two and the API didn't support payload with the action type set correctly. 

This has since been fixed in the API but be never updated the Core payload until. 

However, there are older instances of the API still running that haven't been updated to support this payload yet, so we feature flag it for now. We should be able to move the feature flag very soon. 

Fixes #

We should squash the commits before merging.

### DIFF
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -707,8 +707,6 @@ function dependenciesFromActionConfig(
           message: `${description} references depdendency ${depKey}, but no such action could be found`,
         })
       }
-      // FIXME @eysi: We're setting the "parent" config type as the dep type which is not correct
-      // and we need to the dep type.
       return {
         kind,
         name,
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -700,7 +700,9 @@ function dependenciesFromActionConfig(
   const deps: ActionDependency[] = config.dependencies.map((d) => {
     try {
       const { kind, name } = parseActionReference(d)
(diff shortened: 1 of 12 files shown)


core/src/graph/actions.ts
```diff
@@ -700,15 +700,17 @@ function dependenciesFromActionConfig(
   const deps: ActionDependency[] = config.dependencies.map((d) => {
     try {
       const { kind, name } = parseActionReference(d)
-      return { kind, name, explicit: true, needsExecutedOutputs: false, needsStaticOutputs: false }
+      // FIXME @eysi: We're setting the "parent" config type as the dep type which is not correct
```
eysi09, line 703: @thsig, thoughts on this one?

In short, I'm trying to send the action type to cloud. Both via action status events (where it's readily available) and in the stack graph event where it is not. 

I kind of just worked my way backwards through type errors to be able to add the type to the payload but I'm kind of stuck here.

The type isn't available in this context unless we add it to the `ActionReference` type which causes a lot of type errors and a bunch of code needs to be updated. Besides, it doesn't sound like it belongs there.

So is there a smart to find the type of the dependency in this context?

thsig: The `render` method on `BaseConfigGraph` could be modified to include the type for each node (the underlying action can be found via a constant-time lookup using the action name + kind).

thsig (approved):

eysi09:
Closing in favour of #5447

PR #5291, opened 2023-10-23
fix(cloud): do not fetch secrets on community tier

eysi09:

Before this fix, we'd log a warning about secrets not existing for users on the community tier when in fact we shouldn't be attempting to fetch them in the first place.

Fixes #


shumailxyz (approved):
Thanks for the fix ✨

PR #5292, opened 2023-10-23
fix(logger): log exec build/deploy actions at info level

eysi09:

Previously these actions were logged at the verbose level which doesn't really make sense imo.

In particular since other exec actions are on the info level so to the end user it'll be confusing that some exec results get logged and others not.

Also updated the various places where we render a divider to make sure we also include a browser friendly version.

Fixes #


thsig:
Good stuff!

IIRC, I did something similar verbosity-wise in the Docker Compose plugin.

I think it totally makes sense for these levels to be plugin-appropriate, and to respect the user's expectations for the underlying tools.

thsig (approved):

PR #5297, opened 2023-10-24
Circle ci main tests

twelvemo:
Don't run all the jobs again when a commit is merged to main, since that already happened in the merge queue.
We only need to run the jobs for the edge release on main.
Since almost all jobs require the `build` job, it should be sufficient to apply the filter that ignores the main branch to it.
To enable the edge release, i added a build job specifically for edge.

Fixes #

### DIFF
--- a/.circleci/continue-config.yml
+++ b/.circleci/continue-config.yml
@@ -56,6 +56,11 @@ parameters:
       tags:
         ignore: /.*/
 
+  ignore-main: &ignore-main
+    filters:
+      branches:
+        ignore: main
+
   # Only run jobs on main branch
   only-main: &only-main
     filters:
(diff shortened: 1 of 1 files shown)


.circleci/continue-config.yml
```diff
@@ -981,7 +986,8 @@ workflows:
     jobs:
       ### ALL BRANCHES ###
 
-      - build
+      - build:
+          <<: *ignore-main
```
stefreak, line 990: I think it is slightly surprising that this makes all the tests not run on main. I'm ok with solving it like this, if it works, but maybe we need a comment to explain what's going on.

stefreak, line 990: I think it is slightly surprising that this makes all the tests not run on main. I'm ok with solving it like this, if it works, but maybe we need a comment to explain what's going on.

```suggestion
          # We only want to run tests on branches (e.g. for the merge queue) and pull requests.
          # On main we do not need to re-run all the tests, as the merge queue guarantees that tests already passed.
          # For the edge release we declare the `build-edge` job below that only runs on main. 
          <<: *ignore-main
```

stefreak (commented):
I added a suggestion with a comment for increased clarity.

But otherwise this looks good to me 👍

.circleci/continue-config.yml
twelvemo: Makes sense and added your comment. I think this should make all the tests not run in main, since they all depend on `build` but hard to say if it works before merging.

stefreak (approved):

PR #5302, opened 2023-10-25
fix(k8s): return deployed mode in container Deploy status

thsig:

Before this fix, we were returning the action mode (e.g. `default`, `sync` or `local`) set on the locally generated manifests under `status.detail.mode` for `container` Deploys.

This meant that e.g. the `get status` command would always show the mode as `default`. The fix was to use the action mode set on the deployed resources, like we've been doing for `kubernetes` and `helm` Deploys.

Also added an integ test case for this behavior.

The usefulness of this fix is that it should allow the Cloud UI to rely on the action modes in deploy statuses to show e.g. the mode of Deploys in the graph view.


vvagaytsev (approved):
Thank you! 👍

PR #5307, opened 2023-10-27
fix(helm): handle missing (null) values in version check

### LINKED ISSUE #5303: Crash: Cannot read properties of null (reading '.garden')
Hey guys,

Just to say, I was running deploy and then I had this issue at this point:
ℹ deploy.default-backend → Already deployed
ℹ deploy.ingress-controller → missing
ℹ deploy.ingress-controller → Deploying version v-8286b84dfa...
✖ deploy.ingress-controller → Failed (took 0.1 sec)
✖ deploy.ingress-controller → Failed processing Deploy type=helm name=ingress-controller (took 0.22 sec). This is what happened:

You can help by reporting this on GitHub: https://github.com/garden-io/garden/issues/new?labels=bug,crash&template=CRASH.md&title=Crash%3A%20Cannot%20read%20properties%20of%20null%20(reading%20%27.garden%27)

Failed processing Deploy type=helm name=ingress-controller (took 0.22 sec). This is what happened:
TypeError: Cannot read properties of null (reading '.garden')
    at getReleaseStatus (/snapshot/pkg/core/build/src/plugins/kubernetes/helm/status.js:155:49)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Object.helmDeploy (/snapshot/pkg/core/build/src/plugins/kubernetes/helm/deployment.js:46:27)
    at async /snapshot/pkg/core/build/src/router/base.js:202:28
    at async Router.callHandler (/snapshot/pkg/core/build/src/router/base.js:173:24)
    at async deploy (/snapshot/pkg/core/build/src/router/deploy.js:40:24)
    at async DeployTask.process (/snapshot/pkg/core/build/src/tasks/deploy.js:89:28)
    at async DeployTask.<anonymous> (/snapshot/pkg/core/build/src/tasks/base.js:319:29)
    at async /snapshot/pkg/core/build/src/util/open-telemetry/decorators.js:47:30
    at async ProcessTaskNode.execute (/snapshot/pkg/core/build/src/graph/nodes.js:214:35)
    at async GraphSolver.processNode (/snapshot/pkg/core/build/src/graph/solver.js:281:35)
    
I had installed again garden-nginx because the deployment of ingress controller was encountering issues and I used this command:
    
    helm install garden-nginx ingress-nginx/ingress-nginx -n garden-system --version 4.0.13

I use Ubuntu 22.04, Minikube and garden version: 0.13.18. The issue happened when I run the deployment and it seemed related to Helms release "garden-nginx" in the namespace "garden-system".  Kubernetes and Minikube statuses are ok. The issues seems resolved when I installed the update helm for metricbeat, filebeat and mongdb. I used garden validate and garden logs for help. 

Thank you!

vvagaytsev:

Fixes #5303

### DIFF
--- a/core/src/plugins/kubernetes/helm/status.ts
+++ b/core/src/plugins/kubernetes/helm/status.ts
@@ -205,19 +205,25 @@ export async function getReleaseStatus({
 
     if (state === "ready") {
       // Make sure the right version is deployed
-      values = JSON.parse(
-        await helm({
-          ctx,
-          log,
-          namespace,
-          args: ["get", "values", releaseName, "--output", "json"],
-          // do not send JSON output to Garden Cloud or CLI verbose log
-          emitLogEvents: false,
-        })
-      )
+      const helmResponse = await helm({
+        ctx,
+        log,
+        namespace,
+        args: ["get", "values", releaseName, "--output", "json"],
+        // do not send JSON output to Garden Cloud or CLI verbose log
+        emitLogEvents: false,
+      })
+      values = JSON.parse(helmResponse)
 
-      const deployedVersion = values[".garden"]?.version
-      deployedMode = values[".garden"]?.mode
+      let deployedVersion: string | undefined = undefined
+      // JSON.parse can return null
+      if (values !== null) {
+        log.verbose(
+          `No helm values returned for release ${releaseName}. It looks like the relevant deployments were removed manually.`
+        )
+        deployedVersion = values[".garden"]?.version
+        deployedMode = values[".garden"]?.mode
+      }
 
       if (action.mode() !== deployedMode || !deployedVersion || deployedVersion !== action.versionString()) {
         state = "outdated"
--- a/core/src/plugins/kubernetes/helm/status.ts
+++ b/core/src/plugins/kubernetes/helm/status.ts
@@ -218,9 +218,7 @@ export async function getReleaseStatus({
       let deployedVersion: string | undefined = undefined
       // JSON.parse can return null
       if (values !== null) {
-        log.verbose(
-          `No helm values returned for release ${releaseName}. It looks like the relevant deployments were removed manually.`
-        )
+        log.verbose(`No helm values returned for release ${releaseName}. Is this release managed outside of garden?`)
         deployedVersion = values[".garden"]?.version
         deployedMode = values[".garden"]?.mode
       }
--- a/core/src/plugins/kubernetes/helm/status.ts
+++ b/core/src/plugins/kubernetes/helm/status.ts
@@ -218,9 +218,9 @@ export async function getReleaseStatus({
       let deployedVersion: string | undefined = undefined
       // JSON.parse can return null
       if (values === null) {
+        log.verbose(`No helm values returned for release ${releaseName}. Is this release managed outside of garden?`)
         state = "outdated"
       } else {
-        log.verbose(`No helm values returned for release ${releaseName}. Is this release managed outside of garden?`)
         deployedVersion = values[".garden"]?.version
         deployedMode = values[".garden"]?.mode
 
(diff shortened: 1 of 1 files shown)


vvagaytsev:
The ingress controller-specific issue reported in the target ticket will be resolved in #5136. However, the `getReleaseStatus` function should still be able to handle possible `null` values returned from the `helm` command.

core/src/plugins/kubernetes/helm/status.ts
```diff
@@ -205,19 +205,25 @@ export async function getReleaseStatus({
...
+        // do not send JSON output to Garden Cloud or CLI verbose log
+        emitLogEvents: false,
+      })
+      values = JSON.parse(helmResponse)
+
+      let deployedVersion: string | undefined = undefined
+      // JSON.parse can return null
+      if (values !== null) {
+        log.verbose(
+          `No helm values returned for release ${releaseName}. It looks like the relevant deployments were removed manually.`
```
twelvemo, line 222:
```suggestion
          `No helm values returned for release ${releaseName}. Is this release managed outside of garden?`
```
If a helm chart has been deployed by garden it will always at least include the `.garden` values even if no other custom values are supplied by the user.
If the release status is ready and we don't get any values than a helm release must have been installed outside of garden, as is the case in the issue linked to this PR.
I think handling this case is a good idea, but would suggest that we change the error message a bit, since the deployments might be present and healthy.

twelvemo (changes requested):
Nice! Had one suggesting around the error message.

core/src/plugins/kubernetes/helm/status.ts
```diff
@@ -205,19 +205,25 @@ export async function getReleaseStatus({
...
+        namespace,
+        args: ["get", "values", releaseName, "--output", "json"],
+        // do not send JSON output to Garden Cloud or CLI verbose log
+        emitLogEvents: false,
+      })
+      values = JSON.parse(helmResponse)
+
+      let deployedVersion: string | undefined = undefined
+      // JSON.parse can return null
+      if (values !== null) {
```
twelvemo, line 220:
```suggestion
      if (values === null) {
```
Don't we want to check if the values are null and return `state: "outdated"` if that's the case? Trying to get the `deployedVersion` from the values will still fail.

twelvemo (changes requested):
Took another look and had another question.

core/src/plugins/kubernetes/helm/status.ts
vvagaytsev: I thought about that too. Now, if the `values` is `null`, it will be handled in [L226](https://github.com/garden-io/garden/pull/5307/files#diff-9e897b08cb49cc822f3496aa64e103cad793f9414b6f34960bd21416acf66785R226).

Of course, we can explicitly set the status to `outdated` if the `values` is `null`. Just for better clarity.

vvagaytsev: Changed this in 6ac81b4396f73d6fb8a3c9c82b3576585f658e64.

core/src/plugins/kubernetes/helm/status.ts
```diff
@@ -205,22 +205,28 @@ export async function getReleaseStatus({
...
+        emitLogEvents: false,
+      })
+      values = JSON.parse(helmResponse)
+
+      let deployedVersion: string | undefined = undefined
+      // JSON.parse can return null
+      if (values === null) {
         state = "outdated"
+      } else {
+        log.verbose(`No helm values returned for release ${releaseName}. Is this release managed outside of garden?`)
```
twelvemo, line 223: Shouldn't we log this when the values are null?

vvagaytsev: Ah, right :)

vvagaytsev: Fixed this in 8605647f43cb325638d7900ef9b357f15abec5d0.

twelvemo (approved):
🚀

PR #5325, opened 2023-10-31
fix(pulumi): fix process dependency resolution in plugin command

vvagaytsev:
Method `resolveDependencies` was unused before this fix. However, it was implemented to handle the process dependencies. This commit renames it to override the corresponding method of the parent class, as it was initially intended.

Fixes #


shumailxyz (commented):
Thanks, Good catch ✨

shumailxyz (approved):

PR #5329, opened 2023-11-01
fix(dev): fix reload error when using templates

### LINKED ISSUE #5290: 0.13: [Bug]: Using config templates crashes `garden dev`
### Current Behavior

I have a project which uses config templates. Every now and then the dev console (`garden dev`) will crash with the error:

```Cannot read properties of undefined (reading 'inputsSchema')```

At a glance I think this only happens after the Garden config has been modified (and Garden prints the `Config change detected. Project will be reloaded when the next command is run.`) 

This usually happens when I run commands from within the console or on events triggered by the dashboard if that's open (some of these events trigger internal Garden command where the config resolution fails). 

### Expected behavior

It doesn't crash.

### Reproducible example

I haven't been able to pinpoint exactly when this happens so kind of hoping someone can spot the issue without a solid repro. But it does strike me as some config resolution race condition since it doesn't fail consistently. 

Your best bet to repro would be to:

1. Run `garden dev` in e.g. [this example which uses config templates](https://github.com/garden-io/garden/tree/main/examples/k8s-deploy-config-templates).
2. Open the web dashboard on the live page (if only to have it trigger internal Garden commands which may surface this)
3. Tweak the actual Garden config (e.g. re-naming an action or adding a new one)
4. Wait for the dashboard to request a new graph via the `loadConfig` ws message which is called when the config changes, or run some commands from within the console (e.g. `validate`)
5. Observe as it crashes. Most likely. Sometimes.

### Workaround

No known workaround, it basically makes the dev command un-usable for projects that use config templates.

### Your environment

`garden version` Latest main as of writing.

### PR DESCRIPTION
**Pull request by thsig**

Before this fix, changing a Garden config during a `dev` command session and running two subsequent commands would result in the first command succeeding but the second failing with an error like:

"Build foo references template <some-template> which cannot be found. Vailable templates: <None>"

This was because the config templates scanned by the first subcommand after the config was changed weren't propagated to the parent Garden instance, which was then cloned and passed to the second command (which then had action configs, but no templates).

This was fixed by making the relevant instance variables on the Garden class readonly (since reassignments in a child instance prevent newly scanned configs from propagating to the parent via the shared data structure) and emptying the maps before rescanning instead.

Also added a unit test case that checks this whole flow.

Fixes #5290

This fix is rather hacky. We really shouldn't be relying on shared references to data structures in the instance variables of parent & child Garden instances. This happens to mostly work, but could easily lead to hard-to-diagnose problems in the future.

I thought I'd open the PR to start a discussion on how to proceed properly here.

Would it be better if we stored the action configs and config templates in an instance of a singleton class, to make sure there's ever only one version of the truth (the most up-to-date)?

Or do we explicitly want different child Garden instances to have their own copy of the configs & templates?

I think the choice boils down to either of these two:
1. A single `ConfigRegistry` that's shared by all Garden instances. This would be populated/updated whenever configs are scanned, and there would always be exactly one instance of it in existence at any given time.
2. Deep-cloning the action configs & config templates of the parent Garden when creating a child Garden, and carefully propagating any new configs & templates that are scanned back to the parent.

The former sounds like a better approach to me, but I'm all ears.

CC @eysi09 @edvald

### CODE REVIEW

**Comment by thsig** [comment]
Note: I never did quite repro the error Eythor originally reported, but it looks like it stems from the same cause (the templates not existing on the Garden instance in question).

**Comment by eysi09** [comment]
My first reaction would be to go for option 1 here unless there are use cases for 2 that I'm missing. 

Isn't that something that could live on the `InstanceManager` class?

I'm wondering if we should still merge the hack though since this frequently crashes running Garden processes.

**Comment by thsig** [comment]
I'm going to do a bit of a refactor around the state management for configs in the instance manager in a follow up PR which I've started working on, will include tests around this there.

Merging this fix will get rid of these errors in the meantime.

File: core/src/commands/base.ts [hunk 1]
```diff
@@ -287,7 +287,6 @@ export abstract class Command<
     return withSessionContext({ sessionId, parentSessionId }, () =>
       wrapActiveSpan(this.getFullName(), async () => {
         const commandStartTime = new Date()
-        const server = this.server
```
eysi09, line 290: We really need a lint rule for "no-unused".

core/src/garden.ts
```diff
@@ -1248,6 +1250,23 @@ export class Garden {
...
+  /**
+   * We clear these objects without reassigning them. If this Garden instance was cloned from a parent Garden instance,
+   * this will also clear the configs of the parent instance (which is what we want e.g. when rescanning configs during
+   * a subcommand after a reload is necessary during a dev command session).
+   *
+   * We need to clear before rescanning to make sure old/outdated configs are cleared away, and to avoid duplicate
+   * key errors when adding the newly scanned ones (and those generated from newly scanned config templates).
+   */
+  private clearConfigs() {
+    for (const kind of Object.getOwnPropertyNames(this.actionConfigs)) {
```
eysi09, line 1262: Pardon me ignorance but why use `getOwnPropertyNames` instead of `keys`?

thsig:
With the unit test to cover the behaviour, I think this is OK for now.

I didn't feel that a larger surgery on the state management around the instance manager was worth the effort at this point (also, the config scanning piece may be the main piece of shared state we'll have to deal with in this context).

So I hope this is a practical, simple solution to the problem.

core/src/util/util.ts
```diff
@@ -512,6 +512,17 @@ export function getEnvVarName(identifier: string) {
   return identifier.replace(/-/g, "_").toUpperCase()
 }
 
+/**
+ * Removes all keys from `obj` (except those inherited from the object's prototype).
+ *
+ * Essentially a vanilla object analog of `map.clear()` for ES5 Maps.
+ */
+export function clearObject<T extends object>(obj: T) {
+  for (const key of Object.getOwnPropertyNames(obj)) {
```
eysi09, line 521: Same question as above.

eysi09 (dismissed):
Had a couple of non-blocking questions but otherwise this looks good to me. So I'll go ahead and approve.

eysi09 (approved):

eysi09:
Approving but we need to resolve conflicts.

eysi09:
Bumping this one @thsig :) Could you look into resolving the conflicts so we can get this merged?

thsig:
Ah yes, will do!

eysi09 (approved):

PR #5334, opened 2023-11-01
fix(k8s): ensure patchResources can patch namespace

eysi09:

When applying K8s resources, we create a ConfigMap with metadata about the resources for faster lookup.

This metadata includes the resource namespace among other things.

Before this fix, we'd create the ConfigMap before applying the resource patch which meant that if you patched the actual resource namespace, Garden wouldn't store that information.

Later when looking up the status of the resource, Garden would use the data from the ConfigMap and basically look for the resource in the wrong namespace and return status "not-ready" (when in fact the resource could be ready, just in a different namespace).

Fixes #

- Are there other places in the code you can think of were we could be using the wrong namespace? (At a glance I wouldn't think so, but wanted to flag.)
- Just out of curiosity: Why do we get the deploy status again after waiting for the resources? If you trace through the logs, you see that Garden creates the resources in the correct namespace and logs "Resources ready". We then go ahead and get the statuses again in [this line](https://github.com/garden-io/garden/blob/main/core/src/plugins/kubernetes/kubernetes-type/handlers.ts#L426) and return those. (This is related to the bug because for the second time around we'd look in the wrong namespace). Shouldn't the `waitForResources` function return the correct status?

### DIFF
--- a/core/src/plugins/kubernetes/kubernetes-type/common.ts
+++ b/core/src/plugins/kubernetes/kubernetes-type/common.ts
@@ -150,19 +150,7 @@ export async function getManifests({
   }
 
   const declaredManifests = await readManifests(ctx, action, log)
-
-  if (action.kind === "Deploy") {
-    // Add metadata ConfigMap to aid quick status check
-    const metadataManifest = getMetadataManifest(action, defaultNamespace, declaredManifests)
-    const declaredMetadataManifest: DeclaredManifest = {
-      declaration: { type: "inline", index: declaredManifests.length },
-      manifest: metadataManifest,
-    }
-    declaredManifests.push(declaredMetadataManifest)
-  }
-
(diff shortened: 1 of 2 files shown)


core/src/plugins/kubernetes/kubernetes-type/common.ts
```diff
@@ -177,6 +165,16 @@ export async function getManifests({
     )
   }
 
+  if (action.kind === "Deploy") {
+    // Add metadata ConfigMap to aid quick status check
+    const metadataManifest = getMetadataManifest(action, defaultNamespace, patchedManifests)
```
eysi09, line 170: The fix here is that now we're getting the metadata for the patched manifests.

twelvemo (approved):
Looks good! Nice that you added a test for that!

PR #5349, opened 2023-11-06
chore(circleci): increase executor size for dockerhub and windows

stefreak:
With the goal to reduce the time we have to wait for these jobs.

The dockerhub job only runs after merged to main, or when the
Dockerfiles actually changed.

Fixes #

### DIFF
--- a/.circleci/continue-config.yml
+++ b/.circleci/continue-config.yml
@@ -482,6 +482,7 @@ jobs:
   test-dockerhub:
     machine:
       <<: *shared-machine-config
+    resource_class: large
     steps:
       # Only run test-dockerhub if path-filter noticed changes in ./support/*
       - when:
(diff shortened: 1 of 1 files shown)


.circleci/continue-config.yml
```diff
@@ -482,6 +482,7 @@ jobs:
   test-dockerhub:
     machine:
       <<: *shared-machine-config
+    resource_class: large
```
TimBeyer, line 485: Can we give this one an `xlarge` since we're doing so much in parallel?

TimBeyer, line 485:
```suggestion
    resource_class: xlarge
```

TimBeyer (approved):

TimBeyer (approved):

PR #5356, opened 2023-11-07
chore: module cleanup

TimBeyer:
Removes some unused modules 

Fixes #


e2e/package.json
```diff
@@ -39,6 +39,7 @@
     "minimist": "^1.2.8",
     "mocha": "^10.2.0",
     "prettier": "3.0.0",
+    "replace-in-file": "^7.0.2",
```
stefreak, line 42: we don't use "replace-in-file" anymore in build-pkg– are we using it anywhere else still?

TimBeyer: The release script uses it and the e2e test too

PR #5357, opened 2023-11-07
chore: dependency updates

TimBeyer:
Bumps a whole lot of dependencies now that we're supporting ESM modules

Fixes #


core/src/server/server.ts
```diff
@@ -224,8 +222,7 @@ export class GardenServer extends EventEmitter {
         try {
           this.port = await getPort({
             host: hostname,
-            port: defaultServerPort,
-            portRange: [defaultServerPort + 1, defaultServerPort + 99],
+            port: portNumbers(defaultServerPort, defaultServerPort + 99),
```
stefreak, line 225: @vvagaytsev does this change look good to you?

core/src/util/artifacts.ts
```diff
@@ -77,7 +77,7 @@ export async function copyArtifacts({
   const { default: cpy } = await import("cpy")
 
   try {
-    files = await cpy("**/*", garden.artifactsPath, { cwd: artifactsPath, parents: true })
+    files = await cpy("./**/*", garden.artifactsPath, { cwd: artifactsPath })
```
stefreak, line 80: What did `parents: true` do?

stefreak (commented):
Looks good to me so far; There is still one test failure and I commented on the changes around `cpy` and `getPort`

core/src/util/artifacts.ts
TimBeyer: It's basically the default behavior now. It ensured that it preserved the directory tree while recursing.

stefreak (approved):
Awesome! Now we're up-to-date.

core/src/server/server.ts
vvagaytsev: Yes, that looks totally good! The usage of `get-port-please` was a temporary solution.

vvagaytsev:
This also supersedes the temporary fixes implemented in #4984 and #5163. Great job, thank you so much! 👍 💯 🚀

PR #5359, opened 2023-11-07
refactor(garden-sea): minor rust refactors

stefreak:

Fixes #

### DIFF
--- a/garden-sea/src/extract.rs
+++ b/garden-sea/src/extract.rs
@@ -7,7 +7,7 @@ use std::{fs, io::Write};
 
 use eyre::{Result, WrapErr};
 
-use crate::artifacts::{GardenArtifact, NATIVE_MODULES, NODE_BINARY, SOURCE, STATIC};
+use crate::artifacts::{GardenArtifact, ALL_ARTIFACTS};
 use crate::cleanup::start_cleanup_thread;
 use crate::log::debug;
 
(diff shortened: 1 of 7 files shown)


garden-sea/src/terminate.rs
```diff
@@ -11,7 +11,10 @@ use windows::Win32::System::Console::{
 
 #[cfg(unix)]
 pub fn interrupt(pid: u32) -> Result<(), nix::errno::Errno> {
-    signal::kill(Pid::from_raw(pid.try_into().unwrap()), Signal::SIGINT)?;
+    signal::kill(
+        Pid::from_raw(pid.try_into().expect("Pid could not be coerced into i32")),
```
stefreak, line 15: Maybe we want to just print a debug line, instead of crashing in this case.

For some reasons, the child process pid is u32 and the signal api expects i32.

garden-sea/src/signal.rs
```diff
@@ -23,8 +23,8 @@ enum Signal {
 #[cfg(windows)]
 lazy_static! {
     static ref _C: (Sender<Signal>, Receiver<Signal>) = bounded(100);
-    static ref SEND: Sender<Signal> = _C.0.clone();
-    static ref RECEIVE: Receiver<Signal> = _C.1.clone();
+    static ref SEND: Sender<Signal> = _C.0;
+    static ref RECEIVE: Receiver<Signal> = _C.1;
```
stefreak, line 27: This didn't work; undo
```suggestion
    static ref SEND: Sender<Signal> = _C.0.clone();
    static ref RECEIVE: Receiver<Signal> = _C.1.clone();
```

garden-sea/src/extract.rs
```diff
@@ -55,16 +55,19 @@ pub(crate) fn extract_archives_if_needed(root_path: &Path) -> Result<PathBuf> {
 }
 
 fn extracts_needed(path: &Path) -> Result<bool> {
-    Ok(is_extract_needed(path, NODE_BINARY)?
-        || is_extract_needed(path, NATIVE_MODULES)?
-        || is_extract_needed(path, STATIC)?
-        || is_extract_needed(path, SOURCE)?)
+    for artifact in ALL_ARTIFACTS {
+        if !is_extract_needed(path, artifact)? {
+            return Ok(false);
```
TimBeyer, line 60: Isn't that the wrong way around?
This looks like if for example the node version is still up to date, no other things will get updated.

stefreak: Oops, I believe you are right! Will fix that

PR #5360, opened 2023-11-07
improvement(core): log aborted nodes on dep error

thsig:

When a dependency fails, we now log a line for each aborted node.

Also fixed the counting logic used in the command error summary. Before this fix, we used to include aborted nodes in the failed count.

Note that we're no longer copying the error from the failed dependency to all the failed dependants in the command result. This might be seen as a breaking change.

If we'd rather keep the machine-readable command output the same, I can modify the logic to set the error for the aborted dependant keys in the command result.

### DIFF
--- a/core/src/commands/base.ts
+++ b/core/src/commands/base.ts
@@ -1025,10 +1025,11 @@ export async function handleProcessResults(
   const graphResults = results.results
   const graphResultsForExport = graphResults.export()
 
-  const failed = pickBy(graphResultsForExport, (r) => r && r.error)
+  const failed = pickBy(graphResultsForExport, (r) => r && !r.aborted && !!r.error)
   const failedCount = size(failed)
+  const abortedCount = size(pickBy(graphResultsForExport, (r) => r && !!r.aborted && !r.error))
 
-  const success = failedCount === 0
+  const success = failedCount === 0 && abortedCount === 0
 
   const buildResults = prepareProcessResults("build", graphResults) as ProcessCommandResult["build"]
   const deployResults = prepareProcessResults("deploy", graphResults) as ProcessCommandResult["deploy"]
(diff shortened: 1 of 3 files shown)


core/src/graph/nodes.ts
```diff
@@ -146,14 +146,32 @@ export abstract class TaskNode<T extends Task = Task> {
...
+        if (depKey === task.getKey() && d.executionType === "request" && error) {
+          depAborted = false
+          depError = new GraphNodeError({ resultError: error, node: d })
+        } else {
+          depAborted = true
+          depError = null
+          if (!keys.has(depKey)) {
+            d.task.log.info({
+              msg: `Aborting because upstream dependency failed.`,
+              metadata: metadataForLog(d.task, "error", inputVersion),
```
eysi09, line 164: Not part of this PR per se but we should really try and do away with this metadata field. 

I don't think it's really used anywhere and we already have the `context` field which serves the same purpose.

core/src/commands/base.ts
```diff
@@ -1053,8 +1054,14 @@ export async function handleProcessResults(
       return f && f.error ? [toGardenError(f.error)] : []
     })
 
+    const errMsg = abortedCount
+      ? failedCount
+        ? `${failedCount} requested ${taskType} action(s) failed, ${abortedCount} aborted!`
+        : `${abortedCount} requested ${taskType} action(s) aborted!`
+      : `${failedCount} requested ${taskType} action(s) failed!`
```
stefreak, line 1061: This is hard to review due to the cyclomatic complexity of this function; Do we have tests for this error message?

Looking at tests it would be much easier to tell what the error message looks like now.

stefreak (approved):
The code and the added test looks good to me, but the updated status counting code for the error message is hard to understand. Adding a test for it might be a good idea (optional).

PR #5361, opened 2023-11-07
chore: bump to node 21.1.0

TimBeyer:
Does what it says.

See https://github.com/TypeStrong/ts-node/issues/1997 for the reason for the `ts-node` change even with `esm: true` being set.

Closes https://github.com/garden-io/garden/issues/4158


shumailxyz:
This should close issue https://github.com/garden-io/garden/issues/4158, right?

TimBeyer:
There's an issue with `ts-node` which if fixed using the suggested `node --experimental-loader=ts-node/esm` causes terrible performance for everything run through `ts-node`.
The performance is so bad, that I wouldn't even consider merging it even if it's just impacting CI script times.

We'd need to either wait for a fix in the project, or maybe see if we can somehow skip running things with `ts-node`.

TimBeyer:
The new fix seems to be better so I guess we're good to go with this.

scripts/register-hook.js
```diff
@@ -0,0 +1,5 @@
+import url from 'node:url'
```
stefreak, line 1: It might be good to add a comment here explaining why this file exists.

stefreak (approved):
👍

PR #5363, opened 2023-11-09
fix(sync-mode): use the same source path schemas for all action types

### LINKED ISSUE #4982: 0.13: [Bug]: Helm Module devMode relative sourcePath resolves to non-POSIX full path and then fails validation
### Current Behavior

After upgrade to Garden Bonsai (version 0.13.12), an existing config for Module of type helm with `devMode.sync` set is not valid anymore.
Running `garden validate` raises the error: `"key .sync.paths[0][sourcePath] must be a POSIX-style path"`

Running `garden validate -l=silly` shows that garden transforms this value to a full path that starts with `C:\Users\...` on Windows:

### Expected behavior

`garden validate` should not fail.

### Reproducible example

The attached zip file contains a minimal garden project that triggers the issue:

Module defines: `devMode.sync[0].source: "src"`

My project is in directory `C:\Users\User\proj\demo\`
Running `garden validate -l=silly` shows that the converted value becomes: 
`sync.paths[0].sourcePath: "C:\Users\User\proj\demo\src"`

```
context: spec for Deploy type=helm name=db-local (from module db-local)
schemaMetadata: {}
errorDescription: "key .sync.paths[0][sourcePath] must be a POSIX-style path. Available keys: namespace, portForwards, releaseName, timeout, values, valueFiles, atomic, chart, defaultTarget, sync, localMode)"
errorDetails:
  - message: key .sync.paths[0][sourcePath] must be a POSIX-style path
    path:
      - sync
      - paths
      - 0
      - sourcePath
    type: base
    context:
      label: sync.paths[0].sourcePath
      value: C:\Users\User\proj\demo\src
      key: sourcePath

```

which is then rejected as invalid because it is not a valid POSIX path.

### Workaround

I did not manage to find a workaround - had to disable devMode for this module.

### Additional context

The issue only occurs on Windows.
Behavior same for both `cmd.exe` as from `gitbash`.

### Your environment

* **OS:** Windows 11.
* **How I'm running Kubernetes:** k0s local or gke remote 

`garden version`
[project.garden.yml.zip](https://github.com/garden-io/garden/files/12431947/project.garden.yml.zip)
 : garden version: 0.13.12

### LINKED ISSUE #5031: 0.13: [Bug]: devMode syncs in windows throwing errors
### Current Behavior

using 0.13 on windows we get errors about the devMode relative sync paths that we have in our config files.

this works fine on 0.12 with the same project, as well as on macOs/linux on 0.12 and 0.13

Error validating spec for Deploy type=kubernetes name=backend(from module backend): key .sync.paths[0][sourcePath] must be a POSIX-style path, key .sync.paths[1][sourcePath] must be a POSIX-style path, key .sync.paths[2][sourcePath] must be a POSIX-style path. Available keys: files, kustomize, manifests, namespace, portForwards, timeout, waitForJobs, defaultTarget, sync, localMode)

example:
```
kind: Module
type: kubernetes
name: backend
description: >
  The main backend service
devMode:
  sync:
    - source: backend
      target: /var/code/backend
      exclude:
        [
          backend/dir1,
          backend/dir2,
        ]
      mode: one-way-replica
    - source: backend/dir1
      target: /var/code/backend/dir1
      mode: one-way-replica
    - target: /var/code/backend/dir2
      source: backend/dir2
      mode: two-way
```

### Expected behavior

that it should work with these configs since it does in 12, and on other os's

### Workaround

use garden 12 on windows

### Your environment

* **OS:** win 11
* **How I'm running Kubernetes:** gke

0.13.13

vvagaytsev:
This unifies the way we validate the sync mode's source paths for all action types.

Fixes #4982
Fixes #5031

I tested and verified the changes on the Windows machine. I tried both relative and absolute Windows paths as a sync-mode source path.

I also added 2 test cases for Windows in 40ec1ea0b6f38ea9ac37ab57055cc8e29f50aca1.
Verified the failure of those tests manually and in [CI](https://app.circleci.com/pipelines/github/garden-io/garden/21852/workflows/460299a5-632f-4e11-803c-68d11368279a/jobs/433221).

### DIFF
--- a/core/src/plugins/container/config.ts
+++ b/core/src/plugins/container/config.ts
@@ -31,6 +31,7 @@ import type { RunAction, RunActionConfig } from "../../actions/run.js"
 import { memoize } from "lodash-es"
 import type Joi from "@hapi/joi"
 import type { OctalPermissionMask } from "../kubernetes/types.js"
+import { templateStringLiteral } from "../../docs/common.js"
 
 export const defaultDockerfileName = "Dockerfile"
 
@@ -237,13 +238,21 @@ export const syncDefaultGroupSchema = memoize(() =>
     .description("Set the default group on files and directories at the target. " + ownerDocs)
 )
 
+const exampleActionRef = templateStringLiteral("actions.build.my-container-image.sourcePath")
+
 export const syncSourcePathSchema = memoize(() =>
   joi
     .string()
     .default(".")
     .description(
       deline`
-        POSIX-style or Windows path of the directory to sync to the target. Defaults to the config's directory if no value is provided.
+        POSIX-style or Windows-style local path of the directory to sync to the target.
+        Can be either absolute or relative to the source directory where the Deploy action is defined.
+
+        This should generally be a templated path to another action's source path (e.g. ${exampleActionRef}), or a relative path.
+        If a path is hard-coded, you must make sure the path exists, and that it is reliably the correct path for every user.
+
+        Defaults to the Deploy action's config's directory if no value is provided.
         `
     )
     .example("src")
--- a/core/src/plugins/container/config.ts
+++ b/core/src/plugins/container/config.ts
@@ -237,6 +237,17 @@ export const syncDefaultGroupSchema = memoize(() =>
     .description("Set the default group on files and directories at the target. " + ownerDocs)
 )
 
+export const syncSourcePathSchema = memoize(() =>
+  joi
+    .string()
+    .default(".")
+    .description(
+      deline`
+        POSIX-style or Windows path of the directory to sync to the target. Defaults to the config's directory if no value is provided.
+        `
+    )
+    .example("src")
+)
 export const syncTargetPathSchema = memoize(() =>
   joi
     .posixPath()
@@ -254,15 +265,7 @@ export const syncTargetPathSchema = memoize(() =>
 const containerSyncSchema = createSchema({
   name: "container-sync",
   keys: () => ({
-    source: joi
-      .string()
-      .default(".")
-      .description(
-        deline`
-        POSIX-style or Windows path of the directory to sync to the target. Defaults to the config's directory if no value is provided.
-        `
-      )
-      .example("src"),
+    source: syncSourcePathSchema(),
     target: syncTargetPathSchema(),
     exclude: syncExcludeSchema(),
     mode: syncModeSchema(),
--- a/core/src/plugins/container/config.ts
+++ b/core/src/plugins/container/config.ts
@@ -239,6 +239,8 @@ export const syncDefaultGroupSchema = memoize(() =>
 )
 
 const exampleActionRef = templateStringLiteral("actions.build.my-container-image.sourcePath")
+const backSlash = "`\\`"
+const forwardSlash = "`/`"
 
 export const syncSourcePathSchema = memoize(() =>
   joi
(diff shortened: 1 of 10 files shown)


core/src/plugins/container/config.ts
```diff
@@ -237,6 +238,25 @@ export const syncDefaultGroupSchema = memoize(() =>
...
+const exampleActionRef = templateStringLiteral("actions.build.my-container-image.sourcePath")
+
+export const syncSourcePathSchema = memoize(() =>
+  joi
+    .string()
+    .default(".")
+    .description(
+      deline`
+        POSIX-style or Windows-style local path of the directory to sync to the target.
+        Can be either absolute or relative to the source directory where the Deploy action is defined.
```
stefreak, line 250: Will a Windows-style relative path automatically transformed to a Unix-path when you use the action on Linux or MacOS?

vvagaytsev: Good catch, thanks! I'll check if there are any advanced options to ensure that conversion. I hope it's doable.
Otherwise, we can implement it and apply it in all similar places.

vvagaytsev: It looks like there is no way to do automatic conversion with joi. I've rewritten the docs in 59ca3ff1dd7fefa6917c6adc7f229d0b04256d1a. Now it says that the path should be in a platform-specific format.

stefreak: I'm not sure if `platform-specific` is a good way to express this. 

Please correct me if I'm wrong, but I think the following statements are true: 
1. I am assuming here that windows-style paths won't work with Mutagen under Linux (Have we tested that though? If this assumption is false we are over-worrying here)
2 The Unix-style paths with a forward slash (`/`) will work on Windows computers (I guess? Otherwise you can't use sync mode with coworkers on other platforms?)
3. Windows-style paths (with a backslash) won't be portable to Unix-like-systems like Linux or MacOS (Unless mutagen somehow will convert the paths?)
4. Unix-style absolute paths won't be portable to Windows systems, and Windows-like absolute paths with a drive letter won't be portable to Unix-systems, obviously

My suggestion:
```
        Path to a local local directory to be synchronized with the target.
        
        Can be either absolute or relative to the source directory where the Deploy action is defined.
        
        We recommend sticking with relative paths here, and using forward slashes (`/`) as a delimiter, as Windows-style paths with back slashes (`\`) and absolute paths will work on some platforms, but they are not portable and will not work for users on other platforms.
```

I have not tested the portability of using absolute paths and backslashes – have you?

vvagaytsev: @stefreak right. Windows paths are not portable to Linux and Unix-like systems. Absolute POSIX paths are also not portable to Windows because of the drive letter format. Those can be used under WSL.

Relative POSIX-paths work well on Windows too. So that's the recommended setup. Thanks for the suggestion.

vvagaytsev: Updated it in 13fd82ce59ef80bf5119551adc598802000edfb6.

core/src/plugins/container/config.ts
```diff
@@ -237,6 +238,27 @@ export const syncDefaultGroupSchema = memoize(() =>
     .description("Set the default group on files and directories at the target. " + ownerDocs)
 )
 
+const exampleActionRef = templateStringLiteral("actions.build.my-container-image.sourcePath")
+const backSlash = "`\\`"
```
TimBeyer, line 242: Might be overkill here, but you could do 

```
import * as posixPath from "path/posix"
import * as winPath from "path/win32"
```

and then use `posixPath.sep` and `winPath.sep` for those constants.

TimBeyer: Feels like for a backslash and a forward slash that's not really worth it though.

vvagaytsev: Changed that in 2d332e5462b3f37ff9ca98eb5268cd07700af7c7.

TimBeyer: Now it's missing the backticks for the markdown rendering though.
Maybe the previous solution was a bit nicer in the end.

TimBeyer (approved):

PR #5409, opened 2023-11-14
fix(sync-mode): avoid collisions in sync key prefixes

### LINKED ISSUE #5313: 0.13: [Bug]: Similarly named deploy actions causing sync issues
### Current Behavior

We have two similarly named services, `monolith` and `monolith-frontend`. These both have deploy actions and set up several `sync` paths.

When doing a deploy where both services are in sync mode, the similar name will cause the `monolith` deployment to terminate syncs from the `monolith-frontend` deployment 

Deploy command: 
```
garden deploy --env=dns --sync monolith,monolith-frontend
```

Termination errors:
Screenshot 2023-10-27 at 4 26 08 PM

### Expected behavior
Both services are able to deploy their syncs simultaneously without issue.

### Workaround
After deploying with the above command, we can restart the `monolith-frontend` syncs with `sync restart monolith-frontend`. This lets the syncs get set up successfully without intervention from the `monolith` deployment.

### Suggested solution(s)
The error is coming from [this check](https://github.com/garden-io/garden/blob/31f442029a050b539b369c2869a28894652337f7/core/src/plugins/kubernetes/sync.ts#L617-L622C4) for unexpected or previously created syncs.

Because the [getSyncKeyPrefix](https://github.com/garden-io/garden/blob/31f442029a050b539b369c2869a28894652337f7/core/src/plugins/kubernetes/sync.ts#L825-L827) method is using `kebabCase` and is compared via `startsWith`, the `monolith-frontend` action name collides with the `monolith` action.

```
# Full sync keys from `sync status`:
k-8-s-dns-dev-dentonalford-monolith-frontend-deployment-monolith-frontend-some-sync-path
k-8-s-dns-dev-dentonalford-monolith-deployment-monolith-different-sync-path

# presumed prefix key result:
k-8-s-dns-dev-dentonalford-monolith-frontend-
k-8-s-dns-dev-dentonalford-monolith- # this will match both prefix keys when using `startsWith`
```

Ensuring the full action name is present in the check should resolve the issue.

### Additional context
To validate the issue, I updated the `monolith-frontend` service name to `monolithfrontend`. This resolved the issue in my environment.

### Your environment

* **OS:** macOS 13.5.2
* **How I'm running Kubernetes:** Azure

garden version: 0.13.18

vvagaytsev:

Fixes #5313


stefreak (approved):

shumailxyz (approved):
LGTM, thanks for fixing it 👍

PR #5410, opened 2023-11-14
perf: do not compute log message until required

TimBeyer:
Before letting go of the whole performance epic, I did one last round of CPU profiling and noticed that for large projects we spend a large amount of time calling functions that are only used in higher log levels, which aren't actually really set most of the time. The problem is that we do the heavy processing for the log entries, but then we eventually discard them.
Thus I changed the logger so it can accept a callback, and if nothing logs it, the callback never gets called.

I don't want to spend large amounts of time on this anymore, and this was just a quick search and replace thing, so if it works let's consider merging it, if not, we can still keep this around as inspiration.

For now this was only applied to the `silly` log level since that's where the most expensive log lines were coming from.

This improves performance of `garden validate` on a project with 500 modules from ~32 to ~25 seconds. 
So this change gives an about 20% time saving for very large projects, which is pretty nice for such a simple change.

Fixes #

### DIFF
--- a/core/src/logger/log-entry.ts
+++ b/core/src/logger/log-entry.ts
@@ -132,6 +132,19 @@ interface CreateCoreLogParams
   origin?: string
 }
 
+export type Msg = string | (() => string)
+
+export function resolveMsg(logEntry: LogEntry): string | undefined {
+  return typeof logEntry.msg === "function" ? logEntry.msg() : logEntry.msg
+}
+
+export function transformMsg(msg: Msg, transformer: (input: string) => string): Msg {
+  if (typeof msg === "function") {
+    return () => transformer(msg())
+  }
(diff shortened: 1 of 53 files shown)


core/src/logger/log-entry.ts
```diff
@@ -318,38 +340,38 @@ export abstract class Log<C extends BaseContext = LogContext> implements LogConf
   /**
    * Render a log entry at the silly level. This is the highest verbosity.
    */
-  silly(params: string | LogParams) {
+  silly(params: Msg | LogParams) {
```
stefreak, line 343: I think we might want to force using the closure instead of a string for silly

stefreak (approved):
I love this! It really is a very simple change for a 20% performance improvement

PR #5413, opened 2023-11-14
fix(sync-mode): remove confusing warning on mutagen connectivity error

### LINKED ISSUE #5411: 0.13: [Bug]: Misleading warning on shortening the project path
### Current Behavior

I get this warning:
```
⚠ deploy.api           → Consider making your Garden project path shorter. Syncing could fail because of Unix socket path length limitations. It's recommended that the Garden project path does not exceed 70 characters. The actual value depends on the platform and the mutagen version.
```

How do I even do that?

### Expected behavior

Bug https://github.com/garden-io/garden/issues/4527 is fixed, so I expect no warning

### Reproducible example

`garden deploy` in garden-platform

### Suggested solution(s)

Remove the warning

### Your environment

* **OS:** macos

`garden version` 0.13.19

vvagaytsev:

Fixes #5411

### DIFF
--- a/core/src/mutagen.ts
+++ b/core/src/mutagen.ts
@@ -11,6 +11,7 @@ import dedent from "dedent"
 import type EventEmitter from "events"
 import type { ExecaReturnValue } from "execa"
 import fsExtra from "fs-extra"
+
 const { mkdirp, pathExists } = fsExtra
 import hasha from "hasha"
 import pRetry from "p-retry"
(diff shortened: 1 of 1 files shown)


core/src/mutagen.ts
```diff
@@ -652,10 +652,6 @@ export class Mutagen {
           await this.ensureDaemon()
           await sleep(2000 + loops * 500)
         } else {
-          emitNonRepeatableWarning(
-            this.log,
-            `Consider making your Garden project path shorter. Syncing could fail because of Unix socket path length limitations. It's recommended that the Garden project path does not exceed ${MUTAGEN_DATA_DIRECTORY_LENGTH_LIMIT} characters. The actual value depends on the platform and the mutagen version.`
-          )
```
shumailxyz, line 658: I think we should also refactor this funciton 
https://github.com/garden-io/garden/blob/981b6832e4dd42a0823032f6576587514d90e476/core/src/mutagen.ts#L801C17-L801C34

And remove the check for `MUTAGEN_DATA_DIRECTORY_LENGTH_LIMIT` and always use hash based approach

vvagaytsev: Yes, makes sense. That would be more robust and predictable.

shumailxyz (approved):
LGTM, thanks 👍

PR #5428, opened 2023-11-16
improvement(core): better action lifecycle logs

eysi09:

This commits adds more logging to the action lifecycle. That is logs about getting status of a given action, the results, next steps, etc.

The logging is done in the task decorators that are also used for emitting events.

The goal is obviously to make the logs more informative but also more consistent.

Fixes #

Here's a screenshot of the new logs. Note that the screenshot uses different colors to what we currently have so that the change is more visible. But we'll make those color changes in a separate PR though.

Screenshot 2023-11-16 at 23 54 26

### DIFF
--- a/core/src/logger/renderers.ts
+++ b/core/src/logger/renderers.ts
@@ -18,12 +18,12 @@ import { highlightYaml, safeDumpYaml } from "../util/serialization.js"
 import type { Logger } from "./logger.js"
 import { logLevelMap, LogLevel } from "./logger.js"
 import { toGardenError } from "../exceptions.js"
-import type { Styles } from "./styles.js"
 import { styles } from "./styles.js"
+import type { Chalk } from "chalk"
 
 type RenderFn = (entry: LogEntry, logger: Logger) => string
 
-export const SECTION_PADDING = 20
+export const SECTION_PADDING = 25
 
 export function padSection(section: string, width: number = SECTION_PADDING) {
   const diff = width - stringWidth(section)
@@ -92,7 +92,7 @@ export function renderTimestamp(entry: LogEntry, logger: Logger): string {
 }
 
 export function getStyle(level: LogLevel) {
-  let style: Styles
+  let style: Chalk
   if (level === LogLevel.error) {
     style = styles.error
   } else if (level === LogLevel.warn) {
@@ -116,20 +116,15 @@ export function getSection(entry: LogEntry): string | null {
 
 export function renderMsg(entry: LogEntry): string {
   const { context, level } = entry
-  const msg = resolveMsg(entry)
-  const { origin } = context
+  const msg = resolveMsg(entry) || ""
   const style = getStyle(level)
 
-  if (!msg) {
-    return ""
-  }
+  // For log levels higher than "info" we print the log level name.
+  const logLevelName = entry.level > LogLevel.info ? `[${logLevelMap[entry.level]}] ` : ""
 
-  // TODO: @eysi Should we strip here?
-  // if (level > LogLevel.info) {
-  //   msg = stripAnsi(msg)
-  // }
+  const origin = context.origin ? `[${styles.italic(context.origin)}] ` : ""
 
-  return style(origin ? `[${styles.italic(origin)}] ` + msg : msg)
+  return style(`${logLevelName}${origin}${msg}`)
 }
 
 export function renderData(entry: LogEntry): string {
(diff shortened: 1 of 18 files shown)


core/src/logger/renderers.ts
```diff
@@ -116,20 +116,15 @@ export function getSection(entry: LogEntry): string | null {
...
   const { context, level } = entry
-  const msg = resolveMsg(entry)
-  const { origin } = context
+  const msg = resolveMsg(entry) || ""
   const style = getStyle(level)
 
-  if (!msg) {
-    return ""
-  }
+  // For log levels higher than "info" we print the log level name.
```
eysi09, line 122: I moved the log level name to the log line proper. 

So now it's:

```
section => [level name] message
```

as opposed to:

```
section [level name] => message
```

core/src/logger/styles.ts
```diff
@@ -8,39 +8,18 @@
...
-type StyleKey = "bold" | "underline" | "italic" | "link" | "section" | "command"
-type StyleFn = (s: string) => string
-
-export type Styles = StyleFn & { [key in ThemeKey | StyleKey]: Styles }
-
 /**
  * A map of all the colors we use to render text in the terminal.
  */
 const theme = {
-  primary: chalk.grey as unknown as Styles,
```
eysi09, line 36: This was just plain wrong so I removed it.

core/test/unit/src/logger/renderers.ts
```diff
@@ -121,17 +120,7 @@ describe("renderers", () => {
     it("should print the log level if it's higher then 'info'", () => {
       const entry = logger.createLog().debug({ msg: "hello world" }).getLatestEntry()
 
-      expect(formatForTerminal(entry, logger)).to.equal(
-        `${styles.primary("[debug]")} ${styles.primary("hello world")}\n`
-      )
-    })
-    it("should print the log level if it's higher then 'info' after the section if there is one", () => {
```
eysi09, line 128: I changed this behaviour so we can remove this test. There's already a test case for the "new" behaviour.

core/src/tasks/base.ts
```diff
@@ -540,6 +577,73 @@ export function emitProcessingEvents<
...
+  willRerun,
+}: {
+  action: A
+  operation: "getStatus" | "process"
+  log: ActionLog
+  result: R
+  willRerun: boolean
+}) {
+  if (result.state === "ready") {
+    const readyStr = {
```
TimBeyer, line 613: I feel like this code could use some extracting of those string maps to the outer scope and adding a few line breaks for clarity.

core/src/tasks/base.ts
```diff
@@ -492,6 +512,19 @@ export function emitProcessingEvents<
     const actionKind = this.action.kind.toLowerCase() as Lowercase<A["kind"]>
     const eventName = actionKindToEventNameMap[actionKind]
     const startedAt = new Date().toISOString()
+    const log = this.log.createLog()
+    const version = this.action.versionString()
+    const verb = {
```
TimBeyer, line 517: This mapping here should also probably live outside of the function.

Since they're all always using the same indexing over `Build`, `Deploy` etc, we can probably do something like

```
const PROCESSING_EVENTS_LOG_STRINGS = {
  Build: {
    verb: "Building",
    ready: "built",
    run: "will be built",
    rerun: "re-build",
  },
...
}
```

eysi09:
Thanks for the review @TimBeyer! I originally co-located these map with the log lines because I kept tweaking it and it made things more visible. 

But I agree that your suggestion is cleaner, addressed in the latest commit.

vvagaytsev (approved):

TimBeyer (approved):

core/src/tasks/base.ts
```diff
@@ -428,6 +476,16 @@ export function emitGetStatusEvents<
     try {
       const result = (await method.apply(this, args)) as R & ExecuteActionOutputs<A>
 
+      const willRerun = this.force && !statusOnly
+      if (result.state === "ready" && !willRerun) {
+        log.success({ msg: `Already ${logStrings.ready}`, showDuration: false })
```
vvagaytsev, line 481: @eysi09 I've just noticed that the line `ℹ build.name           → Already built` is no longer displayed in the term2 logs on macOS. Any idea what could be wrong with the `success` function?

vvagaytsev: Hmm.. Looks like this code is not reachable when I run a Build action for an already-built image. BTW, the same line is getting printed in 0.13.20.

vvagaytsev: Hmm, this is no longer reprodubicle. Sorry for bothering you. I might be some dirty npm dev mode state on my side :)

eysi09: No worries! There's a chance this may need some tweaking still. 

I noticed at least one issue already where we're printing extra new lines in dev mode. I'll fix that.

PR #5429, opened 2023-11-16
test: fix flaky test for deploy logs

TimBeyer:
This test is flaky because the order of log entries isn't always stable now.
Thus we find the correct log line first and then assert that it exists.
Actually all the following tests in this file did the same pattern, so maybe something like this has been the case in the past as well and only this test somehow survived.

Fixes #


stefreak:
@TimBeyer I like this! 🥳 
There is a lint error though.

vvagaytsev (approved):

PR #5431, opened 2023-11-17
chore: restructuring `util/util.ts` (part 1)

vvagaytsev:
Currently, the [`util/util.ts`](https://github.com/garden-io/garden/tree/main/core/src/util/util.ts) file is a big pile of helper functions of different kinds.

This PR is the first step to decompose that utility module into smaller, semantically-scoped groups of helpers.

That decomposition will help to navigate the codebase more easily and allow us to deduplicate some helper functions and interfaces, making them more generic and unified.

Fixes #


core/src/util/util.ts
```diff
@@ -85,6 +83,10 @@ export type Unpacked<T> = T extends (infer U)[]
 export type ExcludesFalsy = <T>(x: T | false | null | undefined) => x is T
 export type MakeOptional<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>
 
+export interface Dictionary<T> {
```
TimBeyer, line 86: Just a refactoring so definitely not a blocker, but I guess we could replace that with `Record<string, T>`?

TimBeyer (approved):

core/src/util/util.ts
vvagaytsev: Good point. Maybe, this type was introduced as a short-handed type for `string`-keyed `Record`. In that case, we can just re-declare it as `type Dictionary<T> = Record<string, T>`

vvagaytsev: This can be addressed in the next PR, there will be a series of PRs to refactor `util/util.ts` :)

PR #5435, opened 2023-11-17
chore: move Garden Cloud utils to own module

vvagaytsev:
This is the continuation of the util-refactoring PR series started in #5431.

Fixes #


core/test/unit/src/util/garden-cloud.ts
```diff
@@ -0,0 +1,69 @@
...
+      })
+    })
+
+    context("when top-level domain is .garden", () => {
+      context("when 2nd level domain is .app", () => {
+        it(`returns "Garden Cloud" for https urls`, () => {
+          expect(getCloudDistributionName("https://backend.app.garden")).to.eql("Garden Cloud")
+        })
+
+        // TODO: should we care about http/https here? Do we always get only https:// urls?
```
vvagaytsev, line 28: @mkhq @eysi09 any ideas here? Should we consider introducing a type for domains? (Maybe it's an overkill.) Can we get it as a template variable at some point?

Now we also accept garbage strings like `"aaabbbcc"`, for such ones this function will return `"Garden Enterprise"`.

core/src/util/garden-cloud.ts
```diff
@@ -0,0 +1,42 @@
+/*
```
eysi09, line 1: Can we just call this file `cloud.ts`, it's consistent with other cloud related filenames.

core/test/unit/src/util/garden-cloud.ts
eysi09: Can we use the URL library to construct a URL from the string and just read the hostname so that we don't need to worry about the protocol. 

It could just throw if the name isn't a valid hostname. It'd be a runtime error but we'd have runtime issues anyway if the hostname is bad.

eysi09 (changes requested):

core/src/util/garden-cloud.ts
vvagaytsev: Done in 28bd58bbe50d94e15346443fb6b1f0e11c8d7597.

core/test/unit/src/util/garden-cloud.ts
vvagaytsev: @eysi09 thanks for the suggestion, sounds good!

I made a quickfix in d0e32f5365d5eaf0b7f356f949f9e924ff4090d5. That minimizes the semantic changes made in this PR.

Let's make a separate PR with improved URL validation and examination.

eysi09 (dismissed):

eysi09 (approved):

PR #5233, opened 2023-10-11
chore: ESM

TimBeyer:
**What this PR does / why we need it**:

Switches to ESM. 

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

PR #5440, opened 2023-11-17
chore: fix edge version

stefreak:
Currently edge versions pretend to be production version. This was a regression introduced in #5233 

This commit fixes that, to ensure that we do not confuse edge builds
with production builds and to ensure that our analytics is correct.

Currently edge builds pretend to be production builds:
```
steffen@MacBook-Pro-3 garden % garden self-update edge-bonsai
Update Garden 🗞️

No installation directory specified via --install-dir option. Garden will be re-installed to the current installation directory: /opt/homebrew/Cellar/garden-cli/0.13.19/libexec
Current Garden version: 0.13.20
Target Garden version to be installed: edge-bonsai
Latest release version: 0.13.20

Downloading version edge-bonsai from https://download.garden.io/core/edge-bonsai/garden-edge-bonsai-macos-arm64.tar.gz...
Backing up prior installation to /opt/homebrew/Cellar/garden-cli/0.13.19/libexec/.backup/0.13.20...
Extracting to installation directory /opt/homebrew/Cellar/garden-cli/0.13.19/libexec...

Done!
steffen@MacBook-Pro-3 garden % garden version
garden version: 0.13.20
```

### DIFF
--- a/cli/src/build-pkg.ts
+++ b/cli/src/build-pkg.ts
@@ -199,7 +199,19 @@ async function zipAndHash({ targetDir, archiveName, cwd, prefix, fileNames }: Zi
 
 async function buildBinaries(args: string[]) {
   const argv = minimist(args)
-  const version = argv.version || getPackageVersion()
+
+  let version: string
+
+  if (argv.version === "edge" || argv.version.startsWith("edge-")) {
+    const gitHash = await exec("git", ["rev-parse", "--short", "HEAD"])
+    version = `${getPackageVersion()}-${argv.version}-${gitHash.stdout}`
+  } else if (argv.version) {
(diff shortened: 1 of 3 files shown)


cli/src/build-pkg.ts
```diff
@@ -199,7 +199,19 @@ async function zipAndHash({ targetDir, archiveName, cwd, prefix, fileNames }: Zi
 
 async function buildBinaries(args: string[]) {
   const argv = minimist(args)
-  const version = argv.version || getPackageVersion()
+
+  let version: string
```
vvagaytsev, line 203: Non-blocking nit. We can extract this as a function and use `const version = ...` instead.

PR #5447, opened 2023-11-20
improvement(api): send action type to cloud

eysi09:

Sends the action type to cloud so that it can e.g. be displayed in the graph UI and in command results in general. 

We weren't doing that previously because a bit of a legacy baggage but that has been resolved. 

Fixes #

Garden Cloud itself was updated awhile ago to handle this data so this only required Core fixes.

### DIFF
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -40,7 +40,7 @@ import { describeSchema, parseActionReference } from "../config/common.js"
 import type { GroupConfig } from "../config/group.js"
 import { ActionConfigContext } from "../config/template-contexts/actions.js"
 import { validateWithPath } from "../config/validation.js"
-import { ConfigurationError, PluginError, InternalError, ValidationError, GardenError } from "../exceptions.js"
+import { ConfigurationError, PluginError, InternalError, GardenError } from "../exceptions.js"
 import { overrideVariables, type Garden } from "../garden.js"
 import type { Log } from "../logger/log-entry.js"
 import type { ActionTypeDefinition } from "../plugin/action-types.js"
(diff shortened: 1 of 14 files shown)


core/src/graph/actions.ts
```diff
@@ -777,15 +777,29 @@ function dependenciesFromActionConfig(
   const deps: ActionDependency[] = config.dependencies.map((d) => {
     try {
       const { kind, name } = parseActionReference(d)
-      return { kind, name, explicit: true, needsExecutedOutputs: false, needsStaticOutputs: false }
+      const depKey = actionReferenceToString(d)
+      const depConfig = configsByKey[depKey]
```
eysi09, line 781: Just FYI @thsig, this won't work. 

If the action depends on a module it won't be in that map and `depConfig` will be `undefined`. 

That's why the CI tests are failing.

core/test/integ/src/plugins/kubernetes/kubernetes-type/handlers.ts
```diff
@@ -115,7 +115,7 @@ describe("kubernetes-type handlers", () => {
   }
 
   async function deployInNamespace({ nsName, deployName }: { nsName: string; deployName: string }) {
-    garden.setModuleConfigs([withNamespace(nsModuleConfig, nsName)])
+    garden.mergeModuleConfigs([withNamespace(nsModuleConfig, nsName)])
```
eysi09, line 118: In the previous version these tests were actually not working right because when setting the module config here we replaced the actual module configs from the test projects. 

With this change we now merge the module config we're setting dynamically with the existing configs so that nothing gets overwritten.

vvagaytsev (approved):

core/src/util/testing.ts
```diff
@@ -327,10 +341,6 @@ export class TestGarden extends Garden {
     })
   }
 
-  setWorkflowConfigs(workflowConfigs: WorkflowConfig[]) {
```
vvagaytsev, line 330: This caused multiple build errors in the tests.

vvagaytsev (changes requested):

core/src/util/testing.ts
eysi09: Yeah my bad, it looked like nothing was depending on it but turns the LSP was just borking. Added it back.

vvagaytsev (approved):
👍

PR #5464, opened 2023-11-21
improvement(logger): better provider resolution lifecycle logs

eysi09:

This commits add more logging around the provider resolution phase of the Garden initialization. Also ensures that the log context passed to the plugin handlers has the provider name set.

For context, these log improvements I've been making are part of the "getting started" video work. The idea there is to explain how Garden works and we need better lifecycle logging to support that.

Fixes #

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -224,6 +224,7 @@ ${renderCommands(commands)}
     } = parsedOpts
 
     const parsedCliVars = parseCliVarFlags(cliVars)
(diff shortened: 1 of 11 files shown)


vvagaytsev:
@eysi09 there are some test failures. Please check if those are caused by the changes made in the PR.

eysi09:
> @eysi09 there are some test failures. Please check if those are caused by the changes made in the PR.

Should be fixed now.

vvagaytsev (approved):

PR #5465, opened 2023-11-21
improvement(logger): use 'white' as primary color

eysi09:

Before this commit, our primary log color was chalk.grey and people generally agree that it's too vague.

This commit changes the primary log color we've been using since we first released Garden to chalk.white for better accessibility and more fun.

In general we can make this configurable. But I figured I'd just bite the bullet.

See also discussion in: https://github.com/garden-io/garden/pull/5464

Fixes #5464


shumailxyz (approved):
Let's do this 👍

PR #5470, opened 2023-11-22
Fix exec command strings

### LINKED ISSUE #2957: garden exec splits quoted arguments
### Current Behavior

Running `garden exec api '/usr/bin/env -v echo "hello world"'` gives this output:

```
executing: echo
   arg[0]= 'echo'
   arg[1]= '"hello'
   arg[2]= 'world"'
"hello world"
```
In the above example "hello" and "world" are being passed as two separate arguments even though they are quoted.

### Expected behavior

Running `garden exec api '/usr/bin/env -v echo "hello world"'` should match the output of `kubectl exec pod -- /usr/bin/env -v echo "hello world"`

```
executing: echo
   arg[0]= 'echo'
   arg[1]= '"hello world"'
"hello world"
```

In this example "hello world" is passed as a single argument.

### Your environment

The examples were generated running on mac os with garden `0.12.40`

twelvemo:
Fixes the garden exec command, which wasn't working when given more than two arguments e.g. `garden exec backend "ls -la"` would fail with executable file not found, while `garden exec backend ls`  succeeds. If not quoting the command with two arguments, garden would interpret the `-la` as a garden argument and fail as well.

This PR restores the behavior for the command argument as it was in 0.12.x which means that commands are getting split which still fails in cases where there are quoted substrings see #2957. 
It also  introduces the option to use `--` as a separator to specify commands the same way kubectl does it. Using the `--` separator results in quoted strings to be interpreted as one argument.
e.g.:
`garden exec debian -- /bin/sh -c echo "hello world"`

Adding a deprecation warning for using the command argument without the `--` separator.

Fixes #2957

### DIFF
--- a/core/src/commands/exec.ts
+++ b/core/src/commands/exec.ts
@@ -29,6 +29,10 @@ const execArgs = {
       return Object.keys(configDump.actionConfigs.Deploy)
     },
   }),
+  command: new StringParameter({
+    help: "The command to run.",
+    required: false,
+  }),
 }
 
 const execOpts = {
@@ -57,12 +61,15 @@ export class ExecCommand extends Command<Args, Opts> {
   override description = dedent`
     Finds an active container for a deployed Deploy and executes the given command within the container.
     Supports interactive shells.
+    You can specify the command to run as a parameter, or pass it after a \`--\` separator. For commands
+    with arguments or quoted substrings, use the \`--\` separator.
 
     _NOTE: This command may not be supported for all action types._
 
     Examples:
 
-         garden exec my-service -- /bin/sh   # runs a shell in the my-service Deploy's container
+         garden exec my-service /bin/sh   # runs an interactive shell in the my-service Deploy's container
+         garden exec my-service -- /bin/sh -c echo "hello world" # prints "hello world" in the my-service Deploy's container and exits
   `
 
   override arguments = execArgs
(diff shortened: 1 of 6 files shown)


core/src/commands/exec.ts
```diff
@@ -88,6 +83,11 @@ export class ExecCommand extends Command<Args, Opts> {
   async action({ garden, log, args, opts }: CommandParams<Args, Opts>): Promise<CommandResult<ExecInDeployResult>> {
     const deployName = args.deploy
     const command = this.getCommand(args)
+
+    if (!command.length) {
+      throw new CommandError({ message: `No command specified. Nothing to execute.` })
+    }
+
```
shumailxyz, line 90: I see this as a breaking change as it'll affect the existing workflows.

Perhaps, we can considering supporting both at the moment: the existing `command` parameter and newly added `--`.

shumailxyz (changes requested):
Thanks for fixing it. I left a comment suggesting to support the old workflow as well as the new one using `--`.

core/src/commands/exec.ts
twelvemo: You are right. In 0.12 we just split the command, which will make multi argument commands work again to some degree but not fix #2957 (it was originally reported for 0.12.40). 
But i agree we can't just introduce a breaking change.

shumailxyz (approved):
Looks great, thanks! 💯 

Also thanks for adding the integ tests.

PR #5472, opened 2023-11-22
fix(git): don't rebuild container module on excluded files modification

vvagaytsev:
To avoid duplicate sanity checks and other repeatable code.

Fixes #


core/test/unit/src/vcs/git.ts
```diff
@@ -107,323 +113,634 @@ export const commonGitHandlerTests = (handlerCls: new (params: VcsHandlerParams)
...
+                    pathBuilder: (subDirName: string, deepDirName) => join(subDirName, deepDirName),
+                  },
+                  {
+                    name: "when directory is included by relative path with globs",
+                    pathBuilder: (subDirName: string, deepDirName) => join(subDirName, deepDirName, "**", "*"),
+                  },
+                  {
+                    name: "when directory is included by name with globs",
+                    // FIXME: shouldn't just '**/deepdir' work well too?
+                    pathBuilder: (_subDirName: string, deepDirName) => join("**", deepDirName, "**", "*"),
```
stefreak, line 374: It would be good to use both `path.posix.join` and `path.win32.join` in the future for tests, to make sure that we are compatible with both backslashes and forward slashes.

stefreak (approved):
Looks good to me so far! 👍  as far as I understand the portions of the code that need to change in PRs that land soon are marked using the `// FIXME` comments, right?

PR #5475, opened 2023-11-23
improvement(logger): some minor fixes

eysi09:

From commit message:

[improvement(logger): some minor fixes](https://github.com/garden-io/garden/pull/5475/commits/2120f9bd10b72022588161baedda51a8627e9178) 

Namely:
- Setting logs for some built-in providers to debug because that's just
  more fitting
- Do not print new line after resolving providers and graph. Doesn't
  render nicely.
- Add an "init" log context for the Garden class that logs when Garden
  has been initialized with duration for non-dev commands.
- Print link to Cloud at end of command run as well
- DO NOT log outdated statuses at a "warning" level
- Other minor tweaks

Fixes #

I know these keep raining in. My apologies. They're not super urgent per se and just small tweaks as I bump into more and more things while doing some content work. 

I tend to push them immediately to keep my own content branch sane and rebased.

### DIFF
--- a/core/src/tasks/resolve-provider.ts
+++ b/core/src/tasks/resolve-provider.ts
@@ -33,6 +33,7 @@ import { stableStringify } from "../util/string.js"
 import { OtelTraced } from "../util/open-telemetry/decorators.js"
 import { LogLevel } from "../logger/logger.js"
 import type { Log } from "../logger/log-entry.js"
+import { styles } from "../logger/styles.js"
 
 /**
  * Returns a provider log context with the provider name set.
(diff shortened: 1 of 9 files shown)


core/src/tasks/resolve-provider.ts
```diff
@@ -41,8 +41,8 @@ import type { Log } from "../logger/log-entry.js"
  * resolved per se. A bit hacky but this is just a cosmetic change.
  */
 function getProviderLog(providerName: string, log: Log) {
-  const verboseLogProviders = ["templated", "container"]
-  const fixLevel = verboseLogProviders.includes(providerName) ? LogLevel.verbose : undefined
+  const debugLogProviders = ["templated", "container"]
+  const fixLevel = debugLogProviders.includes(providerName) ? LogLevel.silly : undefined
```
vvagaytsev, line 45: Should this be `LogLevel.debug` instead of `LogLevel.silly`?

vvagaytsev (approved):

PR #5480, opened 2023-11-26
improvement(build): allow buildkit secret with sh

### LINKED ISSUE #5479: [FEATURE]: Allow build time secrets
### Background / Motivation

I am trying to use Garden to replace my workflow with kubernetes. However, a big part of my original workflow with local docker involves using `docker build --secret` to pass in build time secrets. After much effort around trying to hack this with the current system, it doesn't seem possible to do this securely. 

I'm using the buildkit builder.

PS: I've tried to hack this by using the extraFlags to try to use the `<()` syntax to create a tempfile with the token, but it seems like the kube api client sanitizes commands to not run any nested shells.

The user should be able to pass in their build time secrets to their build, securely.

The only way to do this without this feature is to use buildArgs, which exposes the secret in all subsequent layers as well as build logs. We COULD do multi-stage builds, but that is un-ideal if the original Dockerfile doesn't already do this.

There are two options to do this, based on how docker buildkit has exposed the secrets interface - environment and files:

1. Allow users to mount additional secret volumes. We can then use `extraFlags` to do `--secret id=TOKEN,src=<mounted>`
2. Allow users to pass in an environment variable to the `exec`. We can then use `extraFlags` to do `--secret id=TOKEN,env=<ENV>`

markgeejw:

Wrapping into sh allows shell functionality like <() which is useful for injecting build secrets.

Fixes #5479 but alternatives like volume mounting and properly environment variables injection should still be considered.

NA

### DIFF
--- a/core/src/plugins/kubernetes/container/build/buildkit.ts
+++ b/core/src/plugins/kubernetes/container/build/buildkit.ts
@@ -138,6 +138,8 @@ export const buildkitBuildHandler: BuildHandler = async (params) => {
     ...getBuildkitFlags(action),
   ]
 
+  const shCommand = ["sh", "-c", command.join(" ")]
+
   // Execute the build
   const buildTimeout = action.getConfig("timeout")
 
@@ -153,7 +155,7 @@ export const buildkitBuildHandler: BuildHandler = async (params) => {
 
   const buildRes = await runner.exec({
     log,
-    command,
+    shCommand,
     timeoutSec: buildTimeout,
     containerName: buildkitContainerName,
     stdout: outputStream,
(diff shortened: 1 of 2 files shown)


stefreak (changes requested):
Hi @markgeejw First of all, thank you so much for rolling the sleeves up and diving into the code to fix this 🙌 

Unfortunately this is not the way to go here, because letting the shell evaluate the `buildkitCommand` will lead to unexpected effects if the deployment registry or the image names contain special strings that the shell might interpret, and that can lead to shell injection / code execution issues when Garden is working with untrusted input e.g. in a GitHub action.

Can you share with us (perhaps in #5479) how your Garden configuration would look like with the `<()` workaround of providing the secrets? Maybe we can come up with a different way to declare the secrets, for example as `spec.secrets` in the `Build` action, and then I'd be very happy to guide you towards an implementation that we can merge.

vvagaytsev:
Closing this as stale. Feel free to re-open if you'd like to continue working on this.

PR #5502, opened 2023-12-04
feat: add config field and override param for require login

mkhq:

This introduces a new Project-level config parameter that can force users of a project to login via Garden Cloud when **there is a connected cloud project**. The main use case is when a team is using garden together with Garden Cloud and would like to make sure that all users are logged in. Since making this change would be breaking, we introduce a config field + an env. var, `GARDEN_REQUIRE_LOGIN_OVERRIDE`, to override the config. The default value of the config field is `false`, which does not require the user to login. When transitioning to 0.14, we will make the default value `true`. Note, when there is no cloud project configured through the `id`-field in the Project-level config, garden works as usual without cloud and enforced login.

When the config contains `requireLogin: true` or `GARDEN_REQUIRE_LOGIN_OVERRIDE=true` and a user is not logged in to their cloud instance, the user will be notified that they need to be logged in. See example below:

Screenshot 2023-12-12 at 12 18 35

Fixes #

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -16,7 +16,7 @@ import { shutdown, getPackageVersion } from "../util/util.js"
 import type { Command, CommandResult, BuiltinArgs } from "../commands/base.js"
 import { CommandGroup } from "../commands/base.js"
 import type { GardenError } from "../exceptions.js"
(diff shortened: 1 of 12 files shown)


stefreak:
We have implemented login requirement now. See also https://docs.garden.io/bonsai-0.13/guides/deprecations

PR #5504, opened 2023-12-04
fix(git): fix `exclude` files handling in `subtree` Git repo scan mode

### LINKED ISSUE #5317: Unnecessary config version change can happen when excluded files' contents are modified
## Bug

Container modules get rebuilt when the [exclude](https://docs.garden.io/v/acorn-0.12/reference/module-types/container#exclude) files are modified.

From the docs:
>Files that match these paths or globs are excluded when computing the version of the module, when responding to filesystem watch events, and when staging builds.

However, making changes in the excluded files' contents can cause some unnecessary action/module version changes. In that case, the module is rebuilt when it should not be. But the built images have the same digest, therefore no new images are pushed to the registry, instead, the already existing image in the registry is tagged with yet another version.

### Issue

If a file's content from the `excluded: ` list has been modified, it causes the module version change, despite it should be ignored. Here the `excluded: ` config entry is not modified. 

This happens only with `GARDEN_GIT_SCAN_MODE=subtree`.

### Current Behavior

An excluded file is not excluded by the Git repo scanner.
Thus, the changes made in the excluded file's content cause the change of module version and, hence, trigger a new build job.

### Expected Behavior

An excluded file should not be returned from the Git repo scan.
So, the changes made in the excluded files should not affect the module version.

### Reproducible example

```yaml
kind: Project
name: repro
environments:
  - name: local
  - name: remote
    defaultNamespace: ${var.userId}
providers:
  - name: local-kubernetes
    environments: [local]
variables:
  userId: ${local.username}

---
kind: Module
type: container
name: name
image: name-image
exclude:
  - ./dir1
  - ./file1
```

1. Run `garden build`
2. Change any file in the excluded `./dir1`
3. Run `garden build again`

### Suggested solution(s)

Fix the `GitHandler.getFiles()` method to properly exclude all files and directories listed in the `exclude: ` config entry.

### Workaround

Not found.

### Your environment

This is an environment-agnostic issue. It comes from the Core framework and version calculation logic.

`garden version 0.13.18`

See also #5519.

vvagaytsev:

Fixes #5317

Based on test fixes made in #5472.

### DIFF
--- a/core/src/vcs/git.ts
+++ b/core/src/vcs/git.ts
@@ -81,13 +81,11 @@ interface GitSubTreeIncludeExcludeFiles extends BaseIncludeExcludeFiles {
 const getIncludeExcludeFiles: IncludeExcludeFilesHandler<GetFilesParams, GitSubTreeIncludeExcludeFiles> = async (
   params: GetFilesParams
 ) => {
-  let { include, exclude } = params
-
-  if (!exclude) {
-    exclude = []
-  }
+  let include = params.include
 
   // Make sure action config is not mutated.
+  let exclude = !params.exclude ? [] : [...params.exclude]
+
   // It looks like paths with redundant '.' and '..' parts
   // do not work well along with --exclude and --glob-pathspecs flags.
   exclude = [...exclude.map(normalize), "**/.garden/**/*"]
--- a/core/src/vcs/git.ts
+++ b/core/src/vcs/git.ts
@@ -7,7 +7,7 @@
  */
 
 import { performance } from "perf_hooks"
-import { isAbsolute, join, posix, relative, resolve } from "path"
+import { isAbsolute, join, normalize, posix, relative, resolve } from "path"
 import { isString } from "lodash-es"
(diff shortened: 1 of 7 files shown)


core/src/vcs/git.ts
```diff
@@ -66,6 +74,38 @@ export interface GitCli {
...
+
+const getIncludeExcludeFiles: IncludeExcludeFilesHandler<GetFilesParams, GitSubTreeIncludeExcludeFiles> = async (
+  params: GetFilesParams
+) => {
+  let include = params.include
+
+  // Make sure action config is not mutated.
+  let exclude = !params.exclude ? [] : [...params.exclude]
+
+  // It looks like paths with redundant '.' and '..' parts
```
TimBeyer, line 89: Can you expand on this a bit in the comment?
I'm not sure it's easy to understand for someone just coming by and reading this code which cases we're fixing with the `normalize`.

TimBeyer (approved):

PR #5505, opened 2023-12-04
fix: adding an error handler to the segment client

mkhq:

The segment client has a new API in the updated version. It calls an errorHandler function when there are network level failures such as DNS lookup errors. This change logs the error description on the debug level instead of a full error output on the console. 

Fixes #


10ko:
I wonder if we should implement a retry? Or are we ok if we loose some data?

mkhq:
> I wonder if we should implement a retry? Or are we ok if we loose some data?

There is a retry mechanism in the `flush` method. This is a band-aid on the existing library, we should replace this with the newer segment client.

shumailxyz (approved):

PR #5508, opened 2023-12-04
fix(core): avoid unnecessary config version changes

### LINKED ISSUE #5501: Version hash is dependent on repository source folder name
## Bug

I have the suspicion that ever since https://github.com/garden-io/garden/releases/tag/0.12.61 and the introduction of the fix for https://github.com/garden-io/garden/issues/4661 the version depends on the *path* of the source repository.

### Current Behavior

The version hash generated to determine if an action needs to be executed again *seems* to depend on the source folder name of the checked out remote repository.

### Expected behavior

The version hash should only depend on the contents of the repository, not on the folder name.

### Reproducible example
```
$> cat umbrella/project.garden.yml

apiVersion: garden.io/v1
kind: Project
name: umbrella

defaultEnvironment: local

environments:
  - name: local

providers:
  - name: local-kubernetes
    environments:
      - local
  - name: jib
 
sources:
  - name: app
    repositoryUrl: https://github.com/dadadom/app-testing#master

$> garden version
garden version: 0.13.21

$>  git clone git@github.com:dadadom/app-testing.git app1
...
$>  git clone git@github.com:dadadom/app-testing.git app2
...

$> garden --root umbrella link source app app1

$> garden --root umbrella build --log-level=debug
...
Execing ... -Djib.container.args=GARDEN_MODULE_VERSION=v-9c6dacb57c ...
...

$> garden --root umbrella unlink source app
$> garden --root umbrella link source app app2
$> garden --root umbrella build --log-level=debug
...
Execing ... -Djib.container.args=GARDEN_MODULE_VERSION=v-a469a92ed9 ...
...
```

### Suggested solution(s)

As far as I understand the system, the "source" of the files should be irrelevant as long as the contents and locations of the files within a repository don't change.

### Additional context

For us the issue is critical as we re-execute all gradle jib builds for every production deployment although the exact same image was already built for the staging environment.
In our CI, the repository where garden checks out the sources into is different on every run (not sure how the folder names are calculated).

Looking back a few months I saw that this did not happen with Garden 0.12.56 and only happened when we (directly) upgraded to Garden 0.12.61, hence my suspicion about the mentioned release/issue.

### Your environment
Happens on our CI (linux/amd64), but the example above was executed on an M1 MacBook.

### LINKED ISSUE #5519: Unnecessary config version change can happen when the list of excluded paths is modified
## Bug

Container modules get rebuilt when the [exclude](https://docs.garden.io/v/acorn-0.12/reference/module-types/container#exclude) files are modified.

From the docs:
>Files that match these paths or globs are excluded when computing the version of the module, when responding to filesystem watch events, and when staging builds.

However, making changes in the excluded paths list can cause some unnecessary action/module version changes. In that case, the module is rebuilt when it should not be. But the built images have the same digest, therefore no new images are pushed to the registry, instead, the already existing image in the registry is tagged with yet another version.

### Issue

If the `excluded: ` list is modified it's natural to expect some changes in the build context and the module version. However, the modification of the `exclude: ` should not affect the module version if it does not change the actual build context. For instance, if a non-existing file is added to the `excluded: ` list.

### Current Behavior

Adding a non-existing file the the `excluded: ` list triggers the module version change and a new build job.

### Expected Behavior

The module version should remain the same.

### Reproducible example

```yaml
apiVersion: garden.io/v1
kind: Project
name: repro
environments:
  - name: local
  - name: remote
    defaultNamespace: ${var.userId}
providers:
  - name: local-kubernetes
    environments: [local]
variables:
  userId: ${local.username}

---
kind: Build
type: container
name: name
exclude:
  - ./dir1
  - ./file1
```

1. Run `garden build`
2. Add a non-existing file to the `excluded: ` list, e.f. `./dir-that-do-not-exist`
3. Run `garden build again`

### Suggested solution(s)

Module or action config hash is computed based on 2 entities:
- configuration hash
- Git tree version hash

The `include` and `exclude` config entries are resolved while the Git tree version calculation. Thus, those can be excluded from the config hash calculation.

Fix the action config hash calculation to omit `include` and `exclude` config entries. Consider omitting the other config fields that are not important for versioning and can have the same effect. That can help to fix #5501.

See `BaseAction.configVersion()`.

The module version calculation already excludes the necessary config fields, see function `hashModuleVersion`.

### Workaround

Not found.

### Your environment

This is an environment-agnostic issue. It comes from the Core framework and version calculation logic.

`garden version 0.12.66`
`garden version 0.13.18`

edvald:

Fixes #5501
Fixes #5519

### DIFF
--- a/core/src/actions/base.ts
+++ b/core/src/actions/base.ts
@@ -543,17 +543,18 @@ export abstract class BaseAction<
     return this._treeVersion
   }
 
-  @Memoize()
-  private stringifyConfig() {
-    return stableStringify(omit(this._config, "internal"))
-  }
-
   /**
    * The version of this action's config (not including files or dependencies)
    */
   @Memoize()
   configVersion() {
-    return versionStringPrefix + hashStrings([this.stringifyConfig()])
+    // We omit a few fields here that should not affect versioning directly:
+    // - The internal field (basePath, configFilePath etc.) are not relevant to the config version
+    // - The source, include and exclude stanzas are implicitly factored into the tree version
+    // - The description field is just informational, shouldn't affect execution
+    // - The disabled flag is not relevant to the config version, since it only affects execution
+    const configToHash = omit(this._config, "internal", "source", "include", "exclude", "description", "disabled")
+    return versionStringPrefix + hashStrings([stableStringify(configToHash)])
   }
 
   /**
--- a/core/src/actions/base.ts
+++ b/core/src/actions/base.ts
@@ -895,10 +895,12 @@ export function actionIsDisabled(config: ActionConfig, _environmentName: string)
 
 /**
  * We omit a few fields here that should not affect versioning directly:
- * - The internal field (basePath, configFilePath etc.) are not relevant to the config version
- * - The source, include and exclude stanzas are implicitly factored into the tree version
- * - The description field is just informational, shouldn't affect execution
- * - The disabled flag is not relevant to the config version, since it only affects execution
+ * - The internal field (basePath, configFilePath etc.) are not relevant to the config version.
+ * - The source, include and exclude stanzas are implicitly factored into the tree version.
+ *   Those are handled separately in {@link VcsHandler},
+ *   see {@link VcsHandler.getTreeVersion} and {@link VcsHandler.getFiles}.
+ * - The description field is just informational, shouldn't affect execution.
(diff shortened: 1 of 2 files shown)


stefreak:
Does this also happen to fix #5501?

vvagaytsev:
@stefreak it looks like it does.

TimBeyer (approved):

PR #5521, opened 2023-12-05
chore: remove nock hack in test setup

vvagaytsev:
The hack was done for Node 18. The current Node version is 21.

Fixes #

### DIFF
--- a/core/test/setup.ts
+++ b/core/test/setup.ts
@@ -22,14 +22,6 @@ initTestLogger()
 // Global hooks
 export const mochaHooks = {
   async beforeAll() {
-    // override fetch to handle node 18 issue when using nock
-    // https://github.com/nock/nock/issues/2336
-    // TODO: remove when we move to node 20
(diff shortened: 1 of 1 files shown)


vvagaytsev:
> From the issue it didn't sound like it's fixed since this issue is still open: [nock/nock#2183](https://github.com/nock/nock/issues/2183)
> 
> How come this still seems to work without then?

I don't know. All tests are green. @mkhq do you know more context on this?

stefreak:
I think if the tests pass without the hack we can merge this– it means the hack is apparently not needed anymore despite the issue being still open.

mkhq:
Since the tests pass, lets merge this. I think we might be skipping some tests so this would be needed when we enable the tests again. Would be worth looking into using native fetch (undici) and https://undici.nodejs.org/#/docs/api/MockAgent instead of nock.

mkhq (approved):

PR #5522, opened 2023-12-06
fix(sync-mode): showing undefined when sync status is not-deployed

mkhq:

The sync status command would show strings with undefined when the status was 'not-deployed'.

```
The api Deploy action has 1 sync configured:
  → Sync from .../api to /app/api undefined not deployed
  → Mode: one-way-replica
```

Fixes #

### DIFF
--- a/core/src/commands/sync/sync-status.ts
+++ b/core/src/commands/sync/sync-status.ts
@@ -192,6 +192,7 @@ export async function getSyncStatuses({
         "active": "is",
         "failed": "has",
         "not-active": "is",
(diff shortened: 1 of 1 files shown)


vvagaytsev (approved):

vvagaytsev:
Looks good, thank you! 👍 

Currently, the `verbMap` does not cover all the possible values of the `SyncStatus` type. Should we fix that in this PR as well?

vvagaytsev (approved):
🚀

PR #5525, opened 2023-12-06
fix(commands): print section and tags in color

### LINKED ISSUE #4725: 0.13: [Bug]: Service logs not color coded
### Current Behavior

The garden logs command does not color-code service logs when using the `--show-tag` option. In garden acorn (0.12) running the logs command with the `--show-container` option in a garden project resulted in color coded logs based on the container which made it easier to watch logs for several services at the same time. Since `--show-container` does not exist in bonsai any longer it would be great to get color-coding for the `--show-tags` option or the services per se.

### Reproducible example

Run this command with garden acorn:

```
garden logs -f --timestamps --show-container --hide-service
```

Run this command with garden bonsai:

```
garden logs -f --timestamps --show-tags --hide-service
```

### Your environment

* **OS:** macOS

`garden version`
0.13.5 (bonsai) and 0.12.61-edge-acorn-7eac8b701 (acorn)

eysi09:

This fixes a regression from the Bonsai release where log line sections weren't printed in color.

Fixes #4725 

### DIFF
--- a/core/src/monitors/logs.ts
+++ b/core/src/monitors/logs.ts
@@ -8,7 +8,7 @@
 
 import chalk from "chalk"
 import hasAnsi from "has-ansi"
-import { every, some } from "lodash-es"
+import { every, repeat, some } from "lodash-es"
 import { Stream } from "ts-stream"
 import type { DeployAction } from "../actions/deploy.js"
 import type { Resolved } from "../actions/types.js"
@@ -31,6 +31,7 @@ let colorMap: { [name: string]: string } = {}
 let colorIndex = -1
 // If the container name should be displayed, we align the output wrt to the longest container name
 let maxDeployName = 1
+const isoStringLength = new Date().toISOString().length
 
 interface LogMonitorParams extends MonitorBaseParams {
   action: Resolved<DeployAction>
@@ -200,8 +201,7 @@ export class LogMonitor extends Monitor {
   }
 
   private formatLogMonitorEntry(entry: DeployLogEntry) {
-    const style = chalk[LogMonitor.getColorForName(entry.name)]
-    const sectionStyle = style.bold
+    const sectionStyle = chalk[LogMonitor.getColorForName(entry.name)].bold
     const serviceLog = entry.msg
     const entryLevel = entry.level || LogLevel.info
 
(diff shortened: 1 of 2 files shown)


core/src/monitors/logs.ts
```diff
@@ -200,16 +201,15 @@ export class LogMonitor extends Monitor {
   }
 
   private formatLogMonitorEntry(entry: DeployLogEntry) {
-    const style = chalk[LogMonitor.getColorForName(entry.name)]
-    const sectionStyle = style.bold
+    const sectionStyle = chalk[LogMonitor.getColorForName(entry.name)].bold
```
vvagaytsev, line 204: Can `chalk[LogMonitor.getColorForName(entry.name)]` be `undefined` under any circumstances?

core/src/monitors/logs.ts
```diff
@@ -23,11 +23,17 @@ import type { DeployLogEntry } from "../types/service.js"
...
 import { styles } from "../logger/styles.js"
+import type { PickFromUnion } from "../util/util.js"
 
-export const logMonitorColors = ["green", "cyan", "magenta", "yellow", "blueBright", "blue"]
+type LogMonitorColor = PickFromUnion<
+  keyof typeof chalk,
+  "green" | "cyan" | "magenta" | "yellow" | "blueBright" | "blue"
+>
+
+export const logMonitorColors: LogMonitorColor[] = ["green", "cyan", "magenta", "yellow", "blueBright", "blue"]
```
vvagaytsev, line 33: We can avoid the color name repetition here, but it looks a bit more tricky and maybe awkward:
```
const selectedLogMonitorColors = ["green", "cyan", "magenta", "yellow", "blueBright", "blue"] as const
type LogMonitorColor = PickFromUnion<
  keyof typeof chalk,
  typeof selectedLogMonitorColors[number]
>
export const logMonitorColors: LogMonitorColor[] = [...selectedLogMonitorColors]
```

In that case, we can define the list of colors only once.

core/src/monitors/logs.ts
```diff
@@ -201,7 +207,8 @@ export class LogMonitor extends Monitor {
   }
 
   private formatLogMonitorEntry(entry: DeployLogEntry) {
-    const sectionStyle = chalk[LogMonitor.getColorForName(entry.name)].bold
+    const sectionColor = chalk[LogMonitor.getColorForName(entry.name)]
+    const sectionStyle = (sectionColor || styles.primary).bold
```
vvagaytsev, line 211: `LogMonitor.getColorForName(entry.name)` never returns `undefined`.
With the new type introduced above, we can see that the section style always has the type `Chalk` and _should_ never be `undefined`.

But, the extra type-safety clause `|| styles.primary` looks good in any way.

PR #5526, opened 2023-12-06
fix(git): fix `exclude` filter in `repo` Git scan mode

vvagaytsev:
This PR aligns the `exclude` file filter behavior between `GitRepoHandler` and `GitHandler`.
Also, it adds a few more test cases to cover more scenarios.

The behavior of `GitHandler` was taken as a reference, because it's been there for a long time before we switched the default repo scanner to `GitRepoHandler` in #5399 ([0.13.20](https://github.com/garden-io/garden/releases/tag/0.13.20)).

Fixes _some_ of the [`GitRepoHandler` test failures](https://app.circleci.com/pipelines/github/garden-io/garden/22567/workflows/4c86af6d-1ab2-4f3e-9502-1d8460251de2/jobs/453884) that were introduced in #5472. Those were enabled only for `GitHandler` that passed all the tests.

### DIFF
--- a/core/src/vcs/git-repo.ts
+++ b/core/src/vcs/git-repo.ts
@@ -16,7 +16,7 @@ import { normalize, sep } from "path"
 
 const { pathExists } = fsExtra
 
-type ScanRepoParams = Pick<GetFilesParams, "log" | "path" | "pathDescription" | "failOnPrompt">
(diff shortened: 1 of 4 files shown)


stefreak (approved):
Thank you for digging into this! Great work @vvagaytsev ❤️

PR #5535, opened 2023-12-10
improvement(ephemeral-kubernetes): tweak text in error message

eysi09:

The error message contained the hard coded string 'Garden Cloud'. We should instead use the 'getCloudDistributionName' helper.

In this case we can also just skip a reference to the distro name to keep things simple.

Fixes #


shumailxyz (approved):
Nice tweaking!

eysi09:
> Nice tweaking!

It's what I do best

PR #5537, opened 2023-12-11
docs: clean up Use Cases

worldofgeese:
Until we build out learning paths centered around use cases, it's best the docs are somewhat hermetic in their approach. This changes the docs to be less tutorial and more resource page for people looking to make use of them.

### DIFF
--- a/docs/use-cases/jumpstart-idp.md
+++ b/docs/use-cases/jumpstart-idp.md
@@ -19,32 +19,26 @@ If you're already familiar with Garden and just want to get going, click any of 
 
 Navigate to [Examples](#examples) for a selection of pre-configured stacks you can use to quickly explore relevant features.
 
-## Prerequisites
+## Requirements
 
-Before you proceed, make sure you have gone through the following steps:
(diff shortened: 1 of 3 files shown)


eysi09:
I'll some notes here since not all of them apply to the changes proper. Most of these apply to all guides.

- In the `Local development` guide we need to either remove "With Garden for local development you can:" or add bullets.
- The "Requirements" section and the paragraph below don't really apply in this context imo. Instead we just link to instructions under "Resources".
- The "Now that you've installed" Garden paragraph also doesn't fit in this context.
- We should have a callout for encouraging people to join the community and advertise the chatbot, but we can skip the "If you encounter a bug..." part. Feels unneeded here.

worldofgeese:
This should take care of all open requests. I'm partial to Local Development with Remote Clusters if only because Production-like development environments feels less pointed to me and I'd like to maintain focused clarity.

eysi09 (approved):
LGTM! Thanks so much 🙏

PR #5538, opened 2023-12-11
improvement(dashboard): better warning logs

eysi09:

This commit adds a message about the dashboard to all Garden commands. Previously we only showed it for the 'dev' command.

The message can be disabled.

It also consistently logs a warning if a user is logged in across all cloud distros.

Fixes #

### DIFF
--- a/core/src/cli/cli.ts
+++ b/core/src/cli/cli.ts
@@ -35,11 +35,18 @@ import {
   checkRequirements,
   renderCommandErrors,
   cliStyles,
+  getDashboardInfoMsg,
 } from "./helpers.js"
 import type { ParameterObject, GlobalOptions, ParameterValues } from "./params.js"
 import { globalOptions, OUTPUT_RENDERERS } from "./params.js"
 import type { ProjectConfig } from "../config/project.js"
-import { ERROR_LOG_FILENAME, DEFAULT_GARDEN_DIR_NAME, LOGS_DIR_NAME, gardenEnv } from "../constants.js"
+import {
+  ERROR_LOG_FILENAME,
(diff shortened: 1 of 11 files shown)


core/src/cli/cli.ts
```diff
@@ -319,6 +323,16 @@ ${renderCommands(commands)}
...
         } else {
           garden = await wrapActiveSpan("initializeGarden", () => this.getGarden(workingDir, contextOpts))
+          const isLoggedIn = !!cloudApi
+          const isCommunityEdition = garden.cloudDomain === DEFAULT_GARDEN_CLOUD_DOMAIN
+
+          if (!isLoggedIn && isCommunityEdition) {
+            await garden.emitWarning({
+              key: "web-app",
+              log,
+              message: "\n" + getDashboardInfoMsg(),
```
vvagaytsev, line 333: Do we need a new line here? The whole block is repeated in [`serve.ts`](https://github.com/garden-io/garden/pull/5538/files#diff-ebdd3a25b6cc4795bac9dac4a97375cdfb935d9b0d8eea64e9989336ff93d30aR161-R168). The only diff is the new line character.

We could extract a function that encapsulates the local vars `isLoggedIn` and `isCommunityEdition` to avoid code duplication. But, we have only 2 strikes here. The rule of three is not yet in place :)

eysi09: We actually do need the line to make the logs render a little better. 

IMO we can live with this for now.

vvagaytsev: Sure! :)

vvagaytsev (approved):
Thank you! 👍

PR #5539, opened 2023-12-11
docs: add eksctl guide

worldofgeese:
This gives specific advice for creating an EKS cluster ready to be used for in-cluster image building when an EKS cluster is created with Amazon's official eksctl tool.


twelvemo (approved):
🚀

PR #5551, opened 2023-12-12
improvement(mutagen): use faux SSH command to use original Mutagen

### LINKED ISSUE #5420: Move away from Mutagen fork

vvagaytsev:
Garden currently uses a custom fork of Mutagen for its `kubectl exec` transport. This is the last remaining blocker to get Garden back on the official Mutagen releases.

At the moment, Mutagen doesn't support custom transports, but it does allow a custom ssh executable to be specified, which presents an opportunity for using a faux SSH command that calls down to `kubectl exec` using parameters smuggled through the SSH hostname specification. In this case, they are smuggled in a Base64-encoded JSON object (with some slight encoding tweaks to pass through Mutagen's local path detection heuristic).

The faux SSH implementation added in the original PR #3102 will almost certainly be useable (with minor tweaks) as a custom transport implementation once support exists in Mutagen.

The faux SSH implementation has been hosted in its own repo at https://github.com/garden-io/mutagen-faux-ssh.

This is the rebased version of #3102 and it allows to move away from the Mutagen fork.

Fixes #5420

Some questions from the original PR (#3102):

> This PR is fairly close to complete, but there are a few remaining questions:
> 
> 1. Is this approach satisfactory overall?
> 2. Should the faux `ssh` command be rewritten in TypeScript?
> 3. How should the faux `ssh` command be source-controlled and distributed?
> 4. Are there any transition considerations for active `garden dev` commands?
> 
> Regarding (1), I think this is the best approach because it gets Garden back on the official Mutagen releases with the minimal amount of effort and the implementation strategy is very similar to what custom transports will eventually look like in Mutagen, meaning that the faux SSH command will probably be convertible to a `mutagen-transport-garden` (name TBD) command later on.
>
> Regarding (2), that's a policy decision I'd leave to Garden.  The critical aspect is just the proper signal and standard input closure forwarding.
> 
> Regarding (3), that's probably a good point for discussion.  I would imagine it would be distributed/downloaded in a similar manner to the `mutagen` executable, but I don't know the best location to perform its build or serve its binaries.  There's also the consideration of its versioning w.r.t. Garden, especially if the format of the JSON changes, but that's likely to see very little churn once implemented, so manually managing its versioning as some external project might be suitable.  Once this is decided and implemented, the `MUTAGEN_SSH_PATH` environment variable will need to be set for `mutagen daemon run`; a TODO has been indicated for this in the code.
> 
> Regarding (4), the simple answer is that the project would probably just need to be restarted entirely.  If the same `MUTAGEN_DATA_DIRECTORY` is used, then the standard Mutagen daemon will just reject and ignore the prior `exec`-based sessions, but testing of this would be good before shipping.

Comments from @vvagaytsev:
1. Yes, that approach was good and simple.
2. No, the original implementation in Go worked fine.
3. It's now hosted at https://github.com/garden-io/mutagen-faux-ssh.
4. Agree, all running Mutagen processes must be killed before using the new transport and original Mutagen:
4.1. All sync must be stopped.
4.2. All sync monitors must be terminated.
4.3. Mutagen daemon must be stopped.

### DIFF
--- a/core/src/mutagen.ts
+++ b/core/src/mutagen.ts
@@ -866,20 +866,25 @@ export function parseSyncListResult(res: ExecaReturnValue): SyncSession[] {
   return parsed
 }
 
-export const mutagenVersion = "0.15.0"
+export const isNativeMutagenEnabled = () => !!gardenEnv.MUTAGEN_SSH_PATH
+
+const mutagenVersionLegacy = "0.15.0"
+const mutagenVersionNative = "0.15.0"
+
+export const mutagenVersion = isNativeMutagenEnabled() ? mutagenVersionNative : mutagenVersionLegacy
 
 export function mutagenCliSpecLegacy(): PluginToolSpec {
   return {
     name: "mutagen",
-    version: mutagenVersion,
-    description: `The mutagen synchronization tool, v${mutagenVersion}`,
+    version: mutagenVersionLegacy,
+    description: `The mutagen synchronization tool, v${mutagenVersionLegacy}`,
     type: "binary",
     _includeInGardenImage: false,
     builds: [
       {
         platform: "darwin",
         architecture: "amd64",
-        url: `https://github.com/garden-io/mutagen/releases/download/v${mutagenVersion}-garden-1/mutagen_darwin_amd64_v${mutagenVersion}.tar.gz`,
+        url: `https://github.com/garden-io/mutagen/releases/download/v${mutagenVersionLegacy}-garden-1/mutagen_darwin_amd64_v${mutagenVersion}.tar.gz`,
         sha256: "370bf71e28f94002453921fda83282280162df7192bd07042bf622bf54507e3f",
         extract: {
           format: "tar",
@@ -889,7 +894,7 @@ export function mutagenCliSpecLegacy(): PluginToolSpec {
       {
         platform: "darwin",
         architecture: "arm64",
-        url: `https://github.com/garden-io/mutagen/releases/download/v${mutagenVersion}-garden-1/mutagen_darwin_arm64_v${mutagenVersion}.tar.gz`,
+        url: `https://github.com/garden-io/mutagen/releases/download/v${mutagenVersionLegacy}-garden-1/mutagen_darwin_arm64_v${mutagenVersionLegacy}.tar.gz`,
         sha256: "a0a7be8bb37266ea184cb580004e1741a17c8165b2032ce4b191f23fead821a0",
         extract: {
           format: "tar",
@@ -899,7 +904,7 @@ export function mutagenCliSpecLegacy(): PluginToolSpec {
       {
         platform: "linux",
         architecture: "amd64",
-        url: `https://github.com/garden-io/mutagen/releases/download/v${mutagenVersion}-garden-1/mutagen_linux_amd64_v${mutagenVersion}.tar.gz`,
+        url: `https://github.com/garden-io/mutagen/releases/download/v${mutagenVersionLegacy}-garden-1/mutagen_linux_amd64_v${mutagenVersionLegacy}.tar.gz`,
         sha256: "e8c0708258ddd6d574f1b8f514fb214f9ab5d82aed38dd8db49ec10956e5063a",
         extract: {
           format: "tar",
@@ -909,7 +914,7 @@ export function mutagenCliSpecLegacy(): PluginToolSpec {
       {
         platform: "linux",
         architecture: "arm64",
-        url: `https://github.com/garden-io/mutagen/releases/download/v${mutagenVersion}-garden-1/mutagen_linux_arm64_v${mutagenVersion}.tar.gz`,
+        url: `https://github.com/garden-io/mutagen/releases/download/v${mutagenVersionLegacy}-garden-1/mutagen_linux_arm64_v${mutagenVersionLegacy}.tar.gz`,
         sha256: "80f108fc316223d8c3d1a48def18192e666b33a334b75aa3ebcc95938b774e64",
         extract: {
           format: "tar",
@@ -919,7 +924,7 @@ export function mutagenCliSpecLegacy(): PluginToolSpec {
       {
         platform: "windows",
         architecture: "amd64",
-        url: `https://github.com/garden-io/mutagen/releases/download/v${mutagenVersion}-garden-1/mutagen_windows_amd64_v${mutagenVersion}.zip`,
+        url: `https://github.com/garden-io/mutagen/releases/download/v${mutagenVersionLegacy}-garden-1/mutagen_windows_amd64_v${mutagenVersionLegacy}.zip`,
         sha256: "fdae26b43cc418b2525a937a1613bba36e74ea3dde4dbec3512a9abd004def95",
         extract: {
           format: "zip",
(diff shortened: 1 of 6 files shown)


vvagaytsev:
@xenoscopic I rebased #3102 on top of the latest main and tried out the faux SSH transport.

I used the path to the faux SSH executable binary as a value of `MUTAGEN_SSH_PATH`, was that correct? So the command I used was:
```console
# I compiled the `ssh.go` and moved it to `bin/` sub-directory
MUTAGEN_SSH_PATH="~/Repositories/garden/core/src/mutagen/faux-ssh/ssh/bin" gdev deploy --sync
```

That command failed with the error:
```console
Error: unable to connect to beta: unable to connect to endpoint: unable to dial agent endpoint: unable to create agent command: unable to set up SSH invocation: unable to identify 'ssh' command: unable to query file metadata: stat /Users/vladimirvagaytsev/Repositories/garden/core/src/mutagen/faux-ssh/ssh/ssh: not a directory
```

I tried it with the latest Mutagen 0.17.4 and got the same error. I also tried to grant `777` permissions to the faux SSH binary itself and its parent directory, but that did not help.

Could you give any clue on that, please? Thank you!

xenoscopic:
Hey @vvagaytsev, sorry for the slow response - I believe the issue here is that you'll want to set `MUTAGEN_SSH_PATH` to the directory containing the faux `ssh` executable, not the path of the executable itself.

vvagaytsev:
Thank you @xenoscopic! Now it works!

core/src/mutagen.ts
```diff
@@ -912,17 +1060,34 @@ export const mutagenCliSpec: PluginToolSpec = {
...
 
-export const mutagenCli = new PluginTool(mutagenCliSpec)
+export const mutagenFauxSsh = new PluginTool(mutagenFauxSshSpec)
+
+async function getMutagenSshPath(log: Log): Promise<string | undefined> {
+  if (!isNativeMutagenEnabled()) {
+    return undefined
+  }
+
+  const warnMessage = `This version of Garden uses a new file syncing machinery.
```
eysi09, line 1080: Let's just use a normal `.` instead of `!!!`. 

Also, we need to actually explain _how_ you stop Mutagen processes.

Finally, we should avoid using the term Mutagen. Most Garden don't know (and shouldn't have to know) what Mutagen is. Instead we can say that you need to stop active syncs and explain how that's done.

core/src/mutagen.ts
```diff
@@ -860,50 +872,186 @@ export function parseSyncListResult(res: ExecaReturnValue): SyncSession[] {
   return parsed
 }
 
-export const mutagenVersion = "0.15.0"
+export const isNativeMutagenEnabled = () => !gardenEnv.GARDEN_ENABLE_LEGACY_SYNC
```
eysi09, line 875: Do we need this function? 

The following is equally clear (if not clearer):

```
export const mutagenVersion = gardenEnv.GARDEN_ENABLE_LEGACY_SYNC ? mutagenVersionLegacy : mutagenVersionNative
```

This could also be wrapped into a function like `getMutagenVersion()`.

core/src/mutagen.ts
```diff
@@ -912,17 +1060,34 @@ export const mutagenCliSpec: PluginToolSpec = {
...
+        targetPath: "ssh.exe",
       },
     },
   ],
 }
 
-export const mutagenCli = new PluginTool(mutagenCliSpec)
+export const mutagenFauxSsh = new PluginTool(mutagenFauxSshSpec)
+
+async function getMutagenSshPath(log: Log): Promise<string | undefined> {
```
eysi09, line 1075: Can we add a doc string to this function explaining the different return values?

core/src/mutagen.ts
```diff
@@ -912,17 +1060,34 @@ export const mutagenCliSpec: PluginToolSpec = {
...
+  if (!isNativeMutagenEnabled()) {
+    return undefined
```
eysi09, line 1077: This is a bit of a nitpick but I think we should return `null` instead of `undefined` here.

eysi09 (changes requested):
Had a handful of minor comments but otherwise looks great!

core/src/mutagen.ts
vvagaytsev: Added the comment in d835e425aca0824501f3b99ca4e93f0b9d15da50.

core/src/mutagen.ts
vvagaytsev: That will cause some change in the interface `MutagenEnvValues` and the contract of the function `getMutagenEnv`.

Maybe, it's better to always require to pass sshPath explicitly be declaring it as `sshPath: string | null` instead of `sshPath?: string`. But we need to pass it only in a single place. The other calls to `getMutagenEnv` will have `null` or `undefined` value. That looks a bit awkward. That was the reason why the `sshPath` was introduced as an optional value of the `MutagenEnvValues`.

To be honest, I don't see a big semantic difference between `null` and `undefined` in this context. The `sshPath` must be defined only when we use the new sync machinery. Otherwise, it's not defined. So, it's undefined :)

core/src/mutagen.ts
vvagaytsev: Inlined it in 03218996421ef18d831b637d50a9e35aca7d73c8. Also flipped some if-statements to avoid negated conditions, for better readability :)

core/src/mutagen.ts
vvagaytsev: I made it simpler in 3350bbe7594e8ef94adaace8a524f383cecad611 :)
That would be the easiest way to shutdown the old sync machinery. Maybe we should put a similar instruction to the self-update command.

core/src/mutagen.ts
vvagaytsev: The Mutagen version can be resolved statically. We can keep it as it is.

core/src/mutagen.ts
```diff
@@ -513,6 +523,27 @@ export class Mutagen {
...
+              return (
+                // this happens when switching from the old sync to the new
+                msg.includes("ssh: could not resolve hostname") ||
+                // this happens in the opposite scenario
+                msg.includes("unknown or unsupported protocol")
+              )
+            }
+
+            if (isMutagenForkError(err)) {
+              log.warn(
```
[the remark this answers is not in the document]
vvagaytsev: Thanks for the comments @eysi09! I'll make sure to address those tomorrow. Doc link would be the best option 👍 

Regarding the error throwing, it will be thrown after 5 retries anyway, it will not restart anything. So, it makes sense to do it immediately. Good point!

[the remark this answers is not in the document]
vvagaytsev: Changed this in 5873c505837fc520ee77ea23e26023e06b7f5ae3 and 24fa50899d1e2e41c9702b240ccf9032e381ba19.

I'll create a new PR to update the sync guide.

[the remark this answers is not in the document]
vvagaytsev: Added a troubleshooting section in d06e241d971796e60aa4d48ad2f4c18069105150.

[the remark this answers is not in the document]
thsig: > It looks like the underlying syncing machinery was switched
I'd suggest using slightly less colourful prose here, maybe something like:
> It looks like you've changed to a different version of the sync daemon [...] and therefore the sync daemon needs to be restarted.

core/src/mutagen.ts
```diff
@@ -513,6 +524,30 @@ export class Mutagen {
...
+                dedent`
+                It looks like the underlying syncing machinery was switched via the ${GARDEN_ENABLE_NEW_SYNC_FEATURE_FLAG_NAME} env variable and the sync daemon needs to be restarted.
+                Please see our Troubleshooting docs for instructions on how to restart the daemon for your platform: ${styles.link(
+                  makeDocsLink("misc/troubleshooting#restarting-sync-daemon")
```
eysi09, line 545: This link is wrong and should point to the correction section of the code synchronization guide.

eysi09 (changes requested):

core/src/mutagen.ts
vvagaytsev: Thanks! Fixed that in a81a033e48e5533836c3eb9055f2c6ce6dde34bc.

core/src/mutagen.ts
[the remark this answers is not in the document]
vvagaytsev: Thanks! I updated the message in a81a033e48e5533836c3eb9055f2c6ce6dde34bc.

thsig (approved):

vvagaytsev:
A follow-up PR is #5744.

PR #5552, opened 2023-12-12
refactor(mutagen): always use `getMutagenEnv` helper to define Mutagen env

vvagaytsev:

Fixes #

A tiny predecessor for #5551.

### DIFF
--- a/core/src/mutagen.ts
+++ b/core/src/mutagen.ts
@@ -805,14 +805,25 @@ export function getMutagenDataDir(path: string, log: Log) {
   return shortPath
 }
 
-export function getMutagenEnv(dataDir: string, logLevel?: string) {
-  const env = { MUTAGEN_DATA_DIRECTORY: dataDir }
+const mutagenRequiredEnvVars = ["MUTAGEN_DATA_DIRECTORY"] as const
+type MutagenRequiredEnvVar = (typeof mutagenRequiredEnvVars)[number]
+type MutagenRequiredEnv = {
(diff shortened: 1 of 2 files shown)


core/src/mutagen.ts
```diff
@@ -808,10 +805,30 @@ export function getMutagenDataDir(path: string, log: Log) {
...
+  [k in MutagenRequiredEnvVar]: string
+}
+
+const mutagenOptionalEnvVars = ["MUTAGEN_LOG_LEVEL"] as const
+type MutagenOptionalEnvVar = (typeof mutagenOptionalEnvVars)[number]
+type MutagenOptionalEnv = {
+  [k in MutagenOptionalEnvVar]?: string
+}
+
+type MutagenEnv = MutagenRequiredEnv & MutagenOptionalEnv
```
shumailxyz, line 820: I'm wondering if we can merge these 2 types into 1 as we are currently only using `MUTAGEN_DATA_DIRECTORY ` and `MUTAGEN_LOG_LEVEL`. Do we need this flexibility and separate types for required/optional or its too verbose atm?

shumailxyz (approved):
Minor comment, but LGTM otherwise. Thanks :)

core/src/mutagen.ts
vvagaytsev: Good point, thank you! Simplified this in d82f9f97b5d309001ad166862b67eb6b2ab0b9ff.

shumailxyz (approved):
Awesome ✨

PR #5553, opened 2023-12-12
improvement(cloud): log 'not logged in' msg at info level for community

eysi09:

A previous commit (see #5538) logged this at the warning level which is relevant for the commercial edition but the community. This fixes that.

Fixes #

### DIFF
--- a/core/src/garden.ts
+++ b/core/src/garden.ts
@@ -1869,59 +1869,28 @@ export const resolveGardenParams = profileAsync(async function _resolveGardenPar
     const projectApiVersion = config.apiVersion
     const sessionId = opts.sessionId || uuidv4()
     const cloudApi = opts.cloudApi || null
-    const isCommunityEdition = !config.domain
-    const distroName = getCloudDistributionName(cloudApi?.domain || DEFAULT_GARDEN_CLOUD_DOMAIN)
-    const debugLevelCommands = ["dev", "serve", "exit", "quit"]
-    const cloudLogLevel = debugLevelCommands.includes(opts.commandInfo.name) ? LogLevel.debug : undefined
-    const cloudLog = log.createLog({ name: getCloudLogSectionName(distroName), fixLevel: cloudLogLevel })
-
-    let secrets: StringMap = {}
-    let cloudProject: CloudProject | null = null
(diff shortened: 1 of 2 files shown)


core/src/garden.ts
```diff
@@ -2026,6 +1995,87 @@ export const resolveGardenParams = profileAsync(async function _resolveGardenPar
...
+    cloudLog.success("Ready")
+  } else {
+    const msg = `You are not logged in. To use ${distroName}, log in with the ${styles.command(
+      "garden login"
+    )} command.`
+    if (isCommunityEdition) {
+      cloudLog.info(msg)
+      cloudLog.info(`Learn more at: ${styles.underline(`${DOCS_BASE_URL}/using-garden/dashboard`)}`)
+    } else {
+      cloudLog.warn(
```
vvagaytsev, line 2070: It seems that this can be just `log.warn(msg)`.

eysi09: Actually this is intentional because the `cloudLog` has a debug level for e.g. the `dev` command. 

This basically means that we don't print this line when you start the dev command because it looks off in that context but we do print it for commands you run _within_ the dev command.

eysi09: ...for other commands we always print it.

vvagaytsev: Ok, got it, thank you!
Nit. Then we can inline the local vat `msg` - it's created outside the block where it's used, so it's scoped can be narrowed. Also, it's used only once, so it can be inlined :)

eysi09: Ah whoops, it's supposed to use the `msg` variable. So same message, different level. Good catch.

eysi09: Just amended the original commit and forced pushed the changes since it's a one liner fix.

vvagaytsev: Ah, but we still can use the msg variable here, right? By using `cloudLog.warn(msg)`.

vvagaytsev (approved):
Thank you! ✅

core/src/garden.ts
eysi09: Yeah exactly. That's the change I made. Thanks for flagging.

PR #5584, opened 2023-12-19
test: verify toolspecs in ci instead of manually

stefreak:
This PR changes test-framework and test-plugins to save `.test-tools-verification-cache` directory and changes binary verification tests to always run as part of unit tests.

Also changes the tests to download tools into this directory and skip downloading if the url has been downloaded already and the hash still matches.


vvagaytsev:
We also have 2 more scripts for external (i.e. non-embedded) plugins:
* https://github.com/garden-io/garden/blob/main/plugins/jib/test/verify-jib-tools-binary-hashes.ts
* https://github.com/garden-io/garden/blob/main/plugins/pulumi/test/verify-pulumi-binary-hashes.ts
* https://github.com/garden-io/garden/blob/main/plugins/terraform/test/verify-terraform-binary-hashes.ts

Can we remove them or do we need some extra changes to rigger those tests?

stefreak:
@vvagaytsev those should just work as they are using the same helpers. I removed the code that automatically skipped them, as we want to run them in CI

vvagaytsev (approved):
Great job! Thank you! 💯 🚀

PR #5587, opened 2023-12-21
Fix deploy modes

stefreak:

Fixes #


stefreak:
We went in a different direction here: #5589

PR #5589, opened 2023-12-21
fix(core): inherit Build action mode from dependant Deploy action

stefreak:
Override the Build action mode with the Deploy action mode, if it depends on the build.

This enables using ${this.mode} in the build action as a proxy to the
Deploy action mode, simplifying configuration for many users.

In previous versions of Garden all actions went into `sync` Mode when
users ran the command `garden deploy --sync` without explicitly listing
Deploy actions due to the fact that we defaulted to the `*` Minimatch
pattern that also matched build actions.

This commit ensures that if you depend on a Build from a Deploy action,
and that Deploy action is in `sync` mode, the Build action inherits the
mode.

If multiple Deploy actions depend on the same build, and have different
modes, the non-default mode wins. If multiple Deploy actions use
different non-default modes, we print a warning.

In the future, we can make this even more reliable by injecting
different invariants of build actions into the stack graph for each
Deploy action, if the build action is used by multiple Deploy actions
and they are in different modes.

Fixes #

### DIFF
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -59,7 +59,7 @@ import { mergeVariables } from "./common.js"
 import type { ConfigGraph } from "./config-graph.js"
 import { MutableConfigGraph } from "./config-graph.js"
 import type { ModuleGraph } from "./modules.js"
-import type { MaybeUndefined } from "../util/util.js"
+import { isTruthy, type MaybeUndefined } from "../util/util.js"
 import minimatch from "minimatch"
 import type { ConfigContext } from "../config/template-contexts/base.js"
 import type { LinkedSource, LinkedSourceMap } from "../config-store/local.js"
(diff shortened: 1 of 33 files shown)


core/src/graph/actions.ts
```diff
@@ -138,25 +138,83 @@ export const actionConfigsToGraph = profileAsync(async function actionConfigsToG
...
+  // Preprocess all remaining actions (Deploy actions are preprocessed above)
+  // We are preprocessing actions in two batches so we can infer the mode of Build actions that are used by Deploy actions. See the comments above.
+  await preprocessActions((config) => config.kind !== "Deploy")
 
   // Optimize file scanning by avoiding unnecessarily broad scans when project is not in repo root.
   const preprocessedConfigs = Object.values(preprocessResults).map((r) => r.config)
   const allPaths = preprocessedConfigs.map((c) => getSourcePath(c))
   const minimalRoots = await garden.vcs.getMinimalRoots(log, allPaths)
 
+  // If a Build uses this.mode and is used as a build or dependency of a Deploy action that has a non-default mode,
```
thsig, line 215: Let's delete this comment (it's a leftover from earlier).

core/test/unit/src/actions/action-configs-to-graph.ts
```diff
@@ -840,6 +840,51 @@ describe("actionConfigsToGraph", () => {
     expect(action.mode()).to.equal("local")
   })
 
+  it("deploy action mode overrides the build action mode, if deploy action depends on build action", async () => {
```
thsig, line 843:
```suggestion
  it("deploy action mode overrides the mode of a dependency build action", async () => {
```

thsig (dismissed):
Looks good to me! Added a couple of text-related comments there.

We'll take a more proper look at all this in the new year.

thsig (approved):

PR #5660, opened 2024-01-23
fix(core): fix "repo" scan mode for remote actions

### LINKED ISSUE #5625: 0.13: [Bug]: Regression in git scan mode after 0.13.19 with remote repository URLs
### Current Behavior

We have most of our services defined in our Garden project repository, but the source points to the service repository itself, for example:

```
kind: Build
type: container
name: gateway
source:
  repository:
    url: git@github.com:xxxx/gateway#main
include: ["**/*"]
```

We run `garden update-remote all` to pull the contents of those remote repositories which causes Garden to check out the repository to a path like `.garden/sources/action/build.gateway--a3057b675c` within the current Garden project.

At container build time, Garden will sync the files from that directory to a build directory which is then picked up by mutagen. This is according to the include/exclude rules above. In 0.13.19 and earlier, this worked as expected:

```
ℹ git [debug]          → Scanning Build action gateway root at /private/tmp/b/.garden/sources/action/build.gateway--a3057b675c
  → Includes: **/*
  → Excludes: **/.garden/**/*

ℹ git [debug]          → Found 57 files in Build action gateway root /private/tmp/b/.garden/sources/action/build.gateway--a3057b675c

ℹ build.gateway [verbose] → Syncing sources (57 files)...
ℹ build.gateway [debug] → Syncing 57 files from .garden/sources/action/build.gateway--a3057b675c to .garden/build/gateway (and removing any extraneous files)
```

In 0.13.20 and later, Garden finds no files to sync:

```
ℹ build.gateway [debug] → Syncing 0 files from .garden/sources/action/build.gateway--a3057b675c to .garden/build/gateway (and removing any extraneous files)
```

(There's not much other usable output before this).

It appears to be because the git scan mode has changed from `subtree` to `repo`. When this is set to `subtree` again via environment variable, 0.13.20 and later work again as expected.

### Expected behavior

Checked out files from the remote repository are synced to the K8s cluster and built as a container. (These are actually internal details and should not really be a user requirement, however in this case it is the cause of the problem of the service not building in Garden).

### Reproducible example

See snippet above.

### Workaround

```
export GARDEN_GIT_SCAN_MODE=subtree
```

### Suggested solution(s)

It's hard to know what to suggest without knowing more about the internal workings of the git scan mode, which is really not a user-facing implementation detail of Garden. 

```
ℹ graph [silly]        → TreeCache: Setting value for key ["source","/private/tmp/b/.garden/sources/action/build.gateway--a3057b675c","include","ad948e60de"]
ℹ graph [debug]        → Found 0 files in module path, filtering by 1 include and 1 exclude globs
ℹ graph [silly]        → Include globs: **/*
ℹ graph [silly]        → Exclude globs: **/.garden/**/*
ℹ graph [debug]        → Found 0 files in module path after glob matching
```

This may be a long shot, but it seems like the default exclude glob would exclude the path to which the remote repositories are checked out. This doesn't seem correct.

### Your environment

* **OS:** macOS 14.2.1, Intel

`garden version`
0.13.20 introduces the regression, and I've been able to reproduce up until 0.13.23.

thsig:

Before this fix, remote actions (i.e. actions with `source.repository.url` set, not sources declared at the project level) didn't work correctly when using the new `repo` git scanning mode.

Also set the default git scanning mode back to `repo`, since this bug (which we've now fixed) was the reason we reverted to `subtree`.

The error arose because the base path wasn't being set to the locally cloned path of the remote source's repo early enough in the resolution process. This led to the repo scanning logic using the action's config directory instead of the cloned source directory as the repo root (thus ignoring the sources).

We've added a unit test that verifies that the base path is being set correctly when the action graph is initialized.

Fixes #5625.

### DIFF
--- a/core/src/constants.ts
+++ b/core/src/constants.ts
@@ -18,7 +18,7 @@ const extractedRoot = process.env.GARDEN_SEA_EXTRACTED_ROOT
 
 export const gitScanModes = ["repo", "subtree"] as const
 export type GitScanMode = (typeof gitScanModes)[number]
-export const defaultGitScanMode: GitScanMode = "subtree"
+export const defaultGitScanMode: GitScanMode = "repo"
 
 export const GARDEN_CORE_ROOT = !!extractedRoot
   ? resolve(extractedRoot, "src", "core")
--- a/core/src/garden.ts
+++ b/core/src/garden.ts
@@ -1573,7 +1573,7 @@ export class Garden {
   }
 
   /**
-   * Clones the project/module source if needed and returns the path (either from .garden/sources or from a local path)
+   * Clones the project/action/module source if needed and returns the path (either from .garden/sources or from a local path)
    */
   public async resolveExtSourcePath({
     name,
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -152,6 +152,7 @@ export const actionConfigsToGraph = profileAsync(async function actionConfigsToG
           garden,
           config,
           router,
+          linkedSources,
           log,
           mode,
         })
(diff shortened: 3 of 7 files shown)


vvagaytsev (approved):
Looks good! Thanks for fixing this! 👍

PR #5670, opened 2024-01-26
fix(template): do not partially resolve function arg objects with special keys

edvald:
This caused bad handling of e.g. $forEach objects in input variables for templated actions.


vvagaytsev (approved):
Looks good, thanks for fixing this! There were some tests failures, but those didn't look relevant. I rerun the CI pipeline.

PR #5671, opened 2024-01-29
fix(module-conversion): always include build deps

### LINKED ISSUE #5667: 0.13: [Bug]: Module conversion: Build dependency missing in Deploy action converted from Module service
### Current Behavior
Converted deploy action from module is missing build dependencies.

When running `garden deploy`, the `test-image` build action does not get triggerered.

### Expected behavior

`garden deploy` should also trigger the build `test-image`

### Reproducible example

```
# In foo.dockerfile
FROM nginx:latest
```

```
# In garden.yml
kind: Module
type: container
name: test-image
dockerfile: foo.dockerfile

---

kind: Module
type: container
name: test-deploy
build:
  dependencies: [test-image]
image: '${modules.test-image.outputs.deployment-image-id}'
services:
 - name: "test-deploy"
```

Then run `garden deploy`

### Workaround

Convert to actions

### Your environment

* **OS:** macOS
* **How I'm running Kubernetes:** local-kubernetes

`garden version` 0.13.23-edge-bonsai-ce02590d3

thsig:

Before this fix, build dependencies weren't being transferred to the generated actions in all cases.

This was a problem e.g. if using a `helm` module with a build dependency on a `container` module, whose built image was referenced in the chart (since the missing build dependency meant that the image wasn't necessarily ready by the time the deploy started).

This was fixed by simply ensuring that any explicit build dependencies (i.e. under `module.build.dependencies`) are included in the dependencies of any generated runtime actions (Deploys, Tests and Runs).

This reflects the 0.12-era semantics, where every test, service and task had a build dependency on the module's build step, which in turn respected any build dependencies declared on the module (under `build.dependencies`).

Fixes #5667

### DIFF
--- a/core/src/plugins/container/container.ts
+++ b/core/src/plugins/container/container.ts
@@ -328,7 +328,7 @@ export async function convertContainerModule(params: ConvertModuleParams<Contain
   )
   actions.push(...runtimeActions)
   if (buildAction) {
-    buildAction.dependencies = buildAction?.dependencies?.filter((d) => !volumeModulesReferenced.includes(d.name))
+    buildAction.dependencies = buildAction.dependencies?.filter((d) => !volumeModulesReferenced.includes(d.name))
   }
 
   return {
--- a/core/src/plugins/container/container.ts
+++ b/core/src/plugins/container/container.ts
@@ -93,7 +93,6 @@ export async function configureContainerModule({ log, moduleConfig }: ConfigureM
 
     for (const volume of spec.volumes) {
       if (volume.module) {
-        moduleConfig.build.dependencies.push({ name: volume.module, copy: [] })
         spec.dependencies.push(volume.module)
       }
     }
--- a/core/src/plugins/container/helpers.ts
+++ b/core/src/plugins/container/helpers.ts
@@ -393,7 +393,8 @@ const helpers = {
     // If we explicitly set a Dockerfile, we take that to mean you want it to be built.
     // If the file turns out to be missing, this will come up in the build handler.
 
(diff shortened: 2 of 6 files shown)


eysi09:
Does this account for the fact that build dependencies can be specified as:

```
build:
  dependencies: [foo]
```

and 

```
build:
  dependencies:
    - name: foo
```

I'm assuming that's normalized earlier in the flow but just wanted to confirm.

thsig:
> Does this account for the fact that build dependencies can be specified as:
> 
> ```
> build:
>   dependencies: [foo]
> ```
> 
> and
> 
> ```
> build:
>   dependencies:
>     - name: foo
> ```
> 
> I'm assuming that's normalized earlier in the flow but just wanted to confirm.

Good question—yeah, that's indeed handled by the conversion logic. The build & runtime dependencies are passed through these helpers, respectively:

https://github.com/garden-io/garden/blob/main/core/src/resolve-module.ts#L742-L762

vvagaytsev (dismissed):
Thanks for fixing this! 👍

core/src/plugins/container/container.ts
```diff
@@ -93,7 +93,6 @@ export async function configureContainerModule({ log, moduleConfig }: ConfigureM
 
     for (const volume of spec.volumes) {
       if (volume.module) {
-        moduleConfig.build.dependencies.push({ name: volume.module, copy: [] })
```
edvald, line 96: Why remove this?

edvald: Ah, nm, I see it below

edvald (approved):

Issue #3989, opened 2023-03-23
Escaping template strings not working


hnicke:
### Current Behavior

I need the literal value `${foo}` in a resulting k8s template of a garden kubernetes module.
The [garden docs](https://docs.garden.io/using-garden/module-templates#escaping-template-strings) describe how this should be done: Using `$${}` instead of `${}`.

```yaml
...
command: [ bash ]
args:
  - c
  - |
    ....
    -d '$${foo}'
```

However, it ends up literally like this after rendering:
```yaml
...
command: [ bash ]
args:
  - c
  - |
    ....
    -d '$${foo}'
```

I.e., no interpolation at all happened.

Garden doesn't eat the dollar. Maybe this only happens in the free tier :laughing: 

### Expected behavior
I expect this to be the result of the template rendering:
```yaml
...
command: [ bash ]
args:
  - c
  - |
    ....
    -d '${foo}'
```

### Additional idea
@worldofgeese pointed me to the docs, which I haven't found on my own.
I suggest moving the whole `Escaping template strings` paragraph from `Module Templates` to `Variables and Templating`. My understanding is that the `Module Templates` page only holds information in regards to the `kind: Module; type: templated` modules. The interpolation escape mechanism however is a globally available feature, right?

### Your environment

* **OS:** arch linux
* **How I'm running Kubernetes:** GKE
* 
`garden version`
v0.12.53

orymate:
We've workarounded this by rearranging the manifest templates in a way that all variable substitutions come before the parts that need escaping, and adding a comment like `# hack: $${}` between the two. No variable substitutions are done in the second part of the manifest, at least in 0.12.60.

```yaml
...
spec:
  loggingRef: ${environment.namespace}-demo  # TODO garden fails to expand this variable after the first escape, keeping this line before that
  # hack: $${}
  filters:
    - parser:
        metrics-probe:
          key: "probe"
          labels:
            namespace: "${json#kubernetes#namespace_name}"  # the manifests contain these exact label templates that use similar variable expansion syntax as garden
            pod: "${json#kubernetes#pod_name}"
```

vvagaytsev:
@orymate thanks for sharing the workaround!

vvagaytsev:
@orymate we have fixed this in [0.13.39](https://github.com/garden-io/garden/releases/tag/0.13.39). Feel free to try it out!

[...]

PR #5680, opened 2024-01-30
fix(template): unescape escape templates when not doing partial resolution

edvald:
Fixes: #3989

### DIFF
--- a/core/test/unit/src/template-string.ts
+++ b/core/test/unit/src/template-string.ts
@@ -72,8 +72,12 @@ describe("resolveTemplateString", () => {
     expect(res).to.equal("${bar}")
   })
 
-  it("should pass through a template string with a double $$ prefix", () => {
-    const res = resolveTemplateString({ string: "$${bar}", context: new TestContext({}) })
+  it("should pass through a template string with a double $$ prefix if allowPartial=true", () => {
+    const res = resolveTemplateString({
+      string: "$${bar}",
+      context: new TestContext({}),
+      contextOpts: { allowPartial: true },
+    })
     expect(res).to.equal("$${bar}")
   })
 
@@ -82,6 +86,15 @@ describe("resolveTemplateString", () => {
       string: "$${bar}",
       context: new TestContext({}),
       contextOpts: { unescape: true },
+    })
+    expect(res).to.equal("${bar}")
+  })
+
+  it("should unescape a template string with a double $$ prefix if allowPartial=false", () => {
+    const res = resolveTemplateString({
+      string: "$${bar}",
+      context: new TestContext({}),
+      contextOpts: { allowPartial: false },
     })
     expect(res).to.equal("${bar}")
   })
--- a/core/test/unit/src/template-string.ts
+++ b/core/test/unit/src/template-string.ts
@@ -128,6 +128,15 @@ describe("resolveTemplateString", () => {
 
   it("should escape things correctly 3", () => {
     const res = resolveTemplateString({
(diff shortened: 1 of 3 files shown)


eysi09 (dismissed):
LGTM

TimBeyer:
When I was trying to resolve this I had a few more test cases. Maybe it makes sense to check if those pass too?

```
  it("should allow nesting normal strings within escaped strings", () => {
    const res = resolveTemplateString("$${var.${foo}}", new TestContext({ foo: "yes" }), { unescape: true })
    expect(res).to.equal("${var.yes}")
  })

  it("should allow nesting escaped strings within normal strings", () => {
    const res = resolveTemplateString("${foo == 'yes' ? '$${bar}' : 'fail' }", new TestContext({ foo: "yes" }), { unescape: true })
    expect(res).to.equal("${bar}")
  })


  it.only("should escape things correctly", () => {
    const res = resolveTemplateString("$${env.TEST_ENV}", new TestContext({}), {
      unescape: true,
    })
    expect(res).to.equal("${env.TEST_ENV}")
  })

  it.only("should escape things correctly", () => {
    const res = resolveTemplateString("foo $${env.TEST_ENV} bar", new TestContext({}), {
      unescape: true,
    })
    expect(res).to.equal("foo ${env.TEST_ENV} bar")
  })

  it.only("should escape things correctly", () => {
    const res = resolveTemplateString("$${env:TEST_ENV}", new TestContext({}), {
      unescape: true,
    })
    expect(res).to.equal("${env:TEST_ENV}")
  })


  it.only("should escape things correctly", () => {
    const res = resolveTemplateString("foo $${env:TEST_ENV} bar", new TestContext({}), {
      unescape: true,
    })
    expect(res).to.equal("foo ${env:TEST_ENV} bar")
  })

  it("should escape things correctly", () => {
    const res = resolveTemplateString("${foo}-$${env:TEST_ENV}", new TestContext({ foo: "foo" }), {
      unescape: true,
    })
    expect(res).to.equal("foo-${env:TEST_ENV}")
  })
```

eysi09:
@TimBeyer, yeah good catch. 

Also, the framework tests appear to be failling.

core/test/unit/src/template-string.ts
```diff
@@ -71,8 +71,12 @@ describe("resolveTemplateString", () => {
     expect(res).to.equal("${bar}")
   })
 
-  it("should pass through a template string with a double $$ prefix", () => {
-    const res = resolveTemplateString({ string: "$${bar}", context: new TestContext({}) })
+  it("should pass through a template string with a double $$ prefix if allowPartial=true", () => {
+    const res = resolveTemplateString({
+      string: "$${bar}",
+      context: new TestContext({}),
+      contextOpts: { allowPartial: true },
```
stefreak, line 78: It's not really obvious to me how this fixes the end user issue – can we also add an integration test that shows how this fixes the issue in a less synthetic scenario?

edvald:
All fair points, flagging as WIP for now.

Btw the microk8s test failures seem unrelated. Not sure if we've been seeing those as flaky?

edvald:
@TimBeyer this isn't the current behavior, so it would be a breaking change:

> it("should allow nesting normal strings within escaped strings", () => {
    const res = resolveTemplateString("$${var.${foo}}", new TestContext({ foo: "yes" }), { unescape: true })
    expect(res).to.equal("${var.yes}")
  })

eysi09 (approved):

PR #5686, opened 2024-01-31
feat(config): add environments field on actions

### LINKED ISSUE #3221: [FEATURE]: avoid double negation by introducing `enabled:` for modules
## Feature Request
Deprecate `disabled:` module option and introduce `enabled`.

Alternatively, encourage using `enabled` and still allow `disabled` in the cases where it makes the code more readable.

### Background / Motivation
Starting out positive, both the positive and the negative case are easy to understand:

* `enabled: ${environment.name != "local"}` (read: enabled if environment is not local)
* `enabled: ${environment.name == "local"}` (read: enabled if environment is local)

While starting out negative, the negative case can be harder to read:

* `disabled: ${environment.name != "local"}` (read: disabled if environment is not local)

**Examples with simple boolean algebra**:

Easy to understand:

* `enabled: ${environment.name == "kubernetes" || environment.name == "local-kubernetes"}` (read: the module is enabled if the environment is kubernetes or local-kubernetes)
* `enabled: ${environment.name != "kubernetes" && environment.name != "local-kubernetes"}` (read: the module is enabled if the environment is not kubernetes and not local-kubernetes)
* `disabled: ${environment.name == "kubernetes" || environment.name == "local-kubernetes"}` (read: the module is disabled, if environment is kubernetes or local-kubernetes)

Hard to understand

* `disabled: ${environment.name != "kubernetes" && environment.name != "local-kubernetes"}` (read: the module is disabled, if environment is not kubernetes and is not local-kubernetes)

So, to sum it up, enabled is okay to understand in all cases, and disabled is hard to understand in case of double negation (at least for me, but [also to other people](https://maximilianocontieri.com/code-smell-51-double-negatives)).

use `enabled:` to configure wether a module is active, instead of `disabled`.

Readability

There are 2 options:

* We can have both `enabled` and `disabled` side by side
* We can deprecate `disabled`

:rose: It’s a nice to have, but nice things are nice :slightly_smiling_face:

edvald:
This allows setting e.g. `environments: [dev]` on actions instead of the
more verbose `disabled: ${environment.name != "dev"}`.

Resolves #3221

### DIFF
--- a/core/src/actions/base.ts
+++ b/core/src/actions/base.ts
@@ -174,6 +174,14 @@ export const baseActionConfigSchema = createSchema({
       `
       )
       .meta({ templateContext: ActionConfigContext }),
+    environments: joi
+      .sparseArray()
+      .items(joi.string())
+      .description(
+        dedent`
+        If set, the action is only enabled for the listed environment types. This is effectively a cleaner shorthand for the \`disabled\` field with an expression for environments. For example, \`environments: ["prod"]\` is equivalent to \`disabled: \${environment.name != "prod"}\`.
+        `
+      ),
 
     // Version/file handling (Note: Descriptions and behaviors are different on Build actions!)
     include: includeExcludeSchema()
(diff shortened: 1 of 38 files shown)


vvagaytsev (dismissed):

stefreak:
Added a note that this resolves #3221

stefreak (approved):

eysi09:
@edvald, looks like the docs need to be updated.

PR #5687, opened 2024-01-31
chore(ci): change several checks to only run on main branch

edvald:


vvagaytsev (approved):
Looks good, thank you!

Was the intention to run tests only on a single VM? Now we do it only for `vm-1.23-minikube` - the oldest supported k8s version. Was that done intentionally?

We could use the latest supported version, but we do run `test-plugins` with that version already.

.circleci/continue-config.yml
```diff
@@ -1168,7 +1168,7 @@ workflows:
           <<: *only-prs
           requires: [build-dist-linux-windows]
       - test-windows:
-          <<: *only-internal-prs
+          <<: *only-main
```
stefreak, line 1171: At the moment we are not running tests on main at all; We are running them in the merge queue. Is only-main the right one here? I guess we need to run these on the merge queue, instead of main.

stefreak: The problem with only running them in the merge queue is that it's tricky to find which particular test caused the merge queue to fail; That's a limitation of CircleCI's UI and the fact that we are cancelling the workflow when any test fails.

vvagaytsev (changes requested):
There is an [open discussion](https://github.com/garden-io/garden/pull/5687#discussion_r1474353728). Let's resolve it before merging.

stefreak:
I'd like to avoid making this change, if possible. If the tests are not flaky, waiting for 20 minutes is not that big of a deal IMO. Enforcing the tests was very effective at fighting flakyness, because it made flakes more miserable and because of that we eliminated them mostly.

A really effective tool at making our CI more productive would be caching; E.g. if tests passed in PR, it would be nice they would not re-run in the Merge queue.

PR #5688, opened 2024-01-31
fix(cli): only overwrite terminal writer if using Ink

eysi09:

Before this fix, we'd reset the terminal writer type after initialising the command at the CLI level.

That's basically fine in and off itself except that it broke the test logger and overwrote the settings we use there.

Instead of adding even more config options for the log level I opted for simplifying the flow and only overwriting the type if using the 'Ink' terminal write, since that's only use case for it anyway.

That implicitly fixes the issue with the test logger and slightly simplifies the overall flow.

Fixes #


vvagaytsev:
There are some test failures.

thsig (approved):

PR #5690, opened 2024-01-31
fix(k8s): don't throw when log fetching fails

### LINKED ISSUE #5559: Crash: toKubernetesError encountered an unknown error unexpectedly during Kub...
### Error message

✖ deploy.platform-ui        → Failed (took 8.7 sec)
✖ deploy.platform-ui        → Failed resolving status for Deploy type=container name=platform-ui (took 9.32 sec). This is what happened:

────────────────────────────────────────────────────────────────────────────────────────────────────

You can help by reporting this on GitHub: https://github.com/garden-io/garden/issues/new?labels=bug,crash&template=CRASH.md&title=Crash%3A%20toKubernetesError%20encountered%20an%20unknown%20error%20unexpectedly%20during%20Kub...

Failed resolving status for Deploy type=container name=platform-ui (took 9.32 sec). This is what happened:
Error: Cannot parse content. No Content-Type defined.
  at ObjectSerializer.parse (file:///github/home/.local/share/garden/1702543490-UakctbW.r/rollup/garden.mjs:5[67](https://github.com/FQX-AG/fqx/actions/runs/7206185255/job/19631282968?pr=1915#step:9:68)762:21)
  at CoreV1ApiResponseProcessor.readNamespacedPodLogWithHttpInfo (file:///github/home/.local/share/garden/1702543490-UakctbW.r/rollup/garden.mjs:601871:111)
  at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
  at async retry (file:///github/home/.local/share/garden/1702543490-UakctbW.r/rollup/garden.mjs:[69](https://github.com/FQX-AG/fqx/actions/runs/7206185255/job/19631282968?pr=1915#step:9:70)8396:20)
  at async requestWithRetry (file:///github/home/.local/share/garden/1702543490-UakctbW.r/rollup/garden.mjs:698422:20)
  at async file:///github/home/.local/share/garden/1702543490-UakctbW.r/rollup/garden.mjs:702239:19
  at async Promise.all (index 0)
  at async file:///github/home/.local/share/garden/1702543490-UakctbW.r/rollup/garden.mjs:702302:25
  at async Promise.all (index 0)
  at async getFormattedPodLogs (file:///github/home/.local/share/garden/1702543490-UakctbW.r/rollup/garden.mjs:702297:21)
────────────────────────────────────────────────────────────────────────────────────────────────────

toKubernetesError encountered an unknown error unexpectedly during Kubernetes API: readNamespacedPodLog: toKubernetesError encountered an unknown error unexpectedly during readNamespacedPodLog: Cannot parse content. No Content-Type defined.

### What did you do?

Run deploy in CI

### Your environment

gardendev/garden-gcloud:0.13.21-alpine on github actions
### Frequency

Happens every time now on this specific PR with minimal changes.

thsig:

This should fix an intermittent class of errors reported by users, where unhealthy pods result in Garden failing to fetch logs for these same pods (for the purposes of error reporting) and throwing a noisy error that obscures the underlying problem.

See https://github.com/garden-io/garden/issues/5559, for example.

Here, we catch errors thrown during the diagnostic pod log error fetching, and simply log a warning message that the log fetching failed.

Fixes #5559 (hopefully).

It's kind of hard to write tests for this, but at least we're explicitly only catching an exception we weren't catching before, so this should be eminently safe.

### DIFF
--- a/core/src/plugins/kubernetes/status/workload.ts
+++ b/core/src/plugins/kubernetes/status/workload.ts
@@ -25,6 +25,7 @@ import { getFormattedPodLogs, POD_LOG_LINES } from "./pod.js"
 import type { ResourceStatus, StatusHandlerParams } from "./status.js"
 import { getResourceEvents } from "./events.js"
 import { styles } from "../../../logger/styles.js"
+import { KubernetesError } from "../api.js"
 
 const containerStatusFailures = ["CrashLoopBackOff", "CreateContainerConfigError", "ImagePullBackOff"]
 
@@ -76,7 +77,19 @@ export async function checkWorkloadStatus({ api, namespace, resource }: StatusHa
 
     // Attach pod logs for debug output
     const pods = await getPods()
-    const podLogs = (await getFormattedPodLogs(api, namespace, pods)) || undefined
+    let podLogs: string | undefined
+    try {
+      podLogs = await getFormattedPodLogs(api, namespace, pods)
+    } catch (err) {
+      const podNames = pods.map((pod) => styles.highlight(pod.metadata.name)).join(", ")
+      logs += styles.secondary(`Warning: Could not retrieve logs for one or more of the following pods: ${podNames}`)
+      if (err instanceof KubernetesError) {
+        if (err.message && err.message.length > 0) {
+          logs += styles.secondary(`Error message: ${err.message}`)
+        }
+      }
+      podLogs = undefined
+    }
 
     if (podLogs) {
       logs += styles.accent("\n\n━━━ Pod logs ━━━\n")
--- a/core/src/plugins/kubernetes/retry.ts
+++ b/core/src/plugins/kubernetes/retry.ts
@@ -123,7 +123,7 @@ export function toKubernetesError(err: unknown, context: string): KubernetesErro
     // The ErrorEvent does not expose the status code other than as part of the error.message
   } else {
     // In all other cases, we don't know what this is, so let's just throw an InternalError
-    throw InternalError.wrapError(err, `toKubernetesError encountered an unknown error unexpectedly during ${context}`)
+    throw InternalError.wrapError(err, `toKubernetesError encountered an unknown error during ${context}`)
   }
 
   let message = `Error while performing Kubernetes API operation ${context}: ${errorType}\n`


twelvemo (approved):
👍

thsig:
I'm going to add a bit more diagnostic information at a higher log level while I'm at it, might prove useful for debugging the underlying k8s issues that trigger this.

core/src/plugins/kubernetes/status/workload.ts
```diff
@@ -76,7 +82,18 @@ export async function checkWorkloadStatus({ api, namespace, resource }: StatusHa
...
     // Attach pod logs for debug output
     const pods = await getPods()
-    const podLogs = (await getFormattedPodLogs(api, namespace, pods)) || undefined
+    let podLogs: string | undefined
+    try {
+      podLogs = await getFormattedPodLogs(api, namespace, pods)
+    } catch (err) {
+      const podNames = pods.map((pod) => styles.highlight(pod.metadata.name)).join(", ")
+      logs += styles.secondary(`Warning: Could not retrieve logs for one or more of the following pods: ${podNames}`)
+      if (err instanceof GardenError) {
```
twelvemo, line 91: Wouldn't this be an instance of `KubernetesError` error here? Or should we output the error details, no matter which kind?

thsig: `KubernetesError` extends `GardenError`, so I thought I'd catch the slightly more generic case.

We should be able to log details for any `GardenError` here.

twelvemo (dismissed):

core/src/plugins/kubernetes/status/workload.ts
```diff
@@ -76,7 +82,18 @@ export async function checkWorkloadStatus({ api, namespace, resource }: StatusHa
...
+        log.debug(`Error details:`)
+        log.debug(JSON.stringify(err, null, 2))
```
stefreak, line 93: Error details that are relevant to the user are always encoded in the message. We removed error details on purpose and there is no need to stringify the error here. It can even be harmful, if the data on the error happens to be circular or very large.

stefreak: I think we might want to include the error message in the warning, regardless of the error type.

Let's say this fails some users, and they ask for help. We then have to tell them that they need to reproduce the error with debug loglevel. That's why I'd opt for adding the error message to the warning. If that's noisy or misleading in some cases, we need to handle these particular cases excplicitly.

thsig: Yeah, good point—I've updated the PR accordingly.

thsig:
Thanks for the comments, @stefreak. Does this look OK now?

core/src/plugins/kubernetes/status/workload.ts
```diff
@@ -76,7 +82,21 @@ export async function checkWorkloadStatus({ api, namespace, resource }: StatusHa
...
+      podLogs = await getFormattedPodLogs(api, namespace, pods)
+    } catch (err) {
+      const podNames = pods.map((pod) => styles.highlight(pod.metadata.name)).join(", ")
+      logs += styles.secondary(`Warning: Could not retrieve logs for one or more of the following pods: ${podNames}`)
+      if (err instanceof KubernetesError) {
+        if (err.message && err.message.length > 0) {
+          logs += styles.secondary(`Error message: ${err.message}`)
+        }
+        log.debug(styles.secondary(`responseStatusCode: ${err.responseStatusCode}`))
+        log.debug(styles.secondary(`API message: ${err.apiMessage}`))
```
stefreak, line 96: The message is all you need, and it's all you need to ever think about. If something is missing in the message, we need to adapt the code that throws the error to include additional details when necessary. 

See also https://github.com/garden-io/garden/blob/main/core/src/plugins/kubernetes/retry.ts#L135

Additional properties exist only when it's necessary for control flow, e.g. to handle certain status codes like permission denied, bad request, etc.

thsig: Yeah, good point. I've removed those log lines now. Is this good to go?

stefreak (approved):

core/src/plugins/kubernetes/status/workload.ts
stefreak: Yes, this looks good to me!

Sorry that I didn't remember that yesterday: one thing that might be helpful to have in debug loglevel is the stack of the errors. 

I approved the PR anyway 👍

stefreak (approved):

PR #5692, opened 2024-02-01
improvement(core): include path in template errors

thsig:

When available, include the path to the template string in the YAML doc in question.

Also stop truncating the template expression in the message.

### DIFF
--- a/core/test/unit/src/template-string.ts
+++ b/core/test/unit/src/template-string.ts
@@ -20,6 +20,7 @@ import { expectError, getDataDir, makeTestGarden } from "../../helpers.js"
 import { dedent } from "../../../src/util/string.js"
 import stripAnsi from "strip-ansi"
 import { TemplateStringError } from "../../../src/exceptions.js"
+import repeat from "lodash-es/repeat.js"
 
 class TestContext extends ConfigContext {
   constructor(context) {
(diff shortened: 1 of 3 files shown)


core/test/unit/src/template-string.ts
```diff
@@ -156,17 +156,6 @@ describe("resolveTemplateString", () => {
     })
   })
 
-  it("should trim long template string in error messages", () => {
-    void expectError(
-      () =>
-        resolveTemplateString({
-          string: "${some} very very very very very long long long long long template string",
```
stefreak, line 163: Truncation happens because we template files, e.g. kubernetes manifests. Should it print the whole file (e.g. 1000 lines of text) if there is a template error in there?

edvald: I think a more sensible approach is just a higher limit, and easy enough to do.

thsig: Aha yeah, makes sense. Hadn't thought of that use-case.

thsig: Ok, I've updated the PR.

stefreak (approved):

vvagaytsev (approved):

PR #5697, opened 2024-02-02
fix(core): don't execute disabled dependencies

thsig:

When executing an action that depends on a disabled dependency, the disabled dependency should be skipped (unless it's a build).

This is fixed here (in accordance with the docstring for the `disabled` field on action configs).

### DIFF
--- a/core/src/tasks/base.ts
+++ b/core/src/tasks/base.ts
@@ -275,7 +275,7 @@ export abstract class BaseActionTask<T extends Action, O extends ValidResultType
         }
         return [this.getExecuteTask(action)]
       } else if (dep.explicit) {
-        if (this.skipRuntimeDependencies && dep.kind !== "Build") {
+        if ((this.skipRuntimeDependencies || disabled) && dep.kind !== "Build") {
           if (dep.needsStaticOutputs) {
             return [this.getResolveTask(action)]
           } else {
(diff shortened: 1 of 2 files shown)


stefreak:
So before this change, we did execute actions if they were dependencies

Now the behaviour changes to silently skip executing disabled actions if other actions directly depend on them.

I would have expected an error message or at least a warning here;

Another question that comes to my mind is if we need to treat this as a breaking change? How did 0.12 behave?

thsig:
We did execute disabled actions if they were dependencies of a non-disabled action being processed.

Now, we only execute disabled Builds in this way, but not disabled Deploys/Tests/Runs.

This is what the docs say, and that's why [this user was confused about the changed behaviour from 0.12](https://discord.com/channels/817392104711651328/1199182752755503104/1199182752755503104) (where we did indeed ignore disabled dependencies).

I'm not sure whether to treat this as a breaking change, since this change means we're bringing the behaviour in line with the docs for the `disabled` field.

... that said, I'm also cool with holding off on merging this until the next minor/non-patch release.

thsig:
Here's the relevant docstring that populates the reference docs for the `disabled` field:

https://github.com/garden-io/garden/blob/15b0e8de8b847f664c300608230a2acc7ca0b91b/core/src/actions/base.ts#L169-L173

stefreak:
I think if this was what the documentation said and how 0.12 behaved like, then we should move forward with this and merge 👍 Thanks for the clarification @thsig

stefreak (approved):
LGTM 👍

PR #5700, opened 2024-02-03
docs: stop using build field in docs and examples

eysi09:
When we released Bonsai we introduced the following short hand
for referencing container Build actions in container Deploy actions
via the `build` field:

```yaml
kind: Deploy
name: api
build: api
spec:
```

The longer version would be:

```yaml
kind: Deploy
name: api
dependencies: [build.api]
spec:
  image: ${actions.build.api.outputs.deploymentImageId}
  # ...
```

The short hand works for `container` Deploy actions but not for `helm`
or `kubernetes` Deploy actions. In fact using the `build` field there
sets the build context to the `.garden/build` dir of the corresponding
`build`.

This has caused a lot of confusion for users as they copy examples
containing the `build` field to `kubernetes` and `helm` Deploy actions.

A common symptom is that a Garden command will fail because it can't
find the files belonging to the Deploy action.

Furthermore, the `build` field has in general caused confusion and we
_may_ end up deprecating it.

In the meantime, this commit removes any unneeded/false usage of it in our
examples and docs.

The result is that `container` Deploy actions are slightly more
verbose but in turn we've removed a big footgun that users keep bumping
into as they copy/paste example config.

There are still legit use cases for the `build` field and I've left
those examples alone.

Fixes #

Most of the updates were done with the help of ChatGPT and it looks a couple of comments got lost along the way but I think we can skip those anyway.

### DIFF
--- a/docs/getting-started/next-steps.md
+++ b/docs/getting-started/next-steps.md
@@ -174,8 +174,10 @@ kind: Test
 name: api-integ
 type: container
 description: A Test action for integration testing the api after its been deployed
-build: api # <--- Use the api image to run the test
-dependencies: [deploy.api]
+dependencies: [build.api, deploy.api]
+spec:
+  image: ${actions.build.api.outputs.deploymentImageId}
+  command: [./integ-tests.sh]
 ```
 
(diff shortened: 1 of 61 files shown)


stefreak (approved):
Looks good to me!

Issue #5582, opened 2023-12-19
Align includes/excludes with .gitignore pattern format


stefreak:
## Feature Request

Garden includes/ excludes should behave exactly like .gitignore patterns: https://git-scm.com/docs/gitignore

### Background / Motivation

Simplify reasoning about includes / excludes, .gitignore and .gardenignore files by making sure that the patterns all behave exactly the same.

🌵 Not having this feature makes using Garden painful

PR #5710, opened 2024-02-06
fix(git): fix file list caching bug for repo mode

thsig:

This fixes an error that would come up e.g. when an action config in the project root excluded a subdirectory containing another action config.

In that setup, the file list for the action in the subdirectory would be empty because the file list had been cached using the exclude config applied when computing the file list for the action config at the root level (which included no files for the excluded subdirectory).

This was fixed by adding another level of caching to `GitRepoHandler` that accounts for different includes/excludes and filter functions for the same path.

Note the slightly awkward clarification section about the difference between the exclude behavior of `subtree` and `repo` when no includes are configured.

In a follow-up to this, we really should address https://github.com/garden-io/garden/issues/5582.

I think bringing our include/exclude behavior completely in line with Git's would make things easier to reason about for our users.

### DIFF
--- a/core/src/vcs/git-repo.ts
+++ b/core/src/vcs/git-repo.ts
@@ -13,6 +13,8 @@ import fsExtra from "fs-extra"
 import { pathToCacheContext } from "../cache.js"
 import { FileTree } from "./file-tree.js"
 import { normalize, sep } from "path"
+import { stableStringify } from "../util/string.js"
+import { hashString } from "../util/util.js"
 
 const { pathExists } = fsExtra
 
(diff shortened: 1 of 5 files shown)


twelvemo (approved):
Looks good! For others reviewing, we decided to favor the inconsistency between the two file scanning modes, to shimming the behavior of the `subtree` mode which excludes all occurrences `foo.txt` in every subdir leading to unexpected excludes.

vvagaytsev:
Just a note to link the related PR together. This fixed some filtering login implemented in #5526.

PR #5712, opened 2024-02-06
fix(buildkit): run command from context dir

eysi09:

When using the `cluster-buildkit` build mode, Garden syncs the build
context to the `garden-build` directory in the `garden-buildkit` Pod
(using the `garden-sync` volume).

The context will end up at a path like `/garden-build/<hash>/build-name`
but the command itself is run from the root of the `buildkitd`
container in the `garden-buildkit` Pod.

This means that passing secrets as files to the build command won't
work.

E.g., if you add the following to the Build action spec:

```yaml
extraFlags:
  - --secret=id=foo,src=./my-secrets/foo
```

...where `./my-secrets/foo` is a file, then the path will be incorrect
when the command is executed from the root dir.

This commit fixes that by first changing into the correct directory and
then  executing the `buildctl` command from there.

Note 1: There's no way to pass the cwd to the K8s API and this appears
to be the recommended way, e.g. when using `kubectl exec`.

Note 2: Kind of a side note, but passing environment variables instead
of files does not appear to work, potentially due to this buildkit bug:
https://github.com/moby/buildkit/issues/3787

Note 3: Kaniko doesn't support the `--secret` flag. This means that users
can't re-use a Build action that uses the `--secret` flag across
environments where one uses buildkit and the other kaniko.

Fixes #


core/test/unit/src/plugins/kubernetes/container/build/buildkit.ts
```diff
@@ -19,31 +19,11 @@ import {
   getBuildkitDeployment,
   getBuildkitFlags,
   getBuildkitImageFlags,
+  makeBuildkitBuildCommand,
 } from "../../../../../../../src/plugins/kubernetes/container/build/buildkit.js"
 import { getDataDir, makeTestGarden } from "../../../../../../helpers.js"
+import { k8sGetContainerBuildActionOutputs } from "../../../../../../../src/plugins/kubernetes/container/handlers.js"
 
-describe("getBuildkitModuleFlags", () => {
```
eysi09, line 25: This test was basically duplicated and is also run below.

core/src/plugins/kubernetes/container/build/buildkit.ts
```diff
@@ -258,6 +243,47 @@ export async function ensureBuildkit({
   })
 }
 
+/**
+ * Returns the full build command which first changes into the build context directory
+ * and then runs the `buildctl` command.
+ *
+ * We change into to build context directory to e.g. ensure that secret files that are
```
thsig, line 250:
```suggestion
 * We change into the build context directory to e.g. ensure that secret files that are
```

thsig (dismissed):
Good stuff! Had one suggestion there (a typo in a comment), but looks good otherwise.

thsig (approved):

PR #5716, opened 2024-02-09
fix(get-config): omit internal field from output

thsig:

Before this fix, we were including the `internal` field of actions and other configs in the output of the `get config` command. This was undesirable and very verbose.


vvagaytsev:
@thsig there are some test failures, likely some trivial assertion errors because of the dropped `internal` field in the config output.

vvagaytsev (approved):

PR #5717, opened 2024-02-12
fix(module-conversion): always provide dummy build

thsig:

In 0.12, many users had build dependencies on modules with no-op build steps (e.g. `terraform`, `kubernetes` and `helm` modules)

While this was harmless in 0.12, this now results in noisy errors. We fix this in a simple way here by always providing a no-op dummy build to conversion handlers.


thsig:
Closing in favour of #5727, as the implementation here proved to be problematic.

PR #5727, opened 2024-02-13
fix(module-conversion): skip omitted build deps

thsig:

When a module has a build dependency on another module that doesn't have a corresponding Build after being converted into actions (e.g. a build dependency on a `terraform` module), we now log a warning and remove the dependency from the generated actions instead of erroring out.

In 0.12, some users had build dependencies on modules with no-op build steps (e.g. `terraform`, `kubernetes` and `helm` modules)

While this was harmless in 0.12, this resulted in noisy errors before this fix.

### DIFF
--- a/core/src/resolve-module.ts
+++ b/core/src/resolve-module.ts
@@ -26,7 +26,7 @@ import {
   PluginError,
   toGardenError,
 } from "./exceptions.js"
-import { dedent } from "./util/string.js"
+import { dedent, deline } from "./util/string.js"
 import type { GardenModule, ModuleConfigMap, ModuleMap, ModuleTypeMap } from "./types/module.js"
 import { getModuleTypeBases, moduleFromConfig } from "./types/module.js"
 import type { BuildDependencyConfig, ModuleConfig } from "./config/module.js"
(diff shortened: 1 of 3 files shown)


core/src/resolve-module.ts
```diff
@@ -838,6 +839,47 @@ export const convertModules = profileAsync(async function convertModules(
...
+  const convertedBuildNames = new Set(allActions.filter((a) => isBuildActionConfig(a)).map((a) => a.name))
+  const missingBuildNames = new Set(
+    graph
+      .getModules()
+      .map((m) => m.name)
+      .filter((name) => !convertedBuildNames.has(name))
+  )
+
+  for (const action of allActions) {
+    action.dependencies = (action.dependencies || []).filter((dep) => {
```
vvagaytsev, line 856: It might be a bit better to separate the filtering logic from the warnings logging. We can use `partition` to separate existing and missing deps, and then process each collection separately, just for better readability and SRP :)

See #5730.

vvagaytsev:
LGTM, thank you! I've left a comment, that's not critical. I'll approve the PR.

vvagaytsev (approved):

PR #5732, opened 2024-02-14
improvement(core): less noisy missing dep errors

thsig:

Before this fix, we would include a stack trace in the error logged when a requested action has a missing dependency. This was very noisy and indicated a problem with Garden (when it's a totally normal user error).


vvagaytsev (approved):
LGTM 👍

PR #5733, opened 2024-02-14
improvement(sync): purely interactive `sync start` and `stop` commands

vvagaytsev:
BREAKING CHANGE

This PR removes the ability to run `garden sync start` and `garden sync stop` outside the dev console.

* Command `sync start` now automatically starts the dev console like `deploy --sync` does.
* Flag `--monitor` of the `sync start` command has no effect and is marked as removed.
* Command `sync stop` aborts with a warning message if it's started from outside the dev console.

These changes will allow us to fix the existing issues with the syncing resource management, and make the underlying Mutagen processed managed by Garden.

Technically, it is a breaking change. But it should not affect UX too much.

Changes in the lifecycle management of the Mutagen daemon and sync monitors will be done in a separate PR as a fix.


eysi09:
As discussed elsewhere, this is a direction we want to move to but we can't make a breaking change in a patch release. People might e.g. depend on this in scripts and not having the command exit would break those scripts. 

Instead we should log a warning telling users that this behaviour is deprecated and will be changed in a future breaking change release. Instead we recommend running `garden deploy --sync` or starting syncs _inside_ the dev console with either `deploy --sync` or `sync start`.

vvagaytsev:
Thanks @eysi09! I've created a PR with deprecation warnings, see #5747.

vvagaytsev:
Closing this for now. We can re-open this later.

PR #5735, opened 2024-02-15
fix(core): versioning fix for remote sources

### LINKED ISSUE #5501: Version hash is dependent on repository source folder name

thsig:

Before this fix, we were including the relative path to a file from project root in the string to be hashed when calculating versions.

This was problematic when linking remote sources, since the chosen name and path of the locally linked repo would be  factored into the version (resulting in unwanted cache misses).

This was fixed by using the relative path from the directory containing the file's parent config. For example, an action's directory should be able to be moved around or renamed without it affecting the version of that action.

Fixes #5501.


vvagaytsev (dismissed):
Looks good! 💯 Thanks for converting a test project to the action-based config! We should definitely do it incrementally and often.

vvagaytsev:
There are some test failures. Likely just some trivial assertion failures because of the changed hash calculation.

vvagaytsev (approved):

PR #5745, opened 2024-02-19
fix(core): better Zod validation error messages

### LINKED ISSUE #5446: 0.13: [Bug]: Spec validation error message is printed with JSON encoding
### Current Behavior

When the `spec.command` field is not defined in an `exec` `Test` action, I currently get the following error message:

```
Error validating spec for Test type=exec name=foo:
[
  {
    "code": "invalid_type",
    "expected": "array",
    "received": "undefined",
    "path": [
      "command"
    ],
    "message": "Required"
  }
]
```

### Expected behavior

I expect a human-readable error message, something like this:
```
Error validating spec for Test type=exec name=foo:

...
25  | type: exec
26  | name: foo
------------^
Expected an array at 'spec.command', actually received 'undefined'.
````

### Reproducible example

Command: `garden test`
```
# garden.yml

apiVersion: garden.io/v1
kind: Project
name: foo
environments:
  - name: local

---

kind: Test
type: exec
name: foo
```

### Your environment

* **OS:** macOS

`garden version` 0.13.20

### PR DESCRIPTION
**Pull request by thsig**

Before this fix, we were simply passing through the default error messages generated by Zod when validation failed. This resulted in JSON-like error messages when a Zod schema validation error happened.

This was fixed by detecting Joi VS Zod validation errors, and adding a little rendering logic for Zod errors to improve the output.

Fixes #5446.

Before this fix, introducing an `exec` Test with a missing `spec.command` would result in an error message like this:

Screenshot 2024-02-19 at 12 06 18

After this fix:

Screenshot 2024-02-19 at 12 05 33

### CODE REVIEW

File: core/src/config/base.ts [hunk 1]
```diff
@@ -48,6 +48,8 @@ The format of the files is determined by the configured file's extension:
 _NOTE: The default varfile format will change to YAML in Garden v0.13, since YAML allows for definition of nested objects and arrays._
 `.trim()
 
+export type ObjectPath = (string | number)[]
```
vvagaytsev, line 51: There is a duplicate type definition in https://github.com/garden-io/garden/blob/main/core/src/template-string/template-string.ts#L73

That one can be dropped in favor of this one.

vvagaytsev: I did that in babb28cd9e63d379760c53a4fa8e8242531f85fa.

vvagaytsev (approved):
LGTM! 👍 Thank you!

PR #5747, opened 2024-02-19
feat(sync): deprecate non-interactive `sync start` and `stop` commands

vvagaytsev:
The current behaviour of both commands is now deprecated and will be changed in a future breaking change release.
Both commands will be allowed to execute only inside dev console.

We recommend to run these commands from inside dev console starting from now.

### DIFF
--- a/core/src/commands/deploy.ts
+++ b/core/src/commands/deploy.ts
@@ -161,7 +161,7 @@ export class DeployCommand extends Command<Args, Opts> {
   }
 
   override useInkTerminalWriter(params) {
-    return !!this.maybePersistent(params)
+    return this.maybePersistent(params)
   }
 
   override terminate() {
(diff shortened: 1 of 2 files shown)


vvagaytsev:
This PR addresses the [comment from #5733](https://github.com/garden-io/garden/pull/5733#issuecomment-1947426312).

core/src/commands/sync/sync-stop.ts
```diff
@@ -60,6 +61,15 @@ export class SyncStopCommand extends Command<Args, Opts> {
   async action(params: CommandParams<Args, Opts>): Promise<CommandResult<{}>> {
     const { garden, log, args } = params
 
+    if (!params.parentCommand) {
```
eysi09, line 64: I think we should skip this.

We don't want users starting syncs outside of the dev console but the sync stop command can still make sense, e.g. if you need to manually stop them after a process crash.

eysi09 (changes requested):

core/src/commands/sync/sync-stop.ts
vvagaytsev: Thanks! I removed this change.

eysi09 (approved):

PR #5754, opened 2024-02-20
refactor: use function `makeDocsLink` to generate docs links

vvagaytsev:
Some unification for creating and printing doc links:

* Use function `makeDocsLink` to avoid repetitive pattern of concatenation `DOCS_BASE_URL` with doc relative paths.
* Use `styles.link` along `makeDocsLink` in logging and error messages.

The function hasn't been used in the schema and reference docs yet, because it broke the generated docs output.


core/src/config/base.ts
```diff
@@ -374,7 +377,11 @@ function handleMissingApiVersion(log: Log, projectSpec: ProjectConfig): ProjectC
   if (projectSpec["apiVersion"] === undefined) {
     emitNonRepeatableWarning(
       log,
-      `"apiVersion" is missing in the Project config. Assuming "${GardenApiVersion.v0}" for backwards compatibility with 0.12. The "apiVersion"-field is mandatory when using the new action Kind-configs. A detailed migration guide is available at ${DOCS_BASE_URL}/guides/migrating-to-bonsai`
+      `"apiVersion" is missing in the Project config. Assuming "${
+        GardenApiVersion.v0
+      }" for backwards compatibility with 0.12. The "apiVersion"-field is mandatory when using the new action Kind-configs. A detailed migration guide is available at ${styles.link(
```
stefreak, line 382: Could makeDocsLink already apply the style? We can offer an makeUnstyledDocsLink in the rare case you need it without style.

stefreak (approved):
Nice change 👍

core/src/config/base.ts
vvagaytsev: Good point! I'll address that, thank you :)

PR #5759, opened 2024-02-20
fix: do not fail on empty YAML varfiles

### LINKED ISSUE #5741: Empty varfile causes error.
### Current Behavior

For explicit varfile definitions on actions if you have no configuration to put in the yaml file and it is empty you will get an error like this:

```
Error processing config for Deploy action my-deploy:                               
Unable to load varfile at 'my-settings_local.yaml': Error: Configured variable file my-settings_local.yaml must be a single plain YAML mapping. Got: undefined
                    
```

where `my-settings_local.yaml` is just empty and the action config looks like this:

```yaml
kind: ...
type: ...
name: my-deploy
varfiles:
  - my-settings_${environment.name}.yaml
```

### Expected behavior

An empty varfile should be a valid option.

### Reproducible example

The example above should be enough to repro in a test case, which I did not prepare for this report.

### Workaround

I just add some nonsense to the file that isn't used in my stack:

```yaml
nothing: here
```

You could also do something like conditionally use the varfile. I'm not sure that works though and I don't want to try it.

### Suggested solution(s)

I think there would just need to be a special check for blank files is all.

### Additional context

N/A

### Your environment

* **OS:** OpenSuse Tumbleweed linux

`garden version`
`garden version: 0.13.23`

vvagaytsev:

The `js-yaml` parser returns `undefined` instead of an empty object (`{}`) for empty input files.
So, we handle empty YAML varfiles as empty variable dictionaries to avoid runtime errors.

This PR also converts `module-varfiles` test project to action-based configuration.

Fixes #5741


core/src/config/base.ts
```diff
@@ -576,25 +577,30 @@ export async function loadVarfile({
     const filename = basename(resolvedPath.toLowerCase())
 
     if (filename.endsWith(".json")) {
+      // JSON parser throws a JSON syntax error on completely empty input file,
+      // and returns an empty object for an empty JSON.
```
vvagaytsev, line 581: We could inspect the data.toString() before calling any parser and just return {} if data.toString().trimI() === "".
But that would be conflicting with the official JSON spec, because an empty string is not a valid JSON.

stefreak (approved):
Nice work!

I like this change and the effort to convert module tests to actions, but we need to make sure not to convert all module tests or duplicate some of the tests for modules and actions, so we have a good coverage for both.

Otherwise we might end up introducing regressions for modules in future releases, if the test coverage ends up to be too low.

PR #5760, opened 2024-02-20
chore(lint): enable linter for `*.tsx` files

vvagaytsev:

Enable linter for `*.tsx` files too.

### DIFF
--- a/core/src/commands/dev.tsx
+++ b/core/src/commands/dev.tsx
@@ -190,13 +190,16 @@ Use ${styles.bold("up/down")} arrow keys to scroll through your command history.
   }
 
   private async initCommandHandler(params: ActionParams) {
-    // eslint-disable-next-line @typescript-eslint/no-this-alias
-    const _this = this
     const { garden, log, opts } = params
 
     // override the session for this manager to ensure we inherit from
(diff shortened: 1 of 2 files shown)


core/src/commands/dev.tsx
```diff
@@ -189,6 +190,7 @@ Use ${styles.bold("up/down")} arrow keys to scroll through your command history.
   }
 
   private async initCommandHandler(params: ActionParams) {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
```
stefreak, line 193: Slightly cleaner than that would be to declare quit like this:

```
const quit = () => {
      cl?.disable("🌷  Thanks for stopping by, love you! ❤️")
      this.terminate()
    }
```
And then remove the this alias

stefreak (changes requested):
Looking good! I have one minor comment.

core/src/commands/dev.tsx
vvagaytsev: Thanks! Fixed that in 28a66041b209d6b6d756baf4ab21b935e2d536e7.

stefreak (approved):
❤️

PR #5761, opened 2024-02-20
chore: some fixes in tests and code style

vvagaytsev:

Fixes #


core/src/plugins.ts
```diff
@@ -304,7 +304,9 @@ function resolvePlugin(
   for (const [name, handler] of Object.entries(base.handlers || {})) {
     if (!handler) {
       continue
-    } else if (resolved.handlers[name]) {
```
stefreak, line 307: Doesn't this change the behaviour?

core/src/cloud/api.ts
```diff
@@ -580,7 +580,7 @@ export class CloudApi {
     return this.apiFetch<T>(path, {
       method: "GET",
       headers: headers || {},
-      retry: retry === false ? false : true, // defaults to true unless false is explicitly passed
+      retry: retry !== false, // defaults to true unless false is explicitly passed
```
stefreak, line 583: Can we add a lint rule for this? e.g. https://eslint.org/docs/latest/rules/no-unneeded-ternary

core/test/integ/src/plugins/kubernetes/commands/pull-image.ts
```diff
@@ -85,7 +85,7 @@ describe.skip("pull-image plugin command", () => {
         actionKind: resolvedAction.kind,
       })
 
-      builder.handlers.build!({
+      await builder.handlers.build!({
```
stefreak, line 88: We should add eslint rule `https://typescript-eslint.io/rules/no-floating-promises/` to catch this one

core/test/unit/src/logger/log-entry.ts
```diff
@@ -405,7 +405,7 @@ describe("ActionLog", () => {
       })
       const entry = testLog.info("hello").getLatestEntry()
 
-      expect(entry.key).to.be.a.string
+      expect(entry.key).to.be.a("string")
```
stefreak, line 408: Hm, I did not found what the issue is here exactly – the docs use your syntax; did the old syntax have no effect?
Are we missing a linter rule to catch the mistake here too, e.g. to detect accessing an undefined property?

stefreak (changes requested):
Thank you for making these changes. I'd suggest adding linter rules if possible, to avoid toil like this in the future.

core/test/unit/src/logger/log-entry.ts
vvagaytsev: I tested it manually, and the old statement had no effect.

core/test/integ/src/plugins/kubernetes/commands/pull-image.ts
vvagaytsev: We have this rule enabled, it reports warnings instead of errors. I'll check if we can set it to error-level.

core/test/unit/src/logger/log-entry.ts
stefreak: Thank you! Maybe a rule like https://typescript-eslint.io/rules/no-unused-expressions could help here. If you want we can pair on looking at the config, to see why the linter did not complain.

stefreak: There is also https://www.npmjs.com/package/eslint-plugin-chai-friendly – it might be that this plugin is aware of the cases where unused expressions are allowed in Chai.

core/src/plugins.ts
vvagaytsev: It does not. The `else`-clause is unnecessary here because of the `continue` statement.

core/src/cloud/api.ts
vvagaytsev: Added in 5188a9fa4312ea8cd05ea8cde62aa34e5c6f122a.

core/test/integ/src/plugins/kubernetes/commands/pull-image.ts
vvagaytsev: Done in 2fdda16bb5dfa65a9a388177ba8daccbfd4efd98.

core/test/unit/src/logger/log-entry.ts
vvagaytsev: Here is the signature of `string`:
```
Chai.Assertion.string(string: string, message?: string | undefined): Chai.Assertion
```

It must be called with some args.

vvagaytsev: It looks like it can be addressed with https://eslint.org/docs/latest/rules/no-unused-expressions rule.

vvagaytsev: Just tried that rule and it gave 850 errors :) I think we can postpone that rule and try it out later.

stefreak (approved):
Awesome, looks good to me!

PR #5780, opened 2024-02-26
feat: added profile-project command

thsig:

Added a simple utility command for profiling projects, the `profile-project` command.

This command logs the number of files included in each action and module in the project (omitting actions that were converted from modules to avoid duplication), their include/exclude configs (if any), and finally the number of modules, actions and tracked files in the project.

This should be useful when helping users with large projects diagnose slow init performance and how to best structure their includes, excludes and `.gardenignore`.


edvald:
Could we move this to `garden util profile-project`? Feels like an odd one to put top level.

edvald (approved):

PR #5794, opened 2024-03-01
fix(k8s): fix kubernetes workload rollout status check

twelvemo:
For checking the rollout status of workloads we were using the `spec.replicas` and not the `status.replicas` in some cases. This means that during a rollout we would not catch the case where there are currently more `replicas` than `updated` replicas. If we compare the `spec.replicas` to `updated` it is possible that both have the same number, even though `updated` has not passed it's readiness probe yet, which is the trigger for the old pods to be terminated. Only once `status.replicas` and `status.updated` match are we truly done. This led to garden not waiting for deployment rollouts in some cases, which could lead to further errors trying to run execs or tests in non-running pods or pods being canceled mid-execution, due to them being terminated as old pods.

This is also implemented in this way in kubectl see [here](https://github.com/kubernetes/kubernetes/blob/331ced5606fe32d8beb77c974b28dc12f7795012/staging/src/k8s.io/kubectl/pkg/polymorphichelpers/rollout_status.go#L83).

This is an old bug, but it only surfaced recently since #4846 where we removed `compareDeployedResources` from the status checks for kubernetes type actions.

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/status/workload.ts
+++ b/core/src/plugins/kubernetes/status/workload.ts
@@ -186,8 +186,8 @@ async function getRolloutStatus(workload: Workload) {
 
   out.state = "ready"
 
-  // See `https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/rollout_status.go` for a reference
-  // for this logic.
+  // See here as a reference for this logic:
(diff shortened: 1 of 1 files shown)


stefreak (approved):
Wow, good catch!

PR #5809, opened 2024-03-05
improvement: improvements to validate command

### LINKED ISSUE #5695: 0.13: [Bug]: Garden validate fails upon checking templated variables with terraform plugin
### Current Behavior

After doing some tests with the terraform plugin and not understanding why ''garden validate" failed, I finally discovered that the command runs checks on the actual values of the templated variables for ${actions.deploy.[action-name].outputs} while validating the garden config. 

Because of this, in the event that the validate step is run before the terraform plugin apply-action step is done, the validate step will fail regardless if the garden configuration is correct. 

This issue was not present in 0.12. Please take a look at this as soon as possible.

### Expected behavior

Validate should have worked without needing to access the individual variable values since the purpose of the validate command is just to check the configuration to see if the syntax is correct. If this change was intended, then please state why because this seems very much like feature regression compared to 0.12.

### Reproducible example

**Validate pre-apply**

```
$ garden validate --env=dev --yes

ℹ deploy.pricing-pg-backup  → Aborting because upstream dependency failed.
ℹ deploy.a-pricing        → Aborting because upstream dependency failed.
ℹ run.pricing-migrate       → Aborting because upstream dependency failed.
ℹ run.pricing-reseed        → Aborting because upstream dependency failed.
✖ deploy.pricing-pg         → Failed processing resolve Deploy type=helm name=pricing-pg (took 6.65 sec). This is what happened:

──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── 
Invalid template string (${actions.deploy.infrastructure-aw…): Could not find key pricing_db_password under actions.deploy.infrastructure-aws.outputs. Available keys: (none).
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── 
Failed to resolve Deploy type=helm name=b-pg: Error: resolve Deploy type=helm name=b-pg failed: Error: Invalid template string (${actions.deploy.infrastructure-aw…): Could not find key b_db_password under actions.deploy.infrastructure-aws.outputs. Available keys: (none).
```

**Validate post-apply**

```
✖ run.a-config-reseed     → Failed processing resolve Run type=kubernetes-pod name=a-config-reseed (took 8.9 sec). This is what happened:

──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── 
Invalid template string (${actions.deploy.infrastructure-aw…): Could not find key access_key under actions.deploy.infrastructure-aws.outputs. Available keys: b_client_registry, b_db_password, b_s3_document, b_s3_export, b_server_registry, keycloak_admin_password, keycloak_db_password, keycloak_registry, a_client_registry, a_db_password, a_pricing_registry, a_s3_document, a_s3_export, a_s3_pricing_backup, a_server_registry and pricing_db_password.
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── 
Failed to resolve Run type=kubernetes-pod name=b-migrate: Error: resolve Run type=kubernetes-pod name=b-migrate failed: Error: Invalid template string (${actions.deploy.infrastructure-aw…): Could not find key access_key under actions.deploy.infrastructure-aws.outputs. Available keys: b_client_registry, b_db_password, b_s3_document, b_s3_export, b_server_registry, keycloak_admin_password, keycloak_db_password, keycloak_registry, a_client_registry, a_db_password, a_pricing_registry, a_s3_document, a_s3_export, a_s3_pricing_backup, a_server_registry and pricing_db_password.
```

Based on these 2 outputs, it's clear that the validate step is not working like it was previously doing as for our current production Gitlab CI pipeline which is on garden 0.12, the validate step does not check the variable for the output values to be present.

Now with 0.13 this has changed to being the validate step checking not just syntax but if the values exist as well which interferes with just the syntax checking. Was this change intended? If this is so, having templated values for terraform outputs without applying the plan to the garden deployment before actually deploying will make validating the configuration pointless as it will just be false positives.

### Workaround

I can run validate after running terraform apply-plugin but validate should have been able to successfully scan the configuration without needing to check whether the values are present or not. If variable checks are necessary to check values within the template variables, then that should be an error that would come up during the actual deploy/run execution time as that would make sense to happen then.

### Suggested solution(s)

Please only set the validate command to only check syntax and nothing else. Value checks are something that can be present in the error messages during the runtime of the deploy/run action executions. Having the variable checks done during the validate step will just create confusion as it would be an indication that the garden config syntax is incorrect somewhere when clearly it is not in this case.

### Your environment

* **OS:** 
Windows 11 + WSL Ubuntu 22.04.3 LTS
Terraform + AWS

garden version: 0.13.24

thsig:

Before this commit, the `garden validate` command would fully resolve the config graph. This caused problems for users that had references to runtime outputs of actions in their action configs, since fully resolving the graph in such cases means executing the action (e.g. actually building/deploying).

Here, we change the default behavior of the validate command to only partially resolve the config graph, but also introduce a `--resolve` CLI option for resolving some or all actions fully. This allows users to fully resolve action configs for debugging purposes, but also to sidestep action configs with runtime output references.

Fixes #5695.

### DIFF
--- a/core/src/commands/validate.ts
+++ b/core/src/commands/validate.ts
@@ -9,11 +9,26 @@
 import type { CommandParams, CommandResult } from "./base.js"
 import { Command } from "./base.js"
 import { printEmoji, printHeader } from "../logger/util.js"
-import { dedent } from "../util/string.js"
+import { dedent, deline } from "../util/string.js"
 import { styles } from "../logger/styles.js"
 import { resolveWorkflowConfig } from "../config/workflow.js"
+import { StringsParameter } from "../cli/params.js"
+import type { ConfigGraph } from "../graph/config-graph.js"
 
-export class ValidateCommand extends Command {
+const validateOpts = {
+  resolve: new StringsParameter({
+    help: deline`
+      Fully resolve a specific action, including references to runtime outputs from other actions. Actions should be specified as \`<kind>.<name>\` (e.g. \`deploy.my-service\` or \`build.my-image\`). This option can be specified multiple times to fully resolve multiple actions. Use * to fully resolve all actions. Note that this may result in actions being executed during validation (e.g. if a runtime output is referenced by another action, it will be executed in order to fully resolve the config). In such cases, we recommend not using this option.
+    `,
+    getSuggestions: ({ configDump }) => {
+      return Object.keys(configDump.actionConfigs)
+    },
+  }),
+}
+
+type Opts = typeof validateOpts
+
+export class ValidateCommand extends Command<{}, Opts> {
   name = "validate"
   help = "Check your garden configuration for errors."
   emoji = "✔️"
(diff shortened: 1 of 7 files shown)


core/src/commands/validate.ts
```diff
@@ -22,15 +37,28 @@ export class ValidateCommand extends Command {
 
   override description = dedent`
     Throws an error and exits with code 1 if something's not right in your garden config files.
+
+    Examples:
+
+        garden validate                              # validate all configs, but don't fully resolve any actions
+        garden validate --resolve build.my-image     # same as above, but fully resolve the build.my-image action
+        garden validate --resolve deploy.my-service
```
vvagaytsev, line 45: Should we add an explanation to this example too?

vvagaytsev (approved):
Thanks! 💯 
I've left a minor non-blocking comment. Otherwise looks great!

PR #5810, opened 2024-03-06
fix(k8s): use the same service account for pulling images as building

### LINKED ISSUE #5591: 0.13: [Bug]: `garden plugins kubernetes pull-image` does not work with IMDSv2 and hop limit of 1 (IRSA)
## Garden Bonsai (0.13) Bug

Garden pull image does not use existing buildkit or kaniko util container with IRSA policy.

### Current Behavior

If IMDSv2 with hop limit of 1 is used, the `garden plugins kubernetes pull-image` fails as it spawns a new util Pod without using the serviceAccount as necessary for IRSA to work. 

### Expected behavior

`pull-image` should simply succeed. This can be accomplished by using the existing util container instead of spawning a new one.

### Reproducible example

IRSA-enabled project

### Workaround
Unknown

### Suggested solution(s)

Don't spawn a new util container Pod for the skopeo pull command, but use the existing in-cluster-building util Pod.

https://github.com/garden-io/garden/blob/main/core/src/plugins/kubernetes/commands/pull-image.ts#L131

### Your environment

* **OS:** irrelevant

`garden version` 0.13.23

twelvemo:
The pull-image command spawns a pod for each image to pull. If you have set-up IRSA or Google Workload Identity for in-cluster building these pods also need to use the same service account as for in-cluster building.

Fixes #5591

### DIFF
--- a/core/src/plugins/kubernetes/commands/pull-image.ts
+++ b/core/src/plugins/kubernetes/commands/pull-image.ts
@@ -20,7 +20,12 @@ import { dockerAuthSecretKey, getK8sUtilImageName, systemDockerAuthSecretName } 
 import { getAppNamespace, getSystemNamespace } from "../namespace.js"
 import { randomString } from "../../../util/string.js"
 import type { PluginContext } from "../../../plugin-context.js"
-import { ensureBuilderSecret, ensureServiceAccount, inClusterBuilderServiceAccount } from "../container/build/common.js"
+import {
+  ensureBuilderSecret,
+  ensureServiceAccount,
+  getBuilderServiceAccountAnnotations,
+  inClusterBuilderServiceAccount,
+} from "../container/build/common.js"
 import type { ContainerBuildAction } from "../../container/config.js"
 import { k8sGetContainerBuildActionOutputs } from "../container/handlers.js"
 import type { Resolved } from "../../../actions/types.js"
@@ -133,7 +138,7 @@ async function pullFromExternalRegistry({ ctx, log, localId, remoteId }: PullPar
     log,
     api,
     namespace,
-    annotations: ctx.provider.config.kaniko?.serviceAccountAnnotations,
+    annotations: getBuilderServiceAccountAnnotations(ctx.provider.config),
   })
 
   const runner = new PodRunner({
--- a/core/src/plugins/kubernetes/commands/pull-image.ts
+++ b/core/src/plugins/kubernetes/commands/pull-image.ts
@@ -20,7 +20,7 @@ import { dockerAuthSecretKey, getK8sUtilImageName, systemDockerAuthSecretName } 
 import { getAppNamespace, getSystemNamespace } from "../namespace.js"
 import { randomString } from "../../../util/string.js"
 import type { PluginContext } from "../../../plugin-context.js"
-import { ensureBuilderSecret } from "../container/build/common.js"
+import { ensureBuilderSecret, inClusterBuilderServiceAccount } from "../container/build/common.js"
 import type { ContainerBuildAction } from "../../container/config.js"
 import { k8sGetContainerBuildActionOutputs } from "../container/handlers.js"
 import type { Resolved } from "../../../actions/types.js"
@@ -141,6 +141,7 @@ async function pullFromExternalRegistry({ ctx, log, localId, remoteId }: PullPar
         namespace,
       },
       spec: {
+        serviceAccountName: inClusterBuilderServiceAccount,
         containers: [
           {
             name: "main",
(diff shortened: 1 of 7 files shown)


core/src/plugins/kubernetes/commands/pull-image.ts
```diff
@@ -128,6 +128,14 @@ async function pullFromExternalRegistry({ ctx, log, localId, remoteId }: PullPar
     `docker-archive:${tmpTarPath}:${localId}`,
   ]
 
+  await ensureServiceAccount({
```
stefreak, line 131: Instead of creating a new util deployment, could we instead use the existing methods to create a remote-building util deployment? Or can't we do this because it needs to work with local-docker too? 🤔

core/src/plugins/kubernetes/commands/pull-image.ts
```diff
@@ -128,6 +128,14 @@ async function pullFromExternalRegistry({ ctx, log, localId, remoteId }: PullPar
...
+    ctx,
+    log,
+    api,
+    namespace,
+    annotations: ctx.provider.config.kaniko?.serviceAccountAnnotations,
```
stefreak, line 136: What if the `buildMode` setting is `cluster-buildkit`?

core/src/plugins/kubernetes/commands/pull-image.ts
twelvemo: We create one util pod per pulled image. The image-pull command only allows pulling all your images at the same time, so it won't do to use an existing util deployment even if that exists.

core/src/plugins/kubernetes/commands/pull-image.ts
twelvemo: annotations can be undefined, then we create the service account spec without any annotations, see https://github.com/garden-io/garden/blob/010730cefb279de8dfc966afa43a0e8f704a3678/core/src/plugins/kubernetes/container/build/common.ts#L502

stefreak: But what if the user is using buildkit, and specified the service account annotations there? I think we should call existing method to deploy util/build pods and then exec into the existing util container, instead of creating one just for pull-image.

twelvemo: we always create the service account for `kaniko` and `cluster-buildkit`. For `cluster-buildkit` without any annotations.

stefreak: Also, even if we support both modes in pull-image, it will be easy to overlook when we implement a new mode later.

stefreak: Ah ok, cluster buildkit has no annotations? 🤔 Then I seem to misunderstand how IRSA works? 🤔 Shall we pair on this @twelvemo?

core/src/plugins/kubernetes/commands/pull-image.ts
stefreak: Can't we run multiple commands in the same Pod using `exec`? What is the problem of re-using the same Pod? Are we worrying about filename collisions?

core/src/plugins/kubernetes/commands/pull-image.ts
twelvemo: Oh you are right, i misremembered of course we support cluster-buildkit for IRSA as well. Will push a fix.

core/src/plugins/kubernetes/commands/pull-image.ts
twelvemo: It is slow as is, distributing it via multiple pods and thereby nodes seems like a good idea.

stefreak: Makes sense. If that's crucial, I suggest we re-use the same manifest definitions instead of duplicating the code, so we can't forget to update pull-image next time we make a change to util deployments.

stefreak: At least we could move it to `common.ts`, so it lives next to the other manifests

twelvemo: Yes that is a nice idea, however the util deplyoment from common is a deployment, while here we only spawn a pod that runs to completion.

stefreak: I see there are differences, but they also have settings in common. Can we at least co-locate the code in the same file?

stefreak: Looking at the code together, we agreed that it's enough to ensure that we always call `getBuilderServiceAccountAnnotations`, even from Kaniko and Buildkit.

stefreak (approved):
👍 Good work

PR #5811, opened 2024-03-07
fix(core): properly escape shell commands

### LINKED ISSUE #4081: Cluster buildkit with buildarg is empty string ''
In case a build arg is empty string '' incluster build with cluster buildkit fails

MY_ARG=''

invalid opt: invalid value build-arg:MY_ARG

because its forwards as build-arg:MY_ARG and expects build-arg:MY_ARG='' 

and i did not found any way to work around this since the argument on our side is sometimes an empty string, setting it to null has other drawbacks because some languages convert that into string

stefreak:

Shell commands need to be carefully constructed to avoid command
injection issues, and also to avoid issues with the shell interpreting
input strings, e.g. parenthesis or glob characters.

If possible, we should pass a command list for executing commands.

For untrusted input values, one should use environment variables and
access them from the shell script. Make sure to access them wrapped
around double quotes, otherwise the script is prone to command injection
attacks.

Sometimes it is needed to construct a shell script to chain multuiple
commands together. In this case, this commit introduces a utility
function `commandListToShellScript`.

It wraps every parameter in single quotes, escaping contained single quotes (for use in bash scripts). Joins the elements with a space character.

 Examples:

```
// returns `echo 'hello world'`
commandListToShellScript(["echo", "hello world"])

// returns `echo 'hello'"'"'world'`
commandListToShellScript(["echo", "hello'world"])

// returns `echo ''"'"'; exec ls /'`
commandListToShellScript(["echo", "'; exec ls /"])
```

Caveat: This is only safe if the command is directly executed. It is not safe, if you wrap the output of this in double quotes, for instance.

```
// SAFE
exec(["sh", "-c", ${commandListToShellScript(["some", "command", "--with" untrustedInput])}])
exec(["sh", "-c", dedent`
    set -e
    echo "running command..."
    ${commandListToShellScript(["some", "command", "--with" untrustedInput])}
    echo "done"
`])

// UNSAFE! don't do this

const commandWithUntrustedInput = commandListToShellScript(["some", "command", "--with" untrustedInput])
exec(["sh", "-c", `some_var="${commandWithUntrustedInput}"; echo "$some_var"`])
```

The second is UNSAFE, because we can't know that the /double quotes/ need to be escaped here.

If you can, use environment variables instead of this, to pass untrusted values to shell scripts, e.g. if you do not need to construct a command with untrusted input.

```
// SAFE
exec(["sh", "-c", `some_var="$UNTRUSTED_INPUT"; echo "$some_var"`], { env: { UNTRUSTED_INPUT: untrustedInput } })
```

Swooped in: Fixes #4081

### DIFF
--- a/core/test/unit/src/util/escape.ts
+++ b/core/test/unit/src/util/escape.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018-2023 Garden Technologies, Inc. <info@garden.io>
+ * Copyright (C) 2018-2024 Garden Technologies, Inc. <info@garden.io>
  *
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
--- a/core/test/unit/src/util/escape.ts
+++ b/core/test/unit/src/util/escape.ts
@@ -1,0 +1,36 @@
+/*
+ * Copyright (C) 2018-2023 Garden Technologies, Inc. <info@garden.io>
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { expect } from "chai"
+import { commandListToShellScript } from "../../../../src/util/escape.js"
+
+describe("commandListToShellScript", () => {
+  it("transforms a list of command line arguments to a shell script", () => {
+    const commandList = ["echo", "hello", "world"]
+    const commandString = commandListToShellScript(commandList)
+    expect(commandString).to.equal("'echo' 'hello' 'world'")
+  })
+
+  it("escapes single quotes in command line arguments", () => {
+    const commandList = ["echo", "hello", "world's"]
+    const commandString = commandListToShellScript(commandList)
+    expect(commandString).to.equal(`'echo' 'hello' 'world'"'"'s'`)
+  })
+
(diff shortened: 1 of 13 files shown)


core/test/unit/src/util/escape.ts
```diff
@@ -0,0 +1,36 @@
...
+  it("escapes single quotes in command line arguments", () => {
+    const commandList = ["echo", "hello", "world's"]
+    const commandString = commandListToShellScript(commandList)
+    expect(commandString).to.equal(`'echo' 'hello' 'world'"'"'s'`)
+  })
+
+  it("replaces all single quotes", () => {
+    const commandList = ["echo", "'''"]
+    const commandString = commandListToShellScript(commandList)
+    expect(commandString).to.equal(`'echo' ''"'"''"'"''"'"''`)
```
eysi09, line 28: :)

thsig (approved):
Looks good to me, thanks for digging into this!

thsig:
Looks like there are a few test cases that need to be updated (which assume non-quoted strings, where they're now single-quoted). But once that's fixed, we should be good to go here.

eysi09 (commented):
This looks great to be but it would be great to get more eyes on this (e.g. @twelvemo) due to the security implications.

core/test/unit/src/util/escape.ts
stefreak: it works though :) 

```
sh-3.2$ 'echo' ''"'"''"'"''"'"''
'''
```

PR #5827, opened 2024-03-13
feat(helm): store garden metadata in configmap instead of helm values

### LINKED ISSUE #3962: Deploying Helm charts that use strict value validation breaks with Garden
### Current Behavior

It is not possible to install helm charts with Garden that use strict values validation via a values.schema.json that include the `additionalProperties: false` stanza. For an overview of json schema validation for helm values see [here](https://helm.sh/docs/topics/charts/#schema-files).
Garden adds custom values to a chart deployed with Garden to add some metadata to the helm release. This is an example of a values file that Garden produces:

```
service:
  type: LoadBalancer
  ports:
    - name: https
      port: 443
      protocol: TCP
      targetPort: 443
.garden:
  moduleName: istio-gateway
  projectName: istio-example
  version: v-5f013b4127
```

When `additionalProperties: false` is set in the values.schema.json of a chart the deploy will fail with following error:

```
Error: INSTALLATION FAILED: values don't meet the specifications of the schema(s) in the following chart(s):
gateway:
- (root): Additional property .garden is not allowed
```
### Expected behavior

Garden deploys the helm chart.

### Reproducible example

Deploy the istio gateway Helm Chart with Garden e.g.:

```
kind: Deploy
name: istio-gateway
description: Install istio gateway helm chart
type: helm
spec:
  chart:
    name: gateway
    version: 1.17.1
    repo: https://istio-release.storage.googleapis.com/charts
```

### Workaround

If you maintain the chart, consider setting `additionalProperties: true` in the validation schema. If it is a public chart the only way to install it is to download it and store it locally and set `additionalProperties: true`.
There is an open issue in Helm like https://github.com/helm/helm/issues/10398 to override  the validation client-side.

### Suggested solution(s)

Store the Garden related metadata in a configmap or secret in the same namespace similar to test results instead of Helm values.

### Additional context

If you run into this error please give the issue a thumbsup, currently we only saw this with the istio chart.

### Your environment

* **OS:** 

`garden version`

twelvemo:
Store garden metadata for Helm deploy actions in a config map in the same namespace instead of adding them as additional values to the Helm release, which caused schema validation to fail for Helm charts with `additionalProperties: false` in the `values.schema.json`.

Also fixes a skipped integration test on `localMode` for Helm Charts by updating the example project from modules to actions.

Fixes #3962

### DIFF
--- a/core/src/plugins/kubernetes/helm/status.ts
+++ b/core/src/plugins/kubernetes/helm/status.ts
@@ -304,5 +304,8 @@ export async function getHelmGardenMetadataConfigMapData({
     name: `garden-helm-metadata-${action.name}`,
     namespace,
   })
-  return deserializeValues(gardenMetadataConfigMap.data!) as HelmGardenMetadataConfigMapData
+  if (!gardenMetadataConfigMap.data) {
+    throw new Error(`Configmap with garden metadata for release ${action.name} is empty`)
+  }
+  return deserializeValues(gardenMetadataConfigMap.data) as HelmGardenMetadataConfigMapData
 }
--- a/core/src/plugins/kubernetes/helm/status.ts
+++ b/core/src/plugins/kubernetes/helm/status.ts
@@ -9,6 +9,7 @@
 import type { ForwardablePort, ServiceIngress, DeployState, ServiceStatus } from "../../../types/service.js"
 import type { Log } from "../../../logger/log-entry.js"
 import { helm } from "./helm-cli.js"
+import type { HelmGardenMetadataConfigMapData } from "./common.js"
 import { getReleaseName, loadTemplate } from "./common.js"
 import type { KubernetesPluginContext } from "../config.js"
 import { getForwardablePorts } from "../port-forward.js"
@@ -25,6 +26,7 @@ import { deployStateToActionState } from "../../../plugin/handlers/Deploy/get-st
 import { isTruthy } from "../../../util/util.js"
 import { ChildProcessError } from "../../../exceptions.js"
 import { gardenAnnotationKey } from "../../../util/string.js"
+import { deserializeValues } from "../../../util/serialization.js"
 
 export const gardenCloudAECPauseAnnotation = gardenAnnotationKey("aec-status")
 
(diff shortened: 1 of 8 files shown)


core/test/integ/src/plugins/kubernetes/helm/deployment.ts
```diff
@@ -280,6 +285,90 @@ describe("helmDeploy", () => {
...
+      ctx,
+      action,
+      releaseName,
+      log: garden.log,
+    })
+
+    expect(releaseStatus.state).to.equal("ready")
+
+    await deleteHelmDeploy({ ctx, log: actionLog, action })
+    expect(getHelmGardenMetadataConfigMapData({ ctx, action, log: actionLog, namespace })).to.throw
```
vvagaytsev, line 369: This won't work. For sync functions, an error-throwing one must be wrapped into a lambda or an anonymous function, and the `throw` function in the end of the statement must be called:
```js
expect(() => getHelmGardenMetadataConfigMapData({ ctx, action, log: actionLog, namespace })).to.throw()
```

See:
* https://stackoverflow.com/a/49347521/2753863
* https://stackoverflow.com/a/64260632/2753863

For async functions, we should do it in a different way. We can use our pown helper that has been used a lot across the tests:
```js
await expectError(async () => await getHelmGardenMetadataConfigMapData({ ctx, action, log: actionLog, namespace }))
```

vvagaytsev (changes requested):

core/src/plugins/kubernetes/helm/status.ts
```diff
@@ -304,5 +304,8 @@ export async function getHelmGardenMetadataConfigMapData({
     name: `garden-helm-metadata-${action.name}`,
     namespace,
   })
-  return deserializeValues(gardenMetadataConfigMap.data!) as HelmGardenMetadataConfigMapData
+  if (!gardenMetadataConfigMap.data) {
+    throw new Error(`Configmap with garden metadata for release ${action.name} is empty`)
```
vvagaytsev, line 308: Should we throw an instance of some `GardenError` here? Maybe `RuntimeError` is good enough.

vvagaytsev (approved):
Thank you! 💯

PR #5829, opened 2024-03-14
feat(container): allow global extra build flags e.g. for custom remote builders

stefreak:

Allows overriding the docker command, and setting global extra build flags.

This can be useful when configuring custom remote builders.

Example:

```
providers:
 - name: exec
   initScript: |
      docker buildx create --driver ... custom-builder
 - name: container
   dockerBuildExtraFlags: ["--builder", "custom-builder", "--push"]
```

This is useful as a first step for experimental Namespace builder support.

This PR also changes the Kubernetes provider initialization to allow missing `deploymentRegistry` config flag in remote kubernetes clusters. It now prints a warning instead of throwing an error.

### DIFF
--- a/core/src/plugins/container/container.ts
+++ b/core/src/plugins/container/container.ts
@@ -423,7 +423,7 @@ export const gardenPlugin = () =>
       Provides the \`container\` actions and module type.
       _Note that this provider is currently automatically included, and you do not need to configure it in your project configuration._
     `,
-    configSchema: providerConfigBaseSchema(),
+    configSchema: configSchema(),
 
     createActionTypes: {
       Build: [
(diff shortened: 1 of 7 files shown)


core/src/plugins/container/container.ts
```diff
@@ -42,11 +42,31 @@ import { publishContainerBuild } from "./publish.js"
...
-export type ContainerProviderConfig = GenericProviderConfig
+export interface ContainerProviderConfig extends GenericProviderConfig {
+  customDockerCommand?: string[]
+  dockerBuildExtraFlags?: string[]
+}
+
+export const configSchema = () =>
+  providerConfigBaseSchema()
+    .keys({
+      customDockerCommand: joi.array().items(joi.string()).description(dedent`
```
twelvemo, line 58: Does it make sense for this to be an array if you should only use it to specify your custom binary in a string? Any additional args would be passed into `dockerBuildExtraFlags`.

twelvemo (changes requested):

twelvemo (approved):

core/src/plugins/container/container.ts
twelvemo: We went ahead and removed the `customDockerCommand` as it introduces untested configurations with other container image builders. Only adding `dockerBuildExtraFlags` in this PR.

vvagaytsev (approved):
👍

PR #5867, opened 2024-03-25
fix(vcs): use structural path comparison to compute minimal repo roots

vvagaytsev:

This PR includes the following changes:

* fix sub-path matching in `VcsHandler.getMinimalRoots()`
* improve some logging (log duration of the project scanning on the verbose-level)
* refactor `GitHandler` by moving out some utility methods that are not related to the class state and do not require a class instance
* remove some unused code
* make some types stricter and explicit

See individual commits for the details.


core/src/vcs/vcs.ts
```diff
@@ -462,3 +463,36 @@ export function getSourcePath(config: ModuleConfig | BaseActionConfig) {
...
+ * Sub-path means that a candidate must be located inside a reference path.
+ *
+ * Both {@path} and {@ sybPathCandidate} must be absolute paths
+ *
+ * @param path the reference path (absolute)
+ * @param subPathCandidate the path to be checked (absolute)
+ */
+export function isSubPath(path: string, subPathCandidate: string): boolean {
+  if (!isAbsolute(path)) {
+    throw new RuntimeError({ message: `Expected absolute path, but got: ${path}` })
```
thsig, line 478: Would be good to indicate in these error messages which path (`path` or `subPathCandidate`) caused the error.

thsig (approved):
Left one minor comment there, but looks good otherwise!

core/src/plugins/container/helpers.ts
```diff
@@ -413,7 +413,7 @@ const helpers = {
     const dockerfile = action.getSpec("dockerfile")
     // NOTE: it's important to check for the files existence in the build path to allow dynamically copying the Dockerfile from other actions using `copyFrom`.
     const dockerfileSourcePath = getDockerfilePath(action.getBuildPath(), dockerfile)
-    return await pathExists(dockerfileSourcePath)
+    return action.getFullVersion().files.includes(dockerfileSourcePath)
```
stefreak, line 416: Are files copied using `copyFrom` listed in `getFullVersion().files`?

stefreak: I think they are not, because `copyFrom` allows globs, and it copies from the build directory of so we can't know statically

vvagaytsev: I deleted this commit.

stefreak (approved):

stefreak:
@vvagaytsev nice work 🎈

Issue #2382, opened 2021-05-08
"Attempted to copy non-file" when a symlink target is a folder


orarbel:
## Bug

Building breaks when a module contains a symlink to a folder.

The offending line only checks if the target of a symlink is a file, but doesn't check if it's a folder.

https://github.com/garden-io/garden/blob/893e55b6e4932dda671509248ce2acf513eb8526/core/src/build-staging/helpers.ts#L68

### Current Behavior

An error of "Attempted to copy non-file" is raised.

### Expected behavior

The build will be successful.

### Reproducible example

Projects that contain a symlink to a folder.

### Suggested solution(s)

Add a condition that checks if the target is a folder and if so don't raise an error.

### Your environment

* **OS:** macOS BigSur 11.3
* **How I'm running Kubernetes:** Docker Desktop

`garden version`
0.12.21

eysi09:
Thanks for the detailed report @orarbel. Should be a quick fix on our end.

eysi09:
@edvald, could you confirm that the suggestion makes sense? If that's the case I can fix it on my end.

edvald:
Yeah I think that does make sense, but I wanna take a quick look to make sure. Lots of little devils in the details in this stuff .)

eysi09:
All right, I'll assign this to you then. Feel free to re-assign.

vvagaytsev:
@jondeandres is this still a problem in `0.13`?

abMatGit:
@vvagaytsev hey I can attest this is also a problem for me. In `0.13` this code is still checking only if it symlinks a file not a folder. 

This would be fixed if the code was: 
```typescript
if (!sourceStats.isFile() || !sourceStats.isDirectory()) { 
```

alexkuretz:
Getting this trying to build backstage via container action with garden 0.13.39
```
✖ build.backstage           → Failed processing Build type=container name=backstage (took 10.35 sec). This is what happened:

────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Error while copying from '/home/user/work/backstage/backstage/node_modules/app' to '/home/user/work/backstage/.garden/build/backstage/node_modules/app': Source is neither a symbolic link, nor a file.
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
1 requested deploy action(s) aborted!
```
It is indeed a symlink:
```
user@computer:~/work/backstage/backstage$ ll node_modules/app
[...]

PR #5890, opened 2024-03-28
feat(modules): allow opting out of build staging

thsig:

Here, we make the `local` config field (previously only available to `exec` modules) available for all module types.

When used, it maps to the `buildAtSource` field of any generated Build actions in the module conversion process.

This is done because some users/projects that haven't migrated from modules to actions yet need to reference symlinked directories in their builds, where the simplest solution appears to be to simply opt out of build staging for those modules.

Workaround for #2382.

**Notes for reviewer**:

Will this cause problems with build copying? Does allowing this violate any subtle invariants?

### DIFF
--- a/core/src/config/module.ts
+++ b/core/src/config/module.ts
@@ -92,6 +92,7 @@ interface ModuleSpecCommon {
   apiVersion?: string
   allowPublish?: boolean
   build?: BaseBuildSpec
+  local?: boolean
   description?: string
   disabled?: boolean
   exclude?: string[]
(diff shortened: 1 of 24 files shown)


core/src/config/module.ts
```diff
@@ -187,6 +188,20 @@ export const coreModuleSpecSchema = createSchema({
...
   build: baseBuildSpecSchema().unknown(true),
+  local: joi
+    .boolean()
+    .description(
+      dedent`
+      If set to true, Garden will run the build command, services, tests, and tasks in the module source directory,
+      instead of in the Garden build directory (under .garden/build/<module-name>).
+
+      Garden will therefore not stage the build for local modules. This means that include/exclude filters
+      and ignore files are not applied to local modules.
```
edvald, line 199:
```suggestion
      and ignore files are not applied to local modules, except to calculate the module/action versions.
```

core/src/config/module.ts
```diff
@@ -187,6 +188,20 @@ export const coreModuleSpecSchema = createSchema({
...
+
+      Note: \`local\` This maps to the \`buildAtSource\` option in this module's generated Build action (if any).
```
edvald, line 201:
```suggestion
      Note: This maps to the \`buildAtSource\` option in this module's generated Build action (if any).
```

edvald (approved):

thsig:
There were a few test failures, should be good to go now @edvald.

edvald (approved):

edvald (approved):

PR #5903, opened 2024-04-04
improvement(k8s): better pod runner error handling

thsig:

Before this fix, we weren't properly propagating pod runner errors when performing status checks for in-cluster builds.

We also make some tweaks to the logic that checks if the runner pod was killed. Instead of using the local time at the start of the pod run as the reference/cutoff, we now use the latest event timestamp before exec-ing (which should make the logic robust against timezone differences).

Co-authored with @stefreak.


core/src/plugins/kubernetes/run.ts
```diff
@@ -1237,13 +1239,23 @@ export class PodRunner {
...
     const events = await getResourceEvents(this.api, this.pod)
     if (
       // If reason is killed and lastTimestamp doesn't exist or is greater than afterTime
-      some(events, (event) => event.reason === "Killing" && (!event.lastTimestamp || event.lastTimestamp > afterTime))
+      some(events, (event) => {
+        // only consider events that happened after the last event we've seen, and if they have a lastTimestamp property.
+        // we ignore all events without lastTimestamp, because that could lead to permanently unusable Pods.
+        if (lastEventTime && event.lastTimestamp) {
+          return event.lastTimestamp > lastEventTime && event.reason === "Killing"
+        }
```
stefreak, line 1255: I think keeping the time check separate from reason comparison makes the code easier to understand. I think this should have the same effect:

```suggestion
        if (lastEventTime && event.lastTimestamp && event.lastTimestamp > lastEventTime) {
          return false
        }
```

stefreak: Maybe it's more clear if we move the time check to a `.filter` call directly after `getResourceEvents`

vvagaytsev (approved):

PR #5907, opened 2024-04-05
refactor(git): collect Git utils in one class

vvagaytsev:
@stefreak we discussed this yesterday. This PR introduces a standalone class that holds convenience methods for Git operations and supports arbitrary commands execution. Such approach allows to avoid passing the Git executor instance to every utility function.

Continuation of #5902.


stefreak (approved):

stefreak:
Nice, really enjoying these small refactorings that make the code a little bit more straight forward.

PR #5916, opened 2024-04-09
fix(git): increase max proc buffer size and fix error handling

vvagaytsev:

The project scan fails with the following error for large projects:
```console
Command "git --glob-pathspecs ls-files --ignored --cached --exclude /workspace/all-the-things/.garden --exclude **/.garden/**/* --exclude-per-directory .gardenignore" failed with code undefined:

Here are the last 100 lines of the output:

// the logs have been omitted, there were no actual error details
```

We run the underlying git command with some custom options, and we have used the 10MB buffer instead of the default 100MB. That can be the cause of the problem. Let's merge it and ask the issue reporter to try out the edge build.

Fixes #

### DIFF
--- a/core/src/exceptions.ts
+++ b/core/src/exceptions.ts
@@ -77,7 +77,7 @@ export function isEAddrInUseException(err: any): err is EAddrInUseException {
 }
 
 export function isExecaError(err: any): err is ExecaError {
-  return err.exitCode !== undefined
+  return err.exitCode !== undefined && err.exitCode !== null
 }
 
 export type StackTraceMetadata = {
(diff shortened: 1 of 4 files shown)


stefreak (changes requested):
This is good, but let's also change `ChildProcessError` to not eat the upstream error message if `exitCode` is undefined. Here's the relevant code: https://github.com/garden-io/garden/blob/d980ea8eec8cc21603ec5c0da9210118c38c8065/core/src/exceptions.ts#L355

Maybe it needs to happen one layer above that even, where we decide to throw ChildProcessError and throw InternalError instead if execa gives us an error where exitCode is undefined, with the original error message.

stefreak (dismissed):
👍 Nice, LGTM!

stefreak (approved):

PR #5931, opened 2024-04-15
fix(communityEdition): fix unprocessable entity error

stefreak:
Fix an issue where garden failed with the error "unprocessable entity" some times when user has more than 24 projects.

Previously core listed all projects and if it didn't find project with given name in the first list of results, it would attempt creating the project. Creating the project fails then with "unprocessable entity" because a project with the given name already exists.

Depends on a backend change, thus the draft state


core/src/cloud/api.ts
```diff
@@ -359,41 +359,38 @@ export class CloudApi {
...
       throw new CloudApiDuplicateProjectsError({
         message: deline`Found an unexpected state with multiple projects using the same name, ${projectName}.
         Please make sure there is only one project with the given name.
         Projects can be deleted through the Garden Cloud UI at ${this.domain}`,
       })
     }
 
-    return projects[0]
+    if (projectList.length === 0) {
+      return undefined
```
10ko, line 385: Any reason why we return undefined instead of throwing?

stefreak: That's the behaviour the calling function [`getOrCreateProject`](https://github.com/garden-io/garden/pull/5931/files#diff-c2e1d8caa9ddda54623fafa053401b23d34f039f71077069cec6a6f563f027d0R419) depends on.

10ko (dismissed):

core/src/cloud/api.ts
stefreak: I think it's good to return `undefined` in this case; Using exceptions for control flow would be a bad practice in this case in my opinion.

10ko (approved):

10ko (dismissed):

twelvemo (approved):
Nice!

PR #5952, opened 2024-04-17
ci: use namespace remote builder for dockerhub images

stefreak:
Let's benefit from faster local caching using namespace builders


stefreak:
Unfortunately the test needs to be restructured, otherwise if we load the image after remote building it's much slower than before.

PR #5955, opened 2024-04-18
improvement: allow for more concurrency when using cloud builder

stefreak:
- Allow more fine-grained status and execute task node concurrency control in the
plugins using setters on the action.
- Assign concurrency group by node execution type (status, execute), action type
(exec, container) and the concurrency setting (assuming that actions with a separate
max concurrency setting are a different workload class).
- Fine-tune the build concurrency settings in the container and kubernetes plugins in the
`validate` and `getStatus` handlers for every action.
- This speeds up `garden build -f` in an internal project from ~40 seconds to ~14 seconds.

### DIFF
--- a/core/src/graph/nodes.ts
+++ b/core/src/graph/nodes.ts
@@ -215,7 +215,7 @@ export interface TaskRequestParams<T extends Task = Task> extends TaskNodeParams
 
 @Profile()
 export class RequestTaskNode<TaskType extends Task = Task> extends TaskNode<TaskType> {
-  executionType = "request" as NodeType
+  executionType: NodeType = "request"
 
   override get concurrencyLimit() {
     return gardenEnv.GARDEN_HARD_CONCURRENCY_LIMIT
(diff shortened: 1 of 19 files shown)


core/src/graph/nodes.ts
```diff
@@ -212,8 +215,15 @@ export interface TaskRequestParams<T extends Task = Task> extends TaskNodeParams
 
 @Profile()
 export class RequestTaskNode<TaskType extends Task = Task> extends TaskNode<TaskType> {
-  // FIXME: this is a bit of a TS oddity, but it does work...
-  executionType = <NodeType>"request"
+  executionType = "request" as NodeType
```
vvagaytsev, line 218: I've must figured out a possibly better way of expressing this:
```js
override readonly executionType: NodeType = "request"
```

If we use `as` it is still possible to pass an incorrect string value, while explicit declaration of the property type prevents that. The same can be done in `ProcessTaskNode` and `StatusTaskNode`.

stefreak: Nice, I think that's cleaner, yes

core/src/tasks/deploy.ts
```diff
@@ -32,7 +32,8 @@ function printIngresses(status: DeployStatus, log: ActionLog) {
 @Profile()
 export class DeployTask extends ExecuteActionTask<DeployAction, DeployStatus> {
   type = "deploy" as const
-  override concurrencyLimit = 10
+  override defaultStatusConcurrencyLimit = 10
+  override defaultExecuteConcurrencyLimit = 10
```
vvagaytsev, line 36: It looks like these overrides are not necessary because the same limits are already defined in the `ExecuteActionTask` super class.

vvagaytsev (approved):
Looks great! 💯

PR #5967, opened 2024-04-23
refactor(git): git handler class hierarchy

vvagaytsev:

This PR makes the class hierarchy of the VCS handlers more structured by replacing inheritance with composition.

Previous class hierarchy:

```
VscHandler <--- GitHandler <--- GitRepoHandler
```

New class hierarchy:

```
                                        | <--- GitSubTreeHandler
VscHandler <--- AbstractGitHandler <--- |
                                        | <--- GitRepoHandler (contains GitSubTreeHandler as a delegate)
```

Now, each concrete class implements `getFiles(...)` - the only abstract method defined in `AbstractGitHandler`.

This allows to identify the difference between the various Git repo scanning mode implementations in an easier way.
Also, it avoid calls to a virtual method of the super-class from the private method of the sub-class, and makes the code easier to read and understand.
 
Commit-by-commit review should be easier to do.

Classes `VcsHandler` and `AbstractGitHandler` can be merged to a single one. 
It looks like `VcsHandler` was made to be VCS-agnostic, but, if fact, we support and use only Git.
That it not necessary to merge these 2 classes now. It can be done later in a separate PR.

### DIFF
--- a/core/src/vcs/git-sub-tree.ts
+++ b/core/src/vcs/git-sub-tree.ts
@@ -88,15 +88,6 @@ export class GitSubTreeHandler extends AbstractGitHandler {
    * .ignore files, and the specified include/exclude filters.
    */
   override async getFiles(params: GetFilesParams): Promise<VcsFile[]> {
-    return this._getFiles(params)
-  }
-
-  /**
-   * In order for {@link GitRepoHandler} not to enter infinite recursion when scanning submodules,
-   * we need to name the function that recurses in here differently from {@link getFiles}
-   * so that {@link getFiles} won't refer to the method in the subclass.
-   */
-  async _getFiles(params: GetFilesParams): Promise<VcsFile[]> {
     if (params.include && params.include.length === 0) {
       // No need to proceed, nothing should be included
       return []
@@ -211,7 +202,7 @@ export class GitSubTreeHandler extends AbstractGitHandler {
           }
         }
 
-        return this._getFiles({
+        return this.getFiles({
           log: gitLog,
           path: submodulePath,
           pathDescription: "submodule",


core/src/vcs/git-sub-tree.ts
```diff
@@ -0,0 +1,437 @@
...
+   */
+  override async getFiles(params: GetFilesParams): Promise<VcsFile[]> {
+    return this._getFiles(params)
+  }
+
+  /**
+   * In order for {@link GitRepoHandler} not to enter infinite recursion when scanning submodules,
+   * we need to name the function that recurses in here differently from {@link getFiles}
+   * so that {@link getFiles} won't refer to the method in the subclass.
+   */
```
vvagaytsev, line 98: This might be no longer necessary, as this class has no siblings and we use delegation instead of inheritance in `GitRepoHandler`.

Need to simplify this and validate with the existing test suite.

vvagaytsev: Done in 24a85f85f92970763a31fb99940d035992872380.

stefreak (approved):
Nice work! I think this is a good change, increasing clarity and making it easier to ensure correctness for both scan modes.

core/src/vcs/git-sub-tree.ts
stefreak: Good catch!

stefreak:
The second step of merging `VcsHandler` and `AbstractGitHandler` would indeed make this more clear I think. I wondered why the `AbstractGitHandler` was abstract, but didn't declare any abstract methods until I saw that `VcsHandler` actually declares the abstract methods and `AbstractGitHandler` doesn't implement them all.

vvagaytsev:
@stefreak, sure! Let's call it `AbstractVscHandler` and merge 2 abstract classes, as the double-layer abstraction is redundant here :)

I'll create a PR soon.

PR #5968, opened 2024-04-24
fix(helm): disable `--atomic` by default in Helm modules

### LINKED ISSUE #2713: Garden does not support all k8s resources within Helm charts
### Current Behavior

I am trying install my own Helm charts with k8s jobs only, with garden deploy. Unfortunately it keeps stuck at the deployment phase, although Helm release has been already deployed. Same story is with pods. When I change resource to deployment garden won't stuck since detects status of deployment. I think this is related to serviceResource limitation which supports: "Deployment", "DaemonSet", "StatefulSet" only

### Expected behavior

Garden should support also other k8s resources from Helm charts

### Reproducible example

Here is the snippet of the helm chart files: https://snipit.io/public/snippets/61497
When I try deploy them with `garden deploy` command it's stuck and timeouts after default 300 seconds of Helm release process

### Workaround

Add dummy deployment resource to Helm charts.

### Suggested solution(s)

I think garden should watch the Helm releases status at Helm release level not at the k8s resources like Deployments or StatefuSets

### Additional context

---

### Your environment

* **OS:** macOS BigSur

`garden version` 0.12.33

vvagaytsev:

Aligns the default value of the `--atomic` flag in Helm module- and deploy action configs, and sets the default value to `false`.

Fixes #2713


stefreak (approved):
Nice catch! 👍

PR #5969, opened 2024-04-24
refactor(commands): make `Command.parent` field accessible

vvagaytsev:

Just a minor refactoring a.k.a. code tidying.

### DIFF
--- a/core/src/commands/base.ts
+++ b/core/src/commands/base.ts
@@ -195,7 +195,7 @@ export abstract class Command<
   // E.g: const cmd = new Command(); cmd["parent"] = parentCommand.
   // This is so that commands that are initialised via arguments can be cloned which is required
   // for the websocket server to work properly.
-  private parent?: CommandGroup
+  parent?: CommandGroup
 
   // FIXME: This is a little hack so that we can clone commands that are initialised with
   // arbitrary parameters.
@@ -558,7 +558,7 @@ export abstract class Command<
     // See: https://stackoverflow.com/a/64638986
     const clone = new (this.constructor as new (params?: any) => this)(this._params)
     if (this.parent) {
-      clone["parent"] = this.parent
+      clone.parent = this.parent
     }
     return clone
   }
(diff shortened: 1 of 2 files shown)


core/src/commands/base.ts
```diff
@@ -556,9 +565,9 @@ export abstract class Command<
    */
   clone(): Command {
     // See: https://stackoverflow.com/a/64638986
-    const clone = new (this.constructor as new (params?: any) => this)(this._params)
+    const clone = new (this.constructor as new (params?: any) => this)()
```
stefreak, line 568: How is clone going to work for commands with parameters? Are we never cloning? In this case, we should remove the clone helper.

core/src/commands/base.ts
```diff
@@ -250,6 +251,14 @@ export abstract class Command<
     }
   }
 
+  public setParent(commandGroup: CommandGroup): void {
```
stefreak, line 254: is it maybe easier to make `parent` public, instead of adding a setter?

If it always needs to be set from the outside I wonder what's the point of making it private.

core/src/commands/base.ts
```diff
@@ -250,6 +251,14 @@ export abstract class Command<
...
+    this.parent = commandGroup
+  }
+
+  public hasParent(): boolean {
```
stefreak, line 258: Where do we call `hasParent`?

core/src/commands/base.ts
vvagaytsev: Ah, that's my mistake. I didn't notice we declare the `_params` as private field in the constructor signature. It means that we store it. So I need to revert this change. Thank you! Good catch!

vvagaytsev: I think this part wasn't covered with any tests. We do use cloning.

core/src/commands/base.ts
vvagaytsev: One more good catch! I was suppose to be used internally, but that turned out to be useless, because the `this.parent` field is optional, and the `hasParent` method can't be a undefined-guard checker for it.

I'm wondering if we need private field and setter at all. We can just keep the `parent` accessible from outside.

core/src/commands/base.ts
vvagaytsev: Right, I already did that :)

eysi09 (approved):

PR #5975, opened 2024-04-25
fix(cli): do not render hidden commands

eysi09:

Before this fix, hidden commands (and subcommands) were visible when running the help command and in our docs.

This fixes that.

Fixes #


vvagaytsev (approved):
LGTM! 👍 Thank you!

PR #5996, opened 2024-05-04
feat: optional varfiles

### LINKED ISSUE #5994: [FEATURE]: Enhanced Environment Variable Management with Optional Varfiles
### Background / Motivation

While working on my project, I found it cumbersome to manage optional environment variables, especially when dealing with local development configurations. I often need to include certain environment variables from local and gitignored `.env` files, which should only be used if they exist. This makes it challenging for new developers to set up the project and override defaults when necessary.

Users should be able to include optional varfiles, such as local and gitignored `.env` files, which would be utilized only if they exist. This feature would streamline the process of starting the project for new developers and allow for easy overriding of default configurations when needed.

This feature eliminates the inconvenience of managing optional environment variables and simplifies the setup process for new developers. It also provides flexibility to override default configurations when necessary, enhancing the overall development experience.

Implement a feature that allows users to specify optional varfiles, such as local and gitignored `.env` files, in the project configuration. These varfiles should be included dynamically during project initialization if they exist, providing the necessary environment variables to the application.

Maybe same syntax like [optional variables](https://docs.garden.io/using-garden/variables-and-templating#optional-values)
```yaml
varfiles:
  - my-service.${environment.name}.yaml? # question mark to define varfiles as optional?
  # or
  - path: my-service.${environment.name}.yaml
    optional: true
```

ManAnRuck:

This will add the feature to add optional varfiles

Fixes #5994
[Discord Thread](https://discord.com/channels/817392104711651328/1235161947998130237)

my first PR for this awesome project :) 

- [x] add tests
- [x] adjust documentation
- [x] add Varfile object to validation


core/src/config/common.ts
```diff
@@ -845,6 +852,18 @@ export const joiIdentifierMap = memoize((valueSchema: Joi.Schema) =>
     .description("Key/value map. Keys must be valid identifiers.")
 )
 
+export const joiVarfile = memoize(() =>
+  joi
+    .alternatives(
+      joi.string().description("Path to a file containing a path."),
```
vvagaytsev, line 858: Thanks for the PR! 👍 

It looks like this string should also be declared as a `posixPath()` instead of a plain `string()`.

ManAnRuck: Oh yes! Thank you. I will change it as soon as possible 😊

twelvemo:
Thanks for the PR! Looks good at first glance. In general i think we should avoid introducing alternative configurations for the same thing, but this would otherwise be a breaking change. Since the new varfiles object only adds to the old one, i would suggest deprecating the old notation and remove it in the next minor release. What do others think?

ManAnRuck:
> Thanks for the PR! Looks good at first glance. In general i think we should avoid introducing alternative configurations for the same thing, but this would otherwise be a breaking change. Since the new varfiles notation only adds to the old one, i would suggest deprecating the old notation and removing it in the next minor release. What do others think?

@twelvemo  Should i change something in this pr?

PR #5997, opened 2024-05-05
[WIP] feat: add date template helper functions

### LINKED ISSUE #5995: [FEATURE]: Enhanced Date Manipulation in Templating Files for Simplified Environment Configuration
### Background / Motivation

I am currently working on setting up a local development environment where a DATE variable is crucial for configuring and initializing the system. I've been using an environment variable to manage this date, but this method introduces unnecessary complexity into my application. I recently discovered functions similar to what Make.com offers for handling date operations, and I believe adopting a similar approach in our templating files would streamline our setup process.

Users should be able to define dynamic date calculations directly in templating files without needing to write additional logic in their application. Specifically, they should be able to use functions to manipulate dates, such as subtracting days from the current date. This feature would allow for easier and more flexible configuration management.

Implementing date helper functions within templating files will eliminate the need for external scripts or application logic to handle date calculations. This reduces the complexity of the codebase, minimizes potential errors, and improves the scalability and maintainability of the environment setup. It specifically addresses the pain point of manually managing date-related configurations which are subject to frequent changes.

A new set of date helper functions can be introduced, similar to those found in Make.com's templating. For example, functions like `addDays`, `subtractDays`, and `now()` can be incorporated. The usage could be as simple as specifying in a configuration file:
```yaml
- name: my-config-map
  kind: ConfigMap
  patch:
    data:
      DATE_START: ${addDays(now(), -2)}
```
This implementation allows for dynamic and automated date management within configuration files.

This feature will significantly streamline the setup process for our development environments, reducing the need for manual adjustments and decreasing the risk of errors associated with date management.

ManAnRuck:

Fixes #5995


eysi09:
Thanks so much for the PR @ManAnRuck! ❤️

I think it definitely makes sense to have a helper function for printing the current timestamp with options for different formats. 

Adding functionality for manipulating dates feels a little out of scope though tbh and I can see this grow in until we've re-implemented all of date-fs as template functions.

What exactly is your use case? Wondering if there's other ways to achieve that.

ManAnRuck:
> Thanks so much for the PR @ManAnRuck! ❤️
> 
> I think it definitely makes sense to have a helper function for printing the current timestamp with options for different formats.
> 
> Adding functionality for manipulating dates feels a little out of scope though tbh and I can see this grow in until we've re-implemented all of date-fs as template functions.
> 
> What exactly is your use case? Wondering if there's other ways to achieve that.

Hi @eysi09 Thank you for your response :)

That would of course be ideal if we could find a solution without having to introduce additional complexity into the code :)

Specifically, the use case is that an environment variable exists for scraper services and api services, which defines the date from which the scraper or the data should be retrieved.
It is of course possible to enable this in the code itself using another environment variable, but this would be logic that is not part of the business logic of the actual service and I would therefore like to avoid incorporating it into the service.
For example, at the time the service/job is started, all data that is newer than 7 days for dev should be pulled. for production, the environment variable is set based on the date of the last update.

Translated with DeepL.com (free version)

ManAnRuck:
Another alternative could be to make it possible to add your own scripts, which would be very good in the sense that it could work like a plugin system, but I imagine the implementation to be a bit more complex.

For the user it could then look like this: there is a standard file path or/and this can be defined in garden.yaml files.
This path then points to a file which exports an object of the type `TemplateHelperFunction`.
The content of the file could then be spread into the existing `helperFunctionSpecs` variable.

The biggest hurdles from my point of view would be the use of third party libraries for own functions (e.g. `date-fns-tz`) and the dynamic loading of the object without having to rebuild garden. I don't have an overview of this at the moment.

eysi09:
I discussed this a bit better with the team. 

We agree that adding all these helpers functions is a lot and it kind of pollutes the function namespace because we don't have the functionality to tie to some context (e.g. `datetime.parseDate`).  So in fact the majority of template functions would be simple date helpers. 

Also note that we already have the [`datetime` context](https://docs.garden.io/reference/template-strings/projects#usd-datetime).

Would it work for your use case if instead we just:

- Keep the `formatDate` function
- Add a catch all `modifyDate` function that can basically modify any date string

That `modifyDate` function could have the following signature:

```
modifyDate(timestamp: string, unit: "ms" | "s" | "d" | "m" | "y", value: number, format?: string)
```

The `format` character would be optional and only needed if the timestamp is not and ISO timestamp. The timestamp could then be passed through the `date-fns` `parse` function with the format if required.

It could be called like this in the YAML config:

```
yesterday: ${modifyDate(datetime.now, "d", -1)
```

The `formatDate` function could then consume the output of `modifyDate` if needed.

ManAnRuck:
Thank you for the discussion and suggestions regarding the date handling functions. ❤️

I've taken the liberty to flesh out the modifyDate function as per the discussions, ensuring it accommodates both "add" and "set" modes. This allows for greater flexibility, whether adjusting or explicitly setting date units. To avoid ambiguity, especially in units that could be easily conflated like months ('m') and minutes ('minutes'), I opted to use fully spelled-out units. This should help prevent any potential confusion or errors in unit interpretation.

Additionally, I added a `mode` parameter that specifies whether to `add` to or set` the date unit, enhancing the function's adaptability for different use cases.

Regarding testing, I'm encountering an issue related to timezone handling. Currently, the function operates as expected only when the environment variable `TZ='UTC'` is set. Without this setting, local tests do not behave as intended. This should be fixed before merging 🧐

```yaml
yesterday: ${modifyDate(datetime.now, "days", -1)}
first-day: ${modifyDate(datetime.now, "days", 1, "set")}
```

I would be happy if you could rate these adjustments again and share your thoughts :)

_(the code is already customized)_

core/src/template-string/date-functions.ts
```diff
@@ -0,0 +1,89 @@
...
+      { input: ["2021-01-01T00:00:00Z", -1, "days", "add"], output: "2020-12-31T00:00:00.000Z" },
+      { input: ["2021-01-01T00:00:00Z", 30, "seconds", "set"], output: "2021-01-01T00:00:30.000Z" },
+    ],
+    fn: (date: string, amount: number, unit: string, mode: string) => {
+      const dateClone = new Date(date)
+      switch (mode) {
+        case "add":
+          return add(dateClone, { [unit]: amount }).toISOString()
+        case "set":
+          switch (unit) {
```
vvagaytsev, line 61: This switch block does not handle the time unit of type `"weeks"`. I made `unit` argument type-safe in ef87fee586307abfd78b6be594b36e67553a263e and removed the `"weeks"` time unit in 58c4c8ba612547dfd2d5a54f252aae721a7c27b7.

Feel free to re-implement `weeks` unit if necessary. I haven't found an existing library method to sett a week on a date object.

vvagaytsev: I'll do more changes about the time unit types. Those might need to be connected to the mode types.

core/src/template-string/date-functions.ts
```diff
@@ -0,0 +1,92 @@
...
+      date: joi.string().required().description("The date to format."),
+      format: joi
+        .string()
+        .required()
+        .description("The format to use. See https://date-fns.org/v2.21.1/docs/format for details."),
+    },
+    outputSchema: joi.string(),
+    exampleArguments: [
+      { input: ["2021-01-01T00:00:00Z", "yyyy-MM-dd"], output: "2021-01-01", skipTest: true },
+      { input: ["2021-01-01T00:00:00Z", "yyyy-MM-dd HH:mm:ss"], output: "2021-01-01 00:00:00", skipTest: true },
```
vvagaytsev, line 30: What was the reason for skipping these tests? Did the test fail?

ManAnRuck: I had an issue with local time, since I replaced the input from date to string it was solved. Means there is no reason to skip the test anymore 😃

vvagaytsev: Ok, thank you! :)

core/src/template-string/date-functions.ts
ManAnRuck: From my point of view we can leave weeks out :) thank you!

vvagaytsev:
@eysi09 @stefreak I've made some changes and added more test cases. Please review, see individual commits for details.

core/src/template-string/date-functions.ts
```diff
@@ -0,0 +1,118 @@
...
+        .valid(...validModifyDateTimeUnits)
+        .required()
+        .description("The time unit to set."),
+    },
+    outputSchema: joi.string(),
+    exampleArguments: [
+      { input: ["2021-01-01T00:00:00.234Z", 345, "milliseconds"], output: "2021-01-01T00:00:00.345Z" },
+      { input: ["2021-01-01T00:00:05Z", 30, "seconds"], output: "2021-01-01T00:00:30.000Z" },
+      { input: ["2021-01-01T00:01:00Z", 15, "minutes"], output: "2021-01-01T00:15:00.000Z" },
+      { input: ["2021-01-01T12:00:00Z", 11, "hours"], output: "2021-01-01T11:00:00.000Z", skipTest: true },
```
stefreak, line 106: These tests likely fail becaue we instantiate a date with local timezone from the UTC string, and then we modify the local time zone and export as UTC again.

In the first iteration of this, my proposal would be to rename `shiftDate`, `modifyDate` and `formatDate` to `shiftDateUTC`, `modifyDateUTC` and `formatDateUTC` and to only operate UTC dates; if someone needs timezone support later we can consider that by adding `formatDateLocal` or similar functions.

stefreak: Do we not also need `nowUTC()` or something like that?

vvagaytsev: > These tests likely fail becaue we instantiate a date with local timezone from the UTC string, and then we modify the local time zone and export as UTC again.
> 
> In the first iteration of this, my proposal would be to rename `shiftDate`, `modifyDate` and `formatDate` to `shiftDateUTC`, `modifyDateUTC` and `formatDateUTC` and to only operate UTC dates; if someone needs timezone support later we can consider that by adding `formatDateLocal` or similar functions.

Did that for date modifiers in 890adc013ec7032f7caf361e0adc69755c9f8e0a. The `formatDate` will become `formatDateUtc` in the next commit :)

> Do we not also need `nowUTC()` or something like that?

We can do that, it might be useful.

vvagaytsev: And the formatter was updated in 96b2300ed98d62dc6ee64edbb31f0435e45c2a01.

vvagaytsev: I'm not sure if we need `nowUtc()` now. @ManAnRuck wdyt about it?

vvagaytsev:
@stefreak @ManAnRuck please review the current version of the PR. Comments, feedback and improvement ideas are highly appreciated :)

cc @eysi09

core/src/template-string/date-functions.ts
ManAnRuck: I think if it helps we can leave it out for now and from the user's point of view a workaround with a manual adjustment according to the time zone may have to be made first. Then we can first collect what problems may arise and improve this in a further integration.

stefreak (approved):
Thanks for the contribution @ManAnRuck, I think this can be helpful for some folks 👍 
Thanks for revisiting @vvagaytsev

core/src/template-string/date-functions.ts
vvagaytsev: Ok, thank you! Let's collect some feedback and do the necessary improvement in the further PR(s).
We're merging this PR today :)

PR #5998, opened 2024-05-06
refactor(vcs): keep class fields private and readonly if possible

vvagaytsev:
One more tidying and janitorial PR for VCS handling classes. Add-on for #5967.


twelvemo (approved):
Looks good!

PR #5999, opened 2024-05-06
test: verify the hashes of the bundled NodeJS binaries

vvagaytsev:

This will make the next NodeJS version updates a bit smoother and less error-prone.

Fixes #


twelvemo (approved):
Nice work!

PR #6002, opened 2024-05-06
perf(cli): avoid unnecessary module resolution when filtering by name

edvald:
This applies to most common usages of the `build`, `deploy`, `run` and `test` commands when one or more names are specified, as well as in the `get modules` and `get actions` commands.

For now this is enabled specifically with by setting the `GARDEN_ENABLE_PARTIAL_RESOLUTION=true` env variable.


core/src/commands/test.ts
```diff
@@ -149,13 +149,26 @@ export class TestCommand extends Command<Args, Opts> {
...
+
+    // TODO: Optimize partial resolution further when --skip-dependencies=true
+    // TODO: Optimize partial resolution further with --skip flag
+    if (args.names) {
+      actionsFilter = args.names.map((name) => `test.${name}`)
+    }
+
+    if (opts.module) {
+      actionsFilter = [...(actionsFilter || []), `test.${opts.module}-*`]
+    }
```
vvagaytsev, line 162: It looks like `garden run` command also support `--module` option. Should we implement the same logic for the Run actions?

core/src/resolve-module.ts
```diff
@@ -532,7 +757,7 @@ export class ModuleResolver {
     dependencies: GardenModule[]
     repoRoot: string
   }) {
-    this.log.silly(() => `Resolving module ${resolvedConfig.name}`)
+    this.log.debug(() => `Resolving module ${resolvedConfig.name}`)
```
vvagaytsev, line 760: Nit. This message can also be prefixed with `ModuleResolver: ` like the completion message below.

core/src/resolve-module.ts
```diff
@@ -149,7 +136,7 @@ export class ModuleResolver {
       let foundNewDependency = false
 
       const dependencyNames = fullGraph.dependenciesOf(moduleKey)
-      const resolvedDependencies = dependencyNames.map((n) => resolvedModules[n])
+      const resolvedDependencies = dependencyNames.map((n) => resolvedModules[n]).filter(Boolean)
```
vvagaytsev, line 139: Is this filter equivalent to `(n) => !!n`? We have `isTruthy` helper that might make this a bit more clear.

core/src/commands/test.ts
edvald: Ah, didn't notice that. I'd say not strictly necessary for that one, feels more relevant for tests (since it's more natural to run all tests for a module than to run all tasks).

core/src/resolve-module.ts
edvald: This is equivalent, yes. The isTruthy helper is sometimes better for type-safety but that's not a concern in this case.

core/src/resolve-module.ts
```diff
@@ -247,18 +241,249 @@ export class ModuleResolver {
...
+    }
+
+    // Need to make sure we resolve modules that contain runtime dependencies of services, tasks and tests specified
+    // in actionsFilter (if any), including transitive dependencies.
+    let mayNeedAdditionalResolution = false
+
+    for (const f of actionsFilter || []) {
+      // Build dependencies, i.e. module-to-module deps, will already be accounted for above.
+      if (!f.startsWith("build.")) {
+        mayNeedAdditionalResolution = true
```
vvagaytsev, line 262: We can put a `break` statement after this to avoid iterating the tail of the actions, or we can use `Array.some()` that return immediately after the first match:
```
(actionsFilter || []).some((f) => !f.startsWith("build."))
```

Then we can convert `mayNeedAdditionalResolution` to a constant:
```
let mayNeedAdditionalResolution = (actionsFilter || []).some((f) => !f.startsWith("build."))
```

edvald: Ah yes. Initially I had a more complex check in there, but that would be a bit better.

vvagaytsev (approved):
LGTM! Thank you! 👍 

The `ModuleResolver.resolve()` seems go be a bit long and complicated, but let's avoid unnecessary refactoring for now. The tests pass, so we can go ahead :)

edvald:
> The `ModuleResolver.resolve()` seems go be a bit long and complicated, but let's avoid unnecessary refactoring for now. The tests pass, so we can go ahead :)

Fully agree on that one. Since modules are deprecated though, I didn't want to spend too much time on that. If we put more work into it we can clean it up.

core/src/resolve-module.ts
vvagaytsev: Changed this in 269f2a1ea40493743e483ec2df7b6f9be1fe7138.

core/src/resolve-module.ts
vvagaytsev: Simplified this in 78cf1026b52e0314a019a9c517294f809559116f.

vvagaytsev:
@edvald I reverted one behavioural change in 269b79a6e7d3dd7f3e7ad33be009293687f62120 (see the commit message 
for details). Let's ship it as a separate feature PR.

Please take a look. If you're fine with that, let's merge this.

I'm also thinking about some tiny refactoring to unify the build/run/test/deploy action lookups by names/wildcards and the relevant error handling.

vvagaytsev:
Did more precise error handling in 4b86e055e971326c6957a5358bb00c1e068bf149. Let's see if all the tests pass.

vvagaytsev:
Changes from 4b86e055e971326c6957a5358bb00c1e068bf149 caused even more test failures, I'll revert that commit.

vvagaytsev:
Finally, did some explicit check in 1e64797e3cdf4012fd3c4788ede72b2eb715d26c to ensure the old behaviour when `GARDEN_ENABLE_PARTIAL_RESOLUTION=false`.

vvagaytsev (approved):

PR #6012, opened 2024-05-07
fix(helm): fix race condition when updating deps

### LINKED ISSUE #5979: [FEATURE]: Add flag to disable chart dependency updates for Helm Deploy actions
### Background / Motivation

Our organization uses a Helm chart to deploy and configure our services in production environments, and currently, we are in the process of converting Garden to use that chart for development environments as well. What we're aiming to do is to set up Garden modules to deploy individual services, but have run into a problem with using the same chart across multiple modules.

Each Helm Deploy action will always run a `helm dependency update` when deploying, which causes a conflict when parallel modules compete over the `tmpcharts` directory. That directory is used by Helm to download the dependent charts before copying them over to the proper `charts` directory. This conflict over the `tmpcharts` directory can cause deploys to fail if one module cleans up the directory while another is running.

```
Error: open /Users/username/org/repo/charts/mychart/tmpcharts: no such file or directory
```

Some workarounds are: 
- Add a dependency chain in Garden to deploy each module one at a time
- Separate each service into an individual chart

But neither of those options will scale with a larger number of services.

Since the dependencies can be updated manually or already be present in the chart, it is not necessary to always run the `helm dependency update` and therefore should be an option in the Garden Helm Deploy action. This could also default to true to not change the behavior for existing modules.

Specify a boolean value in the Helm Deploy action to disable dependency update.

Allow parallel Helm Deploy actions to deploy using the same chart without conflicts.

Add a value for disabling update of dependencies:

```yaml
kind: Deploy
type: helm
name: my-helm-module
spec:
  updateDependencies: false
```

When false, the dependency update that occurs [here](https://github.com/garden-io/garden/blob/bfcc1fea81d01f99e25fb875332ac8712481e5e3/core/src/plugins/kubernetes/helm/common.ts#L129) would be skipped, and also the `--dependency-update` argument would be omitted from the render template function [here](https://github.com/garden-io/garden/blob/bfcc1fea81d01f99e25fb875332ac8712481e5e3/core/src/plugins/kubernetes/helm/common.ts#L351).

thsig:

When more than one Helm Deploys that referenced the same dependency chart were executed simultaneously, this would trigger a race condition (since the concurrent calls to `helm dependency update` would result in one call cleaning up the temp directory while another call was using it).

This was fixed by adding an async lock (keyed by chart path) and a simple in-memory TTL (making sure we don't re-fetch the same dependency chart if it's already been fetched in the last 300 seconds).

Fixes #5979.

### DIFF
--- a/core/src/plugins/kubernetes/helm/common.ts
+++ b/core/src/plugins/kubernetes/helm/common.ts
@@ -7,6 +7,7 @@
  */
 
 import { flatten, isPlainObject } from "lodash-es"
+import AsyncLock from "async-lock"
 import { join, resolve } from "path"
 import fsExtra from "fs-extra"
 const { pathExists, readFile, remove, writeFile } = fsExtra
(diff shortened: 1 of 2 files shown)


vvagaytsev (approved):
Nice! 👍

vvagaytsev (changes requested):
@thsig there are some test failures. The expected log line no longer exists. Most likely, because we use a cached chart and do run the dependency update command that produces the expected logs.

twelvemo:
Looks good, we just need to adjust [this test](https://github.com/garden-io/garden/blob/1c5c3a09a26a90028d234b96030690a865b4b14c/core/test/integ/src/plugins/kubernetes/helm/common.ts#L681).

thsig:
Ok cool, I'll have a look

thsig:
Should be good to go now.

I fixed this by adding a debug-level log line that's logged when dependencies have already been fetched and looking for that in the failing test cases. Probably a good idea to have that debug log line anyway (in case users need to diagnose something around this flow in the future).

twelvemo (approved):
Nice!

vvagaytsev:
Thanks @thsig 💪

PR #6027, opened 2024-05-13
chore(deps): update Python deps in tests and examples

vvagaytsev:

In this PR:

* Update Python Docker images to be the latest Python-3.12 Alpine ones in tests and examples
* Update `Flask` versions across tests and examples
* Align `requests` and `urllib3` versions across tests and examples

The `psycopg2` dependency should also have an explicit version. Some image build errors occurred while trying to use the latest available version. That library can get an explicit version in a separate PR, it not urgent to do it now.


stefreak (approved):
👍

PR #6028, opened 2024-05-13
ci: use kubernetes 1.30 in microk8s tests 

vvagaytsev:

Update `microk8s` CI tests to use the latest Kubernetes version (`1.30`).

Kubernetes `1.26` is no longer supported, and `1.30` was released recently.
See https://kubernetes.io/releases/

Fixes #

The default `minikube` version was also updated to be `v1.32.0` (which is the currently used one).


twelvemo (approved):
Thanks for taking care of this!

PR #6037, opened 2024-05-14
fix(terraform): prevent deadlock by consuming stdout

stefreak:
Nodejs spawn will cease to emit error or close events when stdout hasn't been consumed.

See also from official docs:

> By default, pipes for stdin, stdout, and stderr are established between the parent Node.js process and the spawned subprocess. These pipes have limited (and platform-specific) capacity. If the subprocess writes to stdout in excess of that limit without the output being captured, the subprocess blocks waiting for the pipe buffer to accept more data. This is identical to the behavior of pipes in the shell. Use the { stdio: 'ignore' } option if the output will not be consumed.

Source: https://nodejs.org/api/child_process.html#optionsstdio

We used to consume the stdout, but in the ESM PR (https://github.com/garden-io/garden/commit/0f535ddfc156494236028eb292860c458cc37b7c) we accidentally removed the pipe to logStream which can cause deadlocks for terraform users.

Fixes #

### DIFF
--- a/plugins/terraform/src/helpers.ts
+++ b/plugins/terraform/src/helpers.ts
@@ -205,6 +205,10 @@ export async function applyStack(params: TerraformParamsWithVariables) {
     })
   }
 
+  if (proc.stdout) {
+    proc.stdout.pipe(logStream)
+  }
+
   logStream.on("data", (line: Buffer) => {
     statusLine.info(styles.primary("→ " + line.toString()))
   })


vvagaytsev (approved):
Wow! Amazing job on investigating this and drilling down to such low-level details! 🚀 
Thank you for fixing this! 👍 💯

stefreak:
@vvagaytsev thanks, hoping this is the sole culprit for the deadlocks we were seeing :)

stefreak:
regression test implemented in https://github.com/garden-io/garden/pull/6040

PR #6038, opened 2024-05-14
chore(deps): update ink to 5.0.0

vvagaytsev:

* remove redundant `@types/wrap-ansi`
* update `ink` to `5.0.0` to get rid of deprecated transitive dependency on `@types/wrap-ansi`


stefreak (approved):
👍

PR #6040, opened 2024-05-14
test(terraform): add assertion for terraform stdout

stefreak:
Follow-up on #6037 (This adds a missing test)

### DIFF
--- a/core/test/integ/src/plugins/kubernetes/kubernetes-type/common.ts
+++ b/core/test/integ/src/plugins/kubernetes/kubernetes-type/common.ts
@@ -624,7 +624,7 @@ describe("getManifests", () => {
       expect(manifests[0].spec.replicas).to.eql(3)
     })
     it("should log a warning if patches don't match manifests", async () => {
-      garden.log.root["entries"] = []
+      garden.log.root["entries"].length = 0
       const action = cloneDeep(graph.getDeploy("deploy-action"))
       action["_config"]["spec"]["patchResources"] = [
         {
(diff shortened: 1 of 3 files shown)


twelvemo (approved):
👍

core/test/integ/src/plugins/kubernetes/kubernetes-type/common.ts
```diff
@@ -624,7 +624,7 @@ describe("getManifests", () => {
       expect(manifests[0].spec.replicas).to.eql(3)
     })
     it("should log a warning if patches don't match manifests", async () => {
-      garden.log.root["entries"] = []
+      garden.log.root["entries"].length = 0
```
vvagaytsev, line 627: Is this a more preferred way of emptying arrays? I'm just curious :)

stefreak: I did this as changing the identity of the private array made the test fail in the terraform tests for whatever reason. Setting length to 0 removes all elements of the array but preserves it's identity, and to prevent others from experiencing the same thing I changed all occurrences of this "reset" code to the more reliable method of preserving the identity.

stefreak: It think it'd be less hacky if we added a "reset" method to the logger interface or implemented a helper somewhere 🤔

vvagaytsev: Thanks for the explanation! I've just looked to the [official JS reference docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/length) and it looks like it's quite a common way of emptying an array :)
So, no need to create a helper)

vvagaytsev: If we need to mutate an array without reassigning it and changing its identity, that's the way to go :)

The official docs say that the exceed elements will be removed.
That can be understood as those won't be kept by any internal reference and will become available for garbage collection :)

vvagaytsev (approved):
👍

PR #6044, opened 2024-05-14
chore: update nodejs to v22.1.0

stefreak:
- chore: update nodejs to v22.1.0
- chore: remove UV_USE_IO_URING=0 fix as libuv 1.48 landed
- chore: update nodejs in circleci
- chore(images): update circleci-runner to use NodeJS 22.1.0

PR #6045, opened 2024-05-15
chore: remove pty ws endpoint

stefreak:
Maintaining PTY is a pain, especially with node upgrades. See #6044 as
an example. It's blocked by [support in node-pty-prebuilt-multiarch for Node ABI 127](https://github.com/homebridge/node-pty-prebuilt-multiarch/issues/42). Also it seems to [seems to be behind the official pty code from microsoft](https://github.com/homebridge/node-pty-prebuilt-multiarch/pull/41#issuecomment-1998611609).
For Desktop we can implement terminal support in Tauri instead of
Garden.

Fixes #


vvagaytsev (approved):
Great stuff! 🎉 
I'm happy see we're dropping one more heavy-to-maintain dependency :)

PR #5400, opened 2023-11-13
chore: update analytics library

stefreak:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

PR #6047, opened 2024-05-15
chore: remove unneeded ci step to save 50 seconds

stefreak:
We use rollup wasm build since
https://github.com/garden-io/garden/pull/5400 so no need to install
linux build of rollup in CI anymore.


vvagaytsev (approved):
Cool! 👍

PR #6049, opened 2024-05-15
improvement: improve cli startup performance using v8 cache

stefreak:
Use the V8 cache feature introduced in Node v22.1.0 to improve the CLI command startup time.

This feature is enabled by default, and `GARDEN_COMPILE_CACHE=0` can be used if it causes a problem as escape hatch.

On my M2 mac it saves ~11-30ms when running `garden --help`; On CircleCI Windows runner it saves 2 seconds for each CLI command invocation, so the difference can be really significant.


vvagaytsev (approved):
LGTM! 👍

vvagaytsev (approved):

Issue #3327, opened 2022-10-31
Validate command creates k8s namespace


hnicke:
### Current Behavior
When running `garden validate`, garden creates the target k8s namespace (if it doesn't exist).
`garden validate` fails if the configured kube-context is not available.

### Expected behavior
* `garden valdiate` should not have any sideeffects or they should be documented
* `garden validate` should not require write permission for the target cluster
* particurlarily, it should not create a namespace
* I'd prefer if `garden validate` wouldn't fail if the configured context is invalid

`garden version`
0.12.46

hnicke:
I've stumbled across a similar issue today: `garden update-remote` will fail if it can't connect to the cluster. I expect that the `update-remote` command does not try to connect to the k8s cluster at all. Or is there a reason..?

Orzelius:
@hnicke thanks for pointing that out! We'll work on fixing both as it's weird behavior and makes these commands way slower than they should be

hnicke:
still relevant

alexkuretz:
This also occurs when using the `garden logs` command, if the namespace doesn't exist (usually occurs when a dev typo's the namespace name while trying to view logs in a deployment) garden creates an empty namespace. 

`BUILDVERSION=garden-test garden logs --env main `

project.garden.yml snippet:
```
- name: main  
    defaultNamespace: ${local.env.BUILDVERSION}  
    providers:  
      - name: kubernetes  
        context: arn:aws:eks:us-west-2:${var.aws-account}:cluster/${var.cluster-name}
```

Per Albert from Garden:
> Currently this is expected behaviour as you're passing the namespace via defaultNamespace: ${local.env.BUILDVERSION}
Garden creating the namespace is a side effect of https://github.com/garden-io/garden/issues/3327
It's currently not an easy fix but one day we'll get do it. 
I'd recommend to create an exec initscript that validates the BUILDVERSION env var and throws if it's not correct. Then you could make the kubernetes provider depend on the exec provider which would validate the namespace before kubernetes provider attempts to create it.

PR #6051, opened 2024-05-15
 fix(provider): allow initialising providers without write ops for validation command

### LINKED ISSUE #6029: Running `garden validate` applies terraform `rootStack` when `autoApply` is set to true
### Current Behavior

When running `garden validate` with a terraform `rootStack` that has `autoApply` set to true, the stack gets applied. 

### Expected behavior

A validate command should not apply changes to infrastructure even when `autoApply` is set.

### Reproducible example

https://github.com/garden-io/garden/tree/main/examples/terraform-gke

### Your environment

* **OS:**  macOS

`garden version`
0.13.30

twelvemo:

While resolving providers all kinds of actions are performed in the different providers e.g. for the kubernetes providers ingress controllers are installed and app namespaces created. The terraform provider with an `initRoot` stack and `autoApply` set to `true` will even apply a terraform stack on provider initialization. This is the intended behavior for a lot of commands like `deploy` etc. However some commands like `validate` also rely on resolving the providers, whereas a `validate` command should not apply any changes.

This PR adds a `statusOnly` param to the graph and provider resolution so providers can be resolved and their status queried without any write operations. To do this the `getEnvironmentStatus` handler on each provider should only implement status checks, while the `prepareEnvironment` handler should handle any write calls. Added tests for all providers that implement these handlers.

Fixes #6029 #3327


core/src/tasks/resolve-provider.ts
```diff
@@ -379,7 +380,7 @@ export class ResolveProviderTask extends BaseTask<Provider> {
     await writeFile(cachePath, serialize(cachedStatus))
   }
 
-  private async ensurePrepared(tmpProvider: Provider) {
+  private async ensurePrepared(tmpProvider: Provider, statusOnly?: boolean) {
```
vvagaytsev, line 383: Argument `statusOnly` can be declared as non-optional here. This method is called from only one place, and the actual non-optional boolean value is supplied.

core/src/graph/solver.ts
```diff
@@ -418,7 +418,7 @@ export class GraphSolver extends TypedEventEmitter<SolverEvents> {
           // Status is either aborted or failed
           this.log.silly(() => `Request ${request.getKey()} status: ${resultToString(status)}`)
           this.completeTask({ ...status, node: request })
-        } else if (request.statusOnly && status !== undefined) {
+        } else if (request.statusOnly && status !== undefined && status.result) {
```
vvagaytsev, line 421: Is the condition `status.result` necessary here? The `GraphResult.result` field is defined as nullable. I'm not sure about the semantics behind the `null`-value here.

core/src/garden.ts
```diff
@@ -742,7 +742,7 @@ export class Garden {
   @OtelTraced({
     name: "resolveProviders",
   })
-  async resolveProviders(log: Log, forceInit = false, names?: string[]): Promise<ProviderMap> {
+  async resolveProviders(log: Log, forceInit = false, statusOnly = false, names?: string[]): Promise<ProviderMap> {
```
vvagaytsev, line 745: We can now convert the args list to a param object. Maybe we should do the same for `resolveProvider(...)` method.

core/test/integ/src/plugins/kubernetes/provider.ts
```diff
@@ -0,0 +1,85 @@
...
+import { getEnvironmentStatus, prepareEnvironment } from "../../../../../src/plugins/kubernetes/init.js"
+import type { PrepareEnvironmentParams } from "../../../../../src/plugin/handlers/Provider/prepareEnvironment.js"
+import type { Garden } from "../../../../../src/index.js"
+
+async function ensureNamespaceDoesNotExist(api: KubeApi, namespaceName: string) {
+  try {
+    await api.core.readNamespace({ name: namespaceName })
+  } catch (e) {
+    return
+  }
```
vvagaytsev, line 25: Could this be simplified with `namespaceExists` helper?

Something like
```js
if (!(await namespaceExists(api, namespaceName))) {
    return
}
```

vvagaytsev (commented):
Looks good, thank you! 👍 

I've left [one question](https://github.com/garden-io/garden/pull/6051/files#r1607826922) and a few non-blocking minor comments. Let's check those and get this merged today.

Huge thanks for adding provider-level tests! 💯

core/src/graph/solver.ts
twelvemo: `statusOnly` wasn't really used before. Now when it is actually `true` and passed in to the solver, it will return a `status` in the first iteration, but `status.result` will still be `null`. However `status.result` is needed in `completeTask` see [here](https://github.com/garden-io/garden/blob/main/core/src/graph/solver.ts#L469).

twelvemo: It is a bit confusing, but we actually try to read `result.result` [here](https://github.com/garden-io/garden/blob/main/core/src/graph/solver.ts#L469) which is still null after `complete`.
Without this check, garden crashes:
```
TypeError: Cannot read properties of null (reading 'status')
    at file:///Users/anna/repos/garden/core/build/src/garden.js:575:56
    at Array.every (<anonymous>)
    at file:///Users/anna/repos/garden/core/build/src/garden.js:575:41
```
Do you think there is a better way to deal with this?

vvagaytsev: Ah, I see. Thanks for the clarification! 👍 
I don't see a better way at the moment, just wanted to double check this and get more context :)
Looks good!

vvagaytsev (approved):
Thank you! 💯

PR #6063, opened 2024-05-21
fix(commands): use statusOnly provider resolution for several read-only commands

### LINKED ISSUE #3327: Validate command creates k8s namespace

twelvemo:
Applies the `statusOnly` option for provider resolution for all kinds of read-only commands:
* get-actions
* get-modules
* get-files
* get-graph
* update-remote
* profile-project
* logs

Fixes #3327

Based on the changes made in #6051.

### DIFF
--- a/core/src/commands/get/get-actions.ts
+++ b/core/src/commands/get/get-actions.ts
@@ -152,7 +152,7 @@ export class GetActionsCommand extends Command {
       actionsFilter = [opts.kind + ".*"]
     }
 
-    const graph = await garden.getResolvedConfigGraph({ log, emit: false, actionsFilter })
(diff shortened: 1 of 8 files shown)


vvagaytsev (approved):
💯

Issue #6048, opened 2024-05-15
0.13: [Bug]: Creating Extra Cloud Secrets from Files fails


vvagaytsev:
## Garden Bonsai (0.13) Bug

Bulk secrets creation from files fails on duplicate secrets. 

Assume there is an `.env` file with some new secrets added. Attempt to create the secrets from that file fails if at least one of the secrets defined in the file already exists.

### Current Behavior

* Bulk secrets creation fails when at least one of the secrets in the batch already exist.
* The error message for single secret failure is confusing and not too informative.

### Expected behavior

* Bulk secrets creation from file should skip already existing secrets and create only the new ones.
* The error message for single secret failure should hide the underlying HTTP error code and be more user-friendly.

### Reproducible example

Create some new secrets in `~/repo-secrets.env` file and add them with `garden cloud secrets create --from-file ~/repo-secrets.env`.

Add some new entries to `~/repo-secrets.env` file and rerun the secrets creation command. It will fail with the error:
```console
> garden cloud secrets create --from-file ~/repo-secrets.env

Create secrets 🔒 
 
ℹ garden                    → Initializing...
ℹ garden-cloud              → Connecting project...
✔ garden-cloud              → Ready
ℹ garden                    → Running in environment dev.true-customer-danieltaylor
ℹ secrets-command           → Creating secrets...
ℹ secrets-command           → Creating secrets... → 1/10
Encountered an unexpected Garden error. This is likely a bug 🍂
 
You can help by reporting this on GitHub: https://github.com/garden-io/garden/issues/new?labels=bug,crash&template=CRASH.md&title=Crash%3A%20Response%20code%20409%20(Conflict)
 
Please attach the following information to the bug report after making sure that the error message does not contain sensitive information:
 
HTTPError: Response code 409 (Conflict)
    at Request.<anonymous> (file:///Users/danieltaylor/Library/Application%20Support/io.garden.garden/1715246255-NGojvKX.r/rollup/garden.mjs:226907:42)
    at Object.onceWrapper (node:events:634:26)
    at Request.emit (node:events:531:35)
    at Request.emit (node:domain:488:12)
    at Request._onResponseBase (file:///Users/danieltaylor/Library/Application%20Support/io.garden.garden/1715246255-NGojvKX.r/rollup/garden.mjs:226411:22)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
[...]

PR #6064, opened 2024-05-21
refactor(cloud): refactoring and code tidying in `cloud`-commands

vvagaytsev:

Refactor the `garden cloud ...` commands code to have less code repetition and see some similarities between the command implementations.

Preliminary refactoring for fixing #6048.

See individual commits for the details. It might be easier to review this commit-by-commit.


core/src/commands/cloud/secrets/secrets-update.ts
```diff
@@ -223,17 +133,13 @@ export class SecretsUpdateCommand extends Command<Args, Opts> {
...
-        // secrets that do not exist yet and can be created
-        secretsToCreate = getSecretsToCreate(secretsToUpdateArgs, secretsToUpdate)
-      }
-    } else {
-      // update secrets by ids
-      secretsToUpdate = sortBy(allSecrets, "name")
-        .filter((secret) => Object.keys(secretsToUpdateArgs).includes(secret.id))
-        .map((secret) => ({ ...secret, newValue: secretsToUpdateArgs[secret.id] }))
+    let secretsToCreate: [string, string][] = []
+    if (isUpsert && !updateById) {
```
twelvemo, line 139: It might be a bit more readable to keep it in the else block where we already know `!updateById` or was there a specific reason to move it out?

twelvemo (commented):
Looks great! Just had one non-blocking question.

core/src/commands/cloud/secrets/secrets-update.ts
vvagaytsev: The reason was to avoid nested and too complex if-conditions. It would be nice to extract initializer methods for both arrays to declare them as `const` instead of `let`.
Moreover, it looks like these 2 flags cannot be combined, I'm gonna test it and add the validation if necessary.
That will be done in the next PR.

Maybe, another way will be to avoid fetching all secrets in advance and splitting the input data to 2 collections. Instead, we can just create secrets one-by-one and apply the necessary behaviour on conflict.

vvagaytsev: Anyway, this will be changed soon, in the next PR :)

This PR just brings some preliminary refactoring to make the further changes in create/update secrets easier.

twelvemo: Ah i see. I guess it makes sense that you can't upsert a secret by id because that would allow you to set the id if it does not exist.

twelvemo (dismissed):
👍

core/src/commands/cloud/secrets/secrets-update.ts
vvagaytsev: Yes, that definitely makes sense. But it looks like Garden CLI can accept both flags now and silently ignore `--upsert` if both are provided. That should be documented or validated with error. That's also will be done in the next PR :)

twelvemo (approved):

PR #6072, opened 2024-05-24
feat(images): add multi-platform images for garden deployed services

twelvemo:

Adds multi-platform builds for all the garden deployed containers like `buildkit`, `skopeo`, `k8s-sync`, etc. This allows users to run garden on arm64 native kubernetes.

We can not use the publish command any longer to publish the images to dockerhub, since the docker-container buildx runner is not able to load the images in the default local docker image store. To facilitate this, we now always push to DockerHub when running `garden build` in this project, but with a `dev` tag. When running `garden build --var publish=true` we push them with the specified production tag in `release_tag`.  This is a temporary workaround until we have properly integrated mutli-platform building into garden.

Also moves the images project from modules to actions ;-)

Fixes #

### DIFF
--- a/images/README.md
+++ b/images/README.md
@@ -2,4 +2,13 @@
 
 Here we place container images that we maintain and reference when using Garden.
 
-To build, just use `garden build`, and to publish, `garden publish`.
+We are building all images except the circleci image multi-platform for linux/arm64 and linux/amd64. To setup building multi-platform with buildx locally check instructions for:
+
+* Docker Desktop: https://docs.docker.com/build/building/multi-platform/#building-multi-platform-images
+* Orbstack: https://docs.orbstack.dev/docker/images#multiplatform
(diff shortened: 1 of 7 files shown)


vvagaytsev:
Nice! Let's also explicitly define `spec.publishId` for all images.

twelvemo:
> Nice! Let's also explicitly define `spec.publishId` for all images.

We can't really use it yet, because we can't use the publish command, instead we have to run `garden build --var publish=true`. I wrote some more detailed explanations in the README.

vvagaytsev:
> We can't really use it yet, because we can't use the publish command, instead we have to run `garden build --var publish=true`. I wrote some more detailed explanations in the README.

Ah, I see! Thank you :)

vvagaytsev (approved):

PR #6073, opened 2024-05-24
fix(core): issue with partial module resolution and module templates

edvald:


core/src/resolve-module.ts
```diff
@@ -496,11 +502,16 @@ export class ModuleResolver {
       buildPath,
       parentName: rawConfig.parentName,
       templateName: rawConfig.templateName,
-      inputs: rawConfig.inputs,
+      inputs: {},
```
eysi09, line 505: I'm probably missing some context, but why is this empty now?

eysi09 (approved):
Just had one non-blocking question but otherwise looks good.

core/src/resolve-module.ts
thsig: That's because the inputs are being resolved below (line 512).

PR #6078, opened 2024-05-27
improvement(helm): use --wait when deploying

### LINKED ISSUE #6053: [FEATURE]: Enable `helm --wait` flag
### Background / Motivation

We use Garden to test our helm charts in CI and we found an error that when deploying with Flux we had a mistake in the order of our hooks that only manifested in this case. We found that if we add the `helm install --wait` flag this should mimic that behavior.

Its not clear how the Helm deploy action in Garden handles this behavior and I would like to enforce the flag to be used so we can fully test hook ordering.

Add a `wait` option to the Helm deploy action configuration that controls the usage of the `--wait` flag when running `helm install`.

Self-explanatory.

There really isn't a workaround short of me writing a ConfigTemplate reimplenting the Helm deploy action, so we know that our CI is not testing completely, but it doesn't strictly block a workflow so it doesn't make things unusable.

thsig:

The `--wait` option for `helm deploy` and `helm upgrade` is now applied by default for `helm` Deploys.

We had previously used `waitForResources` for this (since Helm's rollout statuses weren't reliable enough when using `--wait`).

In the `helm` deploy handler, we now only use `waitForResources` to wait for resources updated for local or sync mode.

Fixes #6053.

### DIFF
--- a/core/src/plugins/kubernetes/helm/deployment.ts
+++ b/core/src/plugins/kubernetes/helm/deployment.ts
@@ -6,7 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import { waitForResources } from "../status/status.js"
+import { checkResourceStatuses, waitForResources } from "../status/status.js"
 import { helm } from "./helm-cli.js"
 import type { HelmGardenMetadataConfigMapData } from "./common.js"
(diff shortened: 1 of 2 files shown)


twelvemo (approved):
Nice work!

PR #6093, opened 2024-05-28
chore(images): update `circleci-runner`'s base images

vvagaytsev:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Partially addresses #6089

**Special notes for your reviewer**:

PR #6094, opened 2024-05-28
chore: update bundled NodeJS to `v22.2.0`

vvagaytsev:

Fixes #

CircleCI runner image was updated in #6093.


cli/test/unit/src/build-pkg.ts
```diff
@@ -42,7 +42,7 @@ describe("build-pkg", () => {
     })
   })
 
-  context("Node binaries", () => {
```
stefreak, line 45: The async is technically not necessary here, as there is no await statement; await statements in nested function declaration do not matter. But it also doesn't hurt 👍

stefreak (approved):
Nice 👍

cli/test/unit/src/build-pkg.ts
vvagaytsev: Thanks! I added this by accident and forgot to remove :)

PR #6096, opened 2024-05-28
chore: update sync images

vvagaytsev:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

PR #6097, opened 2024-05-28
chore(images): refactor `skopeo` image

vvagaytsev:
Use env vars for tool versions and distr names.

### DIFF
--- a/images/skopeo/Dockerfile
+++ b/images/skopeo/Dockerfile
@@ -5,13 +5,14 @@ ARG TARGETARCH
 # ECR credential helper
 RUN cd /usr/local/bin && \
   ECR_HELPER_VERSION="0.8.0" && \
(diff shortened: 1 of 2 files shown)


vvagaytsev:
It looks like we do not use this image anywhere in the code. Instead, we install skopeo as an OS-level dependency, see [this comment](https://github.com/garden-io/garden/pull/6096/files#r1617669139).

The constant for the `skopeo` image name and the relevant helper were removed in `0.13.0`, see [this commit](https://github.com/garden-io/garden/commit/a7d0c705).

stefreak:
It looks like you are right @vvagaytsev! I'd not delete the skopeo /image/ right away, but we can remove the code that builds the skopeo image from the Garden source tree.

vvagaytsev:
> It looks like you are right @vvagaytsev! I'd not delete the skopeo /image/ right away, but we can remove the code that builds the skopeo image from the Garden source tree.

Thanks @stefreak!

BTW, I can't see any code that deals with the `skopeo` image. We only have the relevant `garden.yml` config that define a Build action for that image.

I've created a draft PR #6100 - let's continue the discussion there. Closing this one as not planned/

Issue #6088, opened 2024-05-28
Update buildkit base images and dependencies


twelvemo:

vvagaytsev:
Fixed in #6072 and #6098.

PR #6098, opened 2024-05-28
chore(images): refactor `buildkit` image 

vvagaytsev:
Use env vars for tool versions and distr names.


core/src/plugins/kubernetes/constants.ts
```diff
@@ -53,9 +53,9 @@ export function getK8sSyncUtilImageName(): DockerImageWithDigest {
 export const k8sReverseProxyImageName: DockerImageWithDigest =
   "gardendev/k8s-reverse-proxy:0.1.0@sha256:df2976dc67c237114bd9c70e32bfe4d7131af98e140adf6dac29b47b85e07232"
 export const buildkitImageName: DockerImageWithDigest =
-  "gardendev/buildkit:v0.12.2-1@sha256:5b30f6fa46e1fdb89b2255b4290dd3f9072b8f91fd6927b8d428e92498fbf8d0"
+  "gardendev/buildkit:v0.12.2-3@sha256:0eaaa9aedd78168477a5704908fa14958b33b2b3384713892535f8246b50d097"
 export const buildkitRootlessImageName: DockerImageWithDigest =
-  "gardendev/buildkit:v0.12.2-1-rootless@sha256:d60e79c66832a95b89f67b1dbee255a561b20105f4d3ec9903dcc7dc4c40f19b"
+  "gardendev/buildkit:v0.12.2-3-rootless@sha256:88fabcd39eaf4bfbfd5ec2f42da699fe2ed97302b23c776c087dff27f3b14d1c"
```
stefreak, line 58: the buildkit version is already 0.13.2 (See FROM statement in Dockerfile), but our release version is 0.12.2

vvagaytsev: Right, we should bump it for less confusion.

vvagaytsev: @twelvemo I think I need your help here. Let's pair on #6088 when you have time.

twelvemo (approved):

PR #6101, opened 2024-05-29
chore(images) use latest `k8s-reverse-proxy` image

vvagaytsev:

This PR also fixes some broken local mode examples.

Fixes #

A follow-up PR for #6072. The [linuxserver/openssh-server:version-9.6_p1-r0](https://hub.docker.com/layers/linuxserver/openssh-server/version-9.6_p1-r0/images/sha256-81bcc076d4365b7498cd63384688e4301b0f99304780d1b74e9795255c62c74a?context=explore) tag seems to be rewritable, so the `k8s-reverse-proxy` got a new base image in the version `0.1.1`. This PR takes the latest `k8s-reverse-proxy` into use.

Maybe we should consider sticking to `-ls...` builds like [this](https://hub.docker.com/layers/linuxserver/openssh-server/9.6_p1-r0-ls157/images/sha256-81bcc076d4365b7498cd63384688e4301b0f99304780d1b74e9795255c62c74a?context=explore). Just to keep the base image info transparent and clear.

That can be done in a separate PR later. The local mode doesn't seem to be used too often.

### DIFF
--- a/core/src/plugins/kubernetes/constants.ts
+++ b/core/src/plugins/kubernetes/constants.ts
@@ -51,7 +51,7 @@ export function getK8sSyncUtilImageName(): DockerImageWithDigest {
 }
 
 export const k8sReverseProxyImageName: DockerImageWithDigest =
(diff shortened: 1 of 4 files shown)


twelvemo (approved):
Good catch with the base image tags!

PR #6105, opened 2024-05-29
fix(modules): another fix for the experimental partial module resolution

edvald:


core/src/resolve-module.ts
```diff
@@ -1230,3 +1248,36 @@ function missingBuildDependency(moduleName: string, dependencyName: string) {
...
   })
 }
+
+// We're hardcoding the module types here because the schemas are frozen, so it's an okay shortcut to support
+// partial resolution.
+function getServiceNames(config: ModuleConfig) {
+  const names = config.serviceConfigs.map((s) => s.name)
+  // These all have a services list field
+  if (["container", "jib-container", "exec"].includes(config.type)) {
+    names.push(...(config.spec.services || []).map((s) => s.name).filter(Boolean))
```
thsig, line 1258: Maybe I'm missing something about the module resolution flow (it's been a while since I looked at it in detail), but would this ever add any new names for these module types?

Looking at the `configure` handlers for these types, they all do something like
```typescript
moduleConfig.serviceConfigs = moduleConfig.spec.services.map((s) => ({ name: s.name, ... })
```
Maybe that's fine, but in that case we might want to deduplicate the name list in the `return` statement below?

edvald: What you're missing there is that this function is called before the configure handler is called. That's basically the hack here, to ascertain what's in the module before calling the `configure` handler.

thsig:
Looks good to me! At least the changes here are not hard to follow (and the blast radius should be small).

thsig (approved):
I'll approve this now and let you decide about my comment there.

PR #6068, opened 2024-05-23
refactor(cloud): helpers to create and update secrets

vvagaytsev:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

PR #6106, opened 2024-05-30
refactor(secrets): payload creation for `secrets update` command

vvagaytsev:

A follow-up PR to fix some bugs introduced in  #6068.

This also gets rid of unnecessary type `UpdateSecretBody`. Its `newValue` field was set but never read. The secrets to be updated did never get updated values.

Moreover, the `UpdateSecretBody.newValue` was set incorrectly, because it used  index-based access on array with the secret id/name used as an index value. That resulted to `undefined` value.

Patches #6068

* The secrets updating logic is still a bit complicated and will be simplified in the next PRs.
* This has `refactor` commit header because it's a follow-up PR, not a bug fix for an issue that exists in an official release.
* The changes were tested manually by updating secrets by names and ids


twelvemo (approved):
Looks good!

PR #6107, opened 2024-05-30
feat: allow custom args in `kubectl apply` for kubernetes deployments

### LINKED ISSUE #3839: [FEATURE]: Add server-side apply option to kubernetes action
### Background / Motivation

When applying manifests using `kubectl apply`, theres a limitation of 256kb because of how annotations work.

If you use `kubectl apply --server-side` then it doesn't suffer from this limitation, however there may be added complications with resolving conflicts.

https://kubernetes.io/docs/reference/using-api/server-side-apply/

For a given kubenetes module you should be able to enable the `--server-side` switch when garden executes `kubectl apply`

ManAnRuck:

See the Discord thread for details: https://discord.com/channels/817392104711651328/1077431043482714282

Fixes #3839

### DIFF
--- a/core/src/plugins/kubernetes/kubectl.ts
+++ b/core/src/plugins/kubernetes/kubectl.ts
@@ -63,6 +63,7 @@ export interface ApplyParams {
   pruneLabels?: { [label: string]: string }
   validate?: boolean
   retryOpts?: RetryOpts
+  applyArgs?: string[]
 }
 
 export const KUBECTL_DEFAULT_TIMEOUT = 300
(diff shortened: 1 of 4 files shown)


stefreak (approved):
For server-side apply in particular, I'd like native Garden support and/or to enable it by default if it's supported by all Kubernetes versions we support.

But it might still be useful to also allow additional parameters to kubectl as another escape hatch in the future.

ManAnRuck:
@stefreak I think I missed here something 🫣 

adding tests and the applyArgs to the apply function itself.
so currently applyArgs is not working I think 😕

PR #6108, opened 2024-05-30
chore(terraform): remove old terraform version

twelvemo:
Removes an old version of terraform, that has a dependency with a vulnerability.

Fixes #

### DIFF
--- a/plugins/terraform/cli.ts
+++ b/plugins/terraform/cli.ts
@@ -46,43 +46,6 @@ export class GlobalTerraform extends CliWrapper {
 }
 
 export const terraformCliSpecs: { [version: string]: PluginToolSpec } = {
-  "0.12.26": {
-    name: "terraform-0-12-26",
-    description: "The terraform CLI, v0.12.26",
-    type: "binary",
-    builds: [
-      {
-        platform: "darwin",
-        architecture: "amd64",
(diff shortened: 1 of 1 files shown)


10ko (approved):

vvagaytsev:
The docs should be re-generated. The rest test failures can acceptable. The integration tests are quite flaky in Acorn.

plugins/terraform/cli.ts
```diff
@@ -46,43 +46,6 @@ export class GlobalTerraform extends CliWrapper {
 }
 
 export const terraformCliSpecs: { [version: string]: PluginToolSpec } = {
-  "0.12.26": {
-    name: "terraform-0-12-26",
-    description: "The terraform CLI, v0.12.26",
-    type: "binary",
-    builds: [
```
stefreak, line 53: @twelvemo @10ko it's strange, the terraform binary should only included in the garden image if `_includeInGardenImage` is true, which is not the case for this. Is there a bug somewhere and we accidentally include all images?

I think it would be better not to remove the version completely, as that is a backwards-incompatible change, and only make sure it's not included in our image

vvagaytsev: Agree. We hluld back-port fresher versions from 0.13 instead. I'll prepare a replacement PR.

vvagaytsev: Created #6110

vvagaytsev:
Superseded by #6110

PR #6109, opened 2024-05-30
refactor(cloud): unify input key/value resource reading and mapping

vvagaytsev:

Align the flow with the `users create` command. Stick to the same validation order.

Fixes #

### DIFF
--- a/core/src/commands/cloud/secrets/secrets-create.ts
+++ b/core/src/commands/cloud/secrets/secrets-create.ts
@@ -16,6 +16,7 @@ import { PathParameter, StringParameter, StringsParameter } from "../../../cli/p
 import type { SecretResult } from "./secret-helpers.js"
 import { makeSecretFromResponse } from "./secret-helpers.js"
 import { getEnvironmentByNameOrThrow } from "./secret-helpers.js"
+import type { Secret } from "../../../cloud/api.js"
 
 export const secretsCreateArgs = {
   secrets: new StringsParameter({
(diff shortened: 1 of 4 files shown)


core/src/commands/cloud/secrets/secrets-create.ts
```diff
@@ -86,20 +87,20 @@ export class SecretsCreateCommand extends Command<Args, Opts> {
...
+      resourceName: "secret",
+      log: cmdLog,
+    })
 
     const api = garden.cloudApi
     if (!api) {
       throw new ConfigurationError({ message: noApiMsg("create", "secrets") })
     }
 
+    const secretsToCreate: Secret[] = inputSecrets.map(([key, value]): Secret => ({ name: key, value }))
```
stefreak, line 102: The outer type will be inferred automatically, if you want the code to be less verbose. But it's fine for me like this too 👍 
```suggestion
    const secretsToCreate = inputSecrets.map(([key, value]): Secret => ({ name: key, value }))
```

stefreak (approved):
👍 Thanks, makes sense!

core/src/commands/cloud/secrets/secrets-create.ts
vvagaytsev: I left it intentionally here and in the rest 2 places. Just to ensure the type-safety :) Is for some reason the inner type is removed, the outer guard will still fail if the mapper returns an incompatible value :D

stefreak: A little excessive for my taste to repeat the type in the same one-liner, but both variants are acceptable for me :)

PR #6110, opened 2024-05-30
improvement(terraform): update default terraform tool to 1.4.6

vvagaytsev:

Align it with Garden 0.13 to use the newer terraform version in the Docker image.
The older version has a dependency with a vulnerability.

Fixes #

Supersedes #6108.


twelvemo (approved):
Thanks so much for taking care of this!

PR #6112, opened 2024-05-30
improvement(k8s): allow volume mounts in runners

### LINKED ISSUE #6069: Unclear overrides of `kubernetes-pod` `podSpec` fields
### Current Behavior

I have a `kubernetes-pod` similar to the following:

```
kind: Run
type: kubernetes-pod
[...]
spec:
  cacheResult: True
  podSpec:
    volumes:
      - name: credentials-vol
        secret:
          secretName: secname

    containers:
      - name: main
        image: gcr.io/google.com/cloudsdktool/google-cloud-cli:latest
        volumeMounts:
          - name: credentials-vol
            optional: false
            mountPath: /mnt/secret
            readOnly: true
        env:
          - name: GOOGLE_APPLICATION_CREDENTIALS
            value: /mnt/secret/credentials.json
        command:

```

And I find that it never mounts the volume in the pod.

### Expected behavior

I would expect that the pod would mount the volume since that is normal behavior.

I tried to understand from the documentation what is supported however it is very vague:

> Note that the spec will be modified for the Run, including overriding with other fields you may set here (such as args and env), and removing certain fields that are not supported.

And I found that the `env` section I gave actually did work, so its also counterfactual AFAIU.

### Workaround

I can do a `kubernetes` Deploy then `kubernetes-exec` action or just write out a deployment for the `spec.file` option in `kubernetes-pod`.

### Suggested solution(s)

At minimum some documentation with warnings about what you can and can't do would help as it wastes a lot of time to have to troubleshoot this even after carefully reading the documentation.

### Your environment

* **OS:** OpenSuse Tumbleweed

`garden version`
garden version: 0.13.30

thsig:

If a `kubernetes-pod` Run or Test includes `volumeMounts` on the first container in its `podSpec`, we now include this in the runner pod.

Previously, we omitted the `volumeMounts` field from the first container in the (optional) user-supplied `podSpec` field. This is because the relevant code assumed `container` Runs/Tests—it's now appropriate for `container` and `kubernetes-pod` Runs/Tests.

Fixes #6069.

### DIFF
--- a/core/src/plugins/kubernetes/run.ts
+++ b/core/src/plugins/kubernetes/run.ts
@@ -303,8 +303,6 @@ export async function prepareRunPodSpec({
       name: mainContainerName,
       image,
       env,
-      // TODO: consider supporting volume mounts in ad-hoc runs (would need specific logic and testing)
-      volumeMounts: [],
     },
   ]
 
@@ -317,7 +315,9 @@ export async function prepareRunPodSpec({
     imagePullSecrets,
   }
 
-  if (volumes) {
+  // This logic is only relevant for `container` Runs and Tests, which need to support mounting `persistentvolumeclaim`
+  // and `configmap` actions (which are only supported for `container` actions, and are currently discouraged).
+  if (volumes && volumes.length && action.type === "container") {
     configureVolumes(action, preparedPodSpec, volumes)
   }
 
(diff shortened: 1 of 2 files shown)


edvald:
Looks good at first glance, just have some tests there that need to be updated.

thsig:
Tests are green now, should be good to go.

vvagaytsev (approved):

PR #6114, opened 2024-05-31
feat: add allowFailure option for workflow steps

### LINKED ISSUE #6077: [FEATURE]: allow failure for workflow tasks
### Background / Motivation

In some cases I want to run a workflow task that I know might fail for various reasons (see #6076 but also others).

In the workflow step you should be able to specify whether a step is allowed to fail without the entire workflow exiting as failure.

```
kind: Workflow
steps:
  - name: can-fail
    command: [...]
    allowFailure: true
```

alex-kattathra-johnson:
Fixes #6077

### DIFF
--- a/core/src/commands/workflow.ts
+++ b/core/src/commands/workflow.ts
@@ -182,6 +182,16 @@ export class WorkflowCommand extends Command<Args, {}> {
         }
       } catch (rawErr) {
         const err = toGardenError(rawErr)
+        if (step.continueOnError) {
+          result.steps[stepName] = {
+            number: index + 1,
+            outputs: {
+              stderr: err.toString(),
+            },
+            log: stepBodyLog.toString((entry) => entry.level <= LogLevel.info),
+          }
+          continue
+        }
         garden.events.emit("workflowStepError", getStepEndEvent(index, stepStartedAt))
         stepErrors[index] = [err]
         printStepDuration({ ...stepParams, success: false })
@@ -201,6 +211,9 @@ export class WorkflowCommand extends Command<Args, {}> {
       stepBodyLog.root.storeEntries = initSaveLogState
 
       if (stepResult.errors && stepResult.errors.length > 0) {
+        if (step.continueOnError) {
+          continue
+        }
         garden.events.emit("workflowStepError", getStepEndEvent(index, stepStartedAt))
         logErrors(outerLog, stepResult.errors, index, steps.length, step.description)
         stepErrors[index] = stepResult.errors
--- a/core/src/commands/workflow.ts
+++ b/core/src/commands/workflow.ts
@@ -146,8 +146,6 @@ export class WorkflowCommand extends Command<Args, {}> {
         bodyLog: stepBodyLog,
       }
 
-      let stepResult: CommandResult
-
       garden.events.emit("workflowStepProcessing", { index })
       const stepTemplateContext = new WorkflowStepConfigContext({
         allStepNames,
@@ -161,6 +159,17 @@ export class WorkflowCommand extends Command<Args, {}> {
 
       const initSaveLogState = stepBodyLog.root.storeEntries
       stepBodyLog.root.storeEntries = true
+
+      if ((!step.command && !step.script) || (step.command && step.script)) {
+        garden.events.emit("workflowStepError", getStepEndEvent(index, stepStartedAt))
+        // This should be caught by the validation layer
+        throw new InternalError({
+          message: `Workflow steps must specify either a command or a script. Got: ${JSON.stringify(step)}`,
+        })
+      }
+
+      let stepResult: CommandResult | undefined
+
       try {
         if (step.command) {
           step.command = resolveTemplateStrings({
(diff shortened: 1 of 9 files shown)


stefreak:
@alex-kattathra-johnson there seems to be a typescript error:
```
cli ┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄  ../core/test/unit/src/commands/workflow.ts(74,43): error TS2322: Type 'true' is not assignable to type 'false'.
cli ┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄  ../core/test/unit/src/commands/workflow.ts(75,32): error TS2322: Type 'true' is not assignable to type 'false'.
cli ┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄  4:26:04 AM - Project '../sdk/tsconfig.json' can't be built because its dependency '../core' has errors
```
You'll need NodeJS 22.2 on your machine, and then you can run `npm install` and then `npm run check-all` to find typescript errors

Let us know on Discord in the #dev channel ifyou need any support setting your dev environment up. You can find the Discord link and further information about how to contribute here: https://github.com/garden-io/garden/blob/main/docs/contributing/README.md

alex-kattathra-johnson:
@stefreak: Got the errors resolved and the builds successful. Thank you!

core/test/unit/src/commands/workflow.ts
```diff
@@ -82,6 +82,28 @@ describe("RunWorkflowCommand", () => {
...
+        apiVersion: GardenApiVersion.v0,
+        name: "workflow-a",
+        kind: "Workflow",
+        internal: {
+          basePath: garden.projectRoot,
+        },
+        files: [],
+        envVars: {},
+        resources: defaultWorkflowResources,
+        steps: [{ script: "echo fail!; exit 1", allowFailure: true }],
```
stefreak, line 97: @alex-kattathra-johnson thank you so much for the effort! Looking pretty good to me so far.

Looking at the code and the test cases I have one question: what behaviour would you expect for subsequent steps?

The behaviour of the current implementation would be that subsequent steps only run with `when: always` or `when: onError`, so configuring `allowFailure` does not have an effect on subsequent steps, only on the overall workflow success. If that's what we want, maybe it would be good to add a few words to the docs to clarify.

Can you also add test cases for subsequent steps, and update the docs at https://github.com/garden-io/garden/blob/main/docs/using-garden/workflows.md to mention the new `allowFailure` option?

Thank you so much! Great work ❤️

core/src/commands/workflow.ts
```diff
@@ -205,14 +207,15 @@ export class WorkflowCommand extends Command<Args, {}> {
         logErrors(outerLog, stepResult.errors, index, steps.length, step.description)
         stepErrors[index] = stepResult.errors
         // There may be succeeding steps with `when: onError` or `when: always`, so we continue.
+        shouldError = shouldError || !step.allowFailure
```
stefreak, line 210: The comment was meant for the `continue` line, so let's keep them together
```suggestion
        shouldError = shouldError || !step.allowFailure
        // There may be succeeding steps with `when: onError` or `when: always`, so we continue.
```

core/src/commands/workflow.ts
```diff
@@ -186,6 +187,7 @@ export class WorkflowCommand extends Command<Args, {}> {
         stepErrors[index] = [err]
         printStepDuration({ ...stepParams, success: false })
         logErrors(stepBodyLog, [err], index, steps.length, step.description)
+        shouldError = shouldError || !step.allowFailure
```
stefreak, line 190: It is more verbose, but I'd find this slightly more readable:
```suggestion
        if (!step.allowFailure) {
            shouldError = true
        }
```

alex-kattathra-johnson:
@stefreak : I've updated the PR based on the discussion from the issue. Could you please do another review when a chance is afforded?

core/src/commands/workflow.ts
```diff
@@ -204,15 +209,18 @@ export class WorkflowCommand extends Command<Args, {}> {
...
 
       garden.events.emit("workflowStepComplete", getStepEndEvent(index, stepStartedAt))
       printStepDuration({ ...stepParams, success: true })
     }
 
-    if (size(stepErrors) > 0) {
+    if (shouldFail) {
       printResult({ startedAt, log: outerLog, workflow, success: false })
       garden.events.emit("workflowError", {})
       // TODO: If any of the errors are not instanceof GardenError, we need to log the explanation (with bug report information, etc.)
```
stefreak, line 226: In the code below, we get all the `stepErrors` and construct an error message explaining to the user why the workflow failed.

In the code above however, we always add the error to `stepErrors`, even if the workflow was allowed to continue.

core/src/config/workflow.ts
```diff
@@ -255,21 +256,27 @@ export const workflowStepSchema = createSchema({
...
+
       \`always\`: This step will always be run, regardless of whether any preceding steps have failed.
 
       \`never\`: This step will always be ignored.
 
       See the [workflows guide](${DOCS_BASE_URL}/using-garden/workflows#the-skip-and-when-options) for details
       and examples.
       `),
+    allowFailure: joi
+      .boolean()
```
stefreak, line 273: We should add a default: 
```suggestion
      .boolean()
      .default(false)
```

core/src/config/workflow.ts
```diff
@@ -255,21 +256,27 @@ export const workflowStepSchema = createSchema({
...
+      If the next step has \`when: onFailure\`, it will also be run. Otherwise, all subsequent steps are ignored.
+
       \`always\`: This step will always be run, regardless of whether any preceding steps have failed.
 
       \`never\`: This step will always be ignored.
 
       See the [workflows guide](${DOCS_BASE_URL}/using-garden/workflows#the-skip-and-when-options) for details
       and examples.
       `),
+    allowFailure: joi
```
stefreak, line 272: During deep review together with @twelvemo we noticed that it might be easier to explain / understand for a user if the flag name focuses on the behaviour in regards to the subsequent steps; We'd propose the name `continueOnError` with default being false.

twelvemo (changes requested):
Great job on this! I think adding an `allowFailure` flag on a step is a great idea, however after a longer discussion @stefreak and I figured that in the current format the distinction between `onFailure` and `onError` is a bit difficult to understand. This is not due to your implementation, which is great. Some more suggestions on how to proceed in another comment soon.

stefreak (changes requested):
@alex-kattathra-johnson Wow, great job, really amazing how you drilled down into the workflow code, and your code looks like it behaves exactly as proposed by @salotz in #6077

During deeper review @twelvemo and me took the time to think again about the user perspective, and while initially I thought adding `onFailure` is a good idea, it turns out to make workflows harder to explain and workflow code harder to read.

Our proposal is the following:
1. Let's go back to the older version of the PR where you only added `allowFailure` (And possibly rename that to `continueOnError`. Let's keep the added tests from the current version 👍 
2. Let's open an issue and postpone the `onFailure` handler to another PR; What we think might be ideal is something like the following, but implementation of that is too complex to shove it into this PR.
```
kind: Workflow
name: test
steps:
 - name: allowed_to_fail 
   script: this_will_fail
   continueOnError: true
 - if: ${steps.allowed_to_fail.outcome == "error"} 
   script: echo "Step 1 failed, let's clean it up"
```

core/src/commands/workflow.ts
stefreak: For the next version of this PR: if `continueOnError` was `true`, then if we don't add the step error to `stepErrors`, then the error message will still be correct should there be any other error.

Otherwise users might be confused, as they are trying to solve an error that wasn't actually causing the workflow to fail.

alex-kattathra-johnson:
Sliced the PR with the Occam's razor! :)

twelvemo (approved):
Looks good! Thanks for taking the time to rework it again!

stefreak (approved):
@twelvemo and me made some additional changes to the logging and error handling behaviour. Looks good to go now, if the tests pass

stefreak (approved):

PR #6116, opened 2024-05-31
fix(ext-tools): remove default `cwd` for external tools

vvagaytsev:

All Garden bundled tools' paths are known to be absolute. There is no reason to set `cwd` for running any bundled tool, because it does not have any effect.

But, it can cause some troubles with particular tools. The latest `kubectl`, that uses Go 1.19+, can fail,
because of the change made in Go 1.19,
see https://pkg.go.dev/os/exec#hdr-Executables_in_the_current_directory

So, it is safer to remove the unnecessary default `cwd` definition.

Fixes #

Failures in [`e2e-vote-helm`](https://app.circleci.com/pipelines/github/garden-io/garden/25249/workflows/d19d905d-364b-47e5-9aea-02f41ac920eb/jobs/522306) were fixed by renaming the redis-action directory.

When the Helm binary's `cwd` contains a directory with the same name as a Helm Chart, Helm fails with the error:
```console
INSTALLATION FAILED: Chart.yaml file is missing
```

That's a known Helm issue. The error message is confusing, but it is improved in another PR, see #6117.


core/src/util/ext-tools.ts
```diff
@@ -71,9 +72,6 @@ export abstract class CliWrapper {
     if (!args) {
       args = []
     }
-    if (!cwd) {
-      cwd = dirname(path)
-    }
 
     log.silly(() => `Execing '${path} ${args!.join(" ")}' in ${cwd}`)
```
stefreak, line 76: Let's log the actual CWD, instead of `undefined`
```suggestion
    log.silly(() => `Execing '${path} ${args!.join(" ")}' in ${cwd | resolve('.'}`)
```

stefreak (changes requested):
Couldn't this break customer code as well, if their directory names happen to collide? I got another idea how we can fix the problem: https://github.com/garden-io/garden/pull/6120

WDYT?

vvagaytsev:
@stefreak agree, #6120 looks better and narrower. He might have some unforeseen risks. Closing this.

By the way, it might be still ok to have #6117. I'm not sure if there is a way to configure a `cwd` for the bundled Helm binary. If there is, then the hint from #6117 could be helpful.

PR #6120, opened 2024-05-31
fix(k8s): prevent exec auth script errors on Windows

stefreak:
Prevent Windows from adding the current directory to the PATH implicitly. Since Go 1.19, Go's exec implementation fails to run if binary found in implicitly added relative directory.

Adding this environment variable resolves that problem. See also https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-needcurrentdirectoryforexepatha

Fixes #


stefreak (commented):
Looks good to me 👍

vvagaytsev (approved):

PR #6145, opened 2024-06-05
tool(mutagen): update to version `0.17.6`

vvagaytsev:

Updated the native Mutagen client to the latest available stable version `0.17.6`.
The native Mutagen client is still available via `GARDEN_ENABLE_NEW_SYNC=1` feature flag.

Fixes #

### DIFF
--- a/core/src/mutagen.ts
+++ b/core/src/mutagen.ts
@@ -904,7 +904,7 @@ export function parseSyncListResult(res: ExecaReturnValue): SyncSession[] {
 }
 
 const mutagenVersionLegacy = "0.15.0"
-const mutagenVersionNative = "0.17.5"
+const mutagenVersionNative = "0.17.6"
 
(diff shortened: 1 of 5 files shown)


core/src/mutagen.ts
```diff
@@ -904,7 +904,7 @@ export function parseSyncListResult(res: ExecaReturnValue): SyncSession[] {
 }
 
 const mutagenVersionLegacy = "0.15.0"
-const mutagenVersionNative = "0.17.5"
+const mutagenVersionNative = "0.17.6"
```
stefreak, line 907: Looks good to me 👍 
But should we turn it the other way around, have an escape hatch to go back to the old mutagen in case there are sync issues, and use the new one as a default?

stefreak (approved):
👍

core/src/mutagen.ts
vvagaytsev: @stefreak that's the plan, we'll do it soon in a separate PR. I discussed that with @eysi09 and @thsig yesterday.

PR #5744, opened 2024-02-19
improvement(sync): update Mutagen to 0.17.5

vvagaytsev:
**What this PR does / why we need it**:

This PR updates the Mutagen tool to the latest available version. 

**Which issue(s) this PR fixes**:

This should fix #5053

**Special notes for your reviewer**:

PR #6148, opened 2024-06-05
improvement(sync): better error handling with solution suggestion

vvagaytsev:

Handle more failure cases when switching between sync daemons.
Add more informative error message with suggested solutions.

Fixes #

This is a prerequisite for using the native Mutagen (see #5551 #5744 ) by default.


vvagaytsev:
Converted to draft, need to check pone more test scenario.

vvagaytsev:
Ready for review again. @eysi09 can you take a look, please? After this, switching to the native (i.e. non-forked) Mutagen should be easier from the UX point of view.

core/src/mutagen.ts
```diff
@@ -686,6 +708,13 @@ export class Mutagen {
           this.log.warn(
             styles.primary(`Could not connect to sync daemon, retrying (attempt ${loops}/${maxRetries})...`)
           )
+          if (loops === 1) {
+            // this happens in any way of changing sync modes and when:
```
twelvemo, line 712: How do we know this is an issue with switching the mutagen daemon only because it is the first failure?

twelvemo (commented):
Looks good, the error messages are very clear! Had one question on when we emit them though.

core/src/mutagen.ts
vvagaytsev: In theory, a root caused can be different, but must likely it would be caused by Mutagen switch. There is no way to verify the actual underlying cause from the error message at this point.

The `if`-condition is there just to print the warning and the hint once, to avoid flooding the output console with repeated warnings.

twelvemo: Gotcha! Just wondering if we should adjust the error message a bit in that case, because it might lead you on a wrong path if you haven't updated the mutagen daemon and it is failing due to different reasons?

core/src/mutagen.ts
```diff
@@ -328,10 +328,25 @@ class _MutagenMonitor extends TypedEventEmitter<MonitorEvents> {
...
 function logMutagenDaemonWarning(log: Log) {
+  const daemonStopCommand = `GARDEN_ENABLE_NEW_SYNC=${!gardenEnv.GARDEN_ENABLE_NEW_SYNC} garden util mutagen daemon stop`
+  const redeploySyncCommand = `GARDEN_ENABLE_NEW_SYNC=${gardenEnv.GARDEN_ENABLE_NEW_SYNC} garden deploy --sync`
+
   log.warn(
-    dedent`
-    It looks like you've changed to a different version of the sync daemon and therefore the sync daemon needs to be restarted.
-    Please see our Troubleshooting docs for instructions on how to restart the daemon for your platform: ${makeDocsLinkStyled("guides/code-synchronization", "#restarting-sync-daemon")}`
+    deline`
+    It looks like you've changed to a different version of the sync daemon.\n
```
twelvemo, line 336:
```suggestion
    It looks like you might have changed to a different version of the sync daemon.\n
```

vvagaytsev: Make sense! Thank you for the improvement! 👍

core/src/mutagen.ts
vvagaytsev: Sure, good catch!

twelvemo (approved):

PR #6152, opened 2024-06-07
refactor(vcs): less coupling between `VcsHandler` and `Garden` classes

vvagaytsev:
Class `VcsHandler` already stores the same `projectRoot` as `Garden` does. No need to read the project root from Garden.
Thus, the `VcsHandler` references `Garden` only to use its warning-emitting machinery. That can be decoupled as well.

Fixes #


twelvemo (approved):
Nice 🚀

PR #6153, opened 2024-06-07
chore(cloudbuilder): add runtime information to build status events

stefreak:
Add runtime information to build status events, so we can display them
in the Cloud UI

Because the runtime needs to be sent to Garden Cloud as early as possible to display the "Cloud builder" status in the command result page correctly, this functionality has been added quite deep in the framework.

I think that's ok because those flags can have a meaning for all other kinds (e.g. Run, Deploy, Test) as well.

We might want to be able to add additional information, e.g. the builder id– but that can be done at a later point in time.

**Additional information**:
Garden Cloud discards additional information if it's not supported yet, but still processes events containing unknown information. This is great as it allows backwards-compatibility with older versions of Garden Cloud.


10ko (approved):
Looks good to me, can @vvagaytsev or someone else (perhaps @eysi09 since you worked on this previusly) also take a look?

My only concern is the term "actual" might be slightly confusing but to be honest I cannot find a better term out of the top of my head, so if we agree on that, I am totally ok with it. :)

core/src/plugins/container/cloudbuilder.ts
```diff
@@ -67,6 +68,50 @@ export const cloudBuilder = {
...
+
+    // if cloud builder is configured AND available, that's our actual runtime. Otherwise we fall back to whatever is configured in the plugin.
+    const actual = (await cloudBuilder.isConfiguredAndAvailable(ctx, action)) ? preferred : fallback
+
+    if (actual === preferred) {
+      return {
+        actual,
+      }
+    } else {
+      const unavailable = await getAvailability(ctx, action)
```
twelvemo, line 101: Does it really make sense to call `getAvailability` again here? It is also called by `cloudBuilder.isConfiguredAndAvailable`. The inconsistent state seems rather unlikely.

twelvemo (changes requested):

core/src/plugins/container/cloudbuilder.ts
stefreak: I agree this is confusing; It's needed to get `unavailable.reason`.

The availability will be cached in the LRUCache, because as you said we call it right before. The same thing happens when the plugin calls `withBuilder`; it calls `isConfiguredAndAvailable` before, for decision-making, and then `withBuilder` calls `getAvailability` again to retrieve the credentials from the availability response.

To make sure that we call `getAvailability` only once and pass the result as a parameter to `withBuilder` would require a thorough refactoring of the build handlers in the Kubernetes and container plugins.

stefreak: Usually, we even call `isConfiguredAndAvailable` before `getActionRuntime` e.g. here: https://github.com/garden-io/garden/pull/6153/files#diff-9908a5a1399edfb110bb66a430136f2b9c0a6350a1f48d19ccc8b859fd212173R100

twelvemo: I see, that makes sense. This is okay for me 👍

twelvemo (approved):

eysi09 (approved):
This looks good to me. Great work!

I think it's great in general to include the "runtime" with the action payload and we can add it for more kinds when we need them.

PR #6155, opened 2024-06-10
tool(sync): print warn on upcoming code sync daemon change

vvagaytsev:

A follow-up PR for #6148.

This makes the users aware about the upcoming change of the code syncing daemon.

Fixes #


twelvemo (approved):
Great work!

Issue #6146, opened 2024-06-05
Varfiles with "type: helm" and sources.repository.url not working


ManAnRuck:
### Current behavior

env file not found (is looking in wrong directory for the varfiles in .garden/sources/…)
```sh
Error processing config for Deploy action redis:
Could not find varfile at path './.env'. Absolute path: /…/garden/examples/redis/.garden/sources/action/deploy.redis--88237447a8/.env
```

### Expected behavior

varfiles should be searched within the action directory

### Reproducible example

garden.yml
```yaml
apiVersion: garden.io/v1
kind: Project
name: redis
environments:
  - name: local
    variables:
      baseHostname: redis.local.demo.garden
providers:
  - name: local-kubernetes
    environments: [local]
variables:
  userId: ${kebabCase(local.username)}

```
redis.garden.yml
```yaml
kind: Deploy
description: Redis service for queueing votes before they are aggregated
type: helm
name: redis
varfiles:
  - ./.env
source:
  repository: 
    url: https://github.com/bitnami/charts#main
spec:
  chart:
    path: bitnami/redis
  values:
    auth:
      enabled: false

```
.env
```
TEST=test
```

```sh
garden dev
garden deploy
```

### Workaround

don't use varfiles

### Suggested solution(s)

varfiles location should be relative to the action yaml definition file

### Your environment

* **OS:** macOS
* **How I'm running Kubernetes:** docker desktop

`0.13.30`

ManAnRuck:
i am currently thinking about what is a better solution.
1. trying to resolve varfiles and value files relative to original file
2. or copy varfiles and valuefiles into the .garden action directory
any suggestions or other ideas?

PR #6147, opened 2024-06-05
fix: handling varfiles in remote actions

ManAnRuck:
**What this PR does / why we need it**:

fix read varfiles in helm deployments with `sources.repository.url`

**Which issue(s) this PR fixes**:

Fixes #6146

**Special notes for your reviewer**:
It's just a dirty try to fix it. maybe there are some better solutions

PR #6156, opened 2024-06-10
fix: 🐛 load valuefiles from config file location

ManAnRuck:
What this PR does / why we need it:

fix read `valueFiles` in helm deployments with sources.repository.url

Which issue(s) this PR fixes:

Fixes https://github.com/garden-io/garden/issues/6146

Special notes for your reviewer:
It's same like #6147 just for `valueFiles` @stefreak 

---

I did test it against this (new uncommitted) example:

<details><summary>show example</summary>

examples/remote-helm/garden.yml
```yaml
apiVersion: garden.io/v1
kind: Project
name: remote-helm
environments:
  - name: local
    variables:
      baseHostname: remote-helm.local.demo.garden
providers:
  - name: local-kubernetes
    environments: [local]
variables:
  userId: ${kebabCase(local.username)}
```

examples/remote-helm/redis.garden.yml
```yaml
kind: Deploy
description: Redis service for queueing votes before they are aggregated
type: helm
name: redis
varfiles:
  - ./.env
source:
  repository: 
    url: https://github.com/bitnami/charts#main
spec:
  chart:
    path: bitnami/redis
  valueFiles:
    - values.yml
```

examples/remote-helm/values.yml
```yaml
auth:
  enabled: false
```

examples/remote-helm/.env
```env
TEST=test
```

</details>

### DIFF
--- a/core/src/plugins/kubernetes/helm/common.ts
+++ b/core/src/plugins/kubernetes/helm/common.ts
@@ -294,7 +294,7 @@ export async function getValueArgs({
   // so it's added to the end of the list.
   const valueFiles = action
     .getSpec()
-    .valueFiles.map((f) => resolve(action.getBuildPath(), f))
+    .valueFiles.map((f) => resolve(action.effectiveConfigFileLocation(), f))
     .concat([valuesPath])
 
   const args = flatten(valueFiles.map((f) => ["--values", f]))
--- a/core/src/plugins/kubernetes/helm/common.ts
+++ b/core/src/plugins/kubernetes/helm/common.ts
@@ -10,7 +10,7 @@ import { flatten, isPlainObject } from "lodash-es"
 import AsyncLock from "async-lock"
 import { join, resolve } from "path"
 import fsExtra from "fs-extra"
-const { pathExists, readFile, remove, writeFile } = fsExtra
+const { pathExists, readFile, remove, writeFile, pathExistsSync } = fsExtra
 import { temporaryWrite } from "tempy"
 import cryptoRandomString from "crypto-random-string"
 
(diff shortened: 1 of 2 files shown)


ManAnRuck:
@stefreak maybe u can also have a look here because it's nearly the same like #6147

ManAnRuck:
@vvagaytsev @stefreak maybe u can also have a look here because it's nearly the same like https://github.com/garden-io/garden/pull/6147

stefreak:
@ManAnRuck thank you for. the contribution!

Even though I understand it is confusing, I believe this one is slightly different than #6147, because the current code looks in the `action.getBuildPath()` explicitly, and changing it to `action.effectiveConfigFileLocation()` would potentially be a breaking change for existing users.

If we add test cases for the behaviour as well, I would be ok with a compromise: We first check if the valueFile exists in build path, and if yes we use it (to preserve the current behaviour) – if not, we fall back to looking for it at the Garden config file location.

In the next major version (0.14) we can add a breaking change to stop looking for valueFile existence in the build path.

WDYT? 

ping @vvagaytsev

ManAnRuck:
yes I like this approach :)

I already add a check if the file exists in both directories and a more readable error if the file is not found.
for the test I maybe need some support.

stefreak (changes requested):
thanks for the contribution @ManAnRuck, and thank you for refactoring this to use the async file system function @vvagaytsev 

This PR looks good to me, but what is missing is one test that cements this behaviour and mentioning this in the docs. With those two changes I'd be happy to merge this 👍

ManAnRuck:
I think I need some help.

https://github.com/garden-io/garden/blob/main/docs/contributing/developing-garden.md#tests
(there are some typos which I can fix by another pr (`npm run run e2e`))

when I run `npm run e2e` I get the following error
```
Test action "e2e-tests" was not found. Available actions: e2e-tests-kaniko, e2e-tests-tasks, e2e-tests-variables, e2e-tests-vote-helm, e2e-tests-vote, e2e-tests-remote-sources and e2e-tests-demo-project
```

when I then run `./bin/garden test e2e-tests-kaniko`
```
Error while copying from '/Users/.../garden/e2e/projects/code-synchronization' to '/Users/.../garden/.garden/build/e2e-tests/e2e/projects/code-synchronization': Source is neither a symbolic link, nor a file.
```

---

regarding the documentation, isn't the current description already the one that describes how it should actually work? i.e. put the files in the config instead of the build folder?

> Note that the paths here should be relative to the config root, and the files should be contained in this action config's directory.
https://docs.garden.io/reference/module-types/helm#valuefiles

ManAnRuck:
@stefreak I added now an integration test. hope this is enough and we don't need e2e tests for this.

vvagaytsev:
Thanks for the update, @ManAnRuck! Feel free to create another PR to fix the testing guidelines! Huge thanks for pointing that out! 👍 

The changes look good to me. @stefreak can you take a look too, please?

stefreak (approved):
Looks good to me 👍

PR #6176, opened 2024-06-11
improvement(core): fix missing action type err msg

### LINKED ISSUE #5401: Unhelpful error message when provider configuration is missing
### Current Behavior

When running `garden deploy` with one of the reproducable example configs below, I get

```
Error processing config for Deploy action test1:
Unrecognized action type 'helm' (defined at garden.yml). Are you missing a provider configuration?

Currently available action types: Build, Deploy, Run, Test
```

### Expected behavior

I expect error messages similar to this:

> Error validating Project 'repro':
> 
> ```
> ...
> 34  |     environments:
> 35  |       - locall,
> --------------^
> ```
> providers[0][environments][0] with value "locall" refers to an environment that does not exist. Available environments: "local"

Or, if no such mistake is in the config and the kubernetes provider and the `local` environment are just not associated, then I expect an error like this:

> Error processing config for Deploy action test1:
> Unrecognized action type 'helm' (defined at garden.yml).
> 
> The 'helm' action type is provided by the 'kubernetes', 'local-kubernetes' and 'ephemeral-kubernetes' providers, but 
> the current environment 'local' is lacking the configuration for one of those providers. Consider adding a provider 
> configuration like so in your project-level configuration (`project.garden.yml`):
> ```
>    providers:
>      - name: kubernetes # Or local-kubernetes, or ephemeral-kubernetes
>        environments: [local]
> ```

### Reproducible example

```
apiVersion: garden.io/v1
kind: Project
name: repro
defaultEnvironment: local
environments:
  - name: local
providers:
  - name: local-kubernetes
    setupIngressController: false
    #environments: []
    environments:
      - locall
---
kind: Deploy
type: helm
name: test1
description: test
```

and also for this case the error could be improved:

```
apiVersion: garden.io/v1
kind: Project
name: repro
defaultEnvironment: local
environments:
  - name: local
providers:
  - name: local-kubernetes
    setupIngressController: false
    environments: []
---
kind: Deploy
type: helm
name: test1
description: test
```

### Workaround

None

### Suggested solution(s)

See expected behaviour

### Your environment

* **OS:** macos
* **How I'm running Kubernetes:** orbstack

`garden version` 0.13.19

thsig:

Before this fix, we'd erroneously list the available action kinds instead of the available action types (for the invalid action's kind).

Fixes #5401.

### DIFF
--- a/core/test/unit/src/garden.ts
+++ b/core/test/unit/src/garden.ts
@@ -4312,7 +4312,11 @@ describe("Garden", () => {
       ])
 
       await expectError(() => garden.resolveModules({ log: garden.log }), {
-        contains: ["Unrecognized action type 'invalidtype'", "Are you missing a provider configuration?"],
+        contains: [
+          "Unrecognized action type 'invalidtype'",
+          "Currently available 'Build' action types: 'container', 'exec', 'test'",
+          "Are you missing a provider configuration?",
+        ],
       })
     })
 
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -408,15 +408,16 @@ async function processActionConfig({
   scanRoot?: string
 }) {
   const actionTypes = await garden.getActionTypes()
-  const definition = actionTypes[config.kind][config.type]?.spec
-  const compatibleTypes = [config.type, ...getActionTypeBases(definition, actionTypes[config.kind]).map((t) => t.name)]
+  const { kind, type } = config
+  const definition = actionTypes[kind][type]?.spec
+  const compatibleTypes = [type, ...getActionTypeBases(definition, actionTypes[kind]).map((t) => t.name)]
 
   const configPath = relative(garden.projectRoot, config.internal.configFilePath || config.internal.basePath)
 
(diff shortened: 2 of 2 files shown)


core/test/unit/src/garden.ts
```diff
@@ -4312,7 +4312,11 @@ describe("Garden", () => {
       ])
 
       await expectError(() => garden.resolveModules({ log: garden.log }), {
-        contains: ["Unrecognized action type 'invalidtype'", "Are you missing a provider configuration?"],
+        contains: [
+          "Unrecognized action type 'invalidtype'",
+          "Currently available 'Build' action types: 'container', 'exec', 'test'",
+          "Are you missing a provider configuration?",
```
stefreak, line 4318: There is not a good page in the docs that introduces the concept of providers. The closest would probably be https://docs.garden.io/reference/project-config … should we still help the user a little bit here and ask them to have a look at the docs?

Can we find the provider names that provide the referenced action type and list them as well?

stefreak (approved):
I do have two comments, but not a strictly required to fix before merging this.

PR #6179, opened 2024-06-14
fix(git): fix repo scan result caching

vvagaytsev:

This PR fixes cache key construction in `GitRepoHandler` that accounts for different includes/excludes and filter functions for the same path.

Originally, the caching was implemented in #5710. This PR makes sure that 2 filter-objects are considered tom have the equal cache keys if the difference is only in the order of the `augmentedExcludes` and `augmentedIncludes` arrays, ad there is no difference between the sets of values of both arrays.

A follow-up PR for #5526 and #5710

~The changes made #5710 make the caching in `GitRepoHandler.scanRoot(...)` unnecessary. That caching logic will be removed in a separate PR.~ The caching in `GitRepoHandler.scanRoot(...)` might still make sense. It caches the results for the whole path, without taking include/exclude filters into account.

### DIFF
--- a/core/src/vcs/git-repo.ts
+++ b/core/src/vcs/git-repo.ts
@@ -60,7 +60,7 @@ const getIncludeExcludeFiles: IncludeExcludeFilesHandler<GitRepoGetFilesParams, 
   return { include, exclude, augmentedIncludes, augmentedExcludes }
 }
 
-function getHashedFilterParams({
+export function getHashedFilterParams({
   filter,
   augmentedIncludes,
   augmentedExcludes,
--- a/core/src/vcs/git-repo.ts
+++ b/core/src/vcs/git-repo.ts
@@ -60,6 +60,24 @@ const getIncludeExcludeFiles: IncludeExcludeFilesHandler<GitRepoGetFilesParams, 
   return { include, exclude, augmentedIncludes, augmentedExcludes }
 }
 
+function getHashedFilterParams({
+  filter,
+  augmentedIncludes,
+  augmentedExcludes,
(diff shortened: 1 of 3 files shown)


core/src/vcs/git-repo.ts
```diff
@@ -60,7 +61,25 @@ const getIncludeExcludeFiles: IncludeExcludeFilesHandler<GitRepoGetFilesParams,
...
+  augmentedIncludes,
+  augmentedExcludes,
+}: {
+  filter: ((path: string) => boolean) | undefined
+  augmentedIncludes: string[]
+  augmentedExcludes: string[]
+}) {
+  return hashString(
+    stableStringify({
+      filter: filter ? filter.toString() : undefined, // We hash the source code of the filter function if provided.
```
stefreak, line 75: Because it's such an unusual thing to do, and it has some implications (e.g. captured variables will not affect the hash key) I think it's better to not hide the fact that we are converting the function to a string, by changing the signature of the function, something like the following:

```
export function getHashedFilterParams({
  filterFnSource,
  augmentedIncludes,
  augmentedExcludes,
}: {
  filterFnSource: string | undefined
  augmentedIncludes: string[]
  augmentedExcludes: string[]
}) {
```

stefreak (commented):
Nice, looks good to me and this is a good fix

I have one minor comment about the function signature of `getHashedFilterParams`; I want to make sure that we are not hiding details that are important later to understand implications of making changes to the code

stefreak (approved):
Let's merge this anyway, the nit can also be addressed later.

PR #6196, opened 2024-06-19
refactor: cloud builder availability

stefreak:
During review of #6153, Anna rightfully pointed out that the availability code is quite confusing (https://github.com/garden-io/garden/pull/6153#discussion_r1633020040 ). We've looked a little more closely at it together and concluded that the code even has potential for weird race conditions, where we are reporting building in cloud builder but actually built the image locally.

Worse yet, builds might fail with an InternalError if the user is unlucky ("You must call isConfiguredAndAvailable before calling withBuilder").

This refactor eliminates these weird races and also makes the code easier to understand and reason about at the same time.

Thanks Anna :)

### DIFF
--- a/core/src/plugins/container/cloudbuilder.ts
+++ b/core/src/plugins/container/cloudbuilder.ts
@@ -15,6 +15,7 @@ import { styles } from "../../logger/styles.js"
 import type { KubernetesPluginContext } from "../kubernetes/config.js"
 import { uuidv4 } from "../../util/random.js"
 import fsExtra from "fs-extra"
+
 const { mkdirp, rm } = fsExtra
 import { join } from "path"
 import { tmpdir } from "node:os"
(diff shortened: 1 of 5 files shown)


twelvemo (dismissed):
🚀

core/src/plugins/container/cloudbuilder.ts
```diff
@@ -45,6 +47,28 @@ export const cloudBuilder = {
       return false
     }
```
vvagaytsev, line 49: This function is called `isConfigured`, but it checks only the `isCloudBuilderEnabled` flag of the `CloudBuilderConfiguration` and ignores the value of `isInClusterBuildingConfigured`.

This looks a bit confusing. Should we expose this function as a part of the API? We can call `getConfiguration` instead of this and check the necessary flags on the caller-side.

stefreak: I'd like to avoid exposing information in a public interface that is not needed; In the Kubernetes plugin, the ctx.config object should be consulted directly if one wants to get the in-cluster building config status.

stefreak: Resolved in pairing session

PR #6205, opened 2024-06-21
refactor: helper to get action source path

vvagaytsev:

Avoid code duplication in VCS logic and in the `BaseAction` class.

Fixes #


twelvemo (approved):
Nice 🚀

PR #6207, opened 2024-06-21
improvement(docs): do not use `build` field in migration guide

eysi09:

During the Bonsai release there was some confusion around the best practices for referencing a container build in a `container` Deploy action.

Originally we used the `build` field for this but it has caused confusion and only works for Deploy actions of type `container`, not e.g. `helm` or `kubernetes`.

So for consistency, we decided that the best practice should be to always explicitly set the corresponding Build action as a dependency for ALL Deploy actions (i.e. any type).

This commit updates the Bonsai migration guide to reflect that. Previously we've updated all our examples and other parts of our documentation.

Fixes #

### DIFF
--- a/docs/guides/migrating-to-bonsai.md
+++ b/docs/guides/migrating-to-bonsai.md
@@ -73,6 +73,8 @@ However, there are some caveats:
 - Modules cannot depend on actions
 - Modules cannot reference actions
 - Actions can reference and depend on modules, by referencing the actions that are generated from modules.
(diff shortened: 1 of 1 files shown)


stefreak (approved):

vvagaytsev (approved):

vvagaytsev:
LGTM! 👍 Thanks @eysi09!

Should we also fire a (deprecation) warning if the `build` field is used? That can be done in a separate PR.

PR #6208, opened 2024-06-21
feat: add building and publishing multi-platform images

twelvemo:
Adds the config option and capability to trigger multi-platform builds. To allow publishing multi-platform images with the publish command, the publish command now uses `regctl` whenever possible. This allows us to copy newly tagged images directly without the need to pull them down to a local image store first. This is more performant, enables usage of the publish command in container based CIs and it is necessary in the context of multi-platform builds because most local image stores don't support storing multi-platform images.

- feat(container): add multi platform builds
- feat(container): publish images without the need to pull them locally

Fixes #

Uses regctl instead of skopeo, because it also has windows builds.

### DIFF
--- a/core/src/plugins/kubernetes/container/publish.ts
+++ b/core/src/plugins/kubernetes/container/publish.ts
@@ -8,41 +8,37 @@
 
 import type { ContainerBuildAction } from "../../container/moduleConfig.js"
 import type { KubernetesPluginContext } from "../config.js"
-import { pullBuild } from "../commands/pull-image.js"
 import type { BuildActionHandler } from "../../../plugin/action-types.js"
 import { containerHelpers } from "../../container/helpers.js"
 import { naturalList } from "../../../util/string.js"
+import { cloudBuilder } from "../../container/cloudbuilder.js"
 
 export const k8sPublishContainerBuild: BuildActionHandler<"publish", ContainerBuildAction> = async (params) => {
   const { ctx, action, log, tagOverride } = params
   const k8sCtx = ctx as KubernetesPluginContext
   const provider = k8sCtx.provider
 
+  const cloudBuilderUsed = await cloudBuilder.getAvailability(k8sCtx, action)
+
   const localImageId = action.getOutput("localImageId")
-
-  if (provider.config.buildMode !== "local-docker") {
-    // NOTE: this may contain a custom deploymentRegistry, from the kubernetes provider config
-    const deploymentRegistryImageId = action.getOutput("deploymentImageId")
-
-    // First pull from the deployment registry, then resume standard publish flow.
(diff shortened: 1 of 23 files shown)


core/src/plugins/kubernetes/container/publish.ts
```diff
@@ -8,41 +8,37 @@
...
-    // First pull from the deployment registry, then resume standard publish flow.
-    // This does mean we require a local docker as a go-between, but the upside is that we can rely on the user's
-    // standard authentication setup, instead of having to re-implement or account for all the different ways the
-    // user might be authenticating with their registries.
-    // We also generally prefer this because the remote cluster very likely doesn't (and shouldn't) have
-    // privileges to push to production registries.
-    log.info(`Pulling from deployment registry...`)
-    await pullBuild({ ctx: k8sCtx, action, log, localId: localImageId, remoteId: deploymentRegistryImageId })
-  }
+  const cloudBuilderUsed = await cloudBuilder.getAvailability(k8sCtx, action)
```
stefreak, line 21: Same here, I think `cloudBuilder.isConfigured()` would be better here.

docs/k8s-plugins/guides/in-cluster-building.md
```diff
@@ -694,6 +694,21 @@ providers:
...
+```yaml
+# garden.yml
+kind: Build
+type: container
+name: my-container
+spec:
+  platforms: ["linux/amd64", "linux/arm64"]
+```
+
+Multi-platform builds are available for both `cluster-buildkit` and `kaniko`.
```
stefreak, line 710: Really, how so? Maybe 
```suggestion
Multi-platform builds are **not** available for both `cluster-buildkit` and `kaniko`.
```

docs/other-plugins/container.md
```diff
@@ -90,6 +90,39 @@ spec:
...
+# garden.yml
+kind: Build
+type: container
+name: my-container
+spec:
+  platforms: ["linux/amd64", "linux/arm64"]
+```
+
+Garden interacts with several local and remote builders. Currently support for multi-platform builds varies based on the builder backend.
+The following build backends support multi-platform builds out of the box: `garden-cloudbuilder`, `cluster-buildkit`, `kaniko`.
```
stefreak, line 107: As `garden-cloudbuiler` is not an option in the Kubernetes config:
```suggestion
The following build backends support multi-platform builds out of the box: [Garden Cloud Builder](../reference/providers/container.md), `cluster-buildkit`, `kaniko`.
```

stefreak (commented):
Looks really good! I posted a few small comments 👍

stefreak (dismissed):
Nice work! :) ⚡ Looks good to me ✨

vvagaytsev (approved):

PR #6211, opened 2024-06-24
feat: Windows support for cloud builder

stefreak:
Establish buildx remote builders via MTLS to also support Windows.

Note: At the moment this implementation is mysteriously 5-20 seconds slower per build than the nsc-proxy implementation. There seems to be an additional cost of establishing the connection.
The `referenceCounter` can be used to cache the certificates in Cloud, but to make use of this the cache in the backend still needs to be implemented.

### DIFF
--- a/core/src/plugins/container/cloudbuilder.ts
+++ b/core/src/plugins/container/cloudbuilder.ts
@@ -8,24 +8,38 @@
 import type { PluginContext } from "../../plugin-context.js"
 import type { Resolved } from "../../actions/types.js"
 import type { ContainerBuildAction } from "./config.js"
-import { ChildProcessError, ConfigurationError, InternalError } from "../../exceptions.js"
-import type { ContainerProvider } from "./container.js"
+import { ConfigurationError, InternalError } from "../../exceptions.js"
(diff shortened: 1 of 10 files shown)


core/src/plugins/container/cloudbuilder.ts
```diff
@@ -294,3 +256,237 @@ function getConfiguration(ctx: PluginContext): CloudBuilderConfiguration {
...
+    for (const [i, { builderUrl, platform }] of this.endpoints.entries()) {
+      const result = await containerHelpers.dockerCli({
+        cwd: this.certDir,
+        args: [
+          "buildx",
+          "create",
+          "--name",
+          this.name,
+          "--node",
+          platform.replace("/", "-"),
```
stefreak, line 476:
```suggestion
          platform.replaceAll("/", "-"),
```

twelvemo (dismissed):
Looks great! Added the `platforms` and the `actionVersion` to the build request, so also waiting for another reviewer.

10ko (dismissed):
LGTM

vvagaytsev (approved):

PR #6227, opened 2024-06-25
improvement(sync): use native Mutagen daemon by default

vvagaytsev:

This PR flips the default value of `GARDEN_ENABLE_NEW_SYNC` feature flag to `true`.
It means that the sync mode will use the [native Mutagen](https://github.com/mutagen-io/mutagen) instead of the [Garden-fork](https://github.com/garden-io/mutagen) as a synchronization machinery.

Fixes #


twelvemo (commented):
Some test failures but otherwise looks good!

twelvemo (approved):

PR #6228, opened 2024-06-25
fix(template): allow partially resolved vars in arithmetic expressions

### LINKED ISSUE #6201: 0.13: [Bug]: Template: Arithmetic binary operators do not work together with variable lookup
### Current behavior
`garden run my-test` prints this error:

```console
Invalid template string (${var.prefix.suffix * 4}) at path spec.command.1: Both terms need to be numbers for * operator (got string and number).
```

### Expected behavior
`garden run my-test -l3` should print this stdout from the run, the multiplication should work:
```
8
```

### Reproducible example

The issue happens with all arithmetic operators, i.e. `+`, `-`, `*`, `/`, and others (modulo, comparison, etc.).

```
apiVersion: garden.io/v1
kind: Project
name: project
defaultEnvironment: local
dotIgnoreFile: .gitignore
environments:
  - name: local
---
kind: Run
type: exec
name: my-test
spec:
  command: [echo, '"${var.prefix.suffix * 4}"']
variables:
  prefix:
    suffix: 2
```

### Workaround
Skipping the variable lookup makes the multuplication work, but that's not really a workaround in most cases

```
  command: [echo, '${2 * 4}']
```

### Additional context
It is not the problem that the YAML value is being read as a string. If you use jsonDecode it also happens:

```
apiVersion: garden.io/v1
kind: Project
name: project
defaultEnvironment: local
dotIgnoreFile: .gitignore
environments:
  - name: local
---
kind: Run
type: exec
name: my-test
spec:
  command: [echo, '${jsonDecode(var.prefix.suffix) * 4}']
variables:
  prefix:
    suffix: 2

```

### Suggested solution(s)
N/A

### Your environment

`garden version: 0.13.31`

vvagaytsev:

This PR allows using partially resolved variables in arithmetic operations.

Fixes #6201

If we omit the explicit double-quotes in the `echo` command args in the example in #6201, i.e. replace
```yaml
spec:
  command: [echo, '"${var.prefix.suffix * 4}"']
```

with
```yaml
spec:
  command: [echo, '${var.prefix.suffix * 4}']
```
we'll get a schema validation error:
```console
Error validating spec for Run type=exec name=my-test:

At path spec.command.1: Expected string, received number
```

That is not related to the template string resolution and should be fixed in a separate PR. A possible solution is to stringify all numbers in `spec.command`'s args implicitly.

### DIFF
--- a/core/src/template-string/template-string.ts
+++ b/core/src/template-string/template-string.ts
@@ -69,7 +69,16 @@ interface ConditionalTree {
 }
 
 function getValue(v: Primitive | undefined | ResolvedClause) {
-  return isPlainObject(v) ? (<ResolvedClause>v).resolved : v
+  return isPlainObject(v) ? (v as ResolvedClause).resolved : v
+}
+
+function isPartiallyResolved(v: Primitive | undefined | ResolvedClause): boolean {
+  if (!isPlainObject(v)) {
+    return false
+  }
+
+  const clause = v as ResolvedClause
+  return !!clause.partial
 }
 
 export class TemplateError extends GardenError {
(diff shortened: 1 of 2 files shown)


vvagaytsev:
@twelvemo @10ko it's still a draft, sorry for the early review request.

vvagaytsev:
@twelvemo @10ko it's ready for review now :)

10ko (approved):
Looks good!

PR #6278, opened 2024-07-10
feat(kubernetes): plugin command to remove `garden-util` resources

### LINKED ISSUE #5909: [FEATURE]: `garden-util` deployment does not get removed
### Background / Motivation
We are using Garden in CI/CD pipelines, which performs various deploys over a lot of namespaces. This deploys a ton of `garden-util` deployments in every namespace. Even after all Garden actions this deployment is not needed and could be removed.

Have an option to remove the `garden-util` deployment.

At the end of the day our development clusters are scaled down, but sometimes it is not possible as `garden-util` is still present and costs us money, while this is not needed anymore. 

It would make our life a lot easier to have a way of deleting the deployments at the very end of garden. Now we have to take some extra steps in our pipelines to get rid of the deployment multiple times.

Set a flag in the Garden project (or in another place) to remove the `garden-util` deployment after all actions are completed.

🌵 Not having this feature makes using Garden ~painful~ a bit more expensive.

vvagaytsev:

A kubernetes plugin command to manually remove the unnecessary `garden-util` resources from the cluster.

Fixes #5909

### DIFF
--- a/core/src/plugins/kubernetes/commands/cleanup-garden-util.ts
+++ b/core/src/plugins/kubernetes/commands/cleanup-garden-util.ts
@@ -24,11 +24,10 @@ export const cleanupUtilDeployment: PluginCommand = {
     const k8sCtx = ctx as KubernetesPluginContext
     const provider = k8sCtx.provider
 
-    log.info({ msg: styles.highlight(`\nRemoving ${utilDeploymentName} deployment`) })
-
     const namespace = provider.outputs["app-namespace"]
-    const targetKinds = ["Service", "Deployment"]
+    log.info({ msg: styles.highlight(`\nRemoving ${utilDeploymentName} deployment from namespace ${namespace}`) })
 
+    const targetKinds = ["Service", "Deployment"]
     const resources: KubernetesResource[] = targetKinds.map((kind) => {
       return { apiVersion: "v1", kind, metadata: { name: utilDeploymentName } }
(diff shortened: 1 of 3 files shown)


core/src/plugins/kubernetes/commands/cleanup-garden-util.ts
```diff
@@ -0,0 +1,45 @@
...
+  description: `Remove ${utilDeploymentName} utility deployment from the current namespace`,
+  title: `Cleanup ${utilDeploymentName} deployment`,
+  resolveGraph: true,
+
+  handler: async ({ ctx, log }) => {
+    const result = {}
+    const k8sCtx = ctx as KubernetesPluginContext
+    const provider = k8sCtx.provider
+
+    log.info({ msg: styles.highlight(`\nRemoving ${utilDeploymentName} deployment`) })
```
thsig, line 27: Would be nice to include the namespace name in the message here.

thsig (approved):
Looks good! Had one minor comment there, but I'll leave it up to you.

core/src/plugins/kubernetes/commands/cleanup-garden-util.ts
vvagaytsev: Thanks! I fixed that.

thsig (approved):

PR #6293, opened 2024-07-16
fix: resolve `disabled` action flag before duplicate action names validation

vvagaytsev:

This PR patches the feature implemented in #4805 and #5686.

Without this fix, the disabled flag on actions is not resolved properly if it's defined as a template string.


stefreak (approved):
Nice work!

There are some problems with this, e.g. the mode has the wrong value.

Nonetheless, this is definitely an improvement so I'm ok to merge this and we can improve the other things later.

PR #6295, opened 2024-07-17
fix(template): use stricter context for `disabled` flag resolution

vvagaytsev:

Do not allow `this.*` context for `disabled` field, just to avoid unnecessary complexity in passing action mode to the `scanAndAddConfigs()`.

Now an error is thrown if  `disabled:` flag has a reference to `${this.mode}`.

Fixes the incorrectly defined action mode introduced in #6293.

We can support `${this.*}` context as a separate feature PR if necessary.


stefreak (approved):
👍  Nice, that is safer!

PR #6300, opened 2024-07-18
chore: remove `buildx` from the standard docker build args

vvagaytsev:

Fixes #

### DIFF
--- a/core/src/plugins/container/build.ts
+++ b/core/src/plugins/container/build.ts
@@ -169,7 +169,7 @@ async function buildContainerLocally({
     }
   }
 
-  const cmdOpts = ["buildx", "build", ...dockerFlags, "--file", dockerfilePath]
+  const cmdOpts = ["build", ...dockerFlags, "--file", dockerfilePath]
   try {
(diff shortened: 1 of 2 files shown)


core/src/plugins/container/build.ts
```diff
@@ -169,7 +169,7 @@ async function buildContainerLocally({
     }
   }
 
-  const cmdOpts = ["buildx", "build", ...dockerFlags, "--file", dockerfilePath]
+  const cmdOpts = ["build", ...dockerFlags, "--file", dockerfilePath]
```
stefreak, line 172: The buildx change shouldn't be a breaking change; It does require the buildx plugin to be installed, but I would consider a docker installation where the buildx plugin is missing basically broken.

Also, the fact that we ship the Docker client with Garden and it relies on docker plugins installed in the system (in `~/.docker`) is something we should consider a Garden bug if users run into this.

vvagaytsev: Ok, thanks for the explanation!

PR #6294, opened 2024-07-16
feat(container): first-class BuildKit secrets support

stefreak:
First-class BuildKit secrets support for BuildKit in-cluster building,
Garden Cloud Builder and building locally.

Kaniko is not supported due to lack of support (See also
https://github.com/GoogleContainerTools/kaniko/issues/3028)

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

PR #6301, opened 2024-07-18
fix: retain original process env when call `spawn` helper

vvagaytsev:

Prior to #6294 we have never used the `env` parameter of the function `containerHelpers.dockerCli(...)`.

This PR ensures that the original process environment is always preserved and pass to the Docker process.
Without this fix, any custom `env` would replace the original process environment variables.
Instead of composing the merged env vars on the caller side, it's now done inside the helper function.

Fixes #

### DIFF
--- a/core/src/plugins/container/helpers.ts
+++ b/core/src/plugins/container/helpers.ts
@@ -33,7 +33,7 @@ import type { Resolved } from "../../actions/types.js"
 import pMemoize from "../../lib/p-memoize.js"
 import { styles } from "../../logger/styles.js"
 import type { ContainerProviderConfig } from "./container.js"
-import { type MaybeSecret } from "../../util/secrets.js"
+import type { MaybeSecret } from "../../util/secrets.js"
 
 const { readFile, pathExists, lstat } = fsExtra
 
@@ -393,7 +393,7 @@ const helpers = {
     stdout?: Writable
     stderr?: Writable
     timeout?: number
-    env?: { [key: string]: MaybeSecret }
+    env?: Record<string, MaybeSecret>
   }) {
     const docker = ctx.tools["container.docker"]
 
--- a/core/src/plugins/container/helpers.ts
+++ b/core/src/plugins/container/helpers.ts
@@ -393,7 +393,7 @@ const helpers = {
     stdout?: Writable
     stderr?: Writable
     timeout?: number
-    env?: Record<string, MaybeSecret>
+    env?: Record<string, MaybeSecret | undefined>
   }) {
     const docker = ctx.tools["container.docker"]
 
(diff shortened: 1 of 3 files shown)


core/src/util/secrets.ts
```diff
@@ -46,6 +46,7 @@ export interface Secret {
 }
 
 export type MaybeSecret = string | Secret
+export type MaybeSecretEnv = { [key: string]: MaybeSecret }
```
stefreak, line 49: I'm not sure that this type (with this name) is necessary; 
First of all, a key-value map whose values might be secret can be used for more than just env variables.
Second of all, writing `Record<string, MaybeSecret>` is only slightly more verbose but has the same effect and when you see it, you know exactly what it means.

core/src/plugins/container/helpers.ts
```diff
@@ -610,3 +610,11 @@ function fixDockerVersionString(v: string) {
 function getDockerfilePath(basePath: string, dockerfile = defaultDockerfileName) {
   return joinWithPosix(basePath, dockerfile)
 }
+
+function getProcessEnv(): MaybeSecretEnv {
+  return Object.fromEntries(
+    Object.entries(process.env)
```
stefreak, line 616: Values in `process.env` can never be a `Secret` – they're always `undefined` or `string`.

Of course, users might want to mark inputs from environment variables as secret, but that would need to be a separate feature.

Is the purpose of this function to fix a type error? If yes, then there is another root cause that needs to be addressed.

core/src/plugins/container/helpers.ts
```diff
@@ -393,15 +393,15 @@ const helpers = {
...
+    env?: MaybeSecretEnv
   }) {
     const docker = ctx.tools["container.docker"]
 
     try {
       const res = await docker.spawnAndWait({
         args,
         cwd,
-        env,
+        env: env === undefined ? undefined : { ...getProcessEnv(), ...env },
```
stefreak, line 404: Can we also merge the process env later, right before we call the underlying library (e.g. `crossSpawn`)? I feel like our exec/spawn APIs should be consistent and always merge the process env. To delete certain vars you'd need to pass `undefined` or `""`.

This would help make it easier to reason about code changes, and also help when we switch underlying libraries which might have different semantics of their `env` parameters.

stefreak (changes requested):
Awesome, good work, this is a very important fix and a very good catch!
I had a couple of comments; I'm open to pair with you on them 👍

core/src/util/secrets.ts
stefreak: I'd be also fine with renaming the type:
```suggestion
export type MaybeSecretRecord = { [key: string]: MaybeSecret }
```

stefreak (approved):
Awesome! Looks real good now 👍

PR #6319, opened 2024-07-25
fix(exec): fix error handling in exec run and test actions

### LINKED ISSUE #6261: Crash: Garden crashes on I/O errors while artifacts copying
## Crash report

When `garden test` is run and the test command fails, Garden 0.13 crashes instead of exiting with a runtime error.
So, the error output looks too verbose and confusing. See the details below.

#### Garden 0.13

```console

You can help by reporting this on GitHub: https://github.com/garden-io/garden/issues/new?labels=bug,crash&template=CRASH.md&title=Crash%3A%20Cannot%20copy%20%60.%2Fsrc-dir%60%3A%20the%20file%20doesn%27t%20exist

Failed processing Test type=exec name=test (from module test-module) (took 3.04 sec). This is what happened:
CpyError: Cannot copy `./src-dir`: the file doesn't exist
    at file:///~/io.garden.garden/1721670966-BlpFWH3.r/rollup/index-_bbuTLSX.mjs:5874:11
    at cpy (file:///~/io.garden.garden/1721670966-BlpFWH3.r/rollup/index-_bbuTLSX.mjs:5954:4)
    at file:///~/io.garden.garden/1721670966-BlpFWH3.r/rollup/common-Co5xZozR.mjs:67:15
    at async Promise.all (index 0)
    at async Object.<anonymous> (file:///~/io.garden.garden/1721670966-BlpFWH3.r/rollup/test-BNy_r1mw.mjs:27:5)
    at async Object.assign.handlerType.handlerType (file:///~/io.garden.garden/1721670966-BlpFWH3.r/rollup/garden.mjs:795076:28)
    at async Router.callHandler (file:///~/io.garden.garden/1721670966-BlpFWH3.r/rollup/garden.mjs:795047:24)
    at async run (file:///~/io.garden.garden/1721670966-BlpFWH3.r/rollup/garden.mjs:796054:28)
    at async TestTask.process (file:///~/io.garden.garden/1721670966-BlpFWH3.r/rollup/garden.mjs:375478:28)
    at async descriptor.value (file:///~/io.garden.garden/1721670966-BlpFWH3.r/rollup/garden.mjs:370894:29)
```

#### Garden 0.12

In Garden 0.12 it looks much better and concise:

```console
Failed running test tests in module test-module. Here is the output:
Cannot copy `src-dir`: the file doesn't exist
1 test action(s) failed!
```
### What did you do?

Run `garden test` with the following config:

* `garden.yml`

<details>

<summary>
Click here to expand `garden.yml`
</summary>

```yml
apiVersion: garden.io/v0
kind: Project
name: repro-6261

environments:
  - name: local

providers:
  - name: exec
    environments: [local]

---
kind: Module
type: exec
name: test-module
local: true
tests:
  - name: test
    command: [./test-cmd.sh]
    artifacts:
      - source: src-dir
        target: dst-dir
```

</details>

* `test-cmd.sh` - a simple sh script that prints some outputs and fails

<details>

<summary>
Click here to expand `test-cmd.sh`
</summary>

```bash
echo "Executing..."
exit 1
```

</details>

### Your environment

* **OS:** 
* **Garden version: `0.13.32` 

### Frequency

Permanently.

### Workaround

None.

### Additional context

The same issue happens when `garden run test` command is used.

Comments:

* The output in `0.13` is confusing because this is not a bug in Garden
* The output in `0.12` is much cleaner
* In practice, the test failed and artifacts couldn't be copied, but without the output of the tests, the current output is confusing
* In `0.12`, it was possible to add `--logger-type=basic` in the CI and see the output of the test command, but in `0.13` this is not possible (even if it's supposed to be a default option now - see [docs](https://docs.garden.io/reference/commands#global-options))

vvagaytsev:

This PR fixes the command result processing and error handling in `Run` and `Test` actions of `exec` type, and align the error handling behaviour between the commands.:

* an error is returned immediately if the command exited with non-zero code, so Garden does not attempt to copy missing or incomplete artifacts; the error message includes the command output and the exit code.
* the command execution output is always printed in the console if the command was successful; in case of error, it will be printed as a part of the error message, and won't be printed twice.
* the artifacts are copied only if the command execution as successful
* I/O errors while artifacts copying do not crash the Garden process, and are handled as the other runtime errors

Fixes #6261

The changes have been tested manually. We do not have integration tests for exec provider yet, so let's skip it. We do not change the affected part of the code often.

There is another function [`copyArtifacts` in `util/artifacts.ts`](https://github.com/garden-io/garden/blob/main/core/src/util/artifacts.ts#L63) which is used on the action router-level in:
* [Run action router](https://github.com/garden-io/garden/blob/main/core/src/router/run.ts#L65)
* [Test action router](https://github.com/garden-io/garden/blob/main/core/src/router/test.ts#L65)

Both places seem to be irrelevant to the issue in question. There is a kind of clean-up in the finally block where the produced artifacts are copied form the temp directory before cleaning it up.

I'm not sure if we need to do that at all, but let's leave the code as-is, and change that part in a separate PR if necessary.

### DIFF
--- a/core/src/plugins/exec/common.ts
+++ b/core/src/plugins/exec/common.ts
@@ -16,6 +16,7 @@ import { exec } from "../../util/util.js"
 import type { Log } from "../../logger/log-entry.js"
 import type { PluginContext } from "../../plugin-context.js"
 import type { ResolvedExecAction } from "./config.js"
+import { RuntimeError } from "../../exceptions.js"
 
 export function getDefaultEnvVars(action: ResolvedExecAction) {
   return {
@@ -99,12 +100,22 @@ export async function copyArtifacts(
 ) {
   return Promise.all(
     (artifacts || []).map(async (spec) => {
-      log.verbose(`→ Copying artifacts ${spec.source}`)
+      try {
+        log.verbose(`→ Copying artifacts ${spec.source}`)
 
-      // Note: lazy-loading for startup performance
-      const { default: cpy } = await import("cpy")
+        // Note: lazy-loading for startup performance
+        const { default: cpy } = await import("cpy")
 
-      await cpy(`./${spec.source}`, join(artifactsPath, spec.target || "."), { cwd: from })
+        await cpy(`./${spec.source}`, join(artifactsPath, spec.target || "."), { cwd: from })
+      } catch (err: unknown) {
+        if (!(err instanceof Error)) {
+          throw err
+        }
+
+        if (err.name === "CpyError") {
+          throw new RuntimeError({ message: err.message })
+        }
+      }
     })
   )
 }
--- a/core/src/plugins/exec/common.ts
+++ b/core/src/plugins/exec/common.ts
@@ -16,7 +16,7 @@ import { exec } from "../../util/util.js"
 import type { Log } from "../../logger/log-entry.js"
 import type { PluginContext } from "../../plugin-context.js"
 import type { ResolvedExecAction } from "./config.js"
-import { RuntimeError, toGardenError } from "../../exceptions.js"
+import { RuntimeError } from "../../exceptions.js"
 
 export function getDefaultEnvVars(action: ResolvedExecAction) {
   return {
--- a/core/src/plugins/exec/common.ts
+++ b/core/src/plugins/exec/common.ts
@@ -16,7 +16,7 @@ import { exec } from "../../util/util.js"
 import type { Log } from "../../logger/log-entry.js"
 import type { PluginContext } from "../../plugin-context.js"
 import type { ResolvedExecAction } from "./config.js"
-import { RuntimeError } from "../../exceptions.js"
+import { RuntimeError, toGardenError } from "../../exceptions.js"
 
 export function getDefaultEnvVars(action: ResolvedExecAction) {
   return {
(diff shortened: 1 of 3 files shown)


core/src/plugins/exec/common.ts
```diff
@@ -99,12 +100,22 @@ export async function copyArtifacts(
...
-      await cpy(`./${spec.source}`, join(artifactsPath, spec.target || "."), { cwd: from })
+        await cpy(`./${spec.source}`, join(artifactsPath, spec.target || "."), { cwd: from })
+      } catch (err: unknown) {
+        if (!(err instanceof Error)) {
+          throw err
+        }
+
+        if (err.name === "CpyError") {
+          throw new RuntimeError({ message: err.message })
+        }
```
thsig, line 117: Here, we're not throwing `err` unless `err.name === "CpyError`. Is this done on purpose? We should at least log something in that case, right?

thsig (changes requested):
Had one comment there, looks good otherwise!

core/src/plugins/exec/common.ts
vvagaytsev: No, that was not done on purpose. That kind of error should be re-thrown.
Good catch, thank you! Fixed that in 165e487f00f2b0f7ab774d1f29f80470b937c684.

thsig (approved):

thsig (approved):

PR #6334, opened 2024-07-30
chore: update bundled NodeJS to v22.6.0

twelvemo:

Fixes #


stefreak:
Converted the issue to a draft, as we also need to change the NodeJS version in the CI etc. before merging this.

vvagaytsev:
Looks good! Should we also update the scripts to setup a local dev environment?

* [`.nvmrc`](https://github.com/garden-io/garden/blob/main/.nvmrc)
* [`.tool-versions`](https://github.com/garden-io/garden/blob/main/.tool-versions)

Btw, `.tool-versions` has older values for `kubectl` and `helm` CLI tools, should we update those as well?
Afaik, `.rool-versions` is only used with [`install-asdf-dependencies.sh`](https://github.com/garden-io/garden/blob/main/scripts/install-asdf-dependencies.sh). So it's not used that often, but we should keep it up-to-date. This can be done in a separate PR.

stefreak:
> Btw, `.tool-versions` has older values for `kubectl` and `helm` CLI tools, should we update those as well? Afaik, `.rool-versions` is only used with [`install-asdf-dependencies.sh`](https://github.com/garden-io/garden/blob/main/scripts/install-asdf-dependencies.sh). So it's not used that often, but we should keep it up-to-date. This can be done in a separate PR.

I think those aren't really requirements and should be deleted, unless they are used in tests for some reasons?

vvagaytsev (approved):

PR #6335, opened 2024-07-30
docs(troubleshooting): add entry on using int values in K8s manifests

eysi09:
This is based on an internal learning. Not sure if Garden can be smarter her, so first step is at least to document the issue.

Fixes #


twelvemo (approved):
👍

Issue #4668, opened 2023-06-20
Crash: Custom commands cause crash in `garden dev` shell: Missing cli argument in custom command wrapper.


shumailxyz:
### Current Behavior
Custom commands that use `gardenCommand` do not work in dev mode and throw an error:

```
Missing cli argument in custom command wrapper.
This is a bug. Please report it!
```

![Screenshot 2023-06-20 at 12 24 41](https://github.com/garden-io/garden/assets/5106058/77e157e6-3843-491c-95a4-1c512d7bf372)

### Expected behavior

Custom garden commands should work just fine in dev mode.

### Workaround

Use following `commands.garden.yml` and try to run `c` in dev mode.

```
kind: Command
name: echo
description:
  short: Just echo a string
exec:
  command:
    - sh
    - -c
    - echo "hello world"

---

kind: Command
name: c
description:
  short: Run validate cmd
gardenCommand:
  - validate
 
```

`garden version`: 0.13.2

shumailxyz:
Draft PR for this: https://github.com/garden-io/garden/pull/4679

However, that doesn't address the issue properly and breaks the rendering of logs in dev REPL. The custom command wrapper performs a full run of garden command via GardenCli.run() and that initializes the logger multiple times. This needs to be addressed more carefully.

Until this is fixed, the workaround is to use custom garden command outside dev mode.

shumailxyz:
Might be related to https://github.com/garden-io/garden/issues/4116

stefreak:
Has been fixed in [#6341](https://github.com/garden-io/garden/pull/6341)

PR #6341, opened 2024-08-02
fix(dev): fix custom commands not working in dev console

thsig:

Before this fix, running a custom command within the `dev` command's interactive command line would fail.

Fixes #4688.


vvagaytsev:
@thsig does this fix #4668? I remmeber we attempted to fi that issue in #4679.

thsig:
Ah yes, it should fix that one as well!

vvagaytsev (approved):

PR #6357, opened 2024-08-06
fix: better stream cleanup when child procs exit

thsig:

This fix applies to a couple of helpers that spawn child processes and stream logs from them.

Some users have experienced unhandled stream-related Node.js errors when processing large graphs, and the lack of proper cleanup handling when the stream closes or the process exits might be the underlying cause.

### DIFF
--- a/core/src/util/process.ts
+++ b/core/src/util/process.ts
@@ -25,12 +25,28 @@ export function streamLogs({
 }): void {
   const logStream = split2()
 
+  let streamClosed = false
+
+  const handleStreamClosed = () => {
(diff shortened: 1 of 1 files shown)


stefreak (approved):

stefreak:
This was not causing the problem; Do we still deem the change necessary?

thsig:
I'm not quite sure it's necessary—I think there are lots of other places where we don't do this sort of cleanup, but haven't run into this error.

So I could go either way on this.

stefreak:
Closing in favor of #6377

PR #6408, opened 2024-08-28
fix(template): fix template string escaping and resolution in Module configs

### LINKED ISSUE #3989: Escaping template strings not working

vvagaytsev:

* Skip pre-mature string unescaping in Module configs
* Fix too greedy escaped string pattern definition in PeggyJS

Fixes #3989
Patches #5680

### DIFF
--- a/core/src/resolve-module.ts
+++ b/core/src/resolve-module.ts
@@ -932,7 +932,14 @@ export class ModuleResolver {
     templateContextParams: ModuleConfigContextParams
   ): Promise<DeepPrimitiveMap> {
     const moduleConfigContext = new ModuleConfigContext(templateContextParams)
-    const resolveOpts = { allowPartial: false }
+    const resolveOpts = {
+      allowPartial: false,
+      // Modules will be converted to actions later, and the actions will be properly unescaped.
+      // We avoid premature un-escaping here,
+      // because otherwise it will strip the escaped value in the module config
+      // to the normal template string in the converted action config.
+      unescape: false,
+    }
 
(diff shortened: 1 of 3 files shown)


twelvemo (approved):
Good catch!

vvagaytsev:
@omry-arpaly huge thanks for you contributions to this fix! 💯

PR #6430, opened 2024-09-10
fix: allow relative symlinks to directories when using build staging

### LINKED ISSUE #2382: "Attempted to copy non-file" when a symlink target is a folder

stefreak:

Allows using relative symlinks to directories and files when using build staging.

The goal is to reproduce symlinks in the target directory as-is, without having to resolve them.
The user would be responsible for making sure that both the relative symlink as well as the target
directory are included in the action.

If the symlink points to a file or directory outside the actions's source path, this throws an
error.

To avoid this becoming a breaking change, we keep the old behaviour if the symlink points to a
file. In that case, we resolve the symlink and copy the target file. If the target file is outside
 the action's source or build directory, log a warning.

In a future 0.14 release of Garden, we should remove that branch, only reproduce symlinks (without
resolving them recursively) and
throw an error if the link is out of bounds.

Fixes #2382

### DIFF
--- a/core/src/build-staging/build-staging.ts
+++ b/core/src/build-staging/build-staging.ts
@@ -189,7 +189,7 @@ export class BuildStaging {
 
     // Source root must exist and be a directory
     let sourceStat = await statsHelper.extendedStat({ path: sourceRoot })
-    if (!sourceStat || !(sourceStat.isDirectory() || sourceStat.target?.isDirectory())) {
+    if (!sourceStat || !sourceStat.isDirectory()) {
       throw new InternalError({
         message: `Build staging: Source root ${sourceRoot} must exist and be a directory`,
       })
@@ -271,6 +271,8 @@ export class BuildStaging {
       const to = targetShouldBeDirectory || targetStat?.isDirectory() ? join(targetPath, sourceBasename) : targetPath
 
       await syncFileAsync({
+        log,
+        root: sourceRoot,
         from: sourceRoot,
         to,
         allowDelete: withDelete,
@@ -323,7 +325,7 @@ export class BuildStaging {
               ([fromRelative, toRelative], fileCb) => {
                 const from = joinWithPosix(sourceRoot, fromRelative)
                 const to = joinWithPosix(targetPath, toRelative)
-                cloneFile({ from, to, allowDelete: withDelete, statsHelper }, fileCb)
+                cloneFile({ log, root: sourceRoot, from, to, allowDelete: withDelete, statsHelper }, fileCb)
               },
               cb
             )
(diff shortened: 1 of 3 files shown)


core/src/build-staging/helpers.ts
```diff
@@ -6,22 +6,27 @@
...
 import { promisify } from "util"
+import { styles } from "../logger/styles.js"
+import { emitNonRepeatableWarning } from "../warnings.js"
+import { type Log } from "../logger/log-entry.js"
 
 export type MappedPaths = [string, string][]
 
 export interface CloneFileParams {
+  log: Log
+  root: string
```
vvagaytsev, line 29: Minor nit. Should we call this `sourceRoot`? Root sounds a bit broad.

vvagaytsev (dismissed):
Looks good, thank you! 👍 
Left a couple of non-blocking comments.

thsig (dismissed):

vvagaytsev:
@stefreak I've addressed the suggested changes in the recent refactoring commits.

Can you please take a look and review those changes? Just to be sure I haven't mixed up anything when I transformed positional arguments to a parameter object in the `ResolveSymlinkCallback`.

core/src/build-staging/build-staging.ts
```diff
@@ -189,7 +189,7 @@ export class BuildStaging {
 
     // Source root must exist and be a directory
     let sourceStat = await statsHelper.extendedStat({ path: sourceRoot })
-    if (!sourceStat || !(sourceStat.isDirectory() || sourceStat.target?.isDirectory())) {
+    if (!sourceStat || !sourceStat.isDirectory()) {
```
stefreak, line 192: I think this one must have been an accident
```suggestion
    if (!sourceStat || !(sourceStat.isDirectory() || sourceStat.target?.isDirectory())) {
```

stefreak (commented):
I can't approve your changes @vvagaytsev but it looks ok to me! The callback-style code is in dire need of refactoring to promises & async/await but for now this seems good enough to me.

vvagaytsev (approved):

vvagaytsev:
Thanks @stefreak! I've put a formal approval on it :)

PR #6433, opened 2024-09-12
fix: reproduce symlinks in build staging correctly on windows

stefreak:
Unfortunately I mixed those up due to the confusing data structures in https://github.com/garden-io/garden/pull/6430


core/src/build-staging/helpers.ts
```diff
@@ -183,7 +183,7 @@ function doClone(params: CopyParams) {
           to,
           // relevant on windows
           // nodejs will auto-detect this on windows, but if the symlink is copied before the target then the auto-detection will get it wrong.
-          targetStats?.isDirectory() ? "dir" : "file",
+          sourceStats.target?.isDirectory() ? "dir" : "file",
```
stefreak, line 186: probably would have been less confusing if these parameters were named `fromStats` and `toStats` or something like that

vvagaytsev: Feel free to rename params to avoid confusion.

vvagaytsev (changes requested):
There is a lint error. It complains about unused local variable `targetStats` from the params object defined in [L159](https://github.com/garden-io/garden/pull/6433/files#diff-35d7acac46ec199eb991f6b1296b6ae1d7a11ec53e57668afcb91a194a9cea00R159).

Do we need that variable here?

stefreak:
@vvagaytsev thanks, fixed the issue now, and refactored the code to make the variable names more clear.

stefreak:
@vvagaytsev the parameter has been added in #6430, but it was only due to the misleading name; It was actually correct to use `sourceStats.target` so the `sourceStats` parameter is indeed not necessary.

vvagaytsev (approved):
Thanks! 👍

PR #6447, opened 2024-09-17
fix: avoid throwing 401 when trying to login to cloud

10ko:
This PR fixes a very annoying issue that was causing the `login` command to fail when a stored authentication token was invalid or expired, forcing the user to logout and login back again in order to be able to use Garden.

Fixes #


10ko:
@twelvemo I had to add back a condition to account for the Auth Tokens and removed a test that's not valid anymore.
Could you give it another go? 🫶

twelvemo (approved):

PR #6448, opened 2024-09-17
fix(template): avoid premature `disabled` flag evaluation on actions

vvagaytsev:

* Do early `disabled` flag evaluation only for actions with duplicate names.
* Deny the usage of `var` and `variables` contexts in early `disabled` flag evaluation.

The contexts `var` and `variables` contexts are not fully resolved at that stage.
That can lead to incorrectly resolved `disabled` flag values.

Patches #4805, #5686, #6293, and #6406.

### DIFF
--- a/core/src/garden.ts
+++ b/core/src/garden.ts
@@ -1522,10 +1522,16 @@ export class Garden {
     }
 
     const context = new TemplatableConfigContext(this, config)
-    // Inject action's variables
-    context.variables = context.var = {
-      ...context.variables,
-      ...config.variables,
+    // Hack: deny variables contexts here, because those have not been fully resolved yet.
+    const deniedContexts = ["var", "variables"]
+    for (const deniedContext of deniedContexts) {
+      Object.defineProperty(context, "var", {
+        get: () => {
+          throw new ConfigurationError({
+            message: `If you have duplicate action names, the ${styles.accent("`disabled`")} flag cannot depend on the ${styles.accent(`\`${deniedContext}\``)} context.`,
+          })
+        },
+      })
     }
 
     return resolveTemplateString({
@@ -1543,10 +1549,11 @@ export class Garden {
     const key = actionReferenceToString(config)
     const existing = this.actionConfigs[config.kind][config.name]
 
-    // Resolve the actual values of the `disabled` flag
-    config.disabled = this.evaluateDisabledFlag(config)
-
     if (existing) {
+      // Resolve the actual values of the `disabled` flag
+      config.disabled = this.evaluateDisabledFlag(config)
+      existing.disabled = this.evaluateDisabledFlag(existing)
+
       if (actionIsDisabled(config, this.environmentName)) {
         this.log.silly(
           () => `Skipping action ${key} because it is disabled and another action with the same key exists`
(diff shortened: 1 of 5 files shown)


core/test/data/test-projects/disabled-action-with-var-context/runs.garden.yml
```diff
@@ -0,0 +1,19 @@
...
+---
+
+kind: Run
+name: run-script
+type: exec
+var:
+  foo: bar
+disabled: "${var.foo != 'bar'}"
+spec:
+  command: ["sh", "-c", "echo 'Hello from remote'"]
```
stefreak, line 19: Non-blocking nit, but let's add another comment with the `$merge` example that demonstrates why we can't allow accessing the var context.

stefreak (approved):
Nice, this is probably a workable compromise for now.

I'm still hoping we can either lift this limitation by refactoring how the action resolution works, or introduce a more consistent behaviour in a breaking change e.g. in 0.14

core/test/data/test-projects/disabled-action-with-var-context/runs.garden.yml
vvagaytsev: Added in 53002b4babaedc1d4946ca709667c921419a61ff.

stefreak (dismissed):
Nice! 👏

core/src/garden.ts
```diff
@@ -1522,10 +1522,16 @@ export class Garden {
...
 
     const context = new TemplatableConfigContext(this, config)
-    // Inject action's variables
-    context.variables = context.var = {
-      ...context.variables,
-      ...config.variables,
+    // Hack: deny variables contexts here, because those have not been fully resolved yet.
+    const deniedContexts = ["var", "variables"]
+    for (const deniedContext of deniedContexts) {
+      Object.defineProperty(context, "var", {
```
stefreak, line 1528: Oh, found a small mistake here, otherwise we forbid `var` only. Maybe should add another test case for that
```suggestion
      Object.defineProperty(context, deniedContext, {
```

stefreak (dismissed):
Awesome 👍

core/test/unit/src/garden.ts
```diff
@@ -3052,6 +3052,22 @@ describe("Garden", () => {
...
+      const garden = await makeTestGarden(getDataDir("test-projects", "disabled-action-with-var-context"))
+
+      // There are 2 'run-script' actions defined in the project, one per environment.
+      await expectError(() => garden.getConfigGraph({ log: garden.log, emit: false }), {
+        contains: [
+          "If you have duplicate action names",
+          "the",
+          "disabled",
+          "flag cannot depend on the",
+          "variables",
```
stefreak, line 3065: I think this test might fail now, because it will now log `var` context I believe
```suggestion
          "var",
```

core/test/data/test-projects/disabled-action-with-var-context/runs.garden.yml
```diff
@@ -0,0 +1,18 @@
...
+spec:
+  command: ["sh", "-c", "echo 'Hello from local'"]
+
+---
+
+kind: Run
+name: run-script
+type: exec
+var:
+  $merge: ${actions.run["run-script"].var}
```
stefreak, line 15: What happens if I use `actions` context directly in the `disabled` flag? Should we deny that as well?

vvagaytsev: It's already denied, the error will be thrown:
```
Invalid template string (...) at path disabled: Could not find key actions. Available keys: command, datetime, environment, git, inputs, local, parent, project, secrets, template, this, var and variables.
```

PR #6459, opened 2024-09-23
refactor: convert `Garden.treeCache` from class field to property

vvagaytsev:

To avoid keeping 2 different references to the same cache instance in `VcsHandler` and `Garden`.
Having `Garden.treeCache` as a property delegating to `VcsHandler.cache` makes it more clear.
Otherwise, it's not obvious that the cache instance in shared between 2 classes.

Fixes #

An alternative approach is to keep 2 independent cache instances for module versions and VCS files. Such attempt was taken in #6458.

That would not make the code simpler, because in some scenarios we need to `invalidateUp` or `invalidateDown` both caches synchronously. That would be error-prone, so let's keep the semantics as it is, and avoid more complex refactoring if that's not necessary.

### DIFF
--- a/core/src/garden.ts
+++ b/core/src/garden.ts
@@ -459,6 +459,10 @@ export class Garden {
     }
   }
 
+  /**
+   * We intentionally share the instance of the tree cache between Garden and VCS handler.
+   * Some code flows and legacy logic rely on that fact.
+   */
   get treeCache(): TreeCache {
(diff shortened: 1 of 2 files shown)


stefreak (dismissed):
I feel like it was clear before too when looking at the constructor. but this also does not hurt.

stefreak (approved):

PR #6465, opened 2024-09-23
fix(sync): use right arch of `mutagen-agent` binary in `k8s-sync` image

BraedonLeonard:

The `mutagen-agent` binary in arm64 builds of `k8s-sync` (and therefore also `k8s-util`) were the binaries for amd64 instead. This updates the Dockerfile for `k8s-sync` to copy the correct binary for `mutagen-agent`.


vvagaytsev:
@BraedonLeonard! Huge thanks for fixing this! 💯 Good catch! 👍 
We have overlooked that. We'll review that PR very soon and will publish new images and a Garden release.

vvagaytsev (approved):

PR #6478, opened 2024-09-24
chore(sync): use new `k8s-sync` and `k8s-util` images

vvagaytsev:

A follow-up PR for #6465.

Fixes #

The new images were already built and pushed.


stefreak (approved):
Nice!

PR #6483, opened 2024-09-25
perf(git): use `sub-tree` scan mode for config files scan

vvagaytsev:

This idea was discussed in #6461.

Use `sub-tree` scan mode for config files scanning. Now `Garden` has the second VCS handler - a dedicated `SubTreeGitHandler` that only performs config files scan. Both VCS handlers share the same `TreeCache`.

Fixes #


stefreak:
The big question for me here is: what are the benchmark results? is it faster, slower, same performance?

edvald:
To evaluate that we need to look at various cases, even different commands etc. I can't quite imagine how this would turn out to be slower outside of very rare cases and even then marginally, but I don't object to running a few performance tests to see how it shakes out.

vvagaytsev:
@stefreak sure, sorry, I accidentally created it before I completed the description.

With `sub-tree` scan mode it doesn't matter if we enable or disable the untracked files hashing implemented in #6461, it's equally fast in either way. The Garden repo scan example from #6461 (with `node_modules` included) gets scanned in about 0.6 sec.

With `repo` scan mode and disabled hashing for untracked files, the results are almost the same: about ~0.6 - 0.9 sec.

I think that's because repo does process more files than `sub-tree`. Processing without hash calculation is very fast, but processing less files is even faster.

We can keep both changes:
1. Keep a dedicated VcsHandler for config scan and use `sub-tree` - very fast.
2. Keep an option to disable hashing of untracked files - then we can also use `repo` mode for config scan if we have to do that for some reason.

vvagaytsev:
@edvald @stefreak thanks for the comments!

Sure, let's do more performance tests and benchmarks with different types of projects. I've converted this to draft.

edvald:
No need to overdo it @vvagaytsev :) Just good to see a couple of different variations.

stefreak:
I agree, if y'all think this is definitely an improvement I'm all for it, I just don't understand the vcs code to be honest so that's why I wanted to see proof that this is better than before :)

vvagaytsev:
Just tested this against main branch with 100k files project (used https://github.com/garden-io/garden-large-repo-generator to generate that). All files were tracked.

This PR did the initial scan in 0.2-0.3 sec against 0.5-0.6 sec in main. So, it looks it's ready to go.

PR #6505, opened 2024-10-03
fix(otel): ensure OTEL sends final span when -o opt is used

### LINKED ISSUE #6482: [Bug]: Opentelemetry missing span when using "-o yaml" option
### Current behavior

When opentelemtry is enabled, and using "-o yaml" on any command, the last span generated references an unexisting parent span.

image

On the image:
- Last span has parentId: '2469d70b49e96758'
- No span has been exported with this spanId

In fact, there is no root span at all in this trace - so I expect that the root span is not correctly sent.

### Expected behavior

All spans should be exported. 
E.g. when we do not add the "-o yaml" option:

image

Here we do have the root span exported (last one: parentId is undefined)

### Reproducible example

Any command that supports -o yaml:
- OTEL_TRACES_EXPORTER=console garden get version -o yaml
- OTEL_TRACES_EXPORTER=console garden get status -o yaml
- OTEL_TRACES_EXPORTER=console garden get test result mytest -o yaml

### Workaround

none

### Suggested solution(s)

I suspect some span not correctly finished - but I'm not an expert on opentracing api JS

### Additional context

- There was no trace parent defined in my environment variables (no OTEL_TRACE_PARENT or TRACEPARENT env variable set)
- On the example runs, I did not have the opentelemtry provider. I did try it before, and if I remember correctly it had same result

### Your environment

* **OS: Fedora Linux** 
* **How I'm running Kubernetes: N/A**

`garden version`

garden version: 0.13.38

eysi09:
Before this fix, when the -o/--output option was used the last OTEL span wouldn't be sent.

Fixing this surfaced some other issues and the code changes might look odd out of context so I'll try and explain the motivations below (also explained in the individual commit messages):

**First commit**

First off, the `Cli.run()` method would either return the results to the callee which would then do the clean up OR, if a certain combination of flags was passed, it would trigger a process exit itself (via the `shutdown` function) and the cleanup logic in the callee would never run. This was one of the reasons for the OTEL bug.

This meant that the behaviour of `Cli.run()` was pretty unpredictable and hard to track and the cleanup logic in the callee wasn't guaranteed to run. This was fixed in the first commit and now the `Cli.run()` method always returns the results in a consistent manner and cleanup is done by the callee. 

**Second commit**

After this change, it became obvious that the `exitOnAbort` flag didn't really do anything so it was removed in the second commit.

**Third commit**

Finally, fixing the OTEL bug with the `Cli.run()` changes exposed another bug where calling `otelSdk.shutdown()` would hang in those cases where an exporter hadn't been configured. This happens when the Garden class isn't initialised AND no OTEL exporter set via the `OTEL_*` env var. For example when running `garden help`. This was fixed by checking whether an exporter was set in the first place before calling `otelSdk.shutdown()`. It's a little ugly but the other alternatives I explored didn't feel any better, it's kind of a function of how we use OTEL. 

Fixes #6482 

### DIFF
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -86,9 +86,8 @@ if (gardenEnv.GARDEN_IGNORE_UNCAUGHT_EXCEPTION) {
 export async function runCli({
   args,
   cli,
-  exitOnError = true,
   initLogger = true,
-}: { args?: string[]; cli?: GardenCli; exitOnError?: boolean; initLogger?: boolean } = {}) {
+}: { args?: string[]; cli?: GardenCli; initLogger?: boolean } = {}) {
   let code = 0
   let result: RunOutput | undefined = undefined
 
@@ -105,7 +104,7 @@ export async function runCli({
         }
 
         // Note: We slice off the binary/script name from argv.
-        const results = await cli.run({ args: args || [], exitOnError })
+        const results = await cli.run({ args: args || [] })
 
         return results
       })
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -11,7 +11,7 @@ import type { RunOutput } from "@garden-io/core/build/src/cli/cli.js"
 import { GardenCli } from "@garden-io/core/build/src/cli/cli.js"
 import type { GardenPluginReference } from "@garden-io/core/build/src/plugin/plugin.js"
 import { GlobalConfigStore } from "@garden-io/core/build/src/config-store/global.js"
-import { getOtelSDK } from "@garden-io/core/build/src/util/open-telemetry/tracing.js"
+import { getOtelSDK, isOtelExporterConfigured } from "@garden-io/core/build/src/util/open-telemetry/tracing.js"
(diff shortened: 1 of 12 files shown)


examples/demo-project/garden.yml
```diff
@@ -29,3 +29,11 @@ providers:
...
 variables:
   userId: ${kebabCase(local.username)}
+
+---
+kind: Command
+name: hello
+description:
+  short: "say hi"
+exec:
+  command: [/bin/sh, -c, "echo 'hello Eysi'"]
```
vvagaytsev, line 39: Is this a left-over from local debugging?

vvagaytsev:
Great! 💯 
I left a minor comment, otherwise looks good! 👍

examples/demo-project/garden.yml
eysi09: Ah good catch. Thought I had removed it but maybe messed up a rebase. 

Fixed now.

vvagaytsev (approved):

vvagaytsev (approved):

PR #6526, opened 2024-10-08
fix(pulumi): prevent existing pulumi configs from being overwritten

BraedonLeonard:

This PR addresses an issue where existing pulumi config files fail to be read in, and are then overwritten by Garden, essentially deleting all existing config from the file.

In `applyConfig()`, there's this line here that overwrites the pulumi config that gets read in:

```python
stackConfig.config = vars
```

The docs for the pulumi provider and actions aren't clear about how existing pulumi config files are handled, so I'm not sure if this is intentional or not, but I'm pretty sure this is meant to merge these values into the existing pulumi config, rather than overwriting it. Since I wasn't certain about whether or not this was intentional, I left it alone and just called it out here.

### DIFF
--- a/plugins/pulumi/src/helpers.ts
+++ b/plugins/pulumi/src/helpers.ts
@@ -301,9 +301,6 @@ export async function applyConfig(params: PulumiParams & { previewDirPath?: stri
   }
   vars = <DeepPrimitiveMap>merge(vars, pulumiVars || {})
   log.debug(`merged vars: ${JSON.stringify(vars, null, 2)}`)
-  // TODO [!!!] Note the fact that this overwrites any existing pulumi config contents
-  //            instead of merging them in in the PR, since I'm not sure if that
-  //            behaviour is intentional or not.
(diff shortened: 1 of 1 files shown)


plugins/pulumi/src/helpers.ts
```diff
@@ -300,6 +301,9 @@ export async function applyConfig(params: PulumiParams & { previewDirPath?: stri
   }
   vars = <DeepPrimitiveMap>merge(vars, pulumiVars || {})
   log.debug(`merged vars: ${JSON.stringify(vars, null, 2)}`)
+  // TODO [!!!] Note the fact that this overwrites any existing pulumi config contents
+  //            instead of merging them in in the PR, since I'm not sure if that
+  //            behaviour is intentional or not.
```
BraedonLeonard, line 306:
```suggestion
```

vvagaytsev (approved):
Great catch! Thanks for fixing this @BraedonLeonard! 💯 👍

PR #6530, opened 2024-10-10
fix(k8s): add imagePullSecret from kubernetes provider to sync init container

### LINKED ISSUE #6525: [Bug]: Image pull secrets not set on sync util pod causing Dockerhub rate limiting
### Current behavior

Due to #5336, users can get rate limited when pulling Garden images from our public Dockerhub repo if they haven't added an image pull secret with their Docker credentials to their cluster. Admittedly they shouldn't need that in the first place and we are applying for the "Docker Verified Publisher" program which should fix this.

Adding the secret fixes the issue for some of the util pods but not for the sync pod because the image pull secrets aren't set in its manifest. 

We need to fix that for now, although again, if/when we're a verified publisher it shouldn't be needed in the first place.

### Expected behavior

Creating the sync util pod shouldn't fail with a Dockerhub rate limit error if the user already has created image pull secrets and specified in the Garden config (under the K8s provider `imagePullSecrets` field).

### Reproducible example

N/A

### Workaround

N/A

### Suggested solution(s)

We should set the `imagePullSecrets` on the sync util Deployment like we e.g. do for the buildkit Deployment here: https://github.com/garden-io/garden/blob/main/core/src/plugins/kubernetes/container/build/buildkit.ts#L218

twelvemo:

Fixes #6525

### DIFF
--- a/core/src/plugins/kubernetes/sync.ts
+++ b/core/src/plugins/kubernetes/sync.ts
@@ -72,6 +72,7 @@ import { gardenEnv } from "../../constants.js"
 import { styles } from "../../logger/styles.js"
 import { commandListToShellScript } from "../../util/escape.js"
 import { toClearText } from "../../util/secrets.js"
+import type { V1Container } from "@kubernetes/client-node"
 
 export const builtInExcludes = ["/**/*.git", "**/*.garden"]
 
@@ -462,7 +463,7 @@ export async function configureSyncMode({
     }
     const k8sSyncUtilImageName = getK8sSyncUtilImageName()
     if (!podSpec.initContainers.find((c) => c.image === k8sSyncUtilImageName)) {
-      const initContainer = {
+      const initContainer: V1Container = {
         name: k8sSyncUtilContainerName,
         image: k8sSyncUtilImageName,
         command: [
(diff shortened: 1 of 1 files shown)


eysi09 (dismissed):
Looks great but I think there's some issue with Prettier, just bumped into the same. 

Locally it fails if there's that extra parenthesis, but in CI it fails if it's not there 🙃

core/src/plugins/kubernetes/sync.ts
```diff
@@ -334,7 +334,7 @@ export async function configureSyncMode({
           Override configuration:
           ${(override.command?.length ?? 0) > 0 ? `Command: ${override.command?.join(" ")}` : ""}
           ${(override.args?.length ?? 0) > 0 ? `Args: ${override.args?.join(" ")}` : ""}
-          ${(override.image?.length ?? 0) ? `Image: ${override.image}` : ""}
+          ${override.image?.length ?? 0 ? `Image: ${override.image}` : ""}
```
vvagaytsev, line 337: This causes the [lint failure](https://app.circleci.com/jobs/github/garden-io/garden/567290) because we switched to `prettier 3.3.3` in #6500.

vvagaytsev:
@eysi09 @twelvemo just run `npm i` locally to get the latest `prettier` from #6500 :)

eysi09:
> @eysi09 @twelvemo just run `npm i` locally to get the latest `prettier` from #6500 :)

Ah right, I did that but hadn't pulled latest main. Didn't realise it was that recent a change.

twelvemo:
Thanks for the hint with the linter! Also should've added the imagePullSecret to the pod and not the container spec. Importing the type for the container spec will make sure there is no confusion here again.

PR #6532, opened 2024-10-11
improvement: make sure artifacts are always fetched

### LINKED ISSUE #6524: [Bug]: Test exec doesn't copy artifacts if command exits with non-0 code
### Current behavior

If a the command of a `Test` of type `exec` fails with an exit code, the artifacts are not copied.

### Expected behavior

Garden should try to copy artifacts, at least if any are present.

Example use-case: 
- User runs e2e frontend tests.
- Tests generate comparison screenshots or videos from changes.
- User wants to see the screenshots with the errors in order to debug the tests.

### Reproducible example

```yaml

# example action config

kind: Test
type: exec
name: example-exit-code
description: Example Exit
spec:
  command: 
    - sh
    - example-exit-code.sh
  artifacts:
    - source: tmp/example-exit-code.txt
      target: tests
```

```sh

# example script

echo "data" > tmp/example-exit-code.txt
exit 1
```

### Your environment

* **OS:** 

`garden version`

10ko:

When executing `Run` or `Test` actions of type `exec`, we now always try to export the artifacts.
More specifically, we make sure to fetch artifacts in the case of:
- A `spec.command` failing with a non-zero exit code.
  - In this case we will only log an errror in case of failure while fetching artifacts.
- A `spec.command` throwing unexpectedly.
  - Same as above, only a warning.

In case the `spec.command` runs successfully and the fetching fails, we maintain the current behaviour of throwing an error.

This PR closes #6524

Fixes #


core/src/plugins/exec/test.ts
```diff
@@ -34,45 +34,68 @@ export type ExecTest = GardenSdkActionDefinitionActionType<typeof execTest>
...
+    } else {
+      throw error
+    }
   }
 
-  const result = {
-    state: runResultToActionState(detail),
-    detail,
-    outputs: {
+  if (runCommandErrors.length === 0) {
```
vvagaytsev, line 61: A minor non-blocking nit :)

This `if`-condition can be inverted, and the `else`-block can be then unwrapped. Thus, we'll have an error-check and throw immediately if any errors are there, and the errorless code branch will be outside a long `if`-block.

vvagaytsev (dismissed):
👍

10ko:
Sorry I should have made it a draft. These changes do not account to all the use-cases unfortunately, since if the run command returns a non-zero exit code, it will not throw. So I am amending the code as we speak.

10ko:
Good to go now, @vvagaytsev

core/src/plugins/exec/test.ts
stefreak: @10ko you forgot to invert the condition (also at https://github.com/garden-io/garden/pull/6532/files#diff-fa98144b93cb2ad5871e8c16c14184f17a93eac35549d003bd0e239a134af471R59) 
```suggestion
  if (runCommandErrors.length) {
```

core/src/plugins/exec/run.ts
```diff
@@ -32,8 +32,34 @@ export type ExecRun = GardenSdkActionDefinitionActionType<typeof execRun>
 execRun.addHandler("run", async ({ artifactsPath, log, action, ctx }) => {
   const startedAt = new Date()
   const { command, env, artifacts } = action.getSpec()
+  const runCommandErrors: unknown[] = []
```
stefreak, line 35: If there's always one element in the array, I think it's cleaner to use `let` here:
```suggestion
  let runCommandError: unknown | undefined
```

stefreak (changes requested):

core/src/plugins/exec/run.ts
10ko: `runCommandErrors` is never reassigned, it will give you a type error if you use `let`.

stefreak: it's never reassigned because you push an element to the array, but why use an array in the first place if it'll never have more than one element?

10ko: aaah, that's what you meant. Sorry didn't understand you from your first comment. Sure thing, that makes sense.

stefreak: @10ko I've pushed a commit with the change I had in mind, does it look good to you?

stefreak (approved):
feel free to merge if you think my commit is ok 👍

core/src/plugins/exec/run.ts
stefreak: No worries :D

PR #6533, opened 2024-10-11
fix(k8s): correctly set image pull secret on sync pod

eysi09:
Before this fix, we'd add image pull secrets with name and namespace to the Pod spec but the namespace field isn't allowed so applying the manifest fails. Now we only add the name.

This isn't caught by the TS compiler because it doesn't check for excess properties (except when using object literals). 

That is:

```typescript
interface Foo {
  foo: string
}

const foo1: Foo = {
  foo: "foo",
  bar: "bar"
} // Doesn't compile, 'bar' does not exist on 'Foo'

const someVar = {
  foo: "foo",
  bar: "bar"
}

const foo2: Foo = someVar // Compiles
```

TODO:

- [x] Add tests


twelvemo:
Oh, good catch 👍  I oversaw that.

eysi09:
Just adding some tests before I flag as ready for review

eysi09:
FYI: Added tests and also re-enabled the tests that were being skipped. Not quite sure what the original issue was I refactored the code a bit and now they seem to pass.

PR #6549, opened 2024-10-15
fix: fail-safe dependency config lookup in action config processing

vvagaytsev:

Handle potentially wrong action names produced from configuration templates,
when a templated action name from a config template is evaluated to `null` or `undefined`.
Such resolved template values are interpolated into the string values for the action names.
This is how the current template resolution engine works.

We should skip such dependencies instead of failing fast with an error.

Fixes #

### DIFF
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -996,7 +996,7 @@ function dependenciesFromActionConfig({
         if (weirdActionNames.has(name)) {
           const highlightedWeirdName = styles.highlight(`"${name}"`)
           log.warn(
-            `Found a dependency with suspicious name ${highlightedWeirdName}. It looks like it was rendered as ${highlightedWeirdName} string from the configuration template ${styles.highlight(config.internal.templateName)}`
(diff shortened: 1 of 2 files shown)


core/src/graph/actions.ts
```diff
@@ -979,26 +986,36 @@ function dependenciesFromActionConfig({
...
+    .map((d) => {
+      const { kind, name } = parseActionReference(d)
+      const depKey = actionReferenceToString(d)
+      const depConfig = configsByKey[depKey]
+
+      if (!depConfig) {
+        if (weirdActionNames.has(name)) {
+          const highlightedWeirdName = styles.highlight(`"${name}"`)
+          log.warn(
+            `Found a dependency with suspicious name ${highlightedWeirdName}. It looks like it was rendered as ${highlightedWeirdName} string from the configuration template ${styles.highlight(config.internal.templateName)}`
```
thsig, line 999:
```suggestion
            `Found a dependency with suspicious name ${highlightedWeirdName}. It was rendered from the configuration template ${styles.highlight(config.internal.templateName)}. The template expression for this action name may have resolved to null or undefined, which is probably a mistake. Please take a look at the template expression in question.`
```

thsig (changes requested):
Looks good! Added one suggested change to the warning message text there.

thsig (approved):

PR #6552, opened 2024-10-16
feat(k8s): allow using registry mirror for utility images

eysi09:

Garden uses various utility container images, e.g. for the K8s sync utility pod, and hosts them on Docker Hub.

Some users have their own mirror of Docker Hub that they'd rather use instead of Docker Hub itself. This avoids issues with rate limiting among other things.

This commit adds a config field on the Kubernetes provider spec that allows users to configure what registry is used.

Note that it only allows users to specify the registry domain, not the repository, image name, or tag. So the assumption is that mirror has the exact same "layout" as Docker Hub itself which is usually (always?) the case.

If needed, we can later expose more granular control, e.g. setting the image tag, but that's not required for now.

Fixes #


twelvemo (approved):
Great stuff. Love the FAQ!

PR #6553, opened 2024-10-16
chore: logging corrections and tweaks

vvagaytsev:

Fixes #


thsig (changes requested):
Looks good! I think we should increase the verbosity of these new log lines from `debug` to `silly`, since they're quite fine-grained. Do you agree?

thsig (approved):

PR #6555, opened 2024-10-16
feat(pulumi): add `spec.showSecretsInOutput` config to Pulumi deploy action

BraedonLeonard:

Some outputs in Pulumi are considered secrets, and you need to use the `--show-secrets` flag to show them. If you don't, secret outputs get replaced with `[secret]`. Because the Pulumi provider doesn't include the `--show-secrets` flag, you can't use outputs from Pulumi which are marked as secrets, which can really limit the usefulness of the Pulumi deploy action's outputs.

This PR makes it so that secret values are included in the output of the Pulumi deploy action.

### DIFF
--- a/docs/reference/action-types/Deploy/pulumi.md
+++ b/docs/reference/action-types/Deploy/pulumi.md
@@ -441,6 +441,19 @@ The name of the pulumi stack to use. Defaults to the current environment name.
 | -------- | -------- |
 | `string` | No       |
 
+### `spec.showSecretsInOutput`
+
+[spec](#spec) > showSecretsInOutput
+
+When set to true, stack outputs which are marked as secrets will be shown in the output.
+
+By default, Pulumi will print secret stack outputs as the string '[secret]' instead of
+the true content of the output.
+
+| Type      | Default | Required |
+| --------- | ------- | -------- |
+| `boolean` | `false` | No       |
+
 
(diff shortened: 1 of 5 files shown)


BraedonLeonard:
By the way, I'm not able to add reviewers to PRs for some reason, which is why I haven't been explicitly adding any to my PRs.

eysi09:
Thanks for the contribution @BraedonLeonard! I assigned @thsig, our in-house Pulumi specialist.

thsig:
Thanks for the PR, @BraedonLeonard!

I'm wondering if this should be opt-in on a per-action basis, just so users are explicitly aware of the possible security implications.

Could you add an `showSecretsInOutput` boolean configuration field to `pulumiDeploySchemaKeys` here (with `false` as the default):

https://github.com/garden-io/garden/blob/5e1fa2182d282057d10652dfbe2cf2a7cfe4c68d/plugins/pulumi/src/action.ts#L34

And then only append the `--show-secrets` CLI option to the `stack output` command when this flag is set to `true`?

BraedonLeonard:
@thsig Just pushed the changes to add make this behaviour configurable, let me know if it looks good to you 👍

thsig (approved):
Looks good, thanks for your contribution! 🎉

PR #6559, opened 2024-10-18
docs: fix header on migrating back to old mutagen

highb:

Updates the header on the docs for switching from the new mutagen back to old to match the content (new to old.) Docs are less confusing after the change.

I haven't checked if there is an existing issue. I assumed not, since it is minor.


vvagaytsev (approved):

vvagaytsev:
Thanks for spotting this @highb!

The original header was written when the new Mutagen was introduced as an experimental option available via a feature flag, I forgot to update it when we switched default sync machinery to the new Mutagen :)

Thanks for your contribution! 👍

PR #6572, opened 2024-10-22
feat(exec): separate `stdout` and `stderr` in Run and Test `exec`-action outputs

### LINKED ISSUE #6556: [FEATURE]: Provide separate action output for stdout and stderr
### Background / Motivation

Currently I use exec actions to fetch various credentials, which are then used by downstream actions/deployments by accessing the credentials with `${actions.run.<name>.outputs.log}`.

In addition to `${actions.run.<name>.outputs.log}`, which provides stdout and stderr combined, we should be able to access `${actions.run.<name>.outputs.stdout}` and `${actions.run.<name>.outputs.stderr}`.

The tools I'm running in my exec actions will log the part that I want to use (e.g. an access token) to stdout, while they'll log other useful information for the user to stderr. I only want to use the part that's sent to stdout, but since Garden only provides the two streams combined, I have to redirect stderr to `/dev/null`. Ideally, we'd let stderr get printed since it has useful information, without necessarily combining the streams in the action output.

vvagaytsev:

Fixes #6556

### DIFF
--- a/core/src/plugins/exec/config.ts
+++ b/core/src/plugins/exec/config.ts
@@ -31,6 +31,18 @@ export const execRuntimeOutputsSchema = s.object({
     .describe(
       "The full log output from the executed command. (Pro-tip: Make it machine readable so it can be parsed by dependants)"
     ),
+  stdout: s
+    .string()
+    .default("")
+    .describe(
+      "The stdout log output from the executed command. (Pro-tip: Make it machine readable so it can be parsed by dependants)"
+    ),
+  stderr: s
+    .string()
+    .default("")
+    .describe(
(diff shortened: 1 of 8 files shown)


vvagaytsev:
@eysi09 @thsig this addresses the feature request from #6556.

I've implemented that only for Run and Test action kinds of exec action type. That was quite simple do to.

Should we do the same for container, kubernetes and helm action types? That requires broader changes across the code base and the log retrieval utilities.

thsig:
> Should we do the same for container, kubernetes and helm action types? That requires broader changes across the code base and the log retrieval utilities.

It would definitely make sense to have that too, but maybe we should wait for a user or two to ask for it before starting work on it.

PR #6638, opened 2024-11-14
fix(container): container registry namespace is empty when not specified

twelvemo:

Fixes #


vvagaytsev (approved):
Looks good, thank you! 👍

PR #6653, opened 2024-11-19
refactor: get rid of `fs.Stats` constructor deprecation warning

vvagaytsev:

Remove unnecessary child-class and use type intersection with factory method instead.

Fixes compilation errors in #6622.


stefreak (approved):
Nice!

The build staging code is very complex, great to get rid of that deprecation warning.

PR #6668, opened 2024-11-22
chore(varfiles): avoid unnecessary path existence checks

vvagaytsev:

Fixes #

### DIFF
--- a/core/src/config/base.ts
+++ b/core/src/config/base.ts
@@ -10,13 +10,10 @@ import dotenv from "dotenv"
 import { sep, resolve, relative, basename, dirname, join } from "path"
 import { load } from "js-yaml"
 import { lint } from "yaml-lint"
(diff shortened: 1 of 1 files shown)


stefreak (approved):

core/src/config/base.ts
```diff
@@ -632,6 +614,20 @@ export const loadVarfile = profileAsync(async function loadVarfile({
...
+      throw error
+    }
+
+    if (isErrnoException(error) && error.code === "ENOENT") {
+      if (path && path !== defaultPath && !optional) {
+        throw new ConfigurationError({
+          message: `Could not find varfile at path '${path}'. Absolute path: ${resolvedPath}`,
+        })
+      } else {
+        return {}
```
stefreak, line 627:
```suggestion
        // We return empty object if the defaultPath does not exist
        return {}
```

PR #6669, opened 2024-11-22
perf(template): avoid eager variables merging

vvagaytsev:

Use proxy that scans the given scopes sequentially based on the order pof precedence.

Improves performance


stefreak:
didn't bring expected performance boost.

PR #5496, opened 2023-11-29
WIP: input tracking

stefreak:
**What this PR does / why we need it**:
- Input tracking, to allow excluding inputs in a future PR
- Separate parsing template strings from evaluating template strings by generating an AST in the parser
- Implement lazy evaluation
- implement a proxy so we avoid changing too much existing code that uses templates

Todo:
- Integrate all that with the existing code base (Parsing, validating and processing configs)

**Special notes for your reviewer**:

PR #6685, opened 2024-12-02
perf: optimise template string resolving performance

stefreak:
We optimise the template string resolving performance by
making sure that we only parse each template expression once. Once
parsed, we use the AST to evaluate template expressions.

Fixes #

Changes are inspired from https://github.com/garden-io/garden/pull/5496


core/src/config/template-contexts/base.ts
```diff
@@ -193,16 +194,10 @@ export abstract class ConfigContext {
...
-        throw new TemplateStringPassthroughException({
-          message,
-        })
-      } else if (opts.allowPartial) {
-        throw new TemplateStringMissingKeyException({
-          message,
-        })
+      if (this._alwaysAllowPartial || opts.allowPartial) {
+        return {
+          resolved: CONTEXT_RESOLVE_KEY_NOT_FOUND,
```
stefreak, line 199: We forgot to return `CONTEXT_RESOLVE_KEY_AVAILABLE_LATER`, this needs to be revisited

core/src/config/render-template.ts
```diff
@@ -288,7 +288,14 @@ async function renderConfigs({
       let resolvedName = m.name
 
       try {
-        resolvedName = resolveTemplateString({ string: m.name, context, contextOpts: { allowPartial: false } })
+        const result = resolveTemplateString({ string: m.name, context, contextOpts: { allowPartial: false } })
+        if (typeof result === "string") {
+          resolvedName = result
+        } else {
+          throw new ConfigurationError({
+            message: "must resolve to string",
```
stefreak, line 296: We should pass expected type to `resolveTemplateString`, to make sure that error message points the user to the correct line and file in the code.

core/src/resolve-module.ts
```diff
@@ -529,8 +529,7 @@ export class ModuleResolver {
 
     const configContext = new ModuleConfigContext(contextParams)
 
-    const templateRefs = getModuleTemplateReferences(rawConfig, configContext)
-    const templateDeps = <string[]>templateRefs.filter((d) => d[1] !== rawConfig.name).map((d) => d[1])
+    const templateDeps = getModuleTemplateReferences(rawConfig, configContext)
```
stefreak, line 532: where do we perform the filter step now?

core/src/graph/actions.ts
```diff
@@ -1035,25 +1031,11 @@ const dependenciesFromActionConfig = profile(function dependenciesFromActionConf
...
 
-    if (maybeTemplateString(ref.name)) {
-      try {
-        ref.name = resolveTemplateString({
-          string: ref.name,
-          context: templateContext,
-          contextOpts: { allowPartial: false },
-        })
-      } catch (err) {
-        log.warn(
```
stefreak, line 1051: Todo double check: did we retain old behaviour where we only warn?

stefreak: I find it extremely unlikely that throwing instead of printing this confusing warning is a breaking change. is there a use case?

10ko (approved):

core/src/graph/actions.ts
edvald: I suspect a couple of users may run into this in the context of action templates. Is there a particular reason to skip this entirely, instead of changing to match the new behavior?

core/.mocharc.yml
```diff
@@ -1,5 +1,5 @@
 require:
-  - build/test/setup.js
+  - ../../../../build/test/setup.js
```
edvald, line 2: Seems a little odd - What prompted this change?

core/src/exceptions.ts
```diff
@@ -309,11 +309,23 @@ export class CloudApiError extends GardenError {
...
-  path?: ObjectPath
-
-  constructor(params: GardenErrorParams & { path?: ObjectPath }) {
-    super(params)
-    this.path = params.path
+  loc: Location
+  originalMessage: string
+
+  constructor(params: GardenErrorParams & { loc: Location }) {
+    // TODO: figure out how to get correct path
```
eysi09: Are we leaving this in intentionally?

core/.mocharc.yml
stefreak: Vovas IDE needed it to be able to run tests, and it didn't seem to have an effect on CI

core/src/exceptions.ts
stefreak: has been updated in the latest commit

core/.mocharc.yml
stefreak: But I agree it's odd, will have a closer look at this

core/src/graph/actions.ts
stefreak: I've managed to reproduce triggering the warning in `garden validate` using this template:

```
kind: Build
name: foobar2
type: exec
spec:
  command: [echo, '${actions.build[var.foo].outputs.log}']
variables:
  hello: {}
```

`garden build` fails with this error message (before my changes):
```
Invalid template string (${actions.build[var.foo].outputs.log}) at path command.1: Could not find key [object Object] under actions.build. Available keys: foobar2.
```

With the new changes the error now looks like this, and we fail already during validation:
```
Found invalid action reference: /Users/steffen/repro/project.garden.yml:22
...
20  | spec:
21  |   command: [echo, '${actions.build[var.foo].outputs.log}']
------------------------^
Could not find key foo under var. Available keys: hello.
```

The new error message is way more helpful compared to the warning we used to print (`⚠ graph                     → Unable to infer dependency from action reference in exec Build foobar2, because template string '${var.foo}' could not be resolved. Either fix the dependency or specify it explicitly.`. Failing earlier in the validation avoids the inevitable follow-up error when the action is being resolved.

I'll continue to look if I can find any issues with `ConfigTemplate`. If you have a scenario or example where the new approach will fail, please let me know

core/src/config/render-template.ts
```diff
@@ -273,29 +273,29 @@ async function renderConfigs({
   context: RenderTemplateConfigContext
   renderConfig: RenderTemplateConfig
 }): Promise<TemplatableConfig[]> {
+  const source = { yamlDoc: template.internal.yamlDoc, path: ["configs"] }
```
vvagaytsev, line 276: Should this path be `"configs"` or `"inputs"`?

PR #6695, opened 2024-12-09
fix: retain bug-compatibility for referencing missing variables in unary operators (`!` and `typeof`).

stefreak:
Older versions of garden did not throw an error when referencing
missing variables from unary operators like `!` and `typeof`. This
commit restores backwards-compatiblity that has been broken in #6685


10ko (approved):
🚀

PR #6706, opened 2024-12-10
Simplify provider init

eysi09:

Previously, Garden would first run the `getEnvironmentStatus` handler
and depending on the results, either run the `prepareEnvironment`
handler or not.

Now we always run the `prepareEnvironment` handler (if the provider
status is not cached).

The reason for this change is that the previous flow is an
unneeded abstraction that doesn't serve any purpose at this point and
makes the provider initialisation harder to reason about.

The specific motivation is that there's bug in the Kubernetes
provider where some resources aren't created when the
environment is prepared. It's actually more work to first check
the statuses of these resources, return those, and then create
them. In particular since we have helpers to create them in
an idempotent manner.

That's why now we just always call `prepareEnvironment` and have that
handler figure out what needs to be done.

We still keep the `getEnvironmentStatus` handler for status-only
commands but now it doesn't have side effects. Whereas before the
status handler would e.g. create a namespace.

Plugins can then decide to use that if needed. E.g. the `prepareEnvironment`
handler for the Terraform plugins uses `getEnvironmentStatus` to decide
what needs to happen.

We'll fix the actual bug that prompted this change in a follow up
commit.

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/init.ts
+++ b/core/src/plugins/kubernetes/init.ts
@@ -9,10 +9,10 @@
 import { KubeApi, KubernetesError } from "./api.js"
 import {
   getAppNamespace,
-  prepareNamespaces,
+  prepareNamespace,
   deleteNamespaces,
-  getNamespaceStatus,
   getSystemNamespace,
+  namespaceExists,
 } from "./namespace.js"
 import type { KubernetesPluginContext, KubernetesConfig, KubernetesProvider, ProviderSecretRef } from "./config.js"
 import type {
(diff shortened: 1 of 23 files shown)


core/src/plugins/kubernetes/init.ts
```diff
@@ -72,8 +72,9 @@ export async function getEnvironmentStatus({
   const k8sCtx = <KubernetesPluginContext>ctx
   const provider = k8sCtx.provider
   const api = await KubeApi.factory(log, ctx, provider)
-
-  const namespace = await getNamespaceStatus({ ctx: k8sCtx, log, provider: k8sCtx.provider, skipCreate: true })
+  // TODO @eysi: Avoid casting (the namespace is ensured though when the provider config is resolved)
```
thsig, line 75: I think this cast might become unnecessary with the changes to provider init that @stefreak and @vvagaytsev have been working on, but I'm not sure.

thsig (approved):

Issue #5266, opened 2023-10-18
0.13: [Bug]: deploy actions resolve template strings in manifests in a way that breaks for multi-line strings 


stefreak:
### Current Behavior
Error cannot parse yaml with multiline string in variable templated to a manifest file

### Expected behavior

Should work.

### Reproducible example

manifest:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test-configmap
data:
  hello: ${var.multiline-var}
```

deploy action:
```yaml
kind: Deploy
type: kubernetes
name: multiline-var-in-manifest-deploy
spec:
  files: [ "deployment.yml" ]
variables:
  multiline-var: |
    this is a
    multiline
    variable
  foo: bar
```

### Workaround

Works with inline manifest

### Suggested solution(s)

Resolve variables after parsing 

### Your environment

* **OS:** macos
* **How I'm running Kubernetes:** orbstack

`garden version` latest main

alexkuretz:
I recently ran into this and had to b64 encode some jwt tokens, would be nice to get this fixed

PR #6713, opened 2024-12-12
fix(template): establish backwards bug-compatibility for kubernetes manifest files

### LINKED ISSUE #6711: [Bug]: Templating logic change in garden 0.13.46 breaks deploys
### Current behavior

```
Error from server (Invalid): error when creating "STDIN": Ingress.networking.k8s.io "${this.name}-0" is invalid: [metadata.name: Invalid value: "${this.name}-0": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), spec.rules[0].host: Invalid value: "${providers.kubernetes.outputs.default-hostname}": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), spec.rules[0].http.paths[0].backend.service.name: Invalid value: "${this.name}": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')]
Error from server (Invalid): error when creating "STDIN": Ingress.networking.k8s.io "${this.name}-1" is invalid: [metadata.name: Invalid value: "${this.name}-1": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), spec.rules[0].host: Invalid value: "${providers.kubernetes.outputs.default-hostname}": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), spec.rules[0].http.paths[0].backend.service.name: Invalid value: "${this.name}": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')]
Error from server (Invalid): error when creating "STDIN": Service "${this.name}" is invalid: [metadata.name: Invalid value: "${this.name}": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?'), spec.ports[0].targetPort: Invalid value: "${var.port}": must contain only alpha-numeric characters (a-z, 0-9), and hyphens (-), spec.selector: Invalid value: "${this.name}": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')]
Error from server (BadRequest): error when creating "STDIN": Deployment in version "v1" cannot be handled as a Deployment: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
```

### Expected behavior

To work the way it has

### Reproducible example

Not entirely sure, so here's some snippets.
project.garden.yml:
```
variables:
  service: task
  deploy-extra-env-vars: []

environments:
  - name: dev
    defaultNamespace: user-${local.username}
    variables:
      buildversion: user-${local.username}
      default-hostname: ${var.service}-user-${local.username}.dev.mycompany.com
```
manifest:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: ${this.name}-0
  annotations:
    ingress.kubernetes.io/force-ssl-redirect: "false"
    nginx.ingress.kubernetes.io/affinity: cookie
    nginx.ingress.kubernetes.io/enable-cors: "true"
spec:
  ingressClassName: nginx
  rules:
    - host: ${providers.kubernetes.outputs.default-hostname}
      http:
        paths:
          - pathType: Prefix
            path: "/v1"
            backend:
              service:
                name: ${this.name}
                port:
                  number: 80
```

### Workaround

Reverting to 0.13.45 works.

### Your environment

* **OS:** Ubuntu

`garden version`
0.13.46

stefreak:

Template strings in Kubernetes manifest files are partially resolved for legacy reasons. The partial resolve behaviour has been changed to improve sanity in https://github.com/garden-io/garden/pull/6685 but for kubernetes manifest files, we need to have a version of the old behaviour.

The expected legacy behaviour looks like this:
- If template strings reference variables that do not exist there is no error
- If a template string contains multiple template expressions, each expression can be partially resolved.
- Template expressions are evaluated before parsing yaml, which means that valid yaml with template expressions can be resolved to invalid yaml (e.g. if variable values contain special characters)

This legacy behaviour can lead to quite some surprises and UX problems, for example https://github.com/garden-io/garden/issues/5266

I would suggest that we remove this functionality in 0.14 in favor of the `patchManifests` feature

Fixes #6711

### DIFF
--- a/core/src/util/fs.ts
+++ b/core/src/util/fs.ts
@@ -253,14 +253,14 @@ export async function makeTempDir({
 }: { git?: boolean; initialCommit?: boolean } = {}): Promise<TempDirectory> {
   const tmpDir = await tmp.dir({ unsafeCleanup: true })
   // Fully resolve path so that we don't get path mismatches in tests
-  const tmpPath = await realpath(tmpDir.path)
+  tmpDir.path = await realpath(tmpDir.path)
 
   if (git) {
-    await exec("git", ["init", "--initial-branch=main"], { cwd: tmpPath })
+    await exec("git", ["init", "--initial-branch=main"], { cwd: tmpDir.path })
     if (initialCommit) {
-      await writeFile(join(tmpPath, "foo"), "bar")
-      await exec("git", ["add", "."], { cwd: tmpPath })
-      await exec("git", ["commit", "-m", "first commit"], { cwd: tmpPath })
+      await writeFile(join(tmpDir.path, "foo"), "bar")
+      await exec("git", ["add", "."], { cwd: tmpDir.path })
+      await exec("git", ["commit", "-m", "first commit"], { cwd: tmpDir.path })
     }
   }
 
--- a/core/src/config/template-contexts/base.ts
+++ b/core/src/config/template-contexts/base.ts
@@ -25,6 +25,10 @@ export type ContextKey = ContextKeySegment[]
 export interface ContextResolveOpts {
   // Allow templates to be partially resolved (used to defer runtime template resolution, for example)
   allowPartial?: boolean
+  // This is kept for backwards compatibility of rendering kubernetes manifests
+  // TODO(0.14): Do not allow the use of template strings in kubernetes manifest files
+  // TODO(0.14): Remove legacyAllowPartial
+  legacyAllowPartial?: boolean
   // Allow partial resolution for values that originate from a special context that always returns CONTEXT_RESOLVE_KEY_AVAILABLE_LATER.
   // This is used for module resolution and can be removed whenever we remove support for modules.
   allowPartialContext?: boolean
--- a/core/src/plugins/kubernetes/kubernetes-type/common.ts
+++ b/core/src/plugins/kubernetes/kubernetes-type/common.ts
@@ -428,7 +428,17 @@ async function readFileManifests(
         const absPath = resolve(manifestPath, path)
         log.debug(`Reading manifest for ${action.longDescription()} from path ${absPath}`)
         const str = (await readFile(absPath)).toString()
-        const resolved = ctx.resolveTemplateStrings(str, { allowPartial: true, unescape: true })
+
+        // TODO(0.14): Do not resolve template strings in unparsed YAML and remove legacyAllowPartial
+        // First of all, evaluating template strings can result in invalid YAML that fails to parse, because the result of the
+        // template expressions will be interpreted by the YAML parser later.
+        // Then also, the use of `legacyAllowPartial: true` is quite unfortunate here, because users will not notice
+        // if they reference variables that do not exist.
+        const resolved = ctx.resolveTemplateStrings(str, {
+          legacyAllowPartial: true,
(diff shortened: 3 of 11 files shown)


core/src/util/fs.ts
```diff
@@ -253,14 +253,14 @@ export async function makeTempDir({
 }: { git?: boolean; initialCommit?: boolean } = {}): Promise<TempDirectory> {
   const tmpDir = await tmp.dir({ unsafeCleanup: true })
   // Fully resolve path so that we don't get path mismatches in tests
-  const tmpPath = await realpath(tmpDir.path)
+  tmpDir.path = await realpath(tmpDir.path)
```
stefreak, line 256: This change was needed to make it possible to run the integ tests locally on macos.

vvagaytsev (approved):
Great job! 🚀

thsig:
I very much agree that we should deprecate this in 0.14.

There are also problems in versioning `kubernetes` actions that use template strings in manifest files: If a variable is used in a manifest file but doesn't appear in the action config, changing that variable won't change the action version (since it's computed based on the unresolved manifest file).

stefreak:
If we want to deprecate this, we need to deprecate this in 0.13 and remove it in 0.14. That way users can be sure that their configuration will work with 0.14 if they eliminate the deprecation warnings.

stefreak:
Would be good to have an easy to use alternative, e.g. an `includeYaml` statement or a `templateFile` statement

PR #6714, opened 2024-12-12
fix(templates): fix regression with multiple if statements introduced in 0.13.45

stefreak:
In `buildConditionalTree`, we did not properly reset state and thus we
got a wrong AST if multiple if statements existed in the same nesting
level.

Fixes #


vvagaytsev (approved):
Great! 🚀 
Thanks for the quick fix! 👍

PR #6727, opened 2024-12-17
chore(cloud): login to and logout from new backend API

vvagaytsev:

Fixes #


docs/reference/commands.md
```diff
@@ -4091,7 +4091,7 @@ Logs you in to Garden/Grow Cloud. Subsequent commands will have access to cloud
 
 ### garden logout
 
-**Log out of Garden Cloud.**
+**Log out of Garden/Grow Cloud.**
```
stefreak, line 4094: The existence of Grow Cloud should not reflect in the docs until we launch 0.14

core/src/cloud/backend.ts
```diff
@@ -76,6 +81,30 @@ export class GardenCloudBackend extends AbstractGardenBackend {
...
+
+    if (!cloudApi) {
+      return
+    }
+
+    try {
+      await cloudApi.post("token/logout", { headers: { Cookie: `rt=${clientAuthToken?.refreshToken}` } })
+    } catch (err) {
+      log.debug({ msg: "Failed to revoke token; it was either invalid or already expired." })
+      throw err
```
stefreak, line 102: if we're at it, this pattern is extremely dangerous- even `TypeError` would be catched here. this needs a guard check to only handle known issue (e.g. 401 status code).
For other issues, we want to throw here.

Also, in case we ignore the response due to a 401, we want to change the log level to warning, so the user is aware that the token could not be revoked.

stefreak: This is likely business logic that needs to be the same regardless of the backend, so the caller of `revokeToken` probably needs to implement the error handling.

stefreak: Not saying it needs to happen in this PR; Vova, you are doing an absolutely amazing job here! I love everything you did here, just noticed this when reading through the diff. This can be done in a separate PR, and I know that you probably had similar thoughts already 👍 ❤️

docs/reference/commands.md
vvagaytsev: Thanks! Reverted that in 2cb8e3a38d4f8ae298d84d7bbce1ae56f5836a11.

core/src/cloud/backend.ts
vvagaytsev: Ah, right. Actually, it's similar to `catch (Throwable e)` in Java, where you can catch even an OOM and other irrcoverable errors :)
Let's change this in the same PR to keep the things safe 👍

vvagaytsev: Thanks for spotting this Steffen! That was something I expected to discuss in the code review round :)

vvagaytsev: Actually, we always re-throw it here, I just borrowed an intermediate debug-log statement from Grow login command.

I think it's better to remove catch here, and let it thro, as it was originally in the Garden command.
The thing that we can change is the surrounding catch block that can handle the errors in more accurate way.

stefreak: Ah makes sense, I overlooked that we are throwing, sorry should have had lunch instead of reading the diff :D

stefreak (changes requested):
I have two minor comments, but other than that this is a big step towards easier to maintain code base while supporting another backend at the same time! Awesome

core/src/cloud/grow/trpc.ts
```diff
@@ -0,0 +1,127 @@
...
+  constructor({ cause, meta }: { cause: TRPCClientError<never>; meta: ErrorMeta | undefined }) {
+    let message: string
+    if (meta?.response.status === 401) {
+      message = `Authentication required; please log in with ${styles.highlight("grow login")} and retry.`
+    } else if (
+      // FIXME: use the actual cloudApiUrl
+      // cloudApiOrigin.endsWith("grow.garden.io") &&
+      cause.message.match(/unable to verify the first certificate/)
+    ) {
+      message = `TLS certificate verification failed; try disabling it in development by setting exporting ${styles.highlight("NODE_TLS_REJECT_UNAUTHORIZED=0")} and retry.`
```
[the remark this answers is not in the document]
vvagaytsev: Good point! Fixed that in 51602fab3212208989849bc8741f9828ca19938c.

core/src/cloud/util.ts
```diff
@@ -65,3 +90,20 @@ export function getGardenCloudDomain(configuredDomain: string | undefined): stri
...
+ * Chooses between {@link getGardenCloudDomain} and {@link getGrowCloudDomain}
+ * depending on the `USE_GROW_CLOUD` feature flag.
+ *
+ * To be used in login and logout commands for now.
+ * Later we should use the right Cloud domain insode the Garden instance
+ * and its CloudApi instance.
+ */
+export function getCloudDomain(configuredDomain: string | undefined): string {
+  const cloudDomainGetter = gardenEnv.USE_GROW_CLOUD ? getGrowCloudDomain : getGardenCloudDomain
+  return cloudDomainGetter(configuredDomain)
```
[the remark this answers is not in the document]
vvagaytsev: Either looks good IMO, but later we can hide this in the `gardenBackendFactory` anyway. Changed this in bee2297c8210460d093c3e0a55da55c41aa88ce0.

stefreak (approved):

PR #6734, opened 2024-12-19
chore: ability to use CloudBuilder via backend v2

vvagaytsev:

Fixes #


core/src/plugins/container/cloudbuilder.ts
```diff
@@ -55,20 +56,140 @@ const cloudBuilderAvailability = new LRUCache<string, CloudBuilderAvailabilityV2
...
+  }
+}
+
+async function retrieveAvailabilityFromGardenCloud({
+  ctx,
+  action,
+  config,
+}: RetrieveAvailabilityParams): Promise<CloudBuilderAvailabilityV2> {
+  const { isInClusterBuildingConfigured } = config
+  if (!ctx.cloudApi) {
```
stefreak, line 83: Can we somehow make sure we don't duplicate this if statement, perhaps by moving it to the caller of `retrieveAvailability` and use this if statement:
```suggestion
  if (!ctx.cloudApi && !ctx.cloudApiV2) {
```

vvagaytsev: I extracted a helper function to prepare an error and to reduce code duplication in bdc03692eaa23b26061e6f06c5fabb04c1b18d1e.

Semantically, the login status check was the part of each backend-specific function, and it still can/should be.
If we pull these checks up, then in each backend-specific function we'll have to perform undefined-checks for `ctx.cloudApi` and `ctx.cloudApiV2` again, or assume those are respectively defined by using `!` operator.

The latter way is brittle, the former could be simplified if we introduce predicate function and guard-types `LoggedInPluginContext` and `LoggedInPluginContextV2` for `PluginContext` (like we introduced `LoggedInGarden` for `Garden`).

That would be more type-safe, but requires more changes.
We can improve it later in the follow-up PRs to unlock this and git it merged to the main asap.

Wdyt?

vvagaytsev: Yes, we can have a kind of template function there, and make the backend-specific behavior abstract and parameterizable. Working on it now.

vvagaytsev: Done, pls check the last few commits.

stefreak (dismissed):
I love your work here ❤️ 

Let's spend time in another PR together to find out if there is another style instead of the abstract class for this particular case.

stefreak:
@vvagaytsev there is 1 lint error though

stefreak (dismissed):

vvagaytsev:
Converted to raft. Got some troubles while doing e2e testing. I'll take a closer look later, need to make sure it's still compatible with the latest backend V2 changes.

vvagaytsev:
I deleted the global config store and it helped. The e2e manual test passed.

We need to think about better UX around this case. Keeping this as a draft.

PR #6736, opened 2024-12-20
feat(templates): allow templating in template path

thsig:

Added a new template variable for `RenderTemplate` configs: `${template.path}`. This is the absolute path to the config template being rendered. This is useful when you want to use the full path to a file or directory next to the `ConfigTemplate` that's being rendered (e.g. to use a shared Dockerfile that belongs the template, every time you render that template).

Addresses a limitation raised by a user on Discord: https://discord.com/channels/817392104711651328/1318995192501764151


core/src/config/template-contexts/base.ts
```diff
@@ -316,9 +316,13 @@ export class TemplateContext extends ConfigContext {
   @schema(joiIdentifier().description(`The name of the template.`))
   public name: string
 
-  constructor(root: ConfigContext, name: string) {
-    super(root)
-    this.name = name
+  @schema(joi.string().description(`The absolute path to the directory containing the ConfigTemplate being rendered.`))
+  public path: string
```
stefreak, line 320: Wouldn't this cause an issue where the version calculation is different depending on the directory where the git repository has been checked out in?
Also, most of the time we do not accept absolute paths in the action specs.
Maybe we need to calculate the relative path to the `ConfigTemplate` from the location of the `RenderConfigTemplate`.

See also https://discord.com/channels/817392104711651328/1318995192501764151/1319639199036735500

thsig: Ah yeah, that's a good point. Relative paths would avoid the versioning problem.

core/src/config/render-template.ts
```diff
@@ -361,6 +366,7 @@ async function renderConfigs({
       // Attach metadata
       resource.internal.parentName = renderConfig.name
       resource.internal.templateName = template.name
+      resource.internal.templatePath = getRelativeTemplatePath(garden.projectRoot, template.internal.configFilePath)
```
stefreak, line 369: Can we maybe encapsulate all three values in a single optional `template` property? Then we can remove all those places where we have to throw, as if template is defined, parentName, templateName and templatePath would all be required.

stefreak (commented):
This is great overall! I had a couple of questions and a minor suggestion that would eliminate the need to throw `InternalError` if one of the template fields in `internal` is undefined.

core/src/config/render-template.ts
thsig: Yeah, I agree. I was thinking just that as I wrote this PR. I'll make that change.

Issue #5825, opened 2024-03-12
0.13: [Bug]: Template variables are being interpolated on input values


stefreak:
### Current Behavior

The action prints the literal text `I shouldn't be here.`.

### Expected behavior

I expect the action to print the literal text `${local.env.UNINVOLVED_VAR}`

### Reproducible example

Run `UNINVOLVED_VAR="I shouldn't be here." INCONSPICUOUS_VAR='${local.env.UNINVOLVED_VAR}' garden run debug` with this `garden.yml`:
```
apiVersion: garden.io/v1
kind: Project
name: foo

environments:
  - name: dev

providers: []

---
kind: Run
type: exec
name: debug
spec:
  command:
  - echo
  - '${local.env.INCONSPICUOUS_VAR}'
```

### Workaround

None available

### Suggested solution(s)

Parse templates into an abstract representation, and from there do not interpolate template strings anymore on any of the intermediate evaluation results.

### Additional context

Unfinished, but this would fix this: https://github.com/garden-io/garden/pull/5496

### Your environment

* **OS:** all

`garden version` Bonsai and Acorn affected.

stefreak:
fixed in #6745

Issue #5450, opened 2023-11-20
0.13: [Bug]: Misleading template error on "git dubious ownership" error


stefreak:
### Current Behavior

```
[verbose] garden version: 0.13.20
Validate ✔️

ℹ garden-enterprise [debug] → Initializing Garden Cloud API client.
[silly] Retrieving client auth token from config store
[debug] Checking client auth token with Garden Enterprise: https://staging.enterprise.garden.io/token/verify
[silly] Calling Cloud API with GET token/verify
[silly] Retrieving client auth token from config store
[debug] Checked client auth token with Garden Enterprise - valid: true
ℹ garden-enterprise [debug] → Authorizing...
ℹ garden-enterprise [debug] → Starting refresh interval.
[debug] Will run refresh function every 4500 ms.
[silly] Calling git with args 'rev-parse --abbrev-ref HEAD' in /Users/steffen/work/github/garden-platform
[silly] Calling git with args 'config --get remote.origin.url' in /Users/steffen/work/github/garden-platform
[silly] Tried to retrieve git remote.origin.url but encountered an error: Error: Command "git config --get remote.origin.url" failed with code 1:
ℹ garden-enterprise [verbose] → Connecting to Garden Enterprise...
ℹ garden-enterprise [debug] → Fetching project ccf57c5e-35eb-4e73-9f5a-0327f9e528e5 from https://staging.enterprise.garden.io.
[silly] Calling Cloud API with GET /projects/uid/ccf57c5e-35eb-4e73-9f5a-0327f9e528e5
[silly] Retrieving client auth token from config store
[silly] Calling Cloud API with GET /secrets/projectUid/ccf57c5e-35eb-4e73-9f5a-0327f9e528e5/env/dev
[silly] Retrieving client auth token from config store
ℹ garden-enterprise [debug] → Fetched 33 secrets from https://staging.enterprise.garden.io
✔ garden-enterprise    → Ready
Failed to resolve project configuration.
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Invalid template string (ge-preview-${slice(local.env.GARDE…): Error validating argument 'input' for slice helper function:
value is not allowed to be empty
[silly] Error: Invalid template string (ge-preview-${slice(local.env.GARDE…): Error validating argument 'input' for slice helper function:
value is not allowed to be empty
[...]

PR #6745, opened 2024-12-30
perf: improve preprocess action perf by only resolving template strings when absolutely needed

### LINKED ISSUE #3473: Helm charts (almost) always re-deploy even when nothing has changed
## Bug
I make use of a hefty Helm chart (CockroachDB's) inside a `helm` Garden module. Even when no values have been changed, on `garden deploy` it's very likely Garden will version and release a new chart to the cluster, which can take up to minutes. One of our community members also [experiences this bug](https://discord.com/channels/817392104711651328/1049847539039473747).

### Current Behavior
Here are two separate releases deployed to my cluster of the same chart. You can see one update can take up to 164 seconds so it's quite costly.

Nothing has changed in the chart or the values provided to the chart.

2022-12-28!UNITO-UNDERSCORE!15h34!UNITO-UNDERSCORE!40

2022-12-29!UNITO-UNDERSCORE!11h02!UNITO-UNDERSCORE!27

### Expected behavior
Charts should only cut a new release version when an input value has changed.

### Reproducible example
Use [my full-fat example](https://github.com/worldofgeese/flask-cockroachdb-devcontainer/tree/gardenify) with your own authenticated deployment registry defined.

### Workaround
None.

### Your environment

* **OS:** Windows 11 running WSL
* **How I'm running Kubernetes:** GKE Autopilot

`garden version`
0.12.48

### LINKED ISSUE #5202: 0.13: [Bug]: ConfigTemplate fun: using $forEach in variables results in error
### Current Behavior

Create a `ConfigTemplate` that attemps to use `$forEach` to construct an item under `variables` and get `Invalid template string` error!

```yaml
kind: ConfigTemplate
name: foobar
inputsSchemaPath: foobar.schema.json

configs:
  - kind: Run
    type: exec
    name: ${parent.name}

    variables:
      decorateWords:
        $forEach: ${inputs.someWords}
        $return: '"${item.value}" is at index ${item.key}'

    spec:
      command:
        - echo
        - ${jsonEncode(var.decorateWords)}
```

### Expected behavior

The `Run` action should print out the JSON string representing `var.decorateWords`.  For example:
```json
["\"alef\" is at index 0", "\"bet\" is at index 1", "\"gimel\" is at index 2"]
```

### Reproducible example

[Here is a repo demonstrating the issue](https://github.com/chrispsplash/bonzai-configtemplate-var-foreach-issue/tree/main).  Just clone and run `garden run baz`.  Here's sample output:
```
❯ garden run baz
Run 🏃‍♂️

ℹ garden               → Running in Garden environment default.default
✔ providers            → Cached (took 0.7 sec)
✖ run.baz              → Failed processing resolve Run type=exec name=baz (took 0.01 sec). This is what happened:

──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Invalid template string ({"$forEach":"zero,one,two,three,fo…): Could not find key item. Available keys: actions, command, datetime, environment, git, inputs, local, modules, parent, project, providers, runtime, secrets, template, this, var and variables.
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
1 test action(s) failed!

```

### Workaround

I couldn't find one.  I'm currently using a short python script to generate the values.

### Suggested solution(s)

N/A

### Additional context

N/A

### Your environment

* **OS:** macOS Ventura 13.5.2
* **How I'm running Kubernetes:** N/A

```
❯ garden version
garden version: 0.13.17
```

stefreak:

## Improve preprocess action performance

In large monorepo settings, this change improves the performance of the `garden validate` command (Which processes actions and templates) by about 35%. In smaller projects the performance improvement is not noticeable.

We achieve this by fully separating parsing of configs and evaluation of configs, and also dropping partial evaluation of template expressions. Whenever a value needs to be known right now, we fully resolve the template string.

This will fix long-standing issues like https://github.com/garden-io/garden/issues/5825 (We were prone to template injection before).

We also keep track of the YAML source for all template strings automatically and don't need to worry about manually keeping track of this anymore. This means that error messages will be improved for all users, even for Modules.

The `ModuleResolver` is the only place left where we partially resolve configs, and where we evaluate template strings and then parse them again.

## Exclude variables and varfile contents from version calculation

This PR also excludes values of variables and varfile contents from version calculation, because
what counts is how variables affect the action or module spec. If a variable is actually used and impacts the spec, it will affect version calculation. This should fix #3473

Fixes #5202, #5442, #5450, #5825, #3473 (and possibly more)


core/src/graph/actions.ts
```diff
@@ -803,50 +801,31 @@ export const preprocessActionConfig = profileAsync(async function preprocessActi
...
-      projectRoot: garden.projectRoot,
-      source: { yamlDoc, path: [] },
-    })
-
-    config = { ...config, variables: resolvedVariables, spec }
+    config = {
+      ...config,
+      ...resolvedBuiltin,
+      spec,
+    }
```
stefreak, line 821: TODO: can we somehow keep early validation? Is it dangerous to remove?

core/src/resolve-module.ts
```diff
@@ -537,13 +535,10 @@ export class ModuleResolver {
...
-      // Note: We're not implementing the YAML source mapping for modules
-      source: undefined,
-    })
+    const resolvedDeps = capture(
+      rawConfig.build.dependencies as unknown as ParsedTemplate,
+      configContext
+    ) as unknown as BuildDependencyConfig
 
     // The build.dependencies field may not resolve at all, in which case we can't extract any deps from there
     if (isArray(resolvedDeps)) {
```
stefreak, line 544: todo: this branch needs adaptation, as we do not resolve dependencies anymore.

stefreak: do dependencies need to be fully resolved? If partial is ok, is it then fine not to resolve them at all?

core/src/config/template-contexts/base.ts
```diff
@@ -344,3 +296,231 @@ export function renderKeyPath(key: ContextKeySegment[]): string {
...
+  type = "context-resolve"
+}
+
+/**
+ * Occurs when looking up a key in a context turns out to be circular.
+ */
+export class ContextCircularlyReferencesItself extends ContextResolveError {}
+
+/**
+ * Occurs when attempting to look up a key on non-primitive values.
```
vvagaytsev, line 399: Should it be "primitive" instead of "non-primitive" here? The error is thrown if a value satisfies `isTemplatePrimitive` in L424.

core/src/config/project.ts
```diff
@@ -516,32 +524,41 @@ export function resolveProjectConfig({
...
       name,
       defaultEnvironment: defaultEnvironmentName,
-      // environments are validated later
-      environments: [{ defaultNamespace: null, name: "fake-env-only-here-for-inital-load", variables: {} }],
+      // environments, providers and sources are validated later
+      environments: [{ defaultNamespace: null, name: "fake-env-only-here-for-initial-load", variables: {} }],
+      providers: [],
       sources: [],
+      // this makes sure that the output declaration shape is valid
+      outputs: serialiseUnresolvedTemplates(outputs),
```
stefreak, line 532: @vvagaytsev We neeed to avoid calling serialiseUnresolvedTemplates in production code; It is meant as a test utility. This is not the right appraoch; What we need to do is either change the schema so it accepts unresolved values, or to validate the schema after resolving outputs.

stefreak: Otherwise outputs will never be resolved

vvagaytsev: @stefreak I agree, but thanks for reminding me.
I'm not sure if I did something like this in prod code; this statement was added in 9b55bbe7.

stefreak: Ah oh sorry, I somehow thought this line was changed by you 🤦 OK, I'll address this, will keep this open as a reminder for me

vvagaytsev: It's also used in `dumpConfig` function, but there it looks ok.

core/src/template/capture.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    const result = evaluate(this.wrapped, { ...args, context })
+
+    if (!result.partial) {
+      return result
+    }
+
+    return {
+      partial: true,
+      resolved: deepMap(result.resolved, (v) => capture(v, context)) as Collection<ParsedTemplate>,
```
stefreak, line 54:
```suggestion
      resolved: deepMap(result.resolved, (v) => capture(v, this.context)) as Collection<ParsedTemplate>,
```

core/test/integ/src/plugins/kubernetes/kubernetes-type/common.ts
```diff
@@ -427,9 +429,31 @@ describe("getManifests", () => {
       graph = await garden.getConfigGraph({ log: garden.log, emit: false })
     })
 
+    function patchAction({
```
stefreak, line 432: Is it actually necessary that we get an existing action from the graph for this test, and then patch it? I think it will be easier to read if we use `TestGarden.addAction` instead, and create a new one from scratch. That also would mean we don't need to restore anything.

vvagaytsev: Good idea, we can try it out. Maybe some other tests in this file should be rewritten in such a way.

core/src/config/project.ts
stefreak: My bad, It's actually fine here, because we override the outputs with the original later. We only validate the shape of the declaration is valid here

core/src/config/template-contexts/actions.ts
```diff
@@ -117,16 +109,16 @@ export class ActionReferenceContext extends ConfigContext {
...
-  public sourcePath: string
+  public readonly sourcePath: string
 
   @schema(actionModeSchema)
-  public mode: ActionMode
+  public readonly mode: ActionMode
 
   @schema(joiVariables().required().description("The variables configured on the action.").example({ foo: "bar" }))
-  public var: DeepPrimitiveMap
+  public readonly var: ConfigContext
```
stefreak, line 118:
```suggestion
  public readonly var: VariablesContext
```

core/test/integ/src/plugins/kubernetes/util.ts
```diff
@@ -44,15 +44,124 @@ import type {
...
+      const allPods = await api.core.listNamespacedPod({ namespace })
+
+      const pod = allPods.items[0]
+
+      const pods = await getWorkloadPods({ api, namespace, resource: pod })
+      expect(pods.length).to.equal(1)
+      expect(pods[0].kind).to.equal("Pod")
+      expect(pods[0].metadata.name).to.equal(pod.metadata.name)
+    } finally {
+      garden.close()
```
stefreak: maybe we can do instead 
```
afterEach(() => garden.close())
```
instead of `try { expect().to.be.correct } finally { garden.close() }

vvagaytsev: That has already been addressed in bd8e7a388d1a18a8c6195edfff8249268b25565e.

vvagaytsev (approved):

PR #6750, opened 2025-01-07
feat: allow overriding `source.path` in remote actions (when using `source.repository`)

stefreak:
Allow overriding `source.path` in remote actions (when using `source.repository`).

This is useful when an action source path needs to be a subdirectory in a remote repository.

This feature already was working, I just lifted the limitation in the validation code and added a test, and updated docs.

See also Discord thread: https://discord.com/channels/817392104711651328/1320775997880144005/1320775997880144005

### DIFF
--- a/docs/reference/action-types/Build/exec.md
+++ b/docs/reference/action-types/Build/exec.md
@@ -44,11 +44,9 @@ A description of the action.
 
 By default, the directory where the action is defined is used as the source for the build context.
 
-You can override this by setting either `source.path` to another (POSIX-style) path relative to the action source directory, or `source.repository` to get the source from an external repository.
+You can override the directory that is used for the build context by setting `source.path`.
 
-If using `source.path`, you must make sure the target path is in a git repository.
-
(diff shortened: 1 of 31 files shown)


docs/reference/action-types/Build/exec.md
```diff
@@ -44,11 +44,9 @@ A description of the action.
 
 By default, the directory where the action is defined is used as the source for the build context.
 
-You can override this by setting either `source.path` to another (POSIX-style) path relative to the action source directory, or `source.repository` to get the source from an external repository.
+You can override directory that is used for the build context by setting `source.path`.
```
twelvemo, line 47:
```suggestion
You can override the directory that is used for the build context by setting `source.path`.
```

twelvemo (changes requested):
Looks good! Just one tiny nit on a sentence in the docs.

docs/reference/action-types/Build/exec.md
stefreak: Thank you, I have applied your suggestion ✅

twelvemo (approved):

PR #6751, opened 2025-01-08
fix: resolve parts of template strings, even if another part cannot be resolved yet.

stefreak:

This PR fixes an issue where in certain cases we get errors like "`Could not find key item`" or "`Could not find key inputs`" since `0.13.46`, where the template parser has been rewritten.

Another PR is currently in progress, #6745, which will also fix this issue by capturing contexts instead of partially resolving templates. That PR won't get ready this week anymore though.

This fix is not complete, as the implementtion before 0.13.45 was not either – in certain situations partial resolution isn't cutting it, for example a template string like `${actions.build[input.name].outputs.log}`. Those cases will be solved by #6745

### DIFF
--- a/core/src/config/render-template.ts
+++ b/core/src/config/render-template.ts
@@ -139,6 +139,7 @@ export async function renderConfigTemplate({
     context: templateContext,
     contextOpts: {
       allowPartial: true,
+      legacyAllowPartial: true,
     },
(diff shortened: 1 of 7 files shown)


thsig (approved):
Looks good to me 👍

vvagaytsev (approved):

PR #6753, opened 2025-01-09
ci: replace broken jinja action with sed script

stefreak:

This became necessary due to https://github.com/cuchi/jinja2-action/issues/22 to keep updating our homebrew formula


vvagaytsev (approved):
Looks good! 👍

PR #6759, opened 2025-01-13
fix(core): emit namespaceStatus events during provider init

thsig:
This was fixed by setting up the event forwarding during the initialization of the plugin event bus.

Before this fix, `namespaceStatus` events weren't actually being emitted during provider resolution, since the event bus on the plugin context hadn't been set up to pipe `namespaceStatus` events to the main event bus (this happened later in the flow).

This is the probable cause for some users experiencing dangling namespaces that weren't being cleared up by Cloud's automatic environment cleanup when module/action resolution failed (since the namespace status events that would usually be emitted by the "get status" and "deploy" handlers for those actions weren't being emitted). This still needs to be verified, but most likely fixes the issue.


10ko (approved):
I've tested `main` against this branch and it fixes the issue we had.Good job!
Good job, LGTM!

PR #6760, opened 2025-01-13
fix(cloud): automatically clear locally stored auth token if refresh token is invalid

vvagaytsev:

Fixes #

### DIFF
--- a/core/src/cloud/grow/auth.ts
+++ b/core/src/cloud/grow/auth.ts
@@ -91,7 +91,9 @@ export async function refreshAuthTokenAndWriteToConfigStore(
     }
 
     log.debug({ msg: `Failed to refresh the token.` })
-    if (err.message.toLowerCase().includes("invalid refresh token")) {
+
+    const errHttpStatusCode = err.data?.httpStatus
+    if (errHttpStatusCode === 401) {
       await clearAuthToken(log, globalConfigStore, cloudDomain)
       log.info("Invalid refresh token was removed from the configuration store.")
     }
@@ -100,7 +102,7 @@ export async function refreshAuthTokenAndWriteToConfigStore(
       message: dedent`An error occurred while verifying client auth token with ${getCloudDistributionName(cloudDomain)}: ${err.message}
         Please try again.
         `,
-      responseStatusCode: err.data?.httpStatus,
+      responseStatusCode: errHttpStatusCode,
     })
   }
 }
--- a/core/src/cloud/grow/auth.ts
+++ b/core/src/cloud/grow/auth.ts
@@ -14,8 +14,9 @@ import type { Log } from "../../logger/log-entry.js"
 import { getNonAuthenticatedApiClient } from "./trpc.js"
 import { CloudApiTokenRefreshError } from "../api.js"
 import { CloudApiError } from "../../exceptions.js"
-import { saveAuthToken } from "../auth.js"
+import { clearAuthToken, saveAuthToken } from "../auth.js"
 import { getCloudDistributionName } from "../util.js"
+import dedent from "dedent"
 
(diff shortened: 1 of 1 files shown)


core/src/cloud/grow/auth.ts
```diff
@@ -86,8 +87,15 @@ export async function refreshAuthTokenAndWriteToConfigStore(
     }
 
     log.debug({ msg: `Failed to refresh the token.` })
+    if (err.message.toLowerCase().includes("invalid refresh token")) {
```
stefreak, line 90: Can we also determine the token validity using the response status code?
```suggestion
    if (err.statusCode >= 400 && err.statusCode < 500) {
```

stefreak: Maybe we should even test for 401 explicitly

stefreak: If we ever migrate to other kinds of tokens that need to be decoded, e.g. JWT tokens then we should also delete the token on 400 response though, so I think `400 <= code < 500` is fine

vvagaytsev: Backend V2 did not return any status code for invalid token.

stefreak: We need to change the backend in that case, to make sure that it will emit correct status code in this case then. Maybe @hph can support you on this?

hph: The error returned is `UNAUTHORIZED`, which should map to 401. This status code is available in the `httpStatus` property, at least when using the frontend client, but I think it should be the same in core. [Reference (see error shape above the linked table)](https://trpc.io/docs/server/error-handling#error-codes)

Re. the error itself, currently this can only happen if a) the token doesn't exist in the database or b) the token is expired.

vvagaytsev: I double-check that and did not see the status code in the `TRPCClientError` instance. The `data` property (that should contain `httpStatus`) was `undefined`, I'm not sure why.

vvagaytsev: I found the original error code deep inside the error object.

stefreak: Is it maybe helpful to infer the specific error shape for the router? See also https://trpc.io/docs/client/vanilla/infer-types#infer-trpcclienterror-types

stefreak:
I think it'd be good if we can get this one done before releasing `0.14.0`

vvagaytsev:
Right, let's bring it to the finish line before 0.14.0 release.

vvagaytsev:
@stefreak @hph this is ready for review now.

After the fix made #6964, we do doon have issues with the shape of `TRPCClientError`.
So, we can get the original `httpStatus` and use it to identify auth errors.

core/src/cloud/grow/auth.ts
vvagaytsev: We removed unnecessary error transforming from the TRPC client config in #6964. So, now we can use original httpStatus from a `TRPCClientError` instance.

stefreak (approved):
Nice 👍

core/src/cloud/grow/auth.ts
```diff
@@ -90,9 +91,18 @@ export async function refreshAuthTokenAndWriteToConfigStore(
     }
 
     log.debug({ msg: `Failed to refresh the token.` })
+
+    const errHttpStatusCode = err.data?.httpStatus
+    if (errHttpStatusCode === 401) {
+      await clearAuthToken(log, globalConfigStore, cloudDomain)
+      log.info("Invalid refresh token was removed from the configuration store.")
```
vvagaytsev, line 98: Ah, shouldn't this be printed on as a debug message? 🤔

PR #6768, opened 2025-01-15
fix: do not crash on missing `deploymentRegistry` for in-cluster builds

vvagaytsev:

A `ConfigurationError` is now thrown instead of `InternalError` if the `deploymentRegistry` is not configured on the provider-level when using in-cluster building.

So Garden do not crash and quits gracefully with informative error message.

Fixes #

An attempt to use conditional validation for `deploymentRegistry` of the provider configuration was taken in 250b658cc9c40c4306f92a28bdef7e1b20c09a71.
It was reverted because of the of conditional validation support in the reference docs generation.

### DIFF
--- a/core/src/plugins/kubernetes/config.ts
+++ b/core/src/plugins/kubernetes/config.ts
@@ -95,7 +95,8 @@ interface KubernetesStorage {
   builder: KubernetesStorageSpec
 }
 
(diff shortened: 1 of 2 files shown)


vvagaytsev:
Conditional validation done in 250b658cc9c40c4306f92a28bdef7e1b20c09a71 broke the docs generation. I need to take a closer look at that.

10ko (approved):

PR #6769, opened 2025-01-15
chore: send build report to backend v2

twelvemo:
Run docker build commands with extra flags that output logs in json and collect additional build metrics for a full build report. To parse these logs to human readable output again, we use the progressui binary.

Fixes #


core/src/plugins/container/build.ts
```diff
@@ -250,23 +313,104 @@ async function buildContainerInCloudBuilder(params: {
...
+  try {
+    return JSON.parse(await readFile(filePath, { encoding: "utf-8" }))
+  } catch (e: unknown) {
+    log.debug(`Failed to read docker metadata file: ${e}`)
+    return undefined
+  }
+}
+
+function getBuilderName(dockerMetadata: DockerBuildReport["dockerMetadata"]) {
+  if (dockerMetadata && typeof dockerMetadata?.["buildx.build.ref"] === "string") {
```
vvagaytsev, line 349: The optional check `?.` in `dockerMetadata?.["buildx.build.ref"]` looks redundant, but it's fine to have it.

core/src/plugins/container/build.ts
```diff
@@ -250,23 +313,104 @@ async function buildContainerInCloudBuilder(params: {
...
+          isDefault: false, //TODO
+          driver,
+        },
+      },
+      imageTags,
+      platforms: await getPlatforms(cmdOpts),
+      dockerLogs,
+      dockerMetadata,
+    }
+    const growCloudApi = ctx.cloudApiV2 as GrowCloudApi
```
vvagaytsev, line 406: The `ctx.cloudApiV2` is known to be defined if a user is logged in to the Backend V2.
But, it might be good to also have a defensive check here and print a warning is the `cloudApiV2` is not defined.

vvagaytsev (dismissed):
Looks good! Great work!

I've left a few minor non-blocking comments, those can be addressed in a follow-up PR.

twelvemo:
> Looks good! Great work!
> 
> I've left a few minor non-blocking comments, those can be addressed in a follow-up PR.

Thanks for the feedback. I've addressed them here, if you get a chance to take another look :-)

core/src/plugins/container/build.ts
```diff
@@ -403,7 +402,13 @@ export async function sendBuildReport({
       dockerLogs,
       dockerMetadata,
     }
-    const growCloudApi = ctx.cloudApiV2 as GrowCloudApi
+    const growCloudApi = ctx.cloudApiV2
+    if (!growCloudApi) {
+      // Will be caught below and only result in a warning in the log
+      throw new InternalError({
+        message: "Cloud API v2 not available. Are you logged in?",
+      })
```
vvagaytsev, line 410: The message will be printed on the debug level in the `catch`-block.
We can just print a warning and return here :)

```js
if (!growCloudApi) {
  log.warn("Garden Cloud v2 not available. Are you logged in?")
  return { timeSaved: 0 }
}
```

vvagaytsev: I changed that in d841e266867586a7833cb435427b004f9e149b6b :)

vvagaytsev (approved):

stefreak (commented):
Not completely finished the review, but here are some comments I had pending

stefreak (approved):

PR #6777, opened 2025-01-17
chore: add win binaries code-signing for edge and main releases

10ko:

This PR adds code-signing for Windows binary whenever we release `edge` or a stable release from `main`.

Fixes #


.circleci/continue-config.yml
```diff
@@ -1429,6 +1464,9 @@ workflows:
       - build-dist-linux-windows:
           <<: *only-tags
           requires: [build]
+      - sign-windows-binary:
+          <<: *only-tags
+          requires: [build-dist-linux-windows]
```
twelvemo, line 1469: Do we need to pass the version parameter into this job here?

twelvemo (commented):
This looks good, but had one question.

.circleci/continue-config.yml
```diff
@@ -458,6 +459,40 @@ jobs:
...
+    parameters:
+      version:
+        description: The version tag/name to set
+        type: string
+        default: ""
+    steps:
+      - checkout
+      - *attach-workspace
+      - aws-cli/setup:
+          aws_access_key_id: $AWS_ACCESS_KEY_ID
```
stefreak, line 475: Do you somehow ensure that these environment variables are only available for tag push events?

.circleci/continue-config.yml
```diff
@@ -1429,6 +1464,9 @@ workflows:
       - build-dist-linux-windows:
           <<: *only-tags
           requires: [build]
+      - sign-windows-binary:
```
stefreak, line 1467: I think as `github-release-tag` actually uplaods the binaries to the github release, and it depends on `build-dist-linux-windows`, the `sign-windows-binary` will simply run in parallel and have no effect.

Either we need to change `github-release-tag` to depend on `sign-windows-binary`, or maybe it would be even better to change the script that we run to build the binaries to also sign them, if the AWS credentials are available (`npm run dist`).

the `npm run dist` script also creates checksum files, so it would probably be necessary to sign binaries before generating the checksums. That way you can also test it locally end-to-end

.circleci/continue-config.yml
10ko: Yes, it's needed for renaming the files to avoid collisions.

twelvemo: What i meant was more that i can't see where a value for that param is passed in?

.circleci/continue-config.yml
twelvemo: Secrets are not passed to forked PRs, i believe this is all the granularity you can get.

.circleci/continue-config.yml
10ko: The problem with `npm run dist` is that, if I read it correctly, it will run at every commit and I'd like to avoid signing the binaries at every PR commit.
I think it would be ok to change the dependency of `github-release-tag` to depend on the new job and recalculate the md5 checksum (which I missed, good catch)?

.circleci/continue-config.yml
10ko: You are right, this won't work, I'll fix it.

.circleci/continue-config.yml
stefreak: What we need to do is make sure the AWS credentials are only available in the git push tasks. 

The script can skip signing if credentials are missing.

stefreak: Re-calculating is also an option, it might be a little confusing if someone in the future tries to understand the pipeline though.

stefreak: I'm happy to pair with you on this, let me know

.circleci/continue-config.yml
stefreak: CircleCI Contexts can have expression-based restrictions, see also:

> Expression based context restrictions allow you to specify rules like pipeline.git.branch == "main". For more information on context restrictions see our [documentation](https://circleci.com/docs/contexts/#expression-restrictions).

stefreak: IMO we need to create a "code-signing" context, and apply an expression-based restriction that only makes the context with the AWS credentials available in the main branch as well as Tag push events.

10ko: Aha, that looks like it should work on our case.
Thanks, TIL!

scripts/sign-windows-binary.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
+
```
stefreak, line 2:
```suggestion
set -e -o pipefail
```

stefreak (dismissed):
Nice work!

scripts/sign-windows-binary.sh
```diff
@@ -0,0 +1,94 @@
...
+aws s3 cp $file_folder/$file_name s3://$s3_bucket/ --profile $aws_profile
+
+# Check if signed file exists
+check_for_signed_file=1
+while (( check_for_signed_file == 1 )); do
+  echo "Checking if file signed/$file_name exists"
+  file_present=$(aws s3api head-object --bucket $s3_bucket --key "signed/$file_name" --profile $aws_profile > /dev/null 2>&1; echo $?)
+
+  if [ $file_present == 0 ]; then
+    check_for_signed_file=0
```
stefreak, line 80: FYI, bash has break :) 

```suggestion
while true; do
  echo "Checking if file signed/$file_name exists"
  file_present=$(aws s3api head-object --bucket $s3_bucket --key "signed/$file_name" --profile $aws_profile > /dev/null 2>&1; echo $?)

  if [ $file_present == 0 ]; then
    break
```

scripts/sign-windows-binary.sh
```diff
@@ -0,0 +1,94 @@
...
+  else
+    sleep 1s
```
stefreak, line 82: Usually sleep does not accept units
```suggestion
    sleep 1
```

This is what happened on my mac when I tested with unit: 
```
bash-3.2$ sleep 1s
usage: sleep seconds
```

10ko: Weird, I tried on my machine but it was probably working because I didn't set the `set -e`

stefreak (approved):

PR #6797, opened 2025-01-23
fix: when using create project, the project name is undefined

10ko:

Found this little bug which was causing the `garden create project` command to render a `project.garden.yaml` file with an `undefined` project name.

Fixes #


vvagaytsev (approved):
Good catch!

PR #6802, opened 2025-01-24
improvement: add better error handling when failing to archive test artifacts

10ko:

Improve error handling when trying to create the archive of artifacts. 
Additionally, increases the timeout for the `tar` command to 10 minutes.

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/run.ts
+++ b/core/src/plugins/kubernetes/run.ts
@@ -582,7 +582,7 @@ async function runWithArtifacts({
         log,
         stdout,
         stderr,
-        // Anything above two minutes for this would be unusual
+        // Anything above 10 minutes for this would be unusual
         timeoutSec: 600,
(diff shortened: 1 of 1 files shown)


vvagaytsev (approved):

core/src/plugins/kubernetes/run.ts
```diff
@@ -582,29 +582,22 @@ async function runWithArtifacts({
         log,
         stdout,
         stderr,
-        // Anything above two minutes for this would be unusual
-        timeoutSec: 120,
+        // Anything above 10 minutes for this would be unusual
+        timeoutSec: 600,
```
stefreak, line 586: In my opinion, 10 minutes is way too much, two minutes was already a lot for running tar --help.

stefreak (approved):
Looking great! I wouldn't change the timeout though

PR #6810, opened 2025-01-28
fix(cluster-buildkit): assume image needs rebuild if skopeo command fails and print a warning

stefreak:
Instead of failing completely, assume image needs to be rebuilt and print a warning. 

This is meant as a pain relieve for the issue encountered in this Discord thread: https://discord.com/channels/817392104711651328/1313997959121211463/1313997959121211463

This is not a full solution for the problem encountered in the thread.

### DIFF
--- a/core/src/plugins/kubernetes/container/build/common.ts
+++ b/core/src/plugins/kubernetes/container/build/common.ts
@@ -13,7 +13,7 @@ import type { KubernetesConfig, KubernetesPluginContext, KubernetesProvider } fr
 import { PodRunner, PodRunnerError, PodRunnerTimeoutError } from "../../run.js"
 import type { PluginContext } from "../../../../plugin-context.js"
 import { hashString, sleep } from "../../../../util/util.js"
-import { ConfigurationError, RuntimeError } from "../../../../exceptions.js"
+import { ConfigurationError } from "../../../../exceptions.js"
 import type { Log } from "../../../../logger/log-entry.js"
 import { prepareDockerAuth } from "../../init.js"
 import { prepareSecrets } from "../../secrets.js"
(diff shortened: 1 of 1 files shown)


vvagaytsev (approved):

vvagaytsev (changes requested):
There is one compilation error to be fixed.

core/src/plugins/kubernetes/container/build/common.ts
```diff
@@ -285,12 +285,13 @@ export async function skopeoBuildStatus({
...
 
-      throw new RuntimeError({
-        message: `Unable to query registry for image status: Command "${skopeoCommand.join(
+      log.warn(`Failed to check if the image has already been built: Command "${skopeoCommand.join(
           " "
-        )}" failed: ${err.message}`,
-        wrappedErrors: [err],
-      })
+        )}" failed: ${err.message}`)
+      log.debug({ err })
```
vvagaytsev, line 291:
```suggestion
log.debug(() => `Error details: ${err}`)
```

vvagaytsev (dismissed):

core/src/plugins/kubernetes/container/build/common.ts
```diff
@@ -285,12 +285,15 @@ export async function skopeoBuildStatus({
...
-        message: `Unable to query registry for image status: Command "${skopeoCommand.join(
+      log.warn(
+        `Failed to check if the image has already been built: Command "${skopeoCommand.join(
           " "
-        )}" failed: ${err.message}`,
-        wrappedErrors: [err],
-      })
+        )}" failed: ${err.message}`
+      )
+      log.debug(() => `Error details: ${err}`)
```
stefreak, line 293: I actually wanted that the stack trace is visible in the debug logs; Don't we accept an `error` parameter for structured logs?

vvagaytsev (approved):

PR #6812, opened 2025-01-30
fix: broken path inside window binary zipped file

10ko:

Fixes #

### DIFF
--- a/.circleci/continue-config.yml
+++ b/.circleci/continue-config.yml
@@ -488,7 +488,9 @@ jobs:
           name: Prepare the windows binary zip and cleanup
           command: |
             rm dist/garden-<<parameters.version>>-windows-amd64.zip
-            zip -r dist/garden-<<parameters.version>>-windows-amd64.zip dist/windows-amd64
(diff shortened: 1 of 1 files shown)


.circleci/continue-config.yml
```diff
@@ -488,7 +488,9 @@ jobs:
           name: Prepare the windows binary zip and cleanup
           command: |
             rm dist/garden-<<parameters.version>>-windows-amd64.zip
-            zip -r dist/garden-<<parameters.version>>-windows-amd64.zip dist/windows-amd64
+            cd dist
+            zip -r garden-<<parameters.version>>-windows-amd64.zip windows-amd64
+            cd ..
```
10ko, line 493: Adding this because tbh I cannot figure out if CircleCI keeps the path or not when moving to the following step, which then persists the workspace

stefreak: It won't affect the following step, but I think this is less confusing anyway 👍

stefreak (approved):

PR #6814, opened 2025-01-31
feat: allow cross-referencing variables in the same scope

### LINKED ISSUE #2730: [FEATURE]: Enable the `var` template key in template strings for `variables`
## Feature Request

Currently, the `var`/`variables` template key is not available when resolving template keys in `variables`. So, for example, something like this is currently not possible:
```yaml
# project config
...
variables:
  baseDomain: example.com # We want to reuse this prefix
  apiHostname: "api.${var.baseDomain}"
  webhookHostname: "webhooks.${var.baseDomain}"
```
Supporting this would require us to resolve `var` references inside `variables` blocks in dependency order.

### Background / Motivation

Supporting this would enable users to further reduce repetition in their configs.

When resolving a variables block (e.g. for the project, or for an environment or module), construct a simple dependency graph of variable references (indexed by the variable key).

Then, resolve the variables in dependency order and incrementally populate the key/value pairs available under the `var` key and use them to resolve template strings further up the reference graph.

There should also be circular dependency detection with informative error messages.

stefreak:
Allows cross-referencing variables in the same scope.

Fixes #2730 (and some regressions of variable lookup behaviour introduced in #6745, there is no need to mention them in the release notes because we have not released it so far)

There are some caveats that are covered in the test cases, to maintain backwards-compatibility with older versions of 0.13.

### DIFF
--- a/core/test/unit/src/config/template-contexts/variables.ts
+++ b/core/test/unit/src/config/template-contexts/variables.ts
@@ -12,6 +12,7 @@ import type { TestGarden } from "../../../../helpers.js"
 import { getDataDir, makeTestGarden } from "../../../../helpers.js"
 import { TestContext } from "./base.js"
 import { deepResolveContext } from "../../../../../src/config/template-contexts/base.js"
+import { resolveAction } from "../../../../../src/graph/actions.js"
 
 /*
  * Copyright (C) 2018-2024 Garden Technologies, Inc. <info@garden.io>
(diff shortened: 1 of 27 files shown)


core/test/unit/src/config/template-contexts/variables.ts
```diff
@@ -0,0 +1,310 @@
...
+import { deepResolveContext } from "../../../../../src/config/template-contexts/base.js"
+import { resolveAction } from "../../../../../src/graph/actions.js"
+
+/*
+ * Copyright (C) 2018-2024 Garden Technologies, Inc. <info@garden.io>
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
```
vvagaytsev, line 23: Duplicate license header can be removed :)

vvagaytsev (approved):

PR #6822, opened 2025-02-05
fix: validate secrets before action resolution

vvagaytsev:

 Validate missing secrets when instantiating a `ResolveActionTask` , rather than when adding configs.
 This would ensure we only get the error when missing secrets would actually cause a failure for the given command, while still failing relatively early even if the action happens to use another action’s outputs.

Fixes #


core/test/unit/src/commands/workflow.ts
```diff
@@ -544,6 +545,53 @@ describe("RunWorkflowCommand", () => {
...
+      source: { path: [] },
+    }) as WorkflowConfig[]
+
+    garden.setRawWorkflowConfigs(parsedConfigs)
+
+    await expectError(() => cmd.action({ ...defaultParams, args: { workflow: name } }), {
+      contains: [
+        "The following secret names were referenced in configuration, but are missing from the secrets loaded remotely",
+        `Workflow ${name}: missing`,
+        "No secrets have been loaded. If you have defined secrets for the current project and environment in Garden Cloud, this may indicate a problem with your configuration.",
```
stefreak, line 590: We should only get this part of the error if we're logged in. Not sure, are we logged in in this test?

core/test/unit/src/commands/workflow.ts
```diff
@@ -544,6 +545,53 @@ describe("RunWorkflowCommand", () => {
...
+      },
+    ]
+    // @ts-expect-error todo: correct types for unresolved configs
+    const parsedConfigs = parseTemplateCollection({
+      // @ts-expect-error todo: correct types for unresolved configs
+      value: configs,
+      source: { path: [] },
+    }) as WorkflowConfig[]
+
+    garden.setRawWorkflowConfigs(parsedConfigs)
```
stefreak, line 584: For workflows, we should only get the error if we're actually executing the workflow (Using the `workflow` command)

core/test/unit/src/tasks/resolve-action.ts
```diff
@@ -47,6 +47,34 @@ describe("ResolveActionTask", () => {
...
+
+      expect(garden.secrets).to.be.empty
+      expect(garden.isLoggedIn()).to.be.false
+
+      const task = await getTask("Run", "run-with-missing-secrets")
+      expectError(() => getTask("Run", "run-with-missing-secrets"), {
+        contains: [
+          "The following secret names were referenced in configuration, but are missing from the secrets loaded remotely",
+          "Run run-with-missing-secrets: MISSING",
+          "No secrets have been loaded. If you have defined secrets for the current project and environment in Garden Cloud, this may indicate a problem with your configuration.",
```
stefreak, line 71: In this case we are clearly not logged in, so the error message is very confusing.

Better would be, if you are not logged in:
```suggestion
          "You are not logged in. Log in to get access to Secrets in Garden Cloud",
```

If you are logged in & loading secrets succeeded:
```suggestion
          "These secrets are missing in the project "<project name>" under environment "<env>" in Garden Cloud. Follow this link to manage secrets: <link>. Available secrets: xx, xx, xx",
```

core/test/unit/src/garden.ts
```diff
@@ -3328,6 +3323,35 @@ describe("Garden", () => {
...
+          expect.fail("Expected scanAndAddConfigs not to throw")
+        }
+      })
+
+      it("should not throw when a workflow config references missing secrets", async () => {
+        const garden = await makeTestGarden(getDataDir("missing-secrets", "workflow"))
+        try {
+          await garden.scanAndAddConfigs()
+        } catch (err) {
+          expect.fail("Expected scanAndAddConfigs not to throw")
```
stefreak, line 3351: These are negative tests, where we ensure that we are not throwing during `scanAndAddConfigs`. I like that! 

We also need tests to ensure we are throwing calling the `resolveAction` helper, and instantiating the `ResolveActionTask`.

stefreak (commented):
Great work! 👍  I had a preliminary look and already sharing some things that I noticed

core/test/unit/src/garden.ts
vvagaytsev: That test was added in 0b1096706e4adf7950793aabd0d2415b1f294fc6.

core/test/unit/src/commands/workflow.ts
vvagaytsev: Right, the workflow is executed below, in the L586. Here we just define a workflow configuration.

core/test/unit/src/garden.ts
stefreak: Ah amazing, I overlooked that 👍

core/test/unit/src/commands/workflow.ts
stefreak: Ah sorry, I overlooked the cmd.action() call – my bad! great work 👍

core/test/unit/src/tasks/resolve-action.ts
vvagaytsev: Thank you, good point!

I partially addressed these comments. Now we do print "You are not logged in. Log in to get access to Secrets in Garden Cloud." statement in the message footer if a user is not logged in.
The login status is propagated to the `throwOnMissingSecretKeys` and it's internal functions.

I did not propagated project config yet, that would be required to print the project and env info. Would it be better to do that in this PR?

core/test/unit/src/commands/workflow.ts
vvagaytsev: I updated the error message footers in 4ed41fafb74014d87a14711b463cfe77138619d5. Now the "You are not logged in. Log in to get access to Secrets in Garden Cloud." message is printed if a user is not logged in.

core/test/unit/src/tasks/resolve-action.ts
stefreak: No, this is already way better than it was before 👍 That's amazing

core/src/config/secrets.ts
```diff
@@ -17,31 +17,32 @@ import difference from "lodash-es/difference.js"
...
+  if (!isLoggedIn) {
+    return "You are not logged in. Log in to get access to Secrets in Garden Cloud."
   }
 
-  if (allMissing.length === 0) {
-    return
+  if (loadedKeys.length === 0) {
+    return deline`
+      Note: No secrets have been loaded. If you have defined secrets for the current project and environment in Garden
+      Cloud, this may indicate a problem with your configuration.
```
stefreak, line 28: This is still extremely confusing. I'd suggest the following, to improve the user experience:
```suggestion
      Note: You can manage secrets in Garden Cloud. No secrets have been defined for the current project and environment. See also https://cloud.docs.garden.io/features/secrets
```

core/src/config/secrets.ts
```diff
@@ -17,31 +17,32 @@ import difference from "lodash-es/difference.js"
...
-) {
-  const allMissing: [string, ContextKeySegment[]][] = [] // [[key, missing keys]]
-  for (const config of configs) {
-    const missing = detectMissingSecretKeys(config, context, secrets)
-    if (missing.length > 0) {
-      allMissing.push([config.name, missing])
-    }
+function getMessageFooter({ loadedKeys, isLoggedIn }: { loadedKeys: string[]; isLoggedIn: boolean }) {
+  if (!isLoggedIn) {
+    return "You are not logged in. Log in to get access to Secrets in Garden Cloud."
```
stefreak, line 22:
```suggestion
    return "You are not logged in. Log in to get access to Secrets in Garden Cloud. See also https://cloud.docs.garden.io/features/secrets"
```

stefreak (approved):
This is amazing! Thank you for looking into this.

I have additional suggestions regarding the error messages, those are totally fine as follow-up or in this PR as well, as you wish!

core/test/unit/src/template-string.ts
```diff
@@ -3025,8 +3033,16 @@ describe("throwOnMissingSecretKeys", () => {
       source: { path: [] },
     } as const)
 
-    throwOnMissingSecretKeys(configs, new TestContext({}), {}, "Module")
-    throwOnMissingSecretKeys(configs, new TestContext({}), { someSecret: "123" }, "Module")
+    // The `isLoggedIn` flag affects the error message and has no effect for this test.
+    // Let's set it to true.
+    throwOnMissingSecretKeys({ configs, context: new TestContext({}), secrets: {}, prefix: "Module", isLoggedIn: true })
```
stefreak, line 3038: Minor nit, this statement and the next statement look redundant 👍

core/src/config/secrets.ts
vvagaytsev: Thanks! I've added the link to the guide in fe5f453de226b6e67e756c1f5ace540f7dede841.

core/src/config/secrets.ts
vvagaytsev: Sure, updated this in 5d676e36db6d465cac52397f171453f1849c09be.

PR #6828, opened 2025-02-08
improvement(terraform): support dynamic backends

eysi09:

This change introduces a `backendConfig` field on the Terraform
provider and action types.

This field allows users to dynamically set the backend config for
their Terraform stacks.

Terraform itself doesn't allow variables in the backend config
so this can be used instead to dynamically set the backend
based on e.g. the environment.

Fixes #

I still need to update the docs on the Terraform provider and remove the "`sed` hack". It's also an open question whether we mark this as experimental for now. 

But I can address both in a follow up PR.

### DIFF
--- a/plugins/terraform/src/helpers.ts
+++ b/plugins/terraform/src/helpers.ts
@@ -7,19 +7,48 @@
  */
 
 import { resolve } from "path"
-import { mapValues, startCase } from "lodash-es"
+import { isPlainObject, mapValues, startCase } from "lodash-es"
 
-import { ConfigurationError, PluginError, RuntimeError } from "@garden-io/sdk/build/src/exceptions.js"
+import {
+  ConfigurationError,
+  isErrnoException,
+  isErrorWithMessage,
+  PluginError,
+  RuntimeError,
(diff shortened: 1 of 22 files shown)


plugins/terraform/src/helpers.ts
```diff
@@ -46,67 +76,152 @@ interface TerraformParamsWithVariables extends TerraformParamsWithWorkspace {
...
+    )
+    throw err
+  }
+
+  const currentBackendConfig = tfState?.backend?.["config"] as { [key: string]: string } | undefined
+
+  const hasChanged =
+    (currentBackendConfig &&
+      isPlainObject(currentBackendConfig) &&
+      !Object.entries(backendConfig).every(([key, value]) => {
```
thsig, line 132: Just out of curiosity, is there a reason you're not using something like `isEqual` from `lodash-es` for the object comparison here?

thsig (approved):
Good stuff!

thsig (dismissed):

thsig (approved):

twelvemo (approved):
Great stuff! This is going to be super useful!

PR #6829, opened 2025-02-09
improvement(terraform): enable streaming logs to cloud

eysi09:

This commit introduces a config option to optionally stream Terraform
logs to Garden Cloud/Enterprise.

Defaults to `false` but we may reconsider in a later breaking change
release.

Fixes #


thsig (approved):

twelvemo (approved):
I think we could always stream them, if there are sensitive values they can be hidden with the `sensitive` flag in terraform. But either way happy to see this!

PR #6832, opened 2025-02-10
chore(k8s): deprecate `configmap` and `persistentvolumeclaim` deploy actions

vvagaytsev:

Fixes #

If a migration guide is necessary, it can be written in a follow-up PR. We have a project task to compose a migration guide, so the guide will not be missed.

### DIFF
--- a/core/src/util/deprecations.ts
+++ b/core/src/util/deprecations.ts
@@ -40,7 +40,7 @@ function makeDeployActionTypeDeprecation(actionType: DeprecatedDeployActionType,
   return {
     contextDesc: "Garden action types",
     featureDesc: `The ${style(`${actionType} Deploy`)} action type`,
-    hint: "Do not use this deploy action type in Garden 0.14.",
+    hint: "Do not use this action in Garden 0.14.",
     hintReferenceLink: null,
   }
 }
--- a/core/src/util/deprecations.ts
+++ b/core/src/util/deprecations.ts
@@ -34,6 +34,17 @@ function makePluginDeprecation(pluginName: DeprecatedPluginName, style: (s: stri
   }
 }
 
+export type DeprecatedDeployActionType = "configmap" | "persistentvolumeclaim"
+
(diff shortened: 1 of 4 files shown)


core/src/util/deprecations.ts
```diff
@@ -34,6 +34,17 @@ function makePluginDeprecation(pluginName: DeprecatedPluginName, style: (s: stri
   }
 }
 
+export type DeprecatedDeployActionType = "configmap" | "persistentvolumeclaim"
+
+function makeDeployActionTypeDeprecation(actionType: DeprecatedDeployActionType, style: (s: string) => string) {
+  return {
+    contextDesc: "Garden action types",
+    featureDesc: `The ${style(`${actionType} Deploy`)} action type`,
+    hint: "Do not use this deploy action type in Garden 0.14.",
```
eysi09, line 43:
```suggestion
    hint: "Do not use this action in Garden 0.14.",
```

eysi09 (approved):

docs/guides/deprecations.md
```diff
@@ -146,4 +146,14 @@ For more information, please refer to the [Migration guide for action configs](#
...
 
-Do not use ``GARDEN_LEGACY_BUILD_STAGE`` environment variable in 0.14.
\ No newline at end of file
+Do not use ``GARDEN_LEGACY_BUILD_STAGE`` environment variable in 0.14.
+
+## Garden action types
+
+<h3 id="configmapDeployAction">The `configmap Deploy` action type</h3>
+
+Do not use this action in Garden 0.14.
```
stefreak, line 155: @eysi09 @vvagaytsev I think we want to nudge users to stop using these features in `0.13` already, so I think we should simply recommend to stop using this action type. We must provide some guidence what the alternative would be here.

```patch
-Do not use this action in Garden 0.14.
+Use the `kubernetes` deploy action type instead.
```

vvagaytsev: Good point! Addressed that in #6876

PR #6833, opened 2025-02-10
chore: deprecate `build` config field in runtime actions

vvagaytsev:

The `build` config field is not recommended to use since #5700 and #6207.
This PR officially deprecates that config field.

Fixes #

### DIFF
--- a/core/src/actions/base.ts
+++ b/core/src/actions/base.ts
@@ -287,7 +287,7 @@ export const baseRuntimeActionConfigSchema = createSchema({
       )
       .meta({
         templateContext: ActionConfigContext,
-        deprecated: makeDeprecationMessage({ deprecation: "buildConfigFieldOnRuntimeActions" }),
+        deprecated: makeDeprecationMessage({ deprecation: "buildConfigFieldOnRuntimeActions", includeLink: true }),
       }),
   }),
   extend: baseActionConfigSchema,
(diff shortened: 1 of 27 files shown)


core/src/actions/base.ts
```diff
@@ -696,6 +701,27 @@ export abstract class RuntimeAction<
...
 > extends BaseAction<C, StaticOutputs, RuntimeOutputs> {
+  constructor(params: ActionWrapperParams<C>) {
+    super(params)
+
+    const buildName = this.getConfig("build")
+    if (buildName) {
+      const log = RootLogger.getInstance().createLog()
+      const config = params.config
+      // Report concrete action name for better UX
+      log.warn(
```
eysi09, line 712: Two asks:

- Can we show the path to the config file where this action is?
- Can we add a short doc on how to migrate away from the `build` field and link to it here?

eysi09 (changes requested):
Had one comment, otherwise looks good.

core/src/actions/base.ts
vvagaytsev: Thanks for the comments! Good points!

* Sure, we can! I added that in f834be6ed0bf56d0484e477318f94d46e9a498fa
* Yes, that would be better to have a short guide, I added it as a section in the deprecations guide in c74067f3bd1ec8d70b198e32966a858708a03d31, b3772f736e85d9efba9b9a28d46b1e453d09e2f9 and a9f8a182d666fde23812a2c8a0e4f0e40972232e

eysi09 (approved):
Great stuff!

PR #6840, opened 2025-02-11
chore(container): deprecate `container Deploy` action type

vvagaytsev:

Fixes #


core/src/constants.ts
```diff
@@ -39,9 +39,24 @@ export enum GardenApiVersion {
...
+ *
+ * TODO(0.14): change this to v2
+ */
+export const defaultGardenApiVersion = GardenApiVersion.v1
+
+/**
+ * The version to be used if the project-level `apiVersion` is not defined
+ * in the Garden config file.
+ *
+ * TODO(0.14): bump this to v1
```
stefreak, line 56: if not defined, we always must assume v0 (as that is an indicator for v0 configs, as when using actions, specifying the field is mandatory)

vvagaytsev:
The original task will be addressed in a different way.
Closing this for now, but keeping the branch, because some changes might be re-used in the new approach.

PR #6844, opened 2025-02-12
fix(template): parse template strings in varfiles

### LINKED ISSUE #6843: [Bug]: Templating in sync.paths[].mode not working since 0.13.53
### Current behavior

Given following YAML:
```
sync:
  paths:
    - source: .
      target: /app/
      mode: ${var.syncmode || 'one-way-safe'}
```

produces following output when running `deploy service`

```
Error validating spec for Deploy type=container name=service (../service):

spec.sync.paths[0][mode] must be one of [one-way, one-way-safe, one-way-replica, one-way-reverse, one-way-replica-reverse, two-way, two-way-safe, two-way-resolved]
```

### Expected behavior

It evaluates the expresion and uses the variable `syncmode` if given, otherwise the default `'one-way-safe'`

### Reproducible example

s.a.

### Workaround

Use no expresions:

```
sync:
  paths:
    - source: .
      target: /app/
      mode: 'one-way-safe'
```

### Suggested solution(s)

Revert or fix changes made in 0.13.53

### Additional context

Works as expected untill 0.13.52.

### Your environment

garden version: 0.13.53

vvagaytsev:

A follow-up fix for #6745

Fixes #6843

### DIFF
--- a/core/src/config/template-contexts/variables.ts
+++ b/core/src/config/template-contexts/variables.ts
@@ -30,6 +30,7 @@ import { ActionSpecContext } from "./actions.js"
 import type { CustomCommandContext } from "./custom-command.js"
 import type { CommandResource } from "../command.js"
 import type { GroupConfig } from "../group.js"
+import { parseTemplateCollection } from "../../template/templated-collections.js"
 
 export class VariablesContext extends LayeredContext {
   /**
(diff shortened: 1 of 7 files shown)


thsig (approved):
Looks good to me!

core/src/config/template-contexts/variables.ts
```diff
@@ -145,11 +149,15 @@ export class VariablesContext extends LayeredContext {
...
   ) {
-    const envVarfileVars = await loadVarfile({
+    const rawEnvVarfileVars = await loadVarfile({
       configRoot: projectConfig.path,
       path: environmentConfig.varfile,
       defaultPath: defaultEnvVarfilePath(environment),
     })
+    const envVarfileVars = parseTemplateCollection({
+      value: rawEnvVarfileVars,
+      source: { path: ["varfile"], yamlDoc: projectConfig.internal.yamlDoc },
```
stefreak, line 159: @thsig @vvagaytsev  the source path and yamlDoc are incorrect; This will lead to confusing error messages when varfiles contain invalid template strings.

The minimum correct value here is `source: { path: [] }` if it's not yaml; to get correct source mapping for yaml varfiles we need to call `parseTemplateCollection` in `loadVarfile`, so we can pass the yaml document.

vvagaytsev: Good catch! Thanks for fixing this in the follow-up PR.

PR #6845, opened 2025-02-12
fix: mute false-positive deprecation warning for `deploymentStrategy`

vvagaytsev:

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/container/deployment.ts
+++ b/core/src/plugins/kubernetes/container/deployment.ts
@@ -79,7 +79,7 @@ export const k8sContainerDeploy: DeployActionHandler<"deploy", ContainerDeployAc
     })
   }
 
-  if (deploymentStrategy) {
+  if (deploymentStrategy === "blue-green") {
     reportDeprecatedFeatureUsage({
       apiVersion: ctx.projectApiVersion,
       log,


eysi09 (approved):

core/src/plugins/kubernetes/container/deployment.ts
```diff
@@ -79,7 +79,7 @@ export const k8sContainerDeploy: DeployActionHandler<"deploy", ContainerDeployAc
     })
   }
 
-  if (deploymentStrategy) {
+  if (deploymentStrategy === "blue-green") {
```
stefreak, line 82: Isn't the deploymentStrategy itself deprecated and will be removed? If no, then we also need to update the docs. @eysi09 @vvagaytsev

vvagaytsev: Right! The config field itself is also deprecated and will be removed. So, we should restore the old condition and nudge on any usage of this field in 0.13.

vvagaytsev: I'll make a PR to address that.

vvagaytsev: I just had to remove default value in the joi schema definition to avoid false-positive warnings :)

PR #6849, opened 2025-02-13
fix(template): partially resolve template strings while action config preprocessing

vvagaytsev:

Restores [partial action config resolution](https://github.com/garden-io/garden/commit/c968c616#diff-87efeddad61156c502b8d97ff8a0666e31a9419dbf108b2792fc44f5dbde3946L840-L851) from 0.13.52 which was removed in #6745.

The action config can contain some if/else/then structural operators, or other conditions.

We need to evaluate such conditions here if possible, otherwise the dependency detection may bring false-positive results.

<details>
<summary>Click to see reproducible example</summary>

```yml
apiVersion: garden.io/v1
kind: Project
name: structural-operators
defaultEnvironment: development

environments:
  - name: development
  - name: production

providers:
  - name: exec
    environments: [ development, production ]
  - name: terraform
    environments: [ development, production ]

---
kind: Run
name: run-task
type: exec
environments:
  - production
  - development
#dependencies:
#  $if: ${environment.name == 'production'}
#  $then:
#    - deploy.tf-production
spec:
  $if: ${environment.name != 'production'}
  $then:
    command: [ "sh", "-c", "echo 'Hello from garden-remote in development'" ]
  $else:
    # this should not be evaluated if the branch condition is false,
   # and sould not be considered as dependency
    command: [ "sh", "-c", "echo '${runtime.services.tf-production.outputs.welcome_message}'" ]

---
kind: Deploy
type: terraform
name: tf-production
environments:
  - production
spec:
  root: .

```

</details>

Fixes regression introduced in #6745. 

### DIFF
--- a/core/src/config/template-contexts/actions.ts
+++ b/core/src/config/template-contexts/actions.ts
@@ -76,6 +76,8 @@ export class ActionConfigContext extends TemplatableConfigContext {
   }
 }
 
+export const actionConfigContextKeys = Object.freeze(Object.keys(ActionConfigContext.getSchema().describe()["keys"]))
+
 interface ActionReferenceContextParams {
   name: string
   disabled: boolean
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -757,27 +757,26 @@ export const preprocessActionConfig = profileAsync(async function preprocessActi
       context: builtinFieldContext,
       opts: {},
     })
+    config = { ...config, ...resolvedBuiltin }
+
     const { spec = {} } = config
 
-    config = {
-      ...config,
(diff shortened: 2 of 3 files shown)


stefreak:
I'm wondering if we actually need to partially resolve template strings here, or if we need to have a capability to detect dead code when walking the AST, so we understand which references are actually relevant in the given context. I'll investigate a little bit if we can change how the AST visiting code works to enable that, instead of complicating how we resolve actions

stefreak:
Thanks for identifying the issue @vvagaytsev + great work on this attempt of solving the problem! I found another way to solve the issue more generically by detecting dead code branches, which also works for ternary epxressions and not just for structural `$if` expressions; So this PR has been superseded #6862

PR #6850, opened 2025-02-13
fix(dev): fix unresolved templates in cmd results

thsig:

We were including the raw graph results in the results of commands requested over HTTP from the live page. This was always unnecessary (and the data structures involved were excessively large), but only started resulting in errors with our recent templating improvements (previously, we'd just leave the unresolved template strings in there).

Fixes live-mode commands run from the Cloud UI to fail under certain conditions.

### DIFF
--- a/core/src/actions/base.ts
+++ b/core/src/actions/base.ts
@@ -54,7 +54,7 @@ import type {
   ResolvedAction,
   ResolvedActionWrapperParams,
 } from "./types.js"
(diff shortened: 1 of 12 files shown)


eysi09 (dismissed):

thsig:
There are some test failures here. I'll take a look.

eysi09 (approved):

PR #6862, opened 2025-02-18
fix: do not consider implicit dependencies from action references in dead code branches

stefreak:
This PR changes `getContextLookupReferences` to ignore dead code
branches

Fixes an issue where we considered implicit dependencies from action references in dead code branches.

**Addititional info**:

There was an alternate solution attempt, here: https://github.com/garden-io/garden/pull/6849

One of the benefits of the approach of detecting dead branches is that it now also works with logical AND (`&&`), logical OR (`||`), ternary expressions (`? :`) and if block expressions (`${if ...}`) and not just with structural if operators (`$if:` yaml keys).

Example / minimal repro:

```
kind: Project
apiVersion: garden.io/v1
name: test
environments:
 - name: local
variables:
 disabled: ${true}
---
kind: Run
type: exec
name: myrun
spec:
  command: [echo, "${ !var.disabled ? actions.deploy.mydeploy.outputs.logs : '<not executed>'}"]
---
kind: Deploy
type: exec
name: mydeploy
disabled: ${var.disabled}
spec:
  deployCommand: [echo, "${var.disabled ? 'I should not run' : 'I should run'}"]
```

Command: `garden run myrun`
Expected behaviour: the deploy "mydeploy" should not be executed.

The docs for the `disabled` flag do not reflect the current behaviour where we execute disabled actions, if another action references one of their outputs. This is a separate issue that will be dealt with in a separate PR.

### DIFF
--- a/core/src/template/ast.ts
+++ b/core/src/template/ast.ts
@@ -7,6 +7,7 @@
  */
 
 import { isArray, isNumber, isString } from "lodash-es"
+import type { ConfigContext, ContextResolveOpts } from "../config/template-contexts/base.js"
 import { ContextResolveError, getUnavailableReason, renderKeyPath } from "../config/template-contexts/base.js"
 import { ConfigurationError, InternalError } from "../exceptions.js"
 import { getHelperFunctions } from "./functions/index.js"
@@ -14,7 +15,7 @@ import type { EvaluateTemplateArgs } from "./types.js"
 import { isTemplatePrimitive, type TemplatePrimitive } from "./types.js"
 import type { Collection, CollectionOrValue } from "../util/objects.js"
 import { type ConfigSource, validateSchema } from "../config/validation.js"
-import type { TemplateExpressionGenerator } from "./analysis.js"
+import type { Branch } from "./analysis.js"
 import { TemplateStringError } from "./errors.js"
 import { styles } from "../logger/styles.js"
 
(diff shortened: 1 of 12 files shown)


stefreak:
Alternative approach for #6849

core/src/template/ast.ts
```diff
@@ -605,6 +640,31 @@ export class IfBlockExpression extends AbstractBlockExpression {
...
+  ): TemplateExpression[] {
+    const condition = this.condition.evaluate({ context, opts, yamlSource, optional: true })
+
+    // if the condition fails, we cannot determine the branches yet.
+    // In that case we consider all branches active.
+    if (typeof condition === "symbol") {
+      return this.getChildren()
+    }
+
+    const activeBranch = isTruthy(condition) ? this.consequent : this.alternate
```
twelvemo, line 660: This is some smart reasoning!

core/test/unit/src/template-string.ts
```diff
@@ -2674,96 +2680,337 @@ describe("getContextLookupReferences", () => {
...
+          keyPath: ["reachableConsequent"],
+        },
+        {
+          keyPath: ["reachableAlternate"],
+        },
+      ],
+    },
+  ]
+  for (const testCase of branchTestCases) {
+    it(`correctly avoids dead code branches (test case: ${testCase.name})`, () => {
```
twelvemo, line 2993: super nice to cover all these test cases!

twelvemo (approved):
This is great work and a real fun read! Together with the good test coverage this really looks good to go!

PR #6870, opened 2025-02-20
fix: correct source mapping in varfiles

stefreak:
Ensures that template error messages in YAML varfiles will be correct

Correction for error behaviour for the changes in #6844

### DIFF
--- a/core/src/config/base.ts
+++ b/core/src/config/base.ts
@@ -8,7 +8,6 @@
 
 import dotenv from "dotenv"
 import { sep, resolve, relative, basename, dirname, join } from "path"
-import { load } from "js-yaml"
 import { lint } from "yaml-lint"
 import { omit, isPlainObject } from "lodash-es"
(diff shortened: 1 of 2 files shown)


thsig (approved):

core/src/config/base.ts
```diff
@@ -628,46 +621,72 @@ export const loadVarfile = profileAsync(async function loadVarfile({
...
+        return {
+          data: parsed as PrimitiveMap,
+          // source mapping to JSON has not been implemented at this point
+          source: { path: [] },
+        }
       } else if (filename.endsWith(".yml") || filename.endsWith(".yaml")) {
+        const loaded = await loadAndValidateYaml({
+          content: fileContents.toString("utf-8"),
+          filename: resolvedPath,
+          version: "1.2",
```
stefreak, line 657: we are using a different yaml parsing library here than before; It looks like the version of `js-yaml` we were using for varfiles also used YAML 1.2

thsig (approved):

PR #6879, opened 2025-02-24
feat(template)!: disallow optional values (`?` syntax) in template strings

vvagaytsev:

This fixes the UX issue when using template strings in URLs. For example `${var.baseUrl}?foo=bar` results in `http://example.comfoo=bar` which users may find unexpected.

Users should use `${var.baseUrl || 'defaultValue'}` instead.

Fixes #

### DIFF
--- a/core/test/unit/src/template-string.ts
+++ b/core/test/unit/src/template-string.ts
@@ -73,14 +73,56 @@ describe("parse and evaluate template strings", () => {
     expect(res).to.equal("value")
   })
 
-  it("should correctly resolve if ? suffix is present but value exists", () => {
-    const res = legacyResolveTemplateString({ string: "${foo}?", context: new TestContext({ foo: "bar" }) })
-    expect(res).to.equal("bar")
-  })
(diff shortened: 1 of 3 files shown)


core/test/unit/src/template-string.ts
```diff
@@ -73,14 +73,56 @@ describe("parse and evaluate template strings", () => {
...
+            "Invalid template string (prefix-${some}?-suffix): Could not find key some. Available keys: (none).",
+        }
+      )
+    })
+
+    it("should throw if an expression with ? in member expression cannot be resolved", async () => {
+      await expectError(
+        () =>
+          legacyResolveTemplateString({
+            string: '${actions.build["${parent.name}?"]}',
```
stefreak, line 111: to make this test more similar to what we've tested before, that the `MemberExpression` detects invalid types, you can change this to

`${actions.build["${parent.name || null}"]}`

vvagaytsev:
Superseded by #6939.

PR #6887, opened 2025-02-25
chore(template): feature flag for optional template values

vvagaytsev:

Fixes #

### DIFF
--- a/core/src/template/ast.ts
+++ b/core/src/template/ast.ts
@@ -18,13 +18,15 @@ import { type ConfigSource, validateSchema } from "../config/validation.js"
 import type { Branch } from "./analysis.js"
 import { TemplateStringError } from "./errors.js"
 import { styles } from "../logger/styles.js"
+import { GardenApiVersion } from "../constants.js"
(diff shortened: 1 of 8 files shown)


core/src/template/ast.ts
```diff
@@ -510,13 +516,35 @@ export class FormatStringExpression extends TemplateExpression {
...
+
+      reportDeprecatedFeatureUsage({
+        apiVersion: GardenApiVersion.v1,
+        log,
+        deprecation: "optionalTemplateValueSyntax",
+      })
+
+      emitNonRepeatableWarning(
+        log,
+        `Found deprecated optional template value syntax in ${styles.highlight(this.rawText)}`
```
stefreak, line 544: It's possible that template strings are huge multi-line strings, so we should probably port the logic from `TemplateStringError` and re-use it here to print an excerpt.

vvagaytsev: Great idea, thanks! 👍 
I've done that in #6890, pls review :)

PR #6891, opened 2025-02-26
chore(sync): deprecate `sync` commands outside dev console

vvagaytsev:

This PR deprecates all `sync` sub-commands if those are executed not in the dev console.

In the next major release we can make these commands only available in the dev console.
This will make the sync lifecycle management more stable and robust.

The initial attempt was done in #5733 - there `sync start` and `sync stop` commands were made available only in dev console.

Fixes #


stefreak (approved):
This is great, thank you!

PR #6893, opened 2025-02-28
chore(k8s): flip default value of deploy `spec.waitForJobs` in `apiVersion: garden.io/v2`

vvagaytsev:

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/kubernetes-type/handlers.ts
+++ b/core/src/plugins/kubernetes/kubernetes-type/handlers.ts
@@ -34,7 +34,6 @@ import { configureKubernetesModule } from "./module-config.js"
 import { configureLocalMode, startServiceInLocalMode } from "../local-mode.js"
 import type { ExecBuildConfig } from "../../exec/build.js"
 import type { KubernetesActionConfig, KubernetesDeployAction, KubernetesDeployActionConfig } from "./config.js"
-import { getWaitForJobs } from "./config.js"
 import type { DeployActionHandler } from "../../../plugin/action-types.js"
 import type { ActionLog } from "../../../logger/log-entry.js"
 import type { ActionMode, Resolved } from "../../../actions/types.js"
@@ -386,7 +385,7 @@ export const kubernetesDeploy: DeployActionHandler<"deploy", KubernetesDeployAct
       resources: namespaceManifests,
       log,
       timeoutSec: action.getConfig("timeout"),
-      waitForJobs: getWaitForJobs(spec, log),
+      waitForJobs: spec.waitForJobs,
     })
   }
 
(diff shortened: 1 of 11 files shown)


core/src/plugins/kubernetes/kubernetes-type/handlers.ts
```diff
@@ -75,6 +77,8 @@ export const kubernetesHandlers: Partial<ModuleActionHandlers<KubernetesModule>>
         files,
         manifests,
         sync: convertKubernetesModuleDevModeSpec(module, service, serviceResource),
+        // TODO(0.14): remove this because the default value will be set by schema
+        waitForJobs: getDefaultWaitForJobs().defaultValue,
```
stefreak, line 81: `waitForJobs` wouldn't be required for action configs, right? Does this line in the module to action conversion code have any effect?

core/src/util/deprecations.ts
```diff
@@ -256,3 +262,44 @@ export function reportDeprecatedSyncCommandUsage({
...
+  return lines.join("\n")
+}
+
+export function reportDefaultConfigValueChange({ apiVersion, log, deprecation }: DeprecationWarningParams) {
+  if (apiVersion === GardenApiVersion.v2) {
+    return
+  }
+
+  const warnMessage = makeDefaultConfigChangeMessage({ deprecation, includeLink: true, style: true })
+  emitNonRepeatableWarning(log, `\nDEFAULT CONFIGURATION CHANGE WARNING: ${warnMessage}\n`)
```
stefreak, line 304:
```suggestion
  emitNonRepeatableWarning(log, `\nDEFAULT VALUE WILL CHANGE: ${warnMessage}\n`)
```

Just some thoughts here: This warning message might be a little bit annoying for users, as there is no way to get rid of the warning other than updating `apiVersion` to v2, which we can only really recommend once we have implemented all the changes.
Should we wait with merging this one until we're done with the other changes?

Also, can we maybe detect that jobs will actually be deployed and only then print the warning, as only then this will be a breaking behavioural change (to reduce the number of people who will see this warning in the first place)

stefreak (commented):
Nice work! Shared some thoughts on this in a comment, what do you think?

core/src/plugins/kubernetes/kubernetes-type/handlers.ts
vvagaytsev: Removed in 7f0af50ffb55a86c635849f94fb877798d6282e3.

core/src/util/deprecations.ts
vvagaytsev: Makes sense. It looks like the `waitForJobs` is only used by `Job` status handler. So, we can call the `getWaitForJobs()` helper there and print a warning if no explicit value has been specified.

vvagaytsev: Done in 1994924d0715e5e0f0a48d180662ef99f17d2ef9.

stefreak: Ah, super amazing! That change is fantastic 👍

PR #6894, opened 2025-02-28
fix: improve error message when servicePort does not reference a port by name

twelvemo:

Fixes #


core/src/plugins/kubernetes/container/ingress.ts
```diff
@@ -81,7 +81,12 @@ export async function createIngressResources(
         // make sure the TLS secrets exist in this namespace
         await ensureSecret(api, cert.secretRef, namespace, log)
       }
-
+      if (!findByName(ports, ingress.spec.port)) {
```
vvagaytsev, line 84: Nit. The statement `findByName(ports, ingress.spec.port)` can be extracted as a local var and reused in L89.

vvagaytsev (approved):
LGTM! Thank you!👍

PR #6897, opened 2025-03-03
chore(exec): flip default value of `buildAtSource` for `exec` Build actions in `apiVersion: garden.io/v2`

vvagaytsev:

Fixes #


core/src/actions/build.ts
```diff
@@ -43,10 +48,33 @@ export interface BuildActionConfig<T extends string = string, S extends object =
...
+   * This must be accessed via `getBuildAtSource` helper to ensure the correct default value.
+   */
   buildAtSource?: boolean
   copyFrom?: BuildCopyFrom[]
 }
 
+export function getBuildAtSource(config: BuildActionConfig): boolean {
+  const projectApiVersion = getProjectApiVersion()
+  const buildAtSource = config.buildAtSource
+  if (buildAtSource === undefined && config.type === "exec") {
```
stefreak, line 61: If this only affects `exec` actions, would the `configure` handler in the `exec` plugin maybe be the better place to decide the default value for `buildAtSource`?

Maybe we could then even adapt the schema in the plugin to make sure the docs are correct for exec vs other action types?

vvagaytsev: Did that in fff6faabb2486fdc280743b12966e565efa130c3, but we decided to reject this change in 0.14.

PR #6903, opened 2025-03-04
feat(cli): add more fields to garden get actions detailed output

### LINKED ISSUE #6902: [FEATURE]: Add more details to `garden get actions` to extract action version and build publish information
### Background / Motivation

We are using Garden in our CI/CD pipeline. So far we had been using modules and relying on the output of `garden get modules` to extract  for each module the version and the image name for builds that have been published, to continue deployment to other environments.

We now want to migrate to using Garden actions. However the output of `garden get actions` is missing those fields, even with the `--detail` flag.

When invoking `garden get actions --detail`, the following fields should be included for each action
- `version` including the `versionString` inside, to see the Garden version and the tag that Garden uses for published images.
- `allowPublish`, to see whether Garden does publish the image.
- `publishId`, to see the image name that Garden uses for publishing.

The output of `garden get actions --detail -o json` can be parsed and used for deployment scripts that publish the images built by Garden to other deployment environments not managed by Garden.

The additional fields can be added in `core/src/commands/get/get-actions.ts` for the action output with detail likely rather easily.

The `interface GetActionsCommandResultItem` could have these additional fields:
```javascript
  version?: ActionVersion
  allowPublish?: boolean
  publishId?: string
```
These can be filled in in the `getActionsOutput` of `GetActionsCommand` when `isOutputDetailed` is set as follows:
```javascript
  version: a.getFullVersion(),
  allowPublish: a.getConfig().allowPublish ?? undefined,
  publishId: a.getSpec("publishId") ?? undefined,
```

klara-meyer:

See #6902 for more background. This adds additional fields to the `garden get actions --detail` output, which can be parsed to use images published by Garden for other deployments.

It adds the following new fields to the actions output:
- `version`: the full resolved version of a Garden action, including `versionString` and `files`.
- `allowPublish`: the value of `allowPublish` in the Garden action.
- `publishId`: the value of the `publishId` in the Garden action spec, or the default publish ID derived by Garden from the action name.

Fixes #6902

### DIFF
--- a/core/src/commands/get/get-actions.ts
+++ b/core/src/commands/get/get-actions.ts
@@ -58,6 +58,26 @@ export const getActionsCmdOutputSchema = createSchema({
       .description("The state of the action."),
     path: joi.string().description("The relative path of the action config file."),
     disabled: joi.boolean().description("Flag to identify if action is disabled."),
+    version: joi
+      .object()
+      .keys({
+        configVersion: joi.string().required().description("The version string of the action's config."),
+        sourceVersion: joi.string().required().description("The version string of the action's source."),
+        versionString: joi.string().required().description("The version string of the action."),
+        dependencyVersions: joi
+          .object()
+          .pattern(joi.string(), joi.string())
+          .required()
+          .description("Map with the version strings of the action's dependencies."),
+        files: joiArray(joi.string()).required().description("List of the files included in the action."),
+      })
+      .description("Object with the full version information of the action."),
+    allowPublish: joi
+      .boolean()
+      .description("Flag to identify whether publishing the build is enabled. Only available for build actions."),
+    publishId: joi
+      .string()
+      .description("The image ID used to publish the image of the action. Only available for build actions."),
     moduleName: joi
       .string()
       .description("The name of the module the action is derived from. Only available for converted actions."),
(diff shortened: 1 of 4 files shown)


core/src/commands/get/get-actions.ts
```diff
@@ -25,6 +25,9 @@ interface GetActionsCommandResultItem {
   state?: ActionState
   path?: string
   disabled?: boolean
+  version?: ActionVersion
```
stefreak, line 28: The `getActionsCmdOutputSchema` also needs to be adapted, and once the schema has been changed we also need to regenerate docs by running `npm run generate-docs`

stefreak (changes requested):
Hello @klara-meyer this is looking pretty good! Thank you for your contribution 🚀 

There seem to be some assertion failures in the unit tests, could you adapt the assertions to the new behaviour? You can find these in https://app.circleci.com/jobs/github/garden-io/garden/617563 & you can run these unit tests locally by running `cd core; npm run test -- -g "<test name>"`

Also, before we can accept this you'd need to align the changed `GetActionsCommandResultItem` with the output schema for the get actions command. Changing the output schema will also automatically trigger a reference doc update, you'd need to run `npm run generate-docs` and commit the updated reference docs.

Thanks again & once that's done I'll happily accept this change! 👍

stefreak (approved):
Awesome work! Looks good to me 🥳

PR #6904, opened 2025-03-04
fix(build-staging): fix the detection of paths to be deleted

vvagaytsev:

This fixes the `scanDirectoryForClone` helper to always return paths in POSIX format. It makes the path comparison logic correct for all platforms, including Windows.

Fixes #


stefreak (approved):
Let's make sure we add test coverage in a follow-up PR.

PR #6923, opened 2025-03-10
fix(docs): propagate correct action kind to k8s Run/Test reference docs

vvagaytsev:

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/config.ts
+++ b/core/src/plugins/kubernetes/config.ts
@@ -44,6 +44,7 @@ import { defaultKanikoImageName, defaultUtilImageRegistryDomain, defaultSystemNa
 import type { LocalKubernetesClusterType } from "./local/config.js"
 import type { EphemeralKubernetesClusterType } from "./ephemeral/config.js"
 import { makeDeprecationMessage } from "../../util/deprecations.js"
+import type { ActionKind } from "../../plugin/action-types.js"
 
 export interface ProviderSecretRef {
   name: string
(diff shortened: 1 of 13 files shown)


docs/reference/action-types/Run/container.md
```diff
@@ -593,7 +593,7 @@ Specify an image ID to deploy. Should be a valid Docker image identifier. Requir
 
 [spec](#spec) > cacheResult
 
-Set to false if you don't want the Runs's result to be cached. Use this if the Run needs to be run any time your project (or one or more of the Run's dependants) is deployed. Otherwise the Run is only re-run when its version changes, or when you run `garden run`.
+Set to false if you don't want the Run's result to be cached. Use this if the Run needs to be run any time your project (or one or more of the Run's dependants) is deployed. Otherwise the Run is only re-run when its version changes, or when you run `garden run`.
```
stefreak, line 596: This reads a little bit more natural to me, instead of using the apostrophe:
```suggestion
Set to false if you don't want the Run action result to be cached. Use this if the Run needs to be run any time your project (or one or more of the Run's dependants) is deployed. Otherwise the Run is only re-run when its version changes, or when you run `garden run`.
```

core/src/plugins/kubernetes/config.ts
```diff
@@ -902,15 +903,15 @@ export const portForwardsSchema = () =>
 
 export const runPodSpecWhitelistDescription = () => runPodSpecIncludeFields.map((f) => `* \`${f}\``).join("\n")
 
-export const runCacheResultSchema = () =>
+export const runCacheResultSchema = (kind: ActionKind) =>
   cacheResultSchema().description(
     dedent`
-Set to false if you don't want the Runs's result to be cached. Use this if the Run needs to be run any time your project (or one or more of the Run's dependants) is deployed. Otherwise the Run is only re-run when its version changes, or when you run \`garden run\`.
+Set to false if you don't want the ${kind} action result to be cached. Use this if the ${kind} needs to be run any time your project (or one or more of the ${kind}'s dependants) is deployed. Otherwise the ${kind} is only re-run when its version changes, or when you run \`garden run\`.
```
stefreak, line 909: My suggestion also applied to every other occurrence of Kind in these sentences:
```suggestion
Set to false if you don't want the ${kind} action result to be cached. Use this if the ${kind} action needs to be run any time your project (or one or more of the ${kind} action's dependants) is deployed. Otherwise the ${kind} action is only re-run when its version changes, or when you run \`garden run\`.
```

PR #6924, opened 2025-03-10
fix(k8s): respect `spec.cacheResult` flag in `kubernetes-pod` Test actions

vvagaytsev:

Prior to fix the `spec.cacheResult: false` would not have any effect for Test actions of type `kubernetes-pod`.

This PR fixes it to respect the behaviour of that flag described in the [docs](https://docs.garden.io/reference/action-types/test/kubernetes-pod#spec.cacheresult).

The default value is still `true`. So, there are no breaking changes in the behaviour.

Fixes #


stefreak (approved):

stefreak:
Good catch!

PR #6926, opened 2025-03-10
refactor(k8s): reduce difference in `Run` and `Test` handlers between action types

vvagaytsev:

A tiny janitorial PR to reduce the implementation difference between Run and Test action handlers of the `kubernetes` plugin.

This will allow to continue doing the next steps with ease:
* get rid of `any` type in `getResult` handler
* introduce explicit and strict payload types for cached results
* extract common machinery between handlers and reduce the code duplication

That will allow to introduce abstract for the results caching and will also make 0.13 code more robust and stable.

Fixes #

~This PR could be combined with one of the "next steps" mentioned above; but it was made intentionally tiny for the ease of review.~


vvagaytsev:
Closing in favour of #6927.

PR #6927, opened 2025-03-10
fix(k8s): do not cache duplicate logs in Run and Test results

vvagaytsev:

Prior to this PR, we did not have a well-defined shape for the cacheable Run and Test results in the `kubernetes` plugin.

So, we did cache objects of 2 shapes. One was a `RunResult` with `namespaceStatus` like
```
{
  log: "logs", // potentially a very log string
  ... // other RunResult fields
}
```

Another was `ActionStatus = {state: ActionState, detail: RunResult, outputs: { log: string }}` like
```
{
  actionState: "ready",
  detail: {
    log: "logs", // potentially a very log string
    ... // other RunResult fields
  },
  output: {
    log: "logs" // the same logs as above
  }
}
```

We did the log truncation in the special helper `trimRunOutput` - but it was applied only to the `log` field of the `RunResult`.

Thus, the objects of the second shape with too-long logs would not be cached in the `ConfigMap` because of the object size limitations.

This PR unifies the shape of the cacheable objects, and minimizes it to avoid duplicate logs in the cache.

This PR also adds the missing test for `storeTestResult` helper.

Fixes #

This also includes the refactoring made in #6926.


core/src/plugins/kubernetes/run-results.ts
```diff
@@ -18,83 +18,120 @@ import { hashSync } from "hasha"
...
-    }
+  return toRunActionStatus(cachedResult)
+}
 
-    if (!result.outputs.log) {
-      result.outputs.log = result.log || ""
-    }
+export function getRunResultKey(ctx: PluginContext, action: Action) {
+  // change the result format version if the result format changes breaking backwards-compatibility e.g. serialization format
+  const resultFormatVersion = 2
```
stefreak, line 43: Do we need to increment the result format version here?

Will this version of Garden be fine with the data that older versions of Garden stored under that cache key?
Will older versions of Garden be fine with the new data shape that this version of Garden stores under that cache key?

vvagaytsev: The shape of the cacheable objects was changed. So I changed the key format to make sure that the data of different shape won't share the same keys.

Afaik, incrementing this should ensure the update of the keys for the same actions.

For Run actions, we had different shapes between `container` Run and `kubernetes-pod`/`helm-pod` Runs. Now all 3 are unified.

The old cached data will still be available for the older Garden versions, and will not be reachable via the new keys. This version will trigger re-execution of some tests and runs and store the results under the new keys that do not clash with the old keys.

We can postpone this change to 0.14, and just truncate both `log` in `RunResult` and `outputs`in 0.13.

stefreak (approved):
Amazing work! thank you, this is way easier to change

PR #6936, opened 2025-03-11
chore: update Zod error message helper

thsig:

Include the path to the error when rendering Zod validation error messages.


thsig:
Essentially, we're just rendering the object path to the validation issue, if available (which is a nice bit of extra information to include in these error messages).

stefreak:
How did the error look before, and how does it look after? Do we have unit tests for this helper?

vvagaytsev:
The path was included in the error message in #7217. This PR make it more readable.

Example error before this (with #7212):
```console
Auth redirect request failed with the error: "Invalid query parameters": Expected string at field 'jwt', but received undefined
```
and after (here the path is also highlighted, the colour is not visible here):
```console
Auth redirect request failed with the error: "Invalid query parameters": Expected string at path jwt, but received undefined.
Expected string at path rt, but received undefined.
Issue  at path jwtVal: Invalid input
```

PR #6938, opened 2025-03-11
refactor(k8s): introduce cache abstraction for `getResult` handlers of Run and Test actions

vvagaytsev:

Introduce `ResultCache` interface to avoid usage of a set of standalone helpers and ensure the usage of proper data types.

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/test-results.ts
+++ b/core/src/plugins/kubernetes/test-results.ts
@@ -11,7 +11,6 @@ import { KubeApi, KubernetesError } from "./api.js"
 import type { ContainerTestAction } from "../container/moduleConfig.js"
 import type { PluginContext } from "../../plugin-context.js"
 import type { KubernetesPluginContext, KubernetesProvider } from "./config.js"
-import type { Log } from "../../logger/log-entry.js"
 import { hashSync } from "hasha"
 import { gardenAnnotationKey } from "../../util/string.js"
 import { upsertConfigMap } from "./util.js"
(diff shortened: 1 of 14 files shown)


core/src/plugins/kubernetes/test-results.ts
```diff
@@ -11,146 +11,135 @@ import { KubeApi, KubernetesError } from "./api.js"
...
+  }: LoadResultParams<CacheableTestAction>): Promise<CacheableTestResult | undefined> {
+    const k8sCtx = <KubernetesPluginContext>ctx
+    const api = await KubeApi.factory(log, ctx, k8sCtx.provider)
+    const testResultNamespace = await getSystemNamespace(k8sCtx, k8sCtx.provider, log)
+    const resultKey = this.cacheKey(k8sCtx, action)
+
+    try {
+      const res = await api.core.readNamespacedConfigMap({ name: resultKey, namespace: testResultNamespace })
+      const result = deserializeValues(res.data!)
+      return result as CacheableTestResult
```
stefreak, line 83: is it safe to simply cast here?

stefreak (approved):
nice work 👍

core/src/plugins/kubernetes/test-results.ts
vvagaytsev: This is what we did before implicitly via declaring this as `any`. Now we have better control on what we store and load, so I'd say it's not less safe than it was before :)

We can, ofc, validate the raw body by presence of a of set of certain fields, but this looks safe-enough for now. We can improve the internal validation if we get any issues with this. The key version incrementation (see 04b5d04db6d560a95ddb28c839f85f9d0e638a1c) should protect is from unexpected data shapes here.

PR #6941, opened 2025-03-12
feat(k8s)!: use local filesystem cache for Run and Test actions

vvagaytsev:

This PR removed the old Kubernetes `ConfigMap`-based cache for Run and Test actions of the `kubernetes` plugin, and introduces a local filesystem-based cache instead.

Fixes #


core/src/plugins/kubernetes/results-cache-fs.ts
```diff
@@ -0,0 +1,206 @@
...
+ * because each Run and Test result
+ * is usually read and written only once per Garden command execution.
+ */
+export class SimpleFileSystemCache<T> {
+  private readonly cacheDir: string
+  private readonly lock: AsyncLock
+
+  constructor(cacheDir: string) {
+    this.cacheDir = cacheDir
+    this.lock = new AsyncLock()
```
stefreak, line 51: this lock protects against concurrent access within the same process, but users sometimes run multiple garden commands at the same time.

Should we use the SQLite database that we ship with anyway, and where we store tokens instead? SQLite is very robust against concurrent access even from multiple processes

vvagaytsev: Well, concurrent writes are possible if user run exactly the same command simultaneously from multiple terminals.

I'm not sure if this will be an often scenario. Afaik, we decided to get rid of SQLite in 0.13 because it slowed down the startup time of Garden CLI.

Now, if any issue occurs while the caching, it should not crash the Garden process. So, the current approach might be good enough for now. Let's try to find a better solution if any issues occur.

stefreak: ah, I didn't know we removed sqlite (as the word still occurs in the code when looking it up)

I think that it's definitely possible that two different commands run simultaneously have actions in common that calculate to the same version hash, e.g. deploying to two different environments at the same time.

I think just ignoring errors is not quite enough here; E.g. if we're reading invalid data (e.g. that became invalid due to concurrent writes from 2 different processes), we need to at least remove the cache entry to allow for a valid entry to be saved next time.

We should also probably log a debug-level message when there is an error due to corrupt data.

I don't think we should just catch all and ignore, as is the case in the current PR: If something happens that is unrelated to data / schema validation, then we should throw that error and crash (to avoid programming mistakes to be overlooked)

core/src/plugins/kubernetes/results-cache-fs.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  constructor({
+    cacheDir,
+    cacheKeyProvider,
+    resultValidator,
+    resultTrimmer,
+  }: {
+    cacheDir: string
+    cacheKeyProvider: CacheKeyProvider
+    resultValidator: ResultValidator<R>
```
stefreak, line 154: can this just be a schema parameter?
Do we really need to have a generic validator?

core/src/plugins/kubernetes/results-cache-fs.ts
```diff
@@ -0,0 +1,202 @@
...
+    resultTrimmer: ResultTrimmer<R>
```
stefreak, line 155: can this just be a method in `AbstractResultCache`? I know why you did this, but it makes the code also a little bit more difficult to follow & only if we need more than a maxSize parameter it's really worth it IMO

core/src/plugins/kubernetes/results-cache-fs.ts
stefreak: Ah well, I had another look and noticed that you're using writeFileAtomic. That should prevent issues with concurrent writes.

So only the point about exception handling still stands 👍

core/src/plugins/kubernetes/results-cache-fs.ts
stefreak: We're using `renderZodError` in the error path, so we're tightly coupled with zod anyway.

core/src/plugins/kubernetes/results-cache-fs.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  public async clear(): Promise<void> {
+    await this.lock.acquire(this.cacheDir, async () => {
+      await remove(this.cacheDir)
+    })
+  }
+}
+
+export function getLocalKubernetesRunResultsCacheDir(gardenDirPath: string): string {
+  return join(gardenDirPath, CACHE_DIR_NAME, "local-k8s-plugin-results")
```
stefreak, line 140: if we want to extend the functionality to support other action types in the future, I think it would be good if we're using a directory from the beginning that we can keep using (e.g. just "action-cache")

core/src/plugins/kubernetes/results-cache.ts
```diff
@@ -37,28 +60,111 @@ export interface StoreResultParams<A extends RunAction | TestAction, R extends C
...
+ */
+export const currentResultSchemaVersion: SchemaVersion = "v1"
+
+export type CreateCacheableResultKeyParams = {
+  schemaVersion: SchemaVersion
+  ctx: PluginContext
+  action: CacheableAction
+}
+
+export class StructuredCacheKey {
```
stefreak, line 101: I definitely like that the `StructuredCacheKey` is a separate class 👍

core/src/plugins/kubernetes/results-cache.ts
```diff
@@ -37,28 +60,111 @@ export interface StoreResultParams<A extends RunAction | TestAction, R extends C
...
+    this.projectName = ctx.projectName
+    this.actionKind = action.kind
+    this.actionName = action.name
+    this.actionType = action.type
+    this.actionVersion = action.versionString()
+  }
+
+  @Memoize()
+  public calculate(): string {
+    const key = `${this.projectName}--${this.actionType}.${this.actionName}--${this.actionVersion}--${this.schemaVersion}`
```
stefreak, line 120: for the team cache, it will be the concern of the backend to distinguish cache entries from different schema versions. Let's implement the file cache the same way.
```suggestion
    const key = `${this.projectName}--${this.actionType}.${this.actionName}--${this.actionVersion}`
```

core/src/plugins/kubernetes/run-results.ts
```diff
@@ -47,114 +32,16 @@ export const k8sGetRunResult: RunActionHandler<"getResult", any> = async (params
...
-}): CacheableRunResult {
-  const result = composeCacheableResult(params)
-  return {
-    ...result,
-    taskName: result.actionName,
+export function getRunResultCache(gardenDirPath: string): LocalResultCache<CacheableRunAction, CacheableResult> {
+  if (runResultCache === undefined) {
+    runResultCache = new LocalResultCache<CacheableRunAction, CacheableResult>({
+      cacheDir: getLocalKubernetesRunResultsCacheDir(gardenDirPath),
+      cacheKeyProvider: cacheKeyProviderFactory(currentResultSchemaVersion),
```
stefreak, line 41: cacheKeyProviderFactory doesn't read like idiomatic javascript.
maybe we can pass a function called `getCacheKey`?
If the schema version does not need to be injected, then maybe it can also just be a method on the abstract class?

core/src/plugins/kubernetes/results-cache-fs.ts
vvagaytsev: Agree, I'lll simplify this.

core/src/plugins/kubernetes/results-cache-fs.ts
vvagaytsev: Initially I was thinking about keeping the schema as member of the class here. But I got some troubles with delcaring its type. Then decided to just keep a function.
Maybe we can just have validation method in the abstract base class and reference the schema from there.

stefreak: I'm happy to pair on it if there are type troubles

core/src/plugins/kubernetes/results-cache-fs.ts
stefreak: I thought about it a little, and I believe if we're writing atomically we can completely eliminate the lock, I believe it is not necessary

stefreak: Maybe it's necessary for remove on Windows, but we can also make sure that we never delete keys while we might receive writes (e.g. by deleting old keys during setup)

core/src/plugins/kubernetes/results-cache-fs.ts
vvagaytsev: Ah, moving it to ABC does not seem to be a good idea. In that case we'll have to extends the class to change the schema validation for another plugin type.

But the different backend-based cache implementations should not inherit concrete schemas, so the schema should still be a class member. I'm checking how to define its type and bind it to generic type `R`. Maybe having validator as a function is not a bad idea. It was done in that way not to keep it Zod-agnostic, but just to avoid type issues.

core/src/plugins/kubernetes/results-cache-fs.ts
vvagaytsev: Changed this in b01195a43749321d313d95a0dcb868bfe508fb08.

core/src/plugins/kubernetes/results-cache-fs.ts
vvagaytsev: Fixed error handling in 698e7042a2c8ce928f1fd85e02993eec54a92fc7.

vvagaytsev: Yes, I also thought about the necessity of the lock. Concurrent reads are fine, and write happens atomically.

core/src/plugins/kubernetes/results-cache-fs.ts
vvagaytsev: My idea was to add something like `local-exec-plugin-results` in the same `.garden/cache` dir. Now we store `provider-statuses` in that shared cache directory.

Of course, we can have a common parent dir like:
```
.garden/cache/action-results/
    -> k8s-plugin
    -> exec-plugin
```
Would that be preferred?

core/src/plugins/kubernetes/results-cache.ts
vvagaytsev: Shouldn't we always avoid cache key collisions? Backend can do whatever it needs, but we decided that cache key construction algorithm will be controlled by the client. Also, cache key must be unique, otherwise we'll get collisions.

So, backend will store the schema version in DB as a separate column, it's good and convenient for its needs.
But, if we remove the schema version from the cache key calculation, then 2 result objects of 2 different shapes will result into the same cache key.

Thus, the unique identifier will be a tuple of `<cacheKey, schemaVersion>`. It will require a composite unique index in DB, and will make the FS-based implementation a bit more complex. We'll still have to append the schema version to calculate the unique key for the local cache.

Why don't we just use the current key format? Backend can still maintain its own data structures and validation, but it doesn't look like the backend internal implementation should affect the way the client builds the cache keys.

core/src/plugins/kubernetes/results-cache-fs.ts
stefreak: Delete is also fine on unix, as it's fine to delete a file and still read it or keep file descriptors open. only for delete on windows we need to ensure that we either handle the error, if another app has one of the files open, or we avoid deleting and reading simultaneously somehow.

core/src/plugins/kubernetes/run-results.ts
vvagaytsev: Agree, we don't need a factory here; and we don't need to inject the key construction logic from outside.
This logic can be moved to the abstract class.

I made it too generic initially. Now it's clear that the cache result for other plugins and actions can use the same structured key format, only the shape of cached value can be different. Also, we likely do not plan to maintain co-existence of a few schema versions in one version of Garden CLI.

We can specify the schema version in the abstract class constructor and do the calculation logic in the abstract class method.

Simplified this in 0ece883cfc5e597a848fa3b3107393cc1c80043c and 7d9c49075fd5870b9ca3f214cad1a66aa073a5d2.

core/src/plugins/kubernetes/results-cache-fs.ts
stefreak: I think it's ok to use one directory for all results, as the hash algorithm we chose should make it impossible in practice to have collisions between different plugins.
Of course this assumes we are using a full cryptographic hash like SHA256.

core/src/plugins/kubernetes/results-cache-fs.ts
stefreak: I see now! 👍 Happy to help with the type issues when we see each other in person, thanks for the explanation

core/src/plugins/kubernetes/results-cache.ts
vvagaytsev: Ah, I see, thanks for elaborating this!

Actually, in that case we agree to have a kind of isolated logical buckets for different schema versions.
That makes sense. So the requirement is just to ensure cache key uniqueness in the scope of a single schema version. In that case we also avoid cache key format change on every change of the data shape.

I'll update that part.

core/src/plugins/kubernetes/results-cache-fs.ts
vvagaytsev: Yes, the current cachhe key format ensures the different keys for different plugins, as it involves the action type.

Let's call the dir `action-results` and store everything in it. Simple plain file structure :)

vvagaytsev: Renamed it in 36a0f22093da81be03cabf6222d129a40da7da79.

core/src/plugins/kubernetes/results-cache.ts
vvagaytsev: Done in fec1de1792369c23daa1ea7c00df53e9460565a6.

core/src/plugins/kubernetes/results-cache-fs.ts
vvagaytsev: Removed lock in 8a06ce426ef2741ddc1608651bb08671ecdffcd5.

thsig (commented):
Great stuff! Had a few comments there.

core/src/plugins/kubernetes/container/run.ts
```diff
@@ -43,9 +43,10 @@ export const k8sContainerRun: RunActionHandler<"run", ContainerRunAction> = asyn
     dropCapabilities,
   })
 
-  const detail = composeCacheableRunResult({ result, action, namespaceStatus })
+  const detail = composeKubernetesCacheEntry({ result, namespaceStatus })
 
   if (action.getSpec("cacheResult")) {
+    const runResultCache = getResultCache(ctx.gardenDirPath)
```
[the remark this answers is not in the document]
vvagaytsev: We do not instantiate it on every execution. The getter helper creates it once and caches.

core/src/plugins/kubernetes/results-cache-fs.ts
```diff
@@ -0,0 +1,149 @@
...
+ * because each Run and Test result
+ * is usually read and written only once per Garden command execution.
+ */
+export class SimpleLocalFileSystemCacheStorage implements CacheStorage {
+  private readonly cacheDir: string
+  private readonly schemaVersion: SchemaVersion
+  private readonly log: Log
+
+  constructor(cacheDir: string, schemaVersion: SchemaVersion) {
+    this.cacheDir = cacheDir
```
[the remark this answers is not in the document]
thsig: We probably shouldn't prune unless there are a lot of files in there, so maybe we could have a cutoff of 3K files?

core/src/plugins/kubernetes/container/run.ts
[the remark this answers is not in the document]
stefreak: I think it's ok to implement this purely in the kubernetes plugin for now, as soon as we want to support other plugins like exec we can look into that for sure 👍

core/src/plugins/kubernetes/results-cache-base.ts
```diff
@@ -0,0 +1,207 @@
...
+  constructor({ ctx, action }: { ctx: PluginContext; action: CacheableAction }) {
+    this.projectName = ctx.projectName
+    this.actionKind = action.kind
+    this.actionName = action.name
+    this.actionType = action.type
+    this.actionVersion = action.versionString()
+  }
+
+  @Memoize()
+  public calculate(): string {
```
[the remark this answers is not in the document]
vvagaytsev: Do we need those fields only for Run action or for Tests too?

core/src/plugins/kubernetes/results-cache-base.ts
```diff
@@ -0,0 +1,207 @@
...
+  private readonly projectName: string
+  private readonly actionKind: string
+  private readonly actionName: string
+  private readonly actionType: string
+  private readonly actionVersion: string
+
+  constructor({ ctx, action }: { ctx: PluginContext; action: CacheableAction }) {
+    this.projectName = ctx.projectName
+    this.actionKind = action.kind
+    this.actionName = action.name
```
stefreak, line 86:
```suggestion
    this.actionKind = action.kind
```
I think we don't need these two. Why should we enforce a cache miss if the name changes?

core/src/plugins/kubernetes/results-cache-base.ts
```diff
@@ -0,0 +1,207 @@
...
+    this.actionType = action.type
+    this.actionVersion = action.versionString()
```
stefreak, line 88: We probably want to use an unshortened version string to avoid cache collisions, but that can be done in a followup PR

core/src/plugins/kubernetes/container/run.ts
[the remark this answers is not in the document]
thsig: Ah of course, didn't notice the memoization there. Nice.

core/src/plugins/kubernetes/results-cache-base.ts
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  @Memoize()
+  public calculate(): string {
+    const key = `${this.projectName}--${this.actionType}.${this.actionName}--${this.actionVersion}`
+    const hash = hashSync(key, { algorithm: "sha1" })
+    return `${this.actionKind.toLowerCase()}-result--${hash.slice(0, 32)}`
```
stefreak, line 95:
```suggestion
    const hash = hashSync(key, { algorithm: "sha256" })
    return `${this.actionKind.toLowerCase()}-result--${hash}`
```
should we maybe use an unshortened cryptographically strong hash algorithm to ensure we are guaranteed to be free of collisions? (Assuming we do the same for the action version)

core/src/plugins/kubernetes/results-cache-base.ts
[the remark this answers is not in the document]
thsig: Only for Run actions.

This is because those have different semantics—tests can be cached purely based on the action version, because they're not designed to have side-effects (which is usually what Runs are for).

core/src/plugins/kubernetes/results-cache-base.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  constructor({ ctx, action }: { ctx: PluginContext; action: CacheableAction }) {
+    this.projectName = ctx.projectName
+    this.actionKind = action.kind
+    this.actionName = action.name
+    this.actionType = action.type
+    this.actionVersion = action.versionString()
+  }
+
+  @Memoize()
```
stefreak, line 91: Does memoize do anything here? We're creating a new instance of this class every time we're calculating a cache key right?

core/src/plugins/kubernetes/results-cache-fs.ts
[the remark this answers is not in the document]
stefreak: I think we should enforce an age limit, similar how we do in the backend

stefreak (approved):
For me this is good to go, but we need to probably create tasks or otherwise revisit some of the issues that were raised in the comments

core/src/plugins/kubernetes/results-cache-fs.ts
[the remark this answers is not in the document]
thsig: Yeah, that's probably simplest. Maybe 45 or 60 days would be a decent default.

core/src/plugins/kubernetes/results-cache-base.ts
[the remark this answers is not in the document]
stefreak: yes, this was also in the requirements

core/src/plugins/kubernetes/results-cache-base.ts
vvagaytsev: Agree, did that in 059e47bc369ef8ae3c8c9bbaa8952191af6626e7.

core/src/plugins/kubernetes/results-cache-base.ts
vvagaytsev: Right, simplified this in 059e47bc369ef8ae3c8c9bbaa8952191af6626e7.

core/src/plugins/kubernetes/results-cache-fs.ts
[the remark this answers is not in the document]
10ko: Let's use the same age as for the logs (7 days), we can increase it in the future if it feels too short.

core/src/plugins/kubernetes/results-cache-base.ts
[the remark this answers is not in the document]
vvagaytsev: Implemented in b354ac609c3f212e912ef9b798bb0c473db435b6.

core/src/plugins/kubernetes/results-cache-fs.ts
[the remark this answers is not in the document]
vvagaytsev: Implemented in 2e7dc1a73d7fef8b1bd96420e19f5ad785088f32.

core/src/commands/get/get-actions.ts
```diff
@@ -63,7 +63,8 @@ export const getActionsCmdOutputSchema = createSchema({
       .keys({
         configVersion: joi.string().required().description("The version string of the action's config."),
         sourceVersion: joi.string().required().description("The version string of the action's source."),
-        versionString: joi.string().required().description("The version string of the action."),
+        versionString: joi.string().required().description("The short version string of the action."),
+        versionStringFull: joi.string().required().description("The full version string of the action."),
```
stefreak, line 67: Do we want to expose the full version string here in the schema?

core/src/plugins/kubernetes/results-cache-base.ts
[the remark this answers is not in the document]
vvagaytsev: Made a follow-up fix in 68da3eebc082a75841d24b0f158d7b76fa46439f.

core/src/commands/get/get-actions.ts
vvagaytsev: Maybe we don't have to do that. Should we allow unknown keys then?

core/src/plugins/kubernetes/results-cache-base.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  protected validateResult(data: unknown, log: Log): z.output<ResultSchema> | undefined {
+    const result = this.resultSchema.safeParse(data)
+    if (result.success) {
+      return result.data
+    }
+
+    const errorMessage = deline`
+      The provided result doesn't match the expected schema.
+      Here is the output: ${renderZodError(result.error)}
```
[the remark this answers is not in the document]
stefreak: I'm not sure we'd want to keep all errors hidden, because users expect garden to help them speed up the CI, and if it doesn't for some reason, we need to let them know, otherwise they might blame us for issues that we can't really do much about, for example file system permissions.
This also plays into the CLI log message changes that haven't been adressed in this PR yet. I expect that these will be implemented in a separate PR?

core/src/plugins/kubernetes/container/run.ts
```diff
@@ -24,10 +24,7 @@ export const k8sContainerRun: RunActionHandler<"run", ContainerRunAction> = asyn
   const timeout = action.getConfig("timeout")
   const k8sCtx = ctx as KubernetesPluginContext
   const image = getDeployedImageId(action)
-  let namespaceStatus = await getNamespaceStatus({ ctx: k8sCtx, log, provider: k8sCtx.provider })
-  if (namespaceStatus.state === "missing") {
-    // return { state: "not-ready", detail: null, outputs: { log: "" } }
-  }
+  const namespaceStatus = await getNamespaceStatus({ ctx: k8sCtx, log, provider: k8sCtx.provider })
```
stefreak, line 27: What happens if we throw `InternalError` here if the namespace state is missing?

I don't quite like that we have to compare with namespace state in line 48.

core/src/commands/get/get-actions.ts
stefreak: No, can't we omit this in the get action command output?

core/src/plugins/kubernetes/container/run.ts
vvagaytsev: Changed this in cd3d726bcd0fe1681d4bbe627b0cf8b7023b5607.

stefreak: Nice 👍

core/test/integ/src/plugins/kubernetes/test-results.ts
```diff
@@ -37,39 +36,41 @@ describe("kubernetes Test results", () => {
...
         result: {
-          // command: [],
           log: data,
           startedAt: new Date(),
           completedAt: new Date(),
           success: true,
         },
-        action,
         // mock data
         namespaceStatus: {
```
stefreak, line 56: I think namespaceStatus doesn't need to be part of `composeCacheable*Result` (as discussed in person)

core/src/commands/get/get-actions.ts
vvagaytsev: Excluded it in d529f4551a5917aade920be71d9650ae962ed0f2.

core/test/integ/src/plugins/kubernetes/test-results.ts
vvagaytsev: Right, we do not need to store it as a part of the cached values. This will also simplify the code and the type definitions 👍 .

core/src/plugins/kubernetes/results-cache-fs.ts
```diff
@@ -0,0 +1,199 @@
...
+    this.log = RootLogger.getInstance().createLog({ name: "fs-cache" })
+  }
+
+  public static async getInstance(params: {
+    cacheDir: string
+    schemaVersion: SchemaVersion
+    cacheExpiryDays: number
+  }): Promise<SimpleLocalFileSystemCacheStorage> {
+    const instance = new SimpleLocalFileSystemCacheStorage(params)
+    await instance.invalidate()
```
stefreak, line 69: Do we need an async lock to prevent invalidating the same directory multiple times at once?
Do we really want to wait until invalidation completes before moving on? I think it would be absolutely acceptable in this instance to not await the promise.
```suggestion
    void instance.invalidate()
```

core/test/integ/src/plugins/kubernetes/test-results.ts
vvagaytsev: Done in 2ddaadf2b5096a75b37dfd43768ed08e491be018.

core/src/types/namespace.ts
```diff
@@ -6,14 +6,26 @@
...
-  state: NamespaceState
-}
+export const baseNamespaceStatusSchema = z.object({
+  pluginName: z.string(),
+  namespaceName: z.string(),
+  state: z.union([z.literal("ready"), z.literal("missing")]),
+})
+export type BaseNamespaceStatus = z.infer<typeof baseNamespaceStatusSchema>
+
+export const namespaceStatusSchema = z.discriminatedUnion("state", [
```
stefreak, line 18: do we still need this?

core/test/integ/src/plugins/kubernetes/container/run.ts
```diff
@@ -142,7 +174,22 @@ describe("runContainerTask", () => {
...
+    const namespaceStatus = await getNamespaceStatus({
+      ctx: k8sCtx,
+      log: garden.log,
+      skipCreate: true,
+      provider: k8sCtx.provider,
+    })
+    expect(namespaceStatus.state).to.eql("ready")
+
+    const runResultCache = await getRunResultCache(ctx.gardenDirPath)
+    await runResultCache.clear({
```
stefreak, line 187: Why are we clearing the cache here? Should we maybe just disable cache in the action config if we're not interested to test cache behaviour here?

stefreak: Ah the clear cache was already present from the start right, only ended up here after the refactorings?

core/src/types/namespace.ts
vvagaytsev: Simplified in 4196bad667701e8c6ade78803568e09575185c50.

core/test/integ/src/plugins/kubernetes/container/run.ts
vvagaytsev: Legacy test, it was there before. We can change the test configs the follow-up PR(s).

core/test/integ/src/plugins/kubernetes/test-results.ts
vvagaytsev: Ah, right, my bad. Restored that in e4c711b350b1c1c33a47987ee61f5bcd04bc32ba.
Introduced an explicit type for the action result.

core/src/plugins/kubernetes/results-cache-base.ts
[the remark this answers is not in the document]
vvagaytsev: Some logging changes were done in 16e4f27aa29b8050b1ecc5710a846c2e86c805b8. Should we increase the visibility of the log message for some scenarios?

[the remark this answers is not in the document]
vvagaytsev: And should we invalidate the value with the broken payload here? Otherwise, we'll always get a cache miss here.

stefreak (approved):
Awesome work!

core/src/plugins/kubernetes/results-cache-base.ts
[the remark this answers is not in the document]
vvagaytsev: Let's revisit the logging and error handling in the follow-up PR(s).

PR #6955, opened 2025-03-18
chore: address 0.14 breaking changes the unit and integration test suite

vvagaytsev:

Fixes #


docs/guides/deprecations.md
```diff
@@ -237,13 +237,6 @@ For more information about Jobs, please refer to the [official Kubernetes docume
 
 # Project configuration
```
stefreak, line 239: Small explanation for the changes in this file: We're on main, so this is the cedar deprecation guide => Once we're done with removing the old code branches, this will be the migration guide to `0.15` (codename D....)

I've started removing some code paths here, but the primary focus of this PR was to make all tests pass.

stefreak (approved):

PR #6956, opened 2025-03-18
chore(local-mode): remove `local-mode` feature

vvagaytsev:

Completely removes `--local-mode` for Deploys from Garden.

Fixes #

We might need to raise mode deprecation warnings if the `localMode` configs are used in the configuration files.
For now, the deprecation warning is printed only if the `--local-mode` flag (or its alias `--local`) is used along with `garden deploy` command.

That will be done in a separate PR.


core/src/plugins/container/config.ts
```diff
@@ -33,7 +32,6 @@ import type { OctalPermissionMask } from "../kubernetes/types.js"
 import { templateStringLiteral } from "../../docs/common.js"
 import { syncGuideLink } from "../kubernetes/constants.js"
 import { makeSecret, type Secret } from "../../util/secrets.js"
-import { makeDeprecationMessage } from "../../util/deprecations.js"
```
stefreak, line 36: The deprecation itself can also be removed together with this PR in `deprecations.ts`

stefreak: Ah, let's do that later

stefreak (dismissed):

vvagaytsev:
Some tests are failing, fixing those now.

core/src/plugins/container/config.ts
vvagaytsev: Deleted in d24f710933d1fc49719f337e39b7ecb76d297171

stefreak (approved):

PR #6960, opened 2025-03-19
feat(k8s)!: action results caching in Garden Cloud V1

vvagaytsev:

Fixes #

### DIFF
--- a/core/src/plugins/kubernetes/results-cache-base.ts
+++ b/core/src/plugins/kubernetes/results-cache-base.ts
@@ -38,6 +38,7 @@ export type SchemaVersion = `v${number}`
 export const currentResultSchemaVersion: SchemaVersion = "v1"
 export const kubernetesCacheEntrySchema = runResultSchemaZod
 export type KubernetesCacheEntrySchema = typeof kubernetesCacheEntrySchema
+export type KubernetesCacheEntry = z.output<KubernetesCacheEntrySchema>
 
 export interface LoadResultParams<A extends CacheableAction, AdditionalKeyData> {
   ctx: PluginContext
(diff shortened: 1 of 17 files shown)


core/src/plugins/kubernetes/results-cache-base.ts
```diff
@@ -70,38 +72,46 @@ export class StructuredCacheKey<AdditionalKeyData> {
...
+  get(key: string, action: Action): Promise<JsonObject>
 
   /**
    * Stores the value associated with the {@code key}.
    *
    * Returns the value back if it was written successfully,
    * or throws a {@link CacheStorageError} otherwise.
    */
-  put(key: string, value: JsonObject): Promise<JsonObject>
+  put(key: string, value: ResultShape, action: Action): Promise<ResultShape>
```
stefreak, line 94: It's simpler for this use case to just make the method generic, and not the entire class.
```suggestion
  put<ResultShape>(key: string, value: ResultShape, action: Action): Promise<ResultShape>
```

vvagaytsev: This type must be consistent with the schema injected to the `ResultCache`.
Changing this requires extra refactoring and does not seem to be a blocker. Let's revisit this later.

core/src/plugins/kubernetes/results-cache.ts
```diff
@@ -25,42 +35,65 @@ type TestKeyData = undefined
...
-    }
+  }
+
+  ctx.log.verbose(
+    deline`
+      You are not logged in to Garden Cloud.
+      The ${styles.highlight("Run")} and ${styles.highlight("Test")} actions
+      of ${styles.highlight("kubernetes")} type will be cached locally.
+      `
+  )
```
stefreak, line 84: This is not a blocker for now, but the place where we nudge the user to log in (`enforceLogin`) should explain that

stefreak (approved):

stefreak:
Awesome work! The implementation of the log messages does not adhere to the acceptance criteria yet, but I assume that will be done in a follow up PR.

core/src/plugins/kubernetes/results-cache.ts
vvagaytsev: Sure, I'll finish that part in the next PR.

PR #6961, opened 2025-03-19
feat(cloud): handle organization IDs from new API

thsig:

Added support for organization IDs in project configurations, and updated the Cloud API & login logic accordingly.

I'm not very happy with the way this came out (with the optional projectId and organizationId in certain types where we could be more specific and therefore safe), but I ran out of time refactoring the types and classes to enable that.


core/src/cloud/grow/util.ts
```diff
@@ -6,19 +6,7 @@
...
-}
-
-export type GrowCloudLogSectionName = "garden-cloud-v2"
-
-export function getGrowCloudLogSectionName(): GrowCloudLogSectionName {
-  return "garden-cloud-v2"
-}
+import { DEFAULT_GARDEN_CLOUD_DOMAIN, gardenEnv } from "../../constants.js"
 
 export function getGrowCloudDomain(configuredDomain: string | undefined): string {
```
vvagaytsev, line 11: Do we still need this? It looks like it's imported only once and is never used.

stefreak: no, can be eliminated, I seem to have overlooked this one.

core/src/cloud/backend.ts
```diff
@@ -124,6 +135,7 @@ export class GrowCloudBackend extends AbstractGardenBackend {
       extractAuthToken: (query) => {
         const token = growCloudTokenSchema.safeParse(query)
         if (!token.success) {
+          // TODO: Better error handling
```
stefreak, line 138:
```suggestion
```

core/test/unit/src/garden.ts
```diff
@@ -854,149 +830,12 @@ describe("Garden", () => {
           `)
           expect(error).to.exist
           expect(error!.message).to.eql("Response code 404 (Not Found)")
-          expect(scope.isDone()).to.be.true
+          expect(scope.isDone(), "not all APIs have been called").to.be.true
         })
       })
 
-      // This means the logged in user is on the community edition
```
stefreak, line 861: Community edition doesn't exist anymore, that's why I've removed these tests.

core/src/cloud/api.ts
```diff
@@ -386,13 +401,30 @@ export class GardenCloudApi {
...
       retry: retry !== false, // defaults to true unless false is explicitly passed
       retryDescription,
       maxRetries,
     })
+
+    // NOTE: Here we detect that user is trying to use `id` with the new backend.
+    if ("error" in res) {
+      // The API returned status code 200 with an error response, so this is a response from tRPC server. That's an indicator that we're talking to the new backend.
+      throw new ConfigurationError({
+        // TODO(0.14): provide a link to the "connecting to Garden Cloud" section in the 0.14 docs
```
vvagaytsev, line 416: This doc link is not yet ready, right?

stefreak: I don't think so, let's add it later. The message is very useful as-is, just at some later point it might be even better to provide a link to the docs.

vvagaytsev (approved):
Amazing work! 🚀

PR #6972, opened 2025-03-22
(0.13) fix: reduce scope of the `build` deprecation to the `container` type

stefreak:
- fix: reduce scope of the \`build\` deprecation to the \`container\` action type
- chore: clean up the apiVersion and deprecation code
- test: fix test failure in prepareProjectResource

Fixes the module backwards-compatibility when using `apiVersion: garden.io/v2` in the project-level configuration.

Best to review the two commits separately.

### DIFF
--- a/core/src/plugins/container/container.ts
+++ b/core/src/plugins/container/container.ts
@@ -56,7 +56,7 @@ import { actionReferenceToString } from "../../actions/base.js"
 import { RootLogger } from "../../logger/logger.js"
 import { styles } from "../../logger/styles.js"
 import { reportDeprecatedFeatureUsage } from "../../util/deprecations.js"
-import { getProjectApiVersion } from "../../project-api-version.js"
+import { getGlobalProjectApiVersion } from "../../project-api-version.js"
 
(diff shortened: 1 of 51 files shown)


core/src/plugins/container/container.ts
```diff
@@ -715,6 +719,30 @@ function validateRuntimeCommon(action: Resolved<ContainerRuntimeAction>) {
...
+    const log = RootLogger.getInstance().createLog()
+    const configPath = action.configPath()
+    // Report concrete action name for better UX
+    log.warn(
+      deline`Action ${styles.highlight(action.longDescription())}
+          ${configPath ? `- defined in ${styles.highlight(configPath)} -` : ""}
+          specifies deprecated config field ${styles.highlight("build")}.`
+    )
+    // Report general deprecation warning
+    reportDeprecatedFeatureUsage({
```
stefreak, line 732: For the reviewers: this will make using the `build` field fail when using `apiVersion: garden.io/v2`

thsig (approved):

PR #6977, opened 2025-03-23
chore: improved log copy around login & cache

thsig:

Various tweaks to log copy around action caching and the `login` command.

### DIFF
--- a/core/src/plugins/kubernetes/run-results.ts
+++ b/core/src/plugins/kubernetes/run-results.ts
@@ -12,8 +12,8 @@ import { toActionStatus } from "./util.js"
 import { getNamespaceStatus } from "./namespace.js"
 import type { KubernetesPluginContext } from "./config.js"
 import type { KubernetesRunResult } from "../../plugin/base.js"
-import { printEmoji } from "../../logger/util.js"
 import { renderSavedTime } from "./results-cache-base.js"
+import { styles } from "../../logger/styles.js"
 
 // TODO: figure out how to get rid of the any cast here
 export const k8sGetRunResult: RunActionHandler<"getResult", any> = async (params) => {
@@ -28,13 +28,14 @@ export const k8sGetRunResult: RunActionHandler<"getResult", any> = async (params
   const cachedResult = await cache.load({ action, ctx, keyData: { namespaceUid: namespaceStatus.namespaceUid }, log })
 
   if (!cachedResult.found) {
-    log.info(`Garden ${cache.brandName} miss ${printEmoji("❌", log)} Reason: ${cachedResult.notFoundReason}`)
+    const reason = `(reason: ${cachedResult.notFoundReason})`
+    log.info(`No cached result found in ${cache.brandName} ${styles.secondary(reason)}`)
 
     return { state: "not-ready", detail: null, outputs: { log: "" } }
   }
 
   const result = cachedResult.result
-  log.info(`Garden ${cache.brandName} hit ${printEmoji("✅", log)} ${renderSavedTime(result)}`)
+  log.info(styles.success(`Cached result found in ${cache.brandName} ${renderSavedTime(result)}`))
 
   return toActionStatus<KubernetesRunResult>({ ...cachedResult.result, namespaceStatus })
 }
--- a/core/src/plugins/kubernetes/test-results.ts
+++ b/core/src/plugins/kubernetes/test-results.ts
@@ -12,8 +12,8 @@ import { toActionStatus } from "./util.js"
 import { getActionNamespaceStatus } from "./namespace.js"
 import type { KubernetesPluginContext } from "./config.js"
 import type { KubernetesRunResult } from "../../plugin/base.js"
-import { printEmoji } from "../../logger/util.js"
 import { renderSavedTime } from "./results-cache-base.js"
+import { styles } from "../../logger/styles.js"
 
 // TODO: figure out how to get rid of the any cast
 export const k8sGetTestResult: TestActionHandler<"getResult", any> = async (params) => {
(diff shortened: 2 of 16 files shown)


core/src/plugins/kubernetes/test-results.ts
```diff
@@ -37,7 +37,7 @@ export const k8sGetTestResult: TestActionHandler<"getResult", any> = async (para
   })
 
   const result = cachedResult.result
-  log.info(`Garden ${cache.brandName} hit ${printEmoji("✅", log)} ${renderSavedTime(result)}`)
+  log.info(`Result found in ${cache.brandName} hit ${printEmoji("✅", log)} ${renderSavedTime(result)}`)
```
stefreak, line 40: What does a "result" mean to our users?
It makes sense given our internal terminology, but I'm not sure the meaning of this term is clear for users.

thsig: I thought it made sense given that the final log string looks something like "Result found in Team Cache [...]" (i.e. caches contain results).

stefreak: It's just my opinion, I think the user regards the result not the logs generated and success status, which is what we consider the result to be, but the actual image that has been uploaded to the registry for instance (We'll also support build kinds later)

thsig:
Made some more tweaks to the cache-related log text.

Before:

Screenshot 2025-03-23 at 15 24 38

After:

Screenshot 2025-03-23 at 15 21 04

core/src/plugins/kubernetes/test-results.ts
thsig: Yeah, makes sense. I'm kind of 50/50 on this. I suggest we revisit this after the launch.

What do you think @edvald @eysi09?

edvald: I prefer the updated copy. "Garden Team Cache Hit" feels like very strange phrasing to me. I'd rather get a description of what happened, as opposed to a sequence of four nouns :) At the very least "Garden" is completely superfluous. Agree with @thsig that we can revisit later if it feels unclear to anyone.

eysi09: I personally think we should go for: `Cached result found in Team Cache`. I know that it says "cache" twice but imo it's the most legible.

eysi09: ...or: `Result already exists in Team Cache`

core/src/plugins/kubernetes/run-results.ts
```diff
@@ -28,13 +28,14 @@ export const k8sGetRunResult: RunActionHandler<"getResult", any> = async (params
...
 
   if (!cachedResult.found) {
-    log.info(`Garden ${cache.brandName} miss ${printEmoji("❌", log)} Reason: ${cachedResult.notFoundReason}`)
+    const reason = `(reason: ${cachedResult.notFoundReason})`
+    log.info(`No cached result found in ${cache.brandName} ${styles.secondary(reason)}`)
 
     return { state: "not-ready", detail: null, outputs: { log: "" } }
   }
 
   const result = cachedResult.result
```
eysi09, line 37: We shouldn't print the saved time if the `--force` flag is set. 

Screenshot 2025-03-23 at 16 03 27

core/src/plugins/kubernetes/test-results.ts
```diff
@@ -37,7 +38,7 @@ export const k8sGetTestResult: TestActionHandler<"getResult", any> = async (para
   })
 
   const result = cachedResult.result
-  log.info(`Garden ${cache.brandName} hit ${printEmoji("✅", log)} ${renderSavedTime(result)}`)
+  log.info(styles.success(`Cached result found in ${cache.brandName} ${renderSavedTime(result)}`))
```
eysi09, line 41: Same as above, we shouldn't render saved time if the `--force` flag is set. (Assuming we know that here, even if it's a framework concept).

core/src/plugins/kubernetes/run-results.ts
```diff
@@ -28,13 +28,14 @@ export const k8sGetRunResult: RunActionHandler<"getResult", any> = async (params
...
-  log.info(`Garden ${cache.brandName} hit ${printEmoji("✅", log)} ${renderSavedTime(result)}`)
+  log.info(styles.success(`Cached result found in ${cache.brandName} ${renderSavedTime(result)}`))
```
eysi09, line 38: We shouldn't print the saved time if the `--force` flag is set. 

Screenshot 2025-03-23 at 16 03 27

eysi09 (changes requested):

core/src/plugins/kubernetes/run-results.ts
stefreak: With `--force` set, we shouldn't even look it up in the cache. Why are we even running the status handlers?

In a later iteration, we'll move some of the responsibility to the framework level, and then the framework-messages can be aware of the cache.

thsig: Providing the `--force` flag to the plugin handlers would be a bit of a refactor which we'd then want to revert once we solve this properly.

I suggest we fix this after the release.

vvagaytsev: I overlooked that part while manual testing with `--force` flag on Friday.
@stefreak right, we should not call the handlers if we force the action executions.

The cache-lookups with the `--force` flag are still the case in 0.13. There `k8sGetRunResult` and `k8sGetTestResult` handlers do the lookups to the ConfigMap-based cache, but do not log anything on the info level.

Here, we also do unnecessary cache lookups, but log it. Anyway, we rerun the actions and do not return the cached result.
So, the semantical behaviour is correct, but the logging is very confusing now.

stefreak: I have the feeling not calling the status handlers might be a bigger refactor, due to the way the event system is implemented and works at the moment

stefreak: Stupid idea: Maybe we can pass action log instance with a fixed debug loglevel to the status handler, in case the `--force` has been used

thsig: > Stupid idea: Maybe we can pass action log instance with a fixed debug loglevel to the status handler, in case the `--force` has been used

That would work!

thsig: @stefreak @eysi09 I've updated the PR to this effect. So I think we should be good to go here.

eysi09 (dismissed):

eysi09 (approved):

PR #6978, opened 2025-03-23
docs: change docs baseurl to latest version

stefreak:
I think we should consider prefixing all docs with the version number,
and forwarding from / to the latest version. @eysi do you think that's
possible?
Otherwise older versions will always link to 404 when we restructure
docs, and it's kind of a hassle to change this constant back and forth
between launches

### DIFF
--- a/core/src/constants.ts
+++ b/core/src/constants.ts
@@ -53,7 +53,7 @@ export const SUPPORTED_ARCHITECTURES: NodeJS.Architecture[] = ["x64", "arm64"]
 export const SEGMENT_DEV_API_KEY = "D3DUZ3lBSDO3krnuIO7eYDdtlDAjooKW" // ggignore
 export const SEGMENT_PROD_API_KEY = "b6ovUD9A0YjQqT3ZWetWUbuZ9OmGxKMa" // ggignore
 
-export const DOCS_BASE_URL = "https://docs.garden.io"
+export const DOCS_BASE_URL = "https://docs.garden.io/cedar-0.14"
 
 export const DEFAULT_GARDEN_CLOUD_DOMAIN = "https://beta.app.garden.io"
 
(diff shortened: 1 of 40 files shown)


eysi09:
Yeah I agree, was thinking something similar. 

So basically the `BASE_URL` would always include the version right?

eysi09:
Ok so I've updated the docs settings so that https://docs.garden.io/cedar-0.14 redirects to https://docs.garden.io. 

Is that what you have in mind?

core/src/constants.ts
```diff
@@ -53,7 +53,7 @@ export const SUPPORTED_ARCHITECTURES: NodeJS.Architecture[] = ["x64", "arm64"]
 export const SEGMENT_DEV_API_KEY = "D3DUZ3lBSDO3krnuIO7eYDdtlDAjooKW" // ggignore
 export const SEGMENT_PROD_API_KEY = "b6ovUD9A0YjQqT3ZWetWUbuZ9OmGxKMa" // ggignore
 
-export const DOCS_BASE_URL = "https://docs.garden.io/bonsai-0.13"
+export const DOCS_BASE_URL = "https://docs.garden.io"
```
eysi09, line 56: We can now change this to `"https://docs.garden.io/cedar-0.14"`

See: https://github.com/garden-io/garden/pull/6978#issuecomment-2746330890

eysi09 (changes requested):

stefreak:
I had it in mind the other way around, to redirect from / => cedar-0.14 but that also works for the purposes of this PR, thank you!

eysi09:
> 0.14

Ok, we can always change this. Just saw a quick and easy way to do this with Gitbook this way around.

eysi09 (approved):

PR #6979, opened 2025-03-23
feat(framework): automatically infer action dependencies from references of `Build` actions outputs

thsig:

### DIFF
--- a/core/test/unit/src/actions/action-configs-to-graph.ts
+++ b/core/test/unit/src/actions/action-configs-to-graph.ts
@@ -406,58 +406,6 @@ describe("actionConfigsToGraph", () => {
     ])
   })
 
-  it("does not mark an implicit dependency needing execution if a static output of dependency is referenced", async () => {
-    const graph = await actionConfigsToGraph({
-      garden,
-      log,
-      groupConfigs: [],
-      configs: parseTemplateCollection({
-        value: [
-          {
-            kind: "Build",
-            type: "container",
-            name: "foo",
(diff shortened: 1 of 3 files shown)


stefreak (approved):

stefreak:
Nice work! @thsig

thsig:
... fixed a few typos in the commit message, my bad! Could you re-approve this?

eysi09 (approved):

eysi09 (approved):

stefreak (dismissed):

stefreak:
@thsig test-plugins is a flake, but `test-framework` seems to be a real test failure in `actionConfigsToGraph`

core/test/unit/src/actions/action-configs-to-graph.ts
```diff
@@ -406,58 +406,6 @@ describe("actionConfigsToGraph", () => {
     ])
   })
 
-  it("does not mark an implicit dependency needing execution if a static output of dependency is referenced", async () => {
```
stefreak, line 409: @thsig Do we have a test that checks that we don't add a dependency when referencing static outputs of deploy actions?

stefreak (approved):
Nice, thanks

core/test/unit/src/actions/action-configs-to-graph.ts
stefreak: Yes, just read your test again and we do 👍

PR #6980, opened 2025-03-23
chore: clear deprecation guide and remove dead code

stefreak:
Removed the dead code and cleared the deprecations, to clear the way for what's coming in 0.15 :)


core/src/docs/generate.ts
```diff
@@ -201,6 +202,8 @@ export async function writeConfigReferenceDocs(
   await renderConfigTemplate("render-template", renderConfigReference(renderTemplateConfigSchema()))
 }
 
+// NOTE: this is only temporarily unused, while we don't have any deprecations
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 async function updateDeprecationGuide(docsRoot: string, deprecationGuideFilename: string) {
```
vvagaytsev, line 207: We can also export this function. In that case, the linter won't complain even if the function is not used.

vvagaytsev (approved):
Great!!! 🚀

PR #6981, opened 2025-03-23
(cherry pick #6979) feat(framework): automatically infer action dependencies from references of Build actions outputs

thsig:

improvement(core): fixes to output reference logic

We only do this for API version v2 (and keep the old behavior for v1 and v0).


thsig:
Deleted this one because there was a typo in the branch name.

PR #6983, opened 2025-03-23
(cherry pick #6979) improvement(core): fixes to output reference logic

thsig:

We only do this for API version v2 (and keep the old behavior for v1 and v0).


core/test/setup.ts
```diff
@@ -65,7 +65,7 @@ export const mochaHooks = {
 
   beforeEach() {
     // Init globally stored project-level apiVersion, assuming garden.io/v1 for 0.13.
-    setGloablProjectApiVersion(GardenApiVersion.v1)
+    setGlobalProjectApiVersion(GardenApiVersion.v1)
```
stefreak, line 68: Good catch 😆 thanks

stefreak (approved):
Awesome!

PR #6985, opened 2025-03-23
docs: remove deleted plugin paths from `gitbook.yaml`

vvagaytsev:

Fixes #


eysi09 (approved):

eysi09:
We're sort of doing a hard reset on docs and I honestly think we can remove most of these redirects. The original links (which are probably buried in way old blogs that no one reads anymore) would still take the user to the docs site.

PR #6987, opened 2025-03-24
chore: remove `hotReload` from schemas

vvagaytsev:

Fixes #

### DIFF
--- a/core/src/plugins/container/config.ts
+++ b/core/src/plugins/container/config.ts
@@ -677,8 +677,6 @@ export const containerDeploySchemaKeys = memoize(() => ({
     .description("List of ingress endpoints that the service exposes.")
     .example([{ path: "/api", port: "http" }]),
   healthCheck: healthCheckSchema().description("Specify how the service's health should be checked after deploying."),
-  // TODO: remove in 0.14, keeping around to avoid config failures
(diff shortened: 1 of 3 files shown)


stefreak (approved):
Good catch!

stefreak (changes requested):
Did we have a deprecation warning and error in 0.13 with apiVersion garden.io/v2 for this one? If not I'm afraid we need to remove it in 0.15

vvagaytsev:
> Did we have a deprecation warning and error in 0.13 with apiVersion garden.io/v2 for this one? If not I'm afraid we need to remove it in 0.15

No, we did not. So, technically, it's a breaking change that has not been yet properly deprecated.

There is no code behind that, and the relevant `serviceResouce.hotReloadArgs` was removed in 0.13.
But some users miught still have these garbage `hotReload` entries in the configs.
So, yes, we can just add properly deprecate this in 0.14 and remove in 0.15.

stefreak:
Yes I think that would be better, as it would improve the user experience for those who are affected 👍

vvagaytsev:
Agree, let's make the migration process as smooth as possible :)
I added todos to deprecate the old `hotReload` config syntax properly in #6989.

PR #6990, opened 2025-03-24
chore(local-mode) deprecate `localMode` configuration syntax

vvagaytsev:

The local-mode feature was deprecated in 0.13 only on a command level, see #6827.
The `spec.localMode` in configuration files was not deprecated properly.

Despite the local-mode feature was completely removed in #6956 and is no longer available in 0.14, let's keep the ability to use `spec.localMode` in the config files for better UX. Such configuration entries have no effect and will be reported via deprecation warnings. Let's keep them for compatibility to ensure smooth 0.13 -> 0.14 migration process.

Fixes #

Tested manually by running `deploy` command with local-mode examples (container, k8s, helm) from 0.13 branch.


core/src/plugins/container/config.ts
```diff
@@ -318,6 +319,78 @@ export const containerSyncPathSchema = createSchema({
   rename: [["sync", "paths"]],
 })
 
+const defaultLocalModeRestartDelayMsec = 1000
+const defaultLocalModeMaxRestarts = Number.POSITIVE_INFINITY
+
+export const localModeRestartSchema = createSchema({
```
stefreak, line 325: let's mark local mode schemas as internal, so they're not added to the docs by the docs generator.

core/src/plugins/container/container.ts
```diff
@@ -539,7 +540,11 @@ export const gardenPlugin = () =>
           staticOutputsSchema: containerDeployOutputsSchema(),
           handlers: {
             // Other handlers are implemented by other providers (e.g. kubernetes)
-            async configure({ config }) {
+            async configure({ config, log }) {
+              if (config.spec["localMode"]) {
+                reportDeprecatedFeatureUsage({ log, deprecation: "localMode" })
```
stefreak, line 545: maybe we can even report deprecated feature usage in the schema itself in this case?

vvagaytsev: How can we do that?

stefreak (approved):
Nice work!

PR #6995, opened 2025-03-24
chore: remove 0.13 dead code

vvagaytsev:

Fixes #


core/src/template/ast.ts
```diff
@@ -537,16 +535,6 @@ export class FormatStringExpression extends TemplateExpression {
     const apiVersion = getGlobalProjectApiVersion()
     const isOptional = this.isOptional(apiVersion)
 
-    // TODO(0.14.1): remove this dead code branch - we no longer support the apiVersion v1
-    if (apiVersion === GardenApiVersion.v1 && isOptional) {
```
stefreak, line 541: nice catch! Then, `isOptional` method and `hasOptionalSuffix` constructor argument need to be deleted as well, right? And syntax updated so it does not provide the constructor arg?

vvagaytsev: Right! Did that in fd20a2a89cca91df2de6cf0a9089608b289500d2 and ae645fbb235fe057da9347ca839d19eed6e90a8c.

stefreak (approved):

PR #6964, opened 2025-03-20
fix(cloud): remove unnecessary error translation inside the tRPC client

vvagaytsev:
**What this PR does / why we need it**:

Remove unnecessary error translation inside the tRPC client.

This PR also refactors the Cloud API V2 class by making its `api` field (a reference to the underlying tRPC client) private:
* The API calls are implemented as methods of the class.
* The methods delegate to the underlying tRPC client and do the error handling.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

PR #6998, opened 2025-03-24
fix(cloud): remove unnecessary error translation inside the tRPC client

vvagaytsev:

Back-port #6964 to 0.13.

Fixes #


stefreak:
The tRPC client isn't supposed to be used in the 0.13 track, it was basically only there for development purposes behind an environment variable feature flag.

PR #7027, opened 2025-03-26
fix(module-k8s): fix `deploymentImageId` in module conversion

vvagaytsev:

Back-port #7026 to 0.13.


vvagaytsev:
Used the wrong branch.

PR #7029, opened 2025-03-26
fix(module-k8s): fix `deploymentImageId` in module conversion

vvagaytsev:

Back-port #7026 to 0.13.

### DIFF
--- a/core/src/plugins/container/container.ts
+++ b/core/src/plugins/container/container.ts
@@ -443,7 +443,7 @@ function convertContainerModuleRuntimeActions(
       timeout: service.spec.timeout || DEFAULT_DEPLOY_TIMEOUT_SEC,
       spec: {
         ...omit(service.spec, ["name", "dependencies", "disabled"]),
-        image: needsContainerBuild ? undefined : deploymentImageId,
+        image: deploymentImageId,
         volumes: [],
       },
     }
@@ -465,7 +465,7 @@ function convertContainerModuleRuntimeActions(
 
       spec: {
         ...omit(task.spec, ["name", "description", "dependencies", "disabled", "timeout"]),
-        image: needsContainerBuild ? undefined : deploymentImageId,
+        image: deploymentImageId,
         volumes: [],
       },
     }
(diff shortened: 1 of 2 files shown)


vvagaytsev:
In 0.13 we still set `build` field on runtime actions in the module converter. So, it looks like we don't need to back-port the fix from 0.14.

stefreak:
@vvagaytsev this means we need to backport both this fix, and the code change of not setting the `build` field. Otherwise module users can't use `apiVersion: garden.io/v2`

stefreak:
You can verify what I'm saying by deploying one of the module e2e tests from the main branch (using `apiVersion: garden.io/v2`) using the latest version of Garden 0.13

stefreak:
This is something that I've missed to backport to 0.13 on the weekend, apparently.

vvagaytsev:
> You can verify what I'm saying by deploying one of the module e2e tests from the main branch (using `apiVersion: garden.io/v2`) using the latest version of Garden 0.13

I wasn't able to reproduce the e2e test failures with that setup. I think it's only reproducible when we use `container` _provider_ with the `container` action type.

The e2e tests use `kubernetes` provider that has its own `validate` handler which does not call the base provider's (i.e. `container` providers's) `validate` handler. Thus, the `validate` handler of the `kubernetes` provider never calls the [`validateRuntimeCommon` function](https://github.com/garden-io/garden/blob/0.13/core/src/plugins/container/container.ts#L720) that performs the validation of `build` and `spec.image` fields.

I attempted to delegate to the base validation handler in #6606, but that caused some regression, and the changes were reverted in #6641.

So, we deprecated the `build` field on the runtime actions of the `container` _action type_, should we hold the same behaviour for all implementing providers (i.e. `container`, `local-kubernetes` and `kubernetes`)?

Maybe we should take a closer look at this.

vvagaytsev:
I've made some extra code changes to back-port the semantics to 0.13. So, maybe we can merge this as it is and revisit the [comment above](https://github.com/garden-io/garden/pull/7029#issuecomment-2788999309) separately.

vvagaytsev:
@stefreak this is ready for review now.

stefreak (approved):
Really nice! Good work!

PR #7030, opened 2025-03-26
fix(k8s): update to nginx ingress controller 1.12.1 that is not affected by RCE vulnerability

vvagaytsev:
**What this PR does / why we need it**:

This PR updates the default `ingress-nginx` to version `1.12.1` in order to fix the [CVE-2025-1974](https://nvd.nist.gov/vuln/detail/CVE-2025-1974).

PR #7031, opened 2025-03-26
fix(k8s): update to nginx ingress controller 1.12.1 that is not affected by RCE vulnerability

vvagaytsev:

Back-port #7030 to 0.13.


stefreak (approved):
Nice

PR #7047, opened 2025-03-31
chore: update `eslint` dependencies

vvagaytsev:

Update `eslint` and relevant plugins to the latest versions, and modernize the relevant configurations.

Supersedes #6677.

The original PR was not mergeable because of the breaking changes.

~See my comment below, there is one open question about the dandling promises in the `sync stop` command.~
That warn was suppressed in this PR and addressed as a fix in #7056.

Commit-by-commit review might be an easier way to go. The most of the changes are `package-lock.json` updates.


core/src/commands/sync/sync-stop.ts
```diff
@@ -114,7 +114,7 @@ export class SyncStopCommand extends Command<Args, Opts> {
         await router.deploy.stopSync({ log: actionLog, action, graph })
 
         // Halt any active monitors for the sync
-        garden.monitors.find({ type: "sync", key: action.name }).map((m) => m.stop())
+        await Promise.all(garden.monitors.find({ type: "sync", key: action.name }).map((m) => m.stop()))
```
vvagaytsev, line 117: It might be better to ignore this rule here and make it as a separate fix PR.

core/src/commands/self-update.ts
```diff
@@ -70,8 +70,8 @@ const selfUpdateOpts = {
 export type SelfUpdateArgs = typeof selfUpdateArgs
 export type SelfUpdateOpts = typeof selfUpdateOpts
 
-const versionScopes = ["major", "minor", "patch"] as const
-export type VersionScope = (typeof versionScopes)[number]
+const _versionScopes = ["major", "minor", "patch"] as const
```
vvagaytsev, line 73: The rule `no-unused-vars` is stricter now, see https://typescript-eslint.io/rules/no-unused-vars/#why-does-this-rule-report-variables-used-only-for-types

core/src/commands/sync/sync-stop.ts
vvagaytsev: This fix can also be back-ported to 0.13.

vvagaytsev: This was addressed in #7056.

.eslintrc.bak
eysi09, line 1: Should we remove this file?

eslint.config.mjs
```diff
@@ -0,0 +1,155 @@
...
+      "mocha/no-skipped-tests": "warn",
+      "mocha/no-exclusive-tests": "error",
+      "unused-imports/no-unused-imports": "error",
+
+      "header/header": [
+        2,
+        "block",
+        [
+          "",
+          " * Copyright (C) 2018-2024 Garden Technologies, Inc. <info@garden.io>",
```
eysi09, line 129: Can we update this to 2025? (Not sure if that should be a different PR, but noticed you updated the header of this file).

But let's update in a follow up PR if this impacts all headers.

eysi09 (changes requested):
Great job! Only requested change would be to remove the `.bak` file.

eslint.config.mjs
vvagaytsev: This was updated to 2025 in 04da1805c1aee8af1e68e404245dca2c23dc873c.

.eslintrc.bak
vvagaytsev: Ah, sure! Good catch! I overlooked that :) Removed it in b3d7df2dddfe5768c9ba45f316e36195c2d24827.

eysi09 (approved):

vvagaytsev:
@eysi09 can you please review this again? I resolved the conflicts and fixes the lint errors.

thsig (approved):

PR #7062, opened 2025-04-03
fix: do not throw on unknown action types for disabled actions

### LINKED ISSUE #6957: [Bug]: Disabled action causing "Unrecognized action type" error
### Current behavior

Garden fails with an `Unrecognized action type` error when running in an environment that doesn't have the required provider, even though the action in question is disabled.

### Expected behavior

Disabled actions should not cause errors.

### Reproducible example

First, create a project level Garden config file with the following config:

```yaml
apiVersion: garden.io/v1
kind: Project
name: my-project
environments:
  - name: no-k8s
  - name: k8s

providers:
  - name: local-kubernetes
    namespace: env-repro
    environments: [k8s]
---
kind: Deploy
name: k8s-deploy
type: kubernetes
environments: ["k8s"] # also fails with disabled: true
spec:
  manifests:
    - apiVersion: v1
      kind: ConfigMap
      metadata:
        name: my-configmap
      data:
        foo: bar
---
kind: Run
name: say-hi
type: exec
spec:
  command: ["echo", "Hi ${local.username}"]
```

Then run `garden run say-hi --env no-k8s` and notice that it fails because the `no-k8s` environment doesn't have a K8s provider which the `k8s-deploy` action requires, even though that one is disabled.

### Workaround

Configure all required providers in all envs, which may not be viable. 

### Suggested solution(s)

My naive first suggestion would be to check if the action is disabled in [`processActionConfig` function](https://github.com/garden-io/garden/blob/main/core/src/graph/actions.ts#L430) and skip it if it is.

And/Or alternatively filter out disabled actions before calling that function. 

### Your environment

Running of latest `main` as of writing, using Orbstack for local K8s.

vvagaytsev:

We can avoid throwing an error in `processActionConfig` if the action is disabled.

Disabled actions with configured provider pass the error-check defined in `processActionConfig`. So, this PR ensures that the disabled actions without configured providers also pass that check.
Anyway, the disabled actions will not be executed, so there is no harm is bypassing that error-check.

At that point we know the resolved value of the `disabled` flag on the action config, so we can use it to identify disabled actions.

Fixes #6957

### DIFF
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -443,7 +443,7 @@ export const processActionConfig = profileAsync(async function processActionConf
 
   const configPath = relative(garden.projectRoot, config.internal.configFilePath || config.internal.basePath)
 
-  if (!actionTypes[kind][type]) {
+  if (!actionTypes[kind][type] && !config.disabled) {
     const availableKinds: ActionKind[] = []
     actionKinds.forEach((actionKind) => {
       if (actionTypes[actionKind][type]) {
--- a/core/test/data/test-projects/disabled-action-without-provider/garden.yml
+++ b/core/test/data/test-projects/disabled-action-without-provider/garden.yml
@@ -1,0 +1,32 @@
+apiVersion: garden.io/v2
+kind: Project
+name: my-project
+environments:
+  - name: no-k8s
+  - name: k8s
+
+providers:
+  - name: local-kubernetes
+    namespace: env-repro
+    environments: [k8s]
+
+---
+kind: Deploy
+name: k8s-deploy
+type: kubernetes
+disabled: true
+environments: ["k8s"] # also fails with disabled: true
+spec:
+  manifests:
+    - apiVersion: v1
+      kind: ConfigMap
+      metadata:
+        name: my-configmap
+      data:
+        foo: bar
+---
+kind: Run
+name: say-hi
+type: exec
+spec:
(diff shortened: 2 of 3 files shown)


core/test/data/test-projects/disabled-action-without-provider/garden.yml
```diff
@@ -0,0 +1,32 @@
...
+  - name: local-kubernetes
+    namespace: env-repro
+    environments: [k8s]
+
+---
+kind: Deploy
+name: k8s-deploy
+type: kubernetes
+disabled: true
+environments: ["k8s"] # also fails with disabled: true
```
eysi09, line 18:
```suggestion
environments: ["k8s"]
```

eysi09 (changes requested):

core/src/graph/actions.ts
```diff
@@ -443,7 +443,7 @@ export const processActionConfig = profileAsync(async function processActionConf
 
   const configPath = relative(garden.projectRoot, config.internal.configFilePath || config.internal.basePath)
 
-  if (!actionTypes[kind][type]) {
+  if (!actionTypes[kind][type] && !config.disabled) {
```
stefreak, line 446: Are we also taking into account the `environments` config field here? It should behave the same as the `disabled` flag.

vvagaytsev: Fixed that in #7069.

PR #7064, opened 2025-04-04
docs: minor corrections in deprecations guide

vvagaytsev:

Fixes #

### DIFF
--- a/core/src/util/deprecations.ts
+++ b/core/src/util/deprecations.ts
@@ -41,9 +41,9 @@ export function getDeprecations(style: (s: string) => string = styles.highlight)
     },
     hotReloadArgs: {
       docsSection: "Old configuration syntax",
-      docsHeadline: `${style("serviceResource.hotReloadArgs")} configuration field in modules`,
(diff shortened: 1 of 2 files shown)


thsig (approved):

core/src/util/deprecations.ts
```diff
@@ -41,9 +41,9 @@ export function getDeprecations(style: (s: string) => string = styles.highlight)
     },
     hotReloadArgs: {
       docsSection: "Old configuration syntax",
```
stefreak, line 43: Maybe "module configuration" would be a less ambiguous section headline here?

Also, I noticed that the deprecation guide shows up under "guides" with "Old configuration syntax" as a headline here, as the markdown file is lacking a header and introduction text. (See https://docs.garden.io/guides/deprecations) – we should put it under misc, like the cedar migration guide, and we could call it "Deprecations" and also explain how `apiVersion` works and that updating to `apiVersion: garden.io/v3` is experimental and not recommended yet at this point etc.

vvagaytsev: Thank you! Good point!

Created #7088 as a follow-up fix. Let's continue the discussion there :)

PR #7069, opened 2025-04-07
fix(graph): use a special helper to check if an action is disabled

vvagaytsev:

The helper ensures the correct and consistent behavior, because the action can be disabled not only explicitly by `disabled: true` flag, but also implicitly over the environment configuration.

Follow-up fix for #7062.

### DIFF
--- a/core/test/data/test-projects/disabled-action-without-provider/garden.yml
+++ b/core/test/data/test-projects/disabled-action-without-provider/garden.yml
@@ -12,9 +12,9 @@ providers:
 
 ---
 kind: Deploy
-name: k8s-deploy
+name: k8s-deploy-disabled-via-flag
 type: kubernetes
-disabled: true
+disabled: true # the action is disabled explicitly in all envs
 environments: ["k8s"]
 spec:
   manifests:
@@ -24,6 +24,7 @@ spec:
         name: my-configmap
       data:
         foo: bar
+
 ---
 kind: Run
 name: say-hi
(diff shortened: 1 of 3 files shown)


core/test/data/test-projects/disabled-action-without-provider/garden.yml
```diff
@@ -12,9 +12,9 @@ providers:
 
 ---
 kind: Deploy
-name: k8s-deploy
+name: k8s-deploy-disabled-via-flag
 type: kubernetes
-disabled: true
+disabled: true # the action is disabled explicitly in all envs
 environments: ["k8s"]
```
stefreak, line 18: Specifying environments doesn't have an effect together with `disabled: true`

stefreak (approved):
Nice, thanks for following up on this and writing a test 👍  Good work

core/test/data/test-projects/disabled-action-without-provider/garden.yml
vvagaytsev: Yes, I know. Prior to this fix Garden was failing with this configuration. So, despite the action was completely disabled, the action type lookup took place and an error was throw if the provider was not configured.

vvagaytsev: It might make sense to add an explanatory comment on that. But it's not a blocker. We can do that later :)

vvagaytsev: Right! We should at least cover 2 independent ways of disabling the action:
* only disabled flag
* only env config

The third scenario (both flag and env config) looks redundant, but we can maybe also keep it. I'll open a follow-up PR with the updated test.

Good catch! Thank you :+1:

stefreak: Sounds good, thanks for your work around this 👏

PR #7123, opened 2025-04-14
fix(docker): improve error handling in Docker builds

vvagaytsev:

* Makes the code less confusing by removing `throw` and `return` statements from one `finally` block.
* Makes the error reporting more precise and user-friendly.
* Applies some refactoring.

Before this PR, a configuration error with hints would not be thrown.


core/src/plugins/container/build.ts
```diff
@@ -254,38 +260,57 @@ async function buildxBuildContainer({
           Learn more at https://docs.docker.com/go/build-multi-platform/
         `,
       })
-    } else if (buildError.message.includes("failed to push")) {
-      buildError = new ConfigurationError({
+    } else if (messageLowercase.includes("failed to push")) {
+      dockerBuildError = new ConfigurationError({
         message: dedent`
           The Docker daemon failed to push the image to the registry.
```
stefreak, line 266: @vvagaytsev did you try if this error message works? I wonder how it'd work now, as we requested rawjson logs from the Docker CLI. Maybe we need to look at the output of the docker log transformer binary instead?

vvagaytsev: Changed this in c543312d0f95dcf6cbbf14745c2b7025aa70cac1 and 840971732245a6e9b2f5220808c3e34937e49283.

eysi09 (approved):

PR #7142, opened 2025-04-17
chore: fix `deprecated` and `internal` flags propagation in reference docs 

vvagaytsev:

This PR fixes the propagation of `deprecated` and `internal` meta flags of joi objects to their children.

Now we don't need to explicitly set `.meta({ internal: true })` on every child of an already internal joi schema to hide it from the generated docs.
The same fix applies to the `deprecated` meta flag.

There are also some type-safety improvements.

Fixes #

### DIFF
--- a/docs/reference/commands.md
+++ b/docs/reference/commands.md
@@ -1724,24 +1724,6 @@ providers:
         # The module spec, as defined by the provider plugin.
         spec:
 
-            # POSIX-style filename to write the resolved file contents to, relative to the path of the module source
-            # directory (for remote modules this means the root of the module repository, otherwise the directory of
-            # the module configuration).
-            #
-            # Note that any existing file with the same name will be overwritten. If the path contains one or more
-            # directories, they will be automatically created if missing.
-            targetPath:
-
-            # By default, Garden will attempt to resolve any Garden template strings in source files. Set this to
-            # false to skip resolving template strings. Note that this does not apply when setting the `value` field,
-            # since that's resolved earlier when parsing the configuration.
-            resolveTemplates:
-
-            # The desired file contents as a string.
-            value:
-
-            sourcePath:
-
         # The name of the parent module (e.g. a templated module that generated this module), if applicable.
         parentName:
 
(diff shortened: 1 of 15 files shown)


docs/reference/template-strings/workflows.md
```diff
@@ -291,14 +291,6 @@ Example:
 my-variable: ${git.originUrl}
 ```
 
-### `${secrets.<secret-name>}`
```
stefreak, line 294: Seems off that we're removing secrets from the reference guide here?

docs/reference/commands.md
```diff
@@ -1724,24 +1724,6 @@ providers:
...
-            # By default, Garden will attempt to resolve any Garden template strings in source files. Set this to
-            # false to skip resolving template strings. Note that this does not apply when setting the `value` field,
-            # since that's resolved earlier when parsing the configuration.
-            resolveTemplates:
-
-            # The desired file contents as a string.
-            value:
-
-            sourcePath:
-
```
stefreak, line 1744: Why does this block of code get removed?

docs/reference/template-strings/workflows.md
vvagaytsev: For some reason it was marked as `internal` long time ago. It might be better to re-declare it as non-internal instead.

docs/reference/commands.md
vvagaytsev: This block was a part of the `generatedFiles` schema. The parent `generatedFiles` block is not displayed as it was declared as internal. Here we have orphan children of that block.

The parent-to-child propagation of `deprecated` and `internal` flags were broken.

vvagaytsev: And the definition for the spec schema is `spec: joi.object().meta({ extendable: true })` , so it has no children.
here, it looked like these fields were part of the spec schema. but in fact they weren't.

docs/reference/template-strings/workflows.md
vvagaytsev: Fixed that in a088f0921cc52bb4aece661ecc69b3459e605c4e.

stefreak (approved):
This is great! I've also attempted to fix this one once, but then I gave up. Amazing to see you managed to push through! 💪

PR #7143, opened 2025-04-17
chore: type-safe deprecation flags in joi schemas

vvagaytsev:

Fixes #


core/src/config/common.ts
```diff
@@ -82,7 +83,13 @@ interface MetadataKeys {
   // Flag as an advanced feature, to be advised in generated docs
   advanced?: boolean
   // Flag as deprecated. Set to a string to provide a deprecation message for docs.
+  /**
+   * Deprecated is deprecated :P
+   * @deprecated use {@link #deprecation} flag instead
+   */
   deprecated?: boolean | string
+  // Flag as deprecated. Set to automatically generate a message for docs.
+  deprecation?: Deprecation
```
vvagaytsev, line 92: Not all schemas are used by reference docs or in the user input and configuration validation.
In such cases, we can use just `true` as a value, because we do not need a deprecation message.

Thus, we can re-declare this as `deprecation?: Deprecation | true` to allow declaring a user-visible `Deprecation` for internally used schemas.

Alternatively, the old and new flags can be collapsed to a discriminated union like `{ internal: true, deprecated: true} | { internal: false, deprecation: Deprecation}`. That might look clearer and more understandable, and prevent us from using both flags simultaneously.

core/src/config/common.ts
```diff
@@ -76,13 +77,19 @@ export const includeGuideLink = makeDocsLinkPlain(
...
-interface MetadataKeys {
+export interface MetadataKeys {
   // Unique name to identify in error messages etc
   name?: string
   // Flag as an advanced feature, to be advised in generated docs
   advanced?: boolean
   // Flag as deprecated. Set to a string to provide a deprecation message for docs.
+  /**
+   * deprecated is deprecated :P
+   * @deprecated use {@link #deprecation} flag instead
```
stefreak, line 88: How many times is deprecated still used? Can we maybe eliminate it in this PR, rather than having the confusing state of both of them? It makes it more difficult to review or reason about this change for me

core/src/config/common.ts
vvagaytsev: We still have some legacy deprecations in the format of `deprecated: "some string value"`. First we need to transform them to the new format of `Deprecation` type. We can merge this once the rest deprecations are converted.

core/src/config/common.ts
vvagaytsev: We still have some legacy deprecations in the format of `deprecated: "some string value"`. First we need to transform them to the new format of `Deprecation` type.

We can merge this once the rest deprecations are converted.

PR #7168, opened 2025-04-23
fix(container): always respect env vars to control remote builder

vvagaytsev:

Prior to this this, the env variable `GARDEN_CONTAINER_BUILDER` and its deprecated alias `GARDEN_CLOUD_BUILDER` were only taken into account when the corresponding provider level `gardenContainerBuilder` (or deprecated `gardenCloudBuilder`) configurations were explicitly defined in the `garden.yml` config file.

Now, the override via env vars still takes precedence over the file-based config, but it's also possible to avoid any configs in file and use only env vars.

**Expected behavior:**
```console
GARDEN_CLOUDBUILDER=1 garden build 
```
The flag enables Remote Container Builder.
See the [0.13 docs](https://docs.garden.io/bonsai-0.13/reference/providers/container#providers-.gardencontainerbuilder.enabled):

> by running `GARDEN_CONTAINER_BUILDER=1` garden build you can try Garden Container Builder temporarily without any changes to your Garden configuration.

**Current behavior:**
```console
GARDEN_CLOUDBUILDER=1 garden build 
```
The flag is ignored and Remote Container Builder isn't used.

The flag `GARDEN_CONTAINER_BUILDER` works as expected in 0.14.

### DIFF
--- a/core/src/plugins/container/cloudbuilder.ts
+++ b/core/src/plugins/container/cloudbuilder.ts
@@ -360,6 +360,35 @@ class CloudBuilder {
 
 export const cloudBuilder = new CloudBuilder()
 
+/**
+ * The env variables GARDEN_CONTAINER_BUILDER (and its deprecated alias GARDEN_CLOUD_BUILDER)
+ * can be used to override the {@code gardenContainerBuilder.enabled} config setting.
+ *
+ * It will be {@code undefined}, if the variables are not set,
+ * and {@code true}/{@code false} if GARDEN_CONTAINER_BUILDER=1 or GARDEN_CONTAINER_BUILDER=0 respectively.
+ */
+function getOverrideFromEnv(ctx: PluginContext): boolean | undefined {
+  const overrideFromEnv = gardenEnv.GARDEN_CONTAINER_BUILDER || gardenEnv.GARDEN_CLOUD_BUILDER
+
+  if (overrideFromEnv === true) {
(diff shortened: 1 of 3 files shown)


core/src/plugins/container/cloudbuilder.ts
```diff
@@ -360,6 +360,35 @@ class CloudBuilder {
...
+ *
+ * It will be {@code undefined}, if the variables are not set,
+ * and {@code true}/{@code false} if GARDEN_CONTAINER_BUILDER=1 or GARDEN_CONTAINER_BUILDER=0 respectively.
+ */
+function getOverrideFromEnv(ctx: PluginContext): boolean | undefined {
+  const overrideFromEnv = gardenEnv.GARDEN_CONTAINER_BUILDER || gardenEnv.GARDEN_CLOUD_BUILDER
+
+  if (overrideFromEnv === true) {
+    const varName = gardenEnv.GARDEN_CONTAINER_BUILDER === true ? "GARDEN_CONTAINER_BUILDER" : "GARDEN_CLOUD_BUILDER"
+    ctx.log.debug(`Garden Container Builder is enabled via env variable ${varName}`)
```
vvagaytsev, line 375: We still use "Garden Container Builder" term in the other log and error messages. Should we rename that in 0.13 to "Remote Container Builder" as we did it in 0.14?

stefreak: It's fine to leave it as-is, but it's also ok to rename it.

stefreak (approved):
Good catch!

PR #7171, opened 2025-04-24
fix: prevent unnecessary execution of builds referenced in runtime actions

vvagaytsev:

Fixes #

### DIFF
--- a/core/src/graph/actions.ts
+++ b/core/src/graph/actions.ts
@@ -946,6 +946,7 @@ function dependenciesFromActionConfig({
 
   for (const ref of getActionTemplateReferences(config, templateContext)) {
     let needsExecuted = false
+    let isExplicit = false
 
     const outputType = ref.keyPath[0]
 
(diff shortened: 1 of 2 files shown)


core/src/graph/actions.ts
```diff
@@ -977,10 +978,13 @@ function dependenciesFromActionConfig({
...
       // (i.e. the dependency should be processed) before the runtime resource is processed (i.e. deployed or run).
       // Note: We could also always execute Test actions that are referenced, but we'll stick with only Builds for now.
-      if (!isString(outputKey) || refActionKind === "Build") {
+      if (!isString(outputKey)) {
         needsExecuted = true
       } else {
         needsExecuted = !staticOutputKeys.includes(outputKey) && !refStaticOutputKeys.includes(outputKey)
+        if (refActionKind === "Build") {
+          isExplicit = true
+        }
```
stefreak, line 987: I have the feeling that this if statement should be outside of the if/else block that looks at the type of `outputKey`. I think there is an explicit dependency relationships for builds, even when the output key can't be resolved at this point.

core/src/graph/actions.ts
```diff
@@ -977,10 +978,13 @@ function dependenciesFromActionConfig({
       // resource (like a `helm` Deploy), where the user intent is almost always that the referenced build should exist
       // (i.e. the dependency should be processed) before the runtime resource is processed (i.e. deployed or run).
       // Note: We could also always execute Test actions that are referenced, but we'll stick with only Builds for now.
```
stefreak, line 980: I think we should rework the comments, theres multiple things happening here so maybe we have multiple different comment blocks at different locations, focussing on each thing that's happening separately?

stefreak (approved):
Awesome! Great work

stefreak:
@vvagaytsev This needs to be back-ported to 0.13 as well

PR #7185, opened 2025-04-28
chore: disable npm package publishing in release pipeline

vvagaytsev:

It has been failing starting from 0.13.24: https://www.npmjs.com/package/@garden-io/core

Back-port #7180 to 0.13. Not a critical change, but it will prevent unnecessary GH job run which is know to be failed.

Fixes #


vvagaytsev:
The CI failure is fixed in #7186. This need to be rebased when the fix is merged.

thsig (approved):

PR #7231, opened 2025-05-13
Deprecations-mention-variables-behavioral-change

stefreak:

Fixes #


core/src/util/deprecations.ts
```diff
@@ -165,6 +167,33 @@ export function getDeprecations(style: (s: string) => string = styles.highlight)
         #!/bin/bash
         echo "hello $\${USERNAME:=world}" # <-- This is not a working bash script anymore :(
         \`\`\`
+
+        ### Errors when variables do not exist
+
+        When using \`apiVersion: garden.io/v1\` in your project-level configuration, Garden will not throw errors when referencing variables or secrets that do not exist in Kubernetes manifest template files.
+
+        When using \`apiVersion: garden.io/v2\` in the project-level configurationm Garden will always throw an error when secrets or variables do not exist. If the secret or variable is meant to be optional, you can provide a default value using the logical or operator (Example: \`${var.enable_xyz || false}\`).
```
vvagaytsev, line 175:
```suggestion
        When using \`apiVersion: garden.io/v2\` in the project-level configuration Garden will always throw an error when secrets or variables do not exist. If the secret or variable is meant to be optional, you can provide a default value using the logical or operator (Example: \`${var.enable_xyz || false}\`).
```

core/src/util/deprecations.ts
```diff
@@ -165,6 +167,33 @@ export function getDeprecations(style: (s: string) => string = styles.highlight)
...
+        # manifests/cookie-config.yml
+        apiVersion: v1
+        kind: ConfigMap
+        metadata:
+          name: cookie-config
+        data:
+            cookie-encryption-key: ${secrets.COOKIE_ENCRYPTION_KEY}
+        \`\`\`
+
+        In older versions of Garden or when using  \`apiVersion: garden.io/v1\` in your project-level configuration, the plain string \`${secrets.COOKIE_ENCRYPTION_KEY}\` might end up being used as an insecure encryption key, if the secret named \`COOKIE_ENCRYPTION_KEY\` did not exist.
```
vvagaytsev, line 194:
```suggestion
        In older versions of Garden or when using \`apiVersion: garden.io/v1\` in your project-level configuration, the plain string \`\${secrets.COOKIE_ENCRYPTION_KEY}\` might end up being used as an insecure encryption key, if the secret named \`COOKIE_ENCRYPTION_KEY\` did not exist.
```

core/src/util/deprecations.ts
```diff
@@ -165,6 +167,33 @@ export function getDeprecations(style: (s: string) => string = styles.highlight)
...
+
+
+        \`\`\`yaml
+        # manifests/cookie-config.yml
+        apiVersion: v1
+        kind: ConfigMap
+        metadata:
+          name: cookie-config
+        data:
+            cookie-encryption-key: ${secrets.COOKIE_ENCRYPTION_KEY}
```
vvagaytsev, line 191:
```suggestion
        cookie-encryption-key: \${secrets.COOKIE_ENCRYPTION_KEY}
```

vvagaytsev (changes requested):
Looks good! 👍 
Left some suggestions - otherwise the build fails.

vvagaytsev (approved):
I've addressed the comments that I left earlier :)
Let's merge this.

PR #7234, opened 2025-05-13
chore: shadow launch of first implementation of gRPC-based event transport

vvagaytsev:

Fixes #


core/src/cloud/grow/buffered-event-stream.ts
```diff
@@ -120,31 +162,158 @@ export class GrowBufferedEventStream {
...
+    }
+
+    this.log.debug(() => `GrowBufferedEventStream: Flushing ${this.eventBuffer.size} events from the buffer`)
+    for (const event of this.eventBuffer.values()) {
+      if (!this.outputStream) {
+        this.log.debug(() => `GrowBufferedEventStream: Stream closed during flush`)
+        break
+      }
+      await this.outputStream.write(event)
+    }
```
stefreak, line 317: I think we might not want to await here, as this can lead to out-of-order delivery when new events arrive while the flush is ongoing.

core/src/cloud/grow/buffered-event-stream.ts
```diff
@@ -0,0 +1,319 @@
...
+      (async () => {
+        await sleep(10000)
+        this.log.warn(
+          "Not all events were acknowledged within 10 seconds. Information in Garden Cloud may be incomplete."
+        )
+      })(),
+    ])
+
+    this.outputStream?.close()
+    this.closed = true
```
stefreak, line 153: maybe needed to reorder these two lines, to avoid reconnect

core/src/cloud/grow/buffered-event-stream.ts
```diff
@@ -0,0 +1,319 @@
...
+      }
+    }
+    this.log.root.events.onAny(this.logListener)
+
+    this.eventListener = async (name, payload) => {
+      this.handleEvent(name, payload)
+    }
+    this.garden.events.onAny(this.eventListener)
+
+    setTimeout(async () => {
```
stefreak, line 98: I think it'd be nice if we can live without this hack, by calling the `loop` method explicitly from the call site (e.g. explicitly using `Promise.all` somewhere in the command or CLI implementation for concurrent work items)

stefreak: This would ensure that we're handling errors correctly and they don't end up as an unhandled exception.

core/src/cloud/grow/buffered-event-stream.ts
```diff
@@ -0,0 +1,319 @@
...
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  private handleLogEntry(logEntry: LogEntryEventPayload) {
+    // TODO: logs handling
+  }
+
+  private async streamEvents() {
+    this.outputStream = createWritableIterable<GrpcEvent>()
+
+    const ackStream = this.client.ingestEventStream(this.outputStream)
```
stefreak, line 268: Instead of calling this "client" it might be a more useful mental model if we imagine just calling the method on the service directly.
```suggestion
    const ackStream = this.eventService.ingestEventStream(this.outputStream)
```

images/support-images.garden.yml
```diff
@@ -1,6 +1,7 @@
 apiVersion: garden.io/v2
 kind: Project
 name: garden-support-images
+organizationId: 0dbe175b-0894-46e5-a18f-82f25d2d232a
```
stefreak, line 4: Accidentally committed this, needs to be removed before we merge.

core/src/server/instance-manager.ts
```diff
@@ -160,16 +161,17 @@ export class GardenInstanceManager {
...
           plugins: this.plugins,
           ...this.defaultOpts,
           ...garden?.opts,
           ...opts,
           ...params,
+          // The sessionId should be the same as the surrounding process.
+          // For each command run, this will be set as the parentSessionId,
+          // and the command-specific Garden (cloned in `Command.run()`) gets its own sessionId.
+          sessionUlid: this.sessionUlid,
+          parentSessionUlid: null,
```
stefreak, line 174: I wonder if this should be 
```suggestion
          parentSessionUlid: this.parentSessionUlid,
```
Probably a theoretical question..

vvagaytsev: I'll take a look

core/src/cloud/grow/buffered-event-stream.ts
stefreak:
```suggestion
    this.closed = true
    this.outputStream?.close()
```

core/src/server/instance-manager.ts
vvagaytsev: There is no field `parentSessionUlid` defined in `GardenInstanceManager`, so the `null` looks ok here.

core/src/server/commands.ts
```diff
@@ -438,11 +441,11 @@ export async function resolveRequest({
     args: cmdArgs,
     opts: cmdOpts,
     environmentString: request.environment,
-    sessionUlid: sessionId,
+    sessionUlid,
   })
 
   cmdLog.context.gardenKey = garden.getInstanceKey()
-  cmdLog.context.sessionId = sessionId
+  cmdLog.context.sessionId = sessionUlid
```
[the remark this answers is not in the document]
vvagaytsev: Fixed this and other places in 6e3dd23e54e3ce6464054fca888500a2165ed2d5.

core/test/unit/src/server/server.ts
```diff
@@ -465,7 +466,11 @@ describe("GardenServer", () => {
             return
           }
           const logEntries = messages.filter((m) => m.type === "logEntry" && m.context.gardenKey === gardenKey)
-          const sessionLogEntries = logEntries.filter((m) => m.context.sessionId === id)
+          const sessionLogEntries = logEntries.filter((m) => {
+            const sessionUuid = m.context.sessionId as string
+            // both are UUIDs, but the case can be different
```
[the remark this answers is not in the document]
vvagaytsev: The `uuidv4` is coming from the test data in the [L459](https://github.com/garden-io/garden/pull/7234/files/fe46b39a58f1b8e62a842151aa1c861da5e65fa8..ae5866eb994a044119fa542dfb788e3467bcd00f#diff-51ebcd832892b5b9dc8cb79afa462079eedf716af460e76399747e40a344a023R459).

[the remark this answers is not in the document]
vvagaytsev: The server validates it to be `uuidv4`.

core/src/cli/command-line.ts
```diff
@@ -688,7 +688,8 @@ ${styles.accent.underline("Keys:")}
     args: PrepareParams["args"]
     opts: PrepareParams["opts"]
   }) {
-    const id = uuidv4()
+    // Do not confuse this with the concept of a "command id" that is used in the new events model
+    const runningCommandTrackingId = uuidv4()
```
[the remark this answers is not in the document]
vvagaytsev: Right! This id is redundant.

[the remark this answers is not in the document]
vvagaytsev: Removed it in 476e4fbbc7b93651fc2c357ad52fd26cfb8cfe92.

core/src/cloud/grow/buffered-event-stream.ts
```diff
@@ -0,0 +1,377 @@
...
+      this.outputStream?.close()
+      this.outputStream = undefined
+    }
+  }
+
+  private async consumeAcks(ackStream: AsyncIterable<EventResponse>) {
+    for await (const nextAck of ackStream) {
+      if (!nextAck.success) {
+        // NOTE: We expect the server to also close the stream in case of an error, so we can retry emitting events on the next attempt.
+        this.log.silly(`Server failed to process event with ulid=${nextAck.eventId}`)
```
stefreak, line 336:
```suggestion
        this.log.silly(`Server failed to process event with ulid=${nextAck.eventId}`)
        continue // NOTE: We expect the server to also close the stream in case of an error, but let's first receive all outstanding acks and errors.
```

core/src/server/commands.ts
```diff
@@ -428,7 +430,8 @@ export async function resolveRequest({
 
   const cmdLog = serverLogger.createLog({})
 
-  const sessionId = request.id || uuidv4()
+  const sessionUuid = request.id || uuidv4()
+  const sessionUlid = uuidToULID(sessionUuid)
```
stefreak, line 434: We should never represent uuidv4 in ULID's crockford encoding. Conversion might even fail at runtime.

stefreak: Maybe we should simply log a debug-level message that this particular request id triggered session with ulid=...

vvagaytsev: Right, that should work.

We can generate a valid `sessionUlid` here and pass it to `GardenInstanceManager.getGardenForRequest`; and use the `sessionUuid` only in the log context.

vvagaytsev: Changed this in 2551ca694996e0f12f945ae7594c9d6f92e00e2c.

By the way, do we want to use monotonic UILDs as session ids?

stefreak: It's relatively unlikely that we're starting two commands within the same millisecond, within the same garden CLI instance, and even if we're doing that I don't see the use case for having the monotonic ordering. But if we see a need we can do that 👍

stefreak (approved):

core/src/cloud/grow/buffered-event-stream.ts
vvagaytsev: Should we address this before we merge the PR?

core/src/cloud/grow/buffered-event-stream.ts
```diff
@@ -0,0 +1,469 @@
...
+        this.log.silly("GrowBufferedEventStream: Connecting ...")
+
+        try {
+          await this.streamEvents()
+        } catch (err) {
+          if (err instanceof ConnectError) {
+            this.log.silly(`GrowBufferedEventStream: Error while streaming events: ${err}`)
+            this.log.silly("GrowBufferedEventStream: Retrying in 1 second...")
+            await sleep(1000)
+          } else if (!isNewBackendEventSystemInProduction) {
```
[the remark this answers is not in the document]
vvagaytsev: If the backend with events support is not in prod, we'll fail with a `ConnectError` here, so this branch will not be reached because of the `else if` clause.

core/src/config/template-contexts/project.ts
```diff
@@ -198,9 +198,9 @@ class VcsContext extends ContextWithSchema {
 
   constructor(info: VcsInfo) {
     super()
-    this.branch = info.branch
-    this.commitHash = info.commitHash
-    this.originUrl = info.originUrl
+    this.branch = info.branch || ""
+    this.commitHash = info.commitHash || ""
+    this.originUrl = info.originUrl || ""
```
stefreak, line 203: ah, we can roll this back, forgot to revert this one.

core/src/cloud/grow/buffered-event-stream.ts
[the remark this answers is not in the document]
stefreak: That else if is not about the ConnectError, but actually about reachability of the else clause: I wanted to avoid throwing unexpected errors (crashes) before this is in production. Maybe it's clearer if we comment out code instead of using that const.

stefreak (approved):

core/src/cloud/grow/buffered-event-stream.ts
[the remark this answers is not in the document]
stefreak: Changed it to comments

stefreak (dismissed):

LarsFronius (approved):

PR #7239, opened 2025-05-16
fix(cloud): handle tRPC errors in authentication functions

### LINKED ISSUE #7238: Crash: `TRPCClientError: fetch failed`
### Error message

```
Deploy 🚀

ℹ garden                    → Initializing...
ℹ garden-enterprise         → Authorizing...

You can help by reporting this on GitHub: https://github.com/garden-io/garden/issues/new?labels=bug,crash&template=CRASH.md&title=Crash%3A%20TRPCClientError%3A%20fetch%20failed

TRPCClientError: fetch failed
    at TRPCClientError.from (file:///Users/magnus/Projects/garden/core/node_modules/@trpc/client/dist/TRPCClientError.mjs:53:16)
    at file:///Users/magnus/Projects/garden/core/node_modules/@trpc/client/dist/links/httpLink.mjs:79:52
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
```

### What did you do?

after successfully logging in with `GARDEN_CLOUD_DOMAIN="https://<docker-compose dev instance>/"` g login, when I try to run `GARDEN_CLOUD_DOMAIN="https://<docker-compose dev instance>/"` g deploy I just get a crash. It should indicate the SSL verification failed instead, and not crash.

### Your environment

* **OS:** macOS
* **Garden version:** edge-cedar

### Frequency

Reproducable with docker-compose based dev backend

### Workaround

None (`NODE_TLS_REJECT_UNAUTHORIZED=0` should never be used for non-development purposes!)

vvagaytsev:

* Handle tRPC errors on calls via non-authenticated client
* Align auth error handling in between old and new backends
* Less confusing logging on logout

```console
You're already logged out from https://grow.garden.io.
Successfully logged out from https://grow.garden.io.
```
is now just
```console
You're already logged out from https://grow.garden.io.
```
and the sucessful message is printed only on successful token revocation.

Fixes #7238

### DIFF
--- a/core/src/cloud/grow/trpc.ts
+++ b/core/src/cloud/grow/trpc.ts
@@ -87,8 +87,10 @@ export function isAggregateError(err: Error): err is AggregateError {
   return err.constructor.name === "AggregateError"
 }
 
-export function describeTRPCClientError(err: TRPCClientError<InferrableClientTypes>) {
-  const errorDescriptions: string[] = []
+type ErrorCause = { type: string; msg: string }
+
+export function getErrorCauseChain(err: TRPCClientError<InferrableClientTypes>): ErrorCause[] {
+  const errorCauses: ErrorCause[] = []
 
   let currentError: Error | undefined = err
   while (currentError !== undefined) {
@@ -100,7 +102,7 @@ export function describeTRPCClientError(err: TRPCClientError<InferrableClientTyp
           .join("; ")
       : currentError.message
 
-    errorDescriptions.push(`${errorType}: ${errorMsg}`)
+    errorCauses.push({ type: errorType, msg: errorMsg })
 
     if (isError(currentError.cause)) {
       currentError = currentError.cause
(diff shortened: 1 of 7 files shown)


vvagaytsev:
Some tests failed. I converted it back to draft.

core/src/cloud/grow/trpc.ts
```diff
@@ -80,3 +82,32 @@ export function getNonAuthenticatedApiClient(trpcConfigParams: Omit<TrpcConfigPa
 }
 
 export type ApiClient = CreateTRPCClient<AppRouter>
+
+export function isAggregateError(err: Error): err is AggregateError {
+  return err.constructor.name === "AggregateError"
+}
+
+export function describeTRPCClientError(err: TRPCClientError<InferrableClientTypes>) {
```
stefreak, line 90: Do you have some examples how some of these errors now look in practice? E.g. how does https://github.com/garden-io/garden/issues/7238 now get displayed?

vvagaytsev: @stefreak yes, it now looks like this:
```console
An error occurred while refreshing client auth token with Garden Enterprise: TRPCClientError: fetch failed; caused by: TypeError: fetch failed; caused by: Error: unable to verify the first certificate
Please try again.
```

Without the rendering function it would be just:
```console
An error occurred while refreshing client auth token with Garden Enterprise: fetch failed
Please try again.
```

Here the actual error message comes from the bottommost cause.

vvagaytsev: Printing the whole chain of the causes might look too technical, so maybe we should just print it on the debug level, and print a shortened variant in the error message.
Wdyt?

stefreak: Hm, maybe we can only display the nested message if the outer message is "fetch failed"? or we arrange the messages like so:

`fetch failed: unable to verify the first certificate`

vvagaytsev: I made it less verbose in the error messages in d72ef9df2bce7dd19875c5a14dfd5e26fd5e4148.
Now it satisfies the shape above (constructor names are omitted, and the error message chain is deduplicated).

The debug-level logs still print the detailed error description, that is calculated lazily.

stefreak (approved):
Awesome work!

PR #7272, opened 2025-06-02
refactor(logs): get rid of `ts-stream` library

vvagaytsev:

Fixes #


stefreak:
In the future, we should consider refactoring how we handle logs and events to slow down consumption when we see backpressure in the event sink (i.e. when the backend can't handle the logs). Otherwise we'll always be prone to be unreliable due to OOM issues etc.

PR #7319, opened 2025-06-11
chore: deprecate `podSelector` in sync mode

vvagaytsev:

This PR makes deprecates the usage of `podSelector` in sync mode.
The code was changed to throw a config error if only `podSelector` is specified as a target resource for sync mode.

That `podSelector` did not work properly prior to this fix and could lead to either a runtime error or to a non-working sync mode with warning. Throwing a meaningful configuration error looks like a better UX here.

Fixes #


eysi09 (approved):
Nice, looks good to me. 

 I

PR #7380, opened 2025-06-23
chore(deps): update `json-stable-stringify`

vvagaytsev:

* Update `json-stable-stringify` to 1.3.0
* Fix types
* Fix test config stringification in the module configs

Supersedes #7302

### DIFF
--- a/core/src/types/test.ts
+++ b/core/src/types/test.ts
@@ -52,12 +52,12 @@ export function testFromConfig<M extends GardenModule = GardenModule>(
     recursive: true,
   })
   // We sort the dependencies by type and name to avoid unnecessary cache invalidation due to possible ordering changes.
-  const depHashes = [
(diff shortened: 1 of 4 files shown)


core/src/types/test.ts
```diff
@@ -52,12 +52,12 @@ export function testFromConfig<M extends GardenModule = GardenModule>(
     recursive: true,
   })
   // We sort the dependencies by type and name to avoid unnecessary cache invalidation due to possible ordering changes.
-  const depHashes = [
-    ...sortBy(deps.build, (mod) => mod.name).map((mod) => mod.version),
+  const depConfigs = [
```
vvagaytsev, line 55: This function is only called from [`ModuleGraph.getTests`](https://github.com/garden-io/garden/blob/main/core/src/graph/modules.ts#L387) which, in turn, is only used in tests.

So, this change does not fix any real bug or change the current behaviour.

PR #7381, opened 2025-06-23
chore(dev-deps): update typescript to 5.8.3

vvagaytsev:

Fixes #


scripts/install-osx-dependencies.sh
```diff
@@ -26,7 +26,7 @@ nvm alias default v22
 nvm use default
 
 # install/update global packages
-npm install -g ts-node typescript
+npm install -g typescript
```
vvagaytsev, line 29: We have `tsx` as a dev dependency in the root-level `package.json`.

PR #7417, opened 2025-06-24
chore(renovate): group PRs by package manager and version scope

vvagaytsev:

* Group minor and patches dependencies for npm and cargo package managers separately.
* Always combine major updates for dev dependencies.
* Ignore updates for copied libs like `fsevents`

The goal is to split the dependencies by language/package manager and to minimize the number of produced PRs.
We still keep individual PRs for major updates in prod dependencies, but for dev dependencies it should be fine to combine.

Fixes #

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,7 +3,7 @@
   "extends": [
     "config:recommended"
   ],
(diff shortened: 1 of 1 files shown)


10ko (changes requested):
There are some test failures.

10ko (approved):

PR #7457, opened 2025-07-03
fix: error handling in BiDi gRPC events streaming

vvagaytsev:

* Improve error handling for stream creation, writing and consuming
* Adjust default ack consuming timeout to 1 second

Without these changes Garden CLI would ignore the gRPC auth issues and other I/O and connectivity problems.

Fixes #


edvald (approved):

10ko (approved):

core/src/cloud/grow/grpc-event-stream.ts
```diff
@@ -215,11 +228,14 @@ export class GrpcEventStream {
...
-          `GrpcEventStream: Server failed to process event with ulid=${nextAck.eventUlid}, final=${nextAck.final}: ${JSON.stringify(this.eventBuffer.get(nextAck.eventUlid), (_, v) => (typeof v === "bigint" ? v.toString() : v))}`
+        this.log.debug(
+          `GrpcEventStream: Server failed to process event ulid=${nextAck.eventUlid}, final=${nextAck.final}: ` +
+            `${JSON.stringify(this.eventBuffer.get(nextAck.eventUlid), (_, v) =>
+              typeof v === "bigint" ? v.toString() : v
+            )}`
         )
       } else {
-        this.log.silly(() => `GrpcEventStream: Received ack for event ${nextAck.eventUlid}, final=${nextAck.final}`)
+        this.log.debug(() => `GrpcEventStream: Received ack for event ${nextAck.eventUlid}, final=${nextAck.final}`)
```
stefreak, line 238: @edvald @vvagaytsev the event stream code should not emit any debug logs, as that will lead to massive log amplification / infinite recursion as soon as we send logs from core to the backend (Some action emits a log message, which is transferred to the backend as an event, which emits a log message, which is transferred as an event etc)

stefreak: The use of silly was by design for this reason.

core/src/cloud/grow/grpc-event-stream.ts
```diff
@@ -277,16 +293,25 @@ export class GrpcEventStream {
...
+    // Serialize writes to avoid race/out-of-order issues
+    const writeEvents = async () => {
+      // NOTE: The Map implementation in the javascript runtime guarantees that values will be iterated in the order they were added (FIFO).
+      for (const event of this.eventBuffer.values()) {
+        if (!this.outputStream) {
+          this.log.silly(() => `GrpcEventStream: Stream closed during flush`)
+          break
+        }
+        try {
+          await this.outputStream.write(create(IngestEventsRequestSchema, { event }))
```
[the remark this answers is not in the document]
stefreak: Fix and unit test added here: https://github.com/garden-io/garden/pull/7478

PR #7458, opened 2025-07-03
feat: add preInit.runScript option for all providers

edvald:

Previously, users would need to use the `exec` provider with an `initScript` set on the provider to perform any preparation needed outside of Garden (such as authenticating with clusters). Now this is a more native feature on any provider, just set `preInit.runScript` on any given provider.

### DIFF
--- a/core/src/plugins/exec/exec.ts
+++ b/core/src/plugins/exec/exec.ts
@@ -60,11 +60,19 @@ export const execPlugin = sdk.createGardenPlugin({
 
 export const execProvider = execPlugin.createProvider({
   configSchema: s.object({
-    initScript: s.string().optional().describe(dedent`
-      An optional script to run in the project root when initializing providers. This is handy for running an arbitrary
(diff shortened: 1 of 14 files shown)


core/src/plugins/exec/exec.ts
```diff
@@ -60,11 +60,19 @@ export const execPlugin = sdk.createGardenPlugin({
...
+      .describe(
+        dedent`
+          DEPRECATED: Use the \`preInit.runScript\` field instead on any provider that needs setup outside of Garden.
+
+          An optional script to run in the project root when initializing providers. This is handy for running an arbitrary
+          script when initializing. For example, another provider might declare a dependency on this provider, to ensure
+          this script runs before resolving that provider.
+        `
+      )
+      .setMetadata({ deprecated: true }),
```
stefreak, line 75: @edvald @vvagaytsev should we use the new [deprecations machinery](https://github.com/garden-io/garden/blob/main/core/src/util/deprecations.ts) for this instead? I thought the old boolean `deprecated` flag has been removed?

PR #7480, opened 2025-07-10
fix(examples): update some K8s deploy examples

eysi09:
Before this, we were using `manifestTemplates` were we should've been using `manifestFiles` in some examples.

There were also a couple of dated examples that I replaced with the `k8s-deploy-simple` example.

Fixes #


examples/k8s-deploy-simple/README.md
```diff
@@ -0,0 +1,41 @@
...
+```yaml
+# In garden.yml
+kind: Deploy
+name: db
+type: kubernetes
+spec:
+  manifestFiles: [./postgres-manifests/**/*] # <--- Here we read the manifests from a file
+```
+
+The same patter could also be used for e.g. creating secrets.
```
vvagaytsev, line 30:
```suggestion
The same pattern could also be used for e.g. creating secrets.
```

PR #7498, opened 2025-07-17
refactor: change `cloud` package structure

vvagaytsev:

Separate legacy and new backend-specific components into 2 directories on the same level.

Fixes #


core/src/cloud/backend-base.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export type AuthRedirectConfig = Pick<AuthRedirectServerConfig, "getLoginUrl" | "successUrl" | "extractAuthToken">
+
+export type RevokeAuthTokenParams = {
+  clientAuthToken: ClientAuthToken
+  globalConfigStore: GlobalConfigStore
+  log: Log
+}
+
+export type CloudApiFactory = GardenCloudApiFactory | GrowCloudApiFactory
```
eysi09, line 32: Can we call these `GardenCloudLegacyApiFactory` and `GardenCloudApiFactory` respectively? And in general use `GardenCloudLegacy` and `GardenCloud` when we need to discern between the two? 

Alternatively, if this is a big change and we need to rename a million things from `GardenCloud` to `GardenCloudLegacy`,  can we instead do `GardenCloud` and `GardenCloudV2`? 

Also cc-ing @edvald in case you have opinions, I feel like I sometimes make the wrong call when naming things :)

eysi09: ...as I continue reading I we've already moved a bunch of things under the `legacy` dir which is great.

core/src/cloud/backend-base.ts
```diff
@@ -0,0 +1,59 @@
...
+export interface GardenBackend {
+  config: GardenBackendConfig
+  cloudApiFactory: CloudApiFactory
+
+  getAuthRedirectConfig(): AuthRedirectConfig
+
+  revokeToken(params: RevokeAuthTokenParams): Promise<void>
+}
+
+export abstract class AbstractGardenBackend implements GardenBackend {
```
eysi09, line 43: Just a general question, but is there any benefit in having both Garden Cloud classes (legacy and current) implementing the same abstract class? 

Would it be simpler to just maintain these as separate classes? In particular as they're bound to diverge and one will eventually be phased out.

core/src/cloud/enforce-login.ts
```diff
@@ -0,0 +1,65 @@
...
+  constructor() {
+    super({
+      message: dedent`
+        ${styles.primary(
+          // TODO(0.14): advertise team cache and container builder here as benefit for logging in
+          `Login required: This project is connected to Garden Cloud. Please run ${styles.command("garden login")} to authenticate or set the ${styles.highlight("GARDEN_AUTH_TOKEN")} environment variable.`
+        )}
+
+        ${styles.secondary(
+          `NOTE: If you cannot log in right now, use the option ${styles.command("--offline")} or the environment variable ${styles.command("GARDEN_OFFLINE=true")} to enable offline mode. Team Cache and Container Builder won't be available in the offline mode.`
```
eysi09, line 28:
```suggestion
          `NOTE: If you cannot log in right now, use the option ${styles.command("--offline")} or the environment variable ${styles.command("GARDEN_OFFLINE=true")} to enable offline mode. Team Cache and Container Builder won't be available in offline mode.`
```

core/src/cloud/enforce-login.ts
```diff
@@ -0,0 +1,65 @@
...
+import { getCloudDistributionName, getCloudLogSectionName } from "./util.js"
+
+class LoginRequiredWhenConnected extends GardenError {
+  override type = "login-required"
+
+  constructor() {
+    super({
+      message: dedent`
+        ${styles.primary(
+          // TODO(0.14): advertise team cache and container builder here as benefit for logging in
```
eysi09, line 23:
```suggestion
```

eysi09: I think we can remove this comment since this project is already connected and we mention Team Cache and Container Builder below.

eysi09 (changes requested):
Nice! Had a few comments/questions and a suggestion for the copy. Otherwise 👍

core/src/cloud/backend-base.ts
vvagaytsev: This is a tiny abstraction that we introduced to implement different login machinry.

For the rest APIs call we'll likely have to implement different components behind the conditional logic, because the API contracts can be (and actually are) different.

We can rename this later to emphasize the scope of this abstraction, or embed each part into backend-specific component. This was implemented via the common interface in 0.13, when we had to maintain logging in to both backends from a single CLI version.

core/src/cloud/backend-base.ts
vvagaytsev: Yes, good point! We should also align the naming conventions, because now we use grow/garden, v1/v2, and old/new in the class and function names interchangeably. Let's do that in a follow-up PR.

eysi09 (approved):

PR #7505, opened 2025-07-23
feat(k8s): add pause command to scale workloads to zero

edvald-garden:
This adds a new `garden plugins kubernetes pause -- [names...]` command, to pause all or specified Kubernetes deployments (any of `container`, `kubernetes` or `helm` types), i.e. scale replicas down to zero.


core/src/plugins/kubernetes/commands/pause.ts
```diff
@@ -0,0 +1,179 @@
...
+      statusOnly: true,
+      actionsFilter: deployNames,
+    })
+
+    const deploys = graph
+      .getDeploys({ names: deployNames })
+      .filter((d) => d.isCompatible("container") || d.isCompatible("kubernetes") || d.isCompatible("helm"))
+
+    if (deploys.length === 0) {
+      log.info(`No container, kubernetes or helm deployments found.`)
```
eysi09, line 48: Should we use "Deploy actions" instead of "deployments" here?

And if this refers to Garden actions, how about highlight them with `styles.highlight("container")` etc?

eysi09: ...and the same applies for other usages of "deployment". That is, should we use "Deploy action" as opposed to "deployment" in these contexts?

core/src/plugins/kubernetes/commands/pause.ts
```diff
@@ -0,0 +1,179 @@
...
+    return { result }
+  },
+}
+
+type PausableWorkload =
+  | KubernetesResource<V1Deployment>
+  | KubernetesResource<V1StatefulSet>
+  | KubernetesResource<V1ReplicaSet>
+
+function isPausable(resource: KubernetesResource): resource is PausableWorkload {
```
eysi09, line 145: So I take it DaemonSets are impausable :)

core/test/integ/src/plugins/kubernetes/commands/pause.ts
```diff
@@ -0,0 +1,240 @@
...
+
+  describe("helm", () => {
+    before(async () => {
+      garden = await getHelmTestGarden()
+      await init()
+      await buildHelmModules(garden, graph)
+    })
+
+    after(async () => {
+      // https://app.circleci.com/pipelines/github/garden-io/garden/15885/workflows/6026638c-7544-45a8-bc07-16f8963c5b9f/jobs/265663?invite=true#step-113-577
```
eysi09, line 169: What does this refer to? The link doesn't work for me.

eysi09, line 169: What does this refer to? The link doesn't work for me.

core/src/plugins/kubernetes/commands/pause.ts
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    log.info({ msg: `Getting status for ${deploys.length} deployments` })
+    const router = await garden.getActionRouter()
+    const statuses = await getDeployStatuses(router, graph, log)
+
+    const api = await KubeApi.factory(log, ctx, provider)
+    // const defaultNamespace = provider.outputs["app-namespace"]
+
+    await Promise.all(
```
eysi09, line 59: Should we return some information about the resources that were paused here and include with the results object? Could be a flat list of resource kind, name, and Garden action type/name. Or grouped by actions.

core/src/plugins/kubernetes/commands/pause.ts
```diff
@@ -0,0 +1,179 @@
...
+      log.info(`No container, kubernetes or helm deployments found.`)
+      return { result }
+    }
+
+    log.info({ msg: `Getting status for ${deploys.length} deployments` })
+    const router = await garden.getActionRouter()
+    const statuses = await getDeployStatuses(router, graph, log)
+
+    const api = await KubeApi.factory(log, ctx, provider)
+    // const defaultNamespace = provider.outputs["app-namespace"]
```
eysi09, line 57:
```suggestion
```

eysi09 (changes requested):
Great stuff! Had a couple of minor comments/questions.

PR #7529, opened 2025-07-29
fix(templates): incorrect path resolution for templated modules

edvald-garden:
This addresses an issue where paths for modules generated by module templates weren't correctly set, when the `path` field is set on the module spec in the template.


jondeandres:
💯 🙏🏻

thsig (approved):

PR #7572, opened 2025-08-07
feat(config): add version.excludeValues to action configs

edvald-garden:
This provides a simple way to avoid ephemeral values affecting cache versioning.

From the docs:

-----
Specify one or more string values that should be ignored when computing the version hash for this action. You may use template expressions here. This is useful to avoid dynamic values affecting cache versions.

For example, you might have a variable that naturally changes for every individual test or dev environment, such as a dynamic hostname. You could solve for that with something like this:

```yaml
version:
  excludeValues:
    - ${var.hostname}
```

With the `hostname` variable being defined in the Project configuration.

For each value specified under this field, every occurrence of that string value (even as part of a longer string) will be replaced when calculating the action version. The action configuration is not affected. 

----

### DIFF
--- a/core/src/actions/base.ts
+++ b/core/src/actions/base.ts
@@ -1019,9 +1019,9 @@ const nonVersionedActionConfigKeys = [
 ] as const
 export type NonVersionedActionConfigKey = keyof Pick<BaseActionConfig, (typeof nonVersionedActionConfigKeys)[number]>
 
-const excludeValueReplacement = "!!!GARDEN-EXCLUDED!!!"
+export const excludeValueReplacement = "!!!GARDEN-EXCLUDED!!!"
 
-export function getActionConfigVersion<C extends BaseActionConfig>(log: ActionLog, config: C) {
+export function replaceExcludeValues(config: BaseActionConfig, log: ActionLog) {
   let configToHash: unknown = omit(config, ...nonVersionedActionConfigKeys)
 
   const excludeValues = config.version?.excludeValues || []
@@ -1041,5 +1041,10 @@ export function getActionConfigVersion<C extends BaseActionConfig>(log: ActionLo
     })
   }
 
+  return configToHash
+}
+
+export function getActionConfigVersion<C extends BaseActionConfig>(log: ActionLog, config: C) {
+  const configToHash = replaceExcludeValues(config, log)
   return versionStringPrefix + hashStrings([stableStringify(configToHash)])
 }
--- a/core/test/unit/src/actions/base.ts
+++ b/core/test/unit/src/actions/base.ts
@@ -10,23 +10,29 @@ import { expect } from "chai"
 import { DEFAULT_BUILD_TIMEOUT_SEC } from "../../../../src/constants.js"
 import type { ActionConfig, BaseActionConfig } from "../../../../src/actions/types.js"
 import type { NonVersionedActionConfigKey } from "../../../../src/actions/base.js"
-import { actionIsDisabled, getActionConfigVersion } from "../../../../src/actions/base.js"
+import {
+  actionIsDisabled,
+  excludeValueReplacement,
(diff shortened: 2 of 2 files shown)


core/src/actions/base.ts
```diff
@@ -986,10 +1009,25 @@ const nonVersionedActionConfigKeys = [
...
 export type NonVersionedActionConfigKey = keyof Pick<BaseActionConfig, (typeof nonVersionedActionConfigKeys)[number]>
 
+const excludeValueReplacement = "!!!GARDEN-EXCLUDED!!!"
+
 export function getActionConfigVersion<C extends BaseActionConfig>(config: C) {
-  const configToHash = omit(config, ...nonVersionedActionConfigKeys)
+  let configToHash: unknown = omit(config, ...nonVersionedActionConfigKeys)
+
+  if (config.version?.excludeValues && config.version.excludeValues.length > 0) {
+    configToHash = deepMap(configToHash, (v) => {
```
thsig, line 1022: Since this is a tight loop, let's move the regex instantiation outside the `deepMap`.

Also, would be great to log replaced values (and the key paths to them).

We'd need to pass a `log` param for this (we can get the `keyPath` from the third parameter to the `deepMap` callback) and maybe use `regex.match` to check if there's a match before replacing (this would probably be a bit cheaper than always replacing and then checking if the before & after values are different).

core/src/actions/base.ts
```diff
@@ -257,6 +257,29 @@ export const baseActionConfigSchema = createSchema({
       .example("my-action.env")
       .meta({ templateContext: ActionConfigContext }),
 
+    version: joi
+      .object()
+      .keys({
+        excludeValues: joiSparseArray(joi.string()).description(
```
thsig, line 263: This is definitely easier for the user to work with than the `excudePaths` idea, especially since we can detect the value as a substring in resolved values.

I could imagine this becoming rather slow if used extensively in large projects, but we can always consider introducing `excludePaths` for cases like that (which should be more performant, but might also be more hassle to configure).

core/src/actions/types.ts
```diff
@@ -96,6 +100,9 @@ export interface BaseActionConfig<K extends ActionKind = ActionKind, T = string,
   // -> Templating with ActionConfigContext allowed, including in variables defined in the varfiles
   varfiles?: Varfile[]
 
+  // Version/caching controls
+  version?: ActionVersionConfig
```
thsig, line 104: To avoid confusion, I suggest renaming this field to `versionConfig` while keeping the user-facing name in the config (`version`).

We can use the `rename` option in `createSchema` to map `version` in the user-facing config to `versionConfig` on the internal interface.

core/test/unit/src/garden.ts
```diff
@@ -4356,6 +4356,38 @@ describe("Garden", () => {
       expect(b.isDisabled()).to.be.true
     })
 
+    it("correctly handles version.excludeValues", async () => {
```
thsig, line 4359: Could we add a test case that uses the excluded value as a substring of the value?

thsig (changes requested):
Looks good! Had a few comments there.

core/src/actions/base.ts
edvald-garden: I'm also implementing `excludePaths` for good measure, so users will have options.

core/test/unit/src/garden.ts
edvald-garden: I'll do that in the relevant unit tests though, not in this particular suite.

core/src/actions/base.ts
edvald-garden: Added a commit to address this

core/test/unit/src/garden.ts
edvald-garden: Done

PR #7628, opened 2025-08-20
Experimental support for Cloud v2 variables

eysi09:

This enables Garden to fetch variables from the new cloud backend, currently under a feature flag. 

Note that the functionality is not GA yet in Garden Cloud, we're rolling this out both sides under feature flags.

Fixes #


edvald-garden:
Perusing this now, but also wanted to note that I've never really enjoyed all the `cloudApi?: CloudApi` optionals in the code. Would be much more sensible if there was a `DummyCloudApi` class or something. Might be a hassle to change at this point though.

edvald-garden (approved):

PR #7653, opened 2025-08-27
feat(config): add project-level excludeValuesFromActionVersions field

edvald-garden:
This simplifies action version exclusion for larger projects. Effectively this allows adding a set of values to every action's `version.excludeValues`, which can make filtering out ephemeral values much simpler across many actions.

From the Project `excludeValuesFromActionVersions` reference:

---
A list of string values that should be excluded when computing action versions.

Setting values here is equivalent to adding them to the \`version.excludeValues\` field on all actions in the project.

These values can be templated, and generally should be templated. A typical example is to exclude the namespace of the environment, or a hostname suffix used across many Deploy actions. For example:

```yaml
excludeValuesFromActionVersions:
  - "${var.hostname-suffix}"  # resolving to something like "my-branch.dev.my-org.com"
```

**Important:**
You should be careful to not make these values too broad, since the strings will be replaced for every field in all actions across the project when computing versions. For example, if a value here resolves to a short and generic string like "api", the string "api" will be replaced for every field in all actions across the project when computing versions. This could lead to unexpected issues like tests getting skipped when they shouldn't, deployments not updating etc.

However, something more specific like a branch name, commit hash, PR number etc., ideally with some specific prefix or suffix, is generally safer to do. That said, this field only affects version computation, not the actual action configuration when it's executed.

---

### DIFF
--- a/core/src/garden.ts
+++ b/core/src/garden.ts
@@ -1306,7 +1306,7 @@ export class Garden {
     })
   }
 
-  // @Memoize(() => true)
+  @pMemoizeDecorator()
   async getExcludeValuesForActionVersions(): Promise<string[]> {
     const context = new ExcludeValuesFromActionVersionsContext(this, this.variables)
     const source = { yamlDoc: this.projectConfig.internal.yamlDoc, path: ["excludeValuesFromActionVersions"] }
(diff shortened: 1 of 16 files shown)


core/src/garden.ts
```diff
@@ -1301,9 +1302,22 @@ export class Garden {
       moduleGraph: graph.moduleGraph,
       // TODO: perhaps groups should be resolved here
       groups: graph.getGroups(),
+      excludeValuesFromActionVersions: await this.getExcludeValuesForActionVersions(),
     })
   }
 
+  // @Memoize(() => true)
```
eysi09, line 1309: Should this be commented out?

eysi09 (commented):
Looks good. Only had one comment.

PR #7658, opened 2025-09-01
feat(config): add version.excludeDependencies to action config

edvald-garden:
This allows further control over how action versions are calculated, by excluding certain dependencies of the action from affecting the version.

From the added docs:

---

### `version.excludeDependencies[]`

Specify a list of dependencies that should be ignored when computing the version hash for this action.

Generally, the versions of all dependencies (both implicit and explicitly specified) are used when computing the version hash for this action. However, there are cases where you might want to exclude certain dependencies from the version hash.

For example, you might have a dependency that naturally changes for every individual test or dev environment, such as a setup script that runs before the test. You could solve for that with something like this:

```yaml
version:
  excludeDependencies:
    - run.setup
```

Where `run.setup` refers to a Run action named `setup`. You can also use the full action reference for each dependency to exclude, e.g. `{ kind: "Run", name: "setup" }`.

---

### DIFF
--- a/core/src/actions/base.ts
+++ b/core/src/actions/base.ts
@@ -261,6 +261,27 @@ export const baseActionConfigSchema = createSchema({
     version: joi
       .object()
       .keys({
+        excludeDependencies: joi
+          .sparseArray()
+          .items(joi.actionReference())
+          .description(
+            dedent`
+            Specify a list of dependencies that should be ignored when computing the version hash for this action.
+
+            Generally, the versions of all dependencies (both implicit and explicitly specified) are used when computing the version hash for this action.
+            However, there are cases where you might want to exclude certain dependencies from the version hash.
+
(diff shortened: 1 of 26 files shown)


core/src/actions/base.ts
```diff
@@ -261,6 +261,27 @@ export const baseActionConfigSchema = createSchema({
...
+
+            For example, you might have a dependency that naturally changes for every individual test or dev environment, such as a setup script that runs before the test. You could solve for that with something like this:
+
+            \`\`\`yaml
+            version:
+              excludeDependencies:
+                - run.setup
+            \`\`\`
+
+            Where \`run.setup\` refers to a Run action named \`setup\`. You can also use the full action reference for each dependency to exclude, e.g. \`{ kind: "Run", name: "setup" }\`.
```
eysi09, line 282: Should we show the YAML version here?

eysi09 (approved):
One minor comment, approving regardless.

PR #7730, opened 2025-09-23
feat: allow module-based services+tasks to depend on Deploy+Run actions

edvald-garden:
This helps mitigate some struggles with gradual config migrations. The caveat is that actions cannot be referenced in template strings in modules at all, and this only applies across module-based services and tasks, and Deploy and Run actions.

### DIFF
--- a/core/src/config/common.ts
+++ b/core/src/config/common.ts
@@ -192,6 +192,8 @@ interface ActionReferenceSchema extends Joi.AnySchema {
   kind(kind: ActionKind): this
 
   name(type: string): this
+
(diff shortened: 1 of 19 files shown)


eysi09 (changes requested):
Nice!

Wondering if we can add this to the docs somewhere. We have this: https://docs.garden.io/misc/migrating-to-bonsai.

Perhaps we can start by just noting that the on 0.14 modules _can_ actually depend on actions in some cases. 

And later we can add a dedicated Migrating from actions guide since it's not really version specific.

PR #7794, opened 2025-10-29
chore(CI): update Rust & wine setup

thsig:

The `build-dist-macos` and `build-dist-linux-windows` CirleCI jobs were failing with:

```
feature `edition2024` is required
The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.83.0
(5ffbef321 2024-10-29))
```

This PR bumps Rust to version 1.88.0 to fix that.

After updating Rust in CI, we got errors from wine indicating problems with 32-bit binaries and the lack of a window server. This is also fixed here.

### DIFF
--- a/.circleci/continue-config.yml
+++ b/.circleci/continue-config.yml
@@ -290,8 +290,8 @@ commands:
 
             # Install the latest stable Rust version
             # We maintain it manually to avoid unexpected breaking changes like we faced in Rust 1.78 for Windows builds.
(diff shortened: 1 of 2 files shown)


.circleci/continue-config.yml
```diff
@@ -282,6 +282,17 @@ commands:
       - run:
           name: "Install Rust and Cross"
           command: |
+            if [[ "$(uname -s)" == "Linux" ]]; then
```
eysi09: Should we skip this if we're not using Wine anyway?

PR #7797, opened 2025-10-30
fix(k8s): fix fail-fast Helm deploys

thsig:

This fixes an issue with our fail-fast Helm deploy logic, where if a previously unhealthy workload belonging to the chart was already deployed, that unhealthy state would be treated as a deploy failure before the new version of the resource had time to be committed to the API server and processed by the controller.

This was essentially a race condition because of the way we started waiting for resources immediately after we invoked the Helm CLI (whereas other code locations where we use `waitForResources` await at least the API operations going through).

The fix involved waiting for resources whose specs could be expected to be changed by the Helm operation. To make this more solid we (with help from our LLM friends) made the spec comparison code much more thorough by attempting to exhaustively address default fields etc. across most built-in resource types.

### DIFF
--- a/core/src/plugins/kubernetes/helm/deployment.ts
+++ b/core/src/plugins/kubernetes/helm/deployment.ts
@@ -6,7 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import { checkResourceStatuses, waitForResources } from "../status/status.js"
+import { checkResourceStatuses, specChanged, waitForResources } from "../status/status.js"
 import { helm } from "./helm-cli.js"
 import type { HelmGardenMetadataConfigMapData } from "./common.js"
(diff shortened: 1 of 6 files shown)


core/src/plugins/kubernetes/helm/deployment.ts
```diff
@@ -135,49 +136,34 @@ export const helmDeploy: DeployActionHandler<"deploy", HelmDeployAction> = async
   // We never fail fast with --atomic
   const failFast = spec.atomic === false && spec.waitForUnhealthyResources === false
   let wrappedInstallError: unknown | null = null
-  // This is basically an internal field that's only used for testing. Couldn't think of a better approach -E
+  // This is basically an internal field that's only used for testing. ouldn't think of a better approach -E
```
edvald-garden: Haha

PR #7843, opened 2025-11-25
improvement(cloud): detect v2 backend for migrated projects

edvald-garden:


core/src/cloud/util.ts
```diff
@@ -55,25 +56,62 @@ export function getCloudLogSectionName(distroName: CloudDistroName): CloudLogSec
...
+        if (response.statusCode >= 300 && response.statusCode < 400) {
+          const location = response.headers.location
+          if (location) {
+            const redirectUrl = new URL(location, domainOrigin)
+            if (redirectUrl.origin === DEFAULT_GARDEN_CLOUD_DOMAIN) {
+              return DEFAULT_GARDEN_CLOUD_DOMAIN
+            }
+          }
+        }
+      } catch (error) {
```
eysi09: Should we retry here before falling back? (Or does `got` already do that?).

